PR #3, opened 2020-06-03
README improvements

thecadams:


thecadams:
@mathieud @antoineeudes PTAL - was hesitant to just push this to master, since you might not see it

antoineeudes:
> @mathieud @antoineeudes PTAL - was hesitant to just push this to master, since you might not see it

That's great thanks!

README.md
```diff
@@ -14,6 +14,23 @@ npm start
 npm run build
 ```
 
+## Deploying
+
+tl;dr:
+```
+npm i -g firebase-tools # only need this once
```
mathieud, line 21: Any reason to not have this as a dev dependency?

thecadams: `-g` installs it globally so it’s in the path.

PR #6, opened 2020-06-03
Feature/register form

mathieud:


src/pages/Account/Account.scss
```diff
@@ -0,0 +1,252 @@
...
+
+body {
+  font: 1rem/1.3 "Rubik", "HelveticaNeue", "Helvetica Neue", Helvetica,
+    sans-serif;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
+  background-color: $black;
+  color: white;
+  padding: 0;
+}
```
antoineeudes, line 27: shouldn't we write that in `global.scss`?

mathieud: I copy pasted Mathieu's stylesheet. You're right we need to figure out what's global and what only applies to the account pages.

src/pages/Account/Register.tsx
```diff
@@ -0,0 +1,84 @@
...
+          </div>
+          <div className="input-group">
+            <input
+              name="password"
+              className="input-block input-centered"
+              type="password"
+              placeholder="Password"
+              ref={register({
+                required: true,
+                // minLength: 8,
```
antoineeudes, line 45: 🌵

PR #13, opened 2020-06-04
:sparkles: add users profile avatars on logged in party page

antoineeudes:
Screen Shot 2020-06-04 at 14 31 08


src/components/molecules/UserList/UserList.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface User {
+  id: string;
+  gratefulFor?: string;
+  islandCompanion?: string;
+  likeAboutParties?: string;
+  partyReal?: boolean;
+  respectParty?: boolean;
+  seekFun?: true;
+  username?: string;
```
mathieud, line 12: username is now partyname

src/components/molecules/UserList/UserList.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import "./UserList.scss";
+
+interface User {
+  id: string;
+  gratefulFor?: string;
```
mathieud, line 6: We shouldn't have all these fields here I think. They're not relevant.

PR #17, opened 2020-06-04
Config upload

thecadams:
AA Party Organiser, I can upload the party config with a CLI tool

First commit adds json schema and filename for tool


thecadams:
@antoineeudes @mathieud PTAL

thecadams:
This PR adds the ability to upload the contents of a JS file to firestore as a doc with an ID of the party name, which is then specified in `src/config.js`.

In future, a possible plan is to support multiple parties with this platform via some sort of mapping of environment/domain to party name, so they can share a firestore. This is laying ground work for that future.

It also means a new config can be uploaded (if performers change, or performers get weird about the wording of their program entry) and it will refresh for all users automatically via firestore snapshot, without any page reloads.

thecadams:
Example config: https://console.firebase.google.com/u/0/project/co-reality-map/database/firestore/data~2Fconfig~2Fexample

Doc for the config we will use for the party: https://console.firebase.google.com/u/0/project/co-reality-map/database/firestore/data~2Fconfig~2Fco-reality-5

JSON Schema generator/tester: https://jsonschema.net/home

PR #22, opened 2020-06-05
:sparkles: add room card

antoineeudes:


src/RoomAttendance.tsx
```diff
@@ -3,29 +3,38 @@ import React from "react";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faUser } from "@fortawesome/free-solid-svg-icons";
 
-export default function RoomAttendance({
+interface PropsType {
+  room: any;
```
mathieud, line 7: You've defined Room [here](https://github.com/co-reality/co-reality-map/pull/22/files#diff-a42781d2f5e0c14bbd82846a50608960R10)

PR #36, opened 2020-06-09
:sparkles: add EditProfilePage

antoineeudes:
Screen Shot 2020-06-09 at 10 01 18

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,7 @@ import "firebase/auth";
 import "firebase/functions";
 import { ReactReduxFirebaseProvider } from "react-redux-firebase";
 import { composeWithDevTools } from "redux-devtools-extension/developmentOnly";
+import WithLoadedUsers from "components/organisms/WithLoadedUsers";
 
 import "bootstrap";
 import "scss/global.scss";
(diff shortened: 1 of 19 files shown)


src/pages/EditProfilePage/EditProfilePage.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const logout = () => {
+    firebase.auth().signOut();
+    history.push("/");
+  };
+
+  return user && users ? (
+    <>
+      <NavBar />
+      <div className="container edit-profile-page-container">
```
mathieud, line 28: Is this different from the modal showing a users profile?

src/index.js
```diff
@@ -78,12 +80,16 @@ render(
     <ReactReduxFirebaseProvider {...rrfProps}>
       <Router>
         <Switch>
+          <Route
+            path="/account/edit"
+            component={withLoadedUsers(EditProfilePage)}
```
mathieud, line 85: I'd prefer a component wrapping the routes here. Would look like
```
<Switch>
  ...
  <LoadedUsers>
    <Switch>
      <Route path="/account/edit" />
      <Route path="/" component={withLoadedUsers(App)} />
    </Switch>
  </LoadedUsers>
</Switch>
```

That way we have sections clearly defined in the routing file.

antoineeudes: great idea, I totally approve

PR #49, opened 2020-06-11
Backstage v2

mathieud:


thecadams:
Looks like this one is not needed any more, are we OK to close it?

mathieud:
Yep

PR #62, opened 2020-06-12
:sparkles: new design for the chat

antoineeudes:


firestore.rules
```diff
@@ -2,17 +2,27 @@ rules_version = '2';
...
-        allow write: if request.auth.uid != null;
-        allow read: if true;
+      allow write: if request.auth.uid != null;
+      allow read: if true;
     }
     match /users/{userId} {
-        allow write, read: if request.auth.uid != null;
+      allow write, read: if request.auth.uid != null;
     }
     match /chatsv2/{chatId} {
```
mathieud, line 11: This can go I think

src/actions/index.js
```diff
@@ -19,27 +19,68 @@ function sendRoom(room, uid) {
...
+        to,
+      });
+  };
+}
+
+export function sendTableChat(from, to, text) {
+  return (dispatch) => {
+    const firestore = firebase.firestore();
+    firestore.collection("chatsv3").add({
+      type: "table",
```
mathieud, line 79: I would have kept group and table the same thing. Doesn't matter that much.

src/components/molecules/ChatMessage/ChatMessage.tsx
```diff
@@ -0,0 +1,81 @@
+import React from "react";
+import "./ChatMessage.scss";
+
+interface ChatMessageType {
+  id: string;
+  from: string;
+  text: string;
+  to?: string;
+  room?: string;
+  table?: string;
```
mathieud, line 10: Do these fields really exist? Can't see them in the creation functions.

PR #74, opened 2020-06-16
[WIP] make active user count more accurate

timifasubaa:
Title 
COde currently does not work but over review I will update things


thecadams:
Sounds good. Let's work through reviewing on a zoom call - there might be better approaches that will produce more accurate results!

thecadams:
@timi Looks like this one is not needed any more, are we OK to close it?

PR #76, opened 2020-06-16
:sparkles: by default messages are adressed to the room when a user i…

antoineeudes:


src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -91,31 +127,49 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
...
                   </Dropdown.Menu>
                 </Dropdown>
               </div>
             )}
 
             <ChatForm
-              type={isPrivate ? "private" : chatboxMessageType}
-              discussionPartner={discussionPartner}
+              type={
+                isPrivate || privateRecipient ? "private" : chatboxMessageType
```
mathieud, line 168: Do we have 2 types of private messages?

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -65,11 +77,35 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
...
 
+  let privateRecipient;
+  const lastChat = chatsToDisplay && chatsToDisplay[0];
+  if (
+    !isPrivate &&
+    !hasUserSelectedRecipient &&
+    lastChat &&
+    lastChat.type === "private"
+  ) {
+    if (lastChat?.to === user.uid) {
```
mathieud, line 93: I don't know how active the chat is, but it shouldn't change who you're sending it to while you're writing something.

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -50,26 +56,64 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
...
   }
 
+  const changeChatboxMessageType = (type: string) => {
+    setIsRecipientChangeBlocked(true);
+    setChatboxMessageType(type);
+    setPrivateRecipient(undefined);
+  };
+
+  useEffect(() => {
+    console.log(isRecipientChangeBlocked);
```
mathieud, line 93: remove this

src/components/organisms/Chatbox/Chatbox.tsx
thecadams: Seems like a pretty reasonable point, lets raise this point to chat with Ed and I in TR

thecadams: Ah just noticed this got addressed as I kept reading, please disregard! Great work!

PR #77, opened 2020-06-17
:bugfix: disconnect from video chat when leaving a table

antoineeudes:


thecadams:
Great debugging - glad this one is solved.

PR #79, opened 2020-06-17
Generate types from json schema

mathieud:


thecadams:
@mathieud Looks like this is not needed anymore - we OK to close?

thecadams:
@mathieud @sintinilorenzo @maxb94 any objections to closing this?

PR #82, opened 2020-06-18
Feature/104/correct attendance count

antoineeudes:


src/pages/LoggedInPartyPage/LoggedInPartyPage.jsx
```diff
@@ -22,7 +23,12 @@ const LoggedInPartyPage = () => {
 
   const attendances = users
     ? users.reduce((acc, value) => {
-        acc[value.room] = (acc[value.room] || 0) + 1;
+        if (
+          value.room_entrance_ts_utc &&
+          wasLessThanThreeHoursAgo(value.room_entrance_ts_utc.seconds)
```
mathieud, line 28: Can we do this at the firestore level? (Fetch only the users that match this condition)

antoineeudes: can't filter data like that with firebase

PR #93, opened 2020-06-22
Feature/141/timestamp on chats

antoineeudes:


src/components/molecules/ChatMessage/ChatMessage.scss
```diff
@@ -8,16 +8,29 @@
...
+    width: 100%;
+    margin-top: 2px;
+
+    &.right-side {
+      text-align: right;
+    }
+  }
+}
+
+.message-buble {
```
DotGav, line 29: Should this be bubble?

PR #97, opened 2020-06-24
Weekend party fixes

thecadams:
A little delayed because I accidentally amended a commit that had been merged, here are the fixes made before and during the party.


thecadams:
48ded1e was the commit I accidentally amended.

PR #98, opened 2020-06-24
Fix video flickering - prevent Participant rerender

mathieud:


src/components/organisms/Room/Participant.js
```diff
@@ -76,34 +76,31 @@ const Participant = ({ participant, children }) => {
...
     }
   }, [participant, muted, audioTracks]);
 
   return (
-    <div
-      className={`col participant ${participant.bartender ? "bartender" : ""}`}
-    >
+    <div className={`col participant ${bartender ? "bartender" : ""}`}>
       <video ref={videoRef} autoPlay={true} />
+      {/* <Video videoRef={videoRef} /> */}
```
antoineeudes, line 87: est-ce volontaire?

mathieud: Nope, was a test. I'll remove it now.

PR #121, opened 2020-06-30
Feature/204/text message reactions

antoineeudes:


src/pages/ReactionPage/ReactionPage.tsx
```diff
@@ -63,6 +62,7 @@ const ReactionPage = () => {
                     width={50}
                     height={50}
                   />
+                  {/* @ts-ignore there is a text for messages to band */}
```
mathieud, line 65: I think if you remove the (wrong) type [here](https://github.com/sparkletown/sparkle/pull/121/files#diff-7a3b067fec7fa24e666613c91d177246R45) you won't need the is-ignore

src/pages/ReactionPage/ReactionPage.tsx
```diff
@@ -18,32 +21,31 @@ const ReactionPage = () => {
...
     <WithNavigationBar>
       <div className="full-page-container experience-container reaction-page-container">
-        <h1 className="title">Reactions to the live</h1>
+        <h1 className="title">Audience Reactions</h1>
         <div className="row">
           <div className="col-8">
             {usersById &&
               messagesToTheBand &&
-              messagesToTheBand.map((message: Reaction) => (
+              messagesToTheBand.map((message: MessageToTheBandReaction) => (
```
mathieud, line 48: You can remove this type, should be inferred.

PR #122, opened 2020-07-01
Feature/205/kansas smittys config

antoineeudes:


src/components/context/ExperienceContext.tsx
```diff
@@ -109,19 +110,19 @@ export default ({
...
-        .collection(`experiences/${experienceName}/reactions`)
+        .collection(`experiences/${venueName}/reactions`)
         .add(newReaction);
     },
-    [firebase, experienceName]
+    [firebase, venueName]
   );
 
-  const store = { reactions, addReaction };
+  const store = { reactions, addReaction, venueName };
```
mathieud, line 125: Why not add a name to the Venue, that's the same value as the id? I don't think we should have this information split.

src/components/venues/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -344,7 +345,7 @@ const Jazz: React.FunctionComponent<PropsType> = ({ setUserList }) => {
               width="100%"
               height="100%"
               className="youtube-video"
-              src="https://www.youtube.com/embed/dqZAA8ZIAVE?autoplay=1"
+              src={venue.iframeUrl ? `${venue.iframeUrl}?autoplay=1` : ""}
```
mathieud, line 348: `iframeUrl` should not be optional

src/components/venues/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -406,11 +407,13 @@ const Jazz: React.FunctionComponent<PropsType> = ({ setUserList }) => {
                 isVideoFocused ? "col-5" : "col-12"
               } table-container`}
             >
-              <TableHeader
-                seatedAtTable={seatedAtTable}
-                setSeatedAtTable={setSeatedAtTable}
-                experienceName={experience.associatedRoom}
-              />
+              {experienceContext && (
```
mathieud, line 410: Why do we need to check this all the time now? Can't see what changed.

src/components/context/ExperienceContext.tsx
thecadams: Yep and we can use the document ID as the venue name. That will ensure it’s unique

thecadams: Yep and we can use the document ID as the venue name. That will ensure it’s unique

PR #133, opened 2020-07-02
:sparkles: email form on marketing page works

antoineeudes:


thecadams:
I suspect this is no longer applicable since the email form now works in master.

PR #146, opened 2020-07-07
:sparkles: add eventListener

antoineeudes:


src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -45,6 +48,18 @@ const AppRouter = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [config]);
 
+  const onClickWindow = (event) => {
+    event.target.id &&
+      analytics.logEvent("clickonbutton", event.target.id, user.uid);
```
antoineeudes, line 53: I don't see the event coming in the firebase console

PR #156, opened 2020-07-09
Feature/261/tweaks on chat modal

antoineeudes:


src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
```diff
@@ -30,6 +31,10 @@ const EntranceExperience = () => {
     return <>Loading...</>;
   }
 
+  if (user) {
+    return <Redirect to={`/venue/${venueId}`} />;
```
mathieud, line 35: Any chance we can get that from the router?

PR #186, opened 2020-07-21
Feature/payment on production

antoineeudes:


src/components/molecules/EventPaymentButton/EventPaymentButton.tsx
```diff
@@ -15,27 +19,51 @@ const EventPaymentButton: React.FunctionComponent<PropsType> = ({
   const [isLoading, setIsLoading] = useState(false);
   const [stripeError, setStripeError] = useState<string | undefined>();
 
+  useConnectUserPurchaseHistory();
```
DotGav, line 22: This and the has User already bought ticket the same as in the check payment is needed

PR #213, opened 2020-07-28
[PROD] :sparkles: filter events already finished on splash page

antoineeudes:


src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
```diff
@@ -74,16 +75,33 @@ const JazzbarEntranceExperience: React.FunctionComponent<PropsType> = ({
     purchaseHistoryRequestStatus: boolean;
   };
 
+  const futureOrOngoingVenueEvents =
```
maxb94, line 78: venueEvents?.filter(...

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
```diff
@@ -74,16 +75,33 @@ const JazzbarEntranceExperience: React.FunctionComponent<PropsType> = ({
...
 
   useEffect(() => {
-    if (user && venueEvents && venueId && eventId && redirectTo === "payment") {
-      setSelectedEvent(venueEvents.find((event) => event.id === eventId));
+    if (
+      user &&
+      futureOrOngoingVenueEvents &&
+      venueId &&
+      eventId &&
+      redirectTo === "payment"
```
maxb94, line 97: this is quite hard to parse. Could you create a variable with a nice name e.g. const eventRequiringPayment = ...

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
```diff
@@ -74,16 +75,33 @@ const JazzbarEntranceExperience: React.FunctionComponent<PropsType> = ({
...
+    ) {
+      setSelectedEvent(
+        futureOrOngoingVenueEvents.find((event) => event.id === eventId)
```
maxb94, line 100: what happens if this is undefined?

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
```diff
@@ -206,99 +224,111 @@ const JazzbarEntranceExperience: React.FunctionComponent<PropsType> = ({
...
+                                </div>
+                              ) : (
+                                <div className="price-container">
+                                  Individual tickets £{venueEvent.price / 100}
+                                  <br />
+                                  Group tickets £
+                                  {venueEvent.collective_price / 100}
+                                  {!user && (
+                                    <div className="login-invitation">
+                                      Already have a ticket?{" "}
```
maxb94, line 292: remove the {" "} and use {\`Already have a ticket? \`}

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
antoineeudes: this code isn't usefule anymore as we use modals for the authentication
Thanks!

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
thecadams: Annoyingly, the linter we are using (`npm run lint`) likes to do the {" "} so it can split JSX strings across lines. Whichever we go with it'd be great to immortalise it in a lint rule.

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
thecadams: aye. A possible mitigation is to change its type to something non-nullable to prevent this being a problem.

src/components/venues/Jazzbar/EntranceExperience/EntranceExperience.tsx
maxb94: ah that's annoying :( agree we should fix it at some point

PR #216, opened 2020-07-28
Feature/cd

mathieud:


scripts/init-env.sh
```diff
@@ -0,0 +1,13 @@
...
+ENV_FILE=".env"
+PREFIX=$1
+
+rm $ENV_FILE
+touch $ENV_FILE
+
+for VARIABLE in API_KEY APP_ID MEASUREMENT_ID BUCKET_URL PROJECT_ID STRIPE_PUBLISHABLE_KEY
+do
+  ENV_NAME="${PREFIX}${VARIABLE}"
+  echo "REACT_APP_${VARIABLE}="${!ENV_NAME} >> ${ENV_FILE}
```
antoineeudes, line 12: what does !ENV_NAME mean?

mathieud: It uses the value of ENV_NAME as the variable name, ie the the first iteration becomes
```
echo "REACT_APP_ API_KEY ="${STAGING_API_KEY} >> ${ENV_FILE}
```

PR #221, opened 2020-07-29
Fix: message error

maxb94:


src/components/venues/Jazzbar/components/MessageList.tsx
```diff
@@ -16,29 +16,19 @@ const MessageList: React.FC<MessageListProps> = ({ messages }) => {
...
-              height={profileImageSize}
-            />
-            <div className="message-bubble">{message.text}</div>
-          </div>
+          <>
+            {message.from in usersById && (
+              <Message
+                sender={usersById[message.from]}
+                message={message}
+                onClick={() => setSelectedUserProfile(usersById[message.from])}
```
mathieud, line 28: Not sure you can change this. The `id` is (was) used in the modal.

maxb94: sender prop gives user

PR #222, opened 2020-07-30
useKeyedSelector for adding object keys to values

maxb94:
We have a typing error where the User type has a id field despite this not always being the case. This is a WIP to fix it.


thecadams:
Looking interesting.

This gives me an idea that it'd be interesting to have a job we can run which validates the data in the DB and flags problems. Could come in handy detecting issues where staging data and prod data don't match, for example.

PR #225, opened 2020-07-30
Add prod deploy

mathieud:


.circleci/config.yml
```diff
@@ -30,10 +38,34 @@ jobs:
       - checkout
       - node/with-cache:
           steps:
-            - run: ./scripts/init-env.sh STAGING_
+            - run:
+                command: |
+                  if [ $CIRCLE_BRANCH = "staging" ]; then
```
maxb94, line 43: if the environment variables are preserved between commands, you should set PREFIX, ENV, and TARGET in one conditional

mathieud: I don't think they are.

maxb94: why not making it a single command then

mathieud: Confirmed (https://circleci.com/gh/sparkletown/sparkle/938?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

The environment setup is the same for the functions and hosting, that's why it's separated from the the actual deploy part. I can group them.

maxb94:
ok

PR #230, opened 2020-07-30
:sparkles: left column as drawer

antoineeudes:


src/scss/global.scss
```diff
@@ -385,21 +385,25 @@ h4 {
...
     overflow: auto;
-    height: 100%;
+    position: absolute;
+    top: 40px;
+    z-index: 1;
   }
 
   .left-column.expanded {
     width: 300px;
+    height: 450px;
```
mathieud, line 406: This should be allowed to grow based on the content size

src/components/organisms/InformationLeftColumn/InformationLeftColumn.scss
```diff
@@ -2,13 +2,21 @@
...
   flex: 0 0 auto;
+  position: relative;
 
   .band-logo {
     max-width: 100%;
     max-height: 100%;
 
     &.expanded {
-      width: 200px;
+      width: 100px;
```
mathieud, line 12: It's smaller when expanded?

src/components/venues/Jazzbar/JazzBarSkeletonPage/JazzBarSkeletonPage.tsx
```diff
@@ -19,7 +19,7 @@ const JazzBarSkeletonPage: React.FunctionComponent<PropsType> = ({
   return (
     <>
       <WithNavigationBar>
-        <div className="full-page-container experience-container">
+        <div className="full-page-container experience-container jazz-bar-container">
```
mathieud, line 22: Can we reduce (group) the classes here? Not sure we need all 3.

src/components/organisms/InformationLeftColumn/InformationLeftColumn.scss
antoineeudes: smaller than before but still bigger than when not expanded

src/scss/global.scss
antoineeudes: the div is scrollable. On the mock-ups, it doesn't take full height. Full height doesn't look good at all in my opinion

PR #248, opened 2020-08-01
Fix hang on page load

thecadams:


thecadams:
Merging without review, let's get a review done this week. Looks like adding a key prop to the list of messages was hanging the page on load because the collection differed every time through the map call.

PR #251, opened 2020-08-03
Refactor venues to templates

thecadams:
Trying to resolve some confusion about the distinction between a venue and a template by rationalising this a little


thecadams:
This needs a review, let's chat when the team is in

PR #253, opened 2020-08-03
407: Admin upload images

maxb94:


firebase.json
```diff
@@ -3,50 +3,45 @@
     "rules": "firestore.rules",
     "indexes": "firestore.indexes.json"
   },
+  "storage": {
+    "rules": "firestore.rules"
```
mathieud, line 7: If you can have a separate file, should we? Seems cleaner.

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+type FirestoreVenueInput = Omit<
+  VenueInput,
+  "bannerImageFile" | "logoImageFile"
+> & {
+  bannerImageUrl: string;
+  logoImageUrl: string;
+};
+
+export const createUrlSafeName = (name: string) => name.replace(/\W/g, "");
```
mathieud, line 21: Should we send this to the backend instead of recomputing it?

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const createVenue = async (input: VenueInput, user: UserInfo) => {
+  const storageRef = firebase.storage().ref();
+
+  const logoFile = input.logoImageFile[0];
+  const bannerFile = input.logoImageFile[0];
+
+  const urlVenueName = createUrlSafeName(input.name);
+  console.log("createVenue -> user", user);
```
mathieud, line 30: Remove

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // upload logo file
+  const uploadLogoRef = storageRef.child(
+    `users/${user.uid}/venues/${createUrlSafeName(urlVenueName)}/${
```
mathieud, line 34: You're calling `createUrlSafeName` the result already

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const logoDownloadUrl: string = await uploadLogoRef.getDownloadURL();
+  const bannerDownloadUrl: string = await uploadLogoRef.getDownloadURL();
+
+  const FirestoreVenueInput = {
+    ..._.omit(input, ["bannerImageFile", "logoImageFile"]),
+    bannerImageUrl: bannerDownloadUrl,
+    logoImageUrl: logoDownloadUrl,
+  };
+  console.log("FirestoreVenueInput", FirestoreVenueInput);
```
mathieud, line 56: remove

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+    `users/${user.uid}/venues/${createUrlSafeName(urlVenueName)}/${
+      bannerFile.name
+    }`
+  );
+  await uploadBannerRef.put(input.logoImageFile[0]);
+
+  const logoDownloadUrl: string = await uploadLogoRef.getDownloadURL();
+  const bannerDownloadUrl: string = await uploadLogoRef.getDownloadURL();
+
+  const FirestoreVenueInput = {
```
mathieud, line 51: camelCase

src/api/admin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // upload banner file
+  const uploadBannerRef = storageRef.child(
+    `users/${user.uid}/venues/${createUrlSafeName(urlVenueName)}/${
+      bannerFile.name
+    }`
+  );
+  await uploadBannerRef.put(input.logoImageFile[0]);
+
+  const logoDownloadUrl: string = await uploadLogoRef.getDownloadURL();
```
mathieud, line 48: Not worth changing it, but I prefer uploading the file immediately instead of on submit. Makes a nicer experience overall.

src/pages/Account/AdminVenue.tsx
```diff
@@ -1,62 +1,62 @@
...
     .test("logoImageUrl", "Image required", (val: FileList) => val.length > 0),
   tagline: Yup.string().required("Required"),
   longDescription: Yup.string().required("Required"),
 });
 
 export const AdminVenue: React.FC = () => {
-  const { register, handleSubmit, errors, watch } = useForm<
-    Partial<FormInputs> // bad typing. If not partial, react-hook-forms should force defaultValues to conform to FormInputs but it doesn't
+  const { register, handleSubmit, errors, watch, formState } = useForm<
+    Partial<VenueInput> // bad typing. If not partial, react-hook-forms should force defaultValues to conform to FormInputs but it doesn't
```
mathieud, line 36: Did you try passing a default for everything see if it gets rid of Partial? Not sure it's better but I'm curious

storage.rules
```diff
@@ -0,0 +1,8 @@
+rules_version = '2';
+service firebase.storage {
+  match /b/{bucket}/o {
+    match /users/{uid}/{allPaths=**} {
+        allow read, write, update: if request.auth.uid == uid
+    }
+  }
```
mathieud, line 7: Are we missing the rule for the venue?

firebase.json
mathieud: Actually looks like this should be `storage.rules`

maxb94: woops was a typo, i meant storage.rules

src/api/admin.ts
maxb94: don't think so. I think backend should compute document id's (though doesn't make much of a difference in this case).

src/api/admin.ts
maxb94: what do you mean?

src/pages/Account/AdminVenue.tsx
maxb94: yes. It does but it's not enforced by typescript. Also I'm not sure if it's possible to set an input to undefined in the html form but if it it would make defaults futile.

src/api/admin.ts
mathieud: const urlVenueName = createUrlSafeName(input.name); ([l29](https://github.com/sparkletown/sparkle/pull/253/files#diff-7b49daef5f76b8f0c9a29c54f8c5e376R29))
... 
... createUrlSafeName(urlVenueName)

storage.rules
maxb94: nope. storing the data in /users/{uid}/venues/{venuename}/...

There is no way of querying the firestore to ensure the user own the venueId so hypothetically any {venuename} could be used. This is not a security concern though as the venue in firestore stores the download url.

src/api/admin.ts
maxb94: ah yes.. fix incoming

PR #260, opened 2020-08-04
407: banner img fix

maxb94:


thecadams:
FYI I believe you can rebase a PR and force-push if needed and Github will pick that up.

PR #264, opened 2020-08-04
409: Admin creator wizard

maxb94:


src/settings.ts
```diff
@@ -1,2 +1,30 @@
...
+    name: "A Theme Camp",
+    description: ["Clickable map", "Embed zooms, venues, and so on"],
+  },
+  {
+    name: "Performance Venue",
+    description: ["Performance space", "Tables with video chat"],
+  },
+  {
+    name: "Art Piece",
+    description: ["Clickable map", "Embed zooms, venues, and so on"],
```
antoineeudes, line 39: not exactly the Art Piece venue but don't worry I'll change it

src/api/admin.ts
```diff
@@ -13,20 +13,31 @@ export interface EventInput {
   price: number;
 }
 
-export interface VenueInput {
+interface Question {
```
mathieud, line 16: Can we use the existing question type?

mathieud: Actually name is purely internal, don't think it should be part of the form (can slugify the text again)

src/settings.ts
maxb94: This copy comes from Ed's designs

src/pages/Account/Venue/AdvancedDetailsForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { AdvancedVenueInput } from "api/admin";
+
+interface AdvancedDetailsProps {
+  register: ReturnType<typeof useForm>["register"];
+  control: ReturnType<typeof useForm>["control"];
+  values: Partial<AdvancedVenueInput>;
+  errors: FieldErrors<Partial<AdvancedVenueInput>>;
+}
+
+export const AdvancedDetailsForm: React.FC<AdvancedDetailsProps> = (props) => {
```
mathieud, line 12: Should this be a template specific form (or will be?)

maxb94: not for the moment, but maybe in the future. If so, the template would need to be passed as a prop

PR #267, opened 2020-08-04
:sparkles: add art piece template

antoineeudes:


src/types/VenueTemplate.ts
```diff
@@ -2,4 +2,5 @@ export enum VenueTemplate {
   jazzbar = "jazzbar",
   friendship = "friendship",
   partymap = "partymap",
+  artPiece = "artpiece",
```
thecadams, line 5: Nit: might make sense to stick to lower case for consistency

PR #270, opened 2020-08-04
Venue dashboard - no design

mathieud:


src/pages/Account/Admin.tsx
```diff
@@ -5,61 +5,192 @@ import { useUser } from "hooks/useUser";
...
+  }
+
+  return (
+    <>
+      <ul>
+        {[
+          { url: `${match.url}`, label: "Venue Infos" },
+          { url: `${match.url}/appearance`, label: "Appearance" },
+          { url: `${match.url}/events`, label: "Events" },
+        ].map((l) => (
```
antoineeudes, line 81: I don't like this variable name. could we name it tab?

src/pages/Account/Admin.tsx
```diff
@@ -5,61 +5,192 @@ import { useUser } from "hooks/useUser";
...
+      where: [["owners", "array-contains", user?.uid]],
+    },
+  ]);
+
+  return (
+    <WithNavigationBar>
+      <div className="container admin-container">
+        <AuthenticationModal show={!user} onHide={() => {}} showAuth="login" />
+        <div className="title">Admin</div>
+        <div style={{ flex: "1", flexDirection: "row", display: "flex" }}>
```
antoineeudes, line 181: no need for `display: flex` with `flex: 1`, right?

src/pages/Account/Admin.tsx
```diff
@@ -5,61 +5,192 @@ import { useUser } from "hooks/useUser";
...
+          <div style={{ flex: "0 0 auto", border: "red solid 1px" }}>
```
antoineeudes, line 182: I would prefer to see all this style in a scss file tbh

mathieud: css is coming in another ticket. No need to create classes I'm deleting tomorrow.

src/pages/Account/Admin.tsx
mathieud: fixed

src/pages/Account/Admin.tsx
mathieud: Apparently it is necessary. Doesn't respect the row flow without. Going to disappear tomorrow when we add the styles.

PR #272, opened 2020-08-05
Add design

mathieud:


src/pages/Account/Admin.tsx
```diff
@@ -73,49 +77,84 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId }) => {
 
   return (
     <>
-      <ul>
+      <div className="page-container-adminpanel-tabs">
         {[
           { url: `${match.url}`, label: "Venue Infos" },
-          { url: `${match.url}/appearance`, label: "Appearance" },
+          // { url: `${match.url}/appearance`, label: "Appearance" },
```
antoineeudes, line 83: do we keep it for later?

src/pages/Account/Admin.tsx
```diff
@@ -73,49 +77,84 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId }) => {
...
           { url: `${match.url}/events`, label: "Events" },
         ].map((tab) => (
-          <li
+          <div
             key={tab.url}
-            style={location.pathname === tab.url ? { color: "red" } : {}}
+            className={`page-container-adminpanel-tab ${
+              location.pathname === tab.url && "selected"
```
antoineeudes, line 89: usually I write sth like `${condition ? ... : ""}` otherwise it writes `false` in the className.

src/pages/Account/Admin.tsx
```diff
@@ -27,19 +31,19 @@ const VenueList: React.FC<VenueListProps> = ({ selectedVenueId }) => {
...
     <>
-      <h4>My Venues</h4>
-      <ul>
+      <div className="page-container-adminsidebar-title">My Venues</div>
+      <ul className="page-container-adminsidebar-venueslist">
         {venues?.map((venue) => (
           <li
             key={venue.id}
-            style={venue.id === selectedVenueId ? { color: "red " } : {}}
+            className={`${selectedVenueId === venue.id ? "selected" : ""}`}
```
maxb94, line 39: dont need the template literals

src/pages/Account/Admin.tsx
```diff
@@ -73,49 +77,84 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId }) => {
...
-      </Link>
+      <div className="page-container-adminpanel-content">
+        <div className="container venue-entrance-experience-container">
+          <div
+            className="header"
+            style={{
+              background: `linear-gradient(
+            0deg,
+            rgba(0, 0, 0, 0.8) 2%,
+            rgba(0, 0, 0, 0) 98%
```
maxb94, line 128: prefer if this was defined in the scss file

src/pages/Account/Admin.tsx
```diff
@@ -136,14 +175,47 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
...
+                    </div>
+                    <div className="event-description">
+                      {venueEvent.description}
+                      {venueEvent.descriptions?.map((description, index) => (
+                        <p key={index}>{description}</p>
+                      ))}
+                    </div>
+                    <div className="button-container">
+                      <div className="price-container">
+                        Individual tickets £{venueEvent.price / 100}
```
maxb94, line 208: does this work without template string?

src/pages/Account/Admin.tsx
mathieud: Don't think we can, it uses the image from from the venue config (just below)

src/pages/Account/Admin.tsx
mathieud: Yes

src/pages/Account/Admin.tsx
mathieud: Yes

src/pages/Account/Admin.tsx
mathieud: We'll also have `camp` later, I think the template litteral makes sense.

PR #274, opened 2020-08-05
Feature/420/admin preview

antoineeudes:


src/components/templates/EntranceExperience/EntranceExperience.tsx
```diff
@@ -56,6 +57,8 @@ export const EntranceExperience: React.FunctionComponent<EntranceExperienceProps
       Date.now() / 1000
   );
 
+  const isUserVenueOwner = user && venue?.owners?.includes(user.uid);
```
maxb94, line 60: venue is defined

src/components/templates/EntranceExperience/EntranceExperience.tsx
```diff
@@ -270,6 +273,21 @@ export const EntranceExperience: React.FunctionComponent<EntranceExperienceProps
                   })}
                 </>
               )}
+            {isUserVenueOwner && (
+              <InformationCard title="Enter the venue as an admin">
```
maxb94, line 277: Not keen on this copy. Should be something like "Preview what an event looks like in your venue". Please confirm text with Mathieu Robert.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -53,6 +53,8 @@ const VenuePage = () => {
     currentTimestamp >
       event.start_utc_seconds + event.duration_minutes * ONE_MINUTE_IN_SECONDS;
 
+  const isUserVenueOwner = user && venue?.owners?.includes(user.uid);
```
maxb94, line 56: venue is defined

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -65,30 +67,32 @@ const VenuePage = () => {
...
-    );
+    if (!canUserJoinTheEvent(event)) {
+      return (
+        <CountDown
+          startUtcSeconds={event.start_utc_seconds}
+          textBeforeCountdown="Bar opens in"
+        />
+      );
+    }
   }
```
maxb94, line 97: would be nice to include a div at the top of the page saying something like "This is a preview of your venue during an event" in big bold letters. What do you think?

antoineeudes: good idea

PR #275, opened 2020-08-05
415: Venue page styling

maxb94:


src/api/admin.ts
```diff
@@ -40,7 +41,8 @@ type FirestoreVenueInput = Omit<
   mapIconImageUrl?: string;
 };
 
-export const createUrlSafeName = (name: string) => name.replace(/\W/g, "");
+export const createUrlSafeName = (name: string) =>
+  `${ROOT_URL}/v/${name.replace(/\W/g, "")}`;
```
mathieud, line 45: Why do we need this?

src/pages/Account/Venue/TemplateForm.tsx
```diff
@@ -73,6 +73,7 @@ const TemplateFormLeft: React.FC<TemplateFormLeftProps> = (props) => {
         </div>
       </div>
       <div className="page-container-left-bottombar">
+        <div />
```
mathieud, line 76: What does this do?

src/api/admin.ts
maxb94: for displaying the text 'The url of your venue will be sparkle.space/s/mylovelyvenue'. I noticed you are also using this function so will place the ROOT_URL in the caller.

src/pages/Account/Venue/TemplateForm.tsx
maxb94: 'justify-content: space-between' to ensure the button is to the right

src/api/admin.ts
mathieud: That's one more config variable we have to maintain when we could read it in the browser.

src/pages/Account/Venue/TemplateForm.tsx
mathieud: Can't we have `justify-content: end`?

maxb94: the next screen has a 'go back' button on the left hand side

src/api/admin.ts
maxb94: hmm could do that i suppose

src/pages/Account/Venue/DetailsForm.tsx
```diff
@@ -149,32 +155,36 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
     state.templatePage?.template.type === "ZOOM_ROOM" ||
     state.templatePage?.template.type === "PERFORMANCE_VENUE";
 
-  const urlSafeName = values.name ? createUrlSafeName(values.name) : undefined;
+  const urlSafeName = values.name
+    ? `/${window.location.host}/v/${createUrlSafeName(values.name)}`
```
mathieud, line 159: Does the initial slash work here?

PR #278, opened 2020-08-06
:lock: check existence of venue before creation

antoineeudes:


src/pages/Account/Venue/DetailsValidationSchema.ts
```diff
@@ -2,13 +2,26 @@ import * as Yup from "yup";
...
+        "This venue name is already taken",
+        async (val: string) =>
+          !(
+            await firebase
+              .firestore()
+              .collection("venues")
+              .doc(val.replace(/\W/g, ""))
+              .get()
+          ).exists
+      )
```
maxb94, line 24: nice 🤘

thecadams: quick gut check on this, is it case insensitive?

antoineeudes: the venue names are case sensitive, yes. It might be good to spend 30mins to make it case insensitive

antoineeudes: 5 mins are enough

PR #280, opened 2020-08-06
Redirect /SparkleVerse to new website

thecadams:


src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -57,7 +57,7 @@ const AppRouter = () => {
   return (
     <Router basename="/">
       <Switch>
-        <Route path="/SparkleVerse" component={SparkleSpaceMarketingPage} />
+        <Route path="/SparkleVerse" component={() => <Redirect to="https://sparklever.se/" />} />
```
maxb94, line 60: could we put the URL in a settings.ts? That way it can be referenced throughout the app

PR #281, opened 2020-08-07
Feature/admin venue theme

mathieud:


src/pages/Account/Admin.scss
```diff
@@ -95,9 +95,9 @@ $page-max-width: 1240px;
               &:after {
                 transform: rotate(180deg);
               }
-              ul.page-container-adminsidebar-subvenueslist {
-                display: block;
-              }
+              // ul.page-container-adminsidebar-subvenueslist {
+              //   display: block;
+              // }
```
maxb94, line 100: remove

mathieud: Thought we would use it if we decide to keep and improve the show/hide subvenues behaviour.

maxb94: ok

PR #282, opened 2020-08-07
venue slug is lowercase

antoineeudes:


src/pages/Account/Venue/DetailsValidationSchema.ts
```diff
@@ -18,14 +19,14 @@ export const validationSchema = Yup.object()
             await firebase
               .firestore()
               .collection("venues")
-              .doc(val.replace(/\W/g, ""))
+              .doc(createUrlSafeName(val))
```
maxb94, line 22: need to make sure lowercasing is done on the backend as well. In `venue.js`

PR #284, opened 2020-08-07
Assets and config for end of the universe party

thecadams:


thecadams:
Had to delete the `.title` in `Venue.scss` because it was interfering with the map template; we should scope this in a container related to the component.

PR #305, opened 2020-08-08
Refactor venue routing and preview

thecadams:
Things are getting a little unwieldy. Return to a simpler time when we just relied on components and types, after all isn't that what TypeScript was meant to be for?

Also take the landing page out of the template cos that was a mess too.


src/pages/Account/AdminEvent.tsx
```diff
@@ -67,29 +65,20 @@ const AdminEvent: React.FunctionComponent<PropsType> = ({
...
         start_time: dayjs.unix(event.start_utc_seconds).format("HH:mm"),
-        end_date: dayjs
-          .unix(event.start_utc_seconds)
-          .add(event.duration_minutes, "minute")
-          .format("YYYY-MM-DD"),
-        end_time: dayjs
-          .unix(event.start_utc_seconds)
-          .add(event.duration_minutes, "minute")
-          .format("HH:mm"),
-        price: event.price / 100,
```
mathieud, line 78: Did you remove price on purpose here?

src/pages/Account/AdminEvent.tsx
```diff
@@ -154,39 +143,18 @@ const AdminEvent: React.FunctionComponent<PropsType> = ({
             )}
           </div>
           <div className="input-group">
-            <label>End Time</label>
+            <label htmlFor="duration_hours">Duration (hours)</label>
             <input
-              type="date"
```
mathieud, line 159: Is this for burningman? I thought we were going to (re)create a Schedule, completely separate from Events, more like what we currently have on the PartyMap.

src/pages/Account/CodeOfConduct.tsx
```diff
@@ -43,7 +43,7 @@ const CodeOfConduct: React.FunctionComponent<PropsType> = ({ location }) => {
   const onSubmit = async (data: CodeOfConductFormData) => {
     if (!user) return;
     await updateUserProfile(user.uid, data);
-    history.push(venueId ? `/v/${venueId}/live` : "");
+    history.push(venueId ? `/in/${venueId}` : "");
```
mathieud, line 46: We should create a function for this, it's everywhere in the app.

src/pages/Account/AdminEvent.tsx
thecadams: Yep, not needed for burning man - we can re-add later. It will save with 0 as price for now.

PR #308, opened 2020-08-10
Fix/profile questions

mathieud:


thecadams:
What does adding `WithId` do to fix it?

mathieud:
> What does adding `WithId` do to fix it?

It doesn't. This is to better reflect the object in the app (the object we get from Firebase does not have an id attribute by default).

The actual fix is [this line](https://github.com/sparkletown/sparkle/pull/308/files#diff-593d1995ad8a87a161a89108f1ac2ca8R68), reading the information in the right place.

thecadams:
> The actual fix is [this line](https://github.com/sparkletown/sparkle/pull/308/files#diff-593d1995ad8a87a161a89108f1ac2ca8R68), reading the information in the right place.

Thanks for the explanation! Got it.

PR #320, opened 2020-08-12
441 + 473 tickets

maxb94:


src/settings.ts
```diff
@@ -71,4 +58,17 @@ export const ALL_VENUE_TEMPLATES: Array<Template> = [
...
     description: ["Create a jazzbar."],
   },
+
+  {
+    template: VenueTemplate.artcar,
+    name: "Art Car",
+    description: ["Create a space on the Playa that moves around."],
+  },
+  {
+    template: VenueTemplate.performancevenue,
```
thecadams, line 68: Performance venue currently just renders to a zoom room, so this copy should change at some point, but no prob

PR #324, opened 2020-08-13
Display other venues on the playa edit view

mathieud:
Looks like 
image

### DIFF
--- a/src/pages/Account/Venue/VenueMapEdition/Container.tsx
+++ b/src/pages/Account/Venue/VenueMapEdition/Container.tsx
@@ -120,19 +120,24 @@ export const Container: React.FC<PropsType> = (props) => {
           }}
           src={backgroundImage}
         />
-        {placedVenues?.map((v) => (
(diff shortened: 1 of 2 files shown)


src/pages/Account/Venue/VenueMapEdition/Container.tsx
```diff
@@ -77,6 +120,19 @@ export const Container: React.FC<PropsType> = (props) => {
...
+            key={v.id}
+            src={v.mapIconImageUrl || DEFAULT_MAP_ICON_URL}
+            style={{
+              position: "absolute",
+              top: (v.placement?.x || 0) * scale,
+              left: (v.placement?.y || 0) * scale,
+              width: "50px", // @debt should be at the right scale
+              opacity: 0.4,
+            }}
+          />
```
maxb94, line 134: memoise. You can do
```
return useMemo(() => <div>....</div>
```
instead of puling the map outside of the TSX

PR #326, opened 2020-08-13
Feature/481 loading icon

sintinilorenzo:
To see better test with a Fast/Slow 3G Network option in the developer console


src/components/molecules/LoadingPage/LoadingPage.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+import "./sparkle-styling/loading.scss";
```
thecadams, line 3: Can we remove the redundant subdirectory?

PR #332, opened 2020-08-14
ticket 486

sintinilorenzo:


src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -18,16 +18,24 @@ export const AdminVenuePreview: React.FC<AdminVenuePreview> = ({
...
-            <span className="content">{venue.zoomUrl}</span>
+            <span className="content">
+              <a href={venue.zoomUrl} target="_blank" rel="noopener noreferrer">
+                {venue.zoomUrl}
+              </a>
+            </span>
           </div>
         );
+      default:
+        return <div></div>;
```
mathieud, line 38: That's not an amazing default. Could we have `Sorry we can't show you this experience` And a button `G back to the playa` or something?

src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -59,7 +67,10 @@ export const AdminVenuePreview: React.FC<AdminVenuePreview> = ({
           <span className="title">Banner photo</span>
           <img
             className="banner"
-            src={venue.config.landingPageConfig.coverImageUrl}
+            src={
+              venue.config.landingPageConfig.bannerImageUrl ??
```
mathieud, line 71: Should we render the image conditionally so that nothing is displayed. Here we'd see the `image missing` artefact with the alt text.

src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -78,7 +89,7 @@ export const AdminVenuePreview: React.FC<AdminVenuePreview> = ({
           <span className="content">
             <img
               className="icon"
-              src={venue.mapIconImageUrl ?? venue.host.icon}
+              src={venue.host.icon ?? venue.mapIconImageUrl}
```
mathieud, line 92: Is this to accomodate for different types of venues?

src/pages/Admin/AdminVenuePreview.tsx
sintinilorenzo: this is just for the last line in the preview: if it's an ArtPiece there is a link to the iFrame, if it's a ZoomRoom there is a link to the zoom url, if it's neither of them (camp) it defaults to nothing (there is no line with extra information)

src/pages/Admin/AdminVenuePreview.tsx
sintinilorenzo: basically when the venue is created the banner goes to venue/config/landingPageConfig/checklist/coverImageUrl. However when it is updated it is stored in venue/config/landingPageConfig/bannerImageUrl.

So that line checks if an image exists in bannerImageUrl, if not it means it has not been updated and shows the one in coverImageUrl at the time of creation.

mathieud: That sounds like an error. @maxb94 ?

PR #333, opened 2020-08-14
Allow to delete a venue from the admin

mathieud:


firestore.rules
```diff
@@ -35,7 +35,7 @@ service cloud.firestore {
     }
     match /venues/{restOfPath=**} {
       allow read: if true;
-      allow write: if request.auth.uid != null;
+      allow write: if request.auth.uid in resource.data.owners;
```
thecadams, line 38: Thinking we should use a function with Firebase admin rather than letting the venue owner write any field, otherwise they could change the template etc.

src/pages/Admin/Admin.tsx
```diff
@@ -183,22 +184,69 @@ const VenueInfoComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
...
+                <button
+                  type="button"
+                  className="btn btn-primary btn-danger btn-block btn-centered"
+                  onClick={async () => {
+                    history.push("/admin");
+                    try {
+                      await deleteVenue(venue);
+                      // success toast
+                    } catch (e) {
+                      // error toast
```
sintinilorenzo, line 233: remove comments?

PR #343, opened 2020-08-17
youtube iframe works for art piece venues

sintinilorenzo:


src/components/templates/ArtPiece/ArtPiece.tsx
```diff
@@ -28,7 +28,7 @@ const ArtPiece = () => {
           <iframe
             className="youtube-video"
             title="art-piece-video"
-            src={venue.iframeUrl}
+            src={venue.embedIframeUrl?.replace("watch?v=", "embed/")}
```
mathieud, line 31: This is only going to work for youtube.

PR #350, opened 2020-08-17
enter the room button

sintinilorenzo:


src/components/templates/Camp/components/RoomModal.tsx
```diff
@@ -58,6 +58,15 @@ export const RoomModal: React.FC<PropsType> = ({ show, onHide, room }) => {
                 )}
                 {!room.image_url && room.title}
               </div>
+              <a
+                className="btn btn-primary join-button"
+                href={room.url}
+                target="_blank"
+                rel="noopener noreferrer"
+                style={{ position: "absolute", right: 10, top: 10 }}
```
mathieud, line 66: Can you use flexboxes instead to position it?

thecadams: Agreed; or else you can put it in a css class.

PR #354, opened 2020-08-18
Prevent chat from crashing the page

mathieud:
useEffect was in an infinite loop because of users changing all the time


thecadams:
Amazing find @mathieud - does this mean the chat now performs ok? Or will we still need a refactor to a backend based solution in the future?

mathieud:
@thecadams It now behaves ok, don't think we need to move to a backend based solution (could improve loading, pagination... but that could still all be done in the frontend)

thecadams:
Fantastic that’s great news!

PR #362, opened 2020-08-18
fix bug with form and order the events chronologically, not showing t…

sintinilorenzo:
…he ones that have passed

### DIFF
--- a/src/pages/Admin/Admin.tsx
+++ b/src/pages/Admin/Admin.tsx
@@ -277,15 +277,15 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
     },
   ]);
 
-  const events = useSelector((state) => state.firestore.ordered.events);
+  let events = useSelector((state) => state.firestore.ordered.events);
   const [showCreateEventModal, setShowCreateEventModal] = useState(false);
   const [showDeleteEventModal, setShowDeleteEventModal] = useState(false);
   const [editedEvent, setEditedEvent] = useState<WithId<VenueEvent>>();
   //filter out events that have already finished
-  // events = events?.filter(
-  //   (ev) =>
-  //     (ev.start_utc_seconds + ev.duration_minutes * 60) * 1000 > Date.now()
-  // );
+  events = events?.filter(
+    (ev) =>
+      (ev.start_utc_seconds + ev.duration_minutes * 60) * 1000 > Date.now()
+  );
 
   return (
     <>
--- a/src/pages/Admin/Admin.tsx
+++ b/src/pages/Admin/Admin.tsx
@@ -277,12 +277,12 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
     },
   ]);
 
-  let events = useSelector((state) => state.firestore.ordered.events);
+  const events = useSelector((state) => state.firestore.ordered.events);
   const [showCreateEventModal, setShowCreateEventModal] = useState(false);
   const [showDeleteEventModal, setShowDeleteEventModal] = useState(false);
   const [editedEvent, setEditedEvent] = useState<WithId<VenueEvent>>();
   //filter out events that have already finished
-  events = events?.filter(
+  const filteredEvents = events?.filter(
     (ev) =>
       (ev.start_utc_seconds + ev.duration_minutes * 60) * 1000 > Date.now()
   );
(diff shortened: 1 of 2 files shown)


src/pages/Admin/Admin.tsx
```diff
@@ -277,10 +277,16 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
...
   ]);
 
-  const events = useSelector((state) => state.firestore.ordered.events);
+  let events = useSelector((state) => state.firestore.ordered.events);
   const [showCreateEventModal, setShowCreateEventModal] = useState(false);
   const [showDeleteEventModal, setShowDeleteEventModal] = useState(false);
   const [editedEvent, setEditedEvent] = useState<WithId<VenueEvent>>();
+  events?.sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
+  //filter out events that have already finished
+  events = events?.filter(
```
mathieud, line 286: Did Ed ask you to filter out the past ones? Feels a bit weird for the admin.

src/pages/Admin/AdminEvent.tsx
```diff
@@ -85,7 +85,7 @@ const AdminEvent: React.FunctionComponent<PropsType> = ({
         price: 0,
         collective_price: 0,
         host: data.host,
-        room: data.room,
+        room: data.room ?? "",
```
mathieud, line 88: Can you not send the field at all instead?

src/pages/Admin/Admin.tsx
```diff
@@ -277,10 +277,16 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
     },
   ]);
 
-  const events = useSelector((state) => state.firestore.ordered.events);
+  let events = useSelector((state) => state.firestore.ordered.events);
   const [showCreateEventModal, setShowCreateEventModal] = useState(false);
   const [showDeleteEventModal, setShowDeleteEventModal] = useState(false);
   const [editedEvent, setEditedEvent] = useState<WithId<VenueEvent>>();
+  events?.sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
```
mathieud, line 284: This can be done at the firebase level, would be better.

mathieud: It seems to be the case already: https://github.com/sparkletown/sparkle/blob/staging/src/pages/Admin/Admin.tsx#L275

Are you sure there's an issue with the ordering?

src/pages/Admin/Admin.tsx
sintinilorenzo: No. I thought it would be good to avoid having a long list of past events that already happened at the top. I have commented it out for now and asked Ed.

src/pages/Admin/AdminEvent.tsx
sintinilorenzo: Yes okay

src/pages/Admin/Admin.tsx
sintinilorenzo: Didn't see that, kept the at the firebase level (has to be ascending instead of descending tho)

src/pages/Admin/Admin.tsx
sintinilorenzo: Yeah he said to just keep the upcoming ones

src/pages/Admin/Admin.tsx
```diff
@@ -272,15 +272,20 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
...
       doc: venue.id,
       subcollections: [{ collection: "events" }],
-      orderBy: ["start_utc_seconds", "desc"],
+      orderBy: ["start_utc_seconds", "asc"],
       storeAs: "events",
     },
   ]);
 
-  const events = useSelector((state) => state.firestore.ordered.events);
+  let events = useSelector((state) => state.firestore.ordered.events);
```
maxb94, line 280: we should never mutate redux state

src/pages/Admin/Admin.tsx
```diff
@@ -272,15 +272,20 @@ const EventsComponent: React.FC<VenueDetailsPartProps> = ({ venue }) => {
...
   const [showCreateEventModal, setShowCreateEventModal] = useState(false);
   const [showDeleteEventModal, setShowDeleteEventModal] = useState(false);
   const [editedEvent, setEditedEvent] = useState<WithId<VenueEvent>>();
+  //filter out events that have already finished
+  events = events?.filter(
```
maxb94, line 285: const `filteredEvents = events..`

src/pages/Admin/AdminEvent.tsx
```diff
@@ -85,16 +85,16 @@ const AdminEvent: React.FunctionComponent<PropsType> = ({
         price: 0,
         collective_price: 0,
         host: data.host,
```
maxb94, line 87: room: template === 'themecamp' ? data.room : undefined

PR #367, opened 2020-08-19
523: Template type in create button

maxb94:


src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -488,11 +492,13 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
 interface SubmitButtonProps {
   isSubmitting: boolean;
   editing?: boolean;
+  templateType?: string;
```
mathieud, line 495: Should it be optional? Maybe default the text to Venue then.

mathieud (approved):

PR #370, opened 2020-08-19
changed requirement for open venue + fixed positioning error in navbar

sintinilorenzo:


functions/stats.js
```diff
@@ -31,22 +31,30 @@ exports.getOnlineStats = functions.https.onCall(async (data, context) => {
...
             }
           });
-          if (venueOpen) {
-            const venueWithId = venue.data();
-            venueWithId.id = venue.id;
-            openVenues.push(venueWithId);
-          }
-        });
+      }
+      if (template === "artpiece" || template === "themecamp") {
```
mathieud, line 53: I would change this to be an `else` clause to act as a default for all the other cases.

mathieud (approved):

PR #371, opened 2020-08-19
Feauture/interactive map

thecadams:


thecadams:
@maxb94 PTAL

src/components/templates/Playa/Playa.tsx
```diff
@@ -1,47 +1,539 @@
...
+  const recenter = useMemo(
+    () => () => {
+      if (myX === undefined || myY === undefined) return;
+      setCenterX(myX);
+      setCenterY(myY);
+    },
+    [myX, myY]
+  );
+
+  const getNearbyVenue = useMemo(
```
maxb94, line 329: this is what useCallback is for :)

src/components/templates/Playa/AvatarLayer.tsx
```diff
@@ -0,0 +1,163 @@
...
+          for (const uid of Object.keys(update.updates)) {
+            if (uid === user.uid) {
+              setMyServerSentState((myServerSentState) =>
+                myServerSentState ? myServerSentState : update.updates[uid]
+              );
+            } else {
+              userStateMapRef.current[uid] = update.updates[uid];
+            }
+          }
+          setUserStateMap(userStateMapRef.current);
```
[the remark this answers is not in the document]
thecadams: yep, is that something I can trigger more effectively somehow?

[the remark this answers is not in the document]
maxb94: yes. The object ref for userStateMapRef.current remains the same, so you are actually never triggering a rerender here. I would suggest getting rid of userStateMapRef and placing this logic inside the setUserStateMap callback a la
```
setUserStateMap(prev => {

// if change, copy prev, add changes then return
const newMap = {...prev}
//do stuff to new
return newMap

// if no change, return prev which does not cause rerender
return prev
}

File: src/components/templates/Playa/AvatarLayer.tsx [hunk 3]
```diff
@@ -44,15 +44,19 @@ const AvatarLayer: React.FunctionComponent<PropsType> = ({
...
-        update: state,
-      };
-      wsRef.current.send(JSON.stringify(update));
+      if (!user) return;
       setMyLocation(state.x, state.y);
+
+      if (wsRef.current) {
+        const update: UpdateWsMessage = {
+          type: MessageType.Update,
+          uid: user?.uid,
```
maxb94, line 53: user is well defined

src/components/templates/Playa/MyAvatar.tsx
```diff
@@ -55,6 +53,15 @@ export const MyAvatar: React.FunctionComponent<PropsType> = ({
...
           let needsUpdate = false;
+          // Work around possible bad state that can happen in the presence of scroll jank
+          if (pressedKeys["ArrowLeft"] && pressedKeys["ArrowRight"]) {
+            pressedKeys["ArrowLeft"] = false;
+            pressedKeys["ArrowRight"] = false;
+          }
+          if (pressedKeys["ArrowUp"] && pressedKeys["ArrowDown"]) {
+            pressedKeys["ArrowUp"] = false;
+            pressedKeys["ArrowDown"] = false;
+          }
```
maxb94, line 64: why is this necessary?

src/components/templates/Playa/MyAvatar.tsx
```diff
@@ -0,0 +1,116 @@
...
+          return;
+      }
+      const moveIncrement = walkMode
+        ? ARROW_MOVE_INCREMENT_PX_WALK
+        : ARROW_MOVE_INCREMENT_PX_BIKE;
+      setState((state) => {
+        if (state) {
+          let needsUpdate = false;
+          if (pressedKeys["ArrowLeft"]) {
+            state.x = Math.max(0, state.x - moveIncrement);
```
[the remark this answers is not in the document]
maxb94: Is this not needed

src/components/templates/Playa/MyAvatar.tsx
thecadams: Janky scrolling while I was messing with keys. Then the booleans got stuck because the key release events didn't fire.

maxb94 (approved):

PR #412, opened 2020-08-25
572 radio

sintinilorenzo:
The radio popUp has two options:
the first is to display an audio,
the second is to display an iFrame with a podcast from iHeart Radio (potentially where burning man radio will be)


src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useEffect } from "react";
+import "./RadioModal.scss";
+
+const sound = new Audio(
+  "https://www.soundhelix.com/examples/mp3/SoundHelix-Song-15.mp3"
```
thecadams, line 5: 2 requests
1. Can this please come from the playa venue configuration? Can be passed in via props if you like.
2. Can the venue configuration we use be an array, as we may want to do a small bit of UI during burn and allow multiple radio stations.

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect } from "react";
+import "./RadioModal.scss";
+
+const sound = new Audio("http://bmir-ice.streamguys.com/live");
```
thecadams, line 4: Just a nudge re my earlier comment about this coming from venue config

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const RadioModal: React.FunctionComponent<PropsType> = ({
+  volume,
+  setVolume,
+}) => {
+  const { radioStations } = useSelector((state) => ({
+    radioStations: state.firestore.data.venues?.playa?.radioStations,
+  }));
+
+  const sound = useMemo(
+    () => (radioStations ? new Audio(radioStations[0]) : undefined),
```
thecadams, line 19: Should probably do a length check: `radioStations && radioStations.length ? new Audio(radioStations[0]) : undefined` ?

thecadams: Or `new Audio(undefined)` is OK then no prob.

PR #418, opened 2020-08-26
event schedule for events right now

sintinilorenzo:


src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+    "May",
+    "Jun",
+    "Jul",
+    "Aug",
+    "Sep",
+    "Oct",
+    "Nov",
+    "Dec",
+  ];
+  const DateToString = (Date: Date) =>
```
thecadams, line 110: can we use something similar or based on https://github.com/sparkletown/sparkle/blob/staging/src/utils/time.js#L64 ?

maxb94 (approved):

PR #420, opened 2020-08-27
🐞 Limit dust storm duration

thecadams:


functions/venue.js
```diff
@@ -216,6 +216,32 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
...
+                  const updated = doc.data();
+                  updated.dustStorm = false;
+                  admin
+                    .firestore()
+                    .collection("venues")
+                    .doc("playa")
+                    .update(updated);
+                }
+              }),
+          60 * 1000
```
sintinilorenzo, line 242: Is 1 min long enough? I thought it would have to be long enough (~20min) to force users to go to other venues. After 1min they'll just go back in the playa.

thecadams: Hm, I think it'd be pretty annoying if you wanted to explore and got forced to wait out the dust storm...

And, we don't want the function to time out...

maxb94 (approved):

PR #431, opened 2020-08-27
4k admin placement playa

maxb94:


sintinilorenzo:
Can you set the initial position of the map in the scrollable page so that the venue is in the middle? Right now it's so big that it's hard to find the venue selected.

sintinilorenzo (approved):

PR #446, opened 2020-09-01
Remove of Camp page message at the bottom "what's on now/ loading" 

dibat:
On the Camp page- this no longer has any function in this context, so complete removal (via commenting out)
Remove the whole message at the bottom "what's on now/ loading" (you can comment it out.


thecadams (commented):
Couple of questions

- Can you please remove the commented code?
- Can we replace references to currrentRooms with just rooms, so it shows room cards for every room in the camp?

dibat:
> Couple of questions
> 
> * Can you please remove the commented code?
> * Can we replace references to currrentRooms with just rooms, so it shows room cards for every room in the camp?

Done, see my latest commit.

PR #458, opened 2020-09-02
fixed the useEffect retrgger problem

andefred:


thecadams:
@andefred looks good, got a tiny bit more context?

andefred:
@thecadams The context for this PR is that when the app loaded, it re-ran the useEffect 100+ times. This was caused by the onClickHandler being defined outside so it was being added and removed hundreds of times. I doubt it will make any significant impact on performance, but this is apparently a better approach to ensure fewer renders.

The more renders => the more potential performance issues.

thecadams:
Wicked! Great catch!

PR #477, opened 2020-09-03
add avatars when in video group call

denisdimitrov:
- add avatars in a hexagon when in a video call
- add blinking green indicator
- show purple when someone is in a video
- update some of the texts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18004,6 +18004,14 @@
         }
       }
     },
+    "react-switch": {
+      "version": "5.0.1",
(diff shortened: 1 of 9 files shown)


thecadams (changes requested):
Please implement this by updating the avatars' locations through a single Websocket relay message. That way everyone on playa will see the whole group roaming around.

thecadams:
Also I updated the branch here

thecadams:
Had a quick chat with @denisdimitrov and this implementation is better as it allows people to keep roaming around while in video chat! Going to approve and let's see how it is to use.

thecadams (approved):

PR #493, opened 2020-09-04
add layout for seats and position tweaks

denisdimitrov:
- add layout for seats
- position chat in top right corner (not sure if flex broke it or if it wasn't positioned at all)
- add keys for the components so react can stop crying


thecadams (approved):
Looking good so far, might need to merge!

PR #540, opened 2020-09-20
Seed adult venue

Demarchist1:


thecadams:
Looks good!

One thing is we have venues in the SparkleVerse env which don’t have this flag - what will happen for those?

Demarchist1:
> One thing is we have venues in the SparkleVerse env which don’t have this flag - what will happen for those?

I think they will default to false, but maybe I missed something?

thecadams:
Looks like they'll default to false to me too. Sounds good, that's what I suspected. Thanks for the response, this looks great  =] Cheers!

thecadams (approved):

PR #541, opened 2020-09-20
Seed kids mode

Demarchist1:


src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -85,6 +92,20 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
         onClick={logout}
         type="button"
       />
+
+      <label
+        htmlFor={"chk-kidsMode"}
+        className={`checkbox ${profile?.kidsMode && "checkbox-checked"}`}
+      >
+        Kids Mode
```
thecadams, line 100: Wonder if we should have something extra here to describe it, like:

`Kids Mode (do not show venues with 18+ content)`

thecadams (commented):
Looking good so far, just one thing I can't spot here - looks like venues with adult content are still shown as icons on the playa when in kids mode - maybe I missed that though.

Great stuff on this one!

thecadams (approved):
Going to proactively approve so once you've addressed the comments you're able to merge. Thanks again!

Demarchist1:
> Looking good so far, just one thing I can't spot here - looks like venues with adult content are still shown as icons on the playa when in kids mode - maybe I missed that though.

I was unable to actually test this, but I thought the changes I made in /Playa/Playa.tsx would exclude them. Is there somewhere else I should be looking for the map icons?

thecadams:
Ah no prob - if it doesn’t hide them we can address in a follow up PR. Let’s merge this one and keep going

thecadams:
Thanks again!

PR #554, opened 2020-09-22
add prototype for memrise chats venue

thecadams:


thecadams:
Closing in favor of @denisdimitrov 's PR:

thecadams:
@denisdimitrov there's a whole boatload of changes here that are not necessary like the bike mode, the gate coordinates. The implementation you've brought over is based around holding the button down, whereas what you want is going to be a "press the button once to move one cell over".

Did something go wrong with your merge? Just trying to understand what's happened.

denisdimitrov:
I was not aware that we want 1 cell movement, I thought the functionality was supposed to be like on the playa and that's what I've been working on until now. My bad. I will try to fix it by the start of the call.

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -0,0 +1,278 @@
...
+import PARTY_BACKGROUND from "./party-bg.jpg";
+
+type Props = {
+  venueName: string;
+};
+
+const DEFAULT_COLUMNS = 40;
+const DEFAULT_ROWS = 25;
+
+const WIDTH = 2451;
```
thecadams, line 23: Just wondering - how will this make the grid look on mobile? What about on an iPad, or an old laptop? Is fixed width and height the right approach here? Can we use reponsive breakpoints or flex with max-width so the grid cells shrink but the grid scrolls vertically, for example?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -29,15 +30,15 @@ import { GiftTicketModal } from "../../organisms/GiftTicketModal/GiftTicketModal
...
+import { VenueTemplate } from "types/VenueTemplate";
 
 interface PropsType {
   redirectionUrl?: string;
 }
 
 const NavBar: React.FunctionComponent<PropsType> = ({ redirectionUrl }) => {
   const { user, profile } = useUser();
-  const venueId = useVenueId();
+  const { venueId } = useParams();
```
thecadams, line 41: you might find this doesn't work, you should test very carefully. I was seeing venueId as `undefined` in nested routing contexts such as routes in the `CampRouter`, so I suspect you should use the `useVenieId` hook.

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+const DEFAULT_COLUMNS = 40;
+const DEFAULT_ROWS = 25;
+
+const AvatarGrid = ({ venueName }: Props) => {
+  const [selectedUserProfile, setSelectedUserProfile] = useState<
+    WithId<User>
+  >();
+
+  const venueId = venueName;
```
thecadams, line 27: I suspect this is problematic, as we normally convert the venue ID. You can use `const venueId = useVenueId()` instead though.

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -0,0 +1,324 @@
...
+    return keyPressed;
+  };
+
+  const downPress = useKeyPress("ArrowDown");
+  const upPress = useKeyPress("ArrowUp");
+  const leftPress = useKeyPress("ArrowLeft");
+  const rightPress = useKeyPress("ArrowRight");
+
+  useEffect(() => {
+    const currentPosition = profile?.data?.memrisechats;
```
thecadams, line 114: This is going to affect ALL venues with this template. We need to make it based on venueId.

src/components/templates/Playa/PlayaAdmin.tsx
```diff
@@ -7,17 +7,20 @@ const PlayaAdmin: React.FC = () => {
   const [error, setError] = useState<string | null>();
 
   const playaVenue = useSelector((state) => state.firestore.data.currentVenue);
+  const PLAYA_VENUE_ID = "paddock";
```
thecadams, line 10: Already available in `settings.ts`

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -208,6 +209,12 @@ const VenuePage = () => {
       template = <AudienceRouter />;
       fullscreen = true;
       break;
+    // case VenueTemplate.avatargrid:
```
thecadams, line 212: Should this be commented out? A quick test of a venue should tell you.

thecadams (commented):
Looking good with some minor things to look at, looking forward to getting this merged!

thecadams:
Looking good! Going to merge! Thanks to both of you @denisdimitrov and @aivonix !

PR #557, opened 2020-09-23
hide messages from before seed

Demarchist1:
Is there a way to filter at the firebase back end rather than the front end?

### DIFF
--- a/src/components/organisms/ChatDrawer/ChatDrawer.tsx
+++ b/src/components/organisms/ChatDrawer/ChatDrawer.tsx
@@ -14,6 +14,7 @@ import {
 } from "@fortawesome/free-solid-svg-icons";
 import useRoles from "hooks/useRoles";
 import { useParams } from "react-router-dom";
+import { SEED_START_UTC_SECONDS } from "settings";
 
 interface ChatOutDataType {
   messageToTheBand: string;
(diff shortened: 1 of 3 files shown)


thecadams:
You can filter in the `useFirestoreConnect` call. Maybe we should always filter venue messages older than X milliseconds.

(We should never hide old private messages of course)

thecadams:
This PR is a solid idea, a more general solution would account for the code being used in non burn mode and for other SparkleVerse events; would you be willing to have a think about a more generalised solution? Going to proactively approve so you’re not blocked. Great stuff.

src/components/organisms/ChatDrawer/ChatDrawer.tsx
```diff
@@ -62,6 +63,7 @@ const ChatDrawer: React.FC<PropsType> = ({
       chatContext.sendRoomChat(user.uid, roomName, data.messageToTheBand);
     reset();
   };
+  const BEFORE_SEED = SEED_START_UTC_SECONDS - 5 * 1440000;
```
thecadams, line 66: nit: would you be able to include some justification for the choice of constant here, as well as what 1440000 is in millis

thecadams (approved):
Proactively approving this one so you’re not blocked- but it’d be good to think about making this change not specific to Burning Seed.

Demarchist1:
I moved the filter to the firebase call, hiding messages older than 2 weeks (constant in settings)

PR #560, opened 2020-09-23
Fix/revert basic party modal 

aivonix:
https://trello.com/c/QoSq0lTJ/8-make-basic-party-map-modals-become-the-full-black-popup-as-before

Fix to make a onClick from map to show the same modal as room-list, whilst keeping IS_BURN logic the same.

### DIFF
--- a/src/components/templates/Camp/components/Map.tsx
+++ b/src/components/templates/Camp/components/Map.tsx
@@ -90,7 +90,7 @@ export const Map: React.FC<PropsType> = ({
                     alt={room.title}
                   />
                 </div>
-                <div className={IS_BURN ? `playa-venue-text` : ``}>
+                <div className={`playa-venue-text`}>
                   <div className="playa-venue-maininfo">
                     <div className="playa-venue-title">{room.title}</div>
                     <div className="playa-venue-people">
--- a/src/components/templates/Camp/components/Map.tsx
+++ b/src/components/templates/Camp/components/Map.tsx
@@ -4,7 +4,6 @@ import { CampRoomData } from "types/CampRoomData";
 import "./Map.scss";
 import { enterRoom } from "../../../../utils/useLocationUpdateEffect";
 import { useUser } from "../../../../hooks/useUser";
-import RoomCard from "./RoomCard";
 import { IS_BURN } from "secrets";
 
 interface PropsType {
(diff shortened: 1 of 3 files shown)


src/components/templates/Camp/components/Map.tsx
```diff
@@ -64,18 +69,14 @@ export const Map: React.FC<PropsType> = ({
...
-                  setIsRoomModalOpen(true);
+                if (!IS_BURN) {
+                  openModal(room);
                 } else {
                   setRoomClicked((prevRoomClicked) =>
                     prevRoomClicked === room.title ? undefined : room.title
                   );
                 }
               }}
-              onMouseEnter={() => {
```
thecadams, line 76: You still need this for burn mode, you able to keep it?

src/components/templates/Camp/components/Map.tsx
```diff
@@ -4,6 +4,7 @@ import { CampRoomData } from "types/CampRoomData";
 import "./Map.scss";
 import { enterRoom } from "../../../../utils/useLocationUpdateEffect";
 import { useUser } from "../../../../hooks/useUser";
+import RoomCard from "./RoomCard";
```
thecadams, line 7: Can’t see where this is used, did the build succeed?

src/components/templates/Camp/components/Map.tsx
aivonix: reverted that back in

src/components/templates/Camp/components/Map.tsx
aivonix: removed this line

thecadams (approved):

src/components/templates/Camp/components/RoomModalOngoingEvent.tsx
```diff
@@ -42,18 +42,49 @@ export const RoomModalOngoingEvent: React.FunctionComponent<PropsType> = ({
...
       )}
+      {eventToDisplay && !currentEvent && (
+        <>
+          <div className="title-container">
+            <img
+              src="/sparkle-icon.png"
+              className="sparkle-icon"
+              alt="sparkle-icon"
+            />
+            What's on next
```
thecadams, line 55: Can you please clean up the `whatsOnText` while in here?

More broadly let's keep an eye out for refactors leaving things in a half finished state. Cheers!

thecadams (changes requested):
Looking good but there's some eye for detail type feedback that I'm hoping we can make sure we incorporate. Thanks for this!

thecadams (approved):
Looking good, thanks for taking care of all the feedback.

PR #563, opened 2020-09-24
time in Matong

Demarchist1:


thecadams (commented):
Thanks for this change! Can you please make this a setting?

Demarchist1:
I added a few more things to the settings related to differences between BM and BS

thecadams (approved):
LGTM!

PR #582, opened 2020-09-25
do not unset background image width

Demarchist1:
this seems to save the placement co-ordinates correctly


thecadams (commented):
Looking to understand why this fixes it. Is it that the background width was being stretched, misleading people into thinking their venue was in a different spot than the coordinates they’d dragged it to? If so, is there a 4000 pixel wide container somewhere that we need to fix?

thecadams:
And further I’m wondering if some git blame can be done to uncover why the unset was put there in the first place.

Demarchist1:
I don't know - I just noticed that the icon's x and y co-ords appeared to be double on the actual paddock compared to the admin venue screen. I was digging around trying to see what you had done to convert the coordinates, and noticed the PlayaContainer width being unset from the admin component but not any other components

thecadams:
OK no prob, thanks for this one I'll go merge it.

PR #592, opened 2020-09-26
dust storm

Demarchist1:


thecadams:
This is a great change - I wonder if we could make the copy a setting?

thecadams (approved):

PR #631, opened 2020-10-08
add disable option to location update

denisdimitrov:
- Remove updateLocationData from the hook because it was running on every tab
- Conditionally update the location of the profile only when it's not available in the firebase

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -16,7 +16,10 @@ import { VenueTemplate } from "types/VenueTemplate";
 import { hasUserBoughtTicketForEvent } from "utils/hasUserBoughtTicket";
 import { isUserAMember } from "utils/isUserAMember";
 import { canUserJoinTheEvent, ONE_MINUTE_IN_SECONDS } from "utils/time";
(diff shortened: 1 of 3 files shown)


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -80,14 +83,21 @@ const VenuePage = () => {
...
+  const profileRoom = profile?.room;
+  const location =
+    venueName === profileRoom
+      ? venueName ?? ""
+      : isVenueRoom
+      ? profileRoom ?? ""
+      : venueName ?? "";
+  useLocationUpdateEffect(user, location);
+
+  if (!profileRoom && user) {
```
thecadams, line 98: Seems this is redundant- the location is always updated above?

denisdimitrov: I removed the updateLocationData from the hook, it only updates the time you've spent in there. The problem was that every single open tab was executing the code.

thecadams: Aha I see now, looks good!

thecadams (approved):

PR #633, opened 2020-10-08
Just changing the title in playa admin.

aivonix:
Simple change to display the venue.name


thecadams:
Heya @aivonix thanks for this PR! Going to guess what happened here was Ed and myself noticed this one at the same time as one another and while I fixed it he asked you to fix it! Anyway we're good here as I've implemented basically the same change, so thanks very much for this one & going to close it out to clean up the PR list.

PR #636, opened 2020-10-09
add room resize on playa template

denisdimitrov:


thecadams:
Hi Denis, while I can see and appreciate the approach here, I think we are better off doing this as rooms separate to placement, in the same way as the Camp template works.

This is because I want to keep placement as an isolated feature for burning man like events, and want us to use percentages for room positioning, and want us to be able to reuse the components on the admin UI that are used for positioning rooms on the camp template..

Can you please rework the PR to do this?

PR #647, opened 2020-10-12
add new playa boundaries

denisdimitrov:


src/settings.ts
```diff
@@ -54,7 +54,8 @@ export const MAX_IMAGE_FILE_SIZE_TEXT = "2MB";
 export const GIF_IMAGE_WIDTH_PX = 300;
 
 // playa is 4000x4000 pixels, Burning Seed paddock is 2000x2000
-export const PLAYA_WIDTH_AND_HEIGHT = 2000;
+export const PLAYA_HEIGHT = 1335;
```
thecadams, line 57: Should be 2000 I think?

thecadams (commented):
Looking good, left a comment, I think we aren't changing the dimensions themselves just yet right?

thecadams (approved):

PR #653, opened 2020-10-13
fix jazzbar tables not showing participants

denisdimitrov:
The logic was using the references to display the users inside but the join function was passing title to the firebase


thecadams:
Heya! My bad I didn’t assess the potential for this change and introduced this bug.

The original change was an attempt to show the correct table title in the TableHeader.

You could change that to something like: (note this is off the top of my head)

`TABLES.find(table => table.reference === seatedAtTable.reference)?.title ?? seatedAtTable.reference`

PR #657, opened 2020-10-14
fix venue and room counting

denisdimitrov:
W.I.P

fix venue and room counting by migrating the structure to object with timestamps

### DIFF
--- a/src/components/templates/Playa/Playa.tsx
+++ b/src/components/templates/Playa/Playa.tsx
@@ -24,6 +24,7 @@ import {
   DEFAULT_PARTY_NAME,
   PLAYA_WIDTH,
   PLAYA_HEIGHT,
+  LOC_UPDATE_FREQ_MS,
 } from "settings";
 import VenuePreview from "./VenuePreview";
 import { WithId } from "utils/id";
@@ -54,6 +55,7 @@ import { PlayaIconComponent } from "./PlayaIcon";
 import { IS_BURN } from "secrets";
 import BannerMessage from "components/molecules/BannerMessage";
 import UserList from "components/molecules/UserList";
+import { currentTimeInUnixEpoch } from "utils/time";
 
 export type MenuConfig = {
   prompt?: string;
@@ -348,7 +350,7 @@ const Playa = () => {
     user &&
       updateLocationData(
         user,
-        { [PLAYA_VENUE_NAME]: new Date().getTime() },
+        { [PLAYA_VENUE_NAME]: currentTimeInUnixEpoch },
         profile?.lastSeenIn
       );
   }, [setShowModal, user, profile]);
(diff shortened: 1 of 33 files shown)


thecadams:
Hi @denisdimitrov 
Gave this a go and it seems to uncount you from attendance on the playa map while also removing your avatar, very easy to reproduce, you able to take a look?

Also if you push changes you'll find I've made some updates including fixing the typo `retainAttendance`. Best not to force push but rather to rebase whatever you do, or else to pull latest changes on the branch.

src/components/templates/Playa/Playa.tsx
```diff
@@ -644,7 +669,9 @@ const Playa = () => {
                       {venue?.roomVisibility === RoomVisibility.nameCount && (
                         <div className="playa-venue-title">{v?.name}</div>
                       )}
-                      <div className="playa-venue-people">{usersInVenue}</div>
+                      <div className="playa-venue-people">
+                        {usersInVen.length}
```
thecadams, line 673: usersInVenue?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -64,8 +73,62 @@ const VenuePage = () => {
...
+      clearInterval(interval);
+      if (retainAttendance && user && profile) {
+        updateUserProfile(user.uid, {
+          lastSeenIn: { ...prevLocations, [venueName]: 0 },
+          lastSeenAt: new Date().getTime(),
+          room: null,
+        });
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
thecadams, line 129: Guessing this is the cause of the location unloading? Would be great not to have this in the final PR.

thecadams (commented):
Couple of things needing a second look, but mostly pretty good! Tons of solid work here, almost to the finish line!

src/pages/VenuePage/VenuePage.tsx
denisdimitrov: The problem here is that profile needs to be added in the dependency array and whenever this logic updates the location of the user, the profile is changed as well, and then this reruns and then it becomes an infinite loop

denisdimitrov: If you have any suggestions or if statements that can avoid this feel free to post them here.

thecadams: Understood. This is hiding a bug and usually will prevent renders that should happen..

thecadams: What about changing updateUserProfile to call a function that only needs to receive the new attendance. Something where profile is no longer referenced inside the hook.

const { profile } = useUser()
const updateAttendance(user, loc, ts) { updateUserProfile(user.uid, {...profile, [loc]: ts}) }

Then you can remove the dependency on profile that’s inside the hook.

PR #659, opened 2020-10-15
Fix conversation space table layout

thecadams:


src/scss/global.scss
```diff
@@ -452,7 +452,8 @@ h4 {
 }
 
 .full-page-container,
-.friendship-container {
+.friendship-container,
+.conversation-space-container {
```
thecadams, line 456: This is the fix.

denisdimitrov (approved):

PR #673, opened 2020-10-19
fix chat avatars and names

denisdimitrov:


thecadams:
@denisdimitrov I did a quick test in prod and this fix hangs the page while it loads every user (4000+ right now). Can we make it selective so it loads into partygoers say, any users who the person has chatted with in the past year?

thecadams:
@denisdimitrov did you get a chance to look at this one?

thecadams:
Another possible approach here is to load the users you've private-chatted with into their own `storeAs` collection, as there shouldn't be more than a few hundred for people.

PR #678, opened 2020-10-22
fix counting logic in the other venues

denisdimitrov:
- Add the correct counting logic in the rest of the templates
- Add /room suffix to keep only 1 room at a time

P.S. @thecadams  Please take a look at the .log file that I've committed. I can't remove it and it's being generated all the time, would you mind if I add it to git.ignore?


thecadams (commented):
Looks like we are using the `lastSeenIn` map now, but keying it by adding `/room` ? Won't that lead to incorrect counting, if someone names a room the same name as a venue ID that's elsewhere in the system?

thecadams:
Totally OK to add the log file to the gitignore, no prob!

PR #679, opened 2020-10-23
Fix: avatar being displayed as background image instead of as img tag…

IvanBuljovcic:
- [x] Avatar showing as background-image in _div_ instead of as a _img_
- [ ] Image not having its dimensions distorted

- Code cleanup
- Memoizing _UserProfilePicture_ component render, to reduce re-render amount
- _avatarSrc_ function wrapped in _useCallback_ to reduce re-render amount

### DIFF
--- a/src/components/molecules/UserList/UserList.tsx
+++ b/src/components/molecules/UserList/UserList.tsx
@@ -1,12 +1,22 @@
 import React, { useState } from "react";
+
+// Components
 import UserProfileModal from "components/organisms/UserProfileModal";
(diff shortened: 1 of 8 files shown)


thecadams (approved):

thecadams:
@IvanBuljovcic what about the onerror stuff that is in the conflicts? Do we have the ability to display a mini avatar if there's not an image?

thecadams (approved):
lgtm, neat solution for this.

PR #682, opened 2020-10-23
fix redirect after account creation

denisdimitrov:
- Move returnUrl logic one step deeper to avoid the direct back to the first step
- Make join event button redirect directly to /in/ instead of entrance page
- Add account profile page to the registration flow because it was missing in some of the cases

### DIFF
--- a/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
+++ b/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
@@ -2,12 +2,13 @@ import firebase from "firebase/app";
 import React from "react";
 import { useForm } from "react-hook-form";
 import { CodeOfConductFormData } from "pages/Account/CodeOfConduct";
-import { useHistory, useParams } from "react-router-dom";
+import { useHistory } from "react-router-dom";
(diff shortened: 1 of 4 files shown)


src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -56,7 +57,11 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
   closeAuthenticationModal,
 }) => {
   const history = useHistory();
-  const { venueId } = useParams();
+  const currentVenue = useSelector(
+    (state) => state.firestore.data.currentVenue
```
thecadams, line 61: `firestore.ordered.currentVenue[0]`

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -56,7 +57,11 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
...
+  );
+
+  const venueId = currentVenue?.name;
```
thecadams, line 64: `currentVenue.id`

thecadams (approved):

PR #686, opened 2020-10-27
Fix/auditorium shoutout position

IvanBuljovcic:
* Added [**Styled Components**](https://styled-components.com/docs) to _UserProfilePicture_ as a demo to showcase the benefits of the library

* When the user is seated to the **left** of the screen, reaction/shout-out message transition to the right
* When the user is seated to the **right** of the screen, reaction/shout-out message transition to the left

This (reaction/message position) can be expanded to include the other 6 directions:
top; top-right; bottom-right; bottom; bottom-left; top-left


src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+    transform: scale(1) translateX(-${translateXOffset});
+  }
+`;
+
+const messageLeft = css`
+  right: 0;
+  transform-origin: right center;
+  animation: ${expandBounceLeft} 5s ease;
+`;
+const messageRight = css`
```
thecadams, line 98: just checking, looks like the convention here is golang-style, lowercase for unexported and uppercase for exported?

src/components/templates/Audience/Audience.tsx
```diff
@@ -196,31 +196,38 @@ export const Audience: React.FunctionComponent = () => {
...
 
-    const translateRow = (untranslatedRowIndex: number) =>
-      untranslatedRowIndex - Math.floor(rowsForSizedAuditorium / 2);
-    const translateColumn = (untranslatedColumnIndex: number) =>
-      untranslatedColumnIndex - Math.floor(columnsForSizedAuditorium / 2);
+  const translateColumn = (untranslatedColumnIndex: number) =>
+    untranslatedColumnIndex - Math.floor(columnsForSizedAuditorium / 2);
 
-    const isSeat = (translatedRow: number, translatedColumn: number) => {
+  const isSeat = useCallback(
```
thecadams, line 212: nit: looks like your IDE isn't doing prettier autoformatting?

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+// --- Reaction
+const reactionOffset = "-20px";
+const reactionLeft = css`
+  left: ${reactionOffset};
+`;
+const reactionRight = css`
+  right: ${reactionOffset};
+`;
+
+type TReaction = {
```
thecadams, line 24: nit: wondering about the Hungarian notation here; do you recommend propagating this across the codebase? if so is there a lint rule?

thecadams:
I think I follow, seems similar to material UI. One concern I have is the lack of a lot of nice things SASS/LESS/SCSS provide like class nesting, `&` aliasing, functions such as darken/lighten (https://css-tricks.com/snippets/javascript/lighten-darken-color/) - I'm wondering if we should go with a wrapper component instead, something like:


```
// Replace non-word characters with `-`
const cssify(name: string) => name.replace(/\W/g, "-").toLowerCase();

// Always wrap components in a wrapper div according to name convention
const Container = (of: string) => <div className={`${cssify(of)}-container`}>{children}</div>;

// MyComponent has the stuff that's wrapped by the parent div
const MyComponent = () => <Container of="MyComponent"><div>CONTENT</div></Container>;
```

Styles in `MyComponent.scss`:
```
.my-component-container {
  // ...
}
```


This seems like a solution very similar to what we have today, but enforced by the `Container` component.. thoughts?

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
IvanBuljovcic: Its basically like the default React convention.

so for components its _ExampleComponent = styled.[tag]_
and in this case, for block styles _exampleStyle = css`...`_, like with variable or function names

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
IvanBuljovcic: Its not something im pushing for, rather something im familiar with..

So the prefix _T_ for type and _I_ (upper case i) for interface, followed by the component name

IvanBuljovcic:
styled-components **does** support nesting, & and all the helper features you would find in SCSS, it does not support mixins however.

Not that im aware of, but there is no need for them since we require to extend some component, you would write:

```
import { Card } from 'somewhere'

export const MyCard = styled(Card)`
...additional styles
`
```

IvanBuljovcic:
For page (parent) specific cases where we, for instance, need a component to have different css then its default:

```
export const ParentComponent = styled.div`
    ${ChildComponent} {
        background: 'firebrick'
    }
`
```

This is equivalent to doing:
```
.parent-component {
    .child-component {
        background: 'firebrick';
    }
}
```

thecadams:
Hi again @IvanBuljovcic,
Just thinking some more about this PR, it could be a great opportunity to show off the snapshot testing + styled components approach we discussed last week. Does that sound like something feasible to land in this PR alongside introducing styled components, if that's the approach you're basically recommending here?

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
thecadams: Catching back up here, it is inconsistent with the rest of the code, so let's remove it.. would be great if we could migrate everything but won't be possible unfortunately.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -29,15 +29,14 @@ import {
...
-  avatarClassName?: string;
-  avatarStyle?: object;
-  containerStyle?: object;
-  setSelectedUserProfile: (user: WithId<User>) => void;
-  user: WithId<User>;
+import * as S from "./UserProfilePicture.styles";
+
+// This would be the global variables and configuration,
+// it is here only for the duration of the demo
+const DEMO_THEME = {
```
thecadams, line 36: just wondering, what's the plan for this?

thecadams:
@IvanBuljovcic catching up here, what's your plan for this PR?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
IvanBuljovcic: This will be removed in the next commit, it was here just as the name suggests, as a demo of how styled components would work.

The new implementation of the self-serve venue creation sets-up the theme provider in the index.tsx file so this it is redundant in this file

IvanBuljovcic:
> 
> 
> @IvanBuljovcic catching up here, what's your plan for this PR?

I intended to have this done with tests but due to the nature of the components using redux and context, it was taking too much time from other issues.

This PR will be updated to the latest staging, with the addition of Denis's code regarding avatar styles.

In short, it is in the to-do list very soon

thecadams:
HI @IvanBuljovcic can we make sure to get around to updating this? Luckily still not conflicts but won't be long!

thecadams (approved):
Thanks for cleaning it up, let's merge and test.

PR #688, opened 2020-10-29
Fix: iframe URL updates when the venues iframeUrl changes either via …

IvanBuljovcic:
Using Firestores _onSnapshot_, audiance `iframeUrl` property updates automatically when the venue.iframeUrl changes.

Tested by changing URL in _/admin_ panel of the venue and by direct change in _firestore_


thecadams:
@IvanBuljovcic you able to fix the build so we can merge this?

PR #695, opened 2020-11-05
fix camp grid and remove room glow

denisdimitrov:
- fixes the streching of the background when grid is disabled
- removes the white background glow of the rooms (currently for all of the camp venues)


thecadams (approved):
lgtm

PR #702, opened 2020-11-10
calculate rows based on the aspect ratio of the image

denisdimitrov:


thecadams:
Very cool. May as well not even save the row count into Firestore.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -292,6 +294,19 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
 
   const defaultVenue = createJazzbar({});
 
+  const img = new Image();
```
0xdevalias, line 297: I believe that currently, this image object will be created on every re-render of the functional component (which would then re-trigger the `useEffect`, etc). Is that the desired/intended state?

If it was just created 'outside' of the `useEffect` for 'variable scope' reasons, you could probably use `useRef` or similar to 'store' it in a persistent way.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -583,37 +598,44 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
...
+                defaultValue={rows ?? 1}
+                value={rows ?? 1}
+                name="rows"
                 type="number"
                 ref={register}
-                className="align-left"
-                placeholder={`Number of grid columns`}
+                onChange={(e) => {
+                  setValue("rows", parseInt(e.target.value));
+                }}
```
0xdevalias, line 634: Since this is seemingly always disabled now, does the onChange do anything anymore? When will it change? (and potentially also, does `input` make sense to use anymore if it will never change?)

0xdevalias (commented):
[CRO, NB] A couple of cursory 'code only' (aka: i didn't pull the branch/run/test it locally) comments/questions. Non-blocking.

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: The whole component is now removed because the calculation is happening in the Map.tsx and it's no longer necessary.

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: I just moved it inside the useEffect which has ```[venue.columns, venue.mapBackgroundImageUrl]``` as dependency array. Now new image is going to get created only when necessary.

PR #707, opened 2020-11-12
Fix backend function updateVenue to properly check showGrid/showLiveSchedule

0xdevalias:
(Ref: https://trello.com/c/7MzYxfQE)

When updating venues other than theme camps (eg. Music Venue), `showGrid` isn't a toggle shown on the frontend, so there was no data being sent through for it, resulting in `undefined`.

The backend `updateVenue` function wasn't checking `showGrid`/`showLiveSchedule` values before adding them to the `updated` data.

This led to a backend error within firestore, which eventually resulted in cryptic error `One or more errors occurred when saving the form:` being shown to the user on the frontend, and the following error in the DevTools console:

```
Error: INTERNAL
    at new t (error.ts:66)
    at error.ts:175
    at e.<anonymous> (service.ts:231)
    at tslib.es6.js:100
    at Object.next (tslib.es6.js:81)
    at a (tslib.es6.js:71)
```

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -333,102 +333,110 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
 
   await checkUserIsAdminOrOwner(venueId, context.auth.token.user_id);
 
-  await admin
-    .firestore()
-    .collection("venues")
-    .doc(venueId)
-    .get()
-    .then((doc) => {
-      if (!doc || !doc.exists) {
-        throw new HttpsError("not-found", `Venue ${venueId} not found`);
-      }
-      const updated = doc.data();
-      if (data.bannerImageUrl || data.subtitle || data.description) {
-        if (!updated.config) {
-          updated.config = {};
-        }
-        if (!updated.config.landingPageConfig) {
-          updated.config.landingPageConfig = {};
-        }
-      }
-      if (data.bannerImageUrl) {
-        updated.config.landingPageConfig.bannerImageUrl = data.bannerImageUrl;
-      }
-      if (data.subtitle) {
-        updated.config.landingPageConfig.subtitle = data.subtitle;
-      }
-      if (data.description) {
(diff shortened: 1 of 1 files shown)


functions/venue.js
```diff
@@ -333,102 +333,115 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
...
+
+  if (typeof data.showLiveSchedule === "boolean") {
+    updated.showLiveSchedule = data.showLiveSchedule;
+  }
+
+  if (typeof data.showGrid === "boolean") {
+    updated.showGrid = data.showGrid;
+  }
+
+  switch (updated.template) {
```
0xdevalias, line 427: We could also potentially move these within the `switch` statement, so that they are only handled for the correct `VenueTemplate` types.

thecadams: The direction I think it makes sense to move is more toward template agnostic here, so the less fussy the fields being updated are about the template, the better. But when I rip through this to make the venue config template agnostic, this will all change anyway so your call. I’m ok with it as is.

thecadams (commented):
You may need to make sure your IDE is honouring the prettier config we have for the project, and enable auto format on save, there’s lots of non-code changes in this PR that are making it harder to review..

0xdevalias:
> not: You may need to make sure your IDE is honouring the prettier config we have for the project, and enable auto format on save, there’s lots of non-code changes in this PR that are making it harder to review..

My IDE is honouring the prettier config, that change in whitespace is technically correct (see the commits), as I removed the `then` and continue to use async/await.

You can tell GitHub diffs to ignore whitespace as well (or can do so on git CLI, depending on your personal workflow)

image

functions/venue.js
0xdevalias: Works for me :)

:shipit: !

thecadams (approved):
lgtm!

PR #708, opened 2020-11-12
fix room uncounting when joining room

denisdimitrov:


thecadams (approved):
OK! Let's get this out there and test it!

src/components/templates/Camp/components/Map.tsx
```diff
@@ -479,29 +491,36 @@ export const Map: React.FC<PropsType> = ({
...
-                        roomEnter(room);
-                      }}
-                      className="btn btn-white btn-small btn-block"
-                    >
-                      Join now
-                    </div>
+                    {isExternalLink(room.url) ? (
+                      <div
+                        className="btn btn-white btn-small btn-block"
+                        onClick={(e) => {
```
IvanBuljovcic, line 504: The code in onClick is repeated on line #515, it would make sense to make that a method which gets called onClick instead of 3 functions being called separately

IvanBuljovcic (changes requested):

src/utils/url.ts
```diff
@@ -24,3 +24,14 @@ export const venueRoomUrl = (venue: WithId<AnyVenue>, roomTitle: string) => {
   );
   return venueRoom ? venueRoom.url : venueInsideUrl(venue.id);
 };
+
+export const isExternalUrl = (url: string) => {
```
thecadams, line 28: this isn't referenced anywhere, is that intentional?

thecadams (commented):
noticed something on a second, closer review.. comments below.

denisdimitrov:
@thecadams @IvanBuljovcic Just pushed some changes, feel free to review once again.

IvanBuljovcic (approved):

thecadams (approved):
Looks good!

PR #711, opened 2020-11-12
search for room venue by venue id instead

denisdimitrov:


src/components/templates/Camp/components/Map.tsx
```diff
@@ -183,8 +183,9 @@ export const Map: React.FC<PropsType> = ({
 
   const roomEnter = useCallback(
     (room: CampRoomData) => {
+      console.log(venues);
```
IvanBuljovcic: 🙅

IvanBuljovcic (approved):

PR #716, opened 2020-11-13
Improve Admin component's venues loading performance

0xdevalias:
(ref: https://trello.com/c/aKYvJutd)

The `/admin/venue/:venueId` page would take a long time to load, sometimes hanging the UI for quite a while before letting users interact with it. This was even evident when a user only had access to as little as 2 venues.

This PR should improve the performance by reducing the number of unnecessary re-renders (and simplify/clean up some code patterns along the way).

I think there are likely still some other areas this could be improved still, but a partial fix shipped sooner is better than no fix shipped.

### DIFF
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -175,7 +175,7 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
       user &&
       privateChats
         .filter(isChatValid)
-        .filter((chat) => chat.to === user.uid && chat.isRead === false).length
+        .filter((chat) => chat.to === user.uid && !chat.isRead).length
     );
   }, [privateChats, user]);
 
@@ -293,7 +293,7 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
                       rootClose={true}
                     >
                       <span className="private-chat-icon">
-                        <div className="navbar-link-gift"></div>
+                        <div className="navbar-link-gift" />
                       </span>
                     </OverlayTrigger>
                   )}
@@ -311,7 +311,7 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
                               {numberOfUnreadMessages}
                             </div>
                           )}
-                        <div className="navbar-link-message"></div>
+                        <div className="navbar-link-message" />
                       </span>
                     </OverlayTrigger>
                   )}
@@ -331,7 +331,7 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
                         className={`profile-icon navbar-link-radio ${
                           volume === 0 && "off"
                         }`}
-                      ></div>
+                      />
                     </OverlayTrigger>
                   )}
                   <OverlayTrigger
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -1,18 +1,11 @@
 import React, { useState, useMemo, useRef } from "react";
-import firebase from "firebase/app";
-import "./NavBar.scss";
-import "./playa.scss";
+import { useFirestoreConnect } from "react-redux-firebase";
 import { Link, useHistory } from "react-router-dom";
-import { faTicketAlt } from "@fortawesome/free-solid-svg-icons";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { isChatValid } from "validation";
 import { OverlayTrigger, Popover, Modal } from "react-bootstrap";
-import PrivateChatModal from "components/organisms/PrivateChatModal";
-import { ProfilePopoverContent } from "components/organisms/ProfileModal";
-import { RadioModal } from "../../organisms/RadioModal/RadioModal";
-import UpcomingTickets from "components/molecules/UpcomingTickets";
-import { useUser } from "hooks/useUser";
-import AuthenticationModal from "components/organisms/AuthenticationModal";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faTicketAlt } from "@fortawesome/free-solid-svg-icons";
+
+import firebase from "firebase/app";
 import {
   DEFAULT_PROFILE_IMAGE,
   SPARKLEVERSE_LOGO_URL,
@@ -24,17 +17,29 @@ import {
   HOMEPAGE_URL,
 } from "settings";
 import { IS_BURN } from "secrets";
-import { useSelector } from "hooks/useSelector";
-import OnlineStats from "../OnlineStats";
-import { SchedulePageModal } from "../../organisms/SchedulePageModal/SchedulePageModal";
-import { useRadio } from "hooks/useRadio";
-import { GiftTicketModal } from "../../organisms/GiftTicketModal/GiftTicketModal";
-import PlayaTime from "../PlayaTime";
-import PlayaAddress from "../PlayaAddress";
-import { venueInsideUrl } from "utils/url";
+import { isChatValid } from "validation";
 import { VenueTemplate } from "types/VenueTemplate";
+import { venueInsideUrl } from "utils/url";
+
+import { useRadio } from "hooks/useRadio";
+import { useSelector } from "hooks/useSelector";
+import { useUser } from "hooks/useUser";
 import { useVenueId } from "hooks/useVenueId";
-import { useFirestoreConnect } from "react-redux-firebase";
+
+import AuthenticationModal from "components/organisms/AuthenticationModal";
+import PrivateChatModal from "components/organisms/PrivateChatModal";
+import { GiftTicketModal } from "components/organisms/GiftTicketModal/GiftTicketModal";
+import { ProfilePopoverContent } from "components/organisms/ProfileModal";
+import { RadioModal } from "components/organisms/RadioModal/RadioModal";
+import { SchedulePageModal } from "components/organisms/SchedulePageModal/SchedulePageModal";
+
+import OnlineStats from "components/molecules/OnlineStats";
+import PlayaAddress from "components/molecules/PlayaAddress";
+import PlayaTime from "components/molecules/PlayaTime";
+import UpcomingTickets from "components/molecules/UpcomingTickets";
+
+import "./NavBar.scss";
+import "./playa.scss";
 
 interface PropsType {
   redirectionUrl?: string;
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -1,11 +1,16 @@
-import React, { useState, useMemo, useRef } from "react";
-import { useFirestoreConnect } from "react-redux-firebase";
+import React, { useState, useMemo, useRef, useCallback } from "react";
+import {
+  ReduxFirestoreQuerySetting,
+  useFirestoreConnect,
+} from "react-redux-firebase";
 import { Link, useHistory } from "react-router-dom";
 import { OverlayTrigger, Popover, Modal } from "react-bootstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faTicketAlt } from "@fortawesome/free-solid-svg-icons";
 
 import firebase from "firebase/app";
+
+import { RootState } from "index";
 import {
   DEFAULT_PROFILE_IMAGE,
   SPARKLEVERSE_LOGO_URL,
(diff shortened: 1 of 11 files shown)


src/pages/Admin/Admin.tsx
```diff
@@ -1,12 +1,11 @@
-import AuthenticationModal from "components/organisms/AuthenticationModal";
```
0xdevalias, line 1: Note: It would be really good to split this file up into 1 file per component, but that would be best done in a separate PR with no other changes (for simplicity of reviewing)

src/pages/Admin/VenueOwnersModal.tsx
```diff
@@ -1,66 +1,101 @@
...
+      )
+      .then(setAllUsers);
+  }, [firestore]);
 
-  // @debt reading every user into redux is obviously bad.
-  // A better solution is to create an index on partyName and a firebase function to search
-  useFirestoreConnect({
-    collection: "users",
-    storeAs: "allUsers",
-  });
```
0xdevalias, line 29: Calling this out, as this is my biggest 'hrmmm...' of my changes.

The `@debt` comment already existed, yet the previous code was using `useFirestoreConnect`, which I believe was causing it to be run _many_ times, as higher level components re-rendered, etc.

It could probably be done better, but this seemed the simplest 'isolated change' way to go about it for now. May make sense to clean this up as part of a larger data fetching architecture re-think in the future.

Clicking through, even though it is only fetched the first time the component is rendered, the app seems to continue to function as expected. That said.. if in future we fix higher level components to reduce their re-renders, I'm not 100% certain this will still be the case.

Open to thoughts/alternatives.

src/pages/Admin/VenueOwnersModal.tsx
```diff
@@ -80,23 +115,15 @@ export const VenueOwnersModal: React.FC<VenueOwnersModalProps> = (props) => {
             onChange={(e) => debouncedSearch(e.target.value)}
           />
           <div className="row-container">
-            {typeof filteredUsers === "undefined" ? (
+            {hasResults &&
+              (filteredUsers ?? []).map((user) => (
```
0xdevalias, line 119: `filteredUsers` should always be defined when `hasResults` is, but TypeScript didn't seem to agree. This was a bit of a hacky workaround for that.

src/pages/Admin/VenueOwnersModal.tsx
```diff
@@ -1,66 +1,101 @@
...
+  // @debt reading every user is obviously bad.
+  const firestore = useFirestore();
+  useEffect(() => {
+    firestore
+      .collection("users")
+      .get()
+      .then((result) =>
+        result.docs.map<WithId<User>>(
+          (doc) => ({ ...doc.data(), id: doc.id } as WithId<User>) // TODO: be less hacky with types here?
+        )
```
0xdevalias, line 58: Calling this out as it feels kind of hacky, but wasn't sure if there was a better way to go about it.

A better way would probably be to build a custom validation function with type guards/similar:

- https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/pages/Admin/Admin.tsx
```diff
@@ -409,26 +423,28 @@ const VenueInfoComponent: React.FC<AdminVenueDetailsPartProps> = ({
 
 const Admin: React.FC = () => {
   const { user } = useUser();
+  const userId = user?.uid || "";
+
+  const { isAdminUser, isLoading: isAdminUserLoading } = useIsAdminUser(userId);
```
0xdevalias, line 428: This replaces `useRoles` from the old code, in hopefully a simpler to understand/more performant way.

0xdevalias (commented):
A few comments/etc to help guide you through my changes.

src/pages/Admin/VenueOwnersModal.tsx
thecadams: It at least loads only once. This is the sort of place I have been tempted to go to a Firebase Function - but you can see the disastrous outcome of that with `getOnlineStats` where it takes forever to load and the CPU cycles cost us $. It seems like Firestore should offer a graphql api for use cases like this, especially as we move into 10,000s of users. It's at least not worse than the previous implementation, which also receives snapshots every time a new account is created.

src/pages/Admin/VenueOwnersModal.tsx
thecadams: Agreed, it's a great find. More broadly it speaks to a problem with the "go get all the data in this collection from a distant datasource, and I'll filter it client-side" approach. Firebase really falls down here, and I predict it'll trigger a complete replatforming at some point in the future.

We should approach Firebase account execs & work with them on solutions because I have struggled so far on engineering solutions to this problem. I want to stay on Firebase as long as possible - it's amazing to not be woken up by infra pages.

src/pages/Admin/VenueOwnersModal.tsx
thecadams: yep, its type inference isn't perfect, so no prob at all.

src/pages/Admin/VenueOwnersModal.tsx
```diff
@@ -114,51 +141,44 @@ interface UserRowProps {
 const UserRow: React.FC<UserRowProps> = (props) => {
   const { user, isOwner, venueId } = props;
 
-  const [loading, setLoading] = useState(false);
+  const [isLoading, setLoading] = useState(false);
   const [error, setError] = useState<string>();
 
-  const onRemoveUserClick = useCallback(async () => {
+  const removeOwner = useCallback(async () => {
```
thecadams, line 147: good self explanatory renames here.

src/pages/Admin/VenueOwnersModal.tsx
```diff
@@ -114,51 +141,44 @@ interface UserRowProps {
 const UserRow: React.FC<UserRowProps> = (props) => {
   const { user, isOwner, venueId } = props;
 
-  const [loading, setLoading] = useState(false);
+  const [isLoading, setLoading] = useState(false);
```
thecadams, line 144: +1 for this convention for booleans

src/types/SparkleSelector.ts
```diff
@@ -0,0 +1,3 @@
+import { RootState } from "../index";
+
+export type SparkleSelector<T> = (state: RootState) => T;
```
thecadams, line 3: yep good addition

src/utils/filter.ts
```diff
@@ -0,0 +1,21 @@
...
+ *     somethingFalse && 'B',        // false
+ *     'C',                          // 'C'
+ *     somethingTrue && 'D',         // 'D'
+ *     'etc',                        // 'etc'
+ *   ]
+ *
+ *   myArray.filter(isTruthyFilter)  // ['A', 'C', 'D', 'etc']
+ *
+ */
+export const isTruthyFilter = <T>(item?: T | false): item is T => !!item;
```
thecadams, line 21: same same, good addition

src/pages/Admin/Admin.tsx
```diff
@@ -108,30 +127,41 @@ type VenueDetailsProps = {
...
-    (state) => ({
-      venues: state.firestore.data.venues ?? {},
-    }),
-    ["venues"]
-  );
+  const { url: matchUrl } = useRouteMatch();
+  const { pathname: urlPath } = useLocation();
+
+  const venueSelector = useCallback(makeVenueSelector(venueId), [venueId]);
+  const venue = useSelector(venueSelector, shallowEqual);
```
thecadams, line 134: Very cool, completely your call on this, but I wonder if `hooks/useSelector` could use this behind the scenes, per https://react-redux.js.org/api/hooks#recipe-useshallowequalselector

thecadams (approved):
Exciting stuff, not a very critiquey review, but hopefully helpful. LGTM

src/pages/Admin/Admin.tsx
thecadams: That is, use `shallowEqual` in `hooks/useSelector` if it makes sense.

0xdevalias:
Linking this here for future reference, seems that at least some of the `useFirestoreConnect()` performance issues we are seeing may be related to this:

- https://github.com/prescottprue/react-redux-firebase/issues/914
  - https://github.com/prescottprue/react-redux-firebase/issues/914#issuecomment-655038967

src/pages/Admin/VenueOwnersModal.tsx
0xdevalias: Makes sense. There are probably some aspects of Firebase that i'm not particularly deeply aware of yet RE: performance tuning, best practices/patterns, etc too, so that could extend our ability to keep using it.

0xdevalias: There may also be some underlying bugs in the libraries we are using and/or in how we are using them, eg.

- https://github.com/sparkletown/sparkle/pull/716#issuecomment-727189462

src/pages/Admin/VenueOwnersModal.tsx
0xdevalias: This specific comment was calling out my explicit typecasting with `as WithId<User>`

RE: Firebase functions, _nods_ makes sense. We can look further at tuning costs/etc when we have some spare dev time (lol@spare time ;p)

src/pages/Admin/Admin.tsx
0xdevalias: Even if we need more in future, these similar patterns may work 'better' for our usages/etc. At some point we may want to deprecate/remove/cleanup `useRole` in favour of this (I know it was used elsewhere in the code, though was out of scope (OOS) for this PR)

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -151,33 +175,45 @@ const NavBar: React.FunctionComponent<PropsType> = ({ redirectionUrl }) => {
       user &&
       privateChats
         .filter(isChatValid)
-        .filter((chat) => chat.to === user.uid && chat.isRead === false).length
+        .filter((chat) => chat.to === user.uid && !chat.isRead).length
     );
   }, [privateChats, user]);
 
-  const [showEventSchedule, setShowEventSchedule] = useState(false);
+  const [isEventScheduleVisible, setEventScheduleVisible] = useState(false);
```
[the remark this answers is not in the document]
0xdevalias: Agreed. I have a strong preference for: `isFoo`, `hasBar`, etc style naming for `booleans`. Makes life so much easier.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -265,19 +289,19 @@ const NavBar: React.FunctionComponent<PropsType> = ({ redirectionUrl }) => {
                     <OverlayTrigger
                       trigger="click"
                       placement="bottom-end"
-                      overlay={giftPopover}
+                      overlay={<GiftPopover />}
```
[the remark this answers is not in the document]
0xdevalias: When you say to 'type declare' it, I'm guessing that you read this syntax as a TypeScript thing? If so, that's not what it is (see below)

Previously this was a JSX component set to a `const` within the render of this main component. I separated that `const` out into an actual component, which is being passed to `overlay` here.

src/pages/Admin/Admin.tsx
```diff
@@ -150,61 +180,45 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId, roomIndex }) => {
...
-              <VenueInfoComponent
-                venue={venue}
-                roomIndex={roomIndex}
-                showCreateEventModal={showCreateEventModal}
-                setShowCreateEventModal={setShowCreateEventModal}
-                setShowDeleteEventModal={setShowDeleteEventModal}
-              />
-            )}
-          />
+          <Route path={`${matchUrl}/events`}>
```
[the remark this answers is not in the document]
0xdevalias: I believe I was following this part of the documentation:

- https://reactrouter.com/web/api/Route/children-func

But in re-reading it now, i'm not sure if my usage was necessarily correct..

> It works exactly like render except that it gets called whether there is a match or not.

It's actually for a 'function as children' pattern, which isn't what we want here:

- https://reactpatterns.com/#function-as-children

So really, we should change it back to using the `render` prop I think.

src/pages/Admin/Admin.tsx
0xdevalias: I would definitely want to think through/plan it more before I made a generalised call across all of our stuff. I'm not 100% sure if this is a good/best pattern, or if there is a better way to address the needs.

src/pages/Admin/Admin.tsx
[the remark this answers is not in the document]
0xdevalias: In re-re-reading it (for my current work) I realised my original reading was correct:

- https://reactrouter.com/web/api/Route/route-render-methods
  - > The recommended method of rendering something with a `<Route>` is to use children elements, as shown above.

So yeah, nested is both good, and the recommended method! All good here, no fixes required!

PR #721, opened 2020-11-17
'Large Party Map' Performance - Part 1/x

0xdevalias:
**Note:** this is only the first PR for _'Large Party Map' Performance_. I'm splitting them up to make code review easier, and to reduce the risk of deployments by working incrementally. I will link the others here when they are also ready.

----

**IMPORTANT:** This code should be cloned/checked by running before considering it ready to be merged. Subtle breakages may occur due to fixing re-renders (eg. where a feature was inadvertently dependent on the buggy re-render logic)

----

## Summary

- add firebase helper scripts
- add `useUserLastSeenLimit` (extracted from `useConnectPartyGoers`)
- add selectors
  - 'core' selectors for firestore: `firestoreDataSelector`, `firestoreOrderedSelector`, `firestoreStatusSelector`
  - `authSelector`, `profileSelector`, `currentVenueSelector`, `partygoersSelector`, `currentEventSelector`, `userPurchaseHistorySelector`, `shouldRetainAttendanceSelector`, `isCurrentVenueRequestedSelector`, `isCurrentEventRequestedSelector`, `isUserPurchaseHistoryRequestedSelector `
- add 'core' 'make selector' helpers for firestore
- refactor `Firestore` types to separately export `FirestoreStatus`, `FirestoreData`, `FirestoreOrdered` types + sort keys alphabetically + add `ValidFirestoreKeys` type
- add `useSparkleFirestoreConnect()` hook that uses `ValidFirestoreKeys` to constrain the allowed values for `storeAs`
- refactor `useConnectPartyGoers()` to also export selected `partygoers` data
- refactor a number of components to replace inline selectors with external 'static reference' selectors
  - including `VenuePage`, etc
- refactor a number of components/hooks/helpers/etc to use named exports + deprecate their `default` exports
  - note: refactoring all usages of the default exports was considered OOS/tech debt for future cleanup
- cleanup imports on legacy `PartyMap` component + related components (`RoomModal`, `Map`, `PartyTitle`, `RoomAttendance`, `RoomCard`, `RoomList`, `RoomModalOngoingEvent`, `ScheduleItem`)
  - exported all named components from `src/components/templates/PartyMap/components/index.ts`
- cleanup imports on `VenuePage`
- probably a bunch more things now... check the commits for a full list..

You can also look at the [commits](https://github.com/sparkletown/sparkle/pull/721/commits), which should be fairly descriptive/self contained.

### DIFF
--- a/src/hooks/useSparkleFirestoreConnect.ts
+++ b/src/hooks/useSparkleFirestoreConnect.ts
@@ -1,0 +1,37 @@
+import {
+  ReduxFirestoreQuerySetting,
+  useFirestoreConnect,
+} from "react-redux-firebase";
+import { ValidFirestoreKeys } from "types/Firestore";
+
+/**
+ * This type allows us to automagically constrain the storeAs
+ * parameter to keys that exist within ValidFirestoreKeys, ensuring
+ * that we can't forget to define it in our types when using functions
+ * that rely on this type (eg. useSparkleFirestoreConnect)
+ *
+ * @see ValidFirestoreKeys
+ * @see useSparkleFirestoreConnect
+ * @see ReduxFirestoreQuerySetting
+ */
+export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
+  storeAs?: ValidFirestoreKeys;
+}
+
+/**
+ * A wrapper for useFirestoreConnect() that ensures the config
+ * we pass it can only use a storeAs key that has been defined
+ * in our FirestoreData/FirestoreOrdered types.
+ *
+ * Note: the config does NOT need to be memo'd before being passed
+ * to this function as useSparkleFirestoreConnect() determines equality
+ * through a deep comparison using lodash's isEqual() function.
+ *
+ * @param config the config to be passed to useFirestoreConnect()
+ *
+ * @see SparkleRFQConfig
+ * @see ValidFirestoreKeys
+ * @see ReduxFirestoreQuerySetting
+ */
+export const useSparkleFirestoreConnect = (config: SparkleRFQConfig[]) =>
+  useFirestoreConnect(config);
--- a/src/hooks/useSparkleFirestoreConnect.ts
+++ b/src/hooks/useSparkleFirestoreConnect.ts
@@ -1,8 +1,15 @@
 import {
   ReduxFirestoreQuerySetting,
   useFirestoreConnect,
+  isLoaded as _isLoaded,
+  isEmpty as _isEmpty,
 } from "react-redux-firebase";
 import { ValidFirestoreKeys } from "types/Firestore";
+
+/**
+ * Type helper representing all types of T except undefined
+ */
+export type Defined<T> = T & Exclude<T, undefined>;
 
 /**
  * This type allows us to automagically constrain the storeAs
@@ -35,3 +42,18 @@ export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
  */
 export const useSparkleFirestoreConnect = (config: SparkleRFQConfig[]) =>
   useFirestoreConnect(config);
+
+/**
+ * Use react-redux-firestore's isEmpty helper with
+ * user-defined type guards to properly narrow types
+ * when using this helper.
+ *
+ * @param item item fetched by react-redux-firestore
+ */
+export const hasData = <T>(item: T): item is Defined<T> =>
+  !_isEmpty(item) && item !== undefined;
+
+/**
+ * Re-export react-redux-firestore's isLoaded helper for convenience.
+ */
+export const isLoaded = <T>(item: T) => _isLoaded(item);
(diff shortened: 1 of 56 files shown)


thecadams (approved):
LGTM, good cleanup!

0xdevalias:
@thecadams I moved this back to draft because it seems to break the app. I believe the old logic was relying on some of the buggy re-rendering behaviour that should now be fixed. Which leaves `VenuePage` spinning on the `Loading...` component.

Once I figure out the fixes to make it work properly I will un-draft mode this PR and submit it for review again.

src/hooks/useSparkleFirestoreConnect.ts
```diff
@@ -0,0 +1,37 @@
...
+ * through a deep comparison using lodash's isEqual() function.
+ *
+ * @param config the config to be passed to useFirestoreConnect()
+ *
+ * @see SparkleRFQConfig
+ * @see ValidFirestoreKeys
+ * @see ReduxFirestoreQuerySetting
+ */
+export const useSparkleFirestoreConnect = (config: SparkleRFQConfig[]) =>
+  useFirestoreConnect(config);
```
0xdevalias, line 37: Calling these out so people are aware of them, they should probably be used as our canonical go to replacement for `useFirestoreConnect()`

src/types/Firestore.ts
```diff
@@ -1,88 +1,90 @@
...
+  data: FirestoreData;
+  ordered: FirestoreOrdered;
+  status: FirestoreStatus;
+}
+
+export interface FirestoreStatus {
+  requesting: Record<ValidFirestoreKeys, boolean>;
+  requested: Record<ValidFirestoreKeys, boolean>;
+  timestamps: Record<ValidFirestoreKeys, number>;
+}
```
0xdevalias, line 43: Note that these types are now automagically constructed for us, so long as we correctly add an entry to both `FirestoreData` and `FirestoreOrdered`.

src/types/Firestore.ts
```diff
@@ -1,88 +1,90 @@
...
+
+// note: these entries should be sorted alphabetically
+export interface FirestoreData {
```
0xdevalias, line 46: Keeping these sorted alphabetically should make it easier when looking for keys

src/types/Firestore.ts
```diff
@@ -1,88 +1,90 @@
...
+  userPurchaseHistory: Record<string, Purchase>;
+  userRoles: Record<string, Role>;
+  users: Record<string, User>;
+  venueChats: Record<string, RestrictedChatMessage> | null;
+  venueEvents: Record<string, VenueEvent>;
+  venues?: Record<string, AnyVenue>;
+}
+
+// note: these entries should be sorted alphabetically
+export interface FirestoreOrdered {
```
0xdevalias, line 70: Keeping these sorted alphabetically should make it easier when looking for keys

0xdevalias (commented):
Calling out some of the new additions I made + patterns they enable.

0xdevalias:
Ok, I believe this is fixed now. The commit that touched the most things, and that I haven't had the chance to go through and fully test all of the code paths for is https://github.com/sparkletown/sparkle/pull/721/commits/f7798176b0d77c28653247e15ed8466815cf2c10

Basically, in using `useConnectCurrentVenueNG()` to fix the some of the firestore connection, I had to go through and update any of the selectors that were expecting the data to be there from the old connect.

I think the first step I would take in testing this is to check that all of the updated components are used as children to a component that uses `useConnectCurrentVenueNG()` (currently that should only be `VenuePage`), or that those components that use `useConnectCurrentVenueNG()` have been rendered before them.

The next 'real world test' step to be certain would be to figure out all of the code paths where one of those components is rendered, and to access a page that will hit it in the web browser, and check that it all still works as expected.

I can look at doing this tomorrow if noone else is able to beforehand.

----

A future improvement to this pattern would be to use `createContext` and connect them that way + some nicer output if the context isn't available, but OOS'd it for the time being (at least in this PR)

src/pages/FriendShipPage/index.ts
```diff
@@ -1 +1,6 @@
+export { FriendShipPage } from "./FriendShipPage";
```
denisdimitrov, line 1: Usually named exports are being used when you need to import multiple values from a single file. In this case we always have a single component. It's almost like you are destructuring an object with only 1 key. Just pointing out.

denisdimitrov (approved):
LGTM

src/pages/FriendShipPage/index.ts
0xdevalias: _nods_ this was an explicit choice. I've read a few things over the years that suggest using default exports leads to more potential issues/etc. A few links from a quick skim of google:

- https://basarat.gitbook.io/typescript/main-1/defaultisbad
- https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad
- https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

Obviously this speaks to a stylistic/etc choice, and we should look at the evidence for/against as a team (when we have time) and decide what we want our canonical choice here to be.

PR #723, opened 2020-11-17
Venue parent added to owner list

IvanBuljovcic:
* If parentId is passed as part of URL query , it is included as the default venue parent in the details form
* When a venue is created/edited so that it has a _venue parent_ that venues owner(s) is added to the owners array of the current venue

### DIFF
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -68,6 +68,7 @@ export type VenueInput = AdvancedVenueInput &
     height?: number;
     bannerMessage?: string;
     parentId?: string;
+    owners?: string[];
   };
 
 type FirestoreVenueInput = Omit<VenueInput, VenueImageFileKeys> &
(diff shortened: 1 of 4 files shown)


src/api/admin.ts
```diff
@@ -88,6 +89,14 @@ export type PlacementInput = {
 export const createUrlSafeName = (name: string) =>
   name.replace(/\W/g, "").toLowerCase();
 
+const getVenueOwners = async (venueId: string): Promise<string[]> => {
```
thecadams, line 92: Can we obtain this from the `currentVenue` and rely on snapshots, instead of getting it once and risking it being out-of-date?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -75,7 +79,10 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
         template: state.templatePage?.template,
         editing: !!venueId,
       },
-      defaultValues,
+      defaultValues: {
+        ...defaultValues,
+        parentId: parentIdQuery ?? "",
```
thecadams, line 84: Possibly should get this from `defaultValues`:

`parentId ?? defaultValues?.parentId ?? ""`

thecadams (commented):
Looking good so far, couple of comments on it for you take a look at.

thecadams (approved):

PR #726, opened 2020-11-17
'Large Party Map' Performance - Part 1/x (FIX)

0xdevalias:
## Background

The original PR for these changes was https://github.com/sparkletown/sparkle/pull/721

As noted in https://github.com/sparkletown/sparkle/pull/721#issuecomment-728864894 it was likely that there were unintended consequences of some of the changes, and I needed to look into it more.

It was merged in https://github.com/sparkletown/sparkle/pull/722, but ended up causing issues in production ([ref](https://trello.com/c/RCnzq8Eq): "due seemingly to deployment of party map performance step 1, the jazz bar is showing up black"), so was reverted in https://github.com/sparkletown/sparkle/pull/724

This PR re-instates the original changes from https://github.com/sparkletown/sparkle/pull/721 and will also contain the additional required fixes.

## Solution

I've decided to de-risk this PR by reverting https://github.com/sparkletown/sparkle/commit/f7798176b0d77c28653247e15ed8466815cf2c10 in it (which can be split off and handled as a separate PR), since it touched so many components, and is most likely where the issues that were seen in production were introduced.

---

## TODO

- [x] Ensure that jazz bar venues work as expected now that we've reduced the scope of changes in this PR (see https://github.com/sparkletown/sparkle/pull/726#issuecomment-729380451)

## Future Improvements (OOS of this PR)

- [ ] Create a new branch/PR that undoes the revert in https://github.com/sparkletown/sparkle/pull/726/commits/cbcd3ea0bc25e0d6bd733f9fb505ce52150e3ad5 so we can fix any bugs introduced as a separate PR

**EDIT:** The below TODO items are outdated, i've decided to de-risk this PR by moving the potentially breaking changes that touched lots of things into it's own PR. Will move these across to that when I get to it.

- [ ] Identify the 'code paths' of components that were affected by the changes in https://github.com/sparkletown/sparkle/pull/721/commits/f7798176b0d77c28653247e15ed8466815cf2c10 (as per https://github.com/sparkletown/sparkle/pull/721#issuecomment-728864894)
  - > I think the first step I would take in testing this is to check that all of the updated components are used as children to a component that uses `useConnectCurrentVenueNG()` (currently that should only be `VenuePage`), or that those components that use `useConnectCurrentVenueNG()` have been rendered before them.
  - > [...and] figure out all of the code paths where one of those components is rendered
- [ ] Create/identify appropriate URLs to access/render each of these code paths and observe how they work (or not) in the browser when rendered
  - > The next 'real world test' step to be certain would be [...] to access a page that will hit it in the web browser, and check that it all still works as expected.
- [ ] For any that have issues, dive deeper into what the specific issues are and fix the bugs.
- [ ] ???
- [ ] [Profit!](https://knowyourmeme.com/memes/profit)

### DIFF
--- a/src/components/templates/PartyMap/components/RoomCard/RoomCard.tsx
+++ b/src/components/templates/PartyMap/components/RoomCard/RoomCard.tsx
@@ -1,9 +1,10 @@
 import React from "react";
-import "./RoomCard.scss";
-import RoomAttendance from "../RoomAttendance";
-import { formatMinute } from "utils/time";
+
+import { formatMinute, getCurrentEvent } from "utils/time";
 import { RoomData } from "types/RoomData";
-import { getCurrentEvent } from "utils/time";
+
+import { RoomAttendance } from "..";
+import "./RoomCard.scss";
 
 interface PropsType {
   startUtcSeconds: number;
@@ -12,7 +13,7 @@ interface PropsType {
   onClick: () => void;
 }
 
-const RoomCard: React.FunctionComponent<PropsType> = ({
+export const RoomCard: React.FunctionComponent<PropsType> = ({
   startUtcSeconds,
   room,
   attendance,
@@ -56,5 +57,3 @@ const RoomCard: React.FunctionComponent<PropsType> = ({
     </div>
   );
 };
-
-export default RoomCard;
--- a/src/components/templates/PartyMap/components/RoomList/RoomList.tsx
+++ b/src/components/templates/PartyMap/components/RoomList/RoomList.tsx
@@ -1,9 +1,10 @@
 import React from "react";
-import "./RoomList.scss";
 
-import { RoomData } from "types/RoomData";
-import RoomCard from "../RoomCard";
 import { eventHappeningNow } from "utils/time";
+import { RoomData } from "types/RoomData";
+
+import { RoomCard } from "..";
+import "./RoomList.scss";
 
 interface PropsType {
   startUtcSeconds: number;
@@ -13,7 +14,7 @@ interface PropsType {
   setIsRoomModalOpen: (value: boolean) => void;
 }
 
-const RoomList: React.FunctionComponent<PropsType> = ({
+export const RoomList: React.FunctionComponent<PropsType> = ({
   startUtcSeconds,
   rooms,
   attendances,
@@ -48,5 +49,3 @@ const RoomList: React.FunctionComponent<PropsType> = ({
     </>
   );
 };
-
-export default RoomList;
--- a/src/hooks/useConnectCurrentVenue.ts
+++ b/src/hooks/useConnectCurrentVenue.ts
@@ -1,8 +1,16 @@
 import { useParams } from "react-router-dom";
 import { useFirestoreConnect } from "react-redux-firebase";
+
+import { currentVenueSelector } from "utils/selectors";
 import getQueryParameters from "utils/getQueryParameters";
 
-const useConnectCurrentVenue = () => {
+import { useSelector } from "./useSelector";
+
+/**
+ * @deprecated use useConnectCurrentVenueNG instead
+ * @see useConnectCurrentVenueNG
+ */
+export const useConnectCurrentVenue = () => {
   let { venueId } = useParams();
   if (!venueId) {
     venueId = getQueryParameters(window.location.search)?.venueId;
@@ -25,6 +33,11 @@ const useConnectCurrentVenue = () => {
       storeAs: "venueEvents",
     },
   ]);
+
+  return useSelector(currentVenueSelector);
 };
 
+/**
+ * @deprecated use named export instead
+ */
 export default useConnectCurrentVenue;
--- a/src/types/Firestore.ts
+++ b/src/types/Firestore.ts
@@ -1,88 +1,94 @@
-import { Venue } from "./Venue";
-import { Purchase } from "./Purchase";
-import { User } from "./User";
+import { WithId } from "utils/id";
+import { AdminRole } from "hooks/roles";
+
 import {
   RestrictedChatMessage,
   PrivateChatMessage,
 } from "components/context/ChatContext";
-import { VenueEvent } from "./VenueEvent";
-import { Table } from "./Table";
-import { PartyMapVenue } from "./PartyMapVenue";
 import { Reaction } from "components/context/ExperienceContext";
-import { WithId } from "utils/id";
+
 import { CampVenue } from "./CampVenue";
 import { ChatRequest } from "./ChatRequest";
+import { PartyMapVenue } from "./PartyMapVenue";
+import { Purchase } from "./Purchase";
 import { Role } from "./Role";
-import { AdminRole } from "hooks/roles";
+import { Table } from "./Table";
+import { User } from "./User";
+import { Venue } from "./Venue";
+import { VenueEvent } from "./VenueEvent";
 
-interface VenueStatus {
-  currentVenue: boolean;
-  currentEvent: boolean;
-  eventPurchase: boolean;
-  venueChats: boolean;
-  venueEvents: boolean;
-  userPurchaseHistory: boolean;
-}
+export type AnyVenue = Venue | PartyMapVenue | CampVenue;
 
 interface Experience {
   reactions: Record<string, Reaction>;
   tables: Record<string, Table>;
 }
 
-type VenueTimestamps = Record<keyof VenueStatus, number>;
-export type AnyVenue = Venue | PartyMapVenue | CampVenue;
-
 interface UserVisit {
   timeSpent: number;
 }
 
+export type ValidFirestoreKeys = keyof FirestoreData | keyof FirestoreOrdered;
+
 export interface Firestore {
-  status: {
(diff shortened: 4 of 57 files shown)


src/hooks/useConnectCurrentVenue.ts
```diff
@@ -1,8 +1,16 @@
...
 import getQueryParameters from "utils/getQueryParameters";
 
-const useConnectCurrentVenue = () => {
+import { useSelector } from "./useSelector";
+
+/**
+ * @deprecated use useConnectCurrentVenueNG instead
+ * @see useConnectCurrentVenueNG
+ */
+export const useConnectCurrentVenue = () => {
```
0xdevalias, line 13: Given the issues we saw introduced in https://github.com/sparkletown/sparkle/commit/f7798176b0d77c28653247e15ed8466815cf2c10 (within https://github.com/sparkletown/sparkle/pull/721), we need to be careful about switching things across to use this, so it may not be so much "don't use this, use the other" at the moment.

src/types/Firestore.ts
```diff
@@ -1,88 +1,94 @@
...
+  data: FirestoreData;
+  ordered: FirestoreOrdered;
+  status: FirestoreStatus;
+}
+
+export interface FirestoreStatus {
+  requesting: Record<ValidFirestoreKeys, boolean>;
+  requested: Record<ValidFirestoreKeys, boolean>;
+  timestamps: Record<ValidFirestoreKeys, number>;
+}
```
0xdevalias, line 43: copying from original PR ([ref](https://github.com/sparkletown/sparkle/pull/721/files?file-filters%5B%5D=.ts&hide-deleted-files=true#r524929410))

Note that these types are now automagically constructed for us, so long as we correctly add an entry to both `FirestoreData` and `FirestoreOrdered`.

src/types/Firestore.ts
```diff
@@ -1,88 +1,94 @@
...
+
+// note: these entries should be sorted alphabetically
+export interface FirestoreData {
```
0xdevalias, line 46: copying from original PR ([ref](https://github.com/sparkletown/sparkle/pull/721/files?file-filters%5B%5D=.ts&hide-deleted-files=true#r524930318))

Keeping these sorted alphabetically should make it easier when looking for keys

src/types/Firestore.ts
```diff
@@ -1,88 +1,94 @@
...
+  userPurchaseHistory: Record<string, Purchase>;
+  userRoles: Record<string, Role>;
+  users: Record<string, User>;
+  venueChats: Record<string, RestrictedChatMessage> | null;
+  venueEvents: Record<string, VenueEvent>;
+  venues?: Record<string, AnyVenue>;
+}
+
+// note: these entries should be sorted alphabetically
+export interface FirestoreOrdered {
```
0xdevalias, line 72: copying from original PR ([ref](https://github.com/sparkletown/sparkle/pull/721/files?file-filters%5B%5D=.ts&hide-deleted-files=true#r524930363))

Keeping these sorted alphabetically should make it easier when looking for keys

src/types/Firestore.ts
```diff
@@ -1,88 +1,94 @@
...
+import { VenueEvent } from "./VenueEvent";
 
-interface VenueStatus {
-  currentVenue: boolean;
-  currentEvent: boolean;
-  eventPurchase: boolean;
-  venueChats: boolean;
-  venueEvents: boolean;
-  userPurchaseHistory: boolean;
-}
```
0xdevalias, line 25: This is no longer needed as we generate it with TypeScript helpers automagically. See below.

src/types/Firestore.ts
```diff
@@ -1,88 +1,94 @@
...
-type VenueTimestamps = Record<keyof VenueStatus, number>;
-export type AnyVenue = Venue | PartyMapVenue | CampVenue;
-
 interface UserVisit {
   timeSpent: number;
 }
 
+export type ValidFirestoreKeys = keyof FirestoreData | keyof FirestoreOrdered;
+
 export interface Firestore {
```
0xdevalias, line 33: This was de-structured to make separate types for `FirestoreStatus`, `FirestoreData`, `FirestoreOrdered`

0xdevalias (commented):
Added a few comments/pointers to help with review.

Also note: of the 47 changed files, only ~9 of them have any large/significant changes, and only ~4 or so of those have any significant code changes (beyond additions of new code). So while it make look big and scary to review, I promise it shouldn't be too bad!

0xdevalias:
## Notes from clickthrough testing to ensure `jazzbar`/etc work again now

**Issue:** When trying to access a non-existent venue (eg. http://localhost:3000/in/doesnotexist) we are stuck on `Loading...`

**Production:** checked against production, and this is true there as well, so not introduced by changes in this PR. The refactor/changes in https://github.com/sparkletown/sparkle/commit/f7798176b0d77c28653247e15ed8466815cf2c10 (that will be split into a new PR) fixed this (by accident). I believe the main relevant part was using `react-redux-firebase`'s `isLoaded`/`isEmpty` helpers, rather than using `state.firestore.status.*` things directly.

- https://github.com/prescottprue/react-redux-firebase/blob/master/docs/api/helpers.md

---

Accessing a `jazzbar` seems to work properly ✅

thecadams (commented):
Looks good - great stuff here with the `useSelector` improvements.

Just curious - what's the reason we're doing `useConnectCurrentVenuNG`? Is that to reduce risk we crash the site again?

0xdevalias:
> @thecadams Just curious - what's the reason we're doing useConnectCurrentVenuNG? Is that to reduce risk we crash the site again?

In case you didn't see it, my comment on the previous PR: https://github.com/sparkletown/sparkle/pull/721#issuecomment-728864894

My original desire was to explicitly pass the `venueId` into the hook `useConnectCurrentVenue()`, rather than relying on the somewhat hacky feeling logic to pull it out of the url. Another aim of it was to make a hook that provided similar functionalities as `useConnectCurrentVenue()`, but done in a way to make it easier to use, harder to use wrong, etc. I decided to OOS it at the time, but my thoughts was that we could probably wire it up with React's context providers to then make a more 'solid' connection between a 'firestore connect' config, and the components that want to access the state from that. (Just had an idea, by adding `useCurrentVenue()`, useCurrentVenueEvents()` hooks to that same file (or similar) that essentially wrap the canonical selectors would probably work with less overhead than context. OOS for now, future refinement.)

I decided to make a new hook (`useConnectCurrentVenueNG()`) to de-risk it, letting any older components stay using `useConnectCurrentVenue()`. That backfired a little due to the inline selector usages spread across the components, and that there wasn't an immediately simple way for me to enumerate and ensure which 'child components' would be rendered after `VenuePage` (which is where I was using `useConnectCurrentVenueNG()` in https://github.com/sparkletown/sparkle/commit/f7798176b0d77c28653247e15ed8466815cf2c10 (which is what broke production when it was merged, and what was left out for this PR to de-risk it further)

(accidental deep-dive into my thinking/processes around this.. hope it helped, let me know if you have any other questions around it)

thecadams (approved):
Got it, makes sense. Thanks for the comments!

src/components/templates/PartyMap/components/RoomCard/RoomCard.tsx
```diff
@@ -1,9 +1,10 @@
 import React from "react";
-import "./RoomCard.scss";
-import RoomAttendance from "../RoomAttendance";
-import { formatMinute } from "utils/time";
+
+import { formatMinute, getCurrentEvent } from "utils/time";
 import { RoomData } from "types/RoomData";
-import { getCurrentEvent } from "utils/time";
+
+import { RoomAttendance } from "..";
```
denisdimitrov, line 6: Are you sure this is correct? I've never seen this syntax ".." before and I'm just asking.

src/components/templates/PartyMap/components/RoomList/RoomList.tsx
```diff
@@ -1,9 +1,10 @@
 import React from "react";
-import "./RoomList.scss";
 
-import { RoomData } from "types/RoomData";
-import RoomCard from "../RoomCard";
 import { eventHappeningNow } from "utils/time";
+import { RoomData } from "types/RoomData";
+
+import { RoomCard } from "..";
```
denisdimitrov, line 6: .

denisdimitrov (changes requested):

src/components/templates/PartyMap/components/RoomCard/RoomCard.tsx
0xdevalias: It's correct, but it may not be the best/easiest to read style. It's essentially accessing `../index.ts` (aka `src/components/templates/PartyMap/components/index.ts`)

Something we can add to our code style discussions in future, and decide on our official standpoint (or let an opinionated linter do it for us)

src/components/templates/PartyMap/components/RoomList/RoomList.tsx
0xdevalias: Will resolve this one as above, but will leave the above open for future reference.

PR #729, opened 2020-11-18
Deploy 20201118

denisdimitrov:
Deploying the following changes:

- fix for the popup positioning


src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -50,33 +50,34 @@ import UpcomingTickets from "components/molecules/UpcomingTickets";
...
+const ProfilePopover = (
   <Popover id="profile-popover">
     <Popover.Content>
       <ProfilePopoverContent />
     </Popover.Content>
   </Popover>
 );
 
-const GiftPopover: React.FC = () => (
+const GiftPopover = (
```
0xdevalias, line 80: I wouldn't have changed these. As they were they were react subcomponents.

What was the thoughts behind changing them to this style?

PR #731, opened 2020-11-19
Beef up linting

thecadams:
A little one I've wanted to do for a while. Recreation of https://github.com/sparkletown/sparkle/pull/665 without all the other files changed yet


thecadams:
@0xdevalias @denisdimitrov @IvanBuljovcic PTAL

thecadams:
@thecadams test

thecadams:
@0xdevalias you able to take a look at this one?

thecadams:
@0xdevalias @thecadams testing GH action

.circleci/config.yml
```diff
@@ -14,6 +14,7 @@ jobs:
             - run: npm run init
             - run: npm run lint
```
0xdevalias, line 15: Currently `lint` is defined as:

```
"lint": "prettier --write \"{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
```

We probably don't want to be running `prettier`'s `--write` in CI.

Propose that we add change `scripts` to something like:

```
// ..snip..
    "lint": "npm-run-all eslint:src prettier:lint",
    "eslint:src": "eslint src --ext .js,.jsx,.ts,.tsx",
    "eslint:fix": "eslint --fix ?something like the above?"
    "prettier": "prettier --write \"{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
    "prettier:lint": "prettier \"{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
    "test": "react-scripts test",
// ..snip..
```

- https://www.npmjs.com/package/npm-run-all

.eslintrc
```diff
@@ -1,19 +1,37 @@
 {
```
0xdevalias, line 1: [OOS/Future] Would be cool to switch this over to the eslint JS config style

- https://eslint.org/docs/user-guide/configuring

.eslintrc
```diff
@@ -1,19 +1,37 @@
...
+    "plugin:prettier/recommended",
     "plugin:react/recommended",
     "plugin:@typescript-eslint/eslint-recommended",
-    "plugin:react-hooks/recommended"
+    "plugin:react-hooks/recommended",
+    "prettier",
+    "prettier/react",
+    "prettier/@typescript-eslint",
+    "react-app"
   ],
```
0xdevalias, line 35: These sound good as a starting point, not deeply familiar with them, but let's do it!

- https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
  - > This plugin works best if you disable all other ESLint rules relating to code formatting, and only enable rules that detect patterns in the AST. (If another active ESLint rule disagrees with prettier about how code should be formatted, it will be impossible to avoid lint errors.) You can use `eslint-config-prettier` to disable all formatting-related ESLint rules.
  - > If your desired formatting does not match the prettier output, you should use a different tool such as `prettier-eslint` instead.

Are we using `eslint-config-prettier` here (as recommended in the above)?

- https://nicedoc.io/prettier/eslint-config-prettier

0xdevalias: https://prettier.io/docs/en/integrating-with-linters.html mentions:

- `eslint-config-prettier`
- `tslint-config-prettier`
- `stylelint-config-prettier`

The notes section specifically talks about best practices, what to use, what is less optimal to use, etc:

- https://prettier.io/docs/en/integrating-with-linters.html#notes

.eslintrc
```diff
@@ -27,6 +45,12 @@
...
     ],
     "react/prop-types": "off",
-    "react/display-name": "warn"
+    "react/display-name": "warn",
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
+      }
+    ]
```
0xdevalias, line 54: Google-fu is failing me, where are the options for this documented? And specifically, what does this option do?

package.json
```diff
@@ -111,18 +111,29 @@
...
     "eslint-plugin-jsx-a11y": "^6.3.1",
+    "eslint-plugin-prettier": "^3.1.4",
     "eslint-plugin-react": "^7.20.5",
     "eslint-plugin-react-hooks": "^4.0.8",
+    "eslint-webpack-plugin": "^1.0.0",
     "firebase-admin": "^9.1.1",
     "firebase-tools": "^8.6.0",
     "gifwrap": "^0.9.2",
     "phin": "^3.5.0",
+    "prettier": "^2.1.2",
```
0xdevalias, line 136: Yay for latest!

package.json
```diff
@@ -111,18 +111,29 @@
...
+    "@typescript-eslint/parser": "^4.8.1",
     "cypress": "^4.11.0",
     "cypress-file-upload": "^4.0.7",
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb": "^18.2.1",
+    "eslint-config-airbnb-typescript": "^12.0.0",
+    "eslint-config-prettier": "^6.15.0",
     "eslint-config-react-app": "^5.2.1",
+    "eslint-loader": "^3.0.3",
+    "eslint-plugin-flowtype": "^4.6.0",
```
0xdevalias, line 124: Do we need this? We don't use Flow. There may be some nice to have rules in there that are independent of it, but also wonder if that would conflict with prettier/etc things. Feels like adding extra risk at this stage, maybe can explore it as a separate PR in future if it makes sense to?

Docs mention needing to add `babel-eslint` added for this to work
- https://github.com/gajus/eslint-plugin-flowtype#installation

package.json
```diff
@@ -91,7 +91,7 @@
     ]
   },
   "lint-staged": {
```
0xdevalias, line 93: Is this a script, or it's own separate thing?

0xdevalias: - https://github.com/okonet/lint-staged
  - https://github.com/okonet/lint-staged#configuration
    - > Starting with v3.1 you can now use different ways of configuring it:
      - > `lint-staged` object in your `package.json`
      - > `.lintstagedrc` file in JSON or YML format, or you can be explicit with the file extension: `.lintstagedrc.json`
      - > `.lintstagedrc.yaml`
      - > `.lintstagedrc.yml`
      - > `lint-staged.config.js`, `.lintstagedrc.js`, or `.lintstagedrc.cjs` file in JS format

All good!

[OOS/Future] could potentially be nice to move it to one of those file types like `.lintstagedrc` or `lint-staged.config.js` (my personal preference)

package.json
```diff
@@ -111,18 +111,29 @@
...
     "cypress": "^4.11.0",
     "cypress-file-upload": "^4.0.7",
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb": "^18.2.1",
+    "eslint-config-airbnb-typescript": "^12.0.0",
+    "eslint-config-prettier": "^6.15.0",
     "eslint-config-react-app": "^5.2.1",
+    "eslint-loader": "^3.0.3",
+    "eslint-plugin-flowtype": "^4.6.0",
+    "eslint-plugin-import": "^2.22.1",
```
0xdevalias, line 125: - https://www.npmjs.com/package/eslint-plugin-import#rules

Big yes!

- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-internal-modules.md
  - could be interesting
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-useless-path-segments.md
  - definitely
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-relative-parent-imports.md
  - probably not
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-named-as-default.md
  - unsure
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-named-as-default-member.md
  - unsure
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-deprecated.md
  - warn only at this stage
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-extraneous-dependencies.md
  - sounds good
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/exports-last.md
  - unsure how this will work with `export const foo = () =>` style (which I like)
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/extensions.md
  - yes, never use extensions (except for scss/svg/etc)
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/order.md
  - this is awesome! will help a lot. need to decide our canonical choices
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/newline-after-import.md
  - definitely
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/prefer-default-export.md
  - i don't like this
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-named-default.md
  - this is probably fine
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-default-export.md
  this would potentially be my preference, open to discussion around why we should/shouldn't though
- https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/group-exports.md
  - definitely

0xdevalias (commented):
Added a bunch of comments, some of which may require changes/etc.

There are probably some stylistic things in here that may require amending/discussion for/against too, but that shouldn't necessarily block this (may just cause some churn in a future PR) (eg. I would love to change some of our prettier settings at the moment)

Would be good to know what sort of things this current setup calls as errors across our codebase (though I guess I can clone and run it to see that)

.eslintrc
```diff
@@ -1,19 +1,37 @@
...
+    "airbnb-typescript",
+    "airbnb/hooks",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:jest/recommended",
+    "plugin:prettier/recommended",
     "plugin:react/recommended",
     "plugin:@typescript-eslint/eslint-recommended",
-    "plugin:react-hooks/recommended"
+    "plugin:react-hooks/recommended",
+    "prettier",
```
0xdevalias, line 31: My ideal tweaks to our current `prettier` config:

- https://prettier.io/docs/en/options.html
- https://prettier.io/docs/en/options.html#semicolons
  - turn this off
- https://prettier.io/docs/en/options.html#quotes
  - turn this on
- https://prettier.io/docs/en/options.html#jsx-quotes
  - turn this on
- https://prettier.io/docs/en/options.html#trailing-commas
  - es5
- https://prettier.io/docs/en/options.html#arrow-function-parentheses
  - i prefer 'omit when possible' for this.. though that was in plain JS. If we need to manually add type annotations then this would be ok, but since a lot of the time we can allow it to infer, I still think I prefer avoiding them when possible

0xdevalias:
For future note, when we look at using GitHub actions for things, there is this tool:

- https://github.com/github/super-linter

.eslintrc
thecadams: I'm not 100% certain the difference between `plugins` & `extends`, but we may as well get as strict as reasonably possible, and insist that we gradually move toward zero lint issues..

.eslintrc
thecadams: Looks like when I cargo-culted this section, I managed to bring in a CR/LF fix: https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-delete-cr-prettier-prettier

package.json
thecadams: This part is crucial - it's a performance optimization so husky only runs over the files staged in the index, not the whole repo. Makes commits go from 5-7 seconds to a few milliseconds. Not really related to the linting, but good catch.

.eslintrc
thecadams: Sounds good! Would you be OK to add those after we land this PR? Also, how can we make sure prettier's autoformat-on-save for vscode will honour these tweaks?

thecadams:
Going to recreate this a second time, now that we have a baseline linting setup.

PR #727, opened 2020-11-18
'Large Party Map' Performance - Part 2/x

0xdevalias:
Part 1: https://github.com/sparkletown/sparkle/pull/726 (though this PR isn't directly dependant on it)

This PR contains a bunch of smaller cleanup parts as part of improving the performance of `src/components/templates/Camp/components/Map.tsx`. I decided to split them out into this PR to minimize the complexity of the bigger changes to come.

Main changes:

- 5447ea43 add useCampPartygoers() hook + minor cleanup in useConnectPartyGoers()
- 5eb19306 refactor Camp to use useCampPartygoers() hook + static selector reference for campVenueSelector
- refactor isExternalUrl, getRoomUrl, openRoomUrl, openUrl into utils + updates references
- add currentVenueSelectorData, orderedVenuesSelector + refactor components to use them + some minor cleanup on the way

See [commits](https://github.com/sparkletown/sparkle/pull/727/commits) for the rest.

PR #732, opened 2020-11-19
'Large Party Map' Performance - Part 3/x

0xdevalias:
**Related PR's:** https://github.com/sparkletown/sparkle/pull/726 (part 1), https://github.com/sparkletown/sparkle/pull/727 (part 2)

This PR focuses on splitting up `src/components/templates/Camp/components/Map.tsx` into more manageable 'related chunks', so that we can then properly fix the performance.

**Note:** this is still a WIP, but it should (mostly?) render/still work in it's current state, just won't have all performance fixes done yet.


src/components/templates/Camp/components/Map.tsx
```diff
@@ -109,495 +90,156 @@ export const Map: React.FC<MapProps> = ({
...
-                            borderRadius: "100%",
-                            alignSelf: "center",
-                            backgroundImage: `url(${partygoer?.pictureUrl})`,
-                            backgroundSize: "cover",
-                          }}
-                          avatarClassName={`${
-                            isMe ? "me profile-avatar" : "profile-avatar"
-                          }`}
-                          setSelectedUserProfile={setSelectedUserProfile}
-                          miniAvatars={venue?.miniAvatars}
```
0xdevalias, line 455: this was refactored into `src/components/templates/Camp/components/MapPartygoersOverlay.tsx`

src/components/templates/Camp/components/Map.tsx
```diff
@@ -109,495 +90,156 @@ export const Map: React.FC<MapProps> = ({
...
-                      onClick={(e) => onJoinRoom(e, room)}
-                    >
-                      {venue.joinButtonText ?? "Join now"}
-                    </div>
-                  </div>
-                  <div className="camp-venue-img">
-                    <img
-                      src={room.image_url}
-                      title={room.title}
-                      alt={room.title}
```
0xdevalias, line 525: this was refactored into `src/components/templates/Camp/components/MapRoomsOverlay.tsx`

src/components/templates/Camp/components/Map.tsx
```diff
@@ -109,495 +90,156 @@ export const Map: React.FC<MapProps> = ({
...
-                    />
-                  </div>
-                  {venue.roomVisibility === RoomVisibility.hover &&
-                    roomHovered &&
-                    roomHovered.title === room.title && (
-                      <div className="camp-venue-text">
-                        <div className="camp-venue-maininfo">
-                          <div className="camp-venue-title">{room.title}</div>
-                          <CampAttendance
-                            attendances={attendances}
```
0xdevalias, line 535: this was refactored into `src/components/templates/Camp/components/MapRoomsOverlay.tsx`

src/components/templates/Camp/components/Map.tsx
```diff
@@ -109,495 +90,156 @@ export const Map: React.FC<MapProps> = ({
...
-                      venue.roomVisibility === RoomVisibility.nameCount ||
-                      (venue.roomVisibility === RoomVisibility.count &&
-                        hasAttendance)) && (
-                      <div className="camp-venue-maininfo">
-                        {(!venue.roomVisibility ||
-                          venue.roomVisibility ===
-                            RoomVisibility.nameCount) && (
-                          <div className="camp-venue-title">{room.title}</div>
-                        )}
-                        <CampAttendance
```
0xdevalias, line 554: this was refactored into `src/components/templates/Camp/components/MapRoomsOverlay.tsx`

src/components/templates/Camp/components/Map.tsx
```diff
@@ -109,495 +90,156 @@ export const Map: React.FC<MapProps> = ({
...
+              <MapRow
+                key={`row${rowIndex}`}
+                showGrid={venue.showGrid}
+                hasSeatedPartygoer={hasSeatedPartygoer}
+                seatedPartygoerIsMe={isMe}
+                // TODO: make this a submodule + useCallback()?
+                onSeatClick={() => onSeatClick(row, column, seatedPartygoer)}
+              />
             );
           })}
```
0xdevalias, line 592: this was refactored into `src/components/templates/Camp/components/MapRoomsOverlay.tsx`

src/components/templates/Camp/hooks/useKeyPress.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from "react";
+
+import { MOVEMENT_INTERVAL } from "./useKeyboardControls";
+
+// TODO: We can probably replace this with the mousetrap library
+//   https://craig.is/killing/mice
+export const useKeyPress = function (targetKey: string) {
```
0xdevalias, line 7: [TODO] see if mousetrap works better for this

- https://craig.is/killing/mice

Alternatively, using `useRef()` rather than `useState()` here may allow us to not break memoisation so often, but then we may need an alternative method of telling react to re-render when we actually want it to.

src/components/templates/Camp/hooks/useKeyboardControls.ts
```diff
@@ -0,0 +1,219 @@
...
+    },
+    [profile, user, venue, venues]
+  );
+
+  // TODO: this is duplicated logic from handlePotentialRoomClick in Map component, so we can probably simplify
+  // TODO: do we want to filter this? Or just handle 1 room hit
+  const hitRoom = useCallback(
+    (row: number, column: number) => {
+      let isHitting = false;
+      venue.rooms.forEach((room: CampRoomData) => {
```
0xdevalias, line 104: [TODO] see if we can de-duplicate this logic (and in doing so only have to process the rooms once rather than twice. If there are lots of rooms this could make a significant difference.

I was thinking of filtering the `venue.rooms` so that it only contains the list of 'hit' rooms, which in theory would only be one.

The current logic using `forEach` isn't going to stop when it finds a hit, which the rest of the code seems to imply it might want to do. So even if we just change this to use `.find()` or similar, could help:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/components/templates/Camp/hooks/useKeyboardControls.ts
```diff
@@ -0,0 +1,219 @@
...
+      isHittingRoom,
+      venue.rooms,
+      rows,
+      selectedRoom,
+      setIsHittingRoom,
+      setSelectedRoom,
+    ]
+  );
+
+  useEffect(() => {
```
0xdevalias, line 142: [TODO] splitting this up into multiple `useEffect()`'s would allow us to downscope the amount of things that can cause it to break memo, not sure if that will make a huge difference here or not though.

src/components/templates/Camp/hooks/useKeyboardControls.ts
```diff
@@ -0,0 +1,219 @@
...
+    if ((!currentPosition?.row && !currentPosition?.column) || keyDown) {
+      return;
+    }
+
+    const { row, column } = currentPosition;
+    if (row && column) {
+      const seatTaken = (r: number, c: number) => partygoersBySeat?.[r]?.[c];
+      if (enterPress && selectedRoom) {
+        setKeyDown(true);
+        setTimeout(() => setKeyDown(false), MOVEMENT_INTERVAL);
```
0xdevalias, line 154: [TODO] These `setTimeout`'s + multiple re-renders is most likely where the 'ghost movement' bug is coming from. The `mousetrap` library (mentioned above) is probably a decent solution to use here.

src/components/templates/Camp/hooks/useKeyboardControls.ts
```diff
@@ -0,0 +1,219 @@
...
+        return;
+      }
+      if (downPress) {
+        setKeyDown(true);
+        setTimeout(() => setKeyDown(false), MOVEMENT_INTERVAL);
+        if (row + 1 > DEFAULT_ROWS || seatTaken(row + 1, column)) {
+          return;
+        }
+        hitRoom(row + 1, column);
+        takeSeat(row + 1, column);
```
0xdevalias, line 166: These are some more examples of the arbitrary `+ 1` logic mentioned earlier, so if we can clean that up to be `0-indexed, then we can make all of this simpler too.

src/utils/reducers.ts
```diff
@@ -0,0 +1,28 @@
...
+  const col = selectCol(cur);
+
+  if (acc[row] === undefined) {
+    acc[row] = [];
+  }
+
+  acc[row][col] = cur;
+
+  return acc;
+};
```
0xdevalias, line 28: Calling this out mostly because I thought it was a really clean way to generically abstract the pattern we were using in `partygoersBySeat`.

0xdevalias (commented):
[FYI] I figured given how extensive the refactors here is that it might help to point out which chunks of code ended up where/similar.

I'll mark all of these 'pointer components' as resolved to reduce the spam a bit, but if you're looking at the diff and wondering where it ended up, there should hopefully be a comment pointing you in the right direction now!

This is my WIP state as of our quick sync this evening. I believe there are possibly some small bugs introduced here (eg. when I tried to demo it on the call it didn't seem as though the keyboard controls were working anymore). I also haven't applied a lot of the optimisations to it yet, that's my next step. But even without them, it seems as though it is already a bit more performant as it is.

src/utils/filter.ts
```diff
@@ -19,3 +23,43 @@
...
+
+/**
+ * Make a CampRoomData[] filter for rooms 'hit' by the supplied row/column.
+ *
+ * @param row
+ * @param column
+ * @param totalRows
+ * @param totalColumns
+ */
+export const makeCampRoomHitFilter = ({
```
thecadams, line 45: TODO: test this and see if hit-detection is working

src/utils/url.ts
```diff
@@ -40,3 +40,11 @@ export const openRoomUrl = (url: string) => {
 export const openUrl = (url: string) => {
   window.open(url, isExternalUrl(url) ? "_blank" : "noopener,noreferrer");
 };
+
+export const externalUrlAdditionalProps = {
+  target: "_blank",
+  rel: "noopener noreferrer",
+};
+
+export const getExtraLinkProps = (isExternal: boolean) =>
```
thecadams, line 49: great stuff! Lots more places we can use this in future

thecadams (approved):
lgtm, some stuff to test once we go to staging, looking good!

src/components/templates/Camp/components/MapRoomOverlay.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  const roomUnhovered = useCallback(() => {
+    dispatch(retainAttendance(false));
+    setRoomHovered(undefined);
+  }, [dispatch]);
+
+  // TODO: this is being spread here because the below code modifies it with .push()/etc
+  // const rooms = [...venue.rooms];
+
+  // TODO: what is the intent/purpose of this?
```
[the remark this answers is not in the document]
0xdevalias: The old code took a local copy of the array first, then modified it. So it could be trying to ensure the hovered/clicked/etc room is first in the array? Maybe?

src/components/templates/Camp/components/MapRoomOverlay.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const shouldShowHovered =
+    venue.roomVisibility === RoomVisibility.hover &&
+    isRoomHovered &&
+    isRoomHovered.title === room.title;
+
+  // TODO: what is a better semantic name for this?
+  const shouldShow1 =
+    !venue.roomVisibility || venue.roomVisibility === RoomVisibility.nameCount;
+
+  // TODO: what is a better semantic name for this?
```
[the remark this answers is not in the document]
0xdevalias: This was mostly a quick hack job to pull them out of the JSX. Was more thinking like `doFoo` or similar, based on what the actual effect of it was. I just wasn't sure what the effect of it was at the time, and figured ready is better than perfect at the moment.

src/hooks/useMousetrap.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  // @ts-ignore
+  bindRef,
+}) => {
+  const [boundElement, setBoundElement] = useState<HTMLElement | undefined>(
+    bindRef
+  );
+
+  // Update our boundElement if it ever changes so we trigger useEffect to re-subscribe
+  if (bindRef !== boundElement) {
```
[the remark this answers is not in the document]
0xdevalias: Yeah, I think you're right.

0xdevalias:
**Note:** while this PR was merged as part of 'The Great Merge', we should definitely come back to the comments here + clean things up a bit in future.

PR #733, opened 2020-11-19
GitHub mention bot via GH Action

thecadams:
This forks https://github.com/marketplace/actions/github-mention-to-slack in case it is ever compromised and so we can control the feature set, and makes it available in Sparkle repo as a GH action.


denisdimitrov (approved):
LGTM!

PR #744, opened 2020-11-20
fix root Firestore types to be possibly undefined

0xdevalias:
A number of the root firestore types implied that the data would always be present, even though that's not guaranteed. This has lead to a few unexpected, including https://github.com/sparkletown/sparkle/pull/743

- TODO: There are more type errors that need fixing with proper checks as they are potentially undefined
- see https://github.com/sparkletown/sparkle/pull/744 for more potential issues around how we're using `Record`

**Note:** I will probably rebase this onto latest staging/fix conflicts/etc before going much further with it. So expect a rebase/force-push to come on this branch in the near future.

### DIFF
--- a/src/components/context/ChatContext.tsx
+++ b/src/components/context/ChatContext.tsx
@@ -1,7 +1,9 @@
 import React, { useCallback } from "react";
 import { useFirestoreConnect } from "react-redux-firebase";
+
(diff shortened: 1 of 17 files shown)


mike-lvov:
This PR is related to #1042.

PR #753, opened 2020-11-21
use parentVenue title/subtitle/etc in NavBar  when parent exists + some performance bits along the way

0xdevalias:


denisdimitrov (approved):
LGTM!

PR #754, opened 2020-11-21
fix a bunch of things that should make PartyMap and ThemeCamp Map way more performant!

0xdevalias:
A bunch of performance fixes.


0xdevalias (commented):
Self-review, all looks good here!

denisdimitrov (approved):
LGTM!

PR #755, opened 2020-11-21
add more party map fixes and tweaks

denisdimitrov:
- blue avatar + pulse
- Chatting with: indicator
- add chat drawer to other templates
- increase size of video in firebarrel
- show online burners at the top


0xdevalias (approved):
LGTM!

PR #756, opened 2020-11-21
fix useCallback dependencies not being exhaustive + actually ask for the events that we want...

0xdevalias:
The `useCallback` deps weren't correct and our linters missed it.

Also I forgot to tell it to turn on the part that gives us the events... whoops


denisdimitrov (approved):
LGTM!

PR #758, opened 2020-11-21
Enable code check

thecadams:
The server is specified by an environment variable already present in CircleCI, which is so we do not expose the URL in the codebase

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -35,6 +35,7 @@ export const SPARKLEVERSITY_URL = "https://sparklever.se/sparkleversity";
 export const SPARKLEVERSE_COMMUNITY_URL =
   "https://www.facebook.com/groups/sparkleverse/";
 export const TICKET_URL = "https://sparklever.se/glowforth/";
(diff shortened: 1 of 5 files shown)


src/settings.ts
```diff
@@ -34,7 +34,8 @@ export const CREATE_EDIT_URL = "/admin";
 export const SPARKLEVERSITY_URL = "https://sparklever.se/sparkleversity";
 export const SPARKLEVERSE_COMMUNITY_URL =
   "https://www.facebook.com/groups/sparkleverse/";
-export const TICKET_URL = "https://qkt.io/seed2020";
+export const TICKET_URL = "https://sparklever.se/glowforth/";
+export const CODE_CHECK_ENABLED = true;
```
thecadams, line 38: This is quite hacky, but we needed to enable this systemwide to prevent workarounds.

thecadams: And in future this will be part of the venue's config.

denisdimitrov (approved):
LGTM!

PR #759, opened 2020-11-21
workaround bouncy partygoers without id + admin subvenue room fixes

0xdevalias:
Partygoers seem to be duplicated sometimes, but the duplicates don't have id's.


denisdimitrov (approved):
LGTM!

PR #763, opened 2020-11-21
Allow partygoers to be undefined

thecadams:
Fixes auditorium


thecadams:
@0xdevalias PTAL

0xdevalias (approved):
1 more down! :shipit:

PR #764, opened 2020-11-21
ensure NavBar component doesn't render 'loading...' when no venue 

0xdevalias:
(eg. in Admin panel)

Also seemed to be some minor spacing related churn


thecadams (approved):
lgtm!

PR #769, opened 2020-11-22
Deploy

thecadams:


src/components/templates/PartyMap/PartyMapRouter.tsx
```diff
@@ -10,7 +10,7 @@ export const PartyMapRouter: FC = () => {
 
   return (
     <Switch>
-      <Route exact path={`${match.url}/adminski`} component={PartyMapAdmin} />
+      <Route exact path={`${match.url}/admin`} component={PartyMapAdmin} />
```
0xdevalias, line 13: This naming was chosen explicitly, is the reason for renaming it now that the proper checks are back in place around it?

thecadams: Yep!

0xdevalias: Nice! :)

PR #776, opened 2020-11-23
Allow hiding room attendance

thecadams:


thecadams:
@0xdevalias PTAL - problems with attendance being incorrect mean it was requested to hide it for now.

0xdevalias (dismissed):
Looks good, LGTM! :shipit:

**[OOS] Future improvements:** Lets extract how we show attendance into a single canonical component

thecadams:
@0xdevalias based on this morning's conversation I'll rework this to show attendance if it's >0.

thecadams:
@0xdevalias PTAL - reworked it as promised.

src/components/templates/Audience/Audience.tsx
```diff
@@ -215,12 +215,6 @@ export const Audience: React.FunctionComponent = () => {
   const rowsForSizedAuditorium = MIN_ROWS + auditoriumSize * 2;
   const columnsForSizedAuditorium = MIN_COLUMNS + auditoriumSize * 2;
 
-  const translateRow = (untranslatedRowIndex: number) =>
-    untranslatedRowIndex - Math.floor(rowsForSizedAuditorium / 2);
-
-  const translateColumn = (untranslatedColumnIndex: number) =>
-    untranslatedColumnIndex - Math.floor(columnsForSizedAuditorium / 2);
-
```
thecadams, line 223: For reasons unbeknownst to me, the compiler suddenly complained about this causing re-evaluation of the below `useMemo` at every render. But hey, it was right. So I was happy to oblige and fix.

src/components/templates/Playa/MyAvatar.tsx
```diff
@@ -77,7 +77,7 @@ const MyAvatar: React.ForwardRefRenderFunction<HTMLDivElement, PropsType> = (
     stateInitialized.current = true;
   }, [serverSentState, setMyLocation, setBikeMode, setVideoState]);
 
-  const arrowMoveIntervalRef = useRef<NodeJS.Timeout | null>(null);
+  const arrowMoveIntervalRef = useRef<number | null>(null);
```
thecadams, line 80: This type error has been lurking for a while, not sure why it wasn't caught earlier but the compiler spontaneously started giving errors about it that wouldn't go away. Easy fix, have to include here due to build failures otherwise.

src/components/templates/Audience/Audience.tsx
0xdevalias: Can these be defined outside of the component? If not, they should probably use `useCallback`

0xdevalias: Ah, I see that you've moved them into the giant `useMemo` block below. That should do the trick. Though we also probably want to clean that up in future/potentially turn it into smaller chunks.

src/components/templates/Playa/MyAvatar.tsx
0xdevalias: Is the value you changed it to a valid value for how we use`arrowMoveIntervalRef`? (I assume yes, but i'd rather ask and get an 'all good' than not ask and find out it was an issue.

0xdevalias (approved):
[CRO] Approved on the caveat that my questions don't uncover any unexpected issues.

0xdevalias:
@thecadams your extra bug fixes in here are blocking #786 (and likely will block more things soon too), so probably getting this in sooner rather than later, and handling any extra cleanup required (if any) in a new branch would probably be ideal.

PR #779, opened 2020-11-23
add partygoers as default setting

denisdimitrov:
I really, really, really dislike the name headerTitle but I couldn't think of anything else, any other suggestions would be amazing.

### DIFF
--- a/src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
+++ b/src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
@@ -2,7 +2,11 @@ import React from "react";
 
 import { useSelector } from "hooks/useSelector";
 
-import { currentVenueSelector, partygoersSelector } from "utils/selectors";
+import {
+  currentVenueSelector,
+  parentVenueSelector,
+  partygoersSelector,
+} from "utils/selectors";
 
 import { Venue } from "types/Venue";
 import { User } from "types/User";
@@ -16,8 +20,16 @@ const filterVenuePartygoers = (partygoers: User[], venue: Venue) => {
 export const VenuePartygoers = () => {
   const venue = useSelector(currentVenueSelector);
   const partygoers = useSelector(partygoersSelector) ?? [];
+  const parentVenue = useSelector(parentVenueSelector);
 
+  const currentVenueTitle = venue.headerTitle ?? "partygoers";
+  const headerTitle = parentVenue?.headerTitle ?? currentVenueTitle;
   const currentVenuePartygoers = filterVenuePartygoers(partygoers, venue);
   const numberOfPartygoers = currentVenuePartygoers.length;
-  return <div>{numberOfPartygoers} burners online</div>;
+
+  return (
+    <div>
+      {numberOfPartygoers} {headerTitle} online
+    </div>
+  );
 };
--- a/src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
+++ b/src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
@@ -22,14 +22,14 @@ export const VenuePartygoers = () => {
   const partygoers = useSelector(partygoersSelector) ?? [];
   const parentVenue = useSelector(parentVenueSelector);
 
-  const currentVenueTitle = venue.headerTitle ?? "partygoers";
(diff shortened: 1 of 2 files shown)


src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
```diff
@@ -16,8 +20,16 @@ const filterVenuePartygoers = (partygoers: User[], venue: Venue) => {
 export const VenuePartygoers = () => {
```
0xdevalias, line 20: I know this is departing from our current pattern here, but i'm worried that by having this in molecules people might assume that they can just use it, yet the selector's aren't necessarily guaranteed to have been connected higher up.

I think we're too 'free' with where we connect into redux, and that adds a lot more complexity when we need to think about where a change may impact.

For this component, since it's used only in the `NavBar` at this stage, which appears to already have selectors/etc for the `venu`, `parentVenue`, etc; I think it would be nicer if we make these props on this component, and pass them down to this. Keeping this as a 'dumb' component (eg. only takes its data via props)

Since `partygoers` doesn't seem to be in `NavBar`, let's keep the selector for it in this component for now. We can think about it more in future as to what a better (if any) pattern might be for it.

```suggestion
interface VenuePartygoersProps {
  venue: ?TODO:TYPE?,
  parentVenue: ?TODO:TYPE?,
}

export const VenuePartygoers = ({ venue, parentVenue }) => {
```

src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
```diff
@@ -16,8 +20,16 @@ const filterVenuePartygoers = (partygoers: User[], venue: Venue) => {
...
   const venue = useSelector(currentVenueSelector);
   const partygoers = useSelector(partygoersSelector) ?? [];
+  const parentVenue = useSelector(parentVenueSelector);
 
+  const currentVenueTitle = venue.headerTitle ?? "partygoers";
+  const headerTitle = parentVenue?.headerTitle ?? currentVenueTitle;
   const currentVenuePartygoers = filterVenuePartygoers(partygoers, venue);
```
0xdevalias, line 27: Probably wrap this with `useMemo` so we only have to re-filter the data when current venue or partygoers changes, rather than every render.

0xdevalias (changes requested):
[CRO] Left a few comments/minor changes.

**Question:** Is there additional 'backend' work required for this feature (rg. in firestore, functions, etc)? And if so, is that being handled in another PR, or are we just leaving it undefined/to fall back to the default for now?

0xdevalias:
> I really, really, really dislike the name `headerTitle` but I couldn't think of anything else, any other suggestions would be amazing.

How about `partygoersLabel` or similar? Our default is partygoers, the descriptive text on any settings can be like "What do you want to call your partygoers?", etc. Feels like we are just providing an option to 'overload' what the partygoer label is set to for some events.

src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
thecadams: I wonder if we can inherit from `React.FunctionComponent` and enforce the inability to connect to the store from that inherited class.

0xdevalias: It's probably more of a job for eslint/similar I would think. If/when it becomes a need to enforce in any strong way. But for now I think just being mindful of it and when to use it is good.

Unless you mean to constrain things such that you can only use a certain component if it exists within a certain parent that provides a dependency, in which case, React's context's would be my immediate thought (and i have a few musing thoughts on how to use this pattern in an effective way bubbling in my brain already)

0xdevalias: [NB, Future Improvement]

src/components/molecules/VenuePartygoers/VenuePartygoers.tsx
0xdevalias: [NB, Future Improvement]

0xdevalias (approved):
[CRO] If you've run this and tested that it does the thing, then LGTM! :shipit:

PR #780, opened 2020-11-23
fix schedule info when parent is missing

denisdimitrov:
When ``doc: venueParentId`` is ```undefined``` the database falls back to the first document in the collection.


0xdevalias:
**Note:** would be good to pull this branch down and play with it/try to break it a bit during review just to ensure we've caught all the places (as this is important for upcoming events)

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -98,7 +98,7 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
     }),
     [venueParentId]
   );
-  useFirestoreConnect(venueParentQuery);
+  useFirestoreConnect(venueParentId ? venueParentQuery : undefined);
```
0xdevalias, line 101: I wonder if there is a fix we can apply to the selectors that sort of masked this issue as well? Eg. rather than pulling the first item from the DB and treating it as 'fine', are we able to ensure that the selector would also just return `undefined` if there is no `parentId`?

0xdevalias (commented):
[CRO]

src/components/molecules/NavBar/NavBar.tsx
thecadams: @denisdimitrov per @0xdevalias suggestion, we now have new stricter checks in https://github.com/sparkletown/sparkle/pull/784 - the type system will force this to be fixed.

thecadams:
Great catch here @denisdimitrov and @0xdevalias - the changes intended by this should be incorporated into https://github.com/sparkletown/sparkle/pull/784 - going to close this one, please reopen if you disagree. Great stuff!

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: @thecadams coming back to this with new 🧠 -space, I don't believe your PR addresses my original concern called out above (though is an epic improvement in it's own right!)

What I was calling out here relates to https://github.com/sparkletown/sparkle/blob/9e2dcc2d07416c89b1e0ca76c057bb67c6abc4fb/src/utils/selectors.ts#L137-L139

Specifically the `?.[0]` on the end of it, which, when `undefined`, returns the first item in the DB. This was copied into the selectors as it was the existing pattern used across the codebase when this selector was made.. but I don't think it is a good one/makes sense. By removing it, `parentVenueOrderedSelector` can properly return `undefined` rather than masking it with the wrong venue.

0xdevalias:
Reopening this as per context in https://github.com/sparkletown/sparkle/pull/780#discussion_r530250131

0xdevalias (approved):
[CRO] If the clickthrough on this works then this is good to go. LGTM! :shipit:

PR #783, opened 2020-11-24
Truthiness checking convenience functions

thecadams:


src/utils/types.ts
```diff
@@ -0,0 +1,4 @@
+export const isTruthy = <T>(item: T): boolean => !!item;
+
+export const hasElements = (array: any[] | null | undefined): boolean =>
```
0xdevalias, line 3:
```suggestion
export const hasElements = <T>(array?: T[] | null): boolean =>
```

0xdevalias (approved):
[CRO] Looks good, LGTM! :shipit: 

Future: we should add an eslint warn/ban for using `!!` syntax (`non-restricted-syntax`) and point devs towards the `isTruthy` helper.

PR #784, opened 2020-11-24
Prevent pathological queries when doc is not specified

thecadams:


src/hooks/useFirestoreConnect.ts
```diff
@@ -11,17 +12,36 @@ import { ValidFirestoreKeys } from "types/Firestore";
  */
 export type Defined<T> = T & Exclude<T, undefined>;
 
+/**
+ * This type allows us to avoid bugs when doc is not specified, and
+ * automagically constrain the storeAs parameter to keys that exist
+ * within ValidFirestoreKeys, ensuring that we can't forget to define
+ * it in our types when using functions that rely on this type
+ * (eg. useFirestoreConnect)
```
0xdevalias, line 20: This doesn't make as much sense anymore now that we're renaming this helper + shadowing `useFirestoreConnect`

src/hooks/useFirestoreConnect.ts
```diff
@@ -40,8 +60,9 @@ export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
  * @see ValidFirestoreKeys
  * @see ReduxFirestoreQuerySetting
  */
-export const useSparkleFirestoreConnect = (config: SparkleRFQConfig[]) =>
-  useFirestoreConnect(config);
+export const useFirestoreConnect = (
+  config: (SparkleRFQSetting | SparkleDocRFQSetting)[]
+) => _useFirestoreConnect(config);
```
0xdevalias: Curious why the rename to shadow the default lib? IMO it makes it more confusing when I see it in code, I then have to mentally reason about/go check the imports, etc.

src/hooks/useFirestoreConnect.ts
```diff
@@ -11,17 +12,36 @@ import { ValidFirestoreKeys } from "types/Firestore";
...
+  where: WhereOptions | WhereOptions[];
+  storeAs?: ValidFirestoreKeys;
+}
+
 /**
  * This type allows us to automagically constrain the storeAs
  * parameter to keys that exist within ValidFirestoreKeys, ensuring
  * that we can't forget to define it in our types when using functions
- * that rely on this type (eg. useSparkleFirestoreConnect)
+ * that rely on this type (eg. useFirestoreConnect)
```
0xdevalias, line 36: This doesn't apply to the type it's attached to anymore?

src/hooks/useFirestoreConnect.ts
```diff
@@ -11,17 +12,36 @@ import { ValidFirestoreKeys } from "types/Firestore";
...
+ * automagically constrain the storeAs parameter to keys that exist
+ * within ValidFirestoreKeys, ensuring that we can't forget to define
+ * it in our types when using functions that rely on this type
+ * (eg. useFirestoreConnect)
+ *
+ * @see ValidFirestoreKeys
+ * @see useFirestoreConnect
+ * @see ReduxFirestoreQuerySetting
+ */
+export interface SparkleRFQSetting extends ReduxFirestoreQuerySetting {
```
0xdevalias, line 26: Why the rename? Guessing because we already had a `Q` in `SparkleRFQ`?

We're extending `ReduxFirestoreQuerySetting`, which relates to a query. `SparkleRFQSetting` is less obvious in intent without additional mental process IMO. If we want both, something like `ReduxFirestoreQuerySetting` maybe?

Not sure it's name even really needs to mirror the type it's extending though, so maybe just `SparkleFirestoreQuery` or something?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -89,14 +89,18 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+      venue?.parentId
+        ? [
+            {
+              collection: "venues",
+              doc: venue.parentId,
+              storeAs: "parentVenue",
+            },
+          ]
+        : [],
+    [venue]
```
0xdevalias, line 103: We don't actually need to `useMemo` the queries as it does a deep compare.

It could also be cool to have a different param on our `useFirestoreConnect` overloaded function for handling these 'check boolean and only then give the config', as it's a pattern we use a lot.

Something like (names are hard):

```
useSparkleFirestoreConnectIfDefined(booleanCheck, maybeConfig)
```

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,37 +171,37 @@ export const useConnectRelatedVenues: ReactHook<
   const makeEventsQueryConfig = (
     doc: string,
     storeAs: string
-  ): SparkleRFQConfig =>
+  ): SparkleDocRFQSetting =>
```
0xdevalias, line 174: What's the intent of `SparkleDocRFQSetting` here compared to `SparkleRFQSetting`?

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,37 +171,37 @@ export const useConnectRelatedVenues: ReactHook<
...
-  const subvenueEventsQueries: SparkleRFQConfig[] = withEvents
+  const subvenueEventsQueries: SparkleDocRFQSetting[] = withEvents
     ? subvenues.map((subvenue) =>
         makeEventsQueryConfig(subvenue.id, `subvenueEvents-${subvenue.id}`)
       )
     : [];
 
   // Combine / filter for valid queries
-  const allValidQueries: SparkleRFQConfig[] = [
+  const allValidQueries: (SparkleRFQSetting | SparkleDocRFQSetting)[] = [
```
0xdevalias, line 204: Why do we need to explicitly `|` these here?

src/hooks/useFirestoreConnect.ts
```diff
@@ -40,8 +60,11 @@ export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
  * @see ValidFirestoreKeys
  * @see ReduxFirestoreQuerySetting
  */
-export const useSparkleFirestoreConnect = (config: SparkleRFQConfig[]) =>
-  useFirestoreConnect(config);
+export const useFirestoreConnect = (
+  config:
+    | (SparkleRFQSetting | SparkleDocRFQSetting)
+    | (SparkleRFQSetting | SparkleDocRFQSetting)[]
+) => _useFirestoreConnect(config);
```
0xdevalias, line 67: Why the need for the explicit `|` here? What's the intent?

(and if we are going to keep it, we should make a type that wraps this up so we don't need to duplicate this pattern across the codebase)

src/pages/Admin/Admin.tsx
```diff
@@ -427,7 +427,7 @@ const Admin: React.FC = () => {
 
   const { isAdminUser, isLoading: isAdminUserLoading } = useIsAdminUser(userId);
 
-  const venuesOwnedByUserQuery = useMemo<ReduxFirestoreQuerySetting>(
+  const venuesOwnedByUserQuery = useMemo<SparkleRFQSetting>(
     () => ({
       collection: "venues",
       where: [["owners", "array-contains", userId]],
```
0xdevalias, line 433: We don't actually need to `useMemo` our query configs as it does a deep compare on them anyway.

0xdevalias (commented):
[CRO] Questions/comments, etc. Holding off on approval/changes needed based on responses.

src/hooks/useConnectRelatedVenues.ts
thecadams: It's to ensure that if your intention is to query for a single doc, you only query for that doc.

0xdevalias: From @thecadams on zoom (paraphrased): essentially we had times when a `doc` that we were supplying was `undefined`, and so we were loading the entire collection (because it's an optional param in the underlying `useFirestoreConnect` hook), when our intent was to load a single tem or bailout.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -89,14 +89,18 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
 
-  const venueParentId = venue?.parentId;
-  const venueParentQuery = useMemo<ReduxFirestoreQuerySetting>(
-    () => ({
-      collection: "venues",
-      doc: venueParentId,
-      storeAs: "parentVenue",
-    }),
-    [venueParentId]
+  const venueParentQuery = useMemo<SparkleRFDocQuery[]>(
```
0xdevalias, line 92: We don't need to useMemo on these queries, can we clean it up.

0xdevalias: [OOS, Future Improvement] Let's clean up this pattern in a new PR/as we come across it.

src/hooks/useFirestoreConnect.ts
0xdevalias: Discussed offline, cool to go with it for now.

src/pages/Admin/Admin.tsx
```diff
@@ -427,7 +424,7 @@ const Admin: React.FC = () => {
 
   const { isAdminUser, isLoading: isAdminUserLoading } = useIsAdminUser(userId);
 
-  const venuesOwnedByUserQuery = useMemo<ReduxFirestoreQuerySetting>(
+  const venuesOwnedByUserQuery = useMemo<SparkleRFQuery>(
```
0xdevalias, line 427: [NB, Future Improvement]
Does eslint let us warn/error on things like this? `useMemo<SparkleRFQuery>` (may just make more sense to fix all cases of it instead. But would be interesting to know.

0xdevalias (approved):
Some [OOS, NB, Future Improvement] comments, but aside from that, looks good! This will help enforce us not making silly mistakes. A+! LGTM!

PR #786, opened 2020-11-24
Add custom 'slim' template to circleci slack/notify on fail step

0xdevalias:
This replaces the `basic_fail_1` template ([ref](https://github.com/CircleCI-Public/slack-orb/blob/master/src/message_templates/basic_fail_1.json)) with a custom template that makes the messages slimmer and less spammy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ notify-on-fail: &notify-on-fail
     condition: always
     steps:
       - slack/notify:
-          event: fail
+          event: always
           custom: |
             {
               "blocks": [
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,48 @@ notify-on-fail: &notify-on-fail
     steps:
       - slack/notify:
           event: fail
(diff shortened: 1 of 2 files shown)


thecadams (approved):
lgtm!

.circleci/config.yml
```diff
@@ -3,8 +3,49 @@ notify-on-fail: &notify-on-fail
     condition: always
     steps:
       - slack/notify:
-          event: fail
-          template: basic_fail_1
+          event: always
```
0xdevalias, line 6: This was added to test the notification, we need to remove it once we know it works, and before this PR is merged

0xdevalias: restored in https://github.com/sparkletown/sparkle/pull/786/commits/02cd75177a998f4182269cca99a79d7f55ce85fe

0xdevalias:
The current build failures on this branch are unrelated to changes here.

@thecadams already has a fix for them on https://github.com/sparkletown/sparkle/pull/776

Once that is merged, we can merge into this and then we should be good to go.

PR #789, opened 2020-11-24
Cleanup/update CircleCI + use npm ci instead of npm install

0xdevalias:
`npm ci` is designed for use in CI, is faster, and ensures `package.lock` inconsistencies/similar result in an error:

- https://docs.npmjs.com/cli/v6/commands/npm-ci

We weren't actually running our `eslint` rules in CI due to our confusingly named `npm` scripts (`npm run lint` executed `prettier --write`) Made it so we do, and we also run `prettier` in check mode now.


.circleci/config.yml
```diff
@@ -49,22 +49,20 @@ notify-on-fail: &notify-on-fail
...
-            - run: npm run init
-            - run: npm run lint
-            - run: npm run build # somehow linting rules differ from eslint
-            - *notify-on-fail
+      - node/install-packages
+      - run: npm run init
+      - run: npm run prettier:check
+      - run: npm run eslint:src
+      - run: npm run test
+#      - run: npm run build # is there any point in building here?
```
0xdevalias, line 64: We can probably just delete this line. We build later before deploying hosting anyway.

0xdevalias (commented):
This still has a number of newly uncovered eslint errors that need fixing. My local environment is failing at the moment, but if someone could try running `eslint` `--fix` across the linted files and/or pushing fixed for them to this branch, that would be awesome!

.circleci/config.yml
thecadams: Yep seems like we can given it was last so nothing depends on it

package.json
```diff
@@ -71,8 +71,10 @@
     "start": "react-scripts start",
     "init": "node scripts/init.js",
     "build": "react-scripts build",
-    "lint": "prettier --write \"{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
-    "lint:src": "eslint src --ext .js,.jsx,.ts,.tsx",
+    "lint": "npm run prettier:check && npm run eslint:src",
+    "eslint:src": "eslint src --ext .js,.jsx,.ts,.tsx",
```
thecadams, line 75: Let's also add `eslint:fix` here for discoverability

0xdevalias: done

package.json
```diff
@@ -71,8 +71,11 @@
...
+    "eslint:fix": "eslint src --fix --ext .js,.jsx,.ts,.tsx",
```
0xdevalias, line 76: [OOS, Future Cleanup] Can potentially rewrite this as:
```suggestion
    "eslint:fix": "npm run eslint:src --fix,
```

src/components/templates/Playa/PlayaIcon.tsx
```diff
@@ -34,8 +34,8 @@ export const PlayaIconComponent: React.FunctionComponent<PropsType> = ({
     >
       {playaIcon.fire && (
         <div className="fire">
-          {Array.from(Array(50)).map(() => (
-            <div className="particle" />
+          {Array.from(Array(50)).map((_, index) => (
+            <div className="particle" key={index} />
```
0xdevalias, line 38: [NB, OOS, Future Cleanup] We should probably `useMemo` this

src/pages/Admin/AdminEditComponent.tsx
```diff
@@ -515,12 +515,12 @@ const PlacementForm: React.FC<PlacementFormProps> = (props) => {
             {formError && (
               <div className="input-error">
                 <div>One or more errors occurred when saving the form:</div>
-                {Object.keys(errors).map((fieldName) => (
-                  <div>
+                {Object.keys(errors).map((fieldName, index) => (
+                  <div key={index}>
```
0xdevalias, line 519: [NB, nit] I think this would be fixed better by using `fieldName`? (assuming we only report a single error per field?)

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -688,18 +690,18 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
...
+          You&apos;ll be able to add rooms to your theme camp on the next page
         </div>
       )}
       {formError && (
         <div className="input-error">
           <div>One or more errors occurred when saving the form:</div>
-          {Object.keys(errors).map((fieldName) => (
-            <div>
+          {Object.keys(errors).map((fieldName, index) => (
+            <div key={index}>
```
0xdevalias, line 700: [NB, nit] I think this would be fixed better by using `fieldName`? (assuming we only report a single error per field?)

0xdevalias (commented):
[CRO] looks good to me, but I can't approve my own PR's ;p

0xdevalias (commented):
Changes look good, needs your approval (as I can't approve my own), then LGTM!

thecadams (approved):

PR #791, opened 2020-11-24
fix camp names and join button width

denisdimitrov:
The grid-container was colliding with other styles on global level, now with the concrete name it won't.

### DIFF
--- a/src/components/templates/Camp/components/Map.scss
+++ b/src/components/templates/Camp/components/Map.scss
@@ -122,7 +122,8 @@ $border-radius: 28px;
     left: 0;
     right: 0;
     margin: auto;
-    transition: all 400ms cubic-bezier(0.23, 1, 0.32, 1);
+    transition: width 400ms cubic-bezier(0.23, 1, 0.32, 1) height 400ms
+      cubic-bezier(0.23, 1, 0.32, 1);
     opacity: 0;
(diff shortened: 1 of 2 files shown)


src/components/templates/Camp/components/Map.scss
```diff
@@ -129,6 +131,145 @@ $border-radius: 28px;
...
+    &:active {
+      .camp-venue-img {
+        transform: scale(1);
+      }
+    }
+    .camp-venue-img {
+      position: relative;
+      width: 100%;
+      height: 100%;
+      transition: all 400ms cubic-bezier(0.23, 1, 0.32, 1);
```
IvanBuljovcic, line 167: This is a nitpick:

When adding transition, try not to add the _all_, that basically means that the browser will watch all transitionable properties and apply the timing to them.

Rather, put the specific property that you want to change i.e.

```
transition:
    width 400ms cubic-bezier(0.23, 1, 0.32, 1),
    height 400ms cubic-bezier(0.23, 1, 0.32, 1)
```

IvanBuljovcic (approved):

src/components/templates/Camp/components/Map.scss
0xdevalias: TIL!

PR #792, opened 2020-11-24
Delete/re-create package-lock.json to fix resolution issues + bump some eslint dependencies

0xdevalias:
There seemed to be a bunch of `package-lock.json` related issues that were blocking a few people. This PR deletes the old one, and recreates it.

**Note:** This may cause a bunch of our underlying dependencies to shift versions, so this may pose a risk to production/upcoming events if it's not tested thoroughly first. (thus the `do-not-merge` label) But it would be good to get some 🧠  / 👀  on it.

While my understanding is that they're eventually merging anyway, in the past, projects I have worked on have used `yarn` over `npm` due to issues like this.


0xdevalias:
@thecadams PTAL. I'm not sure if this is necessarily a good fix/solution here, and given it's likely to bump a whole pile of underlying dependencies, if they don't correctly follow semver there may be unexpected bugs. Feels risky to merge before the next event, but maybe after that?

It was blocking me and some other team members due to our local environment being broken/unrunnable. One workaround for now is to just `rm ./package-lock.json` locally, then `npm install`, let it rebuild the `package-lock.json` and then just use this locally, but **NOT** commit the `package-lock.json`. An alternative workaround seems to be using `yarn`, but unsure if that is because it handles the resolutions/etc better, or if its just because it rebuilds it's own lockfile (using the command to import from a `package-lock.json` may be able to test this theory; see https://classic.yarnpkg.com/en/docs/cli/import/)

0xdevalias:
Further reading: 

- https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/
  - > `package-lock` `v2` and support for `yarn.lock`: Our new `package-lock` format will unlock the ability to do deterministically reproducible builds and includes everything `npm` will need to fully build the package tree. Prior to `npm` `7` `yarn.lock` files were ignored, the `npm` cli can now use `yarn.lock` as source of package metadata and resolution guidance.
- https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson
  - > One common question we’ve gotten a few times now, once we announce that npm v7 will include support for yarn.lock files, is “Why keep `package-lock.json` at all, then? Why not just use `yarn.lock` only?”
  - > The simple answer is: because `yarn.lock` doesn’t fully address npm’s needs, and relying on it exclusively would limit our ability to produce optimal package installs or add features in the future.
  - > Yarn installs are guaranteed to be deterministic given a single combination of yarn.lock and Yarn version. It is possible that a different version of Yarn will result in a different tree layout on disk.
  - > A `yarn.lock` file does guarantee deterministic resolutions of dependencies. For example, if `foo@1.x` resolves to foo@1.2.3, it’ll continue to resolve to that version number in subsequent installs for all Yarn versions, given a consistent yarn.lock file. But that (at least, in itself) is not equivalent to guaranteeing a deterministic tree shape!
  - > In other words, the Yarn tree building contract is split between the `yarn.lock` file and the implementation of `Yarn` itself. The `npm` tree building contract is entirely specified by the `package-lock.json` file. This makes it much harder for us to break by accident across `npm` versions, and if we do (whether by mistake or on purpose), the change will be reflected in the file in source control.
  - > Furthermore, there is a class of nesting and deduplication cases where the `yarn.lock` file does not accurately reflect the resolutions that will be used by `npm` in practice, even when `npm` does use it as a source of metadata. While `npm` uses the yarn.lock file as a reliable source of information, it does not treat it as an authoritative set of constraints.

Note: I will edit this with a few more things when I get back, need to finish reading but have an appointment.

thecadams (approved):

0xdevalias:
@IvanBuljovcic noted that https://github.com/sparkletown/sparkle/commit/4befcdea052b4d85512e54b57fdb9f0fd2091b06#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519 appears to work.

Looking at the history in https://github.com/sparkletown/sparkle/commits/staging/package-lock.json it looks like https://github.com/sparkletown/sparkle/commit/4b52e0a7c343c3fa4967d69d58a966aebe5f00a7#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519 may be the culprit, with `24,669 additions, 59 deletions` on `package-lock.json`, though there is also https://github.com/sparkletown/sparkle/commit/8fd2b1386c14703741cbb11cb8fa0170eabfb15f#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519 after that, with `24,588 additions, 93 deletions`

Based on some comments @thecadams made earlier today, it could be that there was a conflict between the versions of `npm` we are using, and since without a `yarn.lock` file in v7+ it doesn't _actually_ lock down the dependency tree, this seems a likely possiblity.

Based on this, I will open a PR to restore the last known working `package-lock.json` and ensure it is up to date with packages added after that. This should unblock people in the short term.

The longer term fix likely relates to my research in https://github.com/sparkletown/sparkle/pull/792#issuecomment-733278102, and probably committing a `yarn.lock` file as well. I think we would also want to enforce that everyone is using `npm` v7+ at that stage, to avoid any issues (I have a WIP eslint rule in a local branch that may let us enforce this)

0xdevalias:
Closing this issue for now as https://github.com/sparkletown/sparkle/pull/802 will be a less risky fix for our current issues as per https://github.com/sparkletown/sparkle/pull/792#issuecomment-733614988

0xdevalias:
FYI, backlogged https://github.com/sparkletown/internal-sparkle-issues/issues/96 for my desire to continue the research mentioned in https://github.com/sparkletown/sparkle/pull/792#issuecomment-733623764 + devise a strategy for going forward

PR #793, opened 2020-11-24
Allow JazzBar to load table configuration from venue.config.tables

0xdevalias:
Note: `do-not-merge` is currently set as I want to double check it all against the running config before merging, but the code can still be reviewed in the meantime.


src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -57,6 +66,9 @@ const Jazz: React.FunctionComponent<PropsType> = ({ setUserList, venue }) => {
 
   const venueToUse = venue ? venue : firestoreVenue;
 
+  const jazzbarConfig = venueToUse?.config?.jazzbarConfig;
+  const jazzbarTables = jazzbarConfig?.tables ?? JAZZBAR_TABLES;
```
thecadams, line 70: This should be generic across templates. So we could do `venue.config.tables` for example. There are other places that would be great to incorporate into this PR so we have a single way to specify table config and the config will apply regardless of any template changes.

0xdevalias: My thoughts with this are that it would be nice to be able to have more tightly constricted types where we know what is expected to be possible, and what isn't. That lead me to this implementation originally.

But thinking about it really quickly now we should be able to do this even if it is more generically named (eg. using type descriminators based on the string value of template or similar, OOS of this PR), so I support @thecadams 's view.

src/types/Venue.ts
```diff
@@ -22,25 +24,7 @@ export interface Venue {
...
-      bannerImageUrl?: string;
-      checkList: string[];
-      iframeUrl?: string;
-      joinButtonText?: string;
-      quotations?: Quotation[];
-    };
-    memberEmails?: string[];
-    showRangers?: boolean;
-  };
+  config?: VenueConfig;
```
0xdevalias, line 27: We have keys in the root of `Venue`, and the `config` key, do we want to keep this in config?

0xdevalias: @thecadams said it makes sense to use `config` here, we can do a v2 of `Venues` type at some stage and cleanup some of the other keys if it makes sense to. But canonically, use `config`

src/types/Venue.ts
```diff
@@ -87,6 +71,35 @@ export interface Venue {
   showSecretPasswordForm?: boolean;
 }
 
+export interface VenueConfig {
+  theme: {
+    primaryColor: string;
+    backgroundColor?: string;
+  };
+
+  jazzbarConfig: VenueJazzbarConfig;
```
0xdevalias, line 80:
```suggestion
  tables?: Table[];
```

src/types/Venue.ts
```diff
@@ -87,6 +71,35 @@ export interface Venue {
...
+  landingPageConfig: VenueLandingPageConfig; // @debt should this be potentially undefined, or is it guaranteed to exist everywhere?
+
+  memberEmails?: string[];
+  showRangers?: boolean;
+}
+
+export interface VenueJazzbarConfig {
+  tables?: Table[];
+}
```
0xdevalias, line 89:
```suggestion
```

0xdevalias (commented):
Documented changes required as discussed (aside from updating things that use them if any exist currently)

@thecadams If you're happy to implement these while i'm out and test it with some config in the DB, then this should be good to go. Happy to review when I return, or if someone else reviews in the meantime, that is cool too.

src/types/Venue.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/793/commits/d38a1bcc34f80038394d3705cf3e917120abc1f4

src/types/Venue.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/793/commits/d38a1bcc34f80038394d3705cf3e917120abc1f4

src/types/Venue.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/793/commits/d38a1bcc34f80038394d3705cf3e917120abc1f4

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/793/commits/df6c5ff6e1ce46d62bfb2cb0ec97cec7f0ea0e65

src/components/molecules/TableHeader/TableHeader.tsx
```diff
@@ -35,8 +35,7 @@ const TableHeader: React.FC<TableHeaderProps> = ({
       (user: User) => user.data?.[venueName]?.table === seatedAtTable
     );
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const firestoreUpdate = (doc: string, update: any) => {
+  const firestoreUpdate = (doc: string, update: firestore.UpdateData) => {
```
0xdevalias, line 38: This change is causing a TypeError currently

0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/793/commits/ab1f717fcfffe2690e1eb95d241d509e6d119236

0xdevalias:
Note: tested this in staging by editing the firestore DB, saw the tables dissapear, edited to remove it, saw the tables come back

denisdimitrov (approved):
LGTM!

PR #796, opened 2020-11-25
Align staging with master

0xdevalias:
Note: this includes an emergency fix that was pushed directly to the `master` branch:

- Revert "Prevent pathological queries when doc is not specified (#784)"

This caused an issue somewhere around the admin role check, which made the admin panel not load.


thecadams:
Hm, seems the checks and access restrictions we have in place are blocking this merging. It seems to have tried to deploy to prod when I merged staging into it.

I'll just push the merge direct to staging branch.

PR #797, opened 2020-11-25
Revert "Prevent pathological queries when doc is not specified (#784)"

thecadams:
This reverts commit 046fd4775ad49eacdd8e9f59ed483a064531297c.

This is an emergency fix to restore admin functionality.


thecadams:
@0xdevalias PTAL

0xdevalias (approved):
Approved (retroactively).

Musing: I think for 'emergency fixes' we need a workflow where we can easily get them through to prod, even if there isn't necessarily someone around who can approve it.

My desire is that everything at least has a PR for visibility/documenting it/etc, but I don't necessarily think that needs to be reviewed/approved.

Have some thoughts i'll let brew on this and we can figure a good pattern/solution moving forward.

0xdevalias:
image

This highlights another issue with the way our current workflows are configured, at least for this use case. It's wanting us to make a squash merge for this.. but I think for a revert that may not be something we actually want to do.. not sure, what do you think?

thecadams:
> Musing: I think for 'emergency fixes' we need a workflow where we can easily get them through to prod, even if there isn't necessarily someone around who can approve it.
> 
> My desire is that everything at least has a PR for visibility/documenting it/etc, but I don't necessarily think that needs to be reviewed/approved.

Yep we need ability to break glass. Asked here: https://stackoverflow.com/questions/64998944/ability-to-break-glass-and-bypass-github-branch-protection-for-emergency-fixes

0xdevalias:
That StackOverflow is great! Opened a draft PR at https://github.com/sparkletown/.github/pull/1 where we can make this happen! 🎉

PR #798, opened 2020-11-25
Fix deploy-hosting circleci job

thecadams:


thecadams:
Went on a journey and noticed that per https://circleci.com/developer/orbs/orb/circleci/node caching is used by default.

So we only needed to add `node/install-packages` in here

0xdevalias (approved):
Haha, yup.. that'll definitely help!

PR #801, opened 2020-11-25
Removes pathological queries; Creates linting rule to prevent raw useFirestoreConnect import

0xdevalias:
**Background Context:** The original PR (https://github.com/sparkletown/sparkle/pull/784) was deployed in https://github.com/sparkletown/sparkle/pull/795, which broke things in production, and was reverted in https://github.com/sparkletown/sparkle/pull/797

**Thoughts:** @denisdimitrov originally had https://github.com/sparkletown/sparkle/pull/780 up to solve the core of our `priority-prod-fix` concerns for an upcoming event. If that solves the immediate need, then maybe we should resurrect that and get it out sooner, which will triage that need, and then we can work on these fixes over a less risky timeframe.

This PR will fix the issues that were breaking the admin page/anything else it broke in production.


0xdevalias:
@thecadams Created this branch/draft PR to get visibility back on the board for this issue + link up it's context, but absolutely happy for you to run with it and make the fixes/etc from your original PR.

thecadams:
@0xdevalias picking this up now

thecadams:
[Performance TroubleShooting](https://trello.com/c/25Oynxx2/559-performance-troubleshooting)

src/hooks/useFirestoreConnect.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @see AnySparkleRFQuery
+ * @see ValidFirestoreKeys
+ * @see ReduxFirestoreQuerySetting
+ */
+export const useFirestoreConnect = (
+  config: AnySparkleRFQuery | AnySparkleRFQuery[],
+  shouldExecuteQuery?: boolean
+) => {
+  if (!shouldExecuteQuery) return;
```
mike-lvov, line 68: As stated in the official React doc, you must have hooks on the top level.

https://reactjs.org/docs/hooks-rules.html

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -95,14 +95,18 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
 
-  const venueParentId = venue?.parentId;
-  const venueParentQuery = useMemo<ReduxFirestoreQuerySetting>(
-    () => ({
-      collection: "venues",
-      doc: venueParentId,
-      storeAs: "parentVenue",
-    }),
-    [venueParentId]
+  const venueParentQuery = useMemo<SparkleRFDocQuery[]>(
```
mike-lvov, line 98: Suggestion: Maybe we could move this query, useConnect and selector into the hook?

src/hooks/useFirestoreConnect.ts
0xdevalias: @mike-lvov What about this isn't at the top level? This looks fine to me, as per:

- https://reactjs.org/docs/hooks-custom.html

Do you agree? If so, we can mark this resolved.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -91,20 +91,25 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
   const { user, profile } = useUser();
   const venueId = useVenueId();
   const venue = useSelector(currentVenueSelectorData);
+  const venueParentId = venue?.parentId;
```
mike-lvov, line 94: I don't know why there was no `venueParentId` var, even though it is used in the useConnect

src/pages/SparkleSpaceMarketingPage/WelcomePage/WelcomePage.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { useState } from "react";
 import firebase from "firebase/app";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
 import { IFRAME_ALLOW, PLAYA_VENUE_NAME } from "settings";
 
 const WelcomePage: React.FunctionComponent = () => {
-  useFirestoreConnect("marketingemails");
+  useSparkleFirestoreConnect("marketingemails");
```
mike-lvov, line 7: This looks like a legacy. I didn't try to refactor it since it is not related to this PR. But probably there should be created a ticket to refactor this

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -189,14 +188,20 @@ export const useConnectRelatedVenues: ReactHook<
   // Sibling Events
   const siblingVenueEventsQueries: SparkleRFQConfig[] = withEvents
     ? siblingVenues.map((sibling) =>
-        makeEventsQueryConfig(sibling.id, `siblingVenueEvents-${sibling.id}`)
+        makeEventsQueryConfig(
+          sibling.id,
+          `siblingVenueEvents-${sibling.id}` as ValidFirestoreKeys // @debt a little hacky, but we're consciously subverting our helper protections;
```
mike-lvov, line 193: I have chosen from a lesser of two evils.

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,15 +170,14 @@ export const useConnectRelatedVenues: ReactHook<
...
     doc: string,
-    storeAs: string
-  ): SparkleRFQConfig =>
-    ({
-      collection: "venues",
-      doc,
-      subcollections: [{ collection: "events" }],
-      orderBy: ["start_utc_seconds", "asc"],
-      storeAs,
-    } as SparkleRFQConfig); // @debt a little hacky, but we're consciously subverting our helper protections;
```
mike-lvov, line 181: https://github.com/sparkletown/sparkle/pull/801/files#r555860210

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -189,14 +188,20 @@ export const useConnectRelatedVenues: ReactHook<
...
       )
     : [];
 
   // Sub Events
   const subvenueEventsQueries: SparkleRFQConfig[] = withEvents
     ? subvenues.map((subvenue) =>
-        makeEventsQueryConfig(subvenue.id, `subvenueEvents-${subvenue.id}`)
+        makeEventsQueryConfig(
+          subvenue.id,
+          `subvenueEvents-${subvenue.id}` as ValidFirestoreKeys // @debt a little hacky, but we're consciously subverting our helper protections;
```
mike-lvov, line 203: https://github.com/sparkletown/sparkle/pull/801/files#r555860210

src/types/Firestore.ts
```diff
@@ -85,6 +85,7 @@ export interface FirestoreOrdered {
   statsOpenVenues?: Array<WithId<AnyVenue>>;
   subvenues?: WithId<AnyVenue>[];
   subvenueEvents?: WithId<VenueEvent>[];
+  marketingemails?: Array<string>;
```
mike-lvov, line 88: Not sure about the type of this entity, but I didn't dig into that.

#https://github.com/sparkletown/sparkle/pull/801/files#r555857316

mike-lvov:
Partially closes https://github.com/sparkletown/internal-sparkle-issues/issues/2
Not sure if I should implement other recommendations from this ticket,
since we basically gain nothing in terms of linting/etc.

@0xdevalias would be glad to hear your approach on this

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: There was seemingly one on line 98: https://github.com/sparkletown/sparkle/pull/801/files#diff-d2dbbd9743869736d76cf73512c615856022b7a3ee80ec9b63bed997c40f7e37L98

0xdevalias: Which was presumably removed because line 105 changed to use `venue.parentId` not `venueParentId`: https://github.com/sparkletown/sparkle/pull/801/files#diff-d2dbbd9743869736d76cf73512c615856022b7a3ee80ec9b63bed997c40f7e37R105

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -31,6 +27,10 @@ import { useRadio } from "hooks/useRadio";
 import { useSelector } from "hooks/useSelector";
 import { useUser } from "hooks/useUser";
 import { useVenueId } from "hooks/useVenueId";
+import {
+  SparkleRFQConfig,
+  useSparkleFirestoreConnect,
+} from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 33:
```suggestion
import { useSelector } from "hooks/useSelector";
import {
  SparkleRFQConfig,
  useSparkleFirestoreConnect,
} from "hooks/useSparkleFirestoreConnect";
import { useUser } from "hooks/useUser";
import { useVenueId } from "hooks/useVenueId";
```

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -91,20 +91,25 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+      venue?.parentId
+        ? [
+            {
+              collection: "venues",
+              doc: venue.parentId,
+              storeAs: "parentVenue",
+            },
+          ]
+        : [],
+    [venue]
```
0xdevalias, line 110: We don't need to `useMemo` on these queries. This was an earlier assumption we made that turned out to be wrong.

Since the reason it was split out of the `useSparkleFirestoreConnect` (memoisation) doesn't even matter anymore, so if it's not used elsewhere, we can just combine it together again (see suggest change in other comment below)

```suggestion
```

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -91,20 +91,25 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
   );
-  useFirestoreConnect(venueParentId ? venueParentQuery : undefined);
+  useSparkleFirestoreConnect(venueParentId ? venueParentQuery : undefined);
```
0xdevalias, line 112: As above, RE: recombining the query here

```suggestion
  useSparkleFirestoreConnect(venueParentId ? {
      collection: "venues",
      doc: venueParentId,
      storeAs: "parentVenue",
    } : undefined);
```

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -57,7 +57,7 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
   const debouncedSearch = debounce((v) => setSearchValue(v), 500);
   const searchRef = useRef<HTMLInputElement>(null);
 
-  useFirestoreConnect(
+  useSparkleFirestoreConnect(
     user && user.uid
```
0xdevalias, line 61:
```suggestion
    user?.uid
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -141,7 +141,7 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
 };
 
 const SuspectedLocation: React.FC<{ user: WithId<User> }> = ({ user }) => {
-  useFirestoreConnect("venues");
+  useSparkleFirestoreConnect("venues");
```
0xdevalias, line 144: [OOS] Not relevant to this PR, but calling this out as a pattern we want to remove, as we shouldn't ever be loading all of the venues like this.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -19,7 +19,7 @@ import { venueInsideUrl, venuePreviewUrl } from "utils/url";
 import { isCampVenue } from "types/CampVenue";
 import { Link } from "react-router-dom";
 import { ENABLE_SUSPECTED_LOCATION, RANDOM_AVATARS } from "settings";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 22: This should be sorted with other `hooks/` imports, not up here.

src/components/templates/Camp/components/RoomModal.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useMemo } from "react";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 2: This should be sorted with other `hooks/` imports, not up here.

src/components/templates/Camp/components/RoomModal.tsx
```diff
@@ -40,7 +40,7 @@ export const RoomModal: React.FC<RoomModalProps> = ({
   room,
   joinButtonText,
 }) => {
-  useFirestoreConnect("venues");
+  useSparkleFirestoreConnect("venues");
```
0xdevalias, line 43: [OOS] Not relevant to this PR, but calling this out as a pattern we want to remove, as we shouldn't ever be loading all of the venues like this.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from "react";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 2: This should be sorted with other `hooks/` imports, not up here.

src/components/templates/Playa/Playa.tsx
```diff
@@ -131,7 +131,7 @@ const isPlaced = (venue: Venue) => {
 const minZoom = () => (window.innerWidth - 2 * PLAYA_MARGIN_X) / PLAYA_WIDTH;
 
 const Playa = () => {
-  useFirestoreConnect("venues");
+  useSparkleFirestoreConnect("venues");
```
0xdevalias, line 134: [OOS] Not relevant to this PR, but calling this out as a pattern we want to remove, as we shouldn't ever be loading all of the venues like this.

src/components/templates/Playa/VenuePreview.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from "settings";
 import UserList from "components/molecules/UserList";
 import { usePartygoers } from "hooks/users";
+import { useSparkleFirestoreConnect } from "useSparkleFirestoreConnect";
```
0xdevalias, line 13: Is this import path correct? Shouldn't it be 

```suggestion
import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```

src/hooks/roles.ts
```diff
@@ -1,18 +1,17 @@
...
-  isLoaded,
-  ReduxFirestoreQuerySetting,
-  useFirestoreConnect,
-} from "react-redux-firebase";
+import { isEmpty, isLoaded } from "react-redux-firebase";
 import { SparkleSelector } from "types/SparkleSelector";
+import {
+  SparkleRFQConfig,
+  useSparkleFirestoreConnect,
+} from "./useSparkleFirestoreConnect";
```
0xdevalias, line 6: I believe `useSparkleFirestoreConnect` already re-exports `isEmpty`/`isLoaded`?

src/hooks/useConnectCurrentEvent.ts
```diff
@@ -1,5 +1,5 @@
 import { useParams } from "react-router-dom";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 2: This should be sorted with other `hooks/` imports, not up here.

src/hooks/useConnectCurrentEvent.ts
```diff
@@ -9,7 +9,7 @@ export const useConnectCurrentEvent = () => {
   const { user } = useUser();
   const [currentTimestamp] = useState(Date.now() / 1000);
 
-  useFirestoreConnect([
+  useSparkleFirestoreConnect([
     {
       collection: "venues",
       doc: venueId,
```
0xdevalias, line 15: I am pretty sure this could potentially be `undefined` based on reading it from `useParams`, which would cause a pathologic query that loads all venues.

src/hooks/useConnectCurrentVenue.ts
```diff
@@ -1,5 +1,5 @@
 import { useParams } from "react-router-dom";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 2: This should be sorted with other `hooks/` imports, not up here.

src/hooks/useConnectCurrentVenue.ts
```diff
@@ -16,15 +16,15 @@ export const useConnectCurrentVenue = () => {
     venueId = getQueryParameters(window.location.search)?.venueId;
   }
 
-  useFirestoreConnect([
+  useSparkleFirestoreConnect([
     {
       collection: "venues",
       doc: venueId,
```
0xdevalias, line 22: I am pretty sure this could potentially be `undefined` based on reading it from `useParams` / `getQueryParameters `, which would cause a pathologic query that loads all venues.

src/hooks/useConnectCurrentVenue.ts
```diff
@@ -16,15 +16,15 @@ export const useConnectCurrentVenue = () => {
...
       storeAs: "currentVenue",
     },
   ]);
 
-  useFirestoreConnect([
+  useSparkleFirestoreConnect([
     {
       collection: "venues",
       doc: venueId,
```
0xdevalias, line 30: I am pretty sure this could potentially be `undefined` based on reading it from `useParams` / `getQueryParameters `, which would cause a pathologic query that loads all venues.

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,15 +170,14 @@ export const useConnectRelatedVenues: ReactHook<
...
-    storeAs: string
-  ): SparkleRFQConfig =>
-    ({
-      collection: "venues",
-      doc,
-      subcollections: [{ collection: "events" }],
-      orderBy: ["start_utc_seconds", "asc"],
-      storeAs,
-    } as SparkleRFQConfig); // @debt a little hacky, but we're consciously subverting our helper protections;
+    storeAs: ValidFirestoreKeys
```
0xdevalias, line 173: What was the thinking/motivation in changing this from `string` to `ValidFirestoreKeys`?

src/hooks/useConnectUserPurchaseHistory.ts
```diff
@@ -1,12 +1,12 @@
 import { useParams } from "react-router-dom";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 2: This should be sorted with other `hooks/` imports, not up here.

src/hooks/useRoles.ts
```diff
@@ -1,15 +1,15 @@
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 1:
```suggestion
import { useSparkleFirestoreConnect } from "./useSparkleFirestoreConnect";
```

src/hooks/useSparkleFirestoreConnect.ts
```diff
@@ -1,6 +1,7 @@
+// eslint-disable-next-line no-restricted-imports
 import {
   ReduxFirestoreQuerySetting,
-  useFirestoreConnect,
+  useFirestoreConnect as _useFirestoreConnect,
```
0xdevalias, line 4: What is the thought process behind renaming this import here?

Assuming it was to match `_isLoaded` and `_isEmpty`, the intention of those being renamed was so that we could export our own versions of them within this same file's scope (though it looks like we didn't end up re-exporting `_isEmpty`.. which we should do)

src/hooks/useSparkleFirestoreConnect.ts
```diff
@@ -21,10 +22,17 @@ export type Defined<T> = T & Exclude<T, undefined>;
...
-export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
+
+export interface SparkleRFQConfigObject extends ReduxFirestoreQuerySetting {
   storeAs?: ValidFirestoreKeys;
 }
 
+export type SparkleRFQConfig =
+  | SparkleRFQConfigObject
+  | SparkleRFQConfigObject[]
+  | ValidFirestoreKeys // @debt Eridicate this as a way to connectData
```
0xdevalias, line 33: Can you explain this `@debt` more please.

src/hooks/useSparkleFirestoreConnect.ts
```diff
@@ -21,10 +22,17 @@ export type Defined<T> = T & Exclude<T, undefined>;
...
+  | undefined;
```
0xdevalias, line 34: I wouldn't include this here, it needlessly hides away important information, and there are better choices for how to handle an explicit undefined where we want it.

```suggestion
```

src/hooks/useSparkleFirestoreConnect.ts
```diff
@@ -40,9 +48,8 @@ export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
  * @see ValidFirestoreKeys
  * @see ReduxFirestoreQuerySetting
  */
-export const useSparkleFirestoreConnect = (
-  config?: SparkleRFQConfig[] | SparkleRFQConfig
-) => useFirestoreConnect(config);
+export const useSparkleFirestoreConnect = (config: SparkleRFQConfig) =>
```
0xdevalias, line 51:
```suggestion
export const useSparkleFirestoreConnect = (config?: SparkleRFQConfig) =>
```

src/pages/Admin/Admin.tsx
```diff
@@ -67,6 +63,10 @@ import VenueDeleteModal from "./Venue/VenueDeleteModal";
 import { VenueOwnersModal } from "./VenueOwnersModal";
 
 import "./Admin.scss";
+import {
+  SparkleRFQConfig,
+  useSparkleFirestoreConnect,
+} from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 69: This should be sorted with other `hooks/` imports, not up here.

src/pages/Admin/Admin.tsx
```diff
@@ -427,14 +427,14 @@ const Admin: React.FC = () => {
...
-  const venuesOwnedByUserQuery = useMemo<ReduxFirestoreQuerySetting>(
+  const venuesOwnedByUserQuery = useMemo<SparkleRFQConfig>(
     () => ({
       collection: "venues",
       where: [["owners", "array-contains", userId]],
     }),
     [userId]
   );
-  useFirestoreConnect(venuesOwnedByUserQuery);
+  useSparkleFirestoreConnect(venuesOwnedByUserQuery);
```
0xdevalias, line 437: This doesn't need to be `useMemo`'d, it was a legacy assumption that proved wrong.

I would also recombine the query directly into `useSparkleFirestoreConnect` unless we need to reuse it elsewhere.

src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -9,7 +9,7 @@ import { RoomInput, upsertRoom } from "api/admin";
 import { useUser } from "hooks/useUser";
 import { useSelector } from "hooks/useSelector";
 import VenueEventDetails from "./VenueEventDetails";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useSparkleFirestoreConnect } from "hooks/useSparkleFirestoreConnect";
```
0xdevalias, line 12: This should be sorted with other `hooks/` imports, not up here.

src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -28,7 +28,7 @@ export const AdminVenueRoomDetails = ({
   setShowCreateEventModal,
   setShowDeleteEventModal,
 }: Props) => {
-  useFirestoreConnect([
+  useSparkleFirestoreConnect([
     {
       collection: "venues",
       doc: venue.id,
```
0xdevalias, line 34: Can this ever be `undefined` here?

src/pages/Admin/EventsComponent.tsx
```diff
@@ -16,7 +16,7 @@ const EventsComponent: React.FC<RoomVenueDetailsPartProps> = ({
   editedEvent,
   setEditedEvent,
 }) => {
-  useFirestoreConnect([
+  useSparkleFirestoreConnect([
     {
       collection: "venues",
       doc: venue.id,
```
0xdevalias, line 22: Can this ever be `undefined` here?

src/pages/SparkleSpaceMarketingPage/WelcomePage/WelcomePage.tsx
0xdevalias: It might be another thing that can be removed as part of https://github.com/sparkletown/sparkle/pull/833 ?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -61,7 +61,7 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
     window.location.hostname = redirectUrl;
   }
 
-  useFirestoreConnect({
+  useSparkleFirestoreConnect({
     collection: "venues",
     doc: venueId,
```
0xdevalias, line 66: I am pretty sure this could potentially be `undefined` based on reading it from `useParams`, which would cause a pathologic query that loads all venues.

src/types/Firestore.ts
0xdevalias: I wouldn't bother adding it back here personally. That will just confuse the fact that it is almost certainly unused.

```suggestion
```

src/types/utility.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * Type helper representing all types of T except undefined
+ */
+export type Defined<T> = T & Exclude<T, undefined>;
```
0xdevalias, line 4: Nice addition! (you can mark this resolved)

0xdevalias:
> Partially closes [sparkletown/internal-sparkle-issues#2](https://github.com/sparkletown/internal-sparkle-issues/issues/2)
> Not sure if I should implement other recommendations from this ticket,
> since we basically gain nothing in terms of linting/etc.
> 
> @0xdevalias would be glad to hear your approach on this

@mike-lvov When you say "since we basically gain nothing in terms of linting/etc", why do you feel we "gain nothing" from it? Even if there aren't any issues it would detect currently, the idea of linting is to ensure that undesirable patterns won't be introduced again in future, and to provide guidance to developers who may try to about what our 'correct' way of implementing it is.

As far as implementing the other recommendations on the ticket, see my comments in https://github.com/sparkletown/internal-sparkle-issues/issues/2#issuecomment-759176808 I believe we should be able to clean up all of those aspects as well, and then we can add a few more eslint rules to ban their usage.

mike-lvov:
After conducting extensive research, I couldn't find a way to distinguish `not passing doc` vs `passing undefined explicitly`

https://github.com/microsoft/TypeScript/issues/13195

I have found a way to solve this issue by replacing all `useParams` for `venueId` with `useVenueId` and return `string | null` from `useVenueId`. I suggest completely removing explicit `undefined` from the codebase as it is an antipattern to my opinion. This PR can be used as an argument to support my opinion

src/components/molecules/NavBar/NavBar.tsx
mike-lvov: What is the difference here? Different sorting order for the hooks among themselves?

mike-lvov:
@0xdevalias regarding Code Climate: still suggesting removing this as it does not provide any value so far. All the unreasonable duplications are spotted during the review phase by humans.

image

mike-lvov:
> RE: [#801 (comment)](https://github.com/sparkletown/sparkle/pull/801#issuecomment-758742016)
> 
> > Partially closes [sparkletown/internal-sparkle-issues#2](https://github.com/sparkletown/internal-sparkle-issues/issues/2)
> > Not sure if I should implement other recommendations from this ticket,
> > since we basically gain nothing in terms of linting/etc.
> > @0xdevalias would be glad to hear your approach on this
> 
> @mike-lvov When you say "since we basically gain nothing in terms of linting/etc", why do you feel we "gain nothing" from it? Even if there aren't any issues it would detect currently, the idea of linting is to ensure that undesirable patterns won't be introduced again in future, and to provide guidance to developers who may try to about what our 'correct' way of implementing it is.
> 
> As far as implementing the other recommendations on the ticket, see my comments in [sparkletown/internal-sparkle-issues#2 (comment)](https://github.com/sparkletown/internal-sparkle-issues/issues/2#issuecomment-759176808) I believe we should be able to clean up all of those aspects as well, and then we can add a few more eslint rules to ban their usage.

I think I conveyed my thoughts in an unclear way. I'm didn't mean that we don't need linting. I meant that replacing `import firebase from "firebase/app"` with `useFirebase` would not result in any additional type safety (as it was with useFirestore vs useSparkleFirestore). I think of other issues as low priority clean-ups.

mike-lvov:
@0xdevalias I think the problem we have been experiencing recently is a lot of unrelated changes happening in one PR. To avoid that, I would rather implement https://github.com/sparkletown/internal-sparkle-issues/issues/2 issues in another PR

src/hooks/useConnectRelatedVenues.ts
mike-lvov: it used to assign the return type of this function with `as`

image

I have refactored it to not use it. To do that, I needed to have `storeAs` as `ValidFirestoreKeys` type, which seems way more logical than having it as `string`

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: Sorting the imports alphabetically within that 'group'. The diff view here makes it look way more complex than it is, was just moving the import line up.

src/hooks/useConnectRelatedVenues.ts
0xdevalias: Makes sense

0xdevalias:
> After conducting extensive research, I couldn't find a way to distinguish `not passing doc` vs `passing undefined explicitly`
> 
> https://github.com/microsoft/TypeScript/issues/13195


> I have found a way to solve this issue by replacing all `useParams` for `venueId` with `useVenueId` and return `string | null` from `useVenueId`.

I think using `useVenueId` instead of the 'raw' `useParams` instances for this pattern is a good idea. I'd like to know more about the arguments behind the `string | null` before we do that part though.

> I suggest completely removing explicit `undefined` from the codebase as it is an antipattern to my opinion. This PR can be used as an argument to support my opinion

Can you please expand on this, and your reasoning for it? In what cases are you seeing 'explicit `undefined`' being used, and in what way do you feel like it could be done better? Why do you see `null` as a better option than it (assuming that is what you are saying here)?

---

The original PR that introduced this was https://github.com/sparkletown/sparkle/pull/784 (which was reverted in https://github.com/sparkletown/sparkle/pull/797 as it broke things in production, and then this current PR was bringing it back + fixes)

Unfortunately https://github.com/sparkletown/sparkle/pull/784 (like many of our PR's, which is something we need to change) doesn't have a useful description to outline the intended changes within it, so we have to rely on memory/inferring from the code.

Pulling from the review comments on that PR:

https://github.com/sparkletown/sparkle/pull/784#discussion_r529175072

> @0xdevalias What's the intent of `SparkleDocRFQSetting` here compared to `SparkleRFQSetting`?
>
> @thecadams It's to ensure that if your intention is to query for a single `doc`, you only query for that `doc`.
>
> @0xdevalias From @thecadams on zoom (paraphrased): essentially we had times when a `doc` that we were supplying was `undefined`, and so we were loading the entire collection (because it's an optional param in the underlying `useFirestoreConnect` hook), when our intent was to load a single tem or bailout.

The relevant type-related code is at https://github.com/sparkletown/sparkle/pull/784/files#diff-254a400761084e5ebd89145b80d201db7f0fd1e2dff8684213ee5d1804ee886aR11-R45

```typescript
export interface SparkleRFQuery extends ReduxFirestoreQuerySetting {
  doc?: never;
  where: WhereOptions | WhereOptions[];
  storeAs?: ValidFirestoreKeys;
}

export interface SparkleRFDocQuery extends ReduxFirestoreQuerySetting {
  doc: string;
  where?: never;
  storeAs?: ValidFirestoreKeys;
}
```

Inferring from this code (without actually reading deeply/playing with it), I believe the intent is that:

- When you pass `doc`, you can't pass `where` (intended to access 1 specific document, so no need for filtering)
- When you pass `where`, you can't pass `doc` (because you aren't accessing a single document, but filtering over a collection of them)

---

You can also see that `useSparkeFirestoreConnect` was deleted in this PR, in favour of the more generalised custom `useFirestoreConnect` implementation:

- https://github.com/sparkletown/sparkle/pull/784/files#diff-53c83c5e9a6c0e4d642a0a04aee496008588554b3abebce7a1a14d5aa88bac24

mike-lvov:
> Inferring from this code (without actually reading deeply/playing with it), I believe the intent is that:
> 
> When you pass doc, you can't pass where (intended to access 1 specific document, so no need for filtering)
> When you pass where, you can't pass doc (because you aren't accessing a single document, but filtering over a collection of them)

Even though I support this intent, it didn't seem to work right. The current implementation of this(`null`) works good because then you try to pass `null` there it doesn't like it. So a type `string | null` will not be accepted by TS. That is the main reasoning behind `useVenueId` change.

0xdevalias:
> regarding Code Climate: still suggesting removing this as it does not provide any value so far.

That's a pretty big stretch from reality, as it has definitely added value, and in my opinion continues to do so.

I do agree that (particularly with these sweeping tech debt cleanup PRs) the 'inline comment on PR' feature can get fairly spammy feeling. I would be open to disabling this part of it.

---

As I originally said in https://github.com/sparkletown/sparkle/pull/1045#pullrequestreview-555072007, I still remain open to us tweaking the config to better suit our needs:

> I’d be curious to know specifically what aspects (or patterns of issue it’s picking up) that you’re finding ‘too spammy’?

> if there are any particular ‘rules’ that you feel are popping up too often/as false positives/etc, that would be useful to know.

---

> All the unreasonable duplications are spotted during the review phase by humans.
"as it does not provide any value so far"

While they may be spotted during review by humans, the more that we can lean on automated tooling to spot it before a human needs to means we have to spend less human time during review, allowing us to be more agile and efficient. Right now I feel like I can barely get to writing actual code because of the number of tasks (including review) that I need to spend my time on.

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/801#issuecomment-759871145

> I believe also if we clicked through to the codeclimate 'review screen' (linked from the comment it adds to PR's) and marked the things that it shows there as valid/invalid (+why), then it probably wouldn't have continued to mention them in every new commit.

---

Using Code Climate's 'summary' comment:
image

---

Clicking 'View more on Code Climate' will open the following url:

- https://codeclimate.com/github/sparkletown/sparkle/pull/801

Which summarises the issues found, whether they are new/existing, etc:
image

---

Clicking on the 'New' status label for one of the issues, we get the following menu, which can be used to change the status of the issue.

**IMPORTANT:** Don't just arbitrarily change this status to something to 'make the noise go away'.

image

---

Reviewing the 3 'New' changes called out in https://codeclimate.com/github/sparkletown/sparkle/pull/801:

- The first 2 relate to repeated `useFirestoreConnect` code that connects to `collection: "purchases"` in different files.
- We can review the code in both of these files to see if it is in fact a duplicate, or a false positive:
  - https://codeclimate.com/github/sparkletown/sparkle/src/hooks/useConnectCurrentEvent.ts/source?from_sha=78069e72&to_sha=97e96cec
  - https://codeclimate.com/github/sparkletown/sparkle/src/hooks/useConnectUserPurchaseHistory.ts/source?from_sha=78069e72&to_sha=97e96cec
- Since this code is exactly the same, this is a valid finding, so I will mark it 'Confirmed':
  - image
- And then explain my reasoning in the comment:
  - image
- We can also use the 'Create a ticket' button (that becomes visible on the far right when we hover over the issue) to create a backlog task for fixing this issue, if we wanted. For most of these little code duplication things I don't normally bother, as we can always review them within the list of remaining Code Climate issues, then work on them from there. But it's good to know that the feature exists.
  - image
- The 3rd issue that is shown here is a false positive, as while they both use `useFireStoreConnect` to connect to `collection: "venues"` with `doc: venueId`, they do very different things after that, so I will mark it as 'Invalid'
  - image
- The rest of the things noted in the summary are previous issues that have been fixed, so we don't need to worry about those.

---

Once we are happy that we have categorized any of the found issues appropriately (be that as 'Confirmed', 'Invalid' or otherwise), then we can use the main 'Approve' button, and give a reason as to why we are doing so.

image

This will remove the red `X` from the PR's Code Climate status check.

0xdevalias:
> I think I conveyed my thoughts in an unclear way. I'm didn't mean that we don't need linting. I meant that replacing `import firebase from "firebase/app"` with `useFirebase` would not result in any additional type safety (as it was with useFirestore vs useSparkleFirestore). I think of other issues as low priority clean-ups.

That makes sense. I agree that they are lower priority. 👍🏻 

> @0xdevalias I think the problem we have been experiencing recently is a lot of unrelated changes happening in one PR. To avoid that, I would rather implement [sparkletown/internal-sparkle-issues#2](https://github.com/sparkletown/internal-sparkle-issues/issues/2) issues in another PR

I'm happy for that to be how we approach it. 👍🏻

src/hooks/useFirestoreConnect.ts
```diff
@@ -40,9 +55,8 @@ export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
  * @see ValidFirestoreKeys
  * @see ReduxFirestoreQuerySetting
  */
-export const useSparkleFirestoreConnect = (
-  config?: SparkleRFQConfig[] | SparkleRFQConfig
-) => useFirestoreConnect(config);
+export const useFirestoreConnect = (config: SparkleRFQConfig) =>
+  _useFirestoreConnect(config);
```
0xdevalias, line 59: This is a better way to implement this, and to not include `undefined` in the definition of `SparkleRFQConfig`
```suggestion
export const useFirestoreConnect = (config?: SparkleRFQConfig) =>
  _useFirestoreConnect(config);
```

src/hooks/useVenueId.ts
```diff
@@ -11,16 +11,18 @@ const venuePaths = [
...
       path,
     });
 
-    if (match && match.params.venueId) {
+    if (match?.params.venueId) {
       return match.params.venueId;
     }
   }
+
+  return null;
```
0xdevalias, line 27: I still don't understand what about using `null` here makes this 'better' or 'more right' than using `undefined`?

src/pages/Admin/Admin.tsx
```diff
@@ -67,6 +62,11 @@ import VenueDeleteModal from "./Venue/VenueDeleteModal";
 import { VenueOwnersModal } from "./VenueOwnersModal";
 
 import "./Admin.scss";
+import {
+  SparkleRFQConfig,
+  useFirestoreConnect,
+} from "hooks/useFirestoreConnect";
+import { useVenueId } from "hooks/useVenueId";
```
0xdevalias, line 69: Sort with other `hooks/` imports

src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -9,7 +9,7 @@ import { RoomInput, upsertRoom } from "api/admin";
 import { useUser } from "hooks/useUser";
 import { useSelector } from "hooks/useSelector";
 import VenueEventDetails from "./VenueEventDetails";
-import { useFirestoreConnect } from "react-redux-firebase";
+import { useFirestoreConnect } from "hooks/useFirestoreConnect";
```
0xdevalias, line 12: Sort with other `hooks/` imports

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -40,10 +41,13 @@ export const RoomsForm: React.FC = () => {
 
   useEffect(() => {
     const fetchVenueFromAPI = async () => {
+      if (!venueId) return history.replace("/admin");
+
```
0xdevalias, line 45: Why are we redirecting within this function? I assume the intent of this addition was to ensure `venueId` isn't `undefined`/similar by the time it gets passed to `.doc()`. Why wouldn't we just `return` early like we do with all of our other guard patterns like this?

```suggestion
      if (!venueId) return
```

0xdevalias: I see that later on this same `history.replace` code is used, so that makes sense.

Though it seems we could move this guard outside of the `async` `fetchVenueFromAPI` function and just have it in the root of the `useEffect`?

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,38 +171,43 @@ export const useConnectRelatedVenues: ReactHook<
...
+        makeEventsQueryConfig(
+          subvenue.id,
+          `subvenueEvents-${subvenue.id}` as ValidFirestoreKeys // @debt a little hacky, but we're consciously subverting our helper protections;
+        )
       )
     : [];
 
   // Combine / filter for valid queries
-  const allValidQueries: SparkleRFQConfig[] = [
+  const allValidQueries: SparkleRFQConfig = [
```
0xdevalias, line 210: This feels more confusing than how it used to be.. probably would have named your 'fusion' type as `AnySparkleRFQConfig` or something specific about the props the function accepts.

Alternatively, given the change has already been done, this would probably be more correct/easy to understand as `SparkleRFQConfigObject[]`. 'Hiding' the array'ness of it makes it confusing, and I need to go any look up the original 'fusion type' to understand whats going on.

src/hooks/useFirestoreConnect.ts
```diff
@@ -1,9 +1,11 @@
...
   ReduxFirestoreQuerySetting,
-  useFirestoreConnect,
+  useFirestoreConnect as _useFirestoreConnect,
   isLoaded as _isLoaded,
   isEmpty as _isEmpty,
 } from "react-redux-firebase";
+
 import { ValidFirestoreKeys } from "types/Firestore";
 
 /**
```
0xdevalias, line 11: `Defined<T>` is already defined in `src/types/utility.ts` so doesn't need to exist here

0xdevalias (commented):
Includes requested changes (I can't use the official 'request changes' review mode as I opened the original PR)

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/801#issuecomment-759815402

> Even though I support this intent, it didn't seem to work right. The current implementation of this(`null`) works good because then you try to pass `null` there it doesn't like it. So a type `string | null` will not be accepted by TS. That is the main reasoning behind `useVenueId` change.

I'll open a sub-PR that fixes this properly shortly, just polishing it locally.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
mike-lvov: I think so, but it still would be the same

src/hooks/useFirestoreConnect.ts
mike-lvov: This was useful when we still had separated queries. And a query could be undefined. 

`const query: SparkleRFQConfig = userId ? { querydata } : undefined;`

src/hooks/useFirestoreConnect.ts
```diff
@@ -15,23 +17,36 @@ export type Defined<T> = T & Exclude<T, undefined>;
...
+ * This type lists all allowed argument types for @useFirestoreConnect function
+ *
+ * @see SparkleUseFirestoreConfigObject
+ * @see ValidFirestoreKeys
+ */
+export type SparkleRFQConfig =
+  | SparkleUseFirestoreConfigObject
+  | SparkleUseFirestoreConfigObject[]
+  | ValidFirestoreKeys
+  | undefined;
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/801#discussion_r557260510

src/hooks/useFirestoreConnect.ts
```diff
@@ -15,23 +17,36 @@ export type Defined<T> = T & Exclude<T, undefined>;
...
- * @see useSparkleFirestoreConnect
+ * @see useFirestoreConnect
  * @see ReduxFirestoreQuerySetting
  */
-export interface SparkleRFQConfig extends ReduxFirestoreQuerySetting {
+
+export type SparkleUseFirestoreConfigObject = ReduxFirestoreQuerySetting & {
   storeAs?: ValidFirestoreKeys;
-}
+};
```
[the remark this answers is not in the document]
mike-lvov: Since you have done the same, I would think this should be a pattern now? We don't really need interfaces since it has basically become the same as the type

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,38 +171,43 @@ export const useConnectRelatedVenues: ReactHook<
...
   // Parent Events
-  const parentVenueEventsQuery: SparkleRFQConfig | undefined =
+  const parentVenueEventsQuery =
     parentId && withEvents
       ? makeEventsQueryConfig(parentId, "parentVenueEvents")
       : undefined;
 
   // Sibling Events
-  const siblingVenueEventsQueries: SparkleRFQConfig[] = withEvents
+  const siblingVenueEventsQueries = withEvents
```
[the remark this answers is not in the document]
mike-lvov: Good point

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,38 +171,43 @@ export const useConnectRelatedVenues: ReactHook<
...
+        makeEventsQueryConfig(
+          sibling.id,
+          `siblingVenueEvents-${sibling.id}` as ValidFirestoreKeys // @debt a little hacky, but we're consciously subverting our helper protections;
+        )
       )
     : [];
 
   // Sub Events
-  const subvenueEventsQueries: SparkleRFQConfig[] = withEvents
+  const subvenueEventsQueries = withEvents
```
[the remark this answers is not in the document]
mike-lvov: Good point

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
0xdevalias: Thats fine. I think it just more confused me being inside the inner function, rather than the root of the `useEffect`

src/hooks/useFirestoreConnect.ts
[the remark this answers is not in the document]
0xdevalias: I'm still not actually sure what the difference between the two styles is in this use case.. but happy to leave it with the `type` thing, as it seems to work.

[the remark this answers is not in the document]
mike-lvov: There was a difference in the previous TS version(you could only use interface to type objects AFAIK). Currently, the difference is strongly subjective

src/hooks/useFirestoreConnect.ts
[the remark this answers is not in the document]
0xdevalias: I fixed this up in my PR, so no longer an issue, but I feel like using `undefined` in a grouping type like this will likely cause more issues/confusion, so would be best to avoid it in future code.

We are still able to pass the `undefined` queries to the `useFirestoreConnect` function because of the `config?` 'optional' syntax.

src/hooks/useFirestoreConnect.ts
0xdevalias: https://github.com/sparkletown/sparkle/pull/801#discussion_r557267695

We're still able to use those `undefined` queries because of the `?` 'optional' operator.

If we still had separated queries, my preference would be fore them to be explicit rather than 'hiding' the fact that they contain `undefined`:

```
const query: SparkleRFQConfig | undefined = userId ? { querydata } : undefined;
```

src/hooks/useFirestoreConnect.ts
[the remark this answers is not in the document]
0xdevalias: Interesting.. maybe that's where I got the `interface` pattern from.

I think we seem to use this way more than the `interface` way, so let's say it's our chosen pattern until we run into a reason why it doesn't work.

src/hooks/useConnectRelatedVenues.ts
```diff
@@ -170,38 +171,43 @@ export const useConnectRelatedVenues: ReactHook<
...
+    collection: "venues",
+    doc,
+    subcollections: [{ collection: "events" }],
+    orderBy: ["start_utc_seconds", "asc"],
+    storeAs,
+  });
 
   // Parent Events
-  const parentVenueEventsQuery: SparkleRFQConfig | undefined =
+  const parentVenueEventsQuery =
```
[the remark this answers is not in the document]
0xdevalias: @mike-lvov can you add the explicit typing back here please

src/hooks/useVenueId.ts
```diff
@@ -11,6 +11,10 @@ const venuePaths = [
   "/v/:venueId",
   "/admin_v2/venue/:venueId",
 ];
+
+// @debt I think we could just use useParams() here instead of needing to loop through all of the venuePath, so long as the use :venueId in their path
+//   https://reactrouter.com/web/api/Hooks/useparams
+//     useParams returns an object of key/value pairs of URL parameters. Use it to access match.params of the current <Route>.
```
0xdevalias, line 17: [OOS] see https://github.com/sparkletown/internal-sparkle-issues/issues/164

0xdevalias (commented):
Skimming over the changed files I only had 1 small request, then i'm happy for this PR to be merged.

Since I was the one who originally opened it, I can't actually approve it. But once you make the change mentioned here, feel free to approve and merge it.

(When squash-merging, make sure to keep the 'co-authored by' text it adds)

mike-lvov (approved):
LGTM!

PR #809, opened 2020-11-25
fix room layout when grid is missing

denisdimitrov:


src/components/templates/Camp/components/Map.tsx
```diff
@@ -269,8 +269,8 @@ export const Map: React.FC<MapProps> = ({
         gridTemplateRows: `repeat(${templateRows}, 1fr)`,
       }}
     >
-      {showGrid && mapGrid}
-      {showGrid && partygoersOverlay}
+      {mapGrid}
+      {partygoersOverlay}
```
0xdevalias, line 273: This is fine because the hooks are already using showGrid and handling this for us. This was a redundant check.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -249,8 +249,8 @@ export const Map: React.FC<PropsType> = ({
               gridTemplateRows: `repeat(${rows}, 1fr)`,
             }}
           >
-            {showGrid && mapGrid}
-            {showGrid && partygoersOverlay}
+            {mapGrid}
+            {partygoersOverlay}
```
0xdevalias, line 253: This is fine because the hooks are already using showGrid and handling this for us. This was a redundant check.

0xdevalias (approved):
Thanks for catching these from my last PR! LGTM! :shipit:

PR #810, opened 2020-11-25
tweak room counting and uncounting

denisdimitrov:
- fixes camp room counting display which was in column instead of row. now number of room attendance is next to the icon
- useCampPartygoers now checks for online users which were in the camp between the last update and now, instead of just now.
- remove mutation of prop
- replace anchors with buttons to disable new tab opening with wheel click and right click

### DIFF
--- a/src/components/molecules/UserList/UserList.tsx
+++ b/src/components/molecules/UserList/UserList.tsx
@@ -31,7 +31,7 @@ interface PropsType {
 }
 
 const UserList: React.FunctionComponent<PropsType> = ({
-  users: userList,
+  users: _users,
   limit = DEFAULT_USER_LIST_LIMIT,
   imageSize = 40,
   activity = "partying",
@@ -44,14 +44,16 @@ const UserList: React.FunctionComponent<PropsType> = ({
   const [selectedUserProfile, setSelectedUserProfile] = useState<
     WithId<User>
   >();
-  const users = userList?.filter(
+  const usersSanitized = _users?.filter(
     (user) => !user.anonMode && user.partyName && user.id
-  ); // quick fix to get rid of anonymous users
-  const usersToDisplay = isExpanded ? users : users?.slice(0, limit);
-  const attendance = users.length + (attendanceBoost ?? 0);
+  );
+  const usersToDisplay = isExpanded
+    ? usersSanitized
+    : usersSanitized?.slice(0, limit);
+  const attendance = usersSanitized.length + (attendanceBoost ?? 0);
   const venue = useSelector(currentVenueSelectorData);
 
-  if (!users || attendance < 1) return <></>;
+  if (!usersSanitized || attendance < 1) return <></>;
   return (
     <>
       <div className="userlist-container">
@@ -61,7 +63,7 @@ const UserList: React.FunctionComponent<PropsType> = ({
             {attendance === 1 ? "person" : "people"}{" "}
             {isCamp && IS_BURN ? "in the camp" : activity}
           </p>
-          {!disableSeeAll && users.length > limit && (
+          {!disableSeeAll && usersSanitized.length > limit && (
             <p
               className="clickable-text"
               onClick={() => setIsExpanded(!isExpanded)}
--- a/src/components/molecules/UserList/UserList.tsx
+++ b/src/components/molecules/UserList/UserList.tsx
@@ -31,7 +31,7 @@ interface PropsType {
 }
 
 const UserList: React.FunctionComponent<PropsType> = ({
-  users,
+  users: userList,
   limit = DEFAULT_USER_LIST_LIMIT,
   imageSize = 40,
   activity = "partying",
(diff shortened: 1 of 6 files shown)


denisdimitrov:
I tested this with a couple of browsers open and looks stable, needs to be added in staging and tested by more ppl.

denisdimitrov:
@0xdevalias @IvanBuljovcic @thecadams This is updated and tested. Behaves good, and the anchors are now replaced with buttons, to guarantee left mouse click opening of the new tab.

src/components/molecules/UserList/UserList.tsx
```diff
@@ -31,7 +31,7 @@ interface PropsType {
 }
 
 const UserList: React.FunctionComponent<PropsType> = ({
-  users,
+  users: userList,
```
0xdevalias, line 34: [NB, nit] I usually use the convention of `_VARNAME` (eg. `_users`) when 'shadowing' a variable like this. I find it helps me to know that I shouldn't really be using that variable 'deeper' within the component, and then I set whatever override/fix I need to it in the first couple of lines of the component.

Alternatively, we could be more semantice with the name here? eg. `allUsers`, and then further down `const visibleUsers` or similar? Or even just keep this as `users` and then call the other `visibleUsers`?

src/components/molecules/UserList/UserList.tsx
```diff
@@ -44,7 +44,9 @@ const UserList: React.FunctionComponent<PropsType> = ({
   const [selectedUserProfile, setSelectedUserProfile] = useState<
     WithId<User>
   >();
-  users = users?.filter((user) => !user.anonMode && user.partyName && user.id); // quick fix to get rid of anonymous users
+  const users = userList?.filter(
+    (user) => !user.anonMode && user.partyName && user.id
+  ); // quick fix to get rid of anonymous users
```
0xdevalias, line 49: Is this still counted as a 'quick fix'? We're also removing the anonMode users.. is that what we want to be doing? And if so, what makes this a 'quick fix' rather than just our intended future functionality here?

src/components/templates/Camp/components/RoomModalOngoingEvent.tsx
```diff
@@ -32,6 +32,19 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
     roomEvents.length > 0 &&
     (currentEvent ? currentEvent : roomEvents[0]);
 
+  const joinRoom = useCallback(() => {
+    enterRoom();
+    window.open(getRoomUrl(room.url), "_blank", "noopener,noreferrer");
```
0xdevalias, line 37: We have a function in `utils/url` that already does all of this I believe. Let's use that here.

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -30,6 +30,19 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
...
+  }, [enterRoom, room.url]);
+
+  const triggerAttendance = useCallback(() => {
+    dispatch(retainAttendance(true));
+  }, [dispatch]);
+
+  const clearAttendance = useCallback(() => {
+    dispatch(retainAttendance(false));
+  }, [dispatch]);
+
```
0xdevalias, line 45: These are the same as in `src/components/templates/Camp/components/RoomModalOngoingEvent.tsx`, should we extract them to DRY up the code?

I see that the other place is in the legacy Camp Map component, so it may not be worth doing since we are going to remove it soon. So probably just the same comment as above RE: using the helper from `utils/url` that we have instead of `window.open`

src/hooks/useCampPartygoers.ts
```diff
@@ -31,7 +31,9 @@ export const useCampPartygoers = (venueName: string): WithId<User>[] => {
   return useMemo(
     () =>
       partygoers?.filter(
-        (partygoer) => partygoer?.lastSeenIn?.[venueName] > lastSeenThresholdMs
+        (partygoer) =>
+          partygoer?.lastSeenIn?.[venueName] >
+          lastSeenThresholdMs - LOC_UPDATE_FREQ_MS
```
0xdevalias, line 36: Nice catch!

0xdevalias (changes requested):
A couple of tiny cleanup things/questions/etc, but then this should be good to go!

Nice catch on the links vs buttons potentially not catching our events! As well as the time update part!

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
thecadams: We are removing the legacy Camp component asap, so won't worry about DRYing this up for this review, but there's bound to be DRYups in future that we'll access once this is done.

thecadams:

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -30,6 +30,19 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
...
+    openUrl(getRoomUrl(room.url));
+  }, [enterRoom, room.url]);
+
+  const triggerAttendance = useCallback(() => {
+    dispatch(retainAttendance(true));
+  }, [dispatch]);
+
+  const clearAttendance = useCallback(() => {
+    dispatch(retainAttendance(false));
+  }, [dispatch]);
```
0xdevalias, line 44: DRY with `src/components/templates/Camp/components/RoomModalOngoingEvent.tsx`, but since I believe that is one that we will be removing shortly, this is fine.

0xdevalias (commented):
Only comment is a probably irrelevant DRY. So this looks good!

PR #818, opened 2020-11-26
fetch private messages by chat user ids

denisdimitrov:


IvanBuljovcic (approved):

src/components/molecules/Sidebar/Sidebar.tsx
```diff
@@ -18,18 +17,34 @@ enum TABS {
...
-    collection: "users",
-    where: ["enteredVenueIds", "array-contains", venueId],
-    storeAs: "chatUsers",
-  });
   const [tab, setTab] = useState(0);
   const privateChats = useSelector(privateChatsSelector);
   const chatUsers = useSelector(chatUsersSelector);
   const isEnabled = chatUsers && privateChats;
 
+  const chatUserIds = privateChats
```
denisdimitrov, line 32: useMemo?

0xdevalias: Yeah, it should.

src/components/molecules/Sidebar/Sidebar.tsx
```diff
@@ -18,18 +17,34 @@ enum TABS {
   LIVE_SCHEDULE = 2,
 }
 
+const DOCUMENT_ID = "__name__";
+
+// Maybe move this to  utils?
+const filterUniqueKeys = (userId: string, index: number, arr: string[]) =>
+  arr.indexOf(userId) === index;
```
0xdevalias, line 24: Agree on moving this to `utils`. I tend to name my filters `fooprefixFilter`, so in this case: `uniqueKeysFilter`. So that when it is written out it will be `foo.filter(uniqueKeysFilter)`

PR #820, opened 2020-11-27
✨ Generate component utility script ✨

IvanBuljovcic:
A small utility script for creating new components, so that we don't have to manually create files and add boilerplating to the new files.

`npm run generate [ComponentName] [ComponentPath]`

**ComponentName** is required
**ComponentPath**:
- if it is not passed, the new component will be created in `src/components/[ComponentName]`
- If the path is passed as an argument, the new component will be created in `src/[ComponentPath]/[ComponentName]`

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
     "react-switch": "^5.0.1",
+    "react-test-renderer": "^17.0.1",
     "react-webcam": "^5.2.0",
     "read": "^1.0.7",
     "redux": "^4.0.5",
@@ -83,7 +84,8 @@
     "cy:install": "cypress install",
     "cy:run": "cypress run",
     "firebase": "npx --no-install -- firebase",
-    "firebase:emulate-functions": "npm run firebase -- emulators:start --only functions"
+    "firebase:emulate-functions": "npm run firebase -- emulators:start --only functions",
+    "generate": "node ./utils/create-component"
   },
   "proxy": "http://localhost:3001",
   "browserslist": {
@@ -118,11 +120,13 @@
     "@types/react-mic": "^12.4.2",
     "@types/react-redux": "^7.1.9",
     "@types/react-router-dom": "^5.1.5",
+    "@types/react-test-renderer": "^16.9.3",
     "@types/styled-components": "^5.1.4",
     "@types/twilio-video": "^2.0.12",
     "@types/yup": "^0.29.3",
     "@typescript-eslint/eslint-plugin": "^3.7.1",
     "@typescript-eslint/parser": "^3.7.1",
+    "colors": "^1.4.0",
     "cypress": "^4.11.0",
     "cypress-file-upload": "^4.0.7",
     "eslint-config-react-app": "^5.2.1",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -53,6 +53,7 @@
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
     "react-switch": "^5.0.1",
+    "react-test-renderer": "^17.0.1",
```
0xdevalias, line 56: Is this being used here?

package.json
```diff
@@ -118,11 +120,13 @@
     "@types/react-mic": "^12.4.2",
     "@types/react-redux": "^7.1.9",
     "@types/react-router-dom": "^5.1.5",
+    "@types/react-test-renderer": "^16.9.3",
```
0xdevalias, line 123: Is this being used here?

utils/create-component.js
```diff
@@ -0,0 +1,56 @@
+require("colors");
```
0xdevalias, line 1: This feels like it might belong in the `scripts/` folder? That's where we have our other helper scripts.

utils/create-component.js
```diff
@@ -0,0 +1,56 @@
...
+    fs.writeFileSync(
+      `${componentDirectory}/index${template.extension}`,
+      template.content
+    );
+  } else {
+    fs.writeFileSync(
+      `${componentDirectory}/${componentName}${template.extension}`,
+      template.content
+    );
+  }
```
0xdevalias, line 51: If there is an existing file at that path, will it overwrite it?

package.json
denisdimitrov, line 56: Since this is not being used runtime we can move it to devDependencies.

denisdimitrov: It's not being used anywhere either, I suppose we are adding this for future snapshot use?

0xdevalias (commented):
A few comments, but I'll let @thecadams make the call on this.

I normally just use a snippets manager (on my dev machine) for this sort of task, though if we wanted something included here, I wonder if an existing template 'code generator' such as https://github.com/jondot/hygen (or something similar, this was just a quick google) might be a better choice?

package.json
0xdevalias: If we're not planning on using it here, then let's add it in a PR that uses it, to keep things clean/changes easy to follow.

utils/create-component.js
IvanBuljovcic: If the specified path exists, it creates a component within that path.

`npm run generate Test atoms/Test`

it will generate in `src/components/atoms/Test/Test`

if the `.../Test/Test` path already exists, it generates one level down so: `src/components/atoms/Test/Test/Test`

That case is, i believe, a user error for which there shouldnt be too much time spent fixing.
The fix for this is to just amend the `/Test` from the path argument

package.json
0xdevalias: If not, It should be removed. I think we should introduce it in it's own PR.

thecadams:
Hi gang,
This PR is a great idea and pushing in a very important direction. However we are not yet at a codebase size or velocity where codegen helps us. There will come a time in the future at which this type of tooling will help but we're not there yet and so shouldn't carry the maintenance burden of this code yet.

Going to close this to help us focus on near term needs. Thanks for the contribution.

PR #822, opened 2020-11-27
[FEATURE] Party map zoomability and preview

denisdimitrov:
- adds preview of the whole map when grid is on
- adds zoomability of the whole map when grid is on
- allows user to change background image in admin for party map


src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
     [attendances, setIsRoomModalOpen, setSelectedRoom, venue]
   );
+
+  const zoomIn = useCallback(() => {
+    setZoom((zoom) => Math.min(zoom + ZOOM_INCREMENT, MAX_ZOOM));
+  }, []);
+
+  const zoomOut = useCallback(() => {
+    setZoom((zoom) => Math.max(zoom - ZOOM_INCREMENT, MIN_ZOOM));
+  }, []);
```
0xdevalias, line 255: Good use of state update pattern! I always forget that this method exists (which is the correct method to avoid using potentially stale data/race conditions/etc relating to updating the state, based on the existing state.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
+
   if (!user || !venue) {
     return <>Loading map...</>;
   }
 
+  const translateX =
+    (((mapSize.width * zoom - (mapSize.width * zoom) / zoom) / zoom) * 2) / 4;
+  const translateY =
+    (((mapSize.height * zoom - (mapSize.height * zoom) / zoom) / zoom) * 2) / 4;
```
0xdevalias, line 264: There's a lot of maths going on here.. can we simplify it at all? If not (or as well as), we should add some comments to help explain whats going on here.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
+  const bgRatio = mapSize.width ? mapSize.width / mapSize.height : 1;
+
+  const previewAvatarPosition = {
+    x: currentPosition?.row ? (currentPosition?.row / rows) * 100 - 2 : 0,
+    y: currentPosition?.column
+      ? (currentPosition?.column / columns) * 100 - 2
+      : 0,
+  };
```
0xdevalias, line 272: `useMemo` as this is an object (that will be recreated on every render) and being passed as a prop, otherwise will cause re-renders.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
+
+  const zoomableMapSize = showGrid
+    ? {
+        width: mapSize.width,
+        height: mapSize.height,
+      }
+    : {};
```
0xdevalias, line 279: `useMemo`

But also all of the `style` blocks in the JSX will need their own `useMemo`'s too

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
+              backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
               gridTemplateColumns: `repeat(${columns}, calc(100% / ${columns}))`,
               gridTemplateRows: `repeat(${rows}, 1fr)`,
             }}
           >
             {mapGrid}
             {partygoersOverlay}
-            {roomOverlay}
           </div>
+          {roomOverlay}
```
0xdevalias, line 303: Why did this one shift outside of the container?

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -227,32 +245,62 @@ export const Map: React.FC<PropsType> = ({
...
+          style={{
+            ...zoomableMapSize,
+            transform: `scale(${zoom}) translate3d(${translateX}px, ${translateY}px, 0)`,
+          }}
+        >
           <div
             className="party-map-grid-container"
             style={{
+              ...zoomableMapSize,
+              backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
```
0xdevalias, line 295: Do we need to spread `zoomableMapSize` in both of these locations? if so, why is that?

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -262,6 +310,26 @@ export const Map: React.FC<PropsType> = ({
...
+              onZoomIn={zoomIn}
+              onZoomOut={zoomOut}
+            />
+
+            <PartyMapPreview
+              expanded={previewExpand}
+              backgroundImage={venue.mapBackgroundImageUrl}
+              backgroundRatio={bgRatio}
+              avatarPosition={previewAvatarPosition}
+              onIconClick={() => setPreviewExpand(!previewExpand)}
```
0xdevalias, line 328: this will need to be extracted + `useCallback`

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -274,6 +274,11 @@ interface DetailsFormLeftProps {
   formError: boolean;
 }
 
+const backgroundTextByVenue: { [key: string]: string } = {
+  [VenueTemplate.themecamp]: "Theme Camp",
+  [VenueTemplate.partymap]: "Party Map",
+};
```
0xdevalias, line 280: I believe `Record<string, string>` is equivalent to this type.

Is it possible for there to be any other venue values here? We don't currently have a default set. Might be good to add one just in case?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -446,7 +451,7 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
           {templateID && BACKGROUND_IMG_TEMPLATES.includes(templateID) && (
             <>
               <h4 className="italic" style={{ fontSize: "20px" }}>
-                {`Choose the background for your Theme Camp`}
+                {`Choose the background for your ${backgroundTextByVenue[templateID]}`}
```
0xdevalias, line 454: add default fallback if `backgroundTextByVenue` doesn't have a match

something like:
```suggestion
                {`Choose the background for your ${backgroundTextByVenue[templateID] ?? 'FOO DEFAULT'}`}
```

src/pages/Admin/Venue/DetailsForm.tsx
0xdevalias: Should this type be resticted to keys in `BACKGROUND_IMG_TEMPLATES` ?

src/components/templates/PartyMap/components/ZoomControls/ZoomControls.tsx
```diff
@@ -0,0 +1,28 @@
...
+  disableZoomOut: boolean;
+  onZoomIn: () => void;
+  onZoomOut: () => void;
+}
+
+export const ZoomControls: FC<ZoomControlsProps> = ({
+  disableZoomIn,
+  disableZoomOut,
+  onZoomIn,
+  onZoomOut,
```
0xdevalias, line 14: `disableZoomIn`/`disableZoomOut` could also potentially be wrapped into `onZoomIn`/`onZoomOut` functions so that they are a noop if called (in addition to setting the `disabled` class as you are here)

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+    y: number;
+  };
+  rooms: PartyMapRoomData[];
+  onIconClick: () => void;
+}
+
+const PREVIEW_WIDTH_PIXELS = 120;
+const PREVIEW_WIDTH_VH = 80;
+
+export const PartyMapPreview: FC<PartyMapPreviewProps> = ({
```
0xdevalias, line 19: [nit] I usually prefer to use `React.FC` here for better context, rather than importing it individually. `FC` on it's own doesn't provide as much immediate/meaningful context as using the full `React.FC` in my opinion.

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { PartyMapRoomData } from "types/PartyMapRoomData";
+
+interface PartyMapPreviewProps {
+  expanded: boolean;
+  backgroundImage?: string;
+  backgroundRatio: number;
+  avatarPosition: {
+    x: number;
+    y: number;
+  };
```
0xdevalias, line 11: This can use `Point` from `src/types/utility.ts`
```suggestion
  avatarPosition: Point;
```

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { FC } from "react";
+import { PartyMapRoomData } from "types/PartyMapRoomData";
+
+interface PartyMapPreviewProps {
+  expanded: boolean;
```
0xdevalias, line 5: Maybe?

```suggestion
  isExpanded: boolean;
```

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+  avatarPosition: {
+    x: number;
+    y: number;
+  };
+  rooms: PartyMapRoomData[];
+  onIconClick: () => void;
+}
+
+const PREVIEW_WIDTH_PIXELS = 120;
+const PREVIEW_WIDTH_VH = 80;
```
0xdevalias, line 17: Should these be set in pixels/etc? Will this play nice with responsive design/etc?

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  // Expanded preview width and height are using vh
+  // Unexpanded preview width and height are using pixels
+  const previewWidth = expanded
+    ? `${PREVIEW_WIDTH_VH}vh`
+    : PREVIEW_WIDTH_PIXELS;
+  const previewHeight = expanded ? expandedHeight : defaultHeight;
+
+  const avatarTop = avatarPosition.x + "%";
+  const avatarLeft = avatarPosition.y + "%";
```
0xdevalias, line 39: Could also do:
```suggestion
  const avatarTop = `${avatarPosition.x}%`;
  const avatarLeft = `${avatarPosition.y}%`;
```

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <div
+      className={`map-zoom-preview ${expanded && "expand"}`}
+      style={{
+        backgroundImage: `url(${backgroundImage})`,
+        width: previewWidth,
+        height: previewHeight,
+      }}
```
0xdevalias, line 47: this should be extracted + `useMemo`

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+    >
+      <div className="map-zoom-icon" onClick={onIconClick}></div>
+      <div className="map-zoom-position"></div>
+      <div
+        className="map-avatar-position"
+        style={{ top: avatarTop, left: avatarLeft }}
```
0xdevalias, line 53: this should be extracted + `useMemo`

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+                width: width + "%",
+                height: height + "%",
+                backgroundImage: `url(${room.image_url})`,
+              }}
+            />
+          );
+        })}
+      </div>
+    </div>
+  );
```
0xdevalias, line 77: this should be extracted + `useMemo` (or made into a sub-component.. but it feels too small to be worth doing for this)

0xdevalias: `index` isn't a great choice for key as it's not guaranteed to be stable. but `room.title` isn't unique.. so.. this is probably fine.. I don't really have a better solution in mind at the moment.

0xdevalias (changes requested):
A bunch of comments, mostly around memo'ing/etc + some minor style things + avoiding adding extra `.png`s to the git repo if we can.

src/components/templates/PartyMap/components/ZoomControls/ZoomControls.tsx
```diff
@@ -0,0 +1,28 @@
...
+}) => {
+  return (
+    <div className="party-map-zoom-container">
+      <div
```
thecadams, line 18: Wondering if a `button` is better for accessibility / browser behavior - eg. it would get a different cursor appearance by default I believe.

0xdevalias: (sidenote: @thecadams if you select multiple lines in the diff when making comments you can show all the relevant context without having to click through to opening the file)

denisdimitrov:
[Zoomed in grid amazingness](https://trello.com/c/DfffrLHN/291-zoomed-in-grid-amazingness)

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -8,14 +8,180 @@ $border-radius: 28px;
...
+    .party-map-zoom-container {
       display: flex;
-      margin: 20px auto 20px auto;
+      flex-direction: column;
+      background-color: rgba(0, 0, 0, 0.5);
+      border-radius: 22px;
+      justify-content: center;
+      position: absolute;
+      bottom: 20px;
+      width: 40px;
```
IvanBuljovcic, line 44: I dont think we should be using any fixed values for dimensions if there is ever an intention to make the app fluid/responsive.

For such a "small" width use something like:

`width: 2.5rem` which would in this case (font-size default: 16px) be 40px

IvanBuljovcic: the best solution for truly fluid apps would be to use `vw` and `vh` units but that requires a lot of reworks.

With the use of `rem` if any person has their default font size changed to be bigger (impaired vision) the map zoom container will be bigger as well

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -8,14 +8,180 @@ $border-radius: 28px;
...
+      cursor: pointer;
+      transition: box-shadow 200ms, width 200ms, height 200ms linear;
+
+      .map-zoom-icon {
+        z-index: 12;
+        cursor: pointer;
+        position: absolute;
+        right: 0;
+        bottom: 0;
+        height: 30px;
```
IvanBuljovcic, line 108: As mentioned above, `rem` in favor of `px`

src/components/templates/PartyMap/components/PartyMapPreview/PartyMapPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <div
+      className={`map-zoom-preview ${expanded && "expand"}`}
+      style={{
+        backgroundImage: `url(${backgroundImage})`,
+        width: previewWidth,
+        height: previewHeight,
+      }}
+    >
+      <div className="map-zoom-icon" onClick={onIconClick}></div>
+      <div className="map-zoom-position"></div>
```
IvanBuljovcic, line 50:
```suggestion
      <div className="map-zoom-position" />
```

IvanBuljovcic, line 50:
```suggestion
      <div className="map-zoom-position" />
```

.vscode/settings.json
```diff
@@ -21,6 +21,6 @@
   "editor.tabSize": 2,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescriptreact]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
thecadams, line 24:
```suggestion
    "editor.defaultFormatter": "esbenp.prettier-vscode"
```

thecadams (commented):
On the way. Many small things here, nothing too major, looking forward to seeing this merged.

denisdimitrov:
Closing this after a discussion on the daily.

PR #823, opened 2020-11-28
Fix room deletion

thecadams:
Whoops.


thecadams:
Glad that we have linting now.

0xdevalias (approved):
Changes look sane, LGTM!

PR #824, opened 2020-11-28
Add ticket code and dob to form

thecadams:
Hack this in for now, we will make these proper venue features when the venue access is implemented.

### DIFF
--- a/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
+++ b/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
@@ -127,7 +127,7 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
             name="date_of_birth"
             className="input-block input-centered"
             type="date"
-            ref={register}
+            ref={register({ required: true })}
           />
           {errors?.date_of_birth?.type === "required" && (
             <span className="input-error">Date of birth is required</span>
--- a/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
+++ b/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
@@ -18,6 +18,8 @@ interface PropsType {
 interface LoginFormData {
   email: string;
   password: string;
+  code: string;
+  date_of_birth: string;
 }
 
 const LoginForm: React.FunctionComponent<PropsType> = ({
@@ -41,7 +43,7 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
 
   const onSubmit = async (data: LoginFormData) => {
     try {
-      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.email);
+      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.code);
       const auth = await signIn(data);
       if (CODE_CHECK_ENABLED && auth.user) {
         firebase
@@ -54,7 +56,8 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
                 .firestore()
                 .doc(`userprivate/${auth.user.uid}`)
                 .update({
-                  codes_used: [...(doc.data()?.codes_used || []), data.email],
+                  codes_used: [...(doc.data()?.codes_used || []), data.code],
+                  date_of_birth: data.date_of_birth,
                 });
             }
           });
@@ -118,6 +121,39 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
           />
           {errors.password && errors.password.type === "required" && (
             <span className="input-error">Password is required</span>
+          )}
+        </div>
+        <div className="input-group">
+          <input
+            name="date_of_birth"
+            className="input-block input-centered"
+            type="date"
+            ref={register}
+          />
+          {errors.date_of_birth && errors.date_of_birth.type === "required" && (
+            <span className="input-error">Date of birth is required</span>
+          )}
+        </div>
+        <div className="input-group">
+          <input
+            name="code"
+            className="input-block input-centered"
+            type="code"
+            placeholder="Ticket Code From Your Email"
+            ref={register({
+              required: true,
+            })}
+          />
+          {errors.code && (
+            <span className="input-error">
+              {errors.code.type === "required" ? (
+                <>
+                  Enter the ticket code from your email. The code is required.
+                </>
+              ) : (
+                errors.code.message
+              )}
+            </span>
           )}
         </div>
         <input
--- a/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
+++ b/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
@@ -130,7 +130,7 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
             type="date"
             ref={register}
           />
-          {errors.date_of_birth && errors.date_of_birth.type === "required" && (
+          {errors?.date_of_birth?.type === "required" && (
             <span className="input-error">Date of birth is required</span>
           )}
         </div>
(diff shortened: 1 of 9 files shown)


src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -41,7 +43,7 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
 
   const onSubmit = async (data: LoginFormData) => {
     try {
-      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.email);
+      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.code);
```
0xdevalias, line 46: [NB, nit] extract `CODE_CHECK_URL + data.code` into a semantically named variable

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -120,6 +123,39 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
...
           )}
         </div>
+        <div className="input-group">
+          <input
+            name="date_of_birth"
+            className="input-block input-centered"
+            type="date"
+            ref={register}
+          />
+          {errors.date_of_birth && errors.date_of_birth.type === "required" && (
```
0xdevalias, line 133: Would this work?

```suggestion
          {errors?.date_of_birth?.type === "required" && (
```

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -120,6 +123,39 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
             <span className="input-error">Password is required</span>
           )}
         </div>
+        <div className="input-group">
+          <input
+            name="date_of_birth"
+            className="input-block input-centered"
+            type="date"
+            ref={register}
```
0xdevalias, line 131: Here you only use `register`, but below you use `register({ require: true })`

Yet in the error message for `date_of_birth`, you're checking for a `required` error.

Is that correct? From reading the code, I would have expected this to also pass through the `required: true`, though i'm not familiar with the library semantics.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -120,6 +123,39 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
...
+            <span className="input-error">
+              {errors.code.type === "required" ? (
+                <>
+                  Enter the ticket code from your email. The code is required.
+                </>
+              ) : (
+                errors.code.message
+              )}
+            </span>
+          )}
```
0xdevalias, line 157: This block seems as though it will show all errors, not just `required`, but the above 1 line check seems like it won't necessarily display all errors? Is the one liner correct?

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -82,11 +85,12 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
 
   const onSubmit = async (data: RegisterFormData) => {
     try {
-      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.email);
+      if (CODE_CHECK_ENABLED) await axios.get(CODE_CHECK_URL + data.code);
```
0xdevalias, line 88: [NB, nit] same as above RE: extracting semantically named variable

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -165,6 +169,39 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
...
+              {errors.code.type === "required" ? (
+                <>
+                  Enter the ticket code from your email. The code is required.
+                </>
+              ) : (
+                errors.code.message
+              )}
+            </span>
+          )}
+        </div>
```
0xdevalias, line 204: This looks like a duplication of the code added in `src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx`? If so, can we extract a reusable 'dumb' component or two?

Eg. something like:

```jsx
<TicketCodeField register={register} error={errors?.date_of_birth} />
<DateOfBirthField register={register} error={errors?.code} />
```

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -82,11 +85,12 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
 
   const onSubmit = async (data: RegisterFormData) => {
```
0xdevalias, line 86: Is this onSubmit function the same between both forms? If so, can we extract it to a common function as well?

src/settings.ts
```diff
@@ -34,7 +34,7 @@ export const CREATE_EDIT_URL = "/admin";
 export const SPARKLEVERSITY_URL = "https://sparklever.se/sparkleversity";
 export const SPARKLEVERSE_COMMUNITY_URL =
   "https://www.facebook.com/groups/sparkleverse/";
-export const TICKET_URL = "https://sparklever.se/glowforth/";
+export const TICKET_URL = "https://www.eventbrite.co.uk/e/129232464781";
```
0xdevalias, line 37: [OOS, future] We really need to get around to making this a param/env var/similar

0xdevalias (commented):
A few comments/potential changes to DRY up the code a little.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
thecadams: Just added required true, it's there on the code field so should be on dob too.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
thecadams: It'll display all the errors because it'll default to the message unless the error is required. This code just makes the required error more informative and passes others through.

src/settings.ts
thecadams: Agreed 👍

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
thecadams: Good catch. There are a few key differences that mean it'd have to be progressively DRYed up:

* `signUp`/`signIn`
* append to `codes_used` (in `LogInForm`) vs set it for the first time (in `RegisterForm`)

OTOH the stuff that's the same includes

* Validation
* Post login steps

Would you be OK to hold off and DRY this up after I've removed the burn-related stuff from it?

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
thecadams: In my recent changes, I've also pushed a change that unifies both form fields to behave the same way at this, since their behaviour appears to be intended to be identical. The DRY-up exposed the similarity.

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
0xdevalias: Happy for it to be a future cleanup thing. I don't think it's worth blocking the other improvements here for it.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -120,6 +122,39 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
             <span className="input-error">Password is required</span>
           )}
         </div>
+        <div className="input-group">
```
0xdevalias, line 125: We can use the new `DateOfBirthField` and `TicketCodeField` components instead of this now can't we?

0xdevalias (approved):
[CRO] One spot of DRY missed I think, but otherwise this looks good to go. Will approve, but if you can clean that last one up as well it'd be 👌

PR #825, opened 2020-11-28
V-com fixes

thecadams:
* Only allow modifying owners via addVenueOwner and removeVenueOwner, since they are edited in a separate form
* Show venue name on schedules
* Make the login and register forms easier to use
* Clarify admin copy for parent ID feature

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -422,12 +422,6 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.parentId = data.parentId;
   }
 
-  let owners = [context.auth.token.user_id];
-  if (data.owners) {
-    owners = [...owners, ...data.owners];
-    updated.owners = owners;
-  }
-
   if (data.columns) {
     updated.columns = data.columns;
   }
--- a/src/components/molecules/LiveSchedule/LiveEvent.tsx
+++ b/src/components/molecules/LiveSchedule/LiveEvent.tsx
@@ -52,7 +52,9 @@ export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
           {event.description}
         </div>
         <div className="schedule-event-info-room">
-          <div onClick={enterLiveEvent}>{event.room ?? "Enter"}</div>
+          <div onClick={enterLiveEvent}>
+            {event.room ?? "Enter"} - {venue.name}
+          </div>
         </div>
       </div>
     </div>
--- a/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
+++ b/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
@@ -82,7 +82,6 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
   };
   return (
     <div className="form-container">
-      <h2>Log in</h2>
       <div className="secondary-action">
         {`Don't have an account yet?`}
         <br />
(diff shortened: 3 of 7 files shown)


src/components/molecules/LiveSchedule/LiveEvent.tsx
```diff
@@ -52,7 +52,9 @@ export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
           {event.description}
         </div>
         <div className="schedule-event-info-room">
-          <div onClick={enterLiveEvent}>{event.room ?? "Enter"}</div>
+          <div onClick={enterLiveEvent}>
+            {event.room ?? "Enter"} - {venue.name}
+          </div>
```
0xdevalias, line 57: [NB, nit, DRY] We could extract this to a common 'dumb' component as it's repeated in `src/components/molecules/EventDisplay/EventDisplay.tsx`

functions/venue.js
```diff
@@ -422,12 +422,6 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.parentId = data.parentId;
   }
 
-  let owners = [context.auth.token.user_id];
-  if (data.owners) {
-    owners = [...owners, ...data.owners];
-    updated.owners = owners;
-  }
-
```
0xdevalias, line 430: Was this just never used here?

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -82,14 +82,14 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
...
   return (
     <div className="form-container">
-      <h2>Log in</h2>
       <div className="secondary-action">
         {`Don't have an account yet?`}
         <br />
         <span className="link" onClick={displayRegisterForm}>
           Register instead!
         </span>
       </div>
```
0xdevalias, line 91: Can we move this below the login form as well, and potentially put it in a `<small></small>` (or equivalent CSS)

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -116,14 +116,14 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
...
   return (
     <div className="form-container">
-      <h2>Create an account!</h2>
       <div className="secondary-action">
         Already have an account?
         <br />
         <span className="link" onClick={displayLoginForm}>
           Login
         </span>
       </div>
```
0xdevalias, line 125: As above, can we move this below the login form as well, and potentially put it in a `<small></small>` (or equivalent CSS)

0xdevalias: [NB, nit, DRY] This is reused between `src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx` and here, we should extract it into a common 'dumb' component

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -623,13 +623,18 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
           )}
         <div className="input-container">
```
0xdevalias, line 624: [OOS, future] We could do some nice stuff around parsing different things here, and ensuring we get the parentVenueId we want out of it, eg. extract it if they paste a full link/etc, check if that venue ID is valid, etc.

0xdevalias (approved):
Approving as it seems all of my previous comments were NB/nit, though some of them (eg. moving text below the form) would be a better solution that improves the intent of this PR.

PR #828, opened 2020-11-28
Fix crash in events

thecadams:
TypeScript did not catch that Record<> can return undefined if the key is not present. This appears to be a deficiency in the Record<> type. More info here: https://fnune.com/typescript/2019/01/30/typescript-series-1-record-is-usually-not-the-best-choice/

This fix wraps the cast in a Partial<> to force handling of the undefined case. Fortunately, no other areas were affeted by this specific bug.

### DIFF
--- a/src/components/molecules/EventDisplay/EventDisplay.tsx
+++ b/src/components/molecules/EventDisplay/EventDisplay.tsx
@@ -17,13 +17,15 @@ import "./EventDisplay.scss";
 
 interface EventDisplayProps {
   event: firebase.firestore.DocumentData;
-  venue: WithId<AnyVenue>;
+  venue?: WithId<AnyVenue>;
 }
 
 export const EventDisplay: FC<EventDisplayProps> = ({ event, venue }) => {
   const { user, profile } = useUser();
 
   const enterEvent = useCallback(() => {
+    if (!venue) return;
+
     const room = venue?.rooms?.find((room) => room.title === event.room);
 
     if (!room) {
@@ -69,7 +71,7 @@ export const EventDisplay: FC<EventDisplayProps> = ({ event, venue }) => {
         </div>
         <div className="schedule-event-info-room">
           <div onClick={enterEvent}>
-            {event.room ?? "Enter"} - {venue.name}
+            {event.room ?? "Enter"} {venue && `- ${venue.name}`}
           </div>
         </div>
       </div>
--- a/src/components/molecules/LiveSchedule/LiveEvent.tsx
+++ b/src/components/molecules/LiveSchedule/LiveEvent.tsx
@@ -53,7 +53,7 @@ export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
         </div>
         <div className="schedule-event-info-room">
           <div onClick={enterLiveEvent}>
-            {event.room ?? "Enter"} - {venue.name}
+            {event.room ?? "Enter"} {venue && `- ${venue.name}`}
           </div>
         </div>
       </div>
--- a/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
+++ b/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
@@ -40,10 +40,10 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
     withEvents: true,
   });
 
-  const relatedVenuesById: Record<
+  const relatedVenuesById: Partial<Record<
     string,
     WithId<AnyVenue>
-  > = relatedVenues.reduce(itemsToObjectByIdReducer, {});
+  >> = relatedVenues.reduce(itemsToObjectByIdReducer, {});
 
   const orderedEvents: DatedEvents = useMemo(() => {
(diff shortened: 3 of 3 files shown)


thecadams:
Quick background on this. Thomas and several others reported that they were getting a black screen - stacktrace from one of the repro's is attached.
WhatsApp Image 2020-11-28 at 11 40 26 PM

thecadams:
@0xdevalias going to break glass as you're asleep, so no need to review before I push, I will test locally before merge, and again in prod after deploy has finished.

thecadams:
No need to review this now but you can review posthumously and I'll happily make a new PR for anything you consider worthwhile!

thecadams:
https://sparkle.space/v/alteredstatesofindependence

thecadams:
Tested in prod and the crash is gone.

0xdevalias:
Frpm your comments + link above, sounds like this is another case where https://github.com/sparkletown/sparkle/pull/744 comes into play.

thecadams:
> From your comments + link above, sounds like this is another case where #744 comes into play.

Yep, the sooner we get that one in the better for the code stability.

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -17,13 +17,15 @@ import "./EventDisplay.scss";
...
+  venue?: WithId<AnyVenue>;
 }
 
 export const EventDisplay: FC<EventDisplayProps> = ({ event, venue }) => {
   const { user, profile } = useUser();
 
   const enterEvent = useCallback(() => {
+    if (!venue) return;
+
     const room = venue?.rooms?.find((room) => room.title === event.room);
```
0xdevalias, line 29: [nit] since we're now explicitly checking `venue` above, we shouldn't need the possibly `undefined` checks here anymore?

src/components/molecules/LiveSchedule/LiveEvent.tsx
```diff
@@ -53,7 +53,7 @@ export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
         </div>
         <div className="schedule-event-info-room">
           <div onClick={enterLiveEvent}>
-            {event.room ?? "Enter"} - {venue.name}
+            {event.room ?? "Enter"} {venue && `- ${venue.name}`}
           </div>
```
0xdevalias, line 57: [nit] We should extract this out into a common component as we're using the same code in `src/components/molecules/EventDisplay/EventDisplay.tsx`

There might even be a more encompassing' chunk we can extract too, I didn't look too closely.

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -40,10 +40,10 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
     withEvents: true,
   });
 
-  const relatedVenuesById: Record<
+  const relatedVenuesById: Partial<Record<
     string,
     WithId<AnyVenue>
-  > = relatedVenues.reduce(itemsToObjectByIdReducer, {});
+  >> = relatedVenues.reduce(itemsToObjectByIdReducer, {});
```
0xdevalias, line 46: Highlighting this as it's the main type related change in this PR.

It's quite a localised fix, but I believe this will be a problem with our root Firestore types as per https://github.com/sparkletown/sparkle/pull/744, so we should solve it there, and make sure to fix all the places it surfaces as potential issues.

0xdevalias (commented):
Changes as they were made for the 'break glass' look fine. A few comments on general cleaning things up + thoughts on a better 'wholistic' fix for this sort of issue in our root Firestore types.

src/components/molecules/EventDisplay/EventDisplay.tsx
thecadams: Agreed; this is the sort of thing a linter can auto fix at scale.

src/components/molecules/LiveSchedule/LiveEvent.tsx
thecadams: Yep definitely opportunities to zoom out on DRYing up. We should DRY at a business-component level (eg. One unified video chat room component that’s reusable everywhere), that’ll guide decisions where to DRY vs rebuild.

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
thecadams: Let’s get #744 finished and reviewed ASAP.

thecadams:
Thanks for the overall deep thoughtful comments @0xdevalias & glad to have the second set of eyes on the fixes.

src/components/molecules/LiveSchedule/LiveEvent.tsx
0xdevalias: _nods_ that definitely makes sense to me.

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
0xdevalias: BIG yes!

PR #830, opened 2020-11-28
Merge back emergency fixes to staging

thecadams:


thecadams:

0xdevalias:
I wonder if there is a better pattern for us keeping `staging` aligned with master? It feels a little weird having to merge back like this all the time.

0xdevalias:
@thecadams were the main emergency changes the commits in:

- https://github.com/sparkletown/sparkle/pull/830/commits/cdb2f1573dd95d18702fa04afebaca6c53e698c0
- https://github.com/sparkletown/sparkle/pull/830/commits/cdb2f1573dd95d18702fa04afebaca6c53e698c0
- https://github.com/sparkletown/sparkle/pull/830/commits/2cf4d883ec1624044c6f89f463764bbf6bc2b553

I'm guessing you pushed these directly on the `master` branch?

If so, I think even in an emergency break-glass situation, we should make them via a PR to `master` rather than pushing directly. Then if you need to break-glass to merge that PR, that's totally fine. That way it will keep all commits referenced back to a PR of some description.

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -75,6 +75,8 @@ const ChatsList: React.FunctionComponent = () => {
 
   const onClickOnSender = useCallback(
     (sender: WithId<User>) => {
+      if (!privateChats) return;
+
       const chatsToUpdate = privateChats.filter(
         (chat) => !chat.isRead && chat.from === sender.id
       );
```
0xdevalias, line 82: Looking at the rest of the logic in this `useCallback`, do we want to bail out early? That would mean `setSelectedUser` never gets called?

Also, for cleanup, we can do the `chatsToUpdate` `filter` -> `map` in a chain without extracting the `const`, and since we aren't doing anything with the return value, it's probably better to use `forEach` or similar instead of `map`

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -88,17 +90,16 @@ const ChatsList: React.FunctionComponent = () => {
...
+        ?.filter(
           (message) =>
             message.deleted !== true &&
             message.type === "private" &&
             (message.to === selectedUser?.id ||
               message.from === selectedUser?.id) &&
             message.ts_utc.seconds > HIDE_BEFORE
         )
-        .sort(chatSort),
+        .sort(chatSort) ?? [],
```
0xdevalias, line 102: Another pattern for this (not sure which is better) would be to do `(privateChats ?? [])` up front, and then the rest doesn't need to handle potentially `undefined`

src/components/molecules/Sidebar/Sidebar.tsx
```diff
@@ -30,15 +32,18 @@ const Sidebar = () => {
   const isEnabled = chatUsers && privateChats;
 
   const chatUserIds = privateChats
-    ?.flatMap((chat) => [chat.from, chat.to])
+    ?.sort(chatSort)
+    .slice(0, NUM_CHAT_UIDS_TO_LOAD)
```
0xdevalias, line 36: This should probably be pushed back to the firestore query we're using to load the private chats, or at the very least within the selector function (so it keeps memo unless that slice we are interested in changes), rather than just filtering them out here on the frontend.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -122,39 +124,8 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
...
-                  Enter the ticket code from your email. The code is required.
-                </>
-              ) : (
-                errors.code.message
-              )}
-            </span>
-          )}
-        </div>
+        <TicketCodeField register={register} error={errors?.date_of_birth} />
+        <DateOfBirthField register={register} error={errors?.code} />
```
0xdevalias, line 128: @thecadams Looks like we're passing the incorrect `error` to each field. Eg. `TicketCodeField` is getting `date_of_birth` errors, and vice versa.

src/components/organisms/PrivateChatModal/PrivateChatModal.tsx
```diff
@@ -59,6 +59,8 @@ const PrivateChatModal: React.FunctionComponent = () => {
     }, {});
 
   const onClickOnSender = (sender: WithId<User>) => {
+    if (!privateChats) return;
+
     const chatsToUpdate = privateChats.filter(
       (chat) => !chat.isRead && chat.from === sender.id
     );
```
0xdevalias, line 66: [DRY] This looks like the same code as used in `src/components/molecules/ChatsList/ChatsList.tsx`. We should extract it to a common place.

0xdevalias (commented):
Comments/future cleanup/etc

src/components/molecules/ChatsList/ChatsList.tsx
thecadams: That sounds better yep. We could catch this with a linter perhaps?

src/components/molecules/ChatsList/ChatsList.tsx
thecadams: Hm, maybe I rushed this fix. I believe privateChats is only not set when data hasn’t been loaded. So it doesn’t need to be functional in that situation.

thecadams: And yes to these other cleanups for sure. Was definitely rushed getting in here.

src/components/molecules/Sidebar/Sidebar.tsx
thecadams: Yes it should. I’ll fix that.

thecadams: Yep this makes sense to push back into the selector, it’ll also make the result of the selector more stable.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
thecadams: Yes these need to be reversed. Will do ASAP.

src/components/organisms/PrivateChatModal/PrivateChatModal.tsx
thecadams: Agreed. Thinking on incremental v rewrite. The sheer quantity of these types of nits makes me lean in favour of a rebuild of key components in compliance with our new, higher standards.

0xdevalias: To me, that still fits within what I would call an 'incremental refactor' rather than a 'rewrite' (which to me implies starting from scratch entirely). May just be a difference in semantics though.

Similar to how I 'ripped apart' the theme camp template as part of my performance refactors, that is what I think we should be doing to the other core components we need up to scratch.

0xdevalias: If you could create a shared list of the components you think fall under this 'need rewrites' (on trello/similar?) then that would help with prioritising + assessing effort required.

src/components/molecules/ChatsList/ChatsList.tsx
0xdevalias: It'd certainly be possible, but would take a little time to figure the semantics of it properly (assuming a rule doesn't already exist)

If you create a backlog ticket for it I can make this happen when we have some breathing room.

src/components/molecules/Sidebar/Sidebar.tsx
0xdevalias: Agreed

PR #832, opened 2020-11-28
Deploy

thecadams:


thecadams:

0xdevalias (commented):
Assuming e2219e3bf9c5259903b51b103d8aa0af8c2ff09e is the only actual new change here, in which case, seems fine.

PR #821, opened 2020-11-27
Accept image/png image types

margulies:
The image type 'image/x-png' did not appear to accept uploads of some .png image files. The addition of 'image/png' as an option appeared to fix that issue.

(New to contributing to this repo, so starting small...)

PR #835, opened 2020-11-30
Allow uploading png images

thecadams:
See https://stackoverflow.com/a/2086441/1013781 for more info on the issue


0xdevalias:
@thecadams this is a dupe of #821

PR #836, opened 2020-11-30
Fix TypeError: Cannot assign to read only property '0' of object '[object Array]' in SideBar component

0xdevalias:
Fixes `TypeError: Cannot assign to read only property '0' of object '[object Array]'` in the `SideBar` component due to trying to `.sort()` the array before the `.slice()` (which appears to have been introduced in https://github.com/sparkletown/sparkle/pull/830)

Ref: https://stackoverflow.com/questions/53420055/error-while-sorting-array-of-objects-cannot-assign-to-read-only-property-2-of/53420326#53420326


denisdimitrov (approved):
LGTM!

thecadams (approved):

PR #845, opened 2020-11-30
fixes and tweaks for the party map venue and private chats

denisdimitrov:
- show search bar for user when private chats are empty
- create or fill enteredVenueIds on venue page open
- filter venue chats by venue chat users
- filter private chat by private chat users


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -200,6 +201,20 @@ const VenuePage = () => {
     };
   }, [prevLocations, user, venueName]);
 
+  useEffect(() => {
+    if (
```
thecadams, line 205: nit: these returns can be combined

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
```diff
@@ -42,14 +42,6 @@ export const VenueEntrancePage: React.FunctionComponent<{}> = () => {
     !venue.entrance.length ||
     venue.entrance.length < parseInt(step)
   ) {
-    const enteredVenueIds = profile.enteredVenueIds ?? [];
```
thecadams, line 45: is this being removed because we've added it to `VenuePage`?

src/components/molecules/VenueChat/VenueChat.tsx
```diff
@@ -27,6 +32,16 @@ interface ChatOutDataType {
 
 const VenueChat: FC = () => {
   const venueId = useVenueId();
+  const usersById = useSelector(venueChatUsersSelector) ?? {};
+  const venueChatUsersQuery = useMemo<ReduxFirestoreQuerySetting>(
```
thecadams, line 36: Should this be `SparkleRFQConfig` instead of `ReduxFirestoreQuerySetting`?

src/components/molecules/ChatList/ChatList.tsx
```diff
@@ -24,29 +16,20 @@ import "./ChatList.scss";
...
 }
 
 const ChatList: React.FC<ChatListProps> = ({
+  usersById,
   messages,
   allowDelete,
   emptyListMessage,
   deleteMessage,
 }) => {
-  const usersById = useSelector(venueChatUsersSelector) ?? {};
```
thecadams, line 39: just wanting to check - did this move to the parent as a DRY up?

thecadams (commented):
Looks good, couple of nits and a question.

src/pages/VenuePage/VenuePage.tsx
0xdevalias: It also looks like it could be simplified to `profile?.enteredVenueIds?.includes(venue?.id)`

0xdevalias: Also, nit, a newline of space after a 'guard' statement/pattern like we're using here makes things much clearer/easier to read.

0xdevalias: It looks like we also want to check `!profile` in the guard too. Would simplify the latter statement.

src/components/molecules/VenueChat/VenueChat.tsx
0xdevalias: The `useMemo` probably isn't needed here as `useFirestoreConnect` does a deep compare on the query it's passed anyway.

src/components/molecules/ChatList/ChatList.tsx
0xdevalias: - `src/components/molecules/Chatbox/Chatbox.tsx` receives `usersById` as a prop, and passes it to this component
- `ChatBox` is embedded by
  - `src/components/molecules/ChatsList/ChatsList.tsx`
  - `src/components/molecules/VenueChat/VenueChat.tsx`
  - `src/components/organisms/PrivateChatModal/PrivateChatModal.tsx`
  - `src/components/organisms/UserProfileModal/UserProfileModal.tsx`
  - `src/pages/FriendShipPage/FriendShipPage.tsx`
- I can see changes in this PR to `VenueChat`/`ChatsList` to wire this up, but I don't think I see changes for `PrivateChatModal`, `UserProfileModal`, or `FriendShipPage`
  - Do these extra components need to be wired up as well?

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -138,17 +138,21 @@ const ChatsList: React.FunctionComponent = () => {
...
   return (
     <Fragment>
+      <UserSearchBar onSelect={setSelectedUser} />
       {hasPrivateChats && (
         <div className="private-container show">
           <div className="private-messages-list">
-            <UserSearchBar onSelect={setSelectedUser} />
             {discussions.map((userId: string) => {
               const sender = { ...chatUsers![userId], id: userId };
               const lastMessageExchanged =
```
0xdevalias, line 158: [OOS, future] This probably should be extracted + `useMemo`, or reworked as a sub component (or both)

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
0xdevalias: Looks like it. And by being there, it should mean that it counts (if unique) no matter what venue URL we end up on.

src/utils/profile.ts
```diff
@@ -0,0 +1,17 @@
+import firebase from "firebase/app";
```
0xdevalias, line 1: Do we want to create a new 'prefix' for firestore stuff? eg. `api/` (as opposed to `utils/`)? Would help to keep things cleaner/isolated/easier to find all of the firebase calls/etc. And then can easily access them from thunks/etc as required in our redux/etc layer.

src/utils/profile.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const updateProfileEnteredVenueIds = async (
+  prevEnteredVenueIds: string[] | undefined,
```
0xdevalias, line 4:
```suggestion
  prevEnteredVenueIds: readonly string[] | undefined,
```

Which I expect will break your `enteredVenueIds.push(venueId)` line. We want non-mutating functions like `array.concat` instead

- https://lorenstewart.me/2017/01/22/javascript-array-methods-mutating-vs-non-mutating/#iiaddnonmutating

0xdevalias (changes requested):
A few nit/cleanup comments that may not matter, but a few things that might be an issue.

src/components/molecules/ChatList/ChatList.tsx
denisdimitrov: I need to pass the users from the parent component because the chatlist is using a different set of users in every component it's being used. In venue chat I need the users from the venue chat, in private chat I need the users from the private chats and etc...

thecadams: No prob, that’s what I suspected! Thanks!

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
denisdimitrov: Yup, the venue page is now responsible for this logic.

src/utils/profile.ts
thecadams: Yes- eventually. Another opportunity along the same lines is pulling `firestore` references out of our selectors across the codebase. A couple key things keep us moving in this direction, which we can keep eyes on in future reviews and make sure we progressively hew to:

* Ensure the selectors we encounter come from `selectors.ts` or get converted to use those selectors as part of the PR, so we can one day make them point somewhere other than `store.firestore`
* Move firestore queries progressively into hooks (combining `useFirestoreConnect` and `useSelector` into one call ala `const venue = useCurrentVenue()`, and then into the store (using thunks if needed)

These types of directions we set offer a low-cost, incremental, PR and code-review oriented way to push Firebase down into its own layer. Any other ideas are welcome!

src/components/molecules/ChatList/ChatList.tsx
0xdevalias: from https://github.com/sparkletown/sparkle/pull/845#discussion_r533060123

> Do these extra components need to be wired up as well?

src/utils/profile.ts
0xdevalias: We can probably eslint warn/ban imports of the 'direct' firebase import references, + the firebase hooks, etc outside of our chosen 'allowed locations' to aid in this.

0xdevalias: ^tracked as https://github.com/sparkletown/internal-sparkle-issues/issues/2

thecadams: Yes; banning hooks/* and firebase imports in molecules will be a good start.

0xdevalias: Not sure banning `hooks/*` in molecules is correct.. is it? That would mean none of our custom hooks?

thecadams: The underlying goal is to prevent store access in the “dumb” components so they can receive data only via props and are closer to being controlled components. So maybe we would separate out the hooks which access the store from the rest, or achieve that goal a different way. Definitely a discussion for the future.

0xdevalias: Ah, makes sense. I need to find time to actually read that intro post/primer on the whole organism/etc pattern we're using at some point.... :p

thecadams: > I need to find time to actually read that intro post/primer on the whole organism/etc pattern we're using at some point.... 😝

So do I! 😂 I inferred from context in the end. We inherited it from an agency a while back and I suspect it’s meant to encourage the controlled/uncontrolled component divide.

0xdevalias: Haha, that makes sense. I have it bookmarked.. 😛

denisdimitrov:
The comments have been fixed, you can take a second look if you want because this is required for the event today.

0xdevalias:
> @denisdimitrov The comments have been fixed, you can take a second look if you want because this is required for the event today.

For future, if you click the 're-request review' button after changes have been made it should properly notify me, otherwise I won't know to look again.

0xdevalias (approved):
LGTM!

PR #846, opened 2020-11-30
party map layout fixes and tweaks

denisdimitrov:
- increase navbar zindex to show it above schedule
- add new chatbox layout to private message modal
- add padding to number of partygoers online
- render room count conditionally to avoid rendering blue box
- open external internal links differently

This PR is pending on #845 and is going to have breaking changes + conflicts

### DIFF
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -390,7 +390,7 @@ $border-radius: 28px;
   position: fixed;
   left: 0;
   top: 0;
-  z-index: 100;
+  z-index: 299;
   width: 100%;
   height: 100%;
   background-color: rgba($black, 0.8);
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -282,7 +282,7 @@ $border-radius: 28px;
 
 .schedule-dropdown-body {
   position: fixed;
-  z-index: 4;
+  z-index: z(schedule);
   top: 0;
   left: 0;
   width: 100%;
@@ -390,7 +390,7 @@ $border-radius: 28px;
   position: fixed;
   left: 0;
   top: 0;
-  z-index: 3;
+  z-index: z(schedule-backdrop);
   width: 100%;
   height: 100%;
   background-color: rgba($black, 0.8);
@@ -409,7 +409,7 @@ $border-radius: 28px;
 .back-map-btn {
   position: fixed;
   top: 60px;
-  z-index: 3;
+  z-index: z(map-back-button);
   left: 0px;
   height: 52px;
   width: auto;
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -390,7 +390,7 @@ $border-radius: 28px;
   position: fixed;
   left: 0;
   top: 0;
-  z-index: 299;
+  z-index: 100;
   width: 100%;
   height: 100%;
   background-color: rgba($black, 0.8);
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -282,7 +282,7 @@ $border-radius: 28px;
 
 .schedule-dropdown-body {
   position: fixed;
-  z-index: 300;
+  z-index: 4;
   top: 0;
   left: 0;
   width: 100%;
@@ -390,7 +390,7 @@ $border-radius: 28px;
   position: fixed;
   left: 0;
   top: 0;
-  z-index: 299;
+  z-index: 3;
   width: 100%;
   height: 100%;
   background-color: rgba($black, 0.8);
@@ -409,7 +409,7 @@ $border-radius: 28px;
 .back-map-btn {
   position: fixed;
   top: 60px;
-  z-index: 200;
+  z-index: 3;
   left: 0px;
   height: 52px;
   width: auto;
--- a/src/components/molecules/NavBar/playa.scss
+++ b/src/components/molecules/NavBar/playa.scss
@@ -16,7 +16,7 @@ $sand: #937c63;
 
 .navbar.navbar_playa {
   position: fixed;
-  z-index: 100;
+  z-index: 301;
   top: 0;
   left: 0;
   width: 100%;
--- a/src/components/molecules/NavBar/playa.scss
+++ b/src/components/molecules/NavBar/playa.scss
@@ -1,3 +1,5 @@
+@import "scss/constants.scss";
+
 $side-padding: 30px;
 $primary: #005ee5;
 $black: #000000;
@@ -14,11 +16,9 @@ $page-max-width: 1240px;
 
 $sand: #937c63;
 
-$header-z-index: 5;
-
 .navbar.navbar_playa {
   position: fixed;
-  z-index: $header-z-index;
+  z-index: z(header);
   top: 0;
   left: 0;
   width: 100%;
--- a/src/components/molecules/NavBar/playa.scss
+++ b/src/components/molecules/NavBar/playa.scss
@@ -14,9 +14,11 @@ $page-max-width: 1240px;
 
 $sand: #937c63;
 
+$header-z-index: 5;
+
 .navbar.navbar_playa {
   position: fixed;
-  z-index: 301;
+  z-index: $header-z-index;
   top: 0;
   left: 0;
   width: 100%;
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -17,7 +17,7 @@ import { NavSearchBarInput } from "./NavSearchBarInput";
 
 interface SearchResult {
   rooms: CampRoomData[];
-  users: WithId<User>[];
+  users: readonly WithId<User>[];
   events: VenueEvent[];
 }
 
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -5,6 +5,7 @@ import React, { useCallback, useEffect, useState } from "react";
 import { CampRoomData } from "types/CampRoomData";
 import { User } from "types/User";
 import { VenueEvent } from "types/VenueEvent";
+import { WithId } from "utils/id";
 import {
   currentVenueSelectorData,
   partygoersSelector,
@@ -16,7 +17,7 @@ import { NavSearchBarInput } from "./NavSearchBarInput";
 
 interface SearchResult {
   rooms: CampRoomData[];
-  users: User[];
+  users: WithId<User>[];
   events: VenueEvent[];
 }
 
@@ -27,7 +28,9 @@ const NavSearchBar = () => {
     users: [],
     events: [],
   });
-  const [selectedUserProfile, setSelectedUserProfile] = useState<User>();
+  const [selectedUserProfile, setSelectedUserProfile] = useState<
+    WithId<User>
+  >();
   const [selectedRoom, setSelectedRoom] = useState<CampRoomData>();
 
   const venue = useSelector(currentVenueSelectorData);
--- a/src/components/molecules/Sidebar/Sidebar.scss
+++ b/src/components/molecules/Sidebar/Sidebar.scss
@@ -6,7 +6,7 @@ $dark: #19181a;
 
 .sidebar-container {
   position: fixed;
-  z-index: 2;
+  z-index: z(sidebar);
   right: 0;
   top: 60px;
   width: 360px;
--- a/src/components/molecules/Sidebar/Sidebar.scss
+++ b/src/components/molecules/Sidebar/Sidebar.scss
@@ -6,7 +6,7 @@ $dark: #19181a;
 
 .sidebar-container {
   position: fixed;
-  z-index: 90;
+  z-index: 2;
   right: 0;
   top: 60px;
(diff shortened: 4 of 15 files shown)


src/components/molecules/NavBar/playa.scss
```diff
@@ -16,7 +16,7 @@ $sand: #937c63;
 
 .navbar.navbar_playa {
   position: fixed;
-  z-index: 100;
+  z-index: 301;
```
thecadams, line 19: Any reason for `301` specifically... let's move this zindex to a variable, and aim to move more later so we have them all in one place.

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -279,6 +279,8 @@ $border-radius: 28px;
     transition: all 400ms cubic-bezier(0.23, 1, 0.32, 1);
     opacity: 0;
     transform: translateY(-20px) scale(0.8);
+    min-width: 100px;
+    max-width: 120px;
```
thecadams, line 283: just wondering - this seems suspect; was it tested at many widths?

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -332,6 +334,8 @@ $border-radius: 28px;
     transition: height 400ms cubic-bezier(0.23, 1, 0.32, 1);
     display: block;
     transition: transform 600ms cubic-bezier(0.23, 1, 0.32, 1);
+    min-width: 150px;
+    max-width: 280px;
```
thecadams, line 338: same here - wondering if this has been tested at many widths

src/utils/url.ts
```diff
@@ -39,7 +39,11 @@ export const openRoomUrl = (url: string) => {
 };
 
 export const openUrl = (url: string) => {
-  window.open(url, "_blank", "noopener,noreferrer");
+  if (!isExternalUrl(url)) {
+    window.open(url, "_blank", "noopener,noreferrer");
```
thecadams, line 43: nit: let's use the utility function for this

thecadams (commented):
Looks good - only a few minor questions.

src/components/molecules/VenuePartygoers/VenuePartygoers.scss
```diff
@@ -0,0 +1,3 @@
+.attendees-container {
```
0xdevalias, line 1: Should this be scoped under something `VenuePartygoers` related?

Generally in the past I've used patterns where the top 'container' uses a 'wrapper' class (eg. `.venue-partygoers`), and then all other more specific rules live under this. That way they will only be applied when they appear within the 'wrapping classes scope'.

That way we can be more confident that our CSS styling doesn't weirdly change/break other things, without having to result to a CSS-in-JS pattern (which in many cases feels heavy/overkill IMO)

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -61,10 +67,29 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
...
 
+  const chatContext = useContext(ChatContext);
+
+  const submitMessage = useCallback(
+    async (data: { messageToTheBand: string }) => {
+      chatContext &&
+        user &&
+        chatContext.sendPrivateChat(
+          user.uid,
+          fullUserProfile!.id,
```
0xdevalias, line 78: Do we want to be using `!` here?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -61,10 +67,29 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
     return undefined;
   }, [rest]);
 
+  const chatContext = useContext(ChatContext);
+
+  const submitMessage = useCallback(
+    async (data: { messageToTheBand: string }) => {
+      chatContext &&
+        user &&
```
0xdevalias, line 75:
```suggestion
      if (!chatContext || !user) return
      
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -61,10 +67,29 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
...
+    [chatContext, fullUserProfile, user]
+  );
+
   if (!fullUserProfile || !fullUserProfile.id || !user) {
     return <></>;
   }
 
+  const chats: WithId<PrivateChatMessage>[] = privateChats.filter(
+    (chat) => chat.from === fullUserProfile.id || chat.to === fullUserProfile.id
+  );
```
0xdevalias, line 91: This may be worth `useMemo`ing

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -61,10 +67,29 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
     return undefined;
   }, [rest]);
 
+  const chatContext = useContext(ChatContext);
+
+  const submitMessage = useCallback(
+    async (data: { messageToTheBand: string }) => {
```
0xdevalias, line 73: I realise this is what `ChatBox` currently expects, but it feels somewhat needlessly complex/an implementation detail. I wonder if we can refactor this to just take a `msg: string` instead?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
0xdevalias: Also, what does `fullUserProfile` refer to here?

`chatContext.sendPrivateChat` seems to take `(from, to, text)` as params, implying we are sending a message to `fullUserProfile.id` here?

I'm guessing that since this is passed down from props, the implication is that `fullUserProfile` is the user that this `UserProfileModal` is for/showing. And `user` is the current person that opened the modal.

The ambiguous use of `user` in this component makes it hard to follow. I think renaming `user` to something implying "the current person" (`loggedInUser`, `authenticatedUser`, `myUser`?), and/or renaming `fullUserProfile` to something implying "the person this modal is about" would make this a lot easier to follow.

src/utils/url.ts
0xdevalias: @thecadams this is in `src/utils/url.ts`, which particular utility function were you referring to? Or didn't realise this was already in `utils/url`?

0xdevalias (changes requested):
(reviewing the comments I made above, I think there are at least a couple that warrant changes)

src/utils/url.ts
thecadams: Pretty sure this is in `openUrl`.

denisdimitrov: This is the `openUrl` utility function, and I'm adding the check to it.

src/components/molecules/VenuePartygoers/VenuePartygoers.scss
thecadams: We will benefit tremendously from a documented convention for wrapper classes around components. Something like:

* A component must render a top-level element with a `className` of `COMPONENTNAME-wrapper`
* The .scss must be entirely wrapped in
```
.COMPONENTNAME-wrapper {
...
}
```

0xdevalias: Agreed. Potentially something we can lint (`stylelint`?), if not with an existing rule, with a custom. But until then, documented/enforced convention sounds good to me.

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -282,7 +282,7 @@ $border-radius: 28px;
 
 .schedule-dropdown-body {
   position: fixed;
-  z-index: 300;
+  z-index: 4;
```
0xdevalias, line 285: Use variables for these as well?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -16,7 +17,7 @@ import { NavSearchBarInput } from "./NavSearchBarInput";
 
 interface SearchResult {
   rooms: CampRoomData[];
-  users: User[];
+  users: WithId<User>[];
```
0xdevalias, line 20:
```suggestion
  users: readonly WithId<User>[];
```

src/components/molecules/Sidebar/Sidebar.scss
```diff
@@ -6,7 +6,7 @@ $dark: #19181a;
 
 .sidebar-container {
   position: fixed;
-  z-index: 90;
+  z-index: 2;
```
0xdevalias, line 9: Use variables for these as well?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -74,24 +84,23 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
             <div className="profile-basics">
               <div className="profile-pic">
                 <img
-                  src={fullUserProfile.pictureUrl || "/default-profile-pic.png"}
+                  src={userProfile.pictureUrl || "/default-profile-pic.png"}
```
0xdevalias, line 87: Maybe?

```suggestion
                  src={userProfile.pictureUrl ?? "/default-profile-pic.png"}
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -74,24 +84,23 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
...
+                          userProfile.id.charCodeAt(0) % RANDOM_AVATARS.length
                         )
                       ];
                   }}
                 />
               </div>
               <div className="profile-text">
                 <h2 className="italic">
-                  {fullUserProfile.partyName || "Captain Party"}
+                  {userProfile.partyName || "Captain Party"}
```
0xdevalias, line 103: Maybe?

```suggestion
                  {userProfile.partyName ?? "Captain Party"}
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -103,7 +112,7 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
                     {/*
                     // @debt typing - need to support known User interface with unknown question keys
                     // @ts-ignore */}
-                    {fullUserProfile[question.name] || //@debt typing - look at the changelog, was this a bug?
+                    {userProfile[question.name] || //@debt typing - look at the changelog, was this a bug?
```
0xdevalias, line 115: Maybe?

```suggestion
                    {userProfile[question.name] ?? //@debt typing - look at the changelog, was this a bug?
```

Is this @debt still an issue?

0xdevalias (commented):
Comments based on changes since my last review

denisdimitrov:
[Denis Ed tweaks](https://trello.com/c/SqX7QyaO/434-denis-ed-tweaks)

src/components/organisms/UserProfileModal/UserProfileModal.tsx
denisdimitrov: Yes, it is.

denisdimitrov:
@thecadams @0xdevalias You can take a second look at the changes. They've been updated.

src/components/organisms/UserProfileModal/UserProfileModal.scss
```diff
@@ -35,8 +35,9 @@ $modal-max-width: 540px;
   }
 
   .private-chat-container {
-    height: 300px;
-    overflow-wrap: break-word;
```
thecadams, line 39: Can we re-add this one? It makes sure we wrap by split-
ting words.

src/utils/url.ts
thecadams: OK. Just thinking we should pull everything into `openUrl` including the `isExternalUrl` check. If opening URLs is in one place we can easily change the systemwide policy for whether to open in new window.

thecadams (commented):
Couple new questions, looking good so far.

src/components/molecules/VenuePartygoers/VenuePartygoers.scss
0xdevalias: @thecadams RE: the above, I have some good patterns I figured out with how we can use CSS modules (enabled by default in `create-react-app` + some other improvements to our styling. Started documenting my context in https://github.com/sparkletown/internal-sparkle-issues/issues/4 the other day to avoid the rabbithole, but the main relevant part to this would be:

- https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

You can use generic names like `.error` and similar in the component's css module, then import it like `import styles from './Button.module.css'`, which then allows you to use it like `<button className={styles.error}>Error Button</button>;`

It generates locally scoped unique classnames, so no global clashes, and isn't as heavy as a lot of CSS-in-JS patterns/libs.

src/scss/constants.scss
```diff
@@ -40,3 +40,22 @@ $playa-avatar-size: 48px;
...
+);
+
+@function z($layer) {
+  @if not map-has-key($z-layers, $layer) {
+    @warn "No layer found for `#{$layer}` in $z-layers map. Property omitted. Falling back to 1.";
+    @return 1;
+  }
+
+  @return map-get($z-layers, $layer);
+}
```
0xdevalias, line 61: 🎉

0xdevalias (approved):
LGTM! Nice addition of the `z-profile` function!

src/components/organisms/UserProfileModal/UserProfileModal.scss
denisdimitrov: I just added it back.

PR #847, opened 2020-12-01
Integrate Bugsnag for error tracking + show build reference outputs in index.html

0xdevalias:
- Display developer enticing <3 + build outputs in `index.html`
  - Use `/**!` syntax to tell `terser-webpack-plugin` not to strip our comments + `prettier-ignore` syntax to maintain it's form
    - https://webpack.js.org/plugins/terser-webpack-plugin/#extractcomments
    - https://prettier.io/docs/en/ignore.html#html
  - Ascii art from http://patorjk.com/software/taag/#p=display&f=Graffiti&t=%20%20%20Sparkle%0A
  - A couple of `<meta>` tag things from https://github.com/joshbuchea/HEAD
  - Makes use of CI env vars: https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

## Bugsnag

- https://docs.bugsnag.com/platforms/javascript/react/
  - Basic configuration for React is done as per this reference
  - https://docs.bugsnag.com/platforms/javascript/react/#capturing-react-render-errors
    - We are using the error boundary, but we aren't currently showing any fallback view. We could improve on this in future.
  - https://docs.bugsnag.com/platforms/javascript/react/#reporting-handled-errors
    - We can manually report handled errors with `Bugsnag.notify(err)`
    - More info at https://docs.bugsnag.com/platforms/javascript/react/reporting-handled-errors/
  - https://docs.bugsnag.com/platforms/javascript/react/#sending-diagnostic-data
    - It's possibly to include additional diagnostic data/context with our errors if we have it
    - More info at https://docs.bugsnag.com/platforms/javascript/react/customizing-error-reports/
  - https://docs.bugsnag.com/platforms/javascript/react/#identifying-users
    - We can set a user to be associated with an error, to help narrow things down
    - More info at https://docs.bugsnag.com/platforms/javascript/react/customizing-error-reports/#adding-user-data
  - https://docs.bugsnag.com/platforms/javascript/react/#logging-breadcrumbs
    - We can log additional custom 'breadcrumbs' with `Bugsnag.leaveBreadcrumb` to give more context as to the 'path' that lead to the error
    - More info at https://docs.bugsnag.com/platforms/javascript/react/customizing-breadcrumbs/
  - https://docs.bugsnag.com/platforms/javascript/react/#tracking-releases
    - For our 'release version' I will be using the git SHA1 hash of the deployed release, so it should be easy to trace back to the code
  - https://docs.bugsnag.com/platforms/javascript/react/configuration-options/
    - Additional config settings
- https://docs.bugsnag.com/platforms/javascript/react/customizing-error-reports/#discarding-events
  - It's possible to choose to discard events (eg. for user opt-out)
- https://docs.bugsnag.com/platforms/javascript/source-maps/#hosting-source-maps
  - Our source maps are already accessible (and makes no sense to hide them since we're open source), so we don't need to do anything special here

**Note:** for our project setup we use the same Project (and thus Bugsnag api key) for all of staging/production/sparkleverse/etc. We distinguish between these environments through the `releaseStage` Bugsnag setting.

I have updated our CircleCI environment variables to include the Bugsnag API key already.

WIP: TBC


0xdevalias (commented):
Best to review with 'hide whitespace changes' to minimize diff you need to eyeball

0xdevalias:
[Set up Bugnsag (or similar) frontend error tracking/reporting service so that we have some visibility into the issues people are seeing in production](https://trello.com/c/m00FfuQJ/444-set-up-bugnsag-or-similar-frontend-error-tracking-reporting-service-so-that-we-have-some-visibility-into-the-issues-people-are-s)

src/index.tsx
```diff
@@ -86,6 +96,80 @@ const rrfProps = {
...
+      event.setUser(
+        currentUser.uid,
+        currentUser.email || undefined,
+        currentUser.displayName || undefined
+      );
+    },
+  });
+}
+
+const BugsnagErrorBoundary = Bugsnag?.getPlugin("react")?.createErrorBoundary(
```
thecadams, line 165: Can we nest error boundaries?

thecadams: Just thinking we'll need smaller boundaries if we go with your plan to have boundaries throughout the app.

public/index.html
```diff
@@ -1,49 +1,86 @@
...
+    <title>Sparkle</title>
+    <script src="https://www.eventbrite.co.uk/static/widgets/eb_widgets.js"></script>
+  </head>
 
+  <body
+    data-sparkle-build-sha1="%REACT_APP_BUILD_SHA1%"
+    data-sparkle-build-tag="%REACT_APP_BUILD_TAG%"
+    data-sparkle-build-branch="%REACT_APP_BUILD_BRANCH%"
+    data-sparkle-build-pull-requests="%REACT_APP_BUILD_PULL_REQUESTS%"
+    data-sparkle-build-repository="%REACT_APP_BUILD_REPOSITORY_URL%"
```
thecadams, line 81: This seems like a little too much info, no need to include it in our page just because it's there, I'd advocate a more minimal approach. What about we only give the sha and repo URL. The rest can be derived from them anyway.

thecadams (approved):

src/index.tsx
0xdevalias: My understanding is that we can have as many as we need within our component tree. They act kind of like a `try { } catch { }` around a tree of components, and stop the error bubbling any higher (eg. to break the entire app)

- https://reactjs.org/docs/error-boundaries.html

If we introduce more at a deeper level (eg. if we wanted to prevent issues in a certain component/feature from being able to bring down the entire app) we would need to look closer at what additional we need to do in them to ensure they still get reported to bugsnag. It may be as simple as using this same `Bugsnag?.getPlugin("react")?.createErrorBoundary` pattern to create the 'deeper' error boundary. Alternatively if it was just a standard boundary, we can manually pass errors through to Bugsnag from it with `Bugsnag.notify` (which we can also use anywhere we want to explicitly pass an already 'handled' error through, eg. instead of `console.error` logging a failed ajax request/similar.

thecadams: Sounds good.

public/index.html
0xdevalias: These were explicit choices narrowed down from what was available. I don't feel strongly about keeping it in the `data-` keys (as it is already in `meta`), but unless there is a strong reason not to, I feel pretty strongly about keeping the information here in the index in general.

- It 'costs' us basically nothing, a few extra characters in the index, trivial compared to the size of the rest of the app's assets.
- We already use this data to pass extra context on Bugsnag errors, so it needs to exist within the app regardless of whether it is exposed here.
- This is part of my plan to 'entice curious developers to look deeper/explore'. It's easy for us to infer/trace back based on a minimal set of information. I want to lower that barrier of required knowledge to do so. If the information is simple and easy to follow, there's a higher chance of follow-through based on it.

Open to hearing your thoughts/reasons why we shouldn't.

thecadams: Couple of thoughts / responses, but I'm not going to push either way as it's a minor change and you're the one leading it so you should make the call.

* Agreed on zero cost. With caching, compression etc. it's even less.

* Re Bugsnag. My thinking was about reducing redundancy in what we surface to humans; I'm a minimalist - and the rest can be derived. It depends what impression we want to give; I feel as though showing all the extra info triggers OCD that perhaps "these people haven't whittled down the things they include to the bare minimum". I like the saying, attributed to Antoine de Saint-Exupéry:

> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.

* Re enticing developers - yes this info is required to initialize Bugsnag but it's not required (in my opinion) to get humans interested in digging further. All it takes to send me spelunking is a sha and a url.

0xdevalias: I hear you on reducing redundancy in general, I also like a clean/minimalist approach to things, and for me, a repo URL/SHA (or even less ;) ) is all it would take me to track down a repo and get involved. Though my thinking with adding this little bit extra here is less for the hardcore/elite, and more to give an extra helping hand to someone who might be interested in a curiously dabbling/exploring the project, and reducing that barrier of entry/discoverability.

My original reason for adding the `data-` tags as well was to increase potential discoverability (as the `index.html` gets quite aggressively minimized during our build process), but reflecting, since I worked out how to keep an uncompressed comment, + it existing in `meta`, I don't think it adds anything extra we don't already have. So I'll delete these `data-` keys, and then will probably merge this PR.

If we find it too overbearing in future, we can always address things in a followup PR then. (eg. right now i'm not familiar enough to know how much extra context passed to bugsnag is useful, I expect we may be able to pare that down a bit once we get some data/real-world usage experience with it, and which bits we use vs which we just kind of ignore/feel redundant.)

thecadams: Just thinking some more. Removing it entirely might not entice people To be clear, here's what I was advocating we do:

```
    data-sparkle-build-sha1="%REACT_APP_BUILD_SHA1%"
    data-sparkle-build-repository="%REACT_APP_BUILD_REPOSITORY_URL%"
```

That'd get me going "ooo I wonder what's in that sha!" - thoughts on doing that?

0xdevalias: The `data-` keys get pretty lost in the minified html anyway, so they're not super discoverable anyway.

Easier to describe visually:
Screen Shot 2020-12-02 at 11 18 14 am

I'm keeping 1 and 2, removed 3 in https://github.com/sparkletown/sparkle/pull/847/commits/abf7d7e9f7e5407431cee02f0f1bb7c0ac1ddfa3

thecadams: Thanks for the visual. Looks good.

0xdevalias:
@thecadams I've got the `/version` page added. Not super happy with it's styling right now, but ran into a rabbithole of potential issues/lack of my knowledge about our patterns while trying to make it nicer, so decided to sideline that as a future cleanup to get this merged.

On the uphand, feels kind of terminal'y:
image

thecadams:
> @thecadams I've got the `/version` page added. Not super happy with it's styling right now, but ran into a rabbithole of potential issues/lack of my knowledge about our patterns while trying to make it nicer, so decided to sideline that as a future cleanup to get this merged.

Yep! Sound judgment to merge rather than rabbit hole in this iteration. As they say, make it work, then make it fast, then make it elegant.

PR #848, opened 2020-12-01
show indicator for unread messages

denisdimitrov:
- shows an indicator when you have unread messages
- shows how long ago the message was sent


src/components/molecules/ChatsList/ChatsList.scss
```diff
@@ -34,15 +34,40 @@ $primary: #005ee5;
         width: 100%;
```
0xdevalias, line 34: note: I'm not deeply code reviewing the styles here, i'm mostly assuming that if they work in testing/clickthrough they should be fine.

src/components/molecules/ChatsList/ChatsList.scss
```diff
@@ -102,3 +127,7 @@ $primary: #005ee5;
   align-items: center;
   justify-content: center;
 }
+
+.not-read-indicator {
+  background-color: blue;
+}
```
0xdevalias, line 133: 🎉

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -153,10 +154,10 @@ const ChatsList: React.FunctionComponent = () => {
               const sender = { ...chatUsers![userId], id: userId };
               const lastMessageExchanged =
                 discussionPartnerWithLastMessageExchanged?.[userId];
+              const isUnreadMessage = !lastMessageExchanged.isRead;
               const profileName = sender.anonMode
                 ? DEFAULT_PARTY_NAME
                 : sender.partyName;
-
```
0xdevalias, line 159: Did `prettier` delete this? I would have expected it to keep a line here.

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -170,15 +171,32 @@ const ChatsList: React.FunctionComponent = () => {
...
+                        isUnreadMessage && "unread"
+                      }`}
+                    >
+                      {profileName}
+                    </div>
+                    <div
+                      className={`private-message-last ${
+                        isUnreadMessage && "unread"
+                      }`}
+                    >
```
0xdevalias, line 185: We should add the `classnames` dependency, as it would allow us to make this sort of thing way clearer/more concise across the app.

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -170,15 +171,32 @@ const ChatsList: React.FunctionComponent = () => {
...
-                  {lastMessageExchanged.from !== user?.uid &&
-                    !lastMessageExchanged.isRead && (
-                      <div className="not-read-indicator" />
-                    )}
+                  {lastMessageExchanged.from !== user?.uid && (
+                    <div
+                      className={`private-message-time ${
+                        isUnreadMessage && "unread"
+                      }`}
+                    >
```
0xdevalias, line 194: You've removed `not-read-indicator` class here, seemingly in favour of `unread`, yet in `src/components/molecules/ChatsList/ChatsList.scss` you set a style for `not-read-indicator`.

Is this correct?

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -170,15 +171,32 @@ const ChatsList: React.FunctionComponent = () => {
...
+                      {formatDistanceToNow(
+                        new Date(lastMessageExchanged.ts_utc.toDate())
```
0xdevalias, line 196: if you're calling `.toDate()`, do you need to wrap this in `new Date()`?

- https://date-fns.org/v2.2.1/docs/formatDistanceToNow#arguments

It seems it can take a `Number` as it's parameter, so can we just pass `.ts_utc` directly?

src/components/molecules/Sidebar/Sidebar.tsx
```diff
@@ -27,10 +29,13 @@ const filterUniqueKeys = (userId: string, index: number, arr: string[]) =>
   arr.indexOf(userId) === index;
 
 const Sidebar = () => {
+  const { user } = useUser();
   const [tab, setTab] = useState(0);
   const privateChats = useSelector(privateChatsSelector) ?? [];
   const chatUsers = useSelector(chatUsersSelector) ?? [];
   const isEnabled = chatUsers && privateChats;
+  const unreadMessages = filterUnreadPrivateChats(privateChats, user);
+  const hasUnreadMessages = !!unreadMessages.length;
```
0xdevalias, line 38: use `isTruthy` helper from `utils`?

src/utils/filter.ts
```diff
@@ -1,6 +1,14 @@
+import { chatSort, PrivateChatMessage } from "components/context/ChatContext";
+import { FirebaseReducer } from "react-redux-firebase";
+import { VENUE_CHAT_AGE_DAYS } from "settings";
 import { CampRoomData } from "types/CampRoomData";
+import { WithId } from "./id";
 
 import { isWithinBounds } from "./isWithinBounds";
+import { getDaysAgoInSeconds, roundToNearestHour } from "./time";
```
0xdevalias, line 8: [nit] import sorting: 3rd party libs at the top, then settings/etc, types, utils, relative imports

src/utils/filter.ts
```diff
@@ -63,3 +71,21 @@ export const makeCampRoomHitFilter = ({
...
+    chats
+      ?.filter(
+        (message) =>
+          message.from !== user?.uid &&
+          message.deleted !== true &&
+          message.type === "private" &&
+          message.ts_utc.seconds > HIDE_BEFORE &&
+          message.isRead === false
+      )
+      .sort(chatSort) ?? []
```
0xdevalias, line 89: [nit] You could either use a guard (eg. `if (!chats) return []`) up front, or possibly move the default value up front?

```suggestion
    if (!chats) return []
    
    return chats
      .filter(
        (message) =>
          message.from !== user?.uid &&
          message.deleted !== true &&
          message.type === "private" &&
          message.ts_utc.seconds > HIDE_BEFORE &&
          message.isRead === false
      )
      .sort(chatSort)
```

```suggestion
    (chats ?? [])
      .filter(
        (message) =>
          message.from !== user?.uid &&
          message.deleted !== true &&
          message.type === "private" &&
          message.ts_utc.seconds > HIDE_BEFORE &&
          message.isRead === false
      )
      .sort(chatSort)
```

src/utils/filter.ts
```diff
@@ -63,3 +71,21 @@ export const makeCampRoomHitFilter = ({
 
   return isWithinBounds(checkPercent, roomBounds);
 };
+
+export const filterUnreadPrivateChats = (
+  chats: WithId<PrivateChatMessage>[],
+  user: FirebaseReducer.AuthState | undefined
```
0xdevalias, line 77: Prefer using the `?` syntax rather than `| undefined` when possible.

```suggestion
  user?: FirebaseReducer.AuthState
```

src/utils/filter.ts
```diff
@@ -63,3 +71,21 @@ export const makeCampRoomHitFilter = ({
 
   return isWithinBounds(checkPercent, roomBounds);
 };
+
+export const filterUnreadPrivateChats = (
+  chats: WithId<PrivateChatMessage>[],
```
0xdevalias, line 76: It's probably a good habit to get into to make all arrays that are passed to things `readonly` in our types, to avoid/catch unintended mutations.

```suggestion
  chats: readonly WithId<PrivateChatMessage>[],
```

0xdevalias (changes requested):
A bunch of comments, a couple of questions, I think one of the classes/styles might be incorrect currently.

src/components/molecules/ChatsList/ChatsList.tsx
denisdimitrov: It needs to be removed, nice catch.

src/components/molecules/ChatsList/ChatsList.tsx
denisdimitrov: .ts_utc  is Timestamp that's why i'm callign .toDate() but I'll se if it works without new Date

src/utils/filter.ts
denisdimitrov: Updated.

src/utils/filter.ts
denisdimitrov: This is now using the optional syntax.

src/utils/filter.ts
denisdimitrov: Applied.

src/utils/filter.ts
denisdimitrov: The imports have been reordered.

src/components/molecules/ChatsList/ChatsList.tsx
denisdimitrov: I'm not sure. But even if it was user deleted the prettier should fix it.

denisdimitrov:
All comments have been fixed.

src/components/molecules/ChatsList/ChatsList.tsx
thecadams: Not to distract us from the prize of getting this merged, but I believe prettier isn't opinionated about 0/1 line, only when it's >1.

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -153,10 +155,10 @@ const ChatsList: React.FunctionComponent = () => {
               const sender = { ...chatUsers![userId], id: userId };
               const lastMessageExchanged =
                 discussionPartnerWithLastMessageExchanged?.[userId];
+              const isUnreadMessage = !isTruthy(lastMessageExchanged.isRead);
```
thecadams, line 158: something for later: looks like `isFalsy` would be another helpful helper.

thecadams (approved):
LGTM. Thanks for addressing the comments!

Pretty sure you can't merge til the requested changes review is cleared out unfortunately; but great to see the progress here.

0xdevalias:
> @thecadams Pretty sure you can't merge til the requested changes review is cleared out unfortunately; but great to see the progress here.

There are controls for either re-requesting my my review directly, or dismissing it if no longer relevant. IMO one or the other should be our workflow in these situations, so people are explicitly notified if they need to do something on a PR they'd previously looked at.

image

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -193,7 +194,7 @@ const ChatsList: React.FunctionComponent = () => {
                       }`}
                     >
                       {formatDistanceToNow(
-                        new Date(lastMessageExchanged.ts_utc.toDate())
+                        lastMessageExchanged.ts_utc.toDate()
```
0xdevalias, line 197: [nit] since `formatDistanceToNow` accepts a `number`, I believe it should be sufficient to do

```suggestion
                        formatDistanceToNow(lastMessageExchanged.ts_utc)
```

src/utils/filter.ts
```diff
@@ -73,19 +76,19 @@ export const makeCampRoomHitFilter = ({
...
-          message.type === "private" &&
-          message.ts_utc.seconds > HIDE_BEFORE &&
-          message.isRead === false
-      )
-      .sort(chatSort) ?? []
-  );
+  if (!chats) return [];
+
+  return chats
+    ?.filter(
```
0xdevalias, line 85: [nit] since `chats` is explicitly checked by the guard above, we shouldn't need the null-safe syntax anymore:

```suggestion
  return chats.filter(
```

src/utils/filter.ts
```diff
@@ -73,19 +76,19 @@ export const makeCampRoomHitFilter = ({
...
+      (message) =>
+        message.from !== user?.uid &&
```
0xdevalias, line 87: Also, if we added `!user` to the guard check (as presumably `message.from !== user?.uid` should never be true when `user` is `undefined`) then we can also remove the null-safe check here (and it makes it quick to visually skim and understand the logic without having to work trough the `&&`'d stuff as it would shortcut early on `!user`

0xdevalias (approved):
Only nits. LGTM!

(edit: whoops.. seemingly made all of those comments outside of this 'review session'... my bad)

PR #851, opened 2020-12-02
make artpiece grid columns five across

denisdimitrov:


src/components/templates/ArtPiece/ArtPiece.scss
```diff
@@ -17,7 +17,7 @@
   .video-chat-wrapper {
     display: grid;
     margin: 1rem 0;
-    grid-template-columns: repeat(6, 1fr);
+    grid-template-columns: repeat(9, 1fr);
```
0xdevalias, line 20: Just so I can understand it better, can you explain what this is doing/how it works please?

0xdevalias (commented):
The title of this PR says make the videos 3 across, but I believe that was the existing functionality.

The trello card I can see talks about making them 4 across. Is that what this PR is actually doing?

denisdimitrov:
[Art piece video arrangement only three across (should go to four)](https://trello.com/c/glKdnkQX/452-art-piece-video-arrangement-only-three-across-should-go-to-four)

src/components/templates/ArtPiece/ArtPiece.scss
denisdimitrov: I just spoke with Ed, and this should be 5 across. 1 video takes 3 columns. Not sure why, but we are in a hurry for the event and I don't have the time to look into it.

0xdevalias: No worries, don't need to clean it up now. Was just curious/for my own knowledge.

0xdevalias (approved):
Needs to be 5 across, videos apparently take up 3 columns. That matches the 15 in the changes here. LGTM!

PR #852, opened 2020-12-02
🐛 Room show/hide on map if enabled/disabled in admin panel

IvanBuljovcic:
Show/Hide room on the partymap/themecamp if it is Enabled/Disabled in the Admin panel

### DIFF
--- a/src/components/templates/Camp/components/MapRoomOverlay.tsx
+++ b/src/components/templates/Camp/components/MapRoomOverlay.tsx
@@ -139,6 +139,8 @@ export const MapRoomOverlay: React.FC<MapRoomOverlayProps> = ({
     (attendances[`${venue.name}/${room.title}`] ?? 0) +
     (room.attendanceBoost ?? 0);
 
+  if (!room.isEnabled) return <></>;
+
   return (
     <div
       className={`room position-absolute ${isSelectedRoom && "isUnderneath"}`}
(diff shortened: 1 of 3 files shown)


src/components/templates/Camp/components/MapRoomOverlay.tsx
```diff
@@ -139,6 +139,8 @@ export const MapRoomOverlay: React.FC<MapRoomOverlayProps> = ({
     (attendances[`${venue.name}/${room.title}`] ?? 0) +
     (room.attendanceBoost ?? 0);
 
+  if (!room.isEnabled) return <></>;
```
0xdevalias, line 142: Fragments are only really meant for grouping other sibling JSX elements together when we don't want to emit a `<div>`/similar to give them a common parent. In this case, we would want to return `null`

0xdevalias (changes requested):
Can you please update the `<></>` returns to `null` as per in-lined comment, then this should be good to go!

0xdevalias:
[Room turning on and off not working on partymap and themecamp templates](https://trello.com/c/689NmRaI/354-room-turning-on-and-off-not-working-on-partymap-and-themecamp-templates)

0xdevalias (approved):
LGTM!

thecadams:
Great to see this merged. Quickly wondering, what about the third stage of a room when it’s disabled. Should we have it be at say 50% opacity and not clickable?

PR #853, opened 2020-12-02
allow party map background change

denisdimitrov:
This code has been extracted from the  following PR #822  Because it's needed for the event


0xdevalias (approved):
LGTM!

PR #854, opened 2020-12-02
Stub out BugsnagErrorBoundary with React.Fragment when we're not using Bugsnag

0xdevalias:
eg. In local development


denisdimitrov (approved):
LGTM!

PR #856, opened 2020-12-02
fix room modal spacing in party map

denisdimitrov:


0xdevalias (approved):
Assuming this does the thing it says it does, LGTM!

PR #857, opened 2020-12-02
Add lint check to precommit hook

IvanBuljovcic:


package.json
```diff
@@ -103,7 +103,8 @@
   },
   "lint-staged": {
     "{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
-      "prettier --write"
+      "prettier --write",
+      "eslint src scripts functions --ext .js,.jsx,.ts,.tsx"
```
0xdevalias, line 107: According to https://github.com/okonet/lint-staged#filtering-files I believe that we probably only need something like:

```suggestion
      "eslint --ext .js,.jsx,.ts,.tsx"
```

or maybe even just:

```suggestion
      "eslint"
```

The glob of files from the initial `lint-staged` selector is passed to the command specified here.

As is, this is still going to let you commit files when there are `eslint` warnings though, which in CI will be treated as errors, and thus require changes to be made to fix it.

If you wanted to emulate this same functionality in this `lint-staged` hook, then I believe you could use `--max-warnings=0`:

- https://eslint.org/docs/user-guide/command-line-interface#options

So combined with the above, would potentially look something like:

```suggestion
      "eslint --max-warnings=0 --ext .js,.jsx,.ts,.tsx"
```

0xdevalias (changes requested):
As I understand your intent on wanting this here, it was to prevent committing when eslint would fail in CI.

I believe this won't have the desired effect, as when run here, they will still be warnings, and so will pass, yet in CI we treat warnings as errors, so that will fail.

Additionally (as mentioned during our call) I have concerns with blocking commits based on running linters/etc. In my opinion, enforcing these controls belongs at the CI level, and if individual developers want to have their own QoL checks at this stage, they should be free to do so, but it shouldn't be a requirement forced on the whole team. Doubly so since we squash merge our features, so there is less 'clutter' that would end up in the main branch from an accidentally missed linting step here or there.

I had a quick look, and it seems that in v4 of husky (which we are using), you need to explicitly opt out of installing the hooks:

- https://github.com/typicode/husky/tree/master#disable-auto-install

Whereas in v5 (which is the recommended version, even though it's still in pre-release) it appears to be an explicit opt-in step:

- https://typicode.github.io/husky/#/?id=install

Though it also describes how you can autoinstall it in a `postinstall` step within `package.json`. If `pinst` is used, then this can be auto-installed only for development:

- https://github.com/typicode/pinst

On the other hand, we are using lint-staged, so it is only going to be running the linters for the files about to be committed, not the entire repo:

- https://github.com/okonet/lint-staged

Which may reduce my "if it prevents me from committing while i'm in the middle of a big half-done refactor job, that would be annoying" concerns I originally had with this PR.

So.. I'm still not sure if I like the flow of forcing this on developers unless they explicitly opt-out, but it also probably isn't as big as a concern as I originally anticipated.

So.. let's try it, and if we don't like it, we can bump up to `husky@next` in a new PR and make it opt-in.

I'll be happy to approve, but for now am requesting changes based on the inline comments to make this change match what I believe your intent here was.

package.json
thecadams: 2 quick notes on this one.

1. The reason for the `src scripts functions` part is that linting `reporting`, annoyingly enough, includes any `node_modules` directory that is in there. So if we want to remove that, we should also ensure `node_modules` is ignored.

2. The list of file extensions is being supplied because (IIRC) otherwise, eslint tries to run over a wider range of file extensions and finds a variety of things it doesn't understand. I may be wrong though, so give that a try.

thecadams (approved):

0xdevalias (approved):

0xdevalias:
@IvanBuljovcic I've hit cases where this would have been useful so many times over the last few days, so while I didn't realise it was something I needed at the time, thanks for putting this up! Excited to have it merged in now.

PR #868, opened 2020-12-03
Create Code Climate config file

0xdevalias:
Adds Code Climate configuration file to the repo as per:

- https://docs.codeclimate.com/docs/advanced-configuration

Note: the settings as configured here are pretty much the defaults, but mapped out explicitly. We can customize it more in a future PR if we want to.


denisdimitrov (approved):
LGTM!

PR #873, opened 2020-12-03
More safety against missing data in firestore

thecadams:


thecadams:
> Code Climate has analyzed commit [c3937a7](https://github.com/sparkletown/sparkle/commit/c3937a70b9f14246dcf924a7b3101a015093e85b) and detected **2 issues** on this pull request.
> ...
> Duplication	2

Seems minor; I'm going to ignore this. We only added this today, so it's still being calibrated, but to my mind this type of duplication is far beneath the level we are concerned about.

src/pages/Admin/Admin.tsx
```diff
@@ -102,7 +102,7 @@ const VenueList: React.FC<VenueListProps> = ({
             <Link to={`/admin/venue/${venue.id}`}>{venue.name}</Link>
             {isVenueWithRooms(venue) && (
               <ul className="page-container-adminsidebar-subvenueslist">
-                {venue.rooms.map((room, idx) => (
+                {venue.rooms?.map((room, idx) => (
```
thecadams, line 105: This is the issue that was crashing the Admin page. I thought it'd be prudent to fix all the rest of the trouble with the PartyMap template.

denisdimitrov: I suppose our Types are incorrect and give false sense of safety. We should probably have a call and discuss this.

denisdimitrov: This could be an issue in other components as well, at this very moment.

0xdevalias: https://github.com/sparkletown/sparkle/pull/744 my 'we should fix the types' PR that I keep not having time to get back to.

src/types/PartyMapVenue.ts
```diff
@@ -8,24 +8,24 @@ export const isPartyMapVenue = (venue: Venue): venue is PartyMapVenue =>
...
+  unhosted_entry_video_url?: string;
+  map_url?: string;
+  map_viewbox?: string;
+  password?: string;
   admin_password?: string;
-  owners: string[];
-  rooms: PartyMapRoomData[];
+  owners?: string[];
+  rooms?: PartyMapRoomData[];
 }
```
0xdevalias, line 31: 👌 🎉

0xdevalias (approved):
[CRO] Changes look good, LGTM!

0xdevalias:
> @thecadams Seems minor; I'm going to ignore this. We only added this today, so it's still being calibrated, but to my mind this type of duplication is far beneath the level we are concerned about.

Yup, agreed. I haven't set it up to block PR's, so we can treat it as guidance/FYI, 'providing info and giving us a feel for what we want to calibrate/change in it, etc.

PR #840, opened 2020-11-30
[Original Outdated WIP PR] Merge missing Camp features into PartyMap, then 💥 the old Camp components

0xdevalias:
Ensure PartyMap has all of the features from the old Camp components, then 💥 the old Camp components

## Preview Deploy

- see URL in https://github.com/sparkletown/sparkle/pull/840#issuecomment-759346391

## Things that still might need doing/deciding

- [x] Once https://github.com/sparkletown/sparkle/pull/1070 gets merged to `staging` we should merge it into this branch for easier QA testing
- [x] Check the styling on PartyMap to ensure it's not broken (see https://github.com/sparkletown/sparkle/pull/840#discussion_r535017810, maybe do this in a new PR?)
- [ ] Does `onSeatClick` in `PartyMap` need to call `checkForRoomHit` anymore? (see https://github.com/sparkletown/sparkle/pull/840#discussion_r532538359)
- [ ] Should we refactor `unselectRoom` in PartyMap to only unselect if it's the same room it was 'bound' to to avoid race conditions? (see https://github.com/sparkletown/sparkle/pull/840#discussion_r532499105)
- [ ] Did we explicitly want to change to using the `img` rather than a css `backgroundImage` for the map background? (see https://github.com/sparkletown/sparkle/pull/840#discussion_r532541935)
- [x] What is the purpose of `party-map-grid-container` and do we want to keep it? (see https://github.com/sparkletown/sparkle/pull/840#discussion_r532542588)
    - backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/172

PR #887, opened 2020-12-03
Always join rooms by opening a new tab

thecadams:
- Make footer links open in new tab
- Make join links open in new tab

### DIFF
--- a/src/components/molecules/Footer/Footer.tsx
+++ b/src/components/molecules/Footer/Footer.tsx
@@ -25,12 +25,7 @@ export const Footer = () => (
     <div>|</div>
 
     <div>
-      <a
-        className="powered-by"
-        href={HOMEPAGE_URL}
-        target="_blank"
-        rel="noopener noreferrer"
-      >
+      <a href={HOMEPAGE_URL} target="_blank" rel="noopener noreferrer">
         Made with{" "}
         <span role="img" aria-label="heart">
           ❤️
--- a/src/components/molecules/Footer/Footer.tsx
+++ b/src/components/molecules/Footer/Footer.tsx
@@ -10,7 +10,13 @@ import "./Footer.scss";
 
 export const Footer = () => (
   <div className="footer-wrapper">
-    <a href={TERMS_AND_CONDITIONS_URL}>Terms of Use</a>
+    <a
+      href={TERMS_AND_CONDITIONS_URL}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      Terms of Use
+    </a>
 
     <div>|</div>
 
@@ -19,7 +25,12 @@ export const Footer = () => (
     <div>|</div>
 
     <div>
-      <a className="powered-by" href={HOMEPAGE_URL}>
+      <a
+        className="powered-by"
+        href={HOMEPAGE_URL}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
         Made with{" "}
         <span role="img" aria-label="heart">
           ❤️
--- a/src/components/molecules/Footer/Footer.tsx
+++ b/src/components/molecules/Footer/Footer.tsx
@@ -5,16 +5,13 @@ import {
   PRIVACY_POLICY,
   TERMS_AND_CONDITIONS_URL,
 } from "settings";
+import { getExtraLinkProps } from "utils/url";
 
 import "./Footer.scss";
 
 export const Footer = () => (
   <div className="footer-wrapper">
-    <a
-      href={TERMS_AND_CONDITIONS_URL}
-      target="_blank"
-      rel="noopener noreferrer"
-    >
+    <a href={TERMS_AND_CONDITIONS_URL} {...getExtraLinkProps(true)}>
       Terms of Use
     </a>
 
@@ -25,7 +22,7 @@ export const Footer = () => (
     <div>|</div>
 
     <div>
-      <a href={HOMEPAGE_URL} target="_blank" rel="noopener noreferrer">
+      <a href={HOMEPAGE_URL} {...getExtraLinkProps(true)}>
         Made with{" "}
         <span role="img" aria-label="heart">
           ❤️
--- a/src/components/molecules/OnlineStats/OnlineStats.tsx
+++ b/src/components/molecules/OnlineStats/OnlineStats.tsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useState, useMemo } from "react";
 import firebase from "firebase/app";
 import "firebase/functions";
 import { OverlayTrigger, Popover } from "react-bootstrap";
-import { venueInsideUrl } from "utils/url";
+import { openUrl, venueInsideUrl } from "utils/url";
 import { WithId } from "utils/id";
 import { AnyVenue } from "types/Firestore";
 import { VenueEvent } from "types/VenueEvent";
@@ -44,7 +44,7 @@ const PotLuckButton: React.FC<PotLuckButtonProps> = ({
     afterSelect();
 
     // there is a bug in useConnectCurrentVenue that does not update correctly on url change
-    window.location.href = venueInsideUrl(randomVenue.id);
+    openUrl(venueInsideUrl(randomVenue.id));
   }, [venues, afterSelect]);
   if (!venues) {
     return <></>;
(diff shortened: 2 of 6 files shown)


src/components/templates/Camp/components/MapRoomOverlay.tsx
```diff
@@ -97,12 +97,7 @@ export const MapRoomOverlay: React.FC<MapRoomOverlayProps> = ({
...
       e.stopPropagation();
 
-      if (isExternalUrl(room.url)) {
-        openRoomUrl(room.url);
-      } else {
-        window.location.href = getRoomUrl(room.url);
-      }
-
+      openRoomUrl(room.url);
       enterCampRoom(room);
```
0xdevalias, line 101: These changes are insufficient to handle all room entering in this component. See https://github.com/sparkletown/sparkle/pull/888

src/components/molecules/Footer/Footer.tsx
```diff
@@ -10,7 +10,13 @@ import "./Footer.scss";
 
 export const Footer = () => (
   <div className="footer-wrapper">
-    <a href={TERMS_AND_CONDITIONS_URL}>Terms of Use</a>
+    <a
+      href={TERMS_AND_CONDITIONS_URL}
+      target="_blank"
+      rel="noopener noreferrer"
```
0xdevalias, line 16: there is a helper in `utils/url` that you can spread to get these props

src/components/molecules/Footer/Footer.tsx
```diff
@@ -19,7 +25,12 @@ export const Footer = () => (
     <div>|</div>
 
     <div>
-      <a className="powered-by" href={HOMEPAGE_URL}>
+      <a
+        className="powered-by"
+        href={HOMEPAGE_URL}
+        target="_blank"
+        rel="noopener noreferrer"
```
0xdevalias, line 32: as above

src/components/molecules/Footer/Footer.tsx
```diff
@@ -19,7 +25,12 @@ export const Footer = () => (
     <div>|</div>
 
     <div>
-      <a className="powered-by" href={HOMEPAGE_URL}>
+      <a
+        className="powered-by"
```
0xdevalias, line 29: i don't think this class exists anymore (I must have missed it before)

src/components/molecules/OnlineStats/OnlineStats.tsx
```diff
@@ -44,7 +44,7 @@ const PotLuckButton: React.FC<PotLuckButtonProps> = ({
     afterSelect();
 
     // there is a bug in useConnectCurrentVenue that does not update correctly on url change
-    window.location.href = venueInsideUrl(randomVenue.id);
+    openUrl(venueInsideUrl(randomVenue.id));
```
0xdevalias, line 47: how does `venueInsideUrl` and the get room url function that `openUrl` uses work together?

src/components/molecules/VenueInfoEvents/VenueInfoEvents.tsx
```diff
@@ -86,7 +86,7 @@ const VenueInfoEvents: React.FunctionComponent<PropsType> = ({
               eventsNow.length ? "btn-primary" : "btn-secondary"
             } btn-block`}
             // @debt would be nice not to refresh the page
-            onClick={() => (window.location.href = venueInsideUrl(venue.id))}
+            onClick={() => openUrl(venueInsideUrl(venue.id))}
```
0xdevalias, line 89: As above RE: potential clashes with `venueInsideUrl` + function inside `openUrl`

src/components/molecules/OnlineStats/OnlineStats.tsx
0xdevalias: apparently I am thinking of `openRoomUrl`. Does that serve the purpose of `venueInsideUrl`?

src/components/molecules/VenueInfoEvents/VenueInfoEvents.tsx
0xdevalias: apparently I am thinking of `openRoomUrl`. Does that serve the purpose of `venueInsideUrl`?

0xdevalias:
My emergency fix in #888 seems to have missed out on an edgecase via `src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx`'s 'enter' button and room opening (we have way too many independent/repeated ways to do things at the moment..)

I believe your PR will probably cover it by changing the underlying helper function, but would be good to ensure that.

src/components/molecules/VenueInfoEvents/VenueInfoEvents.tsx
thecadams: There's no need to use `openRoomUrl` here since we control the generated URL and it'll never not have `http` in it.

thecadams:
> I believe your PR will probably cover it by changing the underlying helper function, but would be good to ensure that.

Yep it will by changing the helper: https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx#L35

```
openUrl(getRoomUrl(room.url));
```

src/components/molecules/OnlineStats/OnlineStats.tsx
thecadams: As below, we won't need `openRoomUrl` here because `venueInsideUrl` will always generate a URL prefixed with `http`.

src/components/templates/Camp/components/MapRoomOverlay.tsx
```diff
@@ -92,10 +92,8 @@ export const MapRoomOverlay: React.FC<MapRoomOverlayProps> = ({
     (e) => {
       e.stopPropagation();
 
+      openRoomUrl(room.url);
       enterCampRoom(room);
```
0xdevalias, line 96: It shouldn't matter since we're opening a new tab, but I think this is better to sit like:
```suggestion
      enterCampRoom(room);
      openRoomUrl(room.url);
```

`enterCampRoom` (a terribly unintuitive name, refactored in 'the great merger' https://github.com/sparkletown/sparkle/pull/840 to something like `trackRoomEntered` or similar) is basically telling firestore they entered, and then we have to 'actually enter' on the client-side (in this case, open the tab)

src/components/molecules/Footer/Footer.tsx
```diff
@@ -5,12 +5,15 @@ import {
...
   TERMS_AND_CONDITIONS_URL,
 } from "settings";
+import { getExtraLinkProps } from "utils/url";
 
 import "./Footer.scss";
 
 export const Footer = () => (
   <div className="footer-wrapper">
-    <a href={TERMS_AND_CONDITIONS_URL}>Terms of Use</a>
+    <a href={TERMS_AND_CONDITIONS_URL} {...getExtraLinkProps(true)}>
```
0xdevalias, line 14: [nit] the `true` in getExtraLinkProps doesn't tell us a lot. We should refactor `getExtraLinkProps` to take a `{external: true}` style 'props' param.

0xdevalias (approved):
A couple of tiny changes that would be good to make before merging, but otherwise this looks good to go.

If we haven't already, at some point it would be good to search the whole codebase for `window.open` and similar and refactor those places to use `utils/url`

src/utils/url.ts
```diff
@@ -39,7 +40,15 @@ export const openRoomUrl = (url: string) => {
 };
 
 export const openUrl = (url: string) => {
-  window.open(url, "_blank", "noopener,noreferrer");
+  isValidUrl(url)
+    ? window.open(url, "_blank", "noopener,noreferrer")
+    : console.error(
+        `Invalid URL ${url} on page ${window.location.href}; ignoring`
+      );
```
0xdevalias, line 47: @thecadams this is a perfect place where we can make use of the `Bugsnag.notify` helper to explicitly report an error, docs for usage:

- https://docs.bugsnag.com/platforms/javascript/#reporting-handled-errors

0xdevalias: Will do a short followup PR to fix this + any other console.error's that we would benefit from gaining insight into

0xdevalias: https://github.com/sparkletown/sparkle/pull/889

PR #889, opened 2020-12-04
Fix bug in utils/url isValidUrl and use Bugsnag.notify instead of console.error logging

0xdevalias:
Using `Bugsnag.notify` will give us visibility into issues that we would otherwise never see if they are only logged with `console.error`

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -1,3 +1,4 @@
+import Bugsnag from "@bugsnag/js";
 import { CODE_CHECK_URL } from "secrets";
 import { VALID_URL_PROTOCOLS } from "settings";
 import { CampVenue } from "types/CampVenue";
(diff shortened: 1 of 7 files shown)


src/utils/url.ts
```diff
@@ -40,15 +41,25 @@ export const openRoomUrl = (url: string) => {
...
+
+  window.open(url, "_blank", "noopener,noreferrer");
 };
 
-export const isValidUrl = (url: string) => {
-  return VALID_URL_PROTOCOLS.includes(new URL(url).protocol);
+export const isValidUrl = (url: string): boolean => {
+  try {
+    return VALID_URL_PROTOCOLS.includes(new URL(url).protocol);
+  } catch (e) {
```
thecadams, line 60: nit: consider catching only `TypeError` here: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

thecadams (approved):
Looks good, only one nit, beginnings of observability greatness here.

src/utils/url.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/889/commits/935aeb69024ccebf57b50e64566a4edd511425d7

thecadams:
[Bring Back Badges](https://trello.com/c/AAmGbZVs/447-bring-back-badges)

PR #833, opened 2020-11-30
WIP: [IMPROVEMENT] Removes outdated and no longer used parts of the codebase

thecadams:
Targeted for removal:
- Old templates
- Old party configs from when the config was uploaded with a script
- Playa
- Dust storm (it's part of the playa)
- Placement
- Playa address
- others still to come

## TODO

- [ ] Remove `FriendShipPage` and related (see https://github.com/sparkletown/sparkle/pull/833#issuecomment-767948624)

PR #890, opened 2020-12-04
add useInterval hook + related refactors/cleanup (including some room counting things)

0xdevalias:
Note: this PR should also align a few disparate places that are likely negatively affecting room counts.


src/components/organisms/DustStorm/DustStorm.tsx
```diff
@@ -53,24 +56,17 @@ export const DustStorm = () => {
     []
   );
 
-  useEffect(() => {
-    const getOnlineStats = firebase
+  useInterval(() => {
```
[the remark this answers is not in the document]
0xdevalias: `DustStorm` will be removed as part of https://github.com/sparkletown/sparkle/pull/833, so not worth doing here.

src/pages/Schedule/SchedulePage.tsx
```diff
@@ -26,27 +29,18 @@ const DAYS_AHEAD = 100;
 
 export const SchedulePage = () => {
   const [openVenues, setOpenVenues] = useState<OpenVenues>();
-  const [, setLoaded] = useState(false);
 
-  useEffect(() => {
-    const getOnlineStats = firebase
+  useInterval(() => {
```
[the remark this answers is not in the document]
0xdevalias: see explanation in https://github.com/sparkletown/sparkle/pull/890#discussion_r535815941

src/utils/time.js
```diff
@@ -1,8 +1,10 @@
 import { format } from "date-fns";
 
 export const ONE_MINUTE_IN_SECONDS = 60;
-const ONE_HOUR_IN_SECONDS = ONE_MINUTE_IN_SECONDS * 60;
-const ONE_DAY_IN_SECONDS = ONE_HOUR_IN_SECONDS * 24;
+export const ONE_HOUR_IN_SECONDS = ONE_MINUTE_IN_SECONDS * 60;
+export const ONE_DAY_IN_SECONDS = ONE_HOUR_IN_SECONDS * 24;
+
+export const FIVE_MINUTES_MS = 5 * 60 * 1000;
```
thecadams, line 7: nit: we should be consistent about whether to include `_IN` here:

* `FIVE_MINUTES_IN_MS` vs. `FIVE_MINUTES_MS`
* `ONE_HOUR_IN_SECONDS` vs `ONE_HOUR_SECONDS`

My leaning is toward including `_IN`, but feel free to make the call.

src/components/templates/Playa/PlayaBackground.tsx
```diff
@@ -12,31 +13,29 @@ interface Props {
...
+  nightCycle,
+}) => {
   const [daypartClassName, setDaypartClassName] = useState("");
 
-  useEffect(() => {
-    const updateStats = () => {
-      const pt = dayjs().tz("America/Los_Angeles");
-      const ptHour = pt.hour();
+  useInterval(() => {
+    const pt = dayjs().tz("America/Los_Angeles");
```
thecadams, line 23: OOS/future: The timezone should be moved to settings or venue config.

src/components/templates/Playa/Playa.tsx
```diff
@@ -394,29 +396,22 @@ const Playa = () => {
...
-  }, [profile]);
+      .httpsCallable("stats-getLiveAndFutureEvents")()
+      .then((result) => {
+        const { openVenues } = result.data as OnlineStatsData;
+        setOpenVenues(
+          profile?.kidsMode
+            ? openVenues.filter((v) => !v.venue.adultContent)
+            : openVenues
+        );
+        //setOpenVenues(openVenues);
```
thecadams, line 411: nit: commented code

src/components/templates/Playa/Playa.tsx
```diff
@@ -429,7 +424,11 @@ const Playa = () => {
   const menuRef = useRef<HTMLDivElement | null>(null);
   const [, setRerender] = useState(0);
   const [shoutText, setShoutText] = useState("");
-  const [nowMs, setNowMs] = useState(new Date().getTime());
+  const [nowMs, setNowMs] = useState(Date.now());
```
thecadams, line 427: Policy note: this is good; just checked and these both use UTC. As long as we're avoiding any local time we are good.

src/components/molecules/PlayaTime/PlayaTime.tsx
```diff
@@ -21,15 +22,10 @@ dayjs.updateLocale("en", {
...
-    const updateTime = () => {
-      const pt = dayjs().tz("Australia/Sydney");
-      setCurrentTime(pt.format("h:mm a"));
-    };
-    updateTime();
-    const id = setInterval(updateTime, 1000);
-    return () => clearInterval(id);
-  }, []);
+  useInterval(() => {
+    const pt = dayjs().tz("Australia/Sydney");
```
thecadams, line 26: OOS/future: we should move this to settings or make it venue config, some time soon.

src/components/molecules/LoadingPage/LoadingPage.jsx
```diff
@@ -27,16 +28,12 @@ const quotes = IS_BURN
...
+  useInterval(
+    () => {
       setQuote(quotes[Math.floor(Math.random() * quotes.length)]);
-    }, 1000);
-    return () => {
-      clearInterval(id);
-    };
-  }, [setQuote]);
+    },
+    IS_BURN ? 1000 : false
```
thecadams, line 35: From code inspection, it appears this won't run `setQuote` in non-burn mode, because `useInterval` will shortcut out. Which means no loading quote will be set. Is that correct?

thecadams (approved):
Looks good overall. Nothing blocking but a few comments so I'll approve for you to merge when you feel you've resolved enough of them. Thanks for this change, great cleanup.

src/utils/time.js
0xdevalias: I would probably cut `ONE_HOUR_SECONDS` down to `ONE_HOUR_SEC` or similar. I have another WIP PR converting this file to `.ts`, and then probably 🔥 a lot of this, it's super overly verbose, and often implemented in an overly complex way.

eg.
```diff
-export const getCurrentTimeInUTCSeconds = () => {
-  const date = new Date();
-  return (
-    Date.UTC(
-      date.getUTCFullYear(),
-      date.getUTCMonth(),
-      date.getUTCDate(),
-      date.getUTCHours(),
-      date.getUTCMinutes(),
-      date.getUTCSeconds()
-    ) / 1000
-  );
-};
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now
+//   The static Date.now() method returns the number of milliseconds elapsed since January 1, 1970 00:00:00 UTC.
+export const getCurrentTimeInUTCSeconds = () => Date.now() / 1000;
```

src/components/templates/Playa/PlayaBackground.tsx
0xdevalias: Or user config.

src/utils/time.js
thecadams: Perhaps we can also use a linter to disallow `new Date()` and `Date.getTime()`.

src/components/templates/Playa/Playa.tsx
0xdevalias: This was existing in the template prior, so I left it as was. Either way this whole thing is 🔥  once https://github.com/sparkletown/sparkle/pull/833 is ready

src/components/templates/Playa/Playa.tsx
0xdevalias: Agree, UTC AllTheThings!

src/components/templates/Playa/PlayaBackground.tsx
thecadams: > Or user config.

That's a good idea. Something like how we do `mirrorVideo`.

src/components/molecules/PlayaTime/PlayaTime.tsx
0xdevalias: User config IMO

src/components/molecules/LoadingPage/LoadingPage.jsx
0xdevalias: _nods_ basically. If no 'interval time' is provided, there's no point setting an interval (as it's a required param in the underlying command), so we just return early. I think I wrote this as `| undefined` somewhere else.

0xdevalias: Though it may make sense to abstract the whole 'show a quote' thing into a sub-component. Then conditionally render that whole component based on `IS_BURN`

thecadams: Yes, or to default the text to `Loading...` and only run the `useInterval` based on `IS_BURN`.

src/utils/time.js
0xdevalias: I was thinking the same myself.

src/components/templates/Playa/PlayaBackground.tsx
0xdevalias: Exactly!

src/components/molecules/PlayaTime/PlayaTime.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/14

src/components/templates/Playa/PlayaBackground.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/14

src/components/molecules/LoadingPage/LoadingPage.jsx
0xdevalias: You can't conditionally render hooks, they have to be after all your branching flows. So you would need to put the 'should this run or should I bail out?' logic inside the hook itself. Which is sort of what i've done here (though I admit it's not the simplest/most discoverable thing.

I think at the very least this warrants another `note:` entry in the JS docs for `src/hooks/useInterval.ts`, and maybe an `@example` entry that demonstrates using it this way.

src/hooks/useInterval.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * React hook that simplifies using setInterval() within functional components.
+ *
+ * Note: You don't need to memoise/useCallback the handler as we store it with
+ * a ref internally.
+ *
+ * @param handler the function to be called every tick of the interval
+ * @param intervalInMs the interval time in milliseconds
+ */
+export const useInterval = (handler: () => void, intervalInMs?: number) => {
```
0xdevalias, line 12: We should add an explicit callout about what happens when `intervalInMs` is `undefined`, and maybe some `@example` about how to use this hook in both the normal case, and the 'conditionally run' case, as called out in https://github.com/sparkletown/sparkle/pull/890#discussion_r536484984

src/components/molecules/LoadingPage/LoadingPage.jsx
0xdevalias: see https://github.com/sparkletown/sparkle/pull/890#discussion_r536485227

thecadams: Sounds good and it's easy to test, just as long as users will see `Loading...` then my concern is solved here.

src/hooks/useInterval.ts
thecadams: Yep agreed; from what I can see the case where `intervalMs` is `undefined` does nothing, so we could shortcut out? Just trying to understand the intent behind making `intervalMs` optional.

It seems equivalent to:

```
export const useInterval = (handler: () => void, intervalInMs?: number) => {
  if (!intervalMs) return;
  ...
```

0xdevalias: You can't conditionally render hooks. If you tried to bail out there, react would complain, as you have branching logic before the `useEffect`'s/etc below.

If you look at https://github.com/sparkletown/sparkle/pull/890/files/82625de2d8cfa326c4b2127dfdc6fe28fd086fbc#diff-1731fae0e569c8786cc66cc2d47102f32a35838a19aa9d417647cab9db39cacdR21-R31 you can see that we have the early bailout in the `useEffect`, so that the interval will never be created.

So all that needs to change here is adding some extra documentation/explanation in the JSdocs to make this obvious to people.

thecadams: Yep, just needs to be documented.

thecadams, line 12:
```suggestion
/**
 * React hook that simplifies using setInterval() within functional components.
 * If `null` or `undefined` are supplied `intervalInMs`, the hook will do nothing.
 * You can use this to avoid conditionally calling the hook.
 *
 * Note: You don't need to memoise/useCallback the handler as we store it with
 * a ref internally.
 *
 * @param handler the function to be called every tick of the interval
 * @param intervalInMs the interval time in milliseconds. If falsy, the hook
 * will do nothing.
 */
export const useInterval = (handler: () => void, intervalInMs?: number) => {
```

src/utils/time.js
0xdevalias: backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/20

src/hooks/useInterval.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/890/commits/2c8050bf8c7a11cd615de82f3f800358a5f41a77

src/components/molecules/LoadingPage/LoadingPage.jsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/890/commits/2c8050bf8c7a11cd615de82f3f800358a5f41a77

0xdevalias:
```suggestion
    IS_BURN ? 1000 : undefined
```

src/utils/time.js
0xdevalias: Going to leave this as is in this PR, we can address it as part of https://github.com/sparkletown/sparkle/pull/897 and/or future. I expect a lot of this stuff may get cut down in future code cleanups/refactors.

0xdevalias (commented):
Just done another visual pass over the code differences here, and looks good to me.

PR #895, opened 2020-12-04
Restore badges

thecadams:
Badges only appear for related venues


src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -145,7 +147,10 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
   );
 };
 
-const Badges: React.FC<{ user: WithId<User> }> = ({ user }) => {
+const Badges: React.FC<{ user: WithId<User>; currentVenue: WithId<Venue> }> = ({
```
0xdevalias, line 150: Would it be possible to pull this out into a separate component file for this PR? Feels 'hidden away' at the moment.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -159,6 +164,12 @@ const Badges: React.FC<{ user: WithId<User> }> = ({ user }) => {
   );
   const venues = useSelector(orderedVenuesSelector);
 
+  const { relatedVenues } = useConnectRelatedVenues({
+    venueId: currentVenue.id,
+    withEvents: false,
+  });
+  const relatedVenueIds = relatedVenues.map((venue) => venue.id);
+
   const playaTime = useMemo(() => {
```
0xdevalias, line 173: Is this just named poorly, or is it something playa-only that we don't want to show here?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -167,18 +178,26 @@ const Badges: React.FC<{ user: WithId<User> }> = ({ user }) => {
...
     return visits
       .filter((visit) => visit.id !== PLAYA_VENUE_NAME) // no badge for the Playa. Also does not have a logo
       .map((visit) => {
         const { venue, room } = findVenueAndRoomByName(visit.id, venues);
-        if (!venue) return undefined;
+        if (
+          !venue ||
+          (currentVenue.id !== venue.id && !relatedVenueIds.includes(venue.id))
+        )
+          return undefined;
```
0xdevalias, line 200: Are these basically doing the same processing? Except the first one just gets the length of it, whereas the second one maps them to some things?

I'm wondering if we can process the array once, then read the count straight off that.

I'm not deeply following the logic here, so unsure if that is possible.

thecadams:
[Bring Back Badges](https://trello.com/c/AAmGbZVs/447-bring-back-badges)

src/components/organisms/UserProfileModal/UserProfileModal.tsx
thecadams: Yes, they can be deduped, just did! Good call.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
thecadams: Yep! Doing that now.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+    {
+      collection: "users",
+      doc: user.id,
+      subcollections: [{ collection: "visits" }],
+      storeAs: "userModalVisits",
+    },
+  ]);
+  const visits = useSelector(
+    (state) => state.firestore.ordered.userModalVisits
+  );
```
0xdevalias, line 35: [nit] extract selector functions (even if it is just to a const outside of the component, but still in the same file

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  return (
+    <>
+      <div className="visits">
+        <div className="visit-item">
+          <span className="visit-item__value">
+            {visitHours > 1 ? `${visitHours}` : "< 1"} hrs
+          </span>
+          <span className="visit-item__label">Time spent in Sparkle</span>
```
0xdevalias, line 99: [Future] extract platform name to a const in settings or similar.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+        </div>
+        <div className="separator"></div>
```
0xdevalias, line 101: [nit]

```suggestion
        <div className="separator" />
```

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+          </span>
+          <span className="visit-item__label">Venues visited</span>
+        </div>
+      </div>
+      <div className="badges-container">
+        <div className="badges-title">{badges.length} Badges</div>
+        <ul className="badge-list">
+          {badges.map(
+            (b) =>
+              b && (
```
0xdevalias, line 114: this should ideally be extracted from the JSX + `useMemo`'d. I think the extra `b &&` check should be irrelevant since you filtered out `undefined` above?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+                  <Link to={getLocationLink(b.venue, b.room)}>
+                    <img
+                      className="badge-list-item-image"
+                      src={b.image}
+                      alt={`${b.label} badge`}
+                    />
+                  </Link>
+                </li>
+              )
+            // label missing on hover - similar to playa
```
0xdevalias, line 125: is this still relevant?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const findVenueAndRoomByName = (
+  nameOrRoomTitle: string,
+  venues: Array<WithId<AnyVenue>>
+) => {
+  const venue = venues.find(
+    (v) =>
+      v.name === nameOrRoomTitle ||
+      (isCampVenue(v) && v.rooms.find((r) => r.title === nameOrRoomTitle))
+  );
```
0xdevalias, line 141: Will this need to use the 'venue with rooms' thing instead of 'camp venue'? We've hit a few bugs where things didn't work as expected because they were using this directly.

Also, does this bigger function make sense to move to a `utils/` or similar for reusability?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+  if (!venue) return {};
+
+  if (venue.name === nameOrRoomTitle) return { venue };
+
+  return {
+    venue,
+    room:
+      isCampVenue(venue) &&
+      venue.rooms.find((r) => r.title === nameOrRoomTitle),
+  };
```
0xdevalias, line 152: This is the same logic as used in the `const venue` above isn't it? Can we de-dupe and use that in both places?

thecadams: I had a shot at this, but there's some bad typing in the `rooms` property on the different templates. That will all go away with #833 so going to leave it for now :(

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { Venue } from "types/Venue";
+
+import { WithId } from "utils/id";
+import { orderedVenuesSelector } from "utils/selectors";
+import { venueInsideUrl, venuePreviewUrl } from "utils/url";
+
+export const Badges: React.FC<{
+  user: WithId<User>;
+  currentVenue: WithId<Venue>;
+}> = ({ user, currentVenue }) => {
```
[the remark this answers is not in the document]
0xdevalias: [OOS] this is minor tweaks to existing code, we can clean this up more in future

src/components/organisms/Badges/Badges.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/895/commits/3e72c18e563d7fbd06de2c6c287223f3a54ec7be

src/components/organisms/Badges/Badges.tsx
```diff
@@ -108,23 +126,7 @@ export const Badges: React.FC<{
...
-                      src={b.image}
-                      alt={`${b.label} badge`}
-                    />
-                  </Link>
-                </li>
-              )
-            // label missing on hover - similar to playa
-          )}
-        </ul>
+        <ul className="badge-list">{badgeList}</ul>
```
0xdevalias, line 129: [nit] I personally probably would have kept the `ul` in the `badgeList` const/memo as well

src/utils/types.ts
```diff
@@ -2,3 +2,6 @@ export const isTruthy = <T>(item: T): boolean => !!item;
 
 export const hasElements = <T>(array?: T[] | null): boolean =>
   isTruthy(array?.length);
+
+export const notEmpty = <T>(value: T | null | undefined): value is T =>
+  value !== null && value !== undefined;
```
0xdevalias, line 7: [OOS, Future, musing] I wonder if these belong in a better place than `utils/types` (honestly not even sure what this filename means in this usage), I know we have `utils/filters`, though these also have usages outside of just being used with filters. `utils/utils`? :p

src/components/organisms/Badges/Badges.tsx
0xdevalias: Resolved in https://github.com/sparkletown/sparkle/pull/895/commits/3e72c18e563d7fbd06de2c6c287223f3a54ec7be

src/components/organisms/Badges/Badges.tsx
0xdevalias: Resolved in https://github.com/sparkletown/sparkle/pull/895/commits/0f78a1c0fe84cdbe3401086ceaa5ce37d85778f3

src/components/organisms/Badges/Badges.tsx
```diff
@@ -85,6 +87,22 @@ export const Badges: React.FC<{
...
+  const badgeList = useMemo(
+    () =>
+      badges.filter(notEmpty).map((b) => (
+        <li className="badge-list-item" key={b.label}>
+          <Link to={getLocationLink(b.venue, b.room)}>
+            <img
+              className="badge-list-item-image"
+              src={b.image}
+              alt={`${b.label} badge`}
+            />
```
0xdevalias, line 99: [nit] I generally prefer to use full names for variables, reduces cognitive complexity when reading. So in this instance, it would be using `badge` instead of `b`

src/components/organisms/Badges/Badges.tsx
```diff
@@ -85,6 +87,22 @@ export const Badges: React.FC<{
       .filter((b) => b !== undefined);
   }, [relevantVisits, venues, currentVenue, relatedVenueIds]);
 
+  const badgeList = useMemo(
+    () =>
+      badges.filter(notEmpty).map((b) => (
+        <li className="badge-list-item" key={b.label}>
+          <Link to={getLocationLink(b.venue, b.room)}>
```
0xdevalias, line 94: [NB] Will this have impacts on counting, since presumably we're linking directly to a room? (as a wider thing.. where possible it would be good if we could capture counting when a page/thing loads (for on-platform stuff), rather than relying on us having a perfect implementation at every 'calling site'.

0xdevalias: @denisdimitrov FYI: if you get to looking at counting things, this may be another pattern of the types of things we're doing where we may not be correctly registering the count. I haven't deeply validated that, so treat this as a potential thing to look into if you get to it, rather than a "this is definitely related"

src/components/organisms/Badges/Badges.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/895/commits/3e72c18e563d7fbd06de2c6c287223f3a54ec7be

src/components/organisms/Badges/Badges.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/895/commits/3e72c18e563d7fbd06de2c6c287223f3a54ec7be#

0xdevalias: [Future] extract `findVenueAndRoomByName` function from here into a `utils/*` file for reusability.

src/components/organisms/Badges/Badges.tsx
0xdevalias: Makes sense.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,167 @@
...
+      )),
+    [badges]
+  );
+
+  if (!relevantVisits) {
+    return <>Visit venues to collect badges!</>;
+  }
+
+  return (
+    <>
```
0xdevalias, line 111: [nit] unneeded `React.Fragment` wrapper as we only have a single 'root `div`' here anyway.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -0,0 +1,167 @@
...
+  if (venue.name === PLAYA_VENUE_NAME) {
+    return venueInsideUrl(venue.id);
+  }
+
+  if (room) {
+    return venuePreviewUrl(venue.id, room.title);
+  }
+
+  return venueInsideUrl(venue.id);
+};
```
0xdevalias, line 167: This may have counting related impacts.

0xdevalias (approved):
A couple of musing comments/thoughts/etc, all NB. If you're happy with this from a clickthrough/functionality perspective then LGTM!

src/utils/types.ts
thecadams: It ended up in here because it's related to type safety - it is more versatile than just a filter IMO. I'll leave it here for now and we can see if we end up needing it for anything else - that might guide where its final home ends up being.

src/components/organisms/Badges/Badges.tsx
thecadams: Hm, I tried to take away the `Fragment` and turns out we have 2 at the top level of this component: `<div className="visits"` and `<div className="badges-container"` so it does need to stay. There's an autofix linter in eslint that'll detect these, so we can & should use that when we're ready.

src/components/organisms/Badges/Badges.tsx
thecadams: Added a `@debt` comment to note this.

thecadams:
[Bring Back Badges](https://trello.com/c/AAmGbZVs/447-bring-back-badges)

PR #898, opened 2020-12-04
Ability to toggle reactions container on-off on auditorium…

IvanBuljovcic:


src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -603,6 +604,23 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
           </div>
         )}
 
+        {templateID && HAS_REACTIONS_TEMPLATES.includes(templateID) && (
+          <div className="toggle-room">
```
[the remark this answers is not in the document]
IvanBuljovcic: Refactored in Admin v2 (self-serve), this specific component will be removed with the updates in that

functions/venue.js
```diff
@@ -443,6 +448,10 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.showRangers = data.showRangers;
   }
 
+  if (typeof data.showReactions === "boolean") {
+    updated.showReactions = data.showReactions;
+  }
```
thecadams, line 453: You could do:

```
if (data.showReactions) {
  updated.showReactions = isTruthy(data.showReactions);
}
```

thecadams (approved):
LGTM.

functions/venue.js
IvanBuljovcic: This approach works only if `data.showReactions` is true, if it is false, the `updated` will not get the value

PR #901, opened 2020-12-04
Auditorium editable seats by column and row

IvanBuljovcic:
Ability to set/edit seat column and row number when creating or editing a Auditorium venue


denisdimitrov (approved):
LGTM!

PR #906, opened 2020-12-04
Disable utils/url isValidUrl check in openUrl

0xdevalias:
This check seems to be failing for some zoom links for some reason. While we debug/figure that out, we'll allow links to be opened as per the old logic without this check.

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -43,12 +43,16 @@ export const openRoomUrl = (url: string) => {
 export const openUrl = (url: string) => {
   if (!isValidUrl(url)) {
     Bugsnag.notify(
-      new Error(`Invalid URL ${url} on page ${window.location.href}; ignoring`),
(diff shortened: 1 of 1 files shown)


src/utils/url.ts
```diff
@@ -43,12 +43,16 @@ export const openRoomUrl = (url: string) => {
...
+      new Error(
+        `Invalid URL ${url} on page ${window.location.href}; allowing for now (workaround)`
+      ),
       (event) => {
         event.addMetadata("utils/url::openUrl", { url });
       }
     );
-    return;
+    // @debt keep the checking in place so we can debug further, but don't block attempts to open
+    // return;
```
0xdevalias, line 55: I'm explicitly choosing to keep this commented code here as this should be a temporary workaround, after which it should be able to be restored (+whatever the bug fix is)

thecadams (approved):

PR #903, opened 2020-12-04
remove join event button when secret form is shown

denisdimitrov:

PR #908, opened 2020-12-04
Revert "remove join event button when secret form is shown (#903)"

0xdevalias:
This reverts commit d5b664dd63da7e71b1be7eaa4fc783c1a78d769b.

This has removed the button for all venues. Reverting to buy time to resolve the underlying logic/issue/etc

### DIFF
--- a/src/pages/VenueLandingPage/VenueLandingPage.tsx
+++ b/src/pages/VenueLandingPage/VenueLandingPage.tsx
@@ -29,7 +29,6 @@ import {
   userPurchaseHistorySelector,
 } from "utils/selectors";
 import { IFRAME_ALLOW } from "settings";
-import { isTruthy } from "utils/types";
 
 export interface VenueLandingPageProps {
   venue: Firestore["data"]["currentVenue"];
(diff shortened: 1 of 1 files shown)


0xdevalias:
@sparkletown/tech-leads 🔨 🍷  breaking glass to deploy urgent production fix.

thecadams:
@denisdimitrov @0xdevalias (not urgent but just following up) looking at this, we deployed #903 and reverted it here then deployed in #909 ; is there any context on what led to the revert and is there a plan to redo this one?

0xdevalias:
@thecadams context in https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1607112856169300 , but essentially this change was causing the join button to not be visible at all on non-passworded events, which obviously was a 🔥 breaking production for all of our events.

Definitely need to resolve the bug in this + re-instate this. I haven't had the ⌛ as yet. Might make the most sense for @denisdimitrov to when he comes online. But we need to make sure we thoroughly review/test this to ensure we aren't breaking production again when we release it.

PR #910, opened 2020-12-04
16:9 support in auditorium

thecadams:


thecadams:
The commits are all messed up here; going to force push.

src/components/templates/ArtPiece/ArtPiece.scss
```diff
@@ -9,9 +9,25 @@
...
+    overflow: hidden;
+    position: relative;
+
+    // Default aspect ratio 2.39:1
+    padding-top: (1/2.39) * 100%;
+
+    &.aspect-16-9 {
+      padding-top: (9/16) * 100%;
+    }
+  }
```
0xdevalias, line 23: [DRY] we could extract these calculations to variables/mixins/etc and share them between scss rather than duplication

src/components/templates/Audience/Audience.scss
```diff
@@ -35,6 +35,11 @@ $light: #ffffff;
         border-radius: 15px;
         box-shadow: 2px 2px 20px $light;
       }
+
+      .aspect-16-9 {
+        min-height: 112.5%;
+        margin-top: -6.5%;
+      }
```
0xdevalias, line 42: [DRY] see previous comment

0xdevalias (approved):
[CRO] Code changes look good, haven't clicked through to validate that it solves the desired outcome, but if you have and are happy with it, LGTM!

0xdevalias:
Merging so we can get it deployed, cleanup can be done in a future PR if desired.

src/components/templates/ArtPiece/ArtPiece.scss
thecadams: Oh I like the mixins idea; my one thing that'd be good to preserve is the per-component 1:1 JSX/SCSS relationship (via container element + css class); and a mixin provides a great method to achieve that while allowing components to pull from a shared library of utils inside their container div.

Perhaps we'd build something akin to what `src/utils` lets us do, but for mixins.

Alternatively we could refactor all our `iframe` elements into an embeddable video component, which would allow all sorts of possibilities, such as deep support for Twitch/Vimeo/YouTube/etc. That would really help in situations like embedding Twitch, which is quite difficult to get right.

thecadams (commented):
Thanks, good callout on doing video aspect in a reusable way. iframe embeds are a core part of the platform and really deserve their own component, so it's great to see the code telling us that.

src/components/templates/ArtPiece/ArtPiece.scss
0xdevalias: I like the idea of encapsulating our 'special cases' (eg. `iframe` with certain tags/etc) into a reusable component, driven by props.

I agree of keeping a 1:1 component -> localises `.scss` file for it's styles + extracting common helpers/mixins/etc into a shared 'utils' type section.

We currently have `src/scss` which is where `constants.scss` lives. I really want to 💥  the `global.scss` file from there as soon as reasonably possible. But in general, this seems like a useful place to locate these common helpers. 

These common mixins filenames should be prefixed with `_` (eg. `_foo.scss`) as this is the convention for Sass Partials, which we can then use various 'at rules' with:

- https://sass-lang.com/guide#topic-4
- https://sass-lang.com/documentation/at-rules/use
- https://sass-lang.com/documentation/at-rules/mixin
- https://sass-lang.com/documentation/at-rules/extend
- https://sass-lang.com/documentation/at-rules/function
- etc

0xdevalias: ^backlogged the above in https://github.com/sparkletown/internal-sparkle-issues/issues/4

PR #913, opened 2020-12-05
Ensure that backend errors on LoginForm / RegisterForm get cleared when form data changes

0xdevalias:
Fixes https://trello.com/c/23nBR4Af

The issue here is that we were setting the backend errors as a 'field level error', but it was only mapped to the `email` field. So if you got a backend error (eg. a password related error), and didn't change the `email` field (eg. because it was a password related error), then the backend error would never get cleared, and you would be unable to submit the form.

This would have blocked user signup/signin if they didn't try refreshing the page.

Fixed by treating it as a full custom error, and explicitly clearing it with `clearErrors` when any field in the form changes:

- https://react-hook-form.com/api#clearErrors

Note: this appears to have existed as a bug since the backend/'firebase' error type was introduced, so it's not a new issue:

- https://github.com/sparkletown/sparkle/commit/b0477d14295b923d494cfff93da2c8b422f43524#diff-738b3ef5cad64ac92b6a7a05362e16b8b486d8e0ce37ce31a7f43558389f0e3dR37

### DIFF
--- a/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
+++ b/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
@@ -23,6 +23,7 @@ interface RegisterFormData {
   password: string;
   code: string;
   date_of_birth: string;
+  backend?: string;
 }
 
 export interface CodeOfConductQuestion {
@@ -49,11 +50,20 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
     return firebase.auth().createUserWithEmailAndPassword(email, password);
   };
 
-  const { register, handleSubmit, errors, formState, setError } = useForm<
-    RegisterFormData
-  >({
+  const {
+    register,
+    handleSubmit,
+    errors,
+    formState,
+    setError,
+    clearError,
+  } = useForm<RegisterFormData>({
     mode: "onChange",
   });
+
+  const clearBackendErrors = () => {
+    clearError("backend");
+  };
 
   if (!venue) {
     return <>Loading...</>;
@@ -64,6 +74,7 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
       if (venue.requiresTicketCode) await axios.get(codeCheckUrl(data.code));
       if (venue.requiresEmailVerification)
         await axios.get(codeCheckUrl(data.email));
+
       const auth = await signUp(data);
       if (
         auth.user &&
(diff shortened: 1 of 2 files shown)


0xdevalias:
@sparkletown/tech-leads 🔨 🍷 breaking glass to deploy as trello card has this marked as an urgent issue to be fixed.

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -133,36 +155,46 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
             placeholder="Password"
             ref={register({
               required: true,
-              // minLength: 8,
-              pattern: /^(?=.*[0-9])(?=.*[a-zA-Z]).{2,}$/,
+              pattern: /^(?=.*[0-9])(?=.*[a-zA-Z]).{6,}$/,
```
0xdevalias, line 158: Note: this was only checking for length 2, whereas the backend was checking for length 6. The discrepancy between frontend/backend validation rules was what lead to the issue as described in the original ticket showing up as it did, though the main bug that was fixed is as described in the PR description.

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -133,36 +155,46 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
...
           <span
             className={`input-${
               errors.password && errors.password.type === "pattern"
                 ? "error"
                 : "info"
             }`}
           >
-            Password must contain letters and numbers
+            Password must contain letters, numbers, and be at least 6 characters
+            long
```
0xdevalias, line 170: Note: this was only checking for length 2, whereas the backend was checking for length 6. The discrepancy between frontend/backend validation rules was what lead to the issue as described in the original ticket showing up as it did, though the main bug that was fixed is as described in the PR description.

thecadams: Yep, spotted this in review, appreciate the change to 6 characters.

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -133,36 +155,46 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
...
         )}
+
         {venue.requiresDateOfBirth && (
           <DateOfBirthField register={register} error={errors?.date_of_birth} />
         )}
+
+        {errors.backend && (
+          <span className="input-error">{errors.backend.message}</span>
+        )}
+
```
thecadams, line 189: Great that this is right above the button which becomes unclickable; and great that the `onChange` event is used to clear backend errors.

thecadams (commented):
Overall a good solid change, and we can work with the team on what constitutes a 🔥 worthy of an emergency fix like this.

PR #919, opened 2020-12-07
open same host urls in the same tab

denisdimitrov:
`isExternalUrl` is now using different logic and that's why I'm using it, oppose to what the ticket suggested (not to use it). There is no need for http or https check in it. We just need to know if the url is external or not.
The function is not being used anywhere else that's why this change is safe.


src/utils/url.ts
```diff
@@ -29,8 +29,7 @@ export const venueRoomUrl = (venue: WithId<AnyVenue>, roomTitle: string) => {
 };
 
 export const isExternalUrl = (url: string) =>
-  url.includes("http") &&
-  window.location.host !== new URL(getRoomUrl(url)).host;
+  window.location.host !== new URL(url).host;
```
0xdevalias, line 32: If `new URL()` fails to parse it will throw an exception. There's some related commented out code for this in this file because we found it was breaking for some reason.. I think for this function though, wrapping it in a `try`/`catch`, and then 'falling back' to a sane default if we can't parse the URL might be a good idea?

My gut feel is that default would be opening in the same window (in case it's a relative URL, etc).. and maybe using `Bugsnag.notify` to give us that information as well (so that we are capturing data to know what is broken, and ensure we come back to validate/fix it)

- https://docs.bugsnag.com/platforms/javascript/reporting-handled-errors/

You can customize the event severity when using this, so since we should fall back to a 'sane default', I think it would be ok to drop the severity for these reported errors to warn or even info:

- https://docs.bugsnag.com/platforms/javascript/reporting-handled-errors/#customizing-diagnostic-data

[OOS/Future] The 'proper' higher level fix is to improve our validation wherever users can enter URLs to ensure that it can only ever be correct, + updating/sanitising all the existing DB data as required to ensure that aligns with our newly chose validation rules as well.

src/utils/url.ts
```diff
@@ -55,7 +54,11 @@ export const openUrl = (url: string) => {
     // return;
   }
 
-  window.open(url, "_blank", "noopener,noreferrer");
+  if (isExternalUrl(url)) {
+    window.open(url, "_blank", "noopener,noreferrer");
+  } else {
+    window.location.href = url;
+  }
```
0xdevalias, line 61: Can/should we be making use of our routing navigation helpers here? I believe this code as is will force a full window refresh, whereas using the routing helpers will do it within the same 'page load'.

Ideally we would be using the `useHistory` hook, but we're not in a react component context here, so we can't use hooks:

- https://reactrouter.com/web/api/Hooks/usehistory

So I _think_ we can just use `history` directly.. but it seems to come with extra caveats/warnings that i'm not immediately sure of the ramifications of:

- https://reactrouter.com/web/api/history

I think you may want to use `history.push` here, but it would be worth doing your own quick research/testing to validate that:

> `push(path, [state])` - (function) Pushes a new entry onto the history stack

0xdevalias: If this ends up being too complex, or like it will be a rabbithole of unknown length, feel free to add a `@debt` comment such as `@debt can we use react-router's navigation helpers here? should we?` and then leave this implementation as is (eg. `window.location.href = url`)

0xdevalias (changes requested):
Left a few comments, the `try`/`catch` is the important one, the other is a nice to have.

0xdevalias:
@denisdimitrov if I'm not still online once you've made the requested changes here, feel free to dismiss my stake review + get someone else to give it a quick check over + approval. My main interest is to ensure this goes out with stability/doesn't cause any production issues, but that doesn't mean it needs to block on my final stamp of approval if I'm not available. I can retroactively review if needed. Also, please remember to click the 're-request review' button when it's ready.

src/utils/url.ts
denisdimitrov: On one side I understand your point and why we should use history/router, on another I feel like using it outside component logic becomes complicated and harder to trace and I haven't seen it anywhere tbh. Importing history directly would feel like importing redux store directly. I'll leave it as it is for now.

src/utils/url.ts
denisdimitrov: Try/catch block has been added with a simple Bugsnag.notify(error) and severity `info`
I'm not sure if this is enough, you seem to be more experienced with this tool, if you have any other suggestions for the handling of the error, they are welcome.

src/utils/url.ts
0xdevalias: I'm happy with leaving it for now.

The reasoning behind it is essentially as a single page application we don't actually want to be navigating to different pages, as what we are really doing is 'faking' route changes within the router/etc itself. That's where using the router helpers comes into play. I expect we haven't done that elsewhere in the app more out of 'tech debt' reasons than 'we explicitly chose to do it this way' reasons.

src/utils/url.ts
```diff
@@ -28,9 +28,19 @@ export const venueRoomUrl = (venue: WithId<AnyVenue>, roomTitle: string) => {
...
 };
 
-export const isExternalUrl = (url: string) =>
-  url.includes("http") &&
-  window.location.host !== new URL(getRoomUrl(url)).host;
+export const isExternalUrl = (url: string) => {
+  try {
+    const urlHost = new URL(url).host;
+    const currentHost = window.location.host;
+    return currentHost !== urlHost;
```
0xdevalias, line 35: [NB] you still could have left this as a single line if you wanted, I think it's simple enough that breaking it out doesn't add much extra clarity.
```suggestion
    return new URL(url).host !== window.location.host;
```

src/utils/url.ts
```diff
@@ -28,9 +28,19 @@ export const venueRoomUrl = (venue: WithId<AnyVenue>, roomTitle: string) => {
...
+  } catch (error) {
+    Bugsnag.notify(new Error(error), (event) => {
+      event.severity = "info";
+      event.addMetadata("utils::url::isExternalUrl", { url });
+    });
+    return false;
+  }
```
0xdevalias, line 42: The Bugsnag part here looks good to me!

0xdevalias (approved):
The only comment I left is a stylistic/fyi, non-blocking. Aside from that, this looks good!

Assume you've clicked through and are happy that this does what it intends to do (which it looks like it should from the code), LGTM!

src/utils/url.ts
0xdevalias: Looks good to me. I think @thecadams mentioned on one of my similar PR's the other day that where possible we should check the type of error, and only catch the ones we are explicitly caring about, letting the others get thrown higher again. So in this case, that would be catching URL parsing errors, and otherwise throwing them higher.

I think that was doing something like:

```
catch (e) {
  if (e.name === 'TypeError') {
    // bugsnag.notify
    // return false
  }
 throw e; // Error wasn't one we wanted to handle, let it bubble up higher.
}
```

src/utils/url.ts
0xdevalias: Added a note on https://github.com/sparkletown/sparkle/pull/919#discussion_r537446605 as to how we could be a bit more explicit in catching just the error we're expecting if you wanted to make this even more robust.

PR #921, opened 2020-12-07
Private chat messages showing in correct order for both participants

IvanBuljovcic:
* When a user sends a message, only the recipient saw the message, not the sender.
* All messages were in random order, now sorted by timestamp **ts_utc** in reverse (last message first)


src/components/molecules/ChatList/ChatList.tsx
```diff
@@ -84,20 +87,26 @@ const ChatList: React.FC<ChatListProps> = ({
   const hideDeleteModal = useCallback(() => setShowDeleteModal(false), []);
 
   const hasMessages = hasElements(messages);
+
   const messageSender = usersById?.[messageToDelete?.from ?? ""]?.partyName;
+
+  // Last (newest) message goes first
+  const sortedMessages = sortBy(messages, ["ts_utc"]).reverse();
```
0xdevalias, line 94: Probably want to useMemo this

src/components/molecules/ChatList/ChatList.tsx
```diff
@@ -84,20 +87,26 @@ const ChatList: React.FC<ChatListProps> = ({
...
                 key={`chat-message-${index}`}
                 usersById={usersById}
                 message={message}
                 allowDelete={allowDelete ?? false}
                 onDeleteClick={toggleDeleteModal}
                 onAvatarClick={showUserProfile}
+                showSenderImage={showSenderImage}
               />
             ))}
         </div>
```
0xdevalias, line 112: Extract this from the JSX + `useMemo`. If it's the only place `sortedMessages` is needed/used, then you could probably combine this + my above comment within the same `useMemo` block.

src/components/molecules/ChatList/ChatList.tsx
```diff
@@ -84,20 +87,26 @@ const ChatList: React.FC<ChatListProps> = ({
...
+
   return (
     <>
       {hasMessages && (
         <div className="chat-messages-container">
           {usersById &&
-            messages.map((message, index) => (
+            sortedMessages.map((message, index) => (
               <ChatMessage
                 key={`chat-message-${index}`}
```
0xdevalias, line 103: Is there something more unique/good that we can use here in place of index? Maybe the `ts_utc` + who sent it or similar if messages don't have an actual UID?

src/components/molecules/ChatList/ChatMessage.tsx
```diff
@@ -55,33 +50,66 @@ export const ChatMessage: FC<ChatMessageProps> = ({
     onDeleteClick(message);
   }, [message, onDeleteClick]);
 
-  return message.from in usersById ? (
+  if (!user || !profile) return null;
+
+  const sender = { ...usersById[message.from], id: message.from };
+  const isMe = message.from === user?.uid;
```
0xdevalias, line 56: [nit] Since we check `!user` above, do we need to null check `user?.uid` here?
```suggestion
  const isMe = message.from === user.uid;
```

src/components/molecules/ChatList/ChatMessage.tsx
```diff
@@ -55,33 +50,66 @@ export const ChatMessage: FC<ChatMessageProps> = ({
...
+              src={getProfileImage()}
+              title={getProfileName()}
+              alt={`${getProfileName()} profile`}
+              width={PROFILE_IMAGE_SIZE}
+              height={PROFILE_IMAGE_SIZE}
+            />
+            <div className="chat-message-pseudo">
+              {getProfileName()}{" "}
+              <span className="timestamp">
+                {formatUtcSeconds(message.ts_utc)}
```
0xdevalias, line 104: This will cause a type error once https://github.com/sparkletown/sparkle/pull/897 gets merged.
```suggestion
                {formatUtcSeconds(message.ts_utc.seconds)}
```

0xdevalias (changes requested):
Some `useMemo` related changes + minor comments. Once those are addressed, re-request review, but if I am not around, feel free to dismiss my stale review + get @denisdimitrov or similar to give the approval stamp.

src/components/molecules/ChatList/ChatList.tsx
```diff
@@ -84,22 +87,38 @@ const ChatList: React.FC<ChatListProps> = ({
...
+
   const messageSender = usersById?.[messageToDelete?.from ?? ""]?.partyName;
+
+  const renderMessages = useMemo(() => {
+    // Last (newest) message goes first
+    const sortedMessages = sortBy(messages, ["ts_utc"]).reverse();
+
+    return sortedMessages.map((message, index) => (
+      <ChatMessage
+        key={`${message.from}-${message.ts_utc.seconds}`}
```
denisdimitrov, line 99: If someone sends 2 messages in less than a second, does that mean that there will be duplicated keys?

0xdevalias: I believe it will. Do we have milliseconds available on that object?

src/components/molecules/ChatList/ChatList.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/921/commits/f69bd1a816115991a5ca4257830f370237e55697

src/components/molecules/ChatList/ChatList.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/921/commits/f69bd1a816115991a5ca4257830f370237e55697

src/components/molecules/ChatList/ChatList.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/921/commits/f69bd1a816115991a5ca4257830f370237e55697 though see also https://github.com/sparkletown/sparkle/pull/921#discussion_r537551003

src/components/molecules/ChatList/ChatMessage.tsx
0xdevalias: https://github.com/sparkletown/sparkle/pull/921/commits/f69bd1a816115991a5ca4257830f370237e55697

src/components/molecules/ChatList/ChatMessage.tsx
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/921/commits/f69bd1a816115991a5ca4257830f370237e55697

0xdevalias (commented):
Thanks for addressing the issues in my last review.

I'll hold off on approving this as due to @denisdimitrov 's comment in https://github.com/sparkletown/sparkle/pull/921#pullrequestreview-546202951, but happy for him to review/approve once he's happy there.

thecadams (approved):

thecadams:
@0xdevalias I'm going to consider your review as having been completed as I've addressed the concern about duplicate keys by adding a reference to `nanoseconds`. Going to dismiss your changes request and merge. Thanks for the review.

PR #924, opened 2020-12-07
Deploy

0xdevalias:
Deploys: 

- Tweak table upload script (#911)
- add useInterval hook + related refactors/cleanup (including some room counting things) (#890)
- Restore badges (#895)
- Support custom tables in conversation space (#922)
- Fixes for times (including properly typed now), room counting, DRY code, etc (#897)
- open same host urls in the same tab (#919)
- sign up tweaks and layout update (#885)
- Conversation space participants changed to grid layout (#920)
- Private chat messages showing in correct order for both participants (#921)

PR #927, opened 2020-12-08
Deploy

0xdevalias:
Deploys:

- Add lint check to precommit hook (#857)
- Update config.yml (#894)


0xdevalias:
See https://github.com/sparkletown/sparkle/pull/924#issuecomment-740481264 for context on why this deploy was required.

thecadams:
Wow, good catch.

PR #928, opened 2020-12-08
Radio popover enabled from venue config

IvanBuljovcic:
* Radio popover enabled from venue config => `venue.playRadio`
* Radio autoplay only if navbar state is `true`
* State set to `true` on button click


src/hooks/useRadio.ts
```diff
@@ -20,15 +21,22 @@ export const useRadio = (audio?: HTMLAudioElement) => {
...
-  }, [volume, audio]);
+  }, [volume, audio, playRadio]);
 
   useEffect(() => {
-    audio?.play();
-  }, [audio]);
+    if (playRadio) {
+      audio?.play();
+    }
+  }, [audio, playRadio]);
```
denisdimitrov, line 39: This second `useEffect` seems to be using only a part of the logic that's in the one above. The dependency arrays are the same. Is this even required?

0xdevalias: Looks like it's redundant from a quick glance? Let's get this cleared up I reckon.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -221,14 +211,20 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
                     placement="bottom-end"
                     overlay={
-                      <RadioPopover
-                        {...{ volume, setVolume, title: venue?.radioTitle }}
-                      />
+                      <Popover id="radio-popover">
+                        <Popover.Content>
+                          <RadioModal
+                            {...{ volume, setVolume, title: venue?.radioTitle }}
+                            onEnableHandler={() => setIsRadioPlaying(true)}
```
0xdevalias, line 223: this should be have `useCallback`

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -69,14 +66,6 @@ const GiftPopover = (
...
 );
 
-const RadioPopover: React.FC<RadioModalPropsType> = (props) => (
-  <Popover id="radio-popover">
-    <Popover.Content>
-      <RadioModal {...props} />
-    </Popover.Content>
-  </Popover>
-);
-
```
0xdevalias, line 79: Not sure why we removed this in favour of inlining it?

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -1,20 +1,38 @@
...
+  onEnableHandler,
+  isRadioPlaying,
 }) => {
-  return (
-    <div className="radio-modal-container">
+  const handleEnableButtonClick = () => {
+    onEnableHandler();
+    // Volume is randomized because if it is set to the same number as before
+    // the radio will not start playing
+    const volume = 20 + Math.random(); // i.e. 20 + 0.3
```
0xdevalias, line 25: That... seems strange. Do we know why that is?

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -1,20 +1,38 @@
...
+
+    setVolume(volume);
+  };
+
+  const renderEnableRadioButton = () => (
+    <Button onClick={() => handleEnableButtonClick()}>Enable radio</Button>
+  );
```
0xdevalias, line 32: `useCallback`

src/components/organisms/RadioModal/RadioModal.tsx
IvanBuljovcic: Leftover hack from previous integration if the volume was "set" to the previous value, the radio would not play.
Removing in new commit

src/components/molecules/NavBar/NavBar.tsx
IvanBuljovcic: For whatever reason, it didnt receive the correct props it needed to be rendered connected to the correct OverlayTrigger component.

With it like this, it rendered randomly on the top left of the screen instead of bellow the radio button on the navbar.

The perfect solution should and, hopefully soon, will be to pull that code out and make it a standalone component

0xdevalias: No worries

src/components/organisms/RadioModal/RadioModal.tsx
0xdevalias: @IvanBuljovcic this still needs doing by the looks of it

0xdevalias: Or alternately it may just need:

```suggestion
  const renderEnableRadioButton = () => (
    <Button onClick={handleEnableButtonClick}>Enable radio</Button>
  );
```

0xdevalias: Though at that point, I wonder what the purpose of `renderEnableRadioButton` is as opposed to just using the component directly in the JSX?

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -1,20 +1,34 @@
...
+
+  const renderRadioBody = () => (
+    <>
       <div className="title-radio">{title ?? `${PLAYA_VENUE_NAME} Radio`}</div>
       <div className="text-radio">
         We recommend turning on the global burner radio station as you rove
```
0xdevalias, line 34: Probably want to change this copy to remove any burner/playa/etc references

0xdevalias (changes requested):
Just a couple of little things left by the looks of it.

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -34,6 +48,14 @@ export const RadioModal: React.FunctionComponent<RadioModalPropsType> = ({
...
         value={volume}
       />
+    </>
+  );
+
+  return (
+    <div className="radio-modal-container">
+      {!isRadioPlaying && renderEnableRadioButton()}
+
+      {isRadioPlaying && renderRadioBody()}
```
thecadams, line 58: Consider ternary:

```suggestion
      {isRadioPlaying ? renderRadioBody() : renderEnableRadioButton()}
```

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -1,20 +1,34 @@
...
+    setVolume(volume);
+  }, [onEnableHandler, setVolume, volume]);
+
+  const renderEnableRadioButton = () => (
+    <Button onClick={() => handleEnableButtonClick()}>Enable radio</Button>
+  );
+
+  const renderRadioBody = () => (
+    <>
       <div className="title-radio">{title ?? `${PLAYA_VENUE_NAME} Radio`}</div>
```
thecadams, line 32:
```suggestion
      <div className="title-radio">{title ?? "Radio"}</div>
```

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -1,20 +1,34 @@
-import React from "react";
+import React, { useCallback } from "react";
+import { Button } from "react-bootstrap";
 import { PLAYA_VENUE_NAME } from "settings";
 import "./RadioModal.scss";
```
thecadams, line 4: Would be good to remove playa from radio entirely:

```suggestion
import "./RadioModal.scss";
```

src/components/organisms/RadioModal/RadioModal.tsx
thecadams, line 34: Also need to delete the next line, which I don't seem to be able to suggest for you.
```suggestion
        We recommend turning on the radio as you explore the map!
```

thecadams (commented):
Looking good, getting close. Only a few little things left!

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -99,17 +119,28 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
 
   const [isRadioPlaying, setIsRadioPlaying] = useState(false);
   const { volume, setVolume } = useRadio(isRadioPlaying, sound);
 
+  const soundcloudIframeRef = useRef<HTMLIFrameElement>(null);
+
+  // Append script to <body>
+  useEffect(() => {
+    loadScript("https://w.soundcloud.com/player/api.js");
+  }, []);
```
0xdevalias, line 142: Should we only do this when `isSoundCloud`?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -1,4 +1,10 @@
-import React, { useState, useMemo, useRef, useCallback } from "react";
+import React, {
```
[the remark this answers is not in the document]
IvanBuljovcic: I love the optimism of codeclimate.

Definitely need a refactor, not a viable case for this PR

src/components/molecules/NavBar/NavBar.tsx
IvanBuljovcic: Good point

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -69,13 +74,17 @@ const GiftPopover = (
...
-  <Popover id="radio-popover">
-    <Popover.Content>
-      <RadioModal {...props} />
-    </Popover.Content>
-  </Popover>
-);
+// TODO: Get rid of this as soon as humanly possible.
+// TODO: Look into properly setting up SoundClouds' API,
+// possibly without the need to append script
+const loadScript = (src: string) => {
```
0xdevalias, line 80: This is backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/62

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -110,15 +119,29 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
 
   const hasUpcomingEvents = futureUpcoming && futureUpcoming.length > 0;
 
+  const hasRadioStations = radioStations && radioStations.length;
+  const isSoundCloud =
+    !!hasRadioStations && RegExp("soundcloud").test(radioStations![0]);
```
0xdevalias, line 124: We could do this like `radioStations?.[0]?.includes('soundcloud')` or similar I think. A more robust way would be to parse the URL and check the domain host/etc. But for now, this is fine.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -156,13 +179,48 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+
   if (!venueId || !venue) return null;
 
   // TODO: ideally this would find the top most parent of parents and use those details
   const navbarTitle = parentVenue?.name ?? venue.name;
 
   const profileImage = profile?.pictureUrl || DEFAULT_PROFILE_IMAGE;
 
+  const renderSoundCloudIframe = () => {
+    const radioStation = !!hasRadioStations && radioStations![0];
```
0xdevalias, line 194: The `!!` of this should probably be done where it is set. And could use our `isTruthy` semantic wrapper.

0xdevalias: I see a lot of `!` usages, can this actually never be `undefined`? Otherwise it may just be avoiding our typechecking that could lead to errors at runtime.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -156,13 +179,48 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+
+    return (
+      <iframe
+        title="venueRadio"
+        ref={soundcloudIframeRef}
+        id="sound-cloud-player"
```
0xdevalias, line 200: Do we need to use `id` on this? I generally avoid using it unless we have an explicit need for it.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -156,13 +179,48 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+        src={`https://w.soundcloud.com/player/?url=${radioStation}&amp;start_track=0&amp;single_active=true&amp;show_artwork=false`}
+      />
+    );
+  };
+
+  const renderRadioTrigger = () => (
+    <S.RadioTrigger>
+      <div
+        className={`profile-icon navbar-link-radio ${volume === 0 && "off"}`}
+        onClick={() => setShowRadioPopover((prevState) => !prevState)}
```
0xdevalias, line 212: `useCallback`?

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -34,6 +42,16 @@ export const RadioModal: React.FunctionComponent<RadioModalPropsType> = ({
         onChange={(ev) => setVolume(Number(ev.target.value))}
         value={volume}
       />
+    </>
+  );
+
+  return (
+    <div className="radio-modal-container">
+      {!isRadioPlaying ? (
+        <Button onClick={() => handleEnableButtonClick()}>Enable radio</Button>
```
0xdevalias, line 51: `useCallback`?

0xdevalias (changes requested):
Requesting changes just for the `useCallback`/etc memoisations. The stylistic/etc things are NB

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -110,15 +119,29 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
+  const { volume, setVolume } = useRadio(isRadioPlaying, sound);
+
+  const soundcloudIframeRef = useRef<HTMLIFrameElement>(null);
+
+  // Append script to <body>
+  useEffect(() => {
+    if (isSoundCloud) {
+      loadScript("https://w.soundcloud.com/player/api.js");
+    }
+  }, [isSoundCloud]);
```
0xdevalias, line 144: I believe we only needed this when we wanted to control the playback, but I don't think we're actually using it anymore?

0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/928/commits/e8fd26ba30e9505036b0f1cda1afcec36d4ab4bf

0xdevalias (approved):

PR #929, opened 2020-12-08
🔥 Quickfix for seated area, padding botom added 🔥

IvanBuljovcic:


denisdimitrov (approved):
LGTM!

0xdevalias (commented):
[Future] I expect we probably want to be using `$footer-height` from `src/scss/constants.scss` to solve this nicely:

- https://github.com/sparkletown/sparkle/blob/staging/src/scss/constants.scss#L34 rather than setting the

PR #931, opened 2020-12-08
onboarding fixes and tweaks

denisdimitrov:


denisdimitrov:
Going to recreate this because the commit history is messed up and I shouldn't force push. It was also created so Ivan could `git checkout` it easier for the yesterday prod fixes.

0xdevalias:
@denisdimitrov sounds good. Just as a future note though, if it hasn't been reviewed yet, a force push probably isn't the end of the world. The main issue is when it is rebased+force pushed when we have half a review in progress/similar.

PR #932, opened 2020-12-08
🔥 Check for existance of profile before pushing url

IvanBuljovcic:

PR #934, opened 2020-12-08
Prevent loops in account creation

thecadams:


thecadams:
Going to 🔨 🍷  and merge this, we can review later.

src/pages/Account/Profile.tsx
```diff
@@ -44,8 +45,10 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
   const onSubmit = async (data: ProfileFormData) => {
     if (!user) return;
     await updateUserProfile(user.uid, data);
-    const accountQuestionsUrl = `/account/questions?venueId=${venueId}&returnUrl=${window.location.pathname}${window.location.search}`;
-    const nextUrl = venueId ? accountQuestionsUrl : `/${DEFAULT_VENUE}`;
```
thecadams, line 48: These are the core of the change. `returnUrl` should never take the user back to part of the account creation flow; that's a recipe for loops.

And the `DEFAULT_VENUE` is going to confusingly send _someone_ to the wrong venue at some point, so it needs to go.

src/pages/Account/Profile.tsx
```diff
@@ -44,8 +45,10 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+    const accountQuestionsUrl = `/account/questions?venueId=${venueId}${
+      returnUrl ? "&returnUrl=" + returnUrl : ""
+    }`;
```
thecadams, line 50: Bringing `returnUrl` along through the flow here.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -295,7 +295,7 @@ const VenuePage = () => {
     return <LoadingPage />;
   }
 
-  if (profile && !(profile?.partyName && profile?.pictureUrl)) {
+  if (profile && !isCompleteProfile(profile)) {
```
thecadams, line 298: `isCompleteProfile` is an attempt to make this code a little more self documenting.

src/hooks/useUser.ts
```diff
@@ -1,7 +1,14 @@
...
+import { FirebaseReducer } from "react-redux-firebase";
+import { User } from "types/User";
 
-export const useUser = () => {
+type UseUserResult = {
+  user: FirebaseReducer.AuthState | undefined;
+  profile: FirebaseReducer.Profile<User> | undefined;
+};
+
+export const useUser = (): UseUserResult => {
```
thecadams, line 11: `profile` was coming through this hook as `any` type. So this was overdue.

src/pages/Account/Profile.tsx
0xdevalias: [Future] We can use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams and similar to handle this in a better way than string concatenation.

src/pages/Account/Profile.tsx
```diff
@@ -44,8 +45,10 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+    const nextUrl = venueId ? accountQuestionsUrl : returnUrl?.toString() ?? "";
     history.push(IS_BURN ? `/enter/step3` : nextUrl);
```
0xdevalias, line 52: [Future] We should be stripping out all of this `IS_BURN` logic from this flow as well.

src/pages/VenuePage/VenuePage.tsx
0xdevalias: Excellent! This is what I wanted in my retroactive review of https://github.com/sparkletown/sparkle/pull/932#pullrequestreview-547762211

0xdevalias (commented):
Retroactive review, but this looks solid!

PR #935, opened 2020-12-08
Deploy - Prevent loops in account creation (#934)

thecadams:
* Prevent loops in account creation

* Pass the returnUrl

* More self-documenting code


0xdevalias:
@thecadams IMO we don't need the 🔨 🍷  label on this as we are always able to merge deploy PRs without approvals. So just having it on the original PR we bypassed (as you did) is sufficient.

thecadams:
> IMO

No prob, noted, will do!

PR #936, opened 2020-12-09
Prevent importing all of firebase and thus a console warning

thecadams:
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/48


0xdevalias (approved):
This looks good as an initial step. I will probably keep the backlog issue open even once this is done, as I feel like there are more aspects to it than are resolved here. But this is a good first step towards those goals!

PR #937, opened 2020-12-09
Fix error in flatMap occurring in some venues

thecadams:
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/8


0xdevalias (approved):
[CRO] changes look good here, can't see it breaking anything as is. Not sure how to test/validate the bug at the moment, so let's get it out to staging/etc and see if it resolves it!

denisdimitrov:
[Turning room and count on is broken](https://trello.com/c/G2oSiy3Q/553-turning-room-and-count-on-is-broken)

PR #938, opened 2020-12-09
shrink video when audience aspect ratio is 16:9

denisdimitrov:
Not sure if this is the fix we want. Instead of enlarging the video and make it hide the top seats, now it shrinks from the sides (width only) and doesn't collide with the seats.

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -37,8 +37,8 @@ $light: #ffffff;
       }
 
       .aspect-16-9 {
-        min-height: 112.5%;
-        margin-top: -6.5%;
+        max-width: 88.5%;
+        margin-left: 6.5%;
       }
     }
 
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -309,8 +309,6 @@ export const Audience: React.FunctionComponent = () => {
       venue.videoAspect === VideoAspectRatio.SixteenNine ? "aspect-16-9" : ""
     }`;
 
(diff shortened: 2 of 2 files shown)


denisdimitrov:
[Lots of people having issues logging in](https://trello.com/c/jQD4RiUw/536-lots-of-people-having-issues-logging-in)

0xdevalias:
@denisdimitrov is this a change only required in this 1 file? I believe the original PR touched a couple of different scss files to implement 16:9 Do we need to make fixes in all of those locations?

denisdimitrov:
@0xdevalias The task is specifically for `audience` template, the other change is in Artpiece but it's using padding and seems to be working fine.

0xdevalias:
@denisdimitrov great. Thanks for the explanation. I'm guessing you've played around with this change, resized things, etc and it seems to be working ok?

denisdimitrov:
[Auditorium video is bleeding into the seats (EVENT: 12/10@4:45am AEDT)](https://trello.com/c/e0j4kMxb/532-auditorium-video-is-bleeding-into-the-seats-event-12-10445am-aedt)

denisdimitrov:
@0xdevalias I found a 16:9 video in one of the production urls and this was looking good and responsive. It also works good when the "16:9" is missing from the firebase, the video is just slightly smaller.

0xdevalias (approved):
As per @denisdimitrov's comments in https://github.com/sparkletown/sparkle/pull/938#issuecomment-741775896 it sounds like this is good to go!

No concerns with the code changes here. LGTM!

thecadams (commented):
👍

PR #939, opened 2020-12-09
onboarding tweaks and fixes

denisdimitrov:
- hover styles for buttons
- date of birth field validation
- date of birth description
- change button order and styles in confirmation modal
- remove upload of default avatars and use static images
- reduce size of default avatars
- allow login screen prompt after more than one errors
- add reValidation


thecadams (approved):
lgtm!

PR #940, opened 2020-12-09
Restore private chat overlay in NavBar when not currently in PartyMap venue

0xdevalias:
This PR restores the 'private chat in `NavBar`' overlay that was removed when we added the `SideBar` to the `PartyMap`, as any sub-venue that isn't the `PartyMap` doesn't have any access to private chat at the moment.

When `currentVenue.template === "partymap"`, the private chat won't be shown in the `NavBar` (as we will have the `SideBar` available)

It uses `ChatsList` (as per the `SideBar`) rather than the legacy `PrivateChatModal` implementation.


0xdevalias:
**Current State of this PR:**

image
image

This one seems to have some issues with profile pictures/etc at the moment (though pretty sure I saw it working before...)
image

----

This is showing correctly in the `SideBar` on the partymap template though (even though we're using the same underlying `ChatsList` component, need to figure out how/why they differ..)

image
image

0xdevalias:
@denisdimitrov @IvanBuljovcic happy for you either of you to take over this + push directly to the branch to get this finalised as per https://trello.com/c/2wJzU2mj

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -158,6 +181,8 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
 
   if (!venueId || !venue) return null;
 
+  const isVenueUsingPartyMap = venue.template === "partymap";
```
thecadams, line 184:
```suggestion
  const isVenueUsingPartyMap = venue.template === VenueTemplate.partymap;
```

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -43,11 +50,41 @@ const noopHandler = () => {};
...
+
+  const chatQuery = useCallback(() => {
+    if (!chatUsers || !Object.entries(chatUsers).length) {
+      if (hasElements(chatUserIds)) {
+        return chatUsersQuery;
+      }
+      return undefined;
+    }
+    return undefined;
+  }, [chatUserIds, chatUsers, chatUsersQuery]);
```
0xdevalias, line 84: Since `chatUsersQuery` seems to only be used within `chatQuery`, probably just inline it there at the `return` statement

0xdevalias: Also, what is the intent of `!Object.entries(chatUsers).length`?

0xdevalias: Also also, does this need to be a `useCallback`? I think `useMemo` would be sufficient, and would mean we don't have to call it below as a function.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -47,6 +47,7 @@ import { NavBarLogin } from "./NavBarLogin";
 
 import "./NavBar.scss";
 import "./playa.scss";
+import { VenueTemplate } from "types/VenueTemplate";
```
0xdevalias, line 50: [nit] sort this at the top with other types imports

0xdevalias (commented):
Recent changes look good. I can't approve this PR as I opened it originally, but I think it may be good to go.

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -43,11 +50,37 @@ const noopHandler = () => {};
...
+      if (hasElements(chatUserIds)) {
+        return {
+          collection: "users",
+          where: chatUsersOption,
+          storeAs: "chatUsers",
+        };
+      }
+      return undefined;
+    }
+    return undefined;
```
thecadams, line 79:
```suggestion
    if (!chatUsers && hasElements(chatUserIds)) {
      return {
        collection: "users",
        where: chatUsersOption,
        storeAs: "chatUsers",
      };
    }
    return undefined;
```

thecadams (approved):
Looks good, one nit.

0xdevalias:
FYI, there was a LOT of merge conflicts that had to be dealt with in https://github.com/sparkletown/sparkle/pull/940/commits/59f0f874ec34f057b79067665d36aadb3bfeb075 so it would be good to triple check this as a sanity check to make sure I didn't accidentally lose anything that was important.

PR #941, opened 2020-12-09
Show badges on profile modal

IvanBuljovcic:
* Venues being saved to redux, from the VenuePage.jsx
* Filtering for badges done by visited.id and venue.name (visited.id is venue.name)

### DIFF
--- a/src/components/organisms/Badges/Badges.tsx
+++ b/src/components/organisms/Badges/Badges.tsx
@@ -41,7 +41,7 @@ export const Badges: React.FC<{
     venueId: currentVenue.id,
     withEvents: false,
   });
-  const relatedVenueIds = relatedVenues.map((venue) => venue.id);
+  const relatedVenueIds = relatedVenues.map((venue) => venue.name);
 
   const visitHours = useMemo(() => {
     if (!visits) return 0;
@@ -55,7 +55,7 @@ export const Badges: React.FC<{
   // Only show visits to related venues
   const relevantVisits = visits?.filter(
     (visit) =>
-      currentVenue.id === visit.id || relatedVenueIds.includes(visit.id)
+      currentVenue.name === visit.id || relatedVenueIds.includes(visit.id)
   );
 
   const badges = useMemo(() => {
--- a/src/components/organisms/Badges/Badges.tsx
+++ b/src/components/organisms/Badges/Badges.tsx
@@ -34,6 +34,7 @@ export const Badges: React.FC<{
       storeAs: "userModalVisits",
     },
   ]);
+  useFirestoreConnect("venues");
   const visits = useSelector(userModalVisitsSelector);
   const venues = useSelector(orderedVenuesSelector);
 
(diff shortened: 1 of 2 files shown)


src/components/organisms/Badges/Badges.tsx
```diff
@@ -41,7 +41,7 @@ export const Badges: React.FC<{
     venueId: currentVenue.id,
     withEvents: false,
   });
-  const relatedVenueIds = relatedVenues.map((venue) => venue.id);
+  const relatedVenueIds = relatedVenues.map((venue) => venue.name);
```
0xdevalias, line 44:
```suggestion
  const relatedVenueNames = relatedVenues.map((venue) => venue.name);
```

src/components/organisms/Badges/Badges.tsx
```diff
@@ -55,7 +55,7 @@ export const Badges: React.FC<{
   // Only show visits to related venues
   const relevantVisits = visits?.filter(
     (visit) =>
-      currentVenue.id === visit.id || relatedVenueIds.includes(visit.id)
+      currentVenue.name === visit.id || relatedVenueIds.includes(visit.id)
```
0xdevalias, line 58:
```suggestion
      currentVenue.name === visit.id || relatedVenueNames.includes(visit.id)
```

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -239,6 +239,11 @@ const VenuePage = () => {
       : undefined
   );
 
+  useFirestoreConnect({
+    collection: "venues",
+    storeAs: "venues",
+  });
+
```
0xdevalias, line 246: What is this required here for? It's not immediately obvious from the rest of the changes in this PR.

IvanBuljovcic: `if (!relevantVisits || !venues) return [];` for one, and venues is always false so its always returning an empty array.

there is several (8 i think) places where we check for `!venues` and in all cases, its false because the key doesnt exist

0xdevalias: I guess more the point of the question was if this is replacing something that was deleted, where was it removed and why? And if it never existed, how did those features ever work (if they did)?

0xdevalias: And why add this in `VenuePage` as opposed to closer to where it is needed in `Badges`?

0xdevalias: Usages (on my local branch)

0xdevalias: `useFirestoreConnect("venues")` is the 'canonical' way of doing this, as there is no need for `storeAs` because it uses the name of the collection by default.

Which exists in a few places currently:
image

So I expect the fix we want is to just add `useFirestoreConnect("venues")` to `src/components/organisms/Badges/Badges.tsx`

0xdevalias (changes requested):
See above comments for changes

0xdevalias (approved):
[CRO] This looks good from the code. If you've tested that it works as expected and are happy with it, then LGTM!

PR #948, opened 2020-12-09
Fix additional things for performance with memoisation, etc

0xdevalias:


thecadams (approved):
Definitely an improvement, thanks!

PR #949, opened 2020-12-09
Guard against not yet loaded currentVenue

thecadams:
This was preventing login for me

### DIFF
--- a/src/components/molecules/LiveSchedule/LiveEvent.tsx
+++ b/src/components/molecules/LiveSchedule/LiveEvent.tsx
@@ -10,16 +10,20 @@ import { WithId } from "utils/id";
 import { useUser } from "hooks/useUser";
 
 interface LiveEventProps {
-  venue: WithId<Venue>;
-  event: VenueEvent;
+  venue?: WithId<Venue>;
+  event?: VenueEvent;
 }
 
 export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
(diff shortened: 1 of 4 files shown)


src/components/molecules/LiveSchedule/LiveEvent.tsx
```diff
@@ -10,17 +10,21 @@ import { WithId } from "utils/id";
...
   const { user, profile } = useUser();
 
   const enterLiveEvent = useCallback(() => {
-    openEventRoomWithCounting({ user, profile, venue, event });
+    if (venue && event) {
+      openEventRoomWithCounting({ user, profile, venue, event });
+    }
   }, [event, profile, user, venue]);
 
+  if (!event) return <></>;
```
0xdevalias, line 26: Returning `null` is the correct pattern here, fragment is only used to group things when you don't want an actual div to be emitted.

```suggestion
  if (!event) return null;
```

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -43,6 +43,8 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
     mode: "onChange",
   });
 
+  if (!venue) return <></>;
```
0xdevalias, line 46:
```suggestion
  if (!venue) return null;
```

0xdevalias (approved):

PR #954, opened 2020-12-10
add scripts/count-users-entered-venue

0xdevalias:
**Note:** this branch is not currently targeting `staging`, it should be merged after https://github.com/sparkletown/sparkle/pull/926 as it relies on changes made there.

This script will show the count of users that have `enteredVenueIds` matching the supplied venue(s).

This can be useful if you want to quickly see how many people have entered a venue, estimate how many have entered 1 venue but not the other yet (by count), etc.


thecadams (approved):
LGTM.

PR #955, opened 2020-12-10
add logrocket

0xdevalias:
This should be all that's required to integrate Logrocket.

We can look at any further customisations/additions later on:

- https://docs.logrocket.com/docs
- https://docs.logrocket.com/reference#identify


denisdimitrov (approved):
LGTM!

0xdevalias:
Just for future reference, I clicked through with no environment variables set on my local dev environment and it seemed to work fine (no weird errors like we had last time!))

thecadams (commented):
Posthumous review - LGTM as well.

PR #960, opened 2020-12-10
filter badges with venues instead of related

denisdimitrov:
This was causing performance issues.


src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,64 +1,81 @@
...
-  const relatedVenueNames = relatedVenues.map((venue) => venue.name);
+  const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
+  const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+
+  const imgRef: RefObject<HTMLImageElement[]> = useRef([]);
+
+  const firestore = useFirestore();
+
+  const fetchAllVenues = useCallback(async () => {
+    const venuesSnapshot = await firestore.collection("venues").get();
```
0xdevalias, line 37: Currently we seem to be loading all venues from the entire DB here, which we then `.map` a littler further down to get an array of venue names.

Do we need to fetch this? What is it being used for? If we do, can we make it only fetch the venues we need from our visits?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,64 +1,81 @@
...
+
+    const userSnapshot = await firestore.collection("users").doc(user.id).get();
+    const visitsSnapshot = await userSnapshot.ref.collection("visits").get();
+
+    const venues =
+      venuesSnapshot.docs.map(
+        (venueSnapshot) =>
+          ({ ...venueSnapshot.data(), id: venueSnapshot.id } as WithId<
+            AnyVenue
+          >)
```
0xdevalias, line 47: I believe the `as` is being used here because firebase uses `DocumentData` (or similar) as it's type. Does that take a generic parameter where we could pass this type without needing to explicitly cast it here?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,64 +1,81 @@
...
+
+    setVenues(venues);
+    setVisits(visits);
+  }, [firestore, user.id]);
+
+  useEffect(() => {
+    fetchAllVenues();
+  }, [fetchAllVenues]);
+
+  const venueNames = venues.map((venue) => venue.name);
```
0xdevalias, line 65: This will need `useMemo`, otherwise it will break the memoisation where it is passed as a dependency (in `visitHours`) later on.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,64 +1,81 @@
...
-
-  const { relatedVenues } = useConnectRelatedVenues({
-    venueId: currentVenue.id,
-    withEvents: false,
-  });
-  const relatedVenueNames = relatedVenues.map((venue) => venue.name);
+  const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
+  const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+
+  const imgRef: RefObject<HTMLImageElement[]> = useRef([]);
```
0xdevalias, line 32: Maybe rename this `badgesRef` or similar to make it a bit clearer what it's being used for?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -84,27 +100,37 @@ export const Badges: React.FC<{
...
+                  imgRef.current.push(ref);
+                }
+              }}
               src={b.image}
               alt={`${b.label} badge`}
+              onError={() => {
+                if (!imgRef.current?.[index]) return;
+
+                imgRef.current[index].src = "/icons/sparkle-nav-logo.png";
+              }}
```
0xdevalias, line 128: these function refs will change on every render

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -72,6 +73,8 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
 
   const usersById = { [userProfile.id]: userProfile };
 
+  const hasBadges = isTruthy(venue?.showBadges);
```
0xdevalias, line 76: Maybe keep this as `showBadges`?

`hasBadges` implies to me that we've checked an array of badges and it's not empty/similar, whereas this just seems to be whether the venue config is turned on or not.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -647,6 +647,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = (props) => {
           </div>
         )}
 
+        <div className="toggle-room">
+          <h4 className="italic" style={{ fontSize: "20px" }}>
+            Show badges
+          </h4>
```
0xdevalias, line 653: Inline style objects like this create a new object on every render, which can cause extra performance things.

Since it appears to be a static value, could we handle this in a `.scss` class/etc instead?

0xdevalias (changes requested):
A few changes required, see above comments

src/components/organisms/Badges/Badges.tsx
denisdimitrov: Venues is being used for filtering of the visits that exist. Otherwise badges for possibly removed venues will appear.

src/components/organisms/Badges/Badges.tsx
denisdimitrov: I looked around but there doesn't seem to be an option for this.

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -0,0 +1,33 @@
...
+  useCallback,
+} from "react";
+
+export const BadgeImage: React.FC<{
+  badgeImgRefs: RefObject<HTMLImageElement[]>;
+  image?: string;
+  label: string;
+  index: number;
+}> = ({ badgeImgRefs, image, label, index }) => {
+  const defaultAvatar = '/icons/sparkle-nav-logo.png'
```
0xdevalias, line 12: Since this is a static const, it can be moved outside the component render

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -0,0 +1,33 @@
+import React, {
+  RefObject,
+  useCallback,
+} from "react";
+
+export const BadgeImage: React.FC<{
+  badgeImgRefs: RefObject<HTMLImageElement[]>;
```
0xdevalias, line 7: It looks like we're mutating an array passed in here? That sort of breaks React's 'single direction flow' with props/etc. Is there a better way we can achieve this?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -22,14 +22,17 @@ import { WithId } from "utils/id";
...
 export const Badges: React.FC<{
   user: WithId<User>;
   currentVenue: WithId<Venue>;
 }> = ({ user, currentVenue }) => {
   const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
   const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+  const [loading, setLoading] = useState<boolean>(false);
 
-  const imgRef: RefObject<HTMLImageElement[]> = useRef([]);
+  const badgeImgRefs: RefObject<HTMLImageElement[]> = useRef([]);
```
0xdevalias: Is this only used inside of the `BadgeImage` component now? If so, might make sense to put it in there as well? I think this would also resolve my concerns about seemingly mutating an array passed as a prop.

0xdevalias (commented):
I think there are some change requests in my comments above? But not 100% sure if my assumptions are correct.

src/components/organisms/Badges/BadgeImage.tsx
denisdimitrov: badgeImgRefs has been removed because each component can now keep it's own ref and this is unnecessary.

0xdevalias: Sounds good!

src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,77 +1,85 @@
...
-  const venues = useSelector(orderedVenuesSelector);
-
-  const { relatedVenues } = useConnectRelatedVenues({
-    venueId: currentVenue.id,
-    withEvents: false,
-  });
-  const relatedVenueNames = relatedVenues.map((venue) => venue.name);
+  const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
+  const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+  const [loading, setLoading] = useState<boolean>(false);
```
0xdevalias, line 26: prefer `isLoading` style
```suggestion
  const [isLoading, setLoading] = useState<boolean>(false);
```

denisdimitrov:
[Make Badges Performant](https://trello.com/c/Mxmbu0Jl/558-make-badges-performant)

src/components/organisms/Badges/Badges.tsx
0xdevalias: So can we get our badges, and then in batches of 10/similar, lookup whether the venue exists? That would save us from pulling every venue to the frontend, to then filter it down to only a few things.

denisdimitrov:
@0xdevalias I came up with a solution for both this and private chats scenarios. Please take a look in the component. Your opinion is heavy on this one.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -48,13 +39,36 @@ export const Badges: React.FC<{
...
+    const venuesRequests = visits.map((visit) =>
+      firestore.collection("venues").where("name", "==", visit.id).get()
+    );
+
+    const venues: WithId<AnyVenue>[] = [];
+    const hasVenuesRequests = isTruthy(venuesRequests);
+
+    // If there are no venues visited avoid sending the request.
+    if (hasVenuesRequests) {
+      const requestSnapshots = await Promise.all(venuesRequests);
```
0xdevalias, line 51: @denisdimitrov This was actually how I was thinking of doing it too. Removes the need to do the 'limited to 10 at a time' batches. The only potential optimisation I considered was that if we use this `promise.all` pattern but also batch with `in` as well. That way we should be able to send 10x less requests (in theory) (though I expect for most people it will be so few requests in reality that it may not even matter that much)

src/components/organisms/Badges/Badges.tsx
```diff
@@ -48,13 +39,36 @@ export const Badges: React.FC<{
...
+
+      // Promise all returns arrays as response. That's why there is so much depth.
+      // TODO: Same logic can be used for the private chats as well.
+      requestSnapshots.forEach((venuesSnapshot) =>
+        venuesSnapshot.docs.forEach((venueSnapshot) =>
```
0xdevalias, line 56: you could potentially `.flatMap` this? Which should then also allow you to avoid using array mutation with the `.push[]` as you can just return the correct values as a new array from the `.flatMap`

src/components/organisms/Badges/Badges.tsx
```diff
@@ -104,14 +118,14 @@ export const Badges: React.FC<{
       badges.filter(notEmpty).map((badge) => (
         <li className="badge-list-item" key={badge.label}>
           <Link to={getLocationLink(badge.venue, badge.room)}>
-            <BadgeImage image={badge.image} label={badge.label} />
+            <BadgeImage image={badge.image} />
```
0xdevalias: Why did we remove the `badge.label` here?

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -18,8 +18,8 @@ export const BadgeImage: React.FC<{
     <img
       className="badge-list-item-image"
       ref={imageRef}
-      src={image ?? defaultAvatar}
-      alt={`${label} badge`}
+      src={image}
+      alt={""}
```
0xdevalias: Why are we removing the label from here?

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    imageRef.current.src = defaultAvatar;
+  }, []);
+
+  return (
+    <img
+      className="badge-list-item-image"
+      ref={imageRef}
+      src={image}
+      alt={""}
```
0xdevalias, line 21: Blank alt text?

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const BadgeImage: React.FC<{
+  image?: string;
+}> = ({ image }) => {
+  const imageRef = useRef<HTMLImageElement>(null);
+
+  const changeBadgeImage = useCallback(() => {
+    if (!imageRef.current) return;
+
+    imageRef.current.src = defaultAvatar;
+  }, []);
```
0xdevalias, line 14: Do you need this? Can you just pass `image ?? defaultAvatar` directly into the `src=` prop in `img`?

thecadams: `onError` is fired if the image URL doesn't exist. We will need both your suggestion and the ref solution, to ensure the `img` tag always has a valid avatar.

src/components/organisms/Badges/Badges.tsx
thecadams: @denisdimitrov Is it supposed to be `name == visit.id` ? Seems sketchy, just wanted to confirm

src/components/organisms/Badges/BadgeImage.tsx
denisdimitrov: Because the image is for decoration only and if the URL is invalid this alt text shows up for a split second until onError is called which defaults to the default avatar.

src/components/organisms/Badges/BadgeImage.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <img
+      className="badge-list-item-image"
+      ref={imageRef}
+      src={image ?? defaultAvatar}
+      alt="Badge"
```
denisdimitrov, line 21: @thecadams @0xdevalias This alt image is going to shown up if the image url fails to open. After that onError will be called and set the defaultAvatar. This will show the text "Badge" for a split second and that's why I removed the text and set it to empty string.

src/components/organisms/Badges/Badges.tsx
denisdimitrov: It's not being used anymore because of the visual `alt` text flicker.

src/components/organisms/Badges/Badges.tsx
denisdimitrov: @thecadams `visit.id` is actually the venue name / room title.
@0xdevalias Chris already added this with his commit.

denisdimitrov:
PR is ready for re-review and required for tonight. @0xdevalias Chris committed your last changes/comments.

src/components/organisms/Badges/BadgeImage.tsx
0xdevalias: Ha! That's a neat trick (this could potentially also be a solution/workaround for implementing Ed's suggestion of handling 'broken' images in chats)

Part of me wants to say that if we're going to set `alt` to an empty string, is there any point in us setting the attribute at all? Part of me wants to say that in the edgecase of a broken image showing up, does it even matter if we see the text for a split second? Alt text is often used for accessibility/screen reader/etc situations.. so I think I am leaning towards "if it shows up the name for a split second in the edge case where this occurs, that fine" to keep the accessibility/etc benefits.

0xdevalias: (so from my memory of the diff where I originally commented on this, I think keeping the venue title in this component is probably a good idea for that, so it can be "venueName badage")

denisdimitrov: Yes, we can remove the alt image but then the eslint starts crying with the following message:
`img elements must have an alt prop, either with meaningful text, or an empty string for decorative images.`

But this is a very very very edge case, where the venue is created but someone went in the firebase and deleted the image.

denisdimitrov: We should leave it as it is.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -1,77 +1,108 @@
-import React, { useMemo } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import _ from "lodash";
```
0xdevalias, line 2: Can we import the explicit bits we need from the direct path, rather than the entire lib.

@thecadams I wonder if we can set an eslint ban on this pattern too? Without affecting the more 'specific' imports.

src/components/organisms/Badges/Badges.tsx
```diff
@@ -84,29 +115,29 @@ export const Badges: React.FC<{
 
         return {
           venue,
-          image: venue.mapIconImageUrl,
+          image: venue?.host?.icon,
```
0xdevalias, line 118: Do we want a default set on this?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -84,29 +115,29 @@ export const Badges: React.FC<{
...
-          <Link to={getLocationLink(b.venue, b.room)}>
-            <img
-              className="badge-list-item-image"
-              src={b.image}
-              alt={`${b.label} badge`}
-            />
+      badges.filter(notEmpty).map((badge) => (
+        <li className="badge-list-item" key={badge.label}>
+          <Link to={getLocationLink(badge.venue, badge.room)}>
+            <BadgeImage image={badge.image} />
```
0xdevalias, line 130: We should restore the passing through of the venue title here so we can use it in the badge's alt text. 

Something like (this may not be accurate as is, treat it as pseudocode):
```suggestion
            <BadgeImage image={badge.image} badgeName={badge.venue.name} />
```

package.json
```diff
@@ -109,7 +109,7 @@
     "{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
       "prettier --write"
     ],
-    "{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx,json}": [
+    "{src,scripts,functions,configs,reporting}/**/*.{js,jsx,ts,tsx}": [
```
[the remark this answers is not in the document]
thecadams: It's needed to be able to commit; this specific change is already in a bunch of other branches.

src/components/organisms/Badges/Badges.tsx
thecadams: Not sure what the default would be.

denisdimitrov: This is handled in the `BadgeImage` something like image ?? defaultAvatar

0xdevalias (commented):
I think this might be ready to go! Will just do a final pass over the main thread view for any outstanding comments for I approve.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -23,47 +27,51 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
...
 
-  if (!user) return <></>;
-
-  const toggleKidsMode = () => {
-    if (profile) {
+  const toggleKidsMode = useCallback(async () => {
+    if (user && profile) {
       profile.kidsMode = !profile?.kidsMode;
-      updateUserProfile(user.uid, { kidsMode: profile.kidsMode });
+      await updateUserProfile(user.uid, { kidsMode: profile.kidsMode });
```
thecadams, line 51: OOS/future - good catch, this would've silently failed / hit and error boundary. Wonder if there's a lint rule for this.

thecadams (approved):
lgtm. One "future" comment otherwise looks good

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/91

denisdimitrov:
[AAU Badges should appear when opening my own profile (top-right on navbar)](https://trello.com/c/oZ1Z7hDo/533-aau-badges-should-appear-when-opening-my-own-profile-top-right-on-navbar)

PR #962, opened 2020-12-10
enable redux devtools in production

0xdevalias:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/34

See https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f

It's also possible to configure/turn of certain features of it in production if/as required, but for now, i don't think there's any real reason to.


denisdimitrov (approved):
LGTM!

denisdimitrov:
Is exposing all of this data a good idea?

denisdimitrov:
This is going to expose almost all of the firestore data. Except for maybe users and venues.

0xdevalias:
> Is exposing all of this data a good idea?

All of this data is already accessible from the frontend anyway. We shouldn't be relying on 'security through obscurity'. If we're exposing things to the frontend that we shouldn't be, that's something we need to tighten up at our firebase security rules layer rather than 'hoping no one sees it' by not having some QoL dev tools available.

PR #964, opened 2020-12-10
🚧 lint only javascript related files

IvanBuljovcic:
Run _ESLint_ only on JavaScript related files

This is due to the script failing while trying to lint _.scss_ or _.css_ files and not allowing the commit to continue


0xdevalias (approved):
Nice! :shipit:

PR #967, opened 2020-12-10
Remove analytics

thecadams:
Alain suspects this is a cause of perf problems in the app.


src/index.tsx
```diff
@@ -104,7 +101,6 @@ const store = createStore(
   composeWithDevTools(
     applyMiddleware(
       thunkMiddleware,
-      trackingMiddleware(analytics),
```
0xdevalias, line 107: This is the only part of this that matters as per the profiling done IMO

0xdevalias (approved):
This looks good. :shipit:

PR #972, opened 2020-12-11
Add Mixpanel analytics

thecadams:
Just tracking one event, we should add more asap


src/hooks/useMixpanel.ts
```diff
@@ -0,0 +1,22 @@
...
+  event_name: string,
+  properties?: Dict,
+  optionsOrCallback?: RequestOptions | Callback,
+  callback?: Callback
+) => void = () => {};
+
+export const useMixpanel = () => {
+  const mixpanelMemo = useMemo(
+    () => ({
+      track: MIXPANEL_PROJECT_TOKEN ? mixpanel.track.bind(mixpanel) : noopTrack,
```
thecadams, line 16: Mixpanel doesn't seem to provide an easy way to know if it's initialized, but I don't want the conditional spread through the code / error prone / repeated. So added a hook to access it, added linting etc.

https://stackoverflow.com/questions/32203420/check-if-mixpanel-library-has-been-loaded

.eslintrc
```diff
@@ -28,6 +28,10 @@
       {
         "name": "firebase",
         "message": "Use firebase/app instead"
+      },
+      {
+        "name": "mixpanel-browser",
+        "message": "Import safe version from hooks/useMixpanel instead"
       }
```
0xdevalias, line 35: 👌🏻  good use of proactive warnings/guidance!

src/hooks/useMixpanel.ts
```diff
@@ -0,0 +1,24 @@
...
+  event_name: string,
+  properties?: Dict,
+  optionsOrCallback?: RequestOptions | Callback,
+  callback?: Callback
+) => void = () => {};
+
+export const useMixpanel = () => {
+  const mixpanelMemo = useMemo(
+    () => ({
+      track: isLoaded() ? mixpanel.track.bind(mixpanel) : noopTrack,
```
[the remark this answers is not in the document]
0xdevalias: Because this is binding once the first time this hook is called, it will never break memo to use the correct callback.

I believe what we would want to do here is something like (untested):
```suggestion
      track: () => isLoaded() ? mixpanel.track : noopTrack,
```

src/hooks/useMixpanel.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// https://stackoverflow.com/questions/32203420/check-if-mixpanel-library-has-been-loaded
+const isLoaded = () => mixpanel.hasOwnProperty("get_distinct_id");
+
+const noopTrack: (
+  event_name: string,
+  properties?: Dict,
+  optionsOrCallback?: RequestOptions | Callback,
+  callback?: Callback
+) => void = () => {};
```
0xdevalias, line 13: Is there a type exported from mixpanel that we can use for this? Maybe could even do (untested):
```suggestion
const noopTrack: typeof mixpanel.track = () => {};
```

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -299,6 +301,8 @@ const VenuePage = () => {
     history.push(`/account/profile?venueId=${venueId}`);
   }
 
+  mixpanel.track("Page load", { venueId: venue.id, template: venue.template });
+
```
0xdevalias, line 305: This will run every single time this component function is run, so a lot. We probably want this in a `useEffect` call that only runs a single time (so empty deps array)

```suggestion
useEffect(() => {
  mixpanel.track("VenuePage loaded", { venueId: venue.id, template: venue.template });
}, []) 
```

This may potentially cause an eslint warning due to not including all of the dependencies.. it depends on your intended usage of this track here as to what makes the most sense, but I think it would be fine if this is called whenever the `venueId` or `template` changes? If so, I would probably do:

```suggestion
const venueId = venue.id
const venueTemplate = venue.template

useEffect(() => {
  mixpanel.track("VenuePage loaded", { venueId, template: venueTemplate });
}, [venueId, venueTemplate]) 
```

0xdevalias (changes requested):
Some questions + some changes requested.

src/hooks/useMixpanel.ts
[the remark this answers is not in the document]
thecadams: The `bind` is needed because `mixpanel.track` has references to `this` in the code; it crashes without the `bind`.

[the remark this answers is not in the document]
thecadams: And the suggestion changes the function signature I think such that you'd need to wrap the `track` call. My goal was to make this safety hook as pass-through as possible, so you don't need to think about how it works. Which follows the precedent of the `useDispatch` and `useSelector` we already have.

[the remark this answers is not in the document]
0xdevalias: We need the wrapper so that it is evaluated on every call, so the suggestion may not be perfectly accurate as is, but we need to do it and wrap our if/etc logic. Otherwise it will be evaluated once at app load and never again.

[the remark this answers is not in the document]
thecadams: OK, yep, I see. Thanks for the explanation. Looking at this.

src/hooks/useMixpanel.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Returns an object to allow us to expose more of Mixpanel's API in future.
+ *
+ * NOTE: Extra level of function indirection ensures `isLoaded` is evaluated not just the first time.
+ */
+export const useMixpanel = () =>
+  useMemo(
+    () => ({
+      track: () => (isLoaded() ? mixpanel.track.bind(mixpanel) : noopTrack),
```
0xdevalias, line 18: I think as is this may not work correctly.. as I forgot that we probably want to pass parameters through to mixpanel's track function.. so we need to add whatever params it accepts here as well.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -239,6 +241,15 @@ const VenuePage = () => {
...
 
+  useEffect(() => {
+    if (user && profile && venue && venueId) {
+      mixpanel.track()("VenuePage loaded", {
+        venueId: venue.id,
+        template: venue.template,
+      });
+    }
+  }, [user, profile, venue, venueId, mixpanel]);
+
```
0xdevalias, line 252: Only pass through the bits of venue we care about, so it doesn't break memo on unrelated things:

```suggestion
  const venueTemplate = venue.template;
  
  useEffect(() => {
    if (user && profile && venueId && venueTemplate) {
      mixpanel.track()("VenuePage loaded", {
        venueId,
        template: venueTemplate,
      });
    }
  }, [user, profile, venueId, venueTemplate, mixpanel]);
```

0xdevalias (approved):
Looks good! If the data comes through, then LGTM! :shipit:

thecadams:
[Integrate Mixpanel for product analytics](https://trello.com/c/WtV4sMX2/571-integrate-mixpanel-for-product-analytics)

PR #973, opened 2020-12-11
fix fontawesome issues introduced in recent version bump 

0xdevalias:
It seems that when https://github.com/sparkletown/sparkle/pull/872 was merged it caused errors while trying to deploy the hosting, due to a TypeScript compilation error:

- https://app.circleci.com/pipelines/github/sparkletown/sparkle/3822/workflows/5974d56e-9d2c-41c1-a041-f4523b99e947/jobs/6370

```
TypeScript error in /home/circleci/project/src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx(33,30):
Type 'IconDefinition' is not assignable to type 'IconProp'.
  Type 'IconDefinition' is not assignable to type 'IconLookup'.  TS2322
```

I believe the versions of the fontawesome libs need to be kept in sync with one another, this PR will fix that.

I'd also like to understand the differences in the `package-lock.json` (i expect there will be multiple versions of a common dependency, causing the issues), and then can probably write a test to ensure this doesn't happen in future.

As a separate follow up PR, I will also be looking how we can ensure our CI notices TypeScript errors at the linting stage, without having to do a full build of the app (as that is quite heavy, and we don't need it till we're deploying)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1794,9 +1794,9 @@
       }
     },
     "@fortawesome/react-fontawesome": {
-      "version": "0.1.10",
-      "resolved": "https://registry.npmjs.org/@fortawesome/react-fontawesome/-/react-fontawesome-0.1.10.tgz",
-      "integrity": "sha512-UGdpJiLBIqR/8xcLrCarf2pChqQFKjDTD02C7ZS/odpOVVl2YuHYRCLEOQ0GpfOk6jtYhmouSFOFoC8qNCe5cg==",
+      "version": "0.1.13",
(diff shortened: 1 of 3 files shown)


0xdevalias:
In https://github.com/sparkletown/sparkle/pull/872 when we introduced `v15.5.1` of `@fortawesome/free-solid-svg-icons` it required `@fortawesome/fontawesome-common-types` `v0.2.32` (whereas the other dependencies still used `v0.2.28`)

https://github.com/sparkletown/sparkle/blob/0a3fa3bbb5787dea03f04ad1c9b1234b9247f0c9/package-lock.json#L1753-L1788

We should be able to parse the keys in the `package-lock.json` and ensure they all have the same version.

thecadams (approved):

PR #976, opened 2020-12-11
Fix partymap height for safari

thecadams:
Safari doesn't seem to use width to set height the way Chrome does:

https://stackoverflow.com/questions/11276367/forcing-aspect-ratio-with-css-doesnt-work-on-safari

This fix explicitly sets the height on the map background image, so it doesn't default to the intrinsic height of the image.

### DIFF
--- a/src/components/templates/PartyMap/components/Map/Map.scss
+++ b/src/components/templates/PartyMap/components/Map/Map.scss
@@ -31,6 +31,7 @@ $border-radius: 28px;
 
 .party-map-background {
   border-radius: 15px;
+  height: 100%;
 }
 
 .camp-background {


0xdevalias (approved):
Looks good.

Also from that same StackOverflow, at least it sounds like it will soon not be a class of issue we need to care about:

> As announced during WWDC20 at What's new for web developers, starting from Safari 13.1, WebKit now calculates the image aspect ratio from the width and height values of the image element. This means that the issue you have faced is fixed on the latest version of the browser.

0xdevalias:
[Safari: Map is about 5x as tall as it should be, added to spreadsheet.](https://trello.com/c/e65olnAs/570-safari-map-is-about-5x-as-tall-as-it-should-be-added-to-spreadsheet)

PR #956, opened 2020-12-10
Remove contexts

thecadams:

PR #980, opened 2020-12-11
Revert "Remove contexts (#956)"

thecadams:
This reverts commit 3a9d1a8ad40912624e708c71c81b549c43037b70.

This was making the chats disappear in the right hand sidebar. We still need to understand the cause and re-introduce this fix later.


denisdimitrov (approved):
LGTM!

PR #981, opened 2020-12-11
Fix chat after removing contexts

thecadams:
Did not remember to connect to venueChats, which meant no chat messages appeared


0xdevalias (commented):
From a quick skim of the code diff this looks good. Haven't clicked through it though, so if it works, :shipit:

PR #984, opened 2020-12-11
Support new environment for Firebase hosting

thecadams:
We have a new Firebase project and a new custom domain.

The hope is this creates some isolation between different parties.


.circleci/config.yml
```diff
@@ -125,6 +128,12 @@ jobs:
               TARGET=sparkleverse
               RELEASE_STAGE=sparkleverse
             fi
+            if [ $CIRCLE_BRANCH = "sparkle1" ]; then
+              PREFIX=SPARKLE1_
+              ENV=prod
+              TARGET=sparkle1
+              RELEASE_STAGE=production
+            fi
```
0xdevalias, line 136: I think this needs to be changed?
```suggestion
            if [ $CIRCLE_BRANCH = "sparkle1" ]; then
              PREFIX=SPARKLE1_
              ENV=sparkl1
              TARGET=sparkle1
              RELEASE_STAGE=sparkle1
            fi
```

0xdevalias (approved):
The code as changed here looks good, not sure if it is 100% of the bits we will need yet though. If you're happy that it does, lets merge it.

0xdevalias (commented):
Need to add sparkle1 to the stages we want reported in bugsnag

0xdevalias (approved):
Looks fine as the code changes are.

PR #987, opened 2020-12-11
add scripts/clone-data-across-firebase-projects

0xdevalias:
This script allows you to clone the specified venue(s) from one firebase project (eg. production) to another (eg. staging)

### DIFF
--- a/scripts/clone-data-across-firebase-projects.ts
+++ b/scripts/clone-data-across-firebase-projects.ts
@@ -1,0 +1,81 @@
+#!/usr/bin/env node -r esm -r ts-node/register
+
+import { AnyVenue } from "../src/types/Firestore";
+import { initFirebaseAdminApp } from "./lib/helpers";
+
+const CONFIRM_VALUE = "i-have-edited-the-script-and-am-sure";
+
+const usage = () => {
+  const scriptName = process.argv[1];
+  const helpText = `
+---------------------------------------------------------  
+${scriptName}: Clone venue(s) between different firebase projects
+
+Usage: node ${scriptName} ${CONFIRM_VALUE}
+
+Example: node ${scriptName} ${CONFIRM_VALUE}
+---------------------------------------------------------
+`;
+
+  console.log(helpText);
+  process.exit(1);
+};
+
+const [confirmationCheck] = process.argv.slice(2);
+if (confirmationCheck !== CONFIRM_VALUE) {
+  usage();
+}
+
+const SOURCE_PROJECT_ID = "co-reality-map";
+const DEST_PROJECT_ID = "co-reality-staging";
+
+const VENUES_TO_CLONE = ["wayspace"];
+
+const sourceApp = initFirebaseAdminApp(SOURCE_PROJECT_ID, "sourceApp");
+const destApp = initFirebaseAdminApp(DEST_PROJECT_ID, "destApp");
+
+// TODO: do we need to copy roles across?
+// TODO: venues (owners will need to be changed)
+// TODO: check if venue already exists (be safe, don't overwrite!)
+
+(async () => {
+  // TODO: @debt use filters so we are only getting batches of the venues we want, use 'in', _.chunk + Promise.all, etc to page through the data
+  const allSourceVenues = await sourceApp
+    .firestore()
+    .collection("venues")
+    .listDocuments();
+
+  const wantedSourceVenues = await Promise.all(
+    allSourceVenues
+      .filter((venue) => VENUES_TO_CLONE.includes(venue.id))
+      .map((venue) =>
+        venue.get().then((v) => ({ ...(v.data() as AnyVenue), id: v.id }))
+      )
+  );
+
+  console.log("total venues:", allSourceVenues.length);
+  console.log("wanted venues:", wantedSourceVenues.length, VENUES_TO_CLONE);
+
+  // TODO: we should save backups before we potentially overwrite things..
+  // const saveToBackupFile = makeSaveToBackupFile(
+  //   `${projectId}-foo`
+  // );
+
+  const destAppBatch = destApp.firestore().batch();
+
+  wantedSourceVenues.forEach((venue) => {
+    const { id, ...venueData } = venue;
+    const destVenueRef = destApp
+      .firestore()
+      .collection("venues")
+      .doc(`devaliastest${id}`);
+    console.log(id, destVenueRef);
+
+    destAppBatch.set(destVenueRef, venueData);
+
+    console.log("added venue to batch:", venue.name);
+  });
+
+  const writeResult = await destAppBatch.commit();
+  console.log(writeResult);
+})();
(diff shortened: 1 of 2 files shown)


scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,81 @@
...
+const DEST_PROJECT_ID = "co-reality-staging";
+
+const VENUES_TO_CLONE = ["wayspace"];
+
+const sourceApp = initFirebaseAdminApp(SOURCE_PROJECT_ID, "sourceApp");
+const destApp = initFirebaseAdminApp(DEST_PROJECT_ID, "destApp");
+
+// TODO: do we need to copy roles across?
+// TODO: venues (owners will need to be changed)
+// TODO: check if venue already exists (be safe, don't overwrite!)
```
0xdevalias, line 39: TODO

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,81 @@
...
+
+(async () => {
+  // TODO: @debt use filters so we are only getting batches of the venues we want, use 'in', _.chunk + Promise.all, etc to page through the data
```
0xdevalias, line 42: TODO

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,81 @@
...
+      )
+  );
+
+  console.log("total venues:", allSourceVenues.length);
+  console.log("wanted venues:", wantedSourceVenues.length, VENUES_TO_CLONE);
+
+  // TODO: we should save backups before we potentially overwrite things..
+  // const saveToBackupFile = makeSaveToBackupFile(
+  //   `${projectId}-foo`
+  // );
```
0xdevalias, line 62: TODO

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const destAppBatch = destApp.firestore().batch();
+
+  wantedSourceVenues.forEach((venue) => {
+    const { id, ...venueData } = venue;
+    const destVenueRef = destApp
+      .firestore()
+      .collection("venues")
+      .doc(`devaliastest${id}`);
```
0xdevalias, line 71: TODO: remove `devaliastest` prefix

0xdevalias (commented):
This should work, but can't currently validate that the data ends up in the other project as it should.

0xdevalias:
> This should work, but can't currently validate that the data ends up in the other project as it should.

Ok, figured this out. Even though we've (theoretically) configured 2 apps, it wrote the data back to the `sourceApp` anyway. Need to figure out why that is.. potentially something to do with the service key we're using?

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const SOURCE_PROJECT_ID = "co-reality-map";
+const DEST_PROJECT_ID = "co-reality-staging";
+
+const SOURCE_CREDENTIAL_FILE =
+  "co-reality-map-firebase-adminsdk-47j7x-530045073b.json";
+const DEST_CREDENTIAL_FILE =
+  "co-reality-staging-firebase-adminsdk-yy5cq-5fd568c2f4.json";
+
+const VENUES_TO_CLONE = ["wayspace"];
```
0xdevalias, line 21: This is how you configure the things for this script

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,101 @@
...
+  const scriptName = process.argv[1];
+  const helpText = `
+---------------------------------------------------------  
+${scriptName}: Clone venue(s) between different firebase projects
+
+Usage: node ${scriptName} ${CONFIRM_VALUE}
+
+Example: node ${scriptName} ${CONFIRM_VALUE}
+---------------------------------------------------------
+`;
```
0xdevalias, line 39: And this is how you call it

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,100 @@
...
+// Configuration (this is the bit you should edit)
+// ---------------------------------------------------------
+
+const SOURCE_PROJECT_ID = "co-reality-map";
+const DEST_PROJECT_ID = "co-reality-staging";
+
+const SOURCE_CREDENTIAL_FILE =
+  "co-reality-map-firebase-adminsdk-47j7x-530045073b.json";
+const DEST_CREDENTIAL_FILE =
+  "co-reality-staging-firebase-adminsdk-yy5cq-5fd568c2f4.json";
```
thecadams, line 19: Should these be command line args?

scripts/clone-data-across-firebase-projects.ts
0xdevalias: resolved in https://github.com/sparkletown/sparkle/pull/987/commits/67e28405436d2a4e6c3862be199d8dc983823353

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const VENUES_TO_CLONE = ["wayspace"];
```
thecadams, line 21: Same as above. Maybe we can do this later

thecadams (approved):

scripts/clone-data-across-firebase-projects.ts
0xdevalias: Eventually, definitely. But I would say we should move to a proper CLI framework as per https://github.com/sparkletown/internal-sparkle-issues/issues/39 before doing so, as otherwise it's just going to be way too unruly, with too many arguments, etc.

0xdevalias: I kind of envision it the same way the table generation script works currently. We edit the config within the script as part of running it.

0xdevalias:
Currently this only copies the venues, but it would be good to be able to copy the events as well. I have a bit of a hacky attempt to copy this script + rework the query for this locally.. but didn't really get it to a finished state.

Here's the main relevant query chunk from it in case it's of any use (may not work in it's current state)

```typescript
// TODO: @debt use filters so we are only getting batches of the venues we want, use 'in', _.chunk + Promise.all, etc to page through the data
  const allSourceVenues = await sourceApp
    .firestore()
    .collection("venues")
    .listDocuments();

  const wantedSourceVenues = allSourceVenues.filter((venue) =>
    VENUES_TO_CLONE.includes(venue.id)
  );

  const wantedSourceVenueEvents = await Promise.all(
    wantedSourceVenues.map(async (venue) => {
      const venueEvents = await venue.collection("events").listDocuments();

      const venueEventsWithId = Promise.all(
        venueEvents.map(async (venueEvent) => ({
          id: venueEvent.id,
          data: await venueEvent.get(),
        }))
      );

      return {
        id: venue.id,
        events: await venueEventsWithId,
      };
    })
  );
```

scripts/clone-data-across-firebase-projects.ts
thecadams: Yep, makes sense.

PR #988, opened 2020-12-12
fix a bunch more performance related things (including a pathological query of all private chats)

0xdevalias:
The pathological query was the same type that https://github.com/sparkletown/sparkle/pull/801 aims to solve/protect against.

If my understanding of the situation is correct, we were likely querying all private chat messages when `user` was `undefined`, which guessing by the performance profiling was probably a lot.

----

Was operating from a few performance profiles I managed to get during some of our more active events of late ([slack ref](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1607725759308900)):

> RE: performance. Seeing tons of `ChatMessage [update]` things in this performance trace. Not 100% how to read it, but I think the top lines might represent reasons why it rendered or similar

image

> if i’m guessing right, the number shown in the tooltip may be the number of times that ‘thing’ has occured during the performance trace.. which if that is true.. the chat stuff has done like 47k updates in… not very long at all

image

> as has `ChatList`

image


src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -46,24 +52,25 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
...
+  useFirestoreConnect(
+    user && user.uid
+      ? {
+          collection: "privatechats",
+          doc: user.uid,
+          subcollections: [{ collection: "chats" }],
+          storeAs: "privatechats",
+        }
+      : undefined
+  );
```
0xdevalias, line 71: Here's the pathological query

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -58,43 +66,39 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
...
+  const venueUsers = useMemo(() => {
+    if (!users) return [];
+
+    return users.filter(
+      (user) =>
+        !!user.lastSeenIn &&
+        user.lastSeenIn[venueToUseName ?? ""] >
+          (nowMs - LOC_UPDATE_FREQ_MS * 2) / 1000
+    );
+  }, [nowMs, users, venueToUseName]);
```
0xdevalias, line 95: It may be possible to replace this whole piece with `useCampPartygoers`

src/utils/selectors.ts
```diff
@@ -124,6 +126,23 @@ export const isUserPurchaseHistoryRequestedSelector: SparkleSelector<boolean> =
...
+
+  return privateChats.some(
+    (message) =>
+      message.from !== user?.uid &&
+      message.deleted !== true &&
+      message.type === "private" &&
+      message.ts_utc.seconds > HIDE_BEFORE &&
+      message.isRead === false
+  );
+};
```
0xdevalias, line 144: Here's an example of a 'smarter' selector. Instead of selecting all of the data, then filtering it once we get to the component, by doing it in the selector we will get extra memoisation/etc (namely that our component will only need to re-render if the value emitted by the selector changes, not the 'dependencies' we use to calculate that final value.

src/components/molecules/Chatbox/Chatbox.tsx
```diff
@@ -29,6 +29,7 @@ interface ChatboxProps {
   isVenueChat?: boolean;
 }
 
+// TODO: we have a ChatBox in organisms but also in molecules.. are they the same? Can we de-dupe them?
```
thecadams, line 32: Yep we should look into this
```suggestion
// @debt TODO: we have a ChatBox in organisms but also in molecules.. are they the same? Can we de-dupe them?
```

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -58,43 +66,39 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
...
-          !!user.lastSeenIn &&
-          user.lastSeenIn[venueToUse?.name ?? ""] >
-            (nowMs - LOC_UPDATE_FREQ_MS * 2) / 1000
-      )
-    : [];
+  // TODO: this will break memo on venueUsers (below) every 5min, does that matter?
+  const [nowMs, setNowMs] = useState(Date.now());
+  useInterval(() => {
+    setNowMs(Date.now());
+  }, LOC_UPDATE_FREQ_MS);
```
thecadams, line 81: Consider wrapping this in `useEffect` so we can undo it on unmount. Although, it is likely unimportant as this component has the same lifecycle as the page view.

thecadams (commented):
Going to have to test this out a little by way of review, but meanwhile only a couple minor thoughts. Great to see such proactive bugfixing.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
0xdevalias: `useInterval` is a wrapping hook i wrote that handles all of that for us in a single reusable way. It clears the interval on unmount:

- https://github.com/sparkletown/sparkle/blob/staging/src/hooks/useInterval.ts

src/utils/selectors.ts
```diff
@@ -124,6 +126,23 @@ export const isUserPurchaseHistoryRequestedSelector: SparkleSelector<boolean> =
...
 );
 
+const DAYS_AGO = getDaysAgoInSeconds(VENUE_CHAT_AGE_DAYS);
+const HIDE_BEFORE = roundToNearestHour(DAYS_AGO);
+
+export const unreadMessagesSelector = (state: RootState) => {
+  const user = authSelector(state);
+  const privateChats = privateChatsSelector(state) ?? [];
+
+  return privateChats.some(
```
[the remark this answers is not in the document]
0xdevalias: Such as?

0xdevalias:
> @thecadams Going to have to test this out a little by way of review, but meanwhile only a couple minor thoughts. Great to see such proactive bugfixing.

Yup, 100% agree. I'm not in any mental state to guarantee it will work perfectly as is, but noticed some things as I was browsing my profiling stuff during the meeting this morning, and wanted to make sure I got it captured and up before I 💤 . I think the pathological query on chats could well have an outsized positive impact on performance, even without the rest of the changes here.

I'm going to /offline for the weekend, but may pop in if I see comments/etc. Happy for you to approve/merge if you get inspired and do theclick throughss/etc, but also, make sure to have a break.

src/components/organisms/Chatbox/Chatbox.tsx
thecadams: Hope it resolves it. Seemed ok in my inspection.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
thecadams: No problem, I see that now.

src/utils/selectors.ts
thecadams: [OOS/future] Yes, and as we pull non-domain logic out of the code, it will be great to change from doing `useSelector(somethingSelector)` to `useAThingWeNeed` so the selector and `useFirestoreConnect` can go into one place; to make it easier to change how we get data in future.

thecadams (approved):
LGTM and works in my testing!

src/utils/selectors.ts
0xdevalias: Definitely! And I believe we could also put the selector first, then use the 'does this exist' helpers or similar to decide if we even need to 'connect' the query at that point, so that it automagically knows if it needs to, and that is abstracted away from us so we just need to call `useTheThing` and not have to 🧠  about it beyond that.

PR #992, opened 2020-12-13
Entrance Experience Customisations

0xdevalias:
**Note:** opening this PR to add visibility to the branch, but I'm not sure if this is something we need/want to continue with. If not, we can close this PR + delete the branch (it can always be restored in future if required)


src/pages/Admin/Venue/EntranceInput.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <div className="input-container" style={{ marginBottom: "1.5rem" }}>
+      <h4 className="italic input-header">Venue entrance</h4>
+      {indexes.map((i) => renderEntranceInput(i))}
+
+      <div className="dynamic-input__button-wrapper">
+        <Button onClick={add}>Add entrance step</Button>
+        {indexes.length > 0 && (
+          <Button onClick={clear} variant="secondary">
+            Remvoe all entrance steps
```
thecadams, line 110: nit: Remove

thecadams (approved):
lgtm, long as you've tested it and can be confident it'll work.

functions/venue.js
```diff
@@ -113,7 +112,6 @@ const createVenueData = (data, context) => {
     case VenueTemplate.partymap:
     case VenueTemplate.themecamp:
       venueData.rooms = data.rooms;
-      venueData.mapBackgroundImageUrl = data.mapBackgroundImageUrl;
```
0xdevalias, line 116: This seems to be the source of the bug raised in https://github.com/sparkletown/internal-sparkle-issues/issues/131 with fix in https://github.com/sparkletown/sparkle/pull/1071

PR #993, opened 2020-12-13
Sync sparkle1 -> staging

0xdevalias:
It seems the `sparkle1` branch was a few commits ahead of `staging`, so to ensure their histories are kept aligned, I have opened this to sync those changes back.

@thecadams would be good to get your thoughts on this.


thecadams (approved):
lgtm; wayspace is working in `sparkle1` so going to assume `firebase-functions` isn't needed in this place, and merge!

PR #1006, opened 2020-12-14
Fix references when cloning venues

thecadams:
Minor fix to ensure users don't cross into the wrong environment when clicking into rooms.


scripts/clone-data-across-firebase-projects.ts
```diff
@@ -90,6 +93,21 @@ const destApp = initFirebaseAdminApp(DEST_PROJECT_ID, {
     const { id, ...venueData } = venue;
     const destVenueRef = destApp.firestore().collection("venues").doc(id);
 
+    Object.keys(venue).forEach((key) => {
+      if (venue[key].toString().includes(`https://${SOURCE_DOMAIN}`)) {
```
thecadams, line 97: Note: prefixing with `https://` here is intended to ensure we don't always replace `sparkle.space` with `1.sparkle.space`, so we won't accidentally end up with `1.1.sparkle.space` :)

0xdevalias: Do we want to handle `http://` and `//` as well? Or just treat them as old/bad data/unlikely to exist.

0xdevalias: This is basically looking for URLs in the top level keys on `venue` yeah? Is that sufficient? Should it be looking at any 'deeper' keys/paths?

thecadams: > Do we want to handle `http://` and `//` as well? Or just treat them as old/bad data/unlikely to exist.

It could do `//`, `http://`, and `https://`

> This is basically looking for URLs in the top level keys on venue yeah? Is that sufficient? Should it be looking at any 'deeper' keys/paths?

That's a good catch. I'll also modify it to examine the properties on each room, if they exist.

PR #1007, opened 2020-12-14
Remove contexts - attempt 2

thecadams:
Revert "Revert "Remove contexts (#956)" (#980)"

This reverts commit e30886ba64e2ba5499139ab11b29eefac60481a4.

This is a resumption of the PR in which we removed contexts from the code; which is suspected to be a cause of much re-rendering. This PR also cleans up some misuses of the contexts, moving things to the store and to hooks.


thecadams:
[Performance TroubleShooting](https://trello.com/c/25Oynxx2/559-performance-troubleshooting)

src/components/organisms/Chatbox/ChatForm.tsx
```diff
@@ -53,16 +52,43 @@ const ChatForm: React.FunctionComponent<PropsType> = ({
     if (!currentUserUID) return;
     switch (type) {
       case "private":
-        return (
-          discussionPartner &&
-          sendPrivateChat(currentUserUID, discussionPartner.id, text)
+        if (!discussionPartner) return;
```
mike-lvov, line 55: ^ spacing above

src/components/templates/FireBarrel/useVideo.tsx
```diff
@@ -2,34 +2,6 @@ import { useCallback, useEffect, useState } from "react";
 import Video, { LocalVideoTrack } from "twilio-video";
 import firebase from "firebase/app";
 
-// export const VideoContext = createContext(null);
```
mike-lvov, line 5: I think it is wise not to have commented code like this in our codebase as it's unclear why it was commented.

src/hooks/useDispatch.ts
```diff
@@ -1,7 +1,9 @@
 // eslint-disable-next-line no-restricted-imports
 import { useDispatch as _useDispatch } from "react-redux";
+import { ThunkDispatch } from "redux-thunk";
 import { RootActions } from "store/actions";
-import { Dispatch as _Dispatch } from "redux";
+import { RootState } from "index";
 
-export const useDispatch = () => _useDispatch<_Dispatch<RootActions>>();
+export const useDispatch = () =>
```
mike-lvov, line 7: Not sure about whether I understand it correctly, but wouldn't it be better to have it like this:

```
export const useDispatch = _useDispatch<ThunkDispatch<RootState, void, RootActions>>;
```

src/hooks/useReactions.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!venueId) return;
+
+    firebase
+      .firestore()
+      .collection(`experiences/${venueId}/reactions`)
+      .where("created_at", ">", Date.now())
+      .onSnapshot((snapshot) => {
+        snapshot.docChanges().forEach((change) => {
+          if (change.type === "added") {
+            const newReaction = change.doc.data() as Reaction;
```
mike-lvov, line 20: Is it possible to somehow omit the type assignment here? Not sure how to do it, just asking.

src/hooks/useReactions.ts
```diff
@@ -0,0 +1,34 @@
...
+            setReactions((prevReactions) => [...prevReactions, newReaction]);
+
+            setTimeout(() => {
+              setReactions((prevReactions) => {
```
mike-lvov, line 24: I think `return` can be omitted here

```
setReactions((prevReactions) => prevReactions.filter((r) => r !== newReaction);

File: src/store/actions/Chat.ts [hunk 6]
```diff
@@ -0,0 +1,161 @@
+import { createAsyncThunk } from "@reduxjs/toolkit";
+import firebase from "firebase/app";
+
+export const SEND_GLOBAL_CHAT: string = "SEND_GLOBAL_CHAT";
```
mike-lvov, line 4: Up for a discussion: 
```
export const GLOBAL_CHAT = "GLOBAL_CHAT";
```

src/store/actions/Chat.ts
```diff
@@ -0,0 +1,161 @@
...
+  to: string;
+}
+
+interface SendGlobalChatFields extends BaseSendChatFields {}
+interface SendPrivateChatFields extends Omit<BaseSendChatToFields, "venueId"> {}
+interface SendRoomChatFields extends BaseSendChatToFields {}
+interface SendTableChatFields extends BaseSendChatToFields {}
+
+interface SendGlobalChatAction extends SendGlobalChatFields {
+  type: typeof SEND_GLOBAL_CHAT;
```
mike-lvov, line 24: I don't think that works as intended. Not sure about the goals ofc, but this could work:

```
interface SendPrivateChatAction extends SendPrivateChatFields {
  type: ChatMessageType.Global;
}
```

And we can delete those consts(SEND_ROOM_CHAT) and export `ChatMessages` enum instead

src/store/actions/Chat.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+};
+
+const saveMessage = async (venueId: string, message: ChatMessage) =>
+  await firebase
+    .firestore()
+    .collection(collectionName(venueId))
+    .add(buildMessage(message));
+
+export const sendGlobalChat = createAsyncThunk<void, SendGlobalChatFields>(
```
mike-lvov, line 113: Up for a discussion:

Is it possible that sendMsgToGlobalChat is more descriptive?

mike-lvov:
Looks good in general! Just a few minor mainly opinionated comments by me. Good job in removing that legacy!

src/components/molecules/VenueChat/VenueChat.tsx
```diff
@@ -16,15 +9,17 @@ import { VENUE_CHAT_AGE_DAYS } from "settings";
...
 import { currentVenueSelectorData, venueChatsSelector } from "utils/selectors";
 
+import { useDispatch } from "hooks/useDispatch";
 import useRoles from "hooks/useRoles";
 import { useUser } from "hooks/useUser";
 import { useSelector } from "hooks/useSelector";
 import { useVenueId } from "hooks/useVenueId";
 
-import { ChatContext, chatSort } from "components/context/ChatContext";
+import { chatSort } from "utils/chat";
```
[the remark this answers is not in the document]
thecadams: This is interesting. Not sure I agree about the suggestion, we should hop on zoom and talk through it.

`chatSort` is not grammatically intended to say that it sorts the chat (that is able to be inferred from context). It's also a sorting predicate, so `Sort` should not be used as a verb, because it doesn't sort - it only describes how to sort. But there's still a `Sort` in there so what's it for? The `Sort` in `chatSort` serves to disambiguate it from other sorting predicates. Including `Sort` in `chatSort` tells the reader: "this is the chat sort predicate, you can expect there to be others elsewhere in the code".

Something called `sortChat` would need to be more self-contained to earn that title. It's the difference between:

```
export const chatSort: (a: BaseChatMessage, b: BaseChatMessage) => number = (
  a: BaseChatMessage,
  b: BaseChatMessage
) => b.ts_utc.valueOf().localeCompare(a.ts_utc.valueOf());
```

And:

```
const sortChat(chat: ???): ??? =>
 chat.sort((a, b) => b.ts_utc.valueOf().localeCompare(a.ts_utc.valueOf()));
```

The latter performs the action, whereas the former is only the predicate.

There's another problem with `sortChat` which is that we're calling these `chats` in the code, so the pluralization doesn't work.

My leaning would be to keep `chatSort` but happy to have a chat and see what you think. And thanks for stimulating the discussion!

src/components/templates/FireBarrel/useVideo.tsx
thecadams: Agreed; this code has been unused for a long time. 💥 and away it goes! Great catch!

src/hooks/useReactions.ts
thecadams: Yes, it would be great! These casts hide potential `undefined` values that bypass TypeScript and crash at runtime.

@0xdevalias has suggested a couple of ORMs we could use and I'd be very happy to see us move in that direction:

* https://github.com/wovalle/fireorm
* https://fireadmin.io/
* https://github.com/prescottprue/fireadmin

src/store/actions/Chat.ts
thecadams: Good call out. My leaning would be toward including the verb so it's immediately obvious when reading the code, that this is a create operation. An action of type `GLOBAL_CHAT` could be interpreted by a reader as "get global chat" or as "hey, a global chat happened, and the UI needs to update".

I'm leaning towards keeping `SEND` in there. But happy to be convinced otherwise!

src/store/actions/Chat.ts
thecadams: Hm, this is an unfortunate one. I'd like to do this simplification but unfortunately the `type` values (`global`, `room`, `table`, `private`) are already in the database. Doing that change would make it impossible to view old chats, because our chats would save to `SEND_{GLOBAL,PRIVATE,TABLE,ROOM}_CHAT`.

So it looks to me like we have to leave it this way unfortunately.

Good catch though, glad you're spotting these!

src/store/actions/Chat.ts
thecadams: Good thinking here. It's more descriptive, yes; and it might help a reader out a little. But I suspect that since the meaning is the same, it makes sense to go with the more minimal expression:

* `sendGlobalChat` can be read as "send a global chat", where "chat" is the object of the sentence.
* `sendMsgToGlobalChat` can be read as "send a message to the global chat", where "message" is the object of the sentence.

They are both identical in meaning, but the shorter one is no more ambiguous, and I'm all for less code to read.

This is a really good suggestion, thanks for stimulating the discussion, very grateful you're suggesting these improvements.

My leaning here is toward the shorter `sendGlobalChat`. But happy to be convinced otherwise.

src/hooks/useReactions.ts
thecadams: Something along these lines is being worked on in https://github.com/sparkletown/sparkle/pull/744 and this will prevent crashes in all sorts of places in the code. Exciting to see the code get more stable.

thecadams:
@mike-lvov ready for another round of review. Thanks for all the thoughtful comments and for stimulating these discussions, very grateful to have you on the team contributing your opinions, keep it up! =]

thecadams:
[chat and experience context pull request](https://trello.com/c/2qtqY9zt/580-chat-and-experience-context-pull-request)

src/components/molecules/VenueChat/VenueChat.tsx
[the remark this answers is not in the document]
mike-lvov: Valid points. Agreed

src/store/actions/Chat.ts
mike-lvov: Agreed

src/store/actions/Chat.ts
mike-lvov: typeof SEND_GLOBAL_CHAT results into string, which is not as good as it could be. A good workaround this could be either creating 
enum SendChatTypes = {
    
}

or create 
type SendTableChat = "SEND_TABLE_CHAT";
export const SEND_TABLE_CHAT: SendTableChat = "SEND_TABLE_CHAT";

denisdimitrov (approved):
LGTM!

PR #1009, opened 2020-12-14
Fixes chat position; Refactors chat component

mike-lvov:


mike-lvov:
[Search for people actually covers up the search bar](https://trello.com/c/vqUOFAxq/568-search-for-people-actually-covers-up-the-search-bar)

denisdimitrov (approved):
LGTM! Nice catch on the unnecessary selectors and the clean up.

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -22,31 +18,32 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
     WithId<User>
   >();
 
-  const venue = useSelector(currentVenueSelectorData);
-  const partygoers = useSelector(partygoersSelector);
-  const venueEvents = useSelector(venueEventsSelector);
+  const partygoers = useSelector(partygoersSelector) ?? [];
```
thecadams, line 21: Good catch removing these. I'm surprised line 25-27 were not causing errors due to those variables not being used in the component, except in the dependency list below.

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -55,37 +52,33 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
...
-                  </div>
-                </div>
-              );
-            })}
+        <div className="user-search-results">
+          <div className="user-search-result-number">
+            <b>{numberOfSearchResults}</b> search results
           </div>
-        </>
+          {searchResult.map((user) => (
```
thecadams, line 59: nit: consider renaming to `searchResults`

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -55,37 +52,33 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
...
+            <div
+              className="row result-user"
+              key={user.id}
```
thecadams, line 62: Glad to see us using Firestore `uid`s as keys.

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -55,37 +52,33 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
...
+            </div>
+          ))}
+        </div>
       )}
       <UserProfileModal
         userProfile={selectedUserProfile}
-        show={selectedUserProfile !== undefined}
-        onHide={() => setSelectedUserProfile(undefined)}
+        show={!!selectedUserProfile}
+        onHide={clearSelectedUserProfile}
```
thecadams, line 81: Thanks for extracting this out; every render we save improves performance!

src/components/molecules/UserSearchBar/UserSearchBar.scss
```diff
@@ -70,9 +70,8 @@
 .user-search-results {
   position: absolute;
   width: 100%;
-  position: fixed;
   z-index: 380;
-  top: 60px;
+  top: 105%;
```
thecadams, line 74: [thinking, not really a review comment] Just thinking about responsive etc. We are at a point where a standard set of recommendations will be very helpful for us to do responsive layouts really well.

A place that answers questions like: what are the recommended places to use % and when to use pixels, how that affects responsive layouts, etc. etc. would be really handy for the team.

thecadams (approved):
This is looking good Mike, thanks for the contribution. Only a couple nits & questions, but approving so you can merge once you're happy with things.

src/components/molecules/UserSearchBar/UserSearchBar.scss
IvanBuljovcic: In this case, its not due to responsiveness. 105% is telling the child (absolute) component to go 5% under the bottom edge of the parent (relative) component.

Same result would be with `bottom: -5%;`

If pixels were used for this result, you would need to know how tall the parent is and add a few more pixels to push the child out of the box of the parent

src/components/molecules/UserSearchBar/UserSearchBar.tsx
mike-lvov: Good catch, thank you!

PR #1010, opened 2020-12-14
Fixes flatmap error in Edge browser

mike-lvov:
Adds polyfill for flatMap

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -47,6 +47,7 @@ import { FIREBASE_CONFIG } from "settings";
 
 import { VenueTemplateReducers, MiscReducers } from "store/reducers";
 import * as serviceWorker from "./serviceWorker";
+import { activatePolyFills } from "./polyfills";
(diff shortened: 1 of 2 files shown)


mike-lvov:
[React error - object or property does not support flatMap](https://trello.com/c/2GUZofrP/545-react-error-object-or-property-does-not-support-flatmap)

thecadams (approved):
Fantastic addition, thanks for this!

0xdevalias:
@mike-lvov I would have to look to be certain, but we shouldn't have any need to be adding manual polyfills like this. We should be getting them through our babel/browserslist/etc settings and transpilation. I'm guessing this PR was raised due to a specific error showing up in Edge? Do you have any more context/details/etc you're able to share here about it?

mike-lvov:
@0xdevalias yes, there are also screenshots about this bug in the Trello ticket.

PR #1011, opened 2020-12-15
Check for incomplete profile on venue entrance

thecadams:


src/pages/VenueEntrancePage/VenueEntrancePage.tsx
```diff
@@ -36,6 +37,10 @@ export const VenueEntrancePage: React.FunctionComponent<{}> = () => {
     return <Login />;
   }
 
+  if (profile && !isCompleteProfile(profile)) {
+    history.push(`/account/profile?venueId=${venueId}`);
```
denisdimitrov, line 41: Correct me if I'm wrong but I believe we redirect to /in/ after the profile is complete. Which means you are going to skip the entrance page.

thecadams: We will go back to the entrance because `/in/` does a check here: https://github.com/sparkletown/sparkle/blob/staging/src/pages/VenuePage/VenuePage.tsx#L266

denisdimitrov (approved):
LGTM!

PR #1013, opened 2020-12-15
If hostname is not same as redirectUrl, redirect user

IvanBuljovcic:


src/types/Venue.ts
```diff
@@ -105,6 +105,8 @@ export interface VenueConfig {
 
   landingPageConfig: VenueLandingPageConfig; // @debt should this be potentially undefined, or is it guaranteed to exist everywhere?
 
+  redirectUrl?: string;
+
```
mike-lvov, line 109: Do we really need this space here?

mike-lvov (approved):

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -53,6 +53,13 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
   );
   const purchaseHistory = useSelector(userPurchaseHistorySelector);
 
+  const redirectUrl = venue?.config?.redirectUrl ?? "";
```
0xdevalias, line 56: There is no need to set a default value with ?? here. It’s semantically more useful/correct that if it isn’t set, that the variable remains undefined. By setting it to an empty string we’re just kind of shadowing/hiding the reality of it.

in this specific case it doesn’t make a huge difference to the logic/flow of things as it’s checked right below, but this is n anti-pattern that can lead to much bigger/more annoying/harder to track down bugs, so calling it out here so people can learn and make sure not to repeat it on a winder scale.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -53,6 +53,13 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
+  const { hostname } = window.location;
+
+  if (redirectUrl && redirectUrl !== hostname) {
```
0xdevalias, line 59: What value gets set in the DB for this? I would have expected to see code here that checks new URL(valueFromDB).hostname or similar. Have we tested that this works as expected?

0xdevalias: And I guess if the value in the DB is just a hostname, then calling it ‘url’ seems misleading/unintuitive.

PR #1014, opened 2020-12-15
Adds initial integration with Zendesk

mike-lvov:


thecadams:
Hi @mike-lvov @ thanks for the contribution. Just wondering, what's the difference between this approach and including zendesk directly in `index.html`? Are there advantages to doing it in React like this?

src/index.tsx
```diff
@@ -54,6 +54,7 @@ import { User } from "types/User";
 
 import { useSelector } from "hooks/useSelector";
 import { authSelector } from "utils/selectors";
+import { initializeZenDesk } from "utils/zenDesk";
```
thecadams, line 57: nit: `Zendesk` - no capital D. (https://www.sec.gov/Archives/edgar/data/1463172/000119312514137878/d665368ds1.htm)

src/utils/zenDesk.ts
```diff
@@ -0,0 +1,12 @@
...
+const ZENDESK_SCRIPT_SRC =
+  "https://static.zdassets.com/ekr/snippet.js?key=57ad1fbc-d174-4561-8102-b3c8d98436d5";
+
+const ZENDESK_SCRIPT_ID = "ze-snippet";
+
+export function initializeZenDesk() {
+  loadScript(ZENDESK_SCRIPT_SRC, () => {}, {
+    id: ZENDESK_SCRIPT_ID,
+  });
+}
```
thecadams, line 12: nite: per above nit, rename this file to `zendesk.ts`.

src/utils/loadScript.ts
```diff
@@ -0,0 +1,17 @@
+import { HTMLAttributes } from "react";
+
+export function loadScript(
+  src: string,
+  onLoad: () => void,
```
thecadams, line 5: nit: considering it's going into a property called `onload`, is this better off being called `onload` instead of `onLoad`? Open to opinions here.

thecadams (commented):
Looking good, pending my higher level comment on the approach.

thecadams:
From Mike:
This way lets us pass in functions to notify when the script is loaded so we have more control for developers. So this is a superior way to do this one.

src/utils/loadScript.ts
```diff
@@ -0,0 +1,12 @@
+type ScriptProps = "src" | "id";
+type LoadScriptProps = Record<ScriptProps, string>;
+
+export function loadScript(options: LoadScriptProps) {
```
denisdimitrov, line 4: Is there a reason for the ES5 syntax? On the next line you are using `const` I suppose this can become an arrow function and keep it consistent.

denisdimitrov (changes requested):

src/utils/loadScript.ts
denisdimitrov: Same question/logic applies for the function `initializeZendesk`

mike-lvov: Good addition, changed that

mike-lvov:
> Hi @mike-lvov & thanks for the contribution. Just wondering, what's the difference between this approach and including zendesk directly in `index.html`? Are there advantages to doing it in React like this?

I think this is a really good question. As far as I understand, there will be more added to this integration and I think it is useful to have all of these things in one place/file

denisdimitrov (approved):
LGTM!

thecadams (approved):
LGTM!

src/utils/zendesk.ts
```diff
@@ -0,0 +1,13 @@
+import { loadScript } from "./loadScript";
+
+const ZENDESK_SCRIPT_SRC =
+  "https://static.zdassets.com/ekr/snippet.js?key=57ad1fbc-d174-4561-8102-b3c8d98436d5";
```
IvanBuljovcic, line 4: nit: Should this be in here or `settings.ts` ?

src/utils/zendesk.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const ZENDESK_SCRIPT_ID = "ze-snippet";
```
IvanBuljovcic, line 6: same as above

src/utils/zendesk.ts
0xdevalias: Personally, I would have implemented it as an environment variable that is then mapped through, rather than as a hardcoded static. That method aligns more to how we handle existing things like this, and makes the code more flexible when others are deploying their own versions of it.

This would be mapped in https://github.com/sparkletown/sparkle/blob/staging/src/secrets.js (which isn't necessarily the best filename, since most of these values aren't actually secret.. particularly when mapped to the frontend.. might be worth renaming it `env` or similar? Realistically.. this file is actually what most of `settings.ts` would end up as one day I think.. so maybe it should be merged there or similar?)

0xdevalias: Backlogged these improvements in https://github.com/sparkletown/internal-sparkle-issues/issues/100

src/utils/zendesk.ts
0xdevalias: See https://github.com/sparkletown/internal-sparkle-issues/issues/100

0xdevalias:
> @thecadams
> From @mike-lvov:
> This way lets us pass in functions to notify when the script is loaded so we have more control for developers. So this is a superior way to do this one.

> @mike-lvov
> I think this is a really good question. As far as I understand, there will be more added to this integration and I think it is useful to have all of these things in one place/file

It's been a while since I had my 🧠  in this feature/the documentation/etc, but from memory I the conclusion I came to when planning it out was that loading the script in the HTML, and then adding the relevant JS/etc parts to `index.ts`/it's own file that is imported/etc would be the 'best' solution here (as documented on https://github.com/sparkletown/internal-sparkle-issues/issues/65#issuecomment-741539135 + the related slack link there):

> tl;dr: `@tech-leads` i’m going to integrate this with the raw zendesk snippet + maybe create our own helpers to wrap around their ‘raw API’ in a typescript friendly way.

If the current method works and doesn't add any inefficiencies/potential issues to our codebase, then happy to leave it as is for now. But just wanted to capture this here for potential future cleanup/knowledge/etc.

PR #1022, opened 2020-12-16
Event duplication

IvanBuljovcic:
Removed `Math.random()` from `key=` assignment.
This was a possible cause to events being duplicated in the UI


denisdimitrov (approved):
LGTM!

PR #1034, opened 2020-12-16
Show zendesk per venue

thecadams:
This PR will make showing Zendesk's popup conditional upon a venue config setting. Open question is whether we should make it default to showing, rather than hiding.


thecadams (commented):
LGTM!

mike-lvov:
LGTM!

mike-lvov (approved):

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -570,6 +570,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
...
+        <input
+          type="checkbox"
+          id={"showZendesk"}
+          name={"showZendesk"}
+          ref={register}
+        />
+        <span className="slider round"></span>
+      </label>
+    </div>
+  );
```
0xdevalias, line 586: This should be a standalone component rather than a render function. The strings in the JSX don't need the `{}` around them either, as you can just use the string directly.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -816,6 +831,7 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
           HAS_GRID_TEMPLATES.includes(templateID) &&
           renderShowGridToggle()}
         {renderShowBadgesToggle()}
+        {renderShowZendeskToggle()}
```
0xdevalias, line 834: As above, this should just be rendering the extracted component rather than calling a function here.

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
```diff
@@ -20,6 +21,12 @@ export const VenueEntrancePage: React.FunctionComponent<{}> = () => {
   useConnectCurrentVenue();
   const venue = useSelector(currentVenueSelectorData);
 
+  useEffect(() => {
+    if (venue?.showZendesk) {
+      showZendeskWidget();
+    }
+  }, [venue]);
```
0xdevalias, line 28: Should this have an `else` clause where it calls `hideZendeskWidget`? As is, if we show the widget, then the `venue` changes to one where Zendesk should be hidden, I believe it will remain visible.

Also, since this code snippet is repeated in a number of files, there is probably a more DRY way we can implement the pattern. Either by making the check in a 'higher' component in the tree if possible, or at the very least by wrapping up this logic as a custom hook.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -98,6 +99,12 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
     }
   }, [shouldOpenPaymentModal, isAuthenticationModalOpen]);
 
+  useEffect(() => {
+    if (venue?.showZendesk) {
+      showZendeskWidget();
+    }
+  }, [venue]);
+
```
0xdevalias, line 107: As above: Should this have an `else` clause where it calls `hideZendeskWidget`?

Also, since this code snippet is repeated in a number of files, there is probably a more DRY way we can implement the pattern. Either by making the check in a 'higher' component in the tree if possible, or at the very least by wrapping up this logic as a custom hook.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -250,6 +251,12 @@ const VenuePage = () => {
     }
   }, [user, profile, venueId, venueTemplate, mixpanel]);
 
+  useEffect(() => {
+    if (venue?.showZendesk) {
+      showZendeskWidget();
+    }
+  }, [venue]);
+
```
0xdevalias, line 259: As above: Should this have an `else` clause where it calls `hideZendeskWidget`?

Also, since this code snippet is repeated in a number of files, there is probably a more DRY way we can implement the pattern. Either by making the check in a 'higher' component in the tree if possible, or at the very least by wrapping up this logic as a custom hook.

src/utils/zendesk.ts
```diff
@@ -1,13 +1,42 @@
...
+  if ((window as any).zE) {
+    const showZendeskForThisPath = ZENDESK_URL_PREFIXES.find((prefix) =>
+      window.location.pathname.startsWith(prefix)
+    );
+
+    if (!showZendeskForThisPath) {
+      (window as any).zE("webWidget", "hide");
+    }
+  }
+};
```
0xdevalias, line 30: This logic + the value in `ZENDESK_URL_PREFIXES` seems to imply that we are only ever going to show the Zendesk widget on the admin pages? Is that the desired logic?

My understanding was that we wanted this widget to face event-going users as well, though be toggleable on/off per event (for situations where we don't intend to provide front-line support for users)

src/utils/zendesk.ts
```diff
@@ -1,13 +1,42 @@
...
+
+export const showZendeskWidget = () => {
+  if ((window as any).zE) {
+    (window as any).zE("webWidget", "show");
+  } else {
+    setTimeout(() => {
+      if ((window as any).zE) {
+        (window as any).zE("webWidget", "show");
+      }
+    }, ZENDESK_WARMUP_DELAY_MS);
```
0xdevalias, line 40: This feels a little hacky. Is there a need for the warmup delay? If so, what are the details about when/why it was happening, as there may be a better solution.

src/utils/zendesk.ts
```diff
@@ -1,13 +1,42 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { ZENDESK_URL_PREFIXES } from "settings";
 import { loadScript } from "./loadScript";
 
 const ZENDESK_SCRIPT_SRC =
   "https://static.zdassets.com/ekr/snippet.js?key=57ad1fbc-d174-4561-8102-b3c8d98436d5";
```
0xdevalias, line 7: We should extract this into env variables + map through as per https://github.com/sparkletown/internal-sparkle-issues/issues/100

0xdevalias (commented):
Retroactive review with a bunch of comments about future cleanup.

src/utils/loadScript.ts
```diff
@@ -1,5 +1,6 @@
-type ScriptProps = "src" | "id";
-type LoadScriptProps = Record<ScriptProps, string>;
+type ScriptProps = "src" | "id" | "onload";
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type LoadScriptProps = Record<ScriptProps, any>;
```
0xdevalias, line 3: Do we really want to use `any` here? No better value?

src/utils/zendesk.ts
```diff
@@ -1,13 +1,42 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
0xdevalias, line 1: If we have a legitimate need to disable checks for `@typescript-eslint/no-explicit-any` here, rather than finding a better type to use that isn't `any`, then we should be doing it with the minimal possible scope (eg. disable it only for the line/lines that it occurs on, not the whole file)

src/pages/Admin/Venue/DetailsForm.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/101

src/pages/Admin/Venue/DetailsForm.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/101

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/103

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/103

src/pages/VenuePage/VenuePage.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/103

src/utils/zendesk.ts
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/104

src/utils/loadScript.ts
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/105

src/utils/zendesk.ts
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/105

src/utils/zendesk.ts
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/104

PR #1036, opened 2020-12-17
Restore venueChats after removal of contexts

thecadams:
This fix brings back the `useFirestoreConnect` in `ChatContext` which was filling out the `venueChats`. It was not brought across in #956 nor in #1007, but is referenced in the code.

### DIFF
--- a/src/components/organisms/Chatbox/Chatbox.tsx
+++ b/src/components/organisms/Chatbox/Chatbox.tsx
@@ -11,6 +11,8 @@ import { User } from "types/User";
 import ChatMessage from "components/molecules/ChatMessage";
 import { useUser } from "hooks/useUser";
 import { useSelector } from "hooks/useSelector";
+import { useVenueId } from "hooks/useVenueId";
+import { useConnectVenueChats } from "hooks/useConnectVenueChats";
 import { useFirestoreConnect } from "react-redux-firebase";
 import { WithId } from "utils/id";
 import { DEFAULT_PARTY_NAME, DEFAULT_PROFILE_IMAGE } from "settings";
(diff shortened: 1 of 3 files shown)


src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -69,6 +71,8 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
         }
       : undefined
   );
+  const venueId = useVenueId();
+  useConnectVenueChats(venueId);
```
mike-lvov, line 75: I think this should be wrapped into a useEffect

src/hooks/useConnectVenueChats.ts
```diff
@@ -0,0 +1,14 @@
+import { useFirestoreConnect } from "react-redux-firebase";
+
+export const useConnectVenueChats = (venueId?: string) => {
+  useFirestoreConnect(
+    venueId
```
denisdimitrov, line 5: As a suggestion, maybe we can try with guards instead of conditionals. That way useFirestoreConnect won't be called at all. I suppose no call is better than a function call with undefined.

```jsx
export const useConnectVenueChats = (venueId?: string) => {
  if (!venueId) {
    return;
  }

  useFirestoreConnct(query)
}
```

denisdimitrov (approved):
LGTM!

src/hooks/useConnectVenueChats.ts
thecadams: Good suggestion. This has the problem that you can't conditionally call a hook, unfortunately. So we have to put it all inside the `useFirestoreConnect` call.

PR #1040, opened 2020-12-17
Prevent multiple listeners

thecadams:
Both these settings affect whether duplicate listeners can be attached to a given path.

This affects the behavior of this function: https://github.com/prescottprue/redux-firestore/blob/master/src/actions/firestore.js

With this change, one heavier venue becomes responsive in 15 seconds vs. 60 seconds without.

We should test whether some other data goes missing with this change present.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -81,6 +81,8 @@ const stripePromise = loadStripe(STRIPE_PUBLISHABLE_KEY ?? "");
 const rrfConfig = {
   userProfile: "users",
   useFirestoreForProfile: true,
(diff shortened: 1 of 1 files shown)


thecadams:
Based on https://github.com/sparkletown/sparkle/blob/staging/package.json#L69 I am fairly sure we are using this version: https://github.com/prescottprue/redux-firestore/blob/v0.14.0/src/actions/firestore.js

denisdimitrov (approved):
LGTM!

mike-lvov (approved):

PR #1043, opened 2020-12-17
make profile popover responsive

denisdimitrov:
Remove previously set height and add fancy design for the scrollbar

### DIFF
--- a/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
+++ b/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
@@ -4,12 +4,12 @@
   &.popover {
     &::-webkit-scrollbar-track {
       border-radius: 3px;
-      background-color: rgba(0, 0, 0, 0);
+      background-color: $white;
     }
 
     &::-webkit-scrollbar {
       width: 8px;
-      background-color: rgba(0, 0, 0, 0);
+      background-color: $white;
     }
 
     &::-webkit-scrollbar-thumb {
--- a/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
+++ b/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
@@ -2,7 +2,20 @@
 
 #profile-popover {
   &.popover {
-    height: unset;
+    &::-webkit-scrollbar-track {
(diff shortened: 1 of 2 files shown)


denisdimitrov:
[Scrollability of the profile page not functional now we have badges in the operation](https://trello.com/c/cRrq8Ekj/588-scrollability-of-the-profile-page-not-functional-now-we-have-badges-in-the-operation)

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
```diff
@@ -2,7 +2,20 @@
 
 #profile-popover {
   &.popover {
-    height: unset;
+    &::-webkit-scrollbar-track {
```
IvanBuljovcic, line 5: So, *BLIP* for _-moz-_ or _-ms-_ users, right?

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
```diff
@@ -2,7 +2,20 @@
...
+      border-radius: 3px;
+      background-color: rgba(0, 0, 0, 0);
```
IvanBuljovcic, line 7: Variable?

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
denisdimitrov: When I first started we weren't supposed to support multiple browsers and the users should always use Chrome, maybe something changed. @thecadams Is the project running on Chrome only?

IvanBuljovcic: I know its only Chrome (for now), its not a required change to include different browser prefixes, its just something to keep in mind long term

IvanBuljovcic (approved):

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
thecadams: Chrome only; tolerable on Safari; this is fine to include.

thecadams (approved):
LGTM!

PR #1045, opened 2020-12-17
Reduce the amount of noise in PRs by Codeclimate

mike-lvov:


thecadams (approved):
Seems reasonable to me, and we can always change it later. LGTM!

0xdevalias (commented):
Setting these at the global level will override all language specific thresholds. The docs suggest that perhaps we want to configure it on a per language level.

Curious, what is the current default that is being overridden here? From memory it was 5 per PR, so reducing it to 4 seems almost pointless?

I’d be curious to know specifically what aspects (or patterns of issue it’s picking up) that you’re finding ‘too spammy’?

there is also the ability to turn off its commenting on PR’s entirely (while still keeping its single ‘summary comment’ that links through to the report/etc on the site itself.

I made the explicit choice to turn on the code commenting part, as I feel like it brings the issues more to the forefront of peoples mind/more likely to work on resolving them. I’m open to us changing that decision if it’s not working for us, but I would like to understand more about the ‘why’ of it and the underlying concerns, as to date I feel like a lot of the ‘spam’ (at least on PR’s I’ve seen) has largely all been relevant, we’ve just been in such a firefighting/triaging mode that instead of taking the time to address the concerns it raises, we don’t have the mental (or literal) bandwidth to make those additional changes, so we ignore it, and then if we push new code with the same issues it just brings it up again and again.

So yeah, if there are any particular ‘rules’ that you feel are popping up too often/as false positives/etc, that would be useful to know.

PR #1046, opened 2020-12-18
Get user badges for specific venue

IvanBuljovcic:


thecadams (approved):
Used this to extract badges, LGTM!

PR #1049, opened 2020-12-20
Add 'nametag' option to optionally display user's names alongside their avatars

margulies:
Add 'nametag' option to optionally display user's names alongside their avatars

---

**@0xdevalias Edit:** If we implement the 'on hover' version described in my reviews here (see https://github.com/sparkletown/sparkle/pull/1049#issuecomment-781740138), then this would fix https://github.com/sparkletown/internal-sparkle-issues/issues/208


.firebaserc
```diff
@@ -1,7 +1,7 @@
 {
   "projects": {
     "default": "co-reality-map",
-    "staging": "co-reality-staging",
+    "staging": "ohbm2021",
```
0xdevalias: This part will need to be fixed up before this will be able to be merged

.gitignore
```diff
@@ -46,3 +46,4 @@ reporting/cookies.json
 
 #IDEs
 .idea/
+workspace.code-workspace
```
0xdevalias: Looking at https://code.visualstudio.com/docs/editor/workspaces it seems like the general patterns for VS Code workspaces are `.code-workspace` or `foo.code-workspace`, so we should be able to generalise this `.gitignore` line to `*.code-workspace` I think, to cover all the bases.

package.json
```diff
@@ -144,7 +144,7 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.20.5",
     "eslint-plugin-react-hooks": "^4.0.8",
-    "firebase-admin": "^9.1.1",
+    "firebase-admin": "^9.4.1",
```
0xdevalias: Are these changes required for this feature specifically? Or just 'came along' from your local changes when you opened the PR? If the latter, it would be good to get the changes here + in `package-lock.json` cleaned up/reverted for this PR.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -794,6 +794,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </div>
   );
 
+  const renderNametagsToggle = () => (
```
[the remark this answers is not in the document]
0xdevalias: Looks like you have a `renderNametagsToggle` on line 779, and then again here on line 821. I'm guessing this was a merge issue when updating this branch at some point.

Could you please pick the one that needs to stay, and remove the other one.

---

Also, as a general style note (I know we're not doing it in this file currently, and it's fine if you just want to continue to match the style pattern here), we're looking to move away from these 'render functions' where possible, in favour of clean abstractions using small components. These would generally be created in a separate file, named the same as the component. That way we can better 'isolate the boundaries' of each 'snippet', and also will let us avoid having such unwieldy huge files like this one has grown to (an almost 1000 line react component file is.. suboptimal to say the least)

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -803,6 +818,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
...
+    <div className="toggle-room">
+      <h4 className="italic input-header">Display profile names on avatars</h4>
+      <label id="showNametags" className="switch">
+        <input
+          type="checkbox"
+          id="showNametags"
+          name="showNametags"
+          ref={register}
+        />
```
0xdevalias: We generally don't need to set the `id` prop on element in React, as it's barely ever used for anything. Technically an `id` must be unique within the page, so having it set on both the label and input here is an error I believe (though one that most browsers will silently ignore)

An exception to this is that using the 'label for' pattern of associating a label with its input requires the input to have an `id` set. Though another way of associating them is to nest the `input` element directly within the `label` element, as you have done here. So I don't believe either will need an `id` here.

Further Reading:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

0xdevalias:
```suggestion
      <label className="switch">
        <input
          type="checkbox"
          name="showNametags"
          ref={register}
        />
```

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -31,6 +33,7 @@ const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
   setSelectedUserProfile,
   reactionPosition,
   user,
+  showNametags,
```
0xdevalias, line 36: Since this is an optional parameter, you can set a default value for it at this point, which can simplify later code.

```suggestion
  showNametags = true,
```

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
```diff
@@ -10,4 +11,6 @@ export interface UserProfilePictureProp {
   setSelectedUserProfile: (user: WithId<User>) => void;
   user: WithId<User>;
   reactionPosition?: "right" | "left" | undefined;
+  showNametags?: boolean;
+  currentVenue?: WithId<AnyVenue>;
```
0xdevalias, line 15: It looks like this param isn't used in the code anymore? I'm guessing this is leftover from before you added `const { currentVenue } = useConnectCurrentVenueNG(venueId);`

```suggestion
```

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -101,8 +105,12 @@ const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
           className={avatarClassName}
           backgroundImage={pictureUrl}
           style={{ ...avatarStyle }}
-        />
-
+        >
+          {showNametags &&
+            (currentVenue?.showNametags ?? SHOW_AVATAR_NAMETAGS) && (
```
0xdevalias, line 111: You could make this line a little cleaner by extracting the boolean logic out into a `const` higher up in this component. Probably something like

```typescript
const shouldShowNameTags = showNametags && (currentVenue?.showNametags ?? SHOW_AVATAR_NAMETAGS);
```

Then here you could just have:
```suggestion
          {shouldShowNametags && (
```

0xdevalias: I'm not 100% what the best logic is here though.

Do you think the local `showNameTags` prop should override the venue-wide setting? My guess is that it was added so that we can force the name tags not to show in certain locations.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -152,12 +160,15 @@ const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
...
+    showNametags,
+    currentVenue,
   ]);
 };
 
 UserProfilePicture.defaultProps = {
   avatarClassName: "profile-icon",
   miniAvatars: false,
+  showNametags: true,
 };
```
0xdevalias, line 173: Ah.. I see that we're using another non-standard (for our codebase) pattern for defaults here. I believe our preference is to add them using the 'inline' ES6 pattern that I suggested above.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: And based on that, perhaps there is a better name we can use for this prop, that better conveys the fact that it controls whether the venue setting is respected/not. As it currently is, I might assume that by setting it to `true`, it should always display, so I would be confused when it didn't (because the venue setting was off)

functions/venue.js
```diff
@@ -550,6 +551,10 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.radioStations = [data.radioStations];
   }
 
+  if (typeof data.showNametags === "boolean") {
+    updated.showNametags = data.showNametags;
+  }
+
   updated.requiresDateOfBirth = data.requiresDateOfBirth || false;
```
0xdevalias, line 558: [OOS] @denisdimitrov @sparkletown/engineering Is this a bug? I would have expected that it would need to follow the same `typeof` pattern that the other `boolean`s do here?

If not, do you know why it's different from the rest?

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -62,6 +62,26 @@ export const Container = styled.div`
...
+    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.2);
+    text-align: center;
+    font-size: 0.8rem;
+    line-height: 0.9;
+    opacity: 1;
+    left: 50%;
+    margin-right: -50%;
+    transform: translate(-50%, 0);
+    min-width: 70%;
+  }
```
0xdevalias, line 84: [note] I haven't deeply reviewed these styles

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -803,6 +818,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </div>
   );
 
+  const renderNametagsToggle = () => (
+    <div className="toggle-room">
+      <h4 className="italic input-header">Display profile names on avatars</h4>
```
0xdevalias: [NB] Not a big thing one way or another, just musing if this wording would work better.

```suggestion
      <h4 className="italic input-header">Display user's names on their avatars</h4>
```

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -878,6 +909,8 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
           HAS_REACTIONS_TEMPLATES.includes(templateID) &&
           renderSeatingNumberInput()}
 
+        {renderNametagsToggle()}
```
0xdevalias: Duplicated with line 900

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -168,12 +168,14 @@ export const validationSchema = Yup.object()
...
     }),
 
+    showNametags: Yup.bool().notRequired(),
     owners: Yup.array<string>().notRequired(),
     placementRequests: Yup.string().notRequired(),
     adultContent: Yup.bool().required(),
     bannerMessage: Yup.string().notRequired(),
     parentId: Yup.string().notRequired(),
     showReactions: Yup.bool().notRequired(),
+    showNametags: Yup.bool().notRequired(),
```
0xdevalias, line 178: Duplicated with line 171

src/settings.ts
```diff
@@ -473,6 +473,9 @@ export const DEFAULT_REDIRECT_URL = IS_BURN ? "/enter" : HOMEPAGE_URL;
 // git clone git@github.com:sparkletown/sparkle-relay && cd sparkle-relay && docker-compose up
 export const DEFAULT_WS_RELAY_URL = "ws://localhost:8080/";
 
+// Default to display profile name on top of avatar (overriden by venue parameter)
+export const SHOW_AVATAR_NAMETAGS = false;
```
0xdevalias, line 477:
```suggestion
export const DEFAULT_SHOW_AVATAR_NAMETAGS = false;
```

src/types/venues.ts
```diff
@@ -142,7 +143,9 @@ export interface Venue {
   videoAspect?: VideoAspectRatio;
   termsAndConditions: TermOfService[];
   showRadio?: boolean;
+  showNametags?: boolean;
   showBadges?: boolean;
+  showNametags?: boolean;
```
0xdevalias: Duplicated on line 73, 146, 148

0xdevalias (changes requested):
Thanks for the PR @margulies, neat feature!

A few notes/bits and pieces that need cleanup before this can be merged.

---

We also chatted briefly about how this feature could potentially have multiple options, similar to `RoomVisibility`, where one 'mode' would be as implemented here, and another 'mode' could be showing the nametag when hovering over the user's avatar.

If you're up for experimenting with what the 'hovered nametag' looks like, I would be interested to hear which you think is 'better', or if it would make sense to have both options available. If the latter, it would be cool to rework the admin controls to be a dropdown that lets us choose between 'off'/'none', 'inline', and 'hover' (or whatever naming seems appropriate)

0xdevalias:
RE: my review in https://github.com/sparkletown/sparkle/pull/1049#pullrequestreview-593727861

Looks like you beat me finishing the review to push some updates! Will have another quick look through again now.

functions/package-lock.json
```diff
@@ -5307,9 +5307,9 @@
       }
```
0xdevalias, line 5307: As there are no changes in the `package.json` now, it would be good to revert the changes in this file as well.

src/types/venues.ts
```diff
@@ -143,6 +144,7 @@ export interface Venue {
   termsAndConditions: TermOfService[];
   showRadio?: boolean;
   showBadges?: boolean;
+  showNametags?: boolean;
```
0xdevalias, line 147: Still seems to be a duplicate here with line 73

0xdevalias (changes requested):
Re-review after your recent changes

src/types/venues.ts
margulies: Thanks for the comments and suggestions. 
This duplicate seems to be the pattern used here, where variables are declared for the Venue and Venue_v2_AdvancedConfig separately.

0xdevalias: Ah, cool, that makes sense (I was only looking at the diff changes specifically, hadn't fully expanded the surrounding code)

This should be fine as is then :)

functions/package-lock.json
margulies: Thanks! Just pushed a commit that should take care of that

src/settings.ts
margulies: Good call

src/pages/Admin/Venue/DetailsValidationSchema.ts
margulies: Resolved. Removed line 171

src/pages/Admin/Venue/DetailsForm.tsx
margulies: fixed

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
margulies: removed line

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
margulies: done

src/pages/Admin/Venue/DetailsForm.tsx
margulies: Thanks! Fixed

.gitignore
margulies: That would be great — I'll update my filename to `.code-workspace` accordingly

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
margulies: This is a good point. I vaguely recall implementing this to get around an issue of the nametags showing up in chat window. I've implemented the change you suggested here, but think about better terminology for the variables

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
margulies: Just to ensure I'm following correctly, after implementing the suggestion you made above, I should simply delete this line?

0xdevalias: Yup. Though if you want to take it one step further, you can move all of these defaults to the same place I suggested the other change, and then remove `UserProfilePicture.defaultProps` entirely.

0xdevalias:
@margulies Recent changes are looking good. Once you're done with everything you're planning here, click the re-request review button and i'll give this another look through :)

image

---

> We also chatted briefly about how this feature could potentially have multiple options, similar to RoomVisibility, where one 'mode' would be as implemented here, and another 'mode' could be showing the nametag when hovering over the user's avatar.
>
> If you're up for experimenting with what the 'hovered nametag' looks like, I would be interested to hear which you think is 'better', or if it would make sense to have both options available. If the latter, it would be cool to rework the admin controls to be a dropdown that lets us choose between 'off'/'none', 'inline', and 'hover' (or whatever naming seems appropriate)

@margulies Curious what your thoughts on this are too.

margulies:
Thanks — I think I've got your suggested fixes in order now. I'll work on the hovering nametags idea a bit before the next round of review

0xdevalias:
> Thanks — I think I've got your suggested fixes in order now. I'll work on the hovering nametags idea a bit before the next round of review

@margulies Sounds good. Just hit the re-request review button when it's ready for another round :)

margulies:
@0xdevalias Thought you'd get a kick out of this hover feature. It's not necessarily right for this PR (so I put it in a new branch), but wanted to share: https://github.com/margulies/sparkle/tree/feature/avatar-hover

0xdevalias:
> Thought you'd get a kick out of this hover feature. It's not necessarily right for this PR (so I put it in a new branch), but wanted to share

Haha, that's neat!

margulies:
@0xdevalias Just pushed a reworked version of the nametag which includes three options: none, inline, inline/hover. Happy for any suggestions for improvement.

functions/venue.js
denisdimitrov: [OOS] @0xdevalias This particular line will work correctly because there is a fallback value and if `requiresDateOfBirth` is `undefined` this will fallback to `false` and the firebase will accept it. If we have a fallback value there is no need for the `typeof` syntax or `if` statements at all.

The newly added code above can be changed to 
```tsx
updated.showNametags = data.showNametags || false
```
and it would still work correctly in every single case.

My suggestion is to replace all of the `typeof` checks with `updated.value = data.value || FALLBACK_VALUE` that way we can get rid of all of the `if` statements. This can be added as a future PR with clean up or something like that.

Keep in mind that some of the other fields are used only in certain venue templates and they should be in a switch case and added only for that venue template. Iframes, columns, rows, showGrid, etc...

denisdimitrov: Simply removing the `typeof` and keeping the `if` statement wouldn't work either because the falsy values won't pass the statement, Ex: `false` for boolean, `""` for string, `0` for number and etc.

```tsx
if (updated.showSomething) {
updated.showSomething= data.showSomething
}
```

If `updated.showSomething` is `false` the statement will skip and the value in the firebase will remain `true`

@0xdevalias You probably already saw this in one of my latest pull requests, I'm commenting again just in case someone missed it or was wondering why use `typeof`

0xdevalias: Cool. Thanks for the insight here :)

0xdevalias:
@margulies Sorry for letting this sit unreviewed for so long.. busy times. I've added it to my 'open source community contributions' project board, so hopefully will be able to get to it soon.

It seems there are a few conflicts that will need fixing up as well.

margulies:
This feature has been refactored and updated with the current staging branch. Happy for any comments/suggestions

mike-lvov (approved):
LGTM after a set of polishes/additions. @0xdevalias this one worth retro reviewing

PR #1055, opened 2020-12-23
Fix InformationLeftColumn component styling + cleanup/refactor related aspects

0xdevalias:
Relates to https://trello.com/c/IeTpyaRm and https://trello.com/c/H8NxSYRh

Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/190
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/286

In the Jazz Bar/Art Piece/similar templates, the width of the left hand 'slide out drawer' was broken due to conflicting styles:

image

This PR fixes that issue, as well as refactoring/cleaning up a bunch of the related tech debt.


0xdevalias (commented):
Best reviewed with 'hide whitespace changes' enabled

src/components/organisms/InformationLeftColumn/InformationLeftColumn.tsx
```diff
@@ -1,83 +1,101 @@
...
+
+    setExpanded((prev) => !prev);
+  }, []);
+
+  // Expose internal state/controls to parent components via ref
+  useImperativeHandle(controlsRef, () => ({
+    isExpanded,
+    setExpanded,
+    toggleExpanded,
+  }));
```
0xdevalias, line 58: FYI: This pattern allows a direct parent to gain access to 'internal state/controls' as defined/provided by a component.

- https://reactjs.org/docs/hooks-reference.html#useimperativehandle

src/components/molecules/DonatePopUp/DonatePopUp.tsx
```diff
@@ -1,26 +1,32 @@
...
+
+import InformationCard from "components/molecules/InformationCard";
+
+import "./DonatePopUp.scss";
 
 export const DonatePopUp = () => {
-  const [isLeftColumnExpanded, setIsLeftColumnExpanded] = useState(false);
+  const controlsRef = useRef<InformationLeftColumnControls>(null);
+
+  const { toggleExpanded } = controlsRef.current ?? {};
```
0xdevalias, line 21: FYI: This pattern allows a direct parent to gain access to 'internal state/controls' as defined/provided by a component.

src/components/molecules/DonatePopUp/DonatePopUp.tsx
```diff
@@ -1,26 +1,32 @@
...
 
   return (
-    <InformationLeftColumn
-      venueLogoPath={"heart"}
-      isLeftColumnExpanded={isLeftColumnExpanded}
-      setIsLeftColumnExpanded={setIsLeftColumnExpanded}
-    >
+    <InformationLeftColumn ref={controlsRef} iconNameOrPath="heart">
```
0xdevalias, line 24: FYI: This pattern allows a direct parent to gain access to 'internal state/controls' as defined/provided by a component.

0xdevalias (commented):
Point out the new pattern using useImperativeHandle

src/components/organisms/InformationLeftColumn/InformationLeftColumn.tsx
```diff
@@ -1,83 +1,101 @@
...
-  setIsLeftColumnExpanded: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
-const InformationLeftColumn: React.FunctionComponent<PropsType> = ({
-  venueLogoPath,
-  children,
-  isLeftColumnExpanded,
-  setIsLeftColumnExpanded,
-}) => {
+export interface InformationLeftColumnControls {
```
mike-lvov, line 35: export type

src/components/organisms/InformationLeftColumn/InformationLeftColumn.tsx
mike-lvov: I feel a bit controversial about this pattern. The example shown in the React lib seems to be a very specific use-case when you can't do it any other way.

> As always, imperative code using refs should be avoided in most cases

Your case seems to be quite straightforward on the other hand. I would probably think twice before introducing patterns like this into our codebase. One way to have it implemented would be to store the UI states in the redux? It can sound like overkill, but this way we can ensure we have consistency.

0xdevalias: - https://levelup.gitconnected.com/reacts-useimperativehandle-made-simple-81035a21eef0
  - > The general React style is to have a unidirectional flow of data and logic; That is to say, you should pass functions and data down through props and a component should only ever be able to access what is passed in as props. In cases where bidirectional or lateral dataflow is needed, we can use libraries such as Redux or React’s context API.
  - > **However, in some cases, importing redux or using context is simply just overkill — This is where useImperativeHandle comes in.** This will provide us a lightweight solution to have bidirectional flow.
  - > It’s not something you’ll use often but in rare cases can be incredibly useful and prevent you from having to overreach for extreme solutions like redux or rack your brain coming up with solutions like passing in multiple props, stateUpdaters and isolated error messages — the later being the hardest.
  - The example they describe here is basically equivalent to how I have used it.
  
---

I don't believe it's an inherently bad pattern to have/use selectively within our codebase. It's definitely not the right answer for every situation (for example if you needed to be able to trigger this functionality from somewhere outside of the direct parent, it wouldn't make sense to use this pattern)

If we need access to it from multiple places, then would generally either use React's Context, or put the state in Redux. Particularly with our current redux patterns, the boilerplate of doing that is a huge amount of overhead, for IMO very little good reason. Moving things to global state that can exist more than once at the same time also necessitates a strategy for splitting those up, so that their global state doesn't clash, which adds additional boilerplate again.

I believe among the usages I saw, there was only 1 place it was used where we even need the parent to be able to 'get' the controls that we are exposing. My usage here is sort of in a similar yet different way to the 'functions as children' pattern (which I believe tends to have it's own performance/usage challenges, that from what I remember of working with it in the past, are more annoying than the API/usage i've got here):

- https://codedaily.io/tutorials/6/Using-Functions-as-Children-and-Render-Props-in-React-Components

The alternative to this (and the way it was before I changed it), is that every single place that wants to use this component, also needs to set up/manage the state for it, just so that it can open/close/etc; and then pass the state setter/getters into the component as props; which is something that can be seen in quite a few places across our codebase at the moment (and even more places in the plast), and it adds needless complexity and boilerplate to handle what is essentially an 'implementation detail' of the component. Solving it once within 

---

**tl;dr** of the above, I don't believe it's an inherently bad pattern, though it is one that we should consider and weigh up if/when we think it would be good to use in other places. I definitely don't think it's a bad enough pattern to block getting this user-facing bugfix/cleanup merged, that has been outstanding since at least December last year, and had been requested to be fixed multiple times since then (including most recently yesterday)

src/components/molecules/InformationCard/InformationCard.tsx
```diff
@@ -1,18 +1,20 @@
 import React from "react";
+import classNames from "classnames";
+
 import "./InformationCard.scss";
 
-interface PropsType {
+export interface InformationCardProps {
```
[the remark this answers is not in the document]
0xdevalias:
```suggestion
export type InformationCardProps {
```

src/components/organisms/InformationLeftColumn/InformationLeftColumn.tsx
0xdevalias:
```suggestion
export type InformationLeftColumnControls = {
```

0xdevalias: I believe my usage here is correct, as per the following:

- https://blog.logrocket.com/types-vs-interfaces-in-typescript/

> **Types vs. interfaces**
>
> The difference between types and interfaces in TypeScript used to be more clear, but with the latest versions of TypeScript, they’re becoming more similar.
> 
> **Interfaces** are basically a way to **describe data shapes, for example, an object.**
> 
> **Type** is a definition of a type of data, for example, **a union, primitive, intersection, tuple, or any other type.**

src/components/molecules/InformationCard/InformationCard.tsx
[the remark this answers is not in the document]
0xdevalias: I believe my usage here is correct as per https://github.com/sparkletown/sparkle/pull/1055#discussion_r569897727

src/components/organisms/InformationLeftColumn/InformationLeftColumn.tsx
0xdevalias: Added this to the documentation PR in https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074

src/components/molecules/InformationCard/InformationCard.tsx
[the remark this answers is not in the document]
0xdevalias: Added this to the documentation PR in https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074

mike-lvov (approved):
You can merge it since there is nothing blocked, but I would continue the discussions

src/components/molecules/InformationCard/InformationCard.tsx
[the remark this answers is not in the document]
mike-lvov: Since there is no difference, why would we divide them? Let's just get rid of interfaces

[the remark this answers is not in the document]
0xdevalias: There is a difference.. the full article spells out the exact specifics, as I believe also the excerpts above do too..?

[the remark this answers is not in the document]
mike-lvov: If you don't plan on using `declaration merging`, which I would be strongly against, having 2 of them seems rather redundant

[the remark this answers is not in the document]
0xdevalias: Let's discuss further on slack: https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1612427117189100

PR #1057, opened 2020-12-23
[OUTDATED] Fix avatar map performance 

mike-lvov:
This is a short-term fix. Unfortunately, the way this template is developed is inefficient and has to be radically changed. Probably, we will use canvas as the main technology to boost up the performance


src/components/templates/Camp/hooks/useMapGrid.tsx
```diff
@@ -9,8 +9,10 @@ import { MapRow } from "../../../molecules/MapRow";
 interface UseMapGrid {
   showGrid?: boolean;
   userUid?: string;
-  columnsArray: JSX.Element[];
-  rowsArray: JSX.Element[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  columnsArray: any[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  rowsArray: any[];
```
0xdevalias, line 15: Definitely don't want to be using `any`/disabling the eslint rules like this (assuming this is more than just a temporary 'work in progress' change)

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -73,65 +100,80 @@ export const Map: React.FC<PropsType> = ({
...
       };
-      const firestore = firebase.firestore();
-      firestore
+
+      await firestore
         .doc(doc)
         .update(update)
         .catch(() => {
           firestore.doc(doc).set(update);
         });
```
0xdevalias, line 145: I know this is a pattern we use in a lot of places currently, but this "try and `update`, then catch the error if it doesn't exist and just `set` it instead" is an anti-pattern/tech-debt in our codebase in my opinion.

I would have to check the firebase docs again to be sure, but I believe there is a function that just handles the 'upsert' pattern anyway, without resorting to this sort of hack. Or if not, we could check if the record exists up front, then choose whether to `update` or `insert` as appropriate.

0xdevalias (commented):
Only did a very quick skim through this, plan to do a more detailed review later on, but wanted to drop a couple of notes here based on what stood out to me so far.

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: Agreed with your that this looks hacky. However, I'd change this in `refactoring` PR, changing all the occurrences of this across our codebase. Will leave it as is in this branch.

0xdevalias: backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/125

src/components/templates/Camp/hooks/useMapGrid.tsx
```diff
@@ -9,8 +9,9 @@ import { MapRow } from "../../../molecules/MapRow";
 interface UseMapGrid {
   showGrid?: boolean;
   userUid?: string;
-  columnsArray: JSX.Element[];
-  rowsArray: JSX.Element[];
+  // FIXME: Convert to number only, when Camp is deleted
+  columnsArray: (number | JSX.Element)[];
+  rowsArray: (number | JSX.Element)[];
```
0xdevalias, line 14: Why do we use `number` here now?

src/components/molecules/MapRow/MapRow.tsx
```diff
@@ -24,7 +24,7 @@ interface MapRowProps {
   ) => void;
 }
 
-export const MapRow: React.FC<MapRowProps> = ({
+export const _MapRaw: React.FC<MapRowProps> = ({
```
0xdevalias, line 27: Since we're using the `_` to imply that it is not the main component that should be used, I think it makes sense to keep it named the same?

```suggestion
export const _MapRow: React.FC<MapRowProps> = ({
```

0xdevalias: `MapRow` is also technically an inaccurate name for this, based on a failed assumption when I was originally refactoring the legacy code. This component actually refers to a 'grid cell/bubble/seat/etc' on the map, not a row.

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -74,3 +70,5 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
     />
   );
 };
+
+export const MapPartygoerOverlay = React.memo(_MapPartygoerOverlay);
```
0xdevalias, line 74: When using `React.memo` we have to ensure that the component is 'pure', with only changes in it's props causing changes any any of the child components below it. Is that the case for this component + all it's children? (eg. `UserProfilePicture`, etc)

If so, awesome! And then also I think the individual `useMemo`'s within it would be redundant/add unneeded overhead.

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -42,7 +39,6 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
       height: `${avatarHeight}%`,
       position: "absolute",
       cursor: "pointer",
-      transition: "all 1400ms cubic-bezier(0.23, 1 ,0.32, 1)",
```
0xdevalias, line 45: Why did we remove the transition here?

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -1,37 +1,34 @@
...
-
+  myUserUid?: string;
   setSelectedUserProfile: (user: WithId<User>) => void;
 }
 
-export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
+export const _MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
   partygoer,
-  venueId,
+  position,
```
0xdevalias, line 22: What was `venueId` used for in the past?

And how was the old code using `position` for `currentRow`/`currentCol` when seemingly (according to this diff at least) it didn't exist as a prop before?

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -1,37 +1,34 @@
...
-  venueId: string;
-  myUserUid: string;
+  position: Experience;
   totalRows: number;
   totalColumns: number;
 
   /** @default false **/
   withMiniAvatars?: boolean;
-
+  myUserUid?: string;
```
0xdevalias, line 16: Why is this optional now?

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -1,37 +1,34 @@
...
+import { Experience, User } from "types/User";
 
 import UserProfilePicture from "components/molecules/UserProfilePicture";
 
 interface MapPartygoerOverlayProps {
   partygoer: WithId<User>;
-
-  venueId: string;
-  myUserUid: string;
+  position: Experience;
```
0xdevalias, line 10: `Experience` contains far more data than what we actually want/need here. We only need a `row` and a `column`. Might even be worth just having 2 different props that are `number` for this, rather than passing an object with extra bits we don't need (which will also simplify memoisation checks)

src/components/molecules/MapRow/MapRow.tsx
```diff
@@ -52,3 +52,5 @@ export const MapRow: React.FC<MapRowProps> = ({
     </div>
   );
 };
+
+export const MapRow = React.memo(_MapRaw);
```
0xdevalias, line 56:
```suggestion
export const MapRow = React.memo(_MapRow);
```

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -13,14 +13,11 @@ interface UsePartygoersOverlay {
...
-  partygoers?: readonly WithId<User>[];
+  partygoers: readonly WithId<User>[];
   setSelectedUserProfile: (user: WithId<User>) => void;
 }
 
-export type UsePartygoersReturn =
-  | ("" | JSX.Element)[]
-  | JSX.Element
-  | undefined;
+export type UsePartygoersReturn = (null | JSX.Element)[] | null;
```
0xdevalias, line 20: Since React allows `null` to be returned in JSX, is there a JSX type that already includes it, such that we don't need to explicitly add it here? One that says "anything that is valid in a React JSX-type position"

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -35,37 +32,33 @@ export const usePartygoersOverlay: ReactHook<
...
-      venueId,
-      userUid,
-      rows,
-      columns,
-      withMiniAvatars,
-      setSelectedUserProfile,
-    ]
-  );
+  return useMemo(() => {
+    if (!showGrid) return null;
```
0xdevalias, line 36: [nit] new line after guard statements

```suggestion
    if (!showGrid) return null;
    
```

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -13,14 +13,11 @@ interface UsePartygoersOverlay {
   withMiniAvatars?: boolean;
   rows: number;
   columns: number;
-  partygoers?: readonly WithId<User>[];
+  partygoers: readonly WithId<User>[];
```
0xdevalias, line 16: Why do we make this non-optional here? Can partygoers ever be `undefined`?

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -35,37 +32,33 @@ export const usePartygoersOverlay: ReactHook<
...
-      columns,
-      withMiniAvatars,
-      setSelectedUserProfile,
-    ]
-  );
+  return useMemo(() => {
+    if (!showGrid) return null;
+    // @debt this can be undefined because our types are broken so check explicitly
+    return partygoers.map((partygoer) => {
+      if (!partygoer.id || !partygoer.data?.[venueId]) return null;
```
0xdevalias, line 39: We could handle this in a `partygoers.filter(THISLOGIC).map(THERESTOFTHIS)`

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -35,37 +32,33 @@ export const usePartygoersOverlay: ReactHook<
...
+
+      return (
+        <MapPartygoerOverlay
+          key={partygoer.id}
+          partygoer={partygoer}
+          position={partygoer?.data?.[venueId]}
+          myUserUid={userUid} // @debt fix this to be less hacky
```
0xdevalias, line 46: I think the `@debt` comment here related to `myUserId` having a default instead of passing through as `undefined`, so I think we can remove that comment with the change here.

See comment earlier RE: `position`, I think it's better to just pass `row` and `column` down as individual props. So we should probably extract them in a `const` higher up.

src/components/templates/Camp/hooks/usePartygoersOverlay.tsx
```diff
@@ -35,37 +32,33 @@ export const usePartygoersOverlay: ReactHook<
...
-                totalRows={rows}
-                totalColumns={columns}
-                withMiniAvatars={withMiniAvatars}
-                setSelectedUserProfile={setSelectedUserProfile}
-              />
-            )
-        )
-      ) : (
-        <div />
-      ),
```
0xdevalias, line 59: I believe this `<div />` was required here for some kind of styling related reasons. Have you changed those/checked that removing it doesn't break anything?

0xdevalias (commented):
A more thorough review on everything except for `src/components/templates/PartyMap/components/Map/Map.tsx` (which I will get back to reviewing when I can.. it was too long to get into for the time being)

0xdevalias:
@mike-lvov This needs to be updated to align with the changes now merged into `staging`, as currently there seem to be way more files changed than I believe this PR intends to, which makes it hard to review.

mike-lvov:
I have looked at this with a fresh look and came up with an idea that this will not have a meaningful impact on performance. One could find some architectural changes(regarding the `Partymap` template) that could be useful. But as we are striving to rewrite the whole thing, I'd prefer not to spend time trying to fix all of the issues and focus instead on creating a viable MVP for Partymap, based on canvas technology

WDYT @denisdimitrov @0xdevalias ?

0xdevalias:
> @mike-lvov I have looked at this with a fresh mind and came up with an idea that this will not have a meaningful impact on performance. One could find some architectural changes(regarding the `Partymap` template) that could be useful. But as we are striving to rewrite the whole thing, I'd prefer not to spend time trying to fix all of the issues and focus instead on creating a viable MVP for Partymap, based on canvas technology

@mike-lvov It's hard to skim through the changed files and see what was done in this PR since the latest changes in staging seem to be complicating the diff. If you think we're better to close this PR and focus on the canvas re-write instead then i'm happy to follow your lead with that. We can always come back to this PR later and cherry pick out any aspects that make sense to keep if we need to.

PR #1059, opened 2020-12-29
[FIX] Chat display fix

mike-lvov:
Chat context was removed by the previous commit. Restored useFirestoreConnect in this commit to fetch chats


0xdevalias:
@mike-lvov When you say 'was removed by the previous commit, which PR and/or commit are you referring to? Can you please link it here. (Also, for future, when possible it's a good idea to link to the PR/etc just to reduce the cognitive burden for future, and to make it easier to 'trace back' and see where things were introduced/etc.)

src/hooks/useChats.ts
```diff
@@ -0,0 +1,23 @@
+import { useFirestoreConnect } from "react-redux-firebase";
```
0xdevalias, line 1: We should try and keep the filenames and the functions they export aligned. So `useChats` doesn't make sense for this filename (as it makes me think that there should be a `useChats` function being exported, which there isn't)

I would either make the filename `useVenueChats` or similar, or make it more generic, and just call it `chat` or similar.

0xdevalias: Also, instead of importing `useFirestoreConnect` directly from the lib, I believe we have our own typed wrappers we should be using instead (assuming that PR has been merged?)

src/hooks/useChats.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { venueChatsSelector } from "utils/selectors";
+
+import { useVenueId } from "hooks/useVenueId";
+import { useSelector } from "hooks/useSelector";
+
+export const useVenueChatConnect = () => {
+  const venueId = useVenueId();
```
0xdevalias, line 9: IMO we shouldn't be calling this sort of thing 'inside' a hook like this, it feels like it couples it too tightly. My preference in the past has been to pass the `venueId` we want to 'connect' to as a prop. Which will make it easier to write nice isolated tests/etc as well.
```suggestion
import { useSelector } from "hooks/useSelector";

export const useVenueChatConnect = (venueId: string) => {
```

If you make this change though (maybe even if you don't) then we'll need to ensure that the `storeAs` `venueChats` always refers to the same venue. It might be better to make the `storeAs` include the `venueId`.. eg.

```
storeAs: `${venueId}Chats`,
```

(and if we do that, we also need to make sure to update the `useVenueChats` below to correctly reference this as well)

src/hooks/useChats.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  useFirestoreConnect({
+    collection: "venues",
+    doc: venueId,
+    subcollections: [{ collection: "chats" }],
+    storeAs: "venueChats",
+  });
+};
```
0xdevalias, line 17: Can `venueId` ever be `undefined`/etc at this point? If so, we risk the same 'pathological query' we've seen in other places that would load all venues. It might be safer to use the following sort of pattern we've used in other places:
```suggestion
  useFirestoreConnect(venueId ? {
    collection: "venues",
    doc: venueId,
    subcollections: [{ collection: "chats" }],
    storeAs: "venueChats",
  } : undefined);
};
```

src/hooks/useChats.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const useVenueChats = () => {
+  useVenueChatConnect();
```
0xdevalias, line 20: [musing] I don't think there is any issue with us calling this multiple times within the app, as I believe the lib will realise it's the same thing, and not 're-subscribe' to it. I was pondering at one stage that we could build our own logic into our 'connect' hooks (eg. `useVenueChatConnect`) that somehow detected if it had already been subscribed, and if so, didn't try again. But if the library does that for us, even better!

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -1,27 +1,31 @@
...
 import {
   partygoersSelector,
   partygoersSelectorData,
   privateChatsSelector,
-  venueChatsSelector,
 } from "utils/selectors";
 
+import { DEFAULT_PARTY_NAME, DEFAULT_PROFILE_IMAGE } from "settings";
+
+import { User } from "types/User";
```
0xdevalias, line 25: [nit] There's a specific 'import sort order' that I find useful, it generally keeps imports from 'main' 3rd party libraries at the top (eg react/redux/firestore/etc), then 'secondary' 3rd party libraries (eg. lodash/etc), then imports from our 'root' files like settings/etc near the top, then types, utils, hooks, templates, organisms, molecules, and then 'relative' imports from `../`, then `./`, and style imports are always basically the last things (or something like that.. from memory) If you look at any of the files where I have actually resorted the imports you'll probably see it.

Not a huge thing, and at some point I intend to encode this pattern via eslint or similar (as it can be a pain to think about manually sometimes), but figured I would raise it here just so you get an idea of how I like to do it.

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -1,27 +1,31 @@
 import React, { useState, useEffect, useMemo, useRef } from "react";
-import UserProfileModal from "components/organisms/UserProfileModal";
+import { useFirestoreConnect } from "react-redux-firebase";
```
0xdevalias, line 2: Should we be importing this from our 'wrapped' version rather than directly from the library?

src/components/organisms/Chatbox/Chatbox.tsx
```diff
@@ -69,7 +73,7 @@ const Chatbox: React.FunctionComponent<PropsType> = ({
         }
       : undefined
   );
-  const chats = useSelector(venueChatsSelector);
+  const chats = useVenueChats();
   const privateChats = useSelector(privateChatsSelector);
```
0xdevalias, line 77: Should we be creating a custom hook for private chats (eg. `usePrivateChats` + `usePrivateChatsConnect`) in a similar way as `useVenueChats`? Maybe not in this PR necessarily.

src/pages/ReactionPage/ReactionPage.tsx
```diff
@@ -17,9 +18,8 @@ const ReactionPage = () => {
   const partygoers = useSelector(partygoersSelector);
   const usersById = partygoers;
   const reactions = useSelector((state) => state.firestore.ordered.reactions);
-  const chats = useSelector((state) =>
-    state.firestore.ordered.venueChats?.filter((chat) => chat.deleted !== true)
-  );
+  const chats = useVenueChats();
+  const filteredChats = chats?.filter((chat) => chat.deleted !== true);
```
0xdevalias, line 22: When the filter is done within the selector, it is memo'd and the component will only be updated if the end result (eg. after filtering) changes. With the change made here, it is going to re-render the component whenever any chat changes, and then re-run the filter each time.

It would be better if we could do this as part of the selector itself. I would probably refactor `useVenueChats` to take some options as props. In this case, I would probably call it something like `includeDeleted` or `withDeleted`, and default it to `false`. Then, based on this prop, we can do the basic selector, or the one that filters deleted.

To get the best memoisation, we want the selector function reference to remain the same, that is generally why we extract them to the selectors file. In this case it may make sense to do it more 'inline' within the hook though, so we could use `useCallback` to ensure our selector function reference only changes when the props it relies on changes. Something like the following (within the `useVenueChats` hook):

```typescript
const venueChatsSelector = useCallback((state) => {
  const venueChats = state.firestore.ordered.venueChats ?? [];

  if (withDeleted) {
    return venueChats;
  } else {
    return venueChats?.filter((chat) => chat.deleted !== true)
  }
}, [withDeleted])
```

If you wanted to keep the selector logic in the `selectors.ts` file, then we could basically extract the logic of the function described above into that file, and then just call that function within the `useCallback` within `useVenueChats` (which is probably the better way, as then all our selector logic is kept 'isolated' in that one file)

`selectors.ts`:
```typescript
const makeVenueChatsSelector = ({ withDeleted = false}) => (state) => {
  const venueChats = state.firestore.ordered.venueChats ?? [];

  if (withDeleted) {
    return venueChats;
  } else {
    return venueChats?.filter((chat) => chat.deleted !== true)
  }
}
```

`useVenueChats.ts`:
```typescript
const venueChatsSelector = useCallback((state) => makeVenueChatsSelector({ withDeleted })(state), [withDeleted])
```

src/pages/ReactionPage/ReactionPage.tsx
```diff
@@ -17,9 +18,8 @@ const ReactionPage = () => {
...
 
   useFirestoreConnect([
```
0xdevalias, line 24: Technically this is checking for `venue` currently, but then uses `venue.name` in the `doc`. So we should make the check `venue?.name`, otherwise it has the (rare) chance to end up `undefined` and trigger a pathological query

src/pages/ReactionPage/ReactionPage.tsx
0xdevalias: And if the above isn't done (which is my preferred choice), at the very very least we should be wrapping `filteredChats` with `useMemo` so it's not re-filtering on every render.

0xdevalias (changes requested):
@mike-lvov A bunch of scattered review comments/suggestions/etc. I think some of them warrant changes before getting this out, but if you think they will take too long/disagree, and that we should merge this to fix prod first, then follow up with another PR to improve these aspects, I am open to that as well.

src/components/organisms/Chatbox/Chatbox.tsx
mike-lvov: Would be nice to have it automatically sorted. Agreed with your approach here

src/components/organisms/Chatbox/Chatbox.tsx
mike-lvov: I didn't import it myself. I have only changed the import order. It would be nice to have a commit to remove `useFirestoreConnect` across our codebase. Didn't do it in this PR as it is unrelated to it. Though, I think we should create a task for that.

src/components/organisms/Chatbox/Chatbox.tsx
mike-lvov: Agreed, I think we should move all selector\connect logic to the hooks. Not in this PR though

src/pages/ReactionPage/ReactionPage.tsx
mike-lvov: I think this approach is what I'm striving to do as well. Did useMemo for now. Though, I don't think there is any memorization happening as we don't have libs like `reselect` for that. As far as I understand, `useSelect` serves only as a pure function to retrieve data from redux store. I can be mistaken though.

0xdevalias: useSelector definitely memo’s in 2 different ways without reselect:

- https://react-redux.js.org/api/hooks

0xdevalias:
@mike-lvov i just saw that this was targeting master rather than staging.. and has been merged. I’m not sure if there is some context I’m missing, but we open PR’s against staging, and then that gets deployed to master, as opposed to merging directly against master.

mike-lvov:
@0xdevalias Yes, the normal flow is creating PRs against staging, merge that and then merge and deploy to master. However, the thing I merged was a hotfix. We didn't have core functionality(chats) working on the production, which I think is a case for hotfixes.

0xdevalias:
They had been broken for a while, and the time difference of a couple
Of minutes to follow our normal 'flow' wouldn't have made a meaningful difference, so we still should follow staging -> prod as normal. The only time a hotfix directly against master would maybe make sense (and even then I'm not so sure to be honest) is if the timing was so critical that the extra couple of minutes would make/break us, which ideally is never going to be the case. Even during our live patching of critical issues during recent events we followed the staging -> prod flow, and it was sufficient for our needs.

By merging directly against master, our staging and master branches are now out of sync, and we have to do extra work to port the changes from master back to staging.

In essence, it gains us very little, and causes extra work to diverge from our standard flow.

0xdevalias:
Chatting to @mike-lvov it seems that the core 'issue' that required the PR to be raised against master here is that there are changes currently in staging that we may not want to deploy to production yet (so staging is essentially 'blocked').

This is an issue we've run into in the past, and hopefully will be able to resolve sooner rather than later, potentially by implementing 'per feature branch' environments so that we can clickthrough/etc before deploying to staging, and thus proactively know if it would break anything.

src/components/organisms/Chatbox/Chatbox.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/109

src/hooks/useChats.ts
0xdevalias: Fixed in 354167a550c42e93a27cc5c1e431145ff1caeef7

src/hooks/useChats.ts
0xdevalias: Partially fixed in 354167a550c42e93a27cc5c1e431145ff1caeef7, but that implementation doesn't account for the `venueId` changing and how that will impact the  `storeAs` name as highlighted in my above comment.

0xdevalias: Backlogged the above in https://github.com/sparkletown/internal-sparkle-issues/issues/110

src/hooks/useChats.ts
0xdevalias: fixed in 354167a550c42e93a27cc5c1e431145ff1caeef7

src/components/organisms/Chatbox/Chatbox.tsx
0xdevalias: https://github.com/sparkletown/sparkle/pull/801 is the WIP PR that is implementing this + eslint banning the direct usage across the codebase

src/pages/ReactionPage/ReactionPage.tsx
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/111

src/pages/ReactionPage/ReactionPage.tsx
0xdevalias: resolved in 354167a550c42e93a27cc5c1e431145ff1caeef7

PR #1062, opened 2021-01-05
[documentation] add code style priciples

mike-lvov:

PR #1065, opened 2021-01-07
Improve helper scripts

0xdevalias:
- fix `initFirebaseAdminApp` to allow it's 2nd argument to be optional as originally intended (a24fad9)
- fix shebang/imports in `scripts/count-users-entered-venue` to align with our patterns (0b3be45)
- refactor `scripts/get-badges` for performance/efficiency (b5ad60f)
- add `secondsToDuration` to time utils (1e79df1)
- add `formatSecondsAsDuration` to time utils (08ea444)
- improve `scripts/get-badges` to output user's `timeSpent` in each `venue`
- allow `scripts/upload-table-config` to use different credential files (36e5673)
- fix `scripts/clone-data-across-firebase-projects` types (d3d3751)
- many many more things, see the commits, as they are fairly individualised/descriptive
  - https://github.com/sparkletown/sparkle/pull/1065/commits

## TODO

- [x] Update https://github.com/sparkletown/sparkle/pull/1067, then merge it into this branch


scripts/get-badges.ts
```diff
@@ -9,61 +14,80 @@ const usage = () => {
...
-  const { users, pageToken } = await app.auth().listUsers(1000);
+  // TODO: extract this as a generic helper function?
+  const usersWhoEnteredAnyOfVenues: WithId<User>[] = await admin
+    .firestore()
+    .collection("users")
+    .where("enteredVenueIds", "array-contains-any", venueIdsArray)
+    .get()
+    .then((snapshot) =>
+      snapshot.docs.map((doc) => withId(doc.data() as User, doc.id))
+    );
```
0xdevalias, line 45: TODO?

scripts/get-badges.ts
```diff
@@ -9,61 +14,80 @@ const usage = () => {
...
+  // TODO: extract this as a generic helper function?
+  // Note: 100 is the max that can be requested with getUsers() per chunk
+  const authUsers: admin.auth.UserRecord[] = await Promise.all(
+    chunk(usersWhoEnteredAnyOfVenues, 100).map(async (usersChunk) => {
+      const usersChunkIds = usersChunk.map((user) => ({ uid: user.id }));
+      const authUsersResult = await admin.auth().getUsers(usersChunkIds);
+
+      return authUsersResult.users;
+    })
+  ).then((result) => result.flat());
```
0xdevalias, line 56: TODO?

scripts/get-badges.ts
```diff
@@ -9,61 +14,80 @@ const usage = () => {
...
 
-  allUsers.push(...users);
-  nextPageToken = pageToken;
+  const authUsersById = authUsers.reduce(
+    (acc, authUser) => ({ ...acc, [authUser.uid]: authUser }),
+    {} as Record<string, admin.auth.UserRecord>
+  );
+
+  // TODO: filter enteredVenueIds so that they only contain related venues?
```
0xdevalias, line 63: TODO?

src/utils/time.ts
```diff
@@ -132,6 +171,7 @@ export function getHoursAgoInSeconds(hours: number) {
   return nowInSec - hours * 60 * 60;
 }
 
+// @debt this is a duplicate of getCurrentTimeInUTCSeconds
```
0xdevalias, line 174: [OOS] `@debt` this is a duplicate of `getCurrentTimeInUTCSeconds`

scripts/get-badges.ts
```diff
@@ -9,61 +19,116 @@ const usage = () => {
...
 (async () => {
-  const allUsers: admin.auth.UserRecord[] = [];
-  let nextPageToken: string | undefined;
-  const { users, pageToken } = await app.auth().listUsers(1000);
+  // TODO: extract this as a generic helper function?
+  const usersWithVisits = await Promise.all(
+    await admin
+      .firestore()
+      .collection("users")
+      .where("enteredVenueIds", "array-contains-any", venueIdsArray)
```
mike-lvov, line 58: Is it possible to have `venueIdsArray.length > 10`?

mike-lvov (commented):
Looks good. Maybe requires a little change to ensure it works with large venueIdsArray arrays

scripts/get-badges.ts
0xdevalias: As the code is written, technically yes.

I'm not sure if we would ever need to do so in practice though, possibly for a large event if we wanted to check all 'related venues' specifically (and not just use the 'parent venue' as a 'gateway'), before https://github.com/sparkletown/internal-sparkle-issues/issues/127 is implemented.

If it ends up being not too hard of a code change, might be worth implementing the 'chunk' pattern here. If it turns out to be a lot of extra work, probably just check that `venueIdsArray.length <= 10` at the start of the script, and throw an error if not.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1065/commits/6dd0b317b5abd0a38dd40905ecf2458904b25649

0xdevalias:
@mike-lvov (or @sparkletown/developers) can you please re-review this. I've fixed the last issue you raised + cleaned up a whole bunch more of the patterns across our scripts.

src/utils/time.ts
denisdimitrov: Can you just remove it and use the other one instead?

denisdimitrov: There is quite a lot of `@debt` notations in the code, we should start fixing the problems and removing them.

scripts/lib/helpers.ts
```diff
@@ -1,63 +1,228 @@
...
   {
     appName,
-    credentialPath = resolve(__dirname, "../prodAccountKey.json"),
-  }: { appName?: string; credentialPath?: string }
-): admin.app.App =>
-  admin.initializeApp(
+    credentialPath: _credentialPath,
+  }: { appName?: string; credentialPath?: string } = {}
+): admin.app.App => {
+  const credentialPath = _credentialPath ?? resolveDefaultCredentialPath();
```
denisdimitrov, line 58: The documentation accepts `serviceAccount` as first param. Maybe you can rename this variable to that and avoid the underscore syntax/name shadowing.

https://firebase.google.com/docs/reference/admin/node/admin.credential#cert

scripts/count-users-entered-venue.ts
```diff
@@ -1,33 +1,41 @@
...
   usage();
 }
 
 const venueIdsArray = venueIds.split(",");
 
-initFirebaseAdminApp(projectId);
+// Note: if we ever need to handle this, we can split our firestore query into 'chunks', each with 10 items per array-contains-any
+if (venueIdsArray.length > 10) {
+  console.error(
+    "Error: This script can only handle up to 10 venueIds at once at the moment."
```
denisdimitrov, line 28: Why? Is this a firebase restriction or just our own limitation to avoid heavy performance issues.

scripts/clone-data-across-firebase-projects.ts
```diff
@@ -64,6 +54,33 @@ const destApp = initFirebaseAdminApp(DEST_PROJECT_ID, {
 // TODO: venues (owners will need to be changed)
 // TODO: check if venue already exists (be safe, don't overwrite!)
 
+const replaceSourceDomainReferences = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  obj: Record<string, any>,
```
denisdimitrov, line 59: We are currently passing only `room` and `venue` objects in this function. Can you import the typings here? If you can it would be nice to add it and remove the `any`, if you can't then keep it as it is.

src/utils/time.ts
0xdevalias: I’m not using it for anything, it’s just something unrelated to the rest of my changes that noticed in my work. It should be fixed as a new PR focussed on cleaning it up.

Which also goes more generally for things labelled debt, they should be raised/captured in backlog issues/similar so we can actually get them fixed.

scripts/count-users-entered-venue.ts
0xdevalias: Firebase restriction in that we can only pass 10 items to the ‘array contains’ things.

We can work around it with the ‘chunk’ pattern (as explained in the attached comment above the check)

I’ve never had to put that many venueId’s into this sort of script, and so didn’t want to spend the extra time/script complexity of implementing the chunk pattern for a need that may not ever arise.

scripts/clone-data-across-firebase-projects.ts
0xdevalias: We can definitely import the types.

I’d have to check the code here to see if it would make sense to use them in this function though (while I wrote the initial script, I didn’t write this part of it, so not as familiar with it off-hand)

scripts/get-badges.ts
mike-lvov: Please, don't resolve comments, since it is hard to keep up with all the changes with them resolved

mike-lvov (commented):
A few non-blocking comments

scripts/get-badges.ts
```diff
@@ -1,69 +1,135 @@
...
+            );
 
-  while (nextPageToken) {
-    const { users, pageToken } = await app
-      .auth()
-      .listUsers(1000, nextPageToken);
-    allUsers.push(...users);
-    nextPageToken = pageToken;
-  }
+          const userWithVisits: UsersWithVisitsResult = { user, visits };
```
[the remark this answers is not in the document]
0xdevalias: I believe I chose to do it to explicitly add the type to it here, rather than needing to do so at the ‘entry point’ of this bit of code, or ‘force’ the cast with ‘as’ (which I believe would overwrite it even if it was inaccurate).

From memory, if I don’t add the type, then later on in type hints/etc we see the full ‘object of types’ (as it can infer the correct types), rather than the much simpler ‘grouping type name’, which takes longer/more cognitive effort to understand/what it means/is referring to/etc.

It might be fine to add it at the ‘entry point’ rather than here, I would have to look deeper at my code again.

scripts/get-badges.ts
0xdevalias: Does the comment saying that it was fixed and where not solve that need?

I know this is the same thing I’ve requested of you in the past, so it’s a reasonable request. I probably resolved it by habit because I knew it was done + referenced where/why, so by doing so it removed it from the things I need to mentally process in ensuring the PR is ready.

I think where I find it ‘harder’ (and likely the original source of my request) is if someone just ‘replies’ with a ‘reaction’ (eg. thumbs up) and no comment, then resolves it; as that doesn’t trigger any of the notifications that I often will use to ensure I have checked it all.

0xdevalias: (referenced this thread in https://github.com/sparkletown/sparkle/pull/1062#issuecomment-764969428 so that we can document it once we figure out our patterns here)

scripts/get-badges.ts
```diff
@@ -1,69 +1,135 @@
...
+        uid: userWithVisits.user.id,
+      }));
+
+      const authUsersResult = await admin.auth().getUsers(chunkUserIds);
+
+      return authUsersResult.users;
+    })
+  ).then((result) => result.flat());
+
+  const authUsersById = authUsers.reduce(
```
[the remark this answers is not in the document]
0xdevalias: What does making this typing explicit add that we aren't already getting from type inference?

On line 83 we explicitly type `authUsers`: `const authUsers: admin.auth.UserRecord[]`

Which is then used in this reducer function.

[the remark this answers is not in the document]
0xdevalias: Ah.. is the request here that we replace the `as Record<string, admin.auth.UserRecord>` on line 97 with different typing? If so, makes sense, and I agree.

Though we would probably just do this by making line 95: `const authUsersById: Record<string, admin.auth.UserRecord> = authUsers.reduce(` + deleting the `as` line

scripts/lib/helpers.ts
0xdevalias: The logic here currently resolves the credentials from this script file's location, then passess it into `.cert` below.

`.cert`'s first param is called `serviceAccountPathOrObject`. Since we haven't actually loaded the credential file, in our usecase, it would basically need to be called `serviceAccountPath` to be accurate. Which is in essence what we are using with `credentialPath` (yet sticking to what we traditionally called this across all of our other scripts)

The `_` syntax + 'shadowing' here is basically a workaround to setting the default value on the parameter when not supplied, to simplify the typing, but for all intents and purposes, it should be considered the same as the param (which is why it was 'shadowed' in this way)

I believe the reasoning behind this (as I remember it) is that if we didn't do it this way, then we would either have to make `credentialPath` non-optional in the typing (which we don't really want to do), or we would need to specify the default value in 2 places (to satisfy both the 'normal' default, and then the default again in the empty object(`{}`) part of the props.

--

I don't think the `_` syntax + 'shadowing' is a big deal here, particularly how 'localised' it is within a helper function that isn't going to need to be changed much/ever. So i'll probably leave it as is.

src/utils/time.ts
0xdevalias: backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/243

scripts/clone-data-across-firebase-projects.ts
0xdevalias: Looked into this for a little bit, while we can explicitly use `Venue | Room` as the type here, it then complicates all of the typing/etc of the rest of the code in this function, as it needs to figure out if the key is a string, then update it, etc.

We can fix this up in a future PR if we decide we care strongly about it then, but it didn't seem like it was worth the time/effort to fix here, for a fairly tightly scoped helper function in an engineering QoL script.

scripts/get-badges.ts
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1065/commits/ad572c1154fe0aff80110c317eac488b662e30b6

scripts/get-badges.ts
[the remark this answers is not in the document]
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1065/commits/0837f02086f5c879eae48fe479231622d1e14dd5

0xdevalias:
@sparkletown/developers still need an approval on this PR

denisdimitrov (approved):
LGTM!

scripts/lib/helpers.ts
denisdimitrov: Sounds good.  It wasn't a change request or something that I strongly want to be changed. Keep it as it is 🎉

scripts/clone-data-across-firebase-projects.ts
denisdimitrov: > We can fix this up in a future PR....
I agree, we can create a PR in the future that cleans as many `any` usages as possible.

PR #1071, opened 2021-01-11
add map background when creating a venue

denisdimitrov:
Not sure when or how this was removed. Anyways, adding it in this PR to the `createVenue` function. `updateVenue` already has this code.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/131


0xdevalias:
@denisdimitrov is this in relation to https://github.com/sparkletown/internal-sparkle-issues/issues/131?

If so, can you add `fixes https://github.com/sparkletown/internal-sparkle-issues/issues/131` to your original PR description. That will link the issues, and will automatically close the backlog item when this PR is merged:

- https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

denisdimitrov:
@0xdevalias Yes this is the issue, however it's in another repo and can not be linked or at least I don't know how.

0xdevalias:
> Yes this is the issue, however it's in another repo and can not be linked or at least I don't know how.

@denisdimitrov my comment in https://github.com/sparkletown/sparkle/pull/1071#issuecomment-758157807 and the related GitHub documentation link tells you how to do it:

> If so, can you add `fixes https://github.com/sparkletown/internal-sparkle-issues/issues/131` to your original PR description. That will link the issues, and will automatically close the backlog item when this PR is merged

0xdevalias:
> Not sure when or how this was removed.

Definitely existed at:
https://github.com/sparkletown/sparkle/blob/d40453919bafc1c1767a2f3eaa588f05205803e8/functions/venue.js#L60

These are the times that line was modified in the commit history (using a super useful `git blame`-like command that I never knew existed):

- https://stackoverflow.com/questions/4404444/how-do-i-git-blame-a-deleted-line/4404551#4404551

```
⇒  git log -S "venueData.mapBackgroundImageUrl = data.mapBackgroundImageUrl;" -- functions/venue.js | cat
commit 78287500322ab62b045f4e20e020236ea35cc14f
Author: Glenn 'devalias' Grant
Date:   Tue Dec 15 22:45:20 2020 +1100

    Entrance Experience Customisations (#992)
    * Themecamp is partymap; Removed placement input; Removed mapicon input
    * :construction: Question inputs component created; Removed unused inputs from DetailsForm
    * :construction: Entrance input component created with Entrance Input Buttons
    * Removed commented out code from venue functions

    Co-authored-by: IvanBuljovcic

commit d40453919bafc1c1767a2f3eaa588f05205803e8

Author: maxbull
Date:   Thu Aug 13 17:02:19 2020 +0100

    443: Write mapbackground only in map venue types
```

Looks like that line was lost in https://github.com/sparkletown/sparkle/pull/992 at:

- https://github.com/sparkletown/sparkle/pull/992/files#diff-d8701293598c7e3a582ee79f4554f303e5b11f670cd47d41c15e642f6ddba3d8L116

I'm not 100% sure of the intent of that PR, but it looks like `mapBackgroundImageUrl` is removed from quite a lot of the codebase within it. For example, it was removed from `src/pages/Admin/Venue/DetailsValidationSchema.ts`, and there doesn't seem to be any equivalent sounding usage of it in the current version still on `staging` either.

There appear to be 12 usages across the codebase currently:

- https://github.com/sparkletown/sparkle/search?q=mapBackgroundImageUrl

@denisdimitrov it may be worth looking over the diff in https://github.com/sparkletown/sparkle/pull/992 more closely, and seeing if we need to restore anything else that changed there as well, particularly in relation to `mapBackgroundImageUrl`/etc.

denisdimitrov:
@0xdevalias Not sure which IDE you are using, but there is a GIT Blame extension for VSCode which is super helpful and shows you at the bottom left corner when and by whom was last modified the current line you are on. Not very useful for removed ones but it's still something. Currently looking at the diff and adding the missing stuff here.

0xdevalias:
> Not sure which IDE you are using, but there is a GIT Blame extension for VSCode which is super helpful and shows you at the bottom left corner when and by whom was last modified the current line you are on. Not very useful for removed ones but it's still something. Currently looking at the diff and adding the missing stuff here.

@mike-lvov I use Jetbrains Webstorm, which probably has a similar feature. Usually when I want to git blame i'll just do it through the GitHub UI, as that allows me to 'step back in time' and see earlier revisions/'blames' easier.

In this instance I was doing a fairly non-standard thing, trying to find the point at which the code changed/was lost, which I don't believe most of the 'normal' git blame UI wrappers tend to handle/account for.

mike-lvov (commented):
Got a small suggestion.

functions/venue.js
```diff
@@ -102,6 +102,10 @@ const createVenueData = (data, context) => {
     }
   }
 
+  if (data.mapBackgroundImageUrl) {
+    venueData.mapBackgroundImageUrl = data.mapBackgroundImageUrl;
```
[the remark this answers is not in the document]
mike-lvov: I think the way it was implemented differs from the one suggested here.

image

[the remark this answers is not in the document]
denisdimitrov: @mike-lvov Yes it differs, since camp is going to be removed in the future. I decided to make it a conditional check, that way in the future when a new venue is added and mapBackgroundImagUrl is provided you won't have to go in the cases and put it in there again.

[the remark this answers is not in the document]
denisdimitrov: > Can we move it here?
> 


Yes, will do.

denisdimitrov:
@mike-lvov Your suggestion has been applied.  Feel free to re-review the PR.

functions/venue.js
```diff
@@ -88,6 +88,9 @@ const createVenueData = (data, context) => {
     requiresDateOfBirth: data.requiresDateOfBirth || false,
     showRadio: data.showRadio || false,
     radioStations: data.radioStations ? [data.radioStations] : [],
+    ...(data.mapBackgroundImageUrl && {
```
mike-lvov, line 91: What are the downsides of doing it like this?

```mapBackgroundImageUrl: data.mapBackgroundImageUrl,```

Comment (denisdimitrov) [reply functions/venue.js:91]: You can not have undefined in the firebase. And since some templates don't have mapBackgroundImageUrl, I added this conditional to make sure this is uploaded to the firebase only when it exists. We can use a default value but then we an extra field in the database which is is not being used for some templates.

Comment (denisdimitrov) [reply functions/venue.js:91]: To explain in more in details:

If we use `mapBackgroundImageUrl: data.mapBackgroundImageUrl` and then create a jazzbar, the function will fail because `mapBackgroundImageUrl` will be `undefined`

If we use `mapBackgroundImageUrl: data.mapBackgroundImageUrl ?? "` and then create a jazzbar, the function will succeed but there will be a field `mapBackgroundImageUrl: ""` in the database that will never be used.

So that leaves us with 2 options: 
1. Add `mapBackgroundImageUrl: data.mapBackgroundImageUrl` to the switch cases for all of the templates that have `mapBackgroundImageUrl` and then keep adding it in the future ones.
2. Add conditional `mapBackgroundImageUrl` that will add the field only when it exists.

**Review by mike-lvov — APPROVED** [review]
LGTM

File: functions/venue.js (hunk 2 above)
Comment (0xdevalias) [reply functions/venue.js:91]: > You can not have `undefined` in the firebase.

> If we use `mapBackgroundImageUrl: data.mapBackgroundImageUrl` and then create a `jazzbar`, the function will fail because `mapBackgroundImageUrl` will be `undefined`

@denisdimitrov With regards to these comments, what specifically happens if we try and set a key that is `undefined`? Does the Firebase SDK throw an error, or it just doesn't set that key, or?

If there is never a valid case where this data structure can contain `undefined` keys, then we could perhaps add our own more generic 'filter' pattern, where we treat `undefined` keys as 'do not set', and then filter them out from this structure before passing them to the Firebase SDK. There are many ways we could do that if we wanted to, but here's a one (and others in that thread for doing it with native JavaScript):

- https://stackoverflow.com/questions/25421233/javascript-removing-undefined-fields-from-an-object/37836669#37836669

---

@denisdimitrov I know you already said that your intent was to not have to make too many future changes as we add new templates, but I'm curious, which templates currently use `mapBackgroundImageUrl`? (guessing `partymap` and `themecamp` (deprecated))

If it's only for these `map` templates, I don't know that there is going to be a lot of 'future churn'. I believe we decided a while back that going forward we're going to 'focus in' and only have the main `partymap`, and then the smaller selection of 'other venues'; so I don't expect we'll be adding many/any new `map` venues.

I kind of feel like the way it's been implemented in this diff is probably 'worse' than how it was originally implemented before the suggestions in https://github.com/sparkletown/sparkle/pull/1071#pullrequestreview-568118430 were implemented.

Comment (denisdimitrov) [reply functions/venue.js:91]: >@denisdimitrov With regards to these comments, what specifically happens if we try and set a key that is undefined? Does the Firebase SDK throw an error, or it just doesn't set that key, or?


@0xdevalias Firebase throws an error and the request fails. That's why some of the values are being added with if statements conditionally.

>If there is never a valid case where this data structure can contain undefined keys, then we could perhaps add our own more generic 'filter' pattern, where we treat undefined keys as 'do not set', and then filter them out from this structure before passing them to the Firebase SDK. There are many ways we could do that if we wanted to, but here's a one (and others in that thread for doing it with native JavaScript):

@0xdevalias Imo this is an overkill, instead of adding undefined values and then filtering them out, why not just add the actual values conditionally? Just the way they are now, this way you skip the object iteration.

> I kind of feel like the way it's been implemented in this diff is probably 'worse' than how it was originally implemented before the suggestions in #1071 (review) were implemented.

@0xdevalias This diff and the previous one are identical, its just the syntax that is different. I'm conditionally adding the field to the object values instead of using if statement.


**These do the same thing:**
```jsx
const venueData = {}
if (data.mapBackgroundImageUrl) {
    venueData.mapBackgroundImageUrl = data.mapBackgroundImageUrl;
}
```

```jsx
const venueData = {
  ...(data.mapBackgroundImageUrl && {
    mapBackgroundImageUrl: data.mapBackgroundImageUrl,
  })
}
```

PR #1047, opened 2020-12-18
Fixes users' fetching. Renames users.

denisdimitrov:
use the same hook for counting in party map room modal and party map room. Also use it for the navigation at the top.

closes https://github.com/sparkletown/internal-sparkle-issues/issues/122
closes https://github.com/sparkletown/internal-sparkle-issues/issues/119
closes https://github.com/sparkletown/internal-sparkle-issues/issues/196
closes https://github.com/sparkletown/internal-sparkle-issues/issues/158

PR #1074, opened 2021-01-12
WIP: Reworks chats according to new designs and structure

mike-lvov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/139
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/110
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/109

The expected output would be to have only 1 version of chats' sidebar, which you will be able to see in every type of venue.

The desired internal structure outcome:
- venue(location), private, world(in plans) chats reuse the same Chatbox component(no logic inside)
- useVenueChat, usePrivateChat, useWorldChat hooks that return everything chat-specific related
- refactor chat's action and types(create api/chat file to handle firestore requests)
- remove all unused chat code
- chat UI is controlled from outside(can open private chat from the profile)

components to be deleted:
- components/molecules
---- Chatbox
---- ChatList
---- ChatsList
---- ChatMessage
---- Message
---- MessageList
---- Sidebar
---- VenueChat

- components/organisms
---- Chatbox
---- Chatdrawer


src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -4,23 +4,21 @@ $white: #ffffff;
 $room-info-bg: #005ee5;
 $border-radius: 28px;
 
-.party-map-content-container {
+.party-map-page-container {
```
0xdevalias, line 7: What was the purpose/reasoning behind this change?

It seems to be used within the `Map` component, which isn't a 'page' (which has a specific defined meaning in the organisms/molecules/etc file structure)

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -234,36 +234,34 @@ export const Map: React.FC<PropsType> = ({
...
-      <div className="party-map-container">
-        <div className="party-map-content">
-          <img
-            width="100%"
-            className="party-map-background"
-            src={venue.mapBackgroundImageUrl}
-            alt=""
-          />
+    <div className="party-map-page-container">
+      <div className="party-map-content-container">
```
0xdevalias, line 238: Are there specific reasons why we need to keep this nesting? And if so, wht are they?

src/components/molecules/Sidebar/Sidebar.scss
```diff
@@ -9,7 +9,7 @@ $dark: #19181a;
   z-index: z(sidebar);
   right: 0;
   top: 60px;
-  width: 360px;
+  min-width: 300px;
```
0xdevalias, line 12: Why the change from `360px` to `300px`? What happens when the window is super-wide? Should we have a max-width as well?

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: I have removed one level of nesting so far. `party-map-page-container` has sidebar(chat) inside

src/components/templates/PartyMap/components/Map/Map.scss
mike-lvov: https://github.com/sparkletown/sparkle/pull/1074/files#r556286544 since it not only has party-map content, but chat as well, I have decided that this naming would be more meaningful

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: Ah, makes sense. Eventually we want to refactor the sidebar out of here, but I guess we'll need to keep that for now.

src/components/templates/PartyMap/components/Map/Map.scss
0xdevalias: I think the pattern we have been using elsewhere is to relate it to the component, so it would be `.party-map-component`.

Then making use of css selectors/nesting, the content can just be `.party-map-component .content` (though using scss' nesting rather than writing out the full selector manually), and the sidebar can be a similarly nested `.party-map-component .sidebar` (but again, using scss nesting)

mike-lvov:
Seems like I'm partianlly blocked by https://github.com/sparkletown/internal-sparkle-issues/issues/109(will not be able to implement world chat for now). But generally speaking, it is a non-blocker for the PR since I will just disable world chat until we either implement world users via context or backendish.

Looks like we could you some context functionality to have `open chat control` exposed to the outer components
@0xdevalias whyt?

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1074#issuecomment-761180342

> Seems like I'm partianlly blocked by [sparkletown/internal-sparkle-issues#109](https://github.com/sparkletown/internal-sparkle-issues/issues/109) not be able to implement world chat for now). But generally speaking, it is a non-blocker for the PR since I will just disable world chat until we either implement world users via context or backendish.

Sounds like a good approach 👍🏻 

> Looks like we could you some context functionality to have `open chat control` exposed to the outer components
> @0xdevalias wdyt?

@mike-lvov How 'outer' are the components we need to expose this to? Eg. is it needing to expose from the child to it's direct parent? Or in a more 'disconnected' way (eg. from potentially anywhere in the app)?

If child -> direct parent there is a pattern I believe I have already used in the app using ``:

- https://reactjs.org/docs/hooks-reference.html#useimperativehandle

If we want access to those controls from potentially anywhere in the app, that's starting to feel a bit more like a 'global state' thing, and so I would probably be looking to implement it with Redux, via actions and such (probably abstracted via a custom hook so that the 'consumers' don't have to be exposed to the 'implementation details' of it)

package.json
```diff
@@ -25,6 +25,7 @@
     "axios": "^0.21.1",
     "bootstrap": "^4.4.1",
     "bootswatch": "^4.5.0",
+    "classnames": "^2.2.6",
```
0xdevalias, line 28: `classnames` is added here, but seemingly no corresponding `package-lock.json` update in this PR? Is that correct?

src/components/molecules/ChatsList/ChatsList.tsx
```diff
@@ -45,28 +45,11 @@ const ChatsList: React.FunctionComponent = () => {
   const { user } = useUser();
   const dispatch = useDispatch();
 
-  const chatUsers = useSelector(chatUsersSelector) ?? {};
+  const chatUsers = useSelector(chatUsersByIdSelector) ?? {};
```
0xdevalias, line 48: Do/can we have a hook that wraps this up along with it's associated 'connect' call (similar to what we are doing in the users PR)

I'll only comment on this once in this review, but my request would be that all occurrences of it are updated in that same way, to use the hook.

src/components/organisms/Chat/Chat.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+type ChatProps = {
+  sendMessage: (message: string) => void;
+};
+
+export const Chat: React.FC<ChatProps> = () => {
+  return <div />;
+};
```
0xdevalias, line 9: Guessing this is here as a placeholder for now?

src/components/organisms/ChatSidebar/components/WorldChat/WorldChat.tsx
```diff
@@ -0,0 +1,5 @@
+import React from "react";
```
0xdevalias, line 1: I see that you have `VenueChat` (which has all of the current `ChatList`/`ChatMessage`/etc underneath it, and you also have `WorldChat` here. Are you intending to duplicate those `ChatList`/`ChatMessage`/etc components underneath here as well? If so, that's definitely not something we want to do. Venue/World/etc chats should all basically be a tiny bit of code that connects them to their data source, but otherwise uses exact same generic underlying 'dumb components' ('dumb' being that they don't connect to the stores/etc by themselves, and are entirely driven by the props passed into them)

src/components/organisms/ChatSidebar/index.ts
```diff
@@ -0,0 +1 @@
+export { ChatSidebar as default } from "./ChatSidebar";
```
0xdevalias, line 1: Prefer named exports rather than default exports.

src/store/actions/Chat.ts
```diff
@@ -1,10 +1,15 @@
 import { createAsyncThunk } from "@reduxjs/toolkit";
```
0xdevalias, line 1: Consider following the Redux style guide/best practices referenced in https://github.com/sparkletown/internal-sparkle-issues/issues/244

Of particular note, which reduces a lot of Redux boilerplate, and is considered best practice:

- https://redux-toolkit.js.org/

src/store/actions/Chat.ts
```diff
@@ -154,8 +159,34 @@ export const sendTableChat = createAsyncThunk<void, SendTableChatFields>(
   }
 );
 
+type ChatSidebarVisibilityAction = {
+  type: ChatActionTypes.SET_CHAT_SIDEBAR_VISIBILITY;
+  payload: boolean;
+};
```
0xdevalias, line 165: - https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters
  - > we recommend trying to treat actions more as "describing events that occurred", rather than "setters". Treating actions as "events" generally leads to more meaningful action names, fewer total actions being dispatched, and a more meaningful action log history. Writing "setters" often results in too many individual action types, too many dispatches, and an action log that is less meaningful.
  
This can be called out in many places, but I will just comment in this one place at the moment. Consider applying it across the entire PR though, as it will reduce the future re-work we have to do to clean things up later.

src/store/reducers/Chat.ts
```diff
@@ -0,0 +1,34 @@
+import { ChatActions, ChatActionTypes } from "store/actions/Chat";
+import { CHAT_TYPES } from "types/Chat";
+
+type PrivateChatState = {
+  openedChatType: CHAT_TYPES.PRIVATE_CHAT;
+  recipientId: string | null;
```
0xdevalias, line 6: What does `null` represent here? As a general rule, we basically never want to be using `null`, or at the very least, there needs to be a very compelling case as to why we would use it. What does it offer us here that `undefined` doesn't?

src/store/reducers/Chat.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const chatReducer = (
+  state = initialChatState,
+  action: ChatActions
+): ChatState => {
+  switch (action.type) {
+    case ChatActionTypes.SET_CHAT_SIDEBAR_VISIBILITY:
+      return { ...state, isChatSidebarVisible: action.payload };
+    case ChatActionTypes.SET_CHAT_TAB:
+      return { ...state, ...action.payload };
```
0xdevalias, line 30: https://redux.js.org/style-guide/style-guide#reducers-should-own-the-state-shape

src/store/reducers/Chat.ts
```diff
@@ -0,0 +1,34 @@
...
+  isChatSidebarVisible: true,
+};
+
+export const chatReducer = (
+  state = initialChatState,
+  action: ChatActions
+): ChatState => {
+  switch (action.type) {
+    case ChatActionTypes.SET_CHAT_SIDEBAR_VISIBILITY:
+      return { ...state, isChatSidebarVisible: action.payload };
```
0xdevalias, line 28: Prefer `action.payload` to be an object with named keys, that we access explicitly

src/types/Chat.ts
```diff
@@ -0,0 +1,16 @@
+export enum CHAT_TYPES {
+  WORLD_CHAT = "WORLD_CHAT",
+  VENUE_CHAT = "VENUE_CHAT",
+  PRIVATE_CHAT = "PRIVATE_CHAT",
+}
+
+type SetPrivateChatTab = {
+  chatType: CHAT_TYPES.PRIVATE_CHAT;
+  recipientId: string | null;
```
0xdevalias, line 9: Why `null` here, and what does it give us over `undefined`? (see other comment in this review for my reasoning)

src/types/Chat.ts
```diff
@@ -0,0 +1,16 @@
+export enum CHAT_TYPES {
```
0xdevalias, line 1:
```suggestion
export enum ChatTypes {
```

I would potentially also suggest that each of the entries also be named similarly, eg. `WorldChat` rather than `WORLD_CHAT`

src/components/organisms/ChatSidebar/components/VenueChat/ChatList/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./ChatList";
```
0xdevalias, line 1: Prefer named exports over instead of `default`

src/components/organisms/ChatSidebar/components/VenueChat/VenueChat.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useEffect, useState, useMemo, FC, useCallback } from "react";
```
0xdevalias, line 1: I expect this component may be able to be simplified using `src/components/molecules/ChatsList/ChatsList.tsx`

src/types/Firestore.ts
```diff
@@ -102,6 +102,7 @@ export interface FirestoreOrdered {
   subvenueEvents?: WithId<VenueEvent>[];
   userModalVisits?: Array<WithId<UserVisit>>;
   userPurchaseHistory?: Array<WithId<Purchase>>;
+  userPrivateChats?: Array<WithId<PrivateChatMessage>>;
```
0xdevalias, line 105:
```suggestion
  userPrivateChats?: WithId<PrivateChatMessage>[];
```

src/utils/selectors.ts
```diff
@@ -210,6 +215,9 @@ export const maybeArraySelector = <T extends SparkleSelector<U[]>, U>(
   selector: SparkleSelector<U[]>
 ) => (ifTrue ? selector : emptyArraySelector);
 
+export const chatUIStateSelector: SparkleSelector<ChatState> = (state) =>
+  state.chat;
```
0xdevalias, line 219: Since everything here is a 'state selector' by definition, would something like this work better?

```suggestion
export const chatUISelector: SparkleSelector<ChatState> = (state) =>
  state.chat;
```

src/hooks/chats.ts
```diff
@@ -0,0 +1,193 @@
...
+  const openChat = (
+    chatOptions: SetAnyChatTabOptions = { chatType: CHAT_TYPES.VENUE_CHAT }
+  ) => {
+    dispatch(setChatSidebarVisibility(true));
+    dispatch(setChatTab(chatOptions));
+  };
+
+  const closeChat = () => {
+    dispatch(setChatSidebarVisibility(false));
+  };
```
0xdevalias, line 37: You'll probably want to `useCallback` on these

src/hooks/chats.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  const { isChatSidebarVisible, openedChatType } = useSelector(
+    chatUIStateSelector
+  );
```
0xdevalias, line 41: Once more things are added to this redux store, we are going to want to use a far more 'refined' selector. As this is written currently. anytime anything in the chat state changes, this hook has to re-run, only to then ignore everything except for `isChatSidebarVisible`, `openedChatType`.

We want our selector function to return the absolute minimal amount of data possible to satisfy our needs. Selectors can also be composed, so you don't need to write each one from the 'root store'.

0xdevalias (commented):
Some interim comments so that you can align early, rather than it waiting till final review for me to mention these sorts of things.

0xdevalias:
The following comments were considered OOS in https://github.com/sparkletown/sparkle/pull/1047, and referenced this PR as where they will be addressed. Capturing here to make sure they aren't lost:

---

> Can we change this to use our standard 'guard' pattern?
> 
> ```suggestion
>   const messagesToDisplay = useMemo(
>     () => {
>       if (!chatsToDisplay || !user) return []
> 
>       return chatsToDisplay.map((chat) => (
> ```
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1047#discussion_r565020201_

---

> Why do we need to use `!!` here? Is `isInProfileModal` not already a `boolean` at its source? If not, that's where we should be doing this.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1047#discussion_r565020525_

package.json
```diff
@@ -26,7 +26,7 @@
     "bootswatch": "^4.5.0",
     "classnames": "^2.2.6",
     "date-fns": "^2.12.0",
-    "dayjs": "^1.8.35",
+    "dayjs": "^1.10.4",
```
0xdevalias, line 29: @mike-lvov What was the need for bumping this version here (if any). Generally it would be good to keep version updates as separate PRs, or at the very least, call out why it was needed for the changes made in this PR.

mike-lvov:
Will close this PR to remove all the noise and open a new one.

0xdevalias:
@mike-lvov for future traceability, can you please mention what the new PR you opened for this is

mike-lvov:
It is in #1179 now

PR #1080, opened 2021-01-18
Merge missing Camp features into PartyMap, then 💥 the old Camp components

0xdevalias:
Ensure the `PartyMap` components have all of the features from the old `Camp` components, then 💥 the old Camp components

**Note:** The bulk of this work was completed/tracked in https://github.com/sparkletown/sparkle/pull/840, along with interim reviews and related. This new PR is to reduce the cognitive effort of performing final reviews/cleanup before merging back into `staging`.

## Preview Deploy

- see URL in https://github.com/sparkletown/sparkle/pull/1080#issuecomment-762057287

## TODO

- [ ] remaining comments/issues in https://github.com/sparkletown/sparkle/pull/1080#issuecomment-762615593


src/api/bannerAdmin.ts
```diff
@@ -0,0 +1,16 @@
...
+export const makeUpdateBanner = (
+  venueId: string,
+  onError: (errorMsg: string) => void
+) => (message?: string) => {
+  const params = {
+    venueId,
+    bannerMessage: message ? message : "",
+  };
+
+  firebase
```
denisdimitrov, line 12: Is this intentionally without async/await?

src/components/organisms/BannerAdmin.tsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    [venueId]
+  );
+
+  const saveBanner = useCallback(() => {
+    if (!inputFieldRef.current) return;
+
+    updateBannerInFirestore(inputFieldRef.current.value);
+
+    return false;
```
denisdimitrov, line 35: return false?

src/components/organisms/BannerAdmin.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { currentVenueSelector } from "utils/selectors";
+
+import { useSelector } from "hooks/useSelector";
+
+export const BannerAdmin: React.FC = () => {
+  const currentVenue = useSelector(currentVenueSelector);
+  const venueId = currentVenue?.id;
+  const existingBannerMessage = currentVenue?.bannerMessage ?? "";
+
+  const inputFieldRef = useRef<HTMLInputElement>(null);
```
denisdimitrov, line 14: Is there a particular reason to use `ref` instead of `useState` ?

src/api/profile.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+  };
+
+  const firestore = firebase.firestore();
+
+  firestore
+    .doc(doc)
+    .update(newData)
+    .catch((e) => {
+      // TODO: Bugsnag.notify(e) ?
```
denisdimitrov, line 41: +1

src/api/bannerAdmin.ts
0xdevalias: I don't think so. I would assume I didn't necessarily think about it because it's not returning a value. But it would still be good to expose the fact that it's technically a `Promise`

src/components/organisms/BannerAdmin.tsx
0xdevalias: My guess is that it's intent was to prevent the default action from the `<button>` `onClick` handler, but that doesn't work in React:

- https://reactjs.org/docs/handling-events.html
  - > Another difference is that you cannot `return false` to prevent default behavior in React. You must call `preventDefault` explicitly.

src/components/organisms/BannerAdmin.tsx
0xdevalias: We're not storing any state here, but we are getting a reference to the `<input>` element/DOM node itself:

  - > You might be familiar with refs primarily as a way to access the DOM. If you pass a `ref` object to React with `<div ref={myRef} />`, React will set its `.current` property to the corresponding DOM node whenever that node changes.

0xdevalias: We're not storing any state here, but we are getting a reference to the `<input>` element/DOM node itself:

  - > You might be familiar with refs primarily as a way to access the DOM. If you pass a `ref` object to React with `<div ref={myRef} />`, React will set its `.current` property to the corresponding DOM node whenever that node changes.
  
This is being used in the `useCallback`'d `saveBanner` function to reference `inputFieldRef.current.value`, so that the `useCallback` memo doesn't change on every update to the input field (because a ref always has a stable reference for comparisons, only it's `.current` changes reference)

0xdevalias:
- [ ] Resolved?

Other comments from @denisdimitrov:
> Overall, looks good. I left some comments and clicked around on the preview link. Some of the styles are not correct and there are some transitions missing. For example, when hovering a room it changes the size directly without the transition animation. The blur is missing on the background of the name/count. And the border radius for the room image is missing.

src/api/profile.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/5379fe84c2ba33c352178141ac8473e8108c2788

src/components/organisms/BannerAdmin.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/d712013d6dc59ef2a8b177f0050bcf5eb30510ce

src/api/bannerAdmin.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/7d64d0f166ee714ac276a4c885ed2c8b1af69c15

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -197,232 +182,21 @@ $border-radius: 28px;
   padding: 10px;
 }
 
-.room {
```
0xdevalias, line 200: Note: The relevant room/etc styles from here were moved into `src/components/templates/PartyMap/components/Map/MapRoom.scss`

src/hooks/useKeyboardControls.ts
```diff
@@ -13,18 +13,17 @@ export const MovementKeys = {
...
   totalColumns: number;
   isSeatTaken: (row: number, column: number) => boolean;
   takeSeat: (row: number | null, column: number | null) => void;
   enterSelectedRoom: () => void;
-  onMove?: () => void;
 }
 
+// TODO: use e.preventDefault() or return false or similar in the keyboard handlers (check mousetrap docs) so we don't scroll on arrow key presses
+// TODO: we may also need some kind of 'centre me on the map' logic somewhere when we do this too? For large zoomed maps.
 // TODO: implement bindRef using useRef() or similar (in Camp?) then remove all withGlobalBind (default is false)
```
0xdevalias, line 27: [OOS]
> - use `e.preventDefault()` or `return false` or similar in the keyboard handlers (check mousetrap docs) so we don't scroll on arrow key presses
> - we may also need some kind of 'centre me on the map' logic somewhere when we do this too? For large zoomed maps.
> - implement `bindRef` using `useRef()` or similar (in PartyMap?) then remove all `withGlobalBind` (default is `false`)

src/pages/VideoAdmin/VideoAdmin.tsx
```diff
@@ -4,6 +4,8 @@ import firebase from "firebase/app";
 import { useVenueId } from "hooks/useVenueId";
 import { currentVenueSelectorData } from "utils/selectors";
 
+// @debt refactor BannerAdmin and this to DRY the common parts.
+//   The only difference in pattern is the update function being called + placeholder/etc labels
```
0xdevalias, line 8: [OOS]
> `@debt` refactor BannerAdmin and this to DRY the common parts.
>   The only difference in pattern is the update function being called + placeholder/etc labels

src/types/venues.ts
```diff
@@ -1,34 +1,34 @@
...
-}
+import { FormValues } from "pages/Admin/Venue/DetailsForm";
 
-interface TermOfService {
-  name: string;
-  text: string;
-  link?: string;
-}
+// TODO: should JazzBarVenue be added to this?
+export type AnyVenue = Venue | PartyMapVenue;
```
0xdevalias, line 13: > should JazzBarVenue be added to this?

src/types/venues.ts
```diff
@@ -98,13 +98,67 @@ export interface Venue {
   showZendesk?: boolean;
 }
 
+// @debt which of these params are exactly the same as on Venue? Can we simplify this?
+export interface PartyMapVenue extends Venue {
```
0xdevalias, line 102: > `@debt` which of these params are exactly the same as on Venue? Can we simplify this?

src/types/venues.ts
```diff
@@ -98,13 +98,67 @@ export interface Venue {
...
+
 export interface VenueConfig {
   theme: {
     primaryColor: string;
     backgroundColor?: string;
   };
 
-  landingPageConfig: VenueLandingPageConfig; // @debt should this be potentially undefined, or is it guaranteed to exist everywhere?
+  // @debt landingPageConfig should probably be undefined, or is it guaranteed to exist everywhere?
+  landingPageConfig: VenueLandingPageConfig;
```
0xdevalias, line 161: [OOS]
> `@debt` landingPageConfig should probably be undefined, or is it guaranteed to exist everywhere?

```suggestion
  landingPageConfig?: VenueLandingPageConfig;
```

src/types/venues.ts
```diff
@@ -156,6 +229,7 @@ export const urlFromImage = (
     : defaultValue;
 };
 
+// TODO: refactor this somewhere more useful, maybe utils?
 export const createJazzbar = (values: FormValues): Venue => {
```
0xdevalias, line 233: > refactor this somewhere more useful, maybe utils?

src/utils/useLocationUpdateEffect.ts
```diff
@@ -47,8 +46,8 @@ export const updateLocationData = (
 };
 
 // get Profile from the firebase
-// @debt rename this trackRoomEntered
-export const enterLocation = (
+// @debt rename this trackLocationEntered?
+export const trackRoomEntered = (
```
0xdevalias, line 50: [OOS]
> rename this trackLocationEntered?

src/components/templates/PartyMap/components/Map/hooks/usePartygoersOverlay.tsx
```diff
@@ -0,0 +1,71 @@
...
+  setSelectedUserProfile,
+}) => {
+  return useMemo(() => {
+    // @debt partygoers can be undefined because our types are broken so check explicitly
+    if (!showGrid || !partygoers || !userUid) return <div />;
+
+    // @debt workaround, sometimes partygoers are duplicated but the new ones don't have id's
+    const filteredPartygoers = partygoers.filter(
+      (partygoer) => partygoer && isTruthy(partygoer.id)
+    );
```
0xdevalias, line 47: > - `@debt` partygoers can be undefined because our types are broken so check explicitly
> - `@debt` workaround, sometimes partygoers are duplicated but the new ones don't have id's

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -1,199 +1,181 @@
...
-import { useUser } from "hooks/useUser";
-import { useSelector } from "hooks/useSelector";
 import { useKeyboardControls } from "hooks/useKeyboardControls";
-import { usePartygoers } from "hooks/users";
 
-import Sidebar from "components/molecules/Sidebar";
+// @debt refactor these hooks into somewhere more sensible
+import { useMapGrid } from "./hooks/useMapGrid";
+import { usePartygoersbySeat } from "./hooks/usePartygoersBySeat";
+import { usePartygoersOverlay } from "./hooks/usePartygoersOverlay";
```
0xdevalias, line 19: > `@debt` refactor these hooks into somewhere more sensible

This comment was made when we were going to be sharing them among multiple maps, but I'm not sure that that is the case anymore, so might be fine where they are?

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -1,199 +1,181 @@
...
-        });
+      makeUpdateUserGridLocation({
+        venueId,
+        userUid: user?.uid,
+        profileData,
+      })(row, column);
     },
-    [profile, user, venueId]
+    [profileData, user, venueId]
   );
```
0xdevalias, line 87: Can we refactor this to use `undefined` instead of `null`?

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -203,32 +185,39 @@ export const Map: React.FC<PropsType> = ({
     onSeatClick,
   });
 
+  // TODO: this probably doesn't even need to be a hook.. it's more of a component if anything. We can clean this up later
   const partygoersOverlay = usePartygoersOverlay({
```
0xdevalias, line 189: > this probably doesn't even need to be a hook.. it's more of a component if anything. We can clean this up later

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -4,78 +4,122 @@ import { useParams } from "react-router-dom";
...
-  );
+  const { user, profile } = useUser();
+  const currentVenue = useSelector(partyMapVenueSelector);
+  const venues = useSelector(orderedVenuesSelector);
 
-  const modalHidden = useCallback(() => {
-    setIsRoomModalOpen(false);
+  // TODO: can we just make this a boolean that is set from isTruthy(selectedRoom)?
+  const [isRoomModalOpen, setRoomModalOpen] = useState(false);
+  const [selectedRoom, setSelectedRoom] = useState<Room | undefined>();
```
0xdevalias, line 36: > can we just make this a boolean that is set from isTruthy(selectedRoom)?

Apparently the nuance of how this is used is used to neatly show/hide the `RoomModal` without it 'losing the data' as it's transitioning to hidden. Not sure if that is a big deal though?

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -4,78 +4,122 @@ import { useParams } from "react-router-dom";
...
+    setRoomModalOpen(true);
+  }, []);
+
+  const unselectRoom = useCallback(() => {
+    setSelectedRoom(undefined);
+    setRoomModalOpen(false);
   }, []);
 
+  // TODO: extract this into a reusable hook/similar
+  const enterRoom = useCallback(
```
0xdevalias, line 49: > extract this into a reusable hook/similar

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -4,78 +4,122 @@ import { useParams } from "react-router-dom";
...
+    (room: Room) => {
+      if (!room || !user) return;
+
+      // TODO: we could process this once to make it look uppable directly? What does the data key of venues look like?
+      const roomVenue = venues?.find((venue) =>
+        room.url.endsWith(`/${venue.id}`)
+      );
```
0xdevalias, line 56: > we could process this once to make it look uppable directly? What does the data key of venues look like?

src/types/venues.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/58906691ed0baa57fa53bf359d66839f100e6c5d

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -54,7 +58,7 @@ const NavSearchBar = () => {
 
     const roomsResults =
       venue && venue.rooms
-        ? (venue?.rooms as CampRoomData[]).filter((room) =>
+        ? (venue?.rooms as Room[]).filter((room) =>
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/12d266eba2172cc43eefd92518effc070d2cf1f1

src/api/profile.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// @debt this is a potential race condition. we should read the profileData + update
+// it in a transaction here (or a function) rather than using potentially stale data
+export const makeUpdateUserGridLocation = ({
+  venueId,
+  userUid,
+  profileData,
+}: MakeUpdateUserGridLocationProps) => (
+  row: number | null,
+  column: number | null
```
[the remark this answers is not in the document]
0xdevalias: I think this would require some data backfilling/otherwise handling things differently.. so feels OOS for this PR

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: I think for the same reason as https://github.com/sparkletown/sparkle/pull/1080#discussion_r559946760 this is probably OOS for this PR

src/components/templates/PartyMap/PartyMap.tsx
denisdimitrov: Can you use the `selectedRoom` instead of the state above to show/hide the modal?

denisdimitrov (dismissed):
LGTM!

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: I thought about that as well, but then I realised that some of the little UI bugs i'm seeing at the moment (eg. the modal closing instantly rather than 'smoothly') is because I joined these concepts together, and when I unset `selectedRoom`, the modal doesn't render anything.

I had some thoughts about how I could fix this, but need to get the 🧠  back into this code to check if they'll work, or if I just need to revert this to being two separate things entirely.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1080/commits/2c8cd7105dd5d01c7b58c1f7f9751c3fb7022ba5

0xdevalias:
From https://github.com/sparkletown/sparkle/pull/1080#issuecomment-762615593:

> * [ ]  the border radius for the room image is missing
>   
>   * @denisdimitrov What are you referring to by this? In looking at the generated styles in `staging` there didn't seem to be a border radius on the room image. Can you please point out more specifically where you mean?

@denisdimitrov are you able to elaborate on the border radius you were referring to? I'm not sure what you mean.

0xdevalias:
@denisdimitrov can I please get you to re-review this. There were a bunch of conflicts I had to merge in again when I updated this against `staging`, and want to get another set of 👀  to ensure I didn't accidentally lose/regress anything with it.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -56,14 +60,12 @@ const NavSearchBar = () => {
     );
 
     const roomsResults =
```
mike-lvov, line 62: `roomsResults: Room[] =`

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -1,31 +1,27 @@
...
-import { AvatarGridRoom } from "types/AvatarGrid";
-import { User } from "types/User";
+import { RoomModal } from "./RoomModal";
+import Announcement from "./Announcement";
+import ChatDrawer from "components/organisms/ChatDrawer";
+import UserProfileModal from "components/organisms/UserProfileModal";
+import UserProfilePicture from "components/molecules/UserProfilePicture";
 
-// Styles
 import "./AvatarGrid.scss";
```
mike-lvov, line 25: [OOS] I think this template should be deleted

src/components/templates/PartyMap/components/Map/hooks/useMapGrid.tsx
```diff
@@ -38,8 +39,8 @@ export const useMapGrid: ReactHook<UseMapGrid, UseMapGridReturn> = ({
             const column = colIndex + 1; // TODO: do these need to be here, can we zero index?
             const row = rowIndex + 1; // TODO: do these need to be here, can we zero index?
 
-            const seatedPartygoer = partygoersBySeat?.[row]?.[column] ?? null;
-            const hasSeatedPartygoer = !!seatedPartygoer;
+            const seatedPartygoer = partygoersBySeat?.[row]?.[column];
+            const hasSeatedPartygoer = seatedPartygoer !== undefined;
```
mike-lvov, line 43: What is the reason for changing it from `!!seatedPartygoer` to `eatedPartygoer !== undefined`?

src/components/templates/PartyMap/components/RoomAttendance.tsx
```diff
@@ -1,50 +1,60 @@
...
     () =>
-      partygoers?.filter(
+      partygoers.filter(
         (partygoer) => partygoer.lastSeenIn?.[`${venue.name}/${room.title}`]
       ),
     [partygoers, venue.name, room.title]
   );
 
-  const numberOfUsersInRoom = usersInRoom?.length;
-  if (numberOfUsersInRoom < 1) return <></>;
```
mike-lvov, line 28: I think the logic was changed here. Was it tested?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -21,12 +21,14 @@ import { usePartygoers } from "hooks/users";
 
 import UserList from "components/molecules/UserList";
 
-import { RoomModalOngoingEvent, ScheduleItem } from "../";
+import { RoomModalOngoingEvent, ScheduleItem } from "..";
```
mike-lvov, line 24: Not sure if we want to leave it as `..`. If so, should we also stick to `.`, instead of `./`?

src/components/templates/Playa/PlayaIcon.tsx
```diff
@@ -1,7 +1,6 @@
 import React from "react";
-import { PlayaIcon, Venue } from "types/Venue";
+import { AnyVenue, PlayaIcon, Venue } from "types/venues";
 import { WithId } from "utils/id";
-import { AnyVenue } from "types/Firestore";
```
mike-lvov, line 4: I thought Playa is completely removed from our codebase?

src/components/templates/Playa/PlayaVenue.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useRef } from "react";
 import { Overlay } from "react-bootstrap";
 import { DEFAULT_MAP_ICON_URL, PLAYA_VENUE_SIZE } from "settings";
-import { AnyVenue } from "types/Firestore";
+import { AnyVenue } from "types/venues";
```
mike-lvov, line 4: I thought Playa is completely removed from our codebase?

src/components/templates/Playa/Router.tsx
```diff
@@ -1,16 +1,19 @@
 import React from "react";
 import { useRouteMatch, Switch, Route } from "react-router-dom";
-import Playa from ".";
-import PlayaAdmin from "./PlayaAdmin";
+
 import useConnectCurrentVenue from "hooks/useConnectCurrentVenue";
 
+import { BannerAdmin } from "components/organisms/BannerAdmin";
+
+import Playa from ".";
```
mike-lvov, line 8: I thought Playa is completely removed from our codebase?

src/components/templates/Playa/VenuePreview.tsx
```diff
@@ -1,7 +1,11 @@
 import React, { useEffect, useState } from "react";
 import { FirebaseReducer } from "react-redux-firebase";
-import { Venue, VenuePlacementState } from "types/Venue";
-import { VenueEvent } from "types/VenueEvent";
+import {
+  Venue,
+  VenueEvent,
+  VenuePlacementState,
+  VenueTemplate,
+} from "types/venues";
```
mike-lvov, line 8: I thought Playa is completely removed from our codebase?

src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -69,13 +67,13 @@ export const AdminVenuePreview: React.FC<AdminVenuePreview> = ({
         );
       case VenueTemplate.partymap:
       case VenueTemplate.themecamp:
-        const campVenue = venue as WithId<CampVenue>;
+        const partyMapVenue = venue as WithId<PartyMapVenue>;
```
mike-lvov, line 70: This type should move to `props` instead of `as` assignment

mike-lvov (changes requested):
A handful of style-related comments. Some changes would be really nice to have in this PR

src/components/templates/Playa/VenuePreview.tsx
0xdevalias: It will be in the PR that removes it, which is separate to this PR, and will be merged after this PR. Both have been hanging around for ages, and this PR is the blocker (in my mental ordering/desire to complete things) of me working on the ‘delete all the old things’ PR.

src/components/templates/Playa/Router.tsx
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1080#discussion_r564114898

src/components/templates/Playa/PlayaVenue.tsx
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1080#discussion_r564114898

src/components/templates/Playa/PlayaIcon.tsx
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1080#discussion_r564114898

src/components/templates/PartyMap/components/Map/hooks/useMapGrid.tsx
0xdevalias: A general preference towards being more explicit in our intent with things, as that reduces the cognitive effort required to reason about it later. Potentially also a scope narrowing thing.

I don’t think I feel particularly strongly about this particular case either way, and I believe this particular field may also be able to be ‘null’ as well (as much as I dislike that it can be), so this change may not be accurate anyway.

src/components/templates/AvatarGrid/AvatarGrid.tsx
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1080#discussion_r564114898

src/components/templates/PartyMap/components/RoomAttendance.tsx
0xdevalias: (as a general side note, it’s far easier to see what you are referring to if you select all of the relevant lines. In this case, all I can see is the deleted lines, and so need to go and open the file/find this section/etc to be able to comment further)

0xdevalias: Previously it would just render nothing at all when there are no users in the room. Now it will still render the ‘container div’, but just won’t have anything in it.

I believe I did visually test this and it seemed to look fine (though for reasons that I believe are outside of the scope of this PR, I don’t think I was ever seeing any users in this room list, which I assume has/will be fixed with your users related PR, and/or will make more sense to track down/solve as a new PR once this + your users PR are merged)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: I don't actually have a strong opinion one way or the other about this. I wonder if there are best practices/linting rules that prefer one over the other?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: I believe the intent of your comment here is to then be able to remove the `roomsResults as Room[]` below. In which case, I agree.

0xdevalias: Looking at this, it is complicated because `venue?.rooms?` is `Room[] | AnyRoom[]`, which is probably why i did the `as` in the first place.

I'll add a `@debt` comment, but probably won't try and change it in this PR.

I believe once we delete all of the old stuff in https://github.com/sparkletown/sparkle/pull/833 we should then be able to clean this up properly.

src/components/templates/PartyMap/components/Map/hooks/useMapGrid.tsx
0xdevalias: Looking at the typing here, that doesn't seem to be any need for `null` to exist? So will probably leave this as is unless there's a good use case why we shouldn't somewhere (eg. the typing lying to us)

Screen Shot 2021-01-27 at 6 58 06 pm

src/pages/Admin/AdminVenuePreview.tsx
0xdevalias: Screen Shot 2021-01-27 at 7 01 22 pm

Looks like it was done this way to work around TypeScript. There would be ways we could 'fix this properly', but since this is existing code that i've just updated the types on, I think I will call it OOS for this PR.

I expect that once we can delete all of the old code in https://github.com/sparkletown/sparkle/pull/833, it will be simpler to see how we can properly fix/unify/simplify our types.

0xdevalias: It's also possible that we might be able to refactor `PartyMapContainer` to be able to take `WithId<Venue>` directly.. which would solve this here. Will check if that breaks anything then comment back here.

0xdevalias: Looks like the changes are going to be annoying to make, and it's not a big enough issue to be worth blocking this PR over (as it's more of a typing issue than a 'real' issue, as we already check that it's one of the correct template types in the `case` statement before this code runs.

So going to call this OOS for this PR.

mike-lvov (approved):
LGTM

PR #1081, opened 2021-01-18
[USERS SUB-PR] Changes profile loading logic

mike-lvov:
#1047 working on this PR, I have discovered that when there was even a lightest delay for users fetching, it produced "one-by-one addition to the store" bug. 

While I was trying to figure out why that was happening (I thought I solved that by having one query of users), I spent a few hours debugging that. My finding was that the `profile` store was actually adding the first item to the `users` store. It means that if we load the profile first, it will update the `users` store to this `worldUsers: [{.MY_PROFILE_OBJECT }]`. Because of this lib's optimization, I had to find a workaround for this.

This PR is a draft proposal. I don't see any other way to handle this so far, I'm opened to any suggestions since I spent the whole day trying to figure out the way to handle this.

### DIFF
--- a/src/hooks/useUser.ts
+++ b/src/hooks/useUser.ts
@@ -6,6 +6,7 @@ import { WithId } from "utils/id";
 import { authSelector, profileSelector } from "utils/selectors";
 
 import { useSelector } from "hooks/useSelector";
+import { useFirestoreConnect } from "hooks/useFirestoreConnect";
 
 type UseUserResult = {
   user: FirebaseReducer.AuthState | undefined;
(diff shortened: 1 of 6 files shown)


mike-lvov:
Just got an enlightening that it won't work for new users since they have a different flow(where the profile is probably getting fetched before the users)

0xdevalias:
This relates to the bug in https://github.com/prescottprue/react-redux-firebase/issues/914

0xdevalias:
@mike-lvov can you please add a comment explaining why you're closing PR's when you do, so that we have some context captured.

mike-lvov:
@0xdevalias This PR became unrelated since in this PR I try to find a workaround for 
https://github.com/prescottprue/react-redux-firebase/issues/914, but this PR
https://github.com/prescottprue/redux-firestore/pull/319 solves the issue. So, no need for workarounds

0xdevalias:
> @0xdevalias This PR became unrelated since in this PR I try to find a workaround for 
> 
> https://github.com/prescottprue/react-redux-firebase/issues/914, but this PR
> 
> https://github.com/prescottprue/redux-firestore/pull/319 solves the issue. So, no need for workarounds

Thanks for that :) The context just helps if we come back to it in future and wonder why it was closed. This way we have a simple summary to sum it all up and don't need to look deeper.

PR #1083, opened 2021-01-20
bug fixing for admin v1

denisdimitrov:
Some of the urls were missing in the `venuePaths` array which was being used to get the `venueId` in the hook. Because of this, the `venueId` was `undefined` when it wasn't. Also the url of the venue creation has been changed because there was a collision between: `/admin/venue/:venueId` and `/admin/venue/creation` 

Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/194

### DIFF
--- a/src/components/organisms/AppRouter/AppRouter.jsx
+++ b/src/components/organisms/AppRouter/AppRouter.jsx
@@ -62,7 +62,7 @@ const AppRouter = () => {
         <Route path="/admin_v2/venue/creation" component={VenueWizard_v2} />
         <Route path="/admin_v2/edit/:venueId" component={VenueWizard_v2} />
 
-        <Route path="/admin/venue/:venueId" component={Admin} />
+        <Route path="/admin/:venueId" component={Admin} />
         <Route path="/admin_v2/:venueId" component={Admin_v2} />
 
         <Route path="/admin" component={Admin} />
--- a/src/hooks/useVenueId.ts
+++ b/src/hooks/useVenueId.ts
@@ -9,7 +9,9 @@ const venuePaths = [
   "/e/:step/:venueId",
   "/in/:venueId",
   "/v/:venueId",
-  "/admin/venue/:venueId",
+  "/admin/:venueId",
+  "/admin/venue/edit/:venueId",
+  "/admin/venue/rooms/:venueId",
   "/admin_v2/:venueId",
   "/admin_v2/edit/:venueId",
 ];
--- a/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
+++ b/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
@@ -41,15 +41,12 @@ export const RoomsForm: React.FC = () => {
 
   useEffect(() => {
     const fetchVenueFromAPI = async () => {
-      console.log(venueId);
       if (!venueId) return history.replace("/admin");
 
       const venueSnapshot = await firestore
         .collection("venues")
         .doc(venueId)
         .get();
-
-      console.log(venueSnapshot);
 
       if (!venueSnapshot.exists) return history.replace("/admin");
       const data = venueSnapshot.data() as Venue;
@@ -58,7 +55,6 @@ export const RoomsForm: React.FC = () => {
         (template) => data.template === template.template
       );
 
-      console.log(template);
       if (!template || !HAS_ROOMS_TEMPLATES.includes(template.template)) {
         history.replace("/admin");
       }
--- a/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
+++ b/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
@@ -41,12 +41,15 @@ export const RoomsForm: React.FC = () => {
 
   useEffect(() => {
     const fetchVenueFromAPI = async () => {
+      console.log(venueId);
       if (!venueId) return history.replace("/admin");
 
       const venueSnapshot = await firestore
         .collection("venues")
         .doc(venueId)
         .get();
+
+      console.log(venueSnapshot);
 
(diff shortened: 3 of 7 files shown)


src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -62,7 +62,7 @@ const AppRouter = () => {
         <Route path="/admin_v2/venue/creation" component={VenueWizard_v2} />
         <Route path="/admin_v2/edit/:venueId" component={VenueWizard_v2} />
 
-        <Route path="/admin/venue/:venueId" component={Admin} />
+        <Route path="/admin/:venueId" component={Admin} />
```
0xdevalias, line 65: Yaaaas! I've wanted to change this for SO long! 🎉

src/hooks/useVenueId.ts
```diff
@@ -9,7 +9,9 @@ const venuePaths = [
   "/e/:step/:venueId",
   "/in/:venueId",
   "/v/:venueId",
-  "/admin/venue/:venueId",
+  "/admin/:venueId",
+  "/admin/venue/edit/:venueId",
+  "/admin/venue/rooms/:venueId",
```
0xdevalias, line 14: [Potentially OOS] I backlogged https://github.com/sparkletown/internal-sparkle-issues/issues/164 in the past, about how we should be able to get rid of this 'paths' hack entirely.

Essentially, if we use `useParams`, it will be able to extract `:venueId` no matter where it exists in the url path:

- https://reactrouter.com/web/api/Hooks/useparams

Which I believe might even reduce this hook down to basically a single line.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -41,20 +41,24 @@ export const RoomsForm: React.FC = () => {
...
+
       if (!venueSnapshot.exists) return history.replace("/admin");
       const data = venueSnapshot.data() as Venue;
       //find the template
       const template = ALL_VENUE_TEMPLATES.find(
         (template) => data.template === template.template
       );
 
+      console.log(template);
       if (!template || !HAS_ROOMS_TEMPLATES.includes(template.template)) {
```
0xdevalias, line 62: Remove these `console.log`s

src/pages/Admin/Venue/VenueWizard.tsx
```diff
@@ -55,6 +55,7 @@ const reducer = (
 export const VenueWizard: React.FC = () => {
   const venueId = useVenueId();
 
+  console.log(venueId);
```
0xdevalias, line 58:
```suggestion
```

src/pages/Admin/Venue/VenueWizard.tsx
```diff
@@ -101,6 +102,7 @@ const VenueWizardCreate: React.FC = () => {
   const { user } = useUser();
   const queryParams = useQuery();
 
+  console.log("asd");
```
0xdevalias, line 105:
```suggestion
```

0xdevalias (changes requested):
[CRO] A few little changes + suggestions to make this more robust

src/hooks/useVenueId.ts
denisdimitrov: I agree with this but lets fix it in a separate PR.

denisdimitrov:
@0xdevalias The console logs have been removed and the conversations about them have been resolved. I agree with the venueId param but lets do it in another PR.

mike-lvov (approved):

denisdimitrov:
@0xdevalias I think merging is blocked because of your previous request changes. Please merge this as soon as you approve it because Thomas needs these changes.

src/hooks/useVenueId.ts
0xdevalias: Sure

0xdevalias:
> @0xdevalias I think merging is blocked because of your previous request changes. Please merge this as soon as you approve it because Thomas needs these changes.

Yup, that does appear to be why merging is blocked at the moment. I'll review/approve/merge in a sec.

While in general it's good practice to let the original reviewer re-review to remove this in the 'normal way' (so that they can ensure all of their concerns were adequately fixed), in 'special cases' where it's an urgent bug fix/similar, it's fine to use GitHub's 'dismiss review' feature:

image

When you use this, please make sure to include a bit of context/description so that we can look back and see why that review was dismissed in future:

image

If a review was dismissed in this way, but the PR was still approved by another dev overall, then I don't consider this a 'break glass' fix, and so you don't need to apply that label (see below for more details)

---

At a high level, the 'engineering culture' I want to encourage is that each individual contributor is empowered to be able to make the changes they need, as they need to. When we have the luxury of time, then I definitely want to make sure we follow our standard review flows (which is approved by at least 1 other dev), but I also trust the judgement of the team to know when it is ok to 'step outside those flows' to unblock something. In the cases where we do 'step outside those flows', I want to make sure that we have enough context/traceability to be able to see where that happened, and why (eg. standardizing how we 'mark' an issue when we use a 'break glass' approval bypass (see https://github.com/sparkletown/internal-sparkle-issues/issues/13#issuecomment-739734800), etc)

(I'll also make sure to copy/reference this in https://github.com/sparkletown/sparkle/pull/1062 so we can document it properly going forward)

0xdevalias (approved):
[CRO] Previous review comments were fixed up. LGTM!

src/hooks/useVenueId.ts
0xdevalias: FYI: I believe https://github.com/sparkletown/internal-sparkle-issues/issues/194#issuecomment-765080430 is related to the change I requested here.

0xdevalias: see https://github.com/sparkletown/sparkle/pull/1091 for fix

PR #1076, opened 2021-01-14
[sub-PR] Constrain useFirestoreConnect query types to enforce our firestore query patterns

0xdevalias:
Sub-PR to https://github.com/sparkletown/sparkle/pull/801 that constrains the `useFirestoreConnect` query types to enforce our firestore query patterns.

This reinstates and improves upon the original type narrowing intended by https://github.com/sparkletown/sparkle/pull/784, as called out in https://github.com/sparkletown/sparkle/pull/801#issuecomment-759752827

PR #1091, opened 2021-01-22
Refactor useVenueId to use useParams instead of brittle venuePaths hack

0xdevalias:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/164
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/194

This should also fix an edge case bug that was missed in https://github.com/sparkletown/sparkle/pull/1083 that causes the 'map banner' to not be editable (gets stuck on an infinite loading screen)

> I think we could just use `useParams()` here instead of needing to loop through all of the `venuePath`s, so long as they use `:venueId` in their path
> 
> - https://reactrouter.com/web/api/Hooks/useparams
>   - > `useParams` returns an object of key/value pairs of URL parameters. Use it to access `match.params` of the current `<Route>`
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1076#discussion_r557069059_

### DIFF
--- a/src/hooks/useVenueId.ts
+++ b/src/hooks/useVenueId.ts
@@ -1,37 +1,14 @@
-import { matchPath, useHistory } from "react-router-dom";
+import { useParams } from "react-router-dom";
 
-type VenueRoute = {
-  venueId: string;
-};
-
-// Sometimes in a nested route we want venue ID
-const venuePaths = [
-  "/e/:step/:venueId",
-  "/in/:venueId",
-  "/v/:venueId",
-  "/admin/:venueId",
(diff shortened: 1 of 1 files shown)


0xdevalias:
@sparkletown/engineering-leads 🔨 🍷 breaking glass to deploy this as it is breaking functionality in production at the moment.

0xdevalias (commented):
@sparkletown/engineering-leads 🔨 🍷 breaking glass to deploy this as it is breaking functionality in production at the moment.

PR #1095, opened 2021-01-22
fix audience avatars and route paths

denisdimitrov:
Fixes the z-index in the audience template, avatars are now showing up underneath the video.

Fixes the route paths, they were using `match.url` instead of `match.path`. Example:
If path is `venue/:venueId` and url is `venue/denismusic` then the url replaces the path, and the params object is empty because path now looks like this: `venue/denismusic` and the venueId param doesn't exist anymore. Which then means that `useVenueId` returns `undefined` which is missleading.  This is necessary change and made here on purpose because audience templates were not working because of this issue.

Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/201

### DIFF
--- a/src/hooks/useVenueId.ts
+++ b/src/hooks/useVenueId.ts
@@ -1,14 +1,18 @@
 import { useParams } from "react-router-dom";
 
+interface ParamsTypes {
+  venueId?: string;
+}
+
 /**
  * Retrieve the venueId from the URL path.
  *
  * @see https://reactrouter.com/web/api/Hooks/useparams
  */
 export const useVenueId: () => string | undefined = () => {
-  const { venueId } = useParams();
+  const { venueId } = useParams() as ParamsTypes;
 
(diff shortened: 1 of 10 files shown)


src/hooks/useVenueId.ts
```diff
@@ -1,14 +1,18 @@
...
  * Retrieve the venueId from the URL path.
  *
  * @see https://reactrouter.com/web/api/Hooks/useparams
  */
 export const useVenueId: () => string | undefined = () => {
-  const { venueId } = useParams();
+  const { venueId } = useParams() as ParamsTypes;
 
-  if (typeof venueId === "string") return venueId;
+  if (venueId && typeof venueId === "string") return venueId;
```
0xdevalias, line 15: Nice catch!

0xdevalias (commented):
Retroactively approved. Nice catch and clean solution!

PR #1098, opened 2021-01-23
fix NavSearchBar search result key stability + allow vertical scrolling

0xdevalias:
- Ensure we don't use `index` for React component's `key` property.
  - Because it doesn't uniquely identify the individual result, it causes issues when the item changes (but still has the same array index), which causes update issues (in this case, user profile pictures were showing for the wrong user, as it wasn't updated when the person in that 'search result slot' changed)
- Sets a max-height on `.nav-search-results` + enable scrolling

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/209
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/210

### DIFF
--- a/src/components/molecules/NavSearchBar/NavSearchBar.scss
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.scss
@@ -1,3 +1,5 @@
+@import "scss/constants.scss";
+
 .nav-search-links {
   display: flex;
   position: relative;
@@ -68,12 +70,17 @@
 }
 
 .nav-search-results {
-  position: absolute;
-  min-width: 290px;
+  z-index: z(nav-search-results);
+
   position: fixed;
-  z-index: 380;
   top: 60px;
+
+  min-width: 290px;
+
   height: auto;
+  max-height: 75vh;
+  overflow-y: auto;
+
   padding: 20px;
   box-shadow: 0 10px 30px 0 rgba(0, 0, 0, 0.5);
   border-radius: 0 0 24px 24px;
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -20,14 +20,17 @@ interface SearchResult {
 
 const NavSearchBar = () => {
   const [searchQuery, setSearchQuery] = useState("");
+
   const [searchResult, setSearchResult] = useState<SearchResult>({
     rooms: [],
     users: [],
     events: [],
   });
+
   const [selectedUserProfile, setSelectedUserProfile] = useState<
     WithId<User>
   >();
+
   const [selectedRoom, setSelectedRoom] = useState<CampRoomData>();
 
   const venue = useSelector(currentVenueSelectorData);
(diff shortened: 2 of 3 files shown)


src/components/molecules/NavSearchBar/NavSearchBar.scss
```diff
@@ -68,12 +70,17 @@
 }
 
 .nav-search-results {
-  position: absolute;
```
0xdevalias, line 71: Removed as it was already overridden by the `position: fixed` below, so it wasn't actually doing anything.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -79,72 +82,78 @@ const NavSearchBar = () => {
     <div className="nav-search-links">
       <div className="nav-search-icon" />
       <NavSearchBarInput value={searchQuery} onChange={setSearchQuery} />
+
       {isTruthy(searchQuery) && (
         <div className="nav-search-close-icon" onClick={clearSearchQuery} />
       )}
-      {isTruthy(numberOfSearchResults) && (
-        <>
```
0xdevalias, line 86: Removed the unnecessary fragment, as we are already within a single `div` anyway.

(Reviewing this with 'show whitespace changes' turned off will minimize the changes you need to review)

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -79,72 +82,78 @@ const NavSearchBar = () => {
...
+          <div className="nav-search-result-number">
+            <b>{numberOfSearchResults}</b> search results
+          </div>
+
+          {/* @debt we really shouldn't be using the index as part of the key here, it's unstable.. but rooms don't have a unique identifier */}
+          {searchResult.rooms.map((room, index) => {
+            return (
+              <div
+                className="row"
+                key={`room-${room.title}-${index}`}
```
0xdevalias, line 101: While we shouldn't be using index at all, this is sort of a hacky workaround since rooms don't technically have a unique identifier.

It's far less likely that we will find a different room that has the exact same `title` that ends up in the same `index`, so while it's still possible for this to be incorrect.. it's WAY less likely to occur.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -79,72 +82,78 @@ const NavSearchBar = () => {
...
-                    <div>{event.name}</div>
-                    <div>Event</div>
-                  </div>
+              </div>
+            );
+          })}
+
+          {searchResult.events.map((event) => {
+            return (
+              <div className="row" key={`event-${event.id ?? event.name}`}>
```
0xdevalias, line 120: The typing says that `event.id` can apparently be `undefined` here, so I figured defaulting to `event.name` was a good fallback. 

Again, not a perfect solution, ideally we would just always use `event.id` as it would be a unique value.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -79,72 +82,78 @@ const NavSearchBar = () => {
...
-                  <div className="result-info">
-                    <div key={`user-${index}`}>{user.partyName}</div>
-                  </div>
+                  className="result-avatar"
+                  style={{
+                    backgroundImage: `url(${user.pictureUrl})`,
+                  }}
+                />
+                <div className="result-info">
+                  <div>{user.partyName}</div>
```
0xdevalias, line 143: There is no need to set the `key` prop on an element unless it is the first child component of a `map` or similar. I assume this was leftover from older code before we wrapped it in the other `div`.

src/scss/constants.scss
```diff
@@ -42,6 +42,7 @@ $playa-venue-live: rgb(125, 223, 194);
 $profile-image-bg-color: #999999;
 
 $z-layers: (
+  nav-search-results: 380,
```
0xdevalias, line 45: This is the same `z-index` value that was already being used, just moved to our centralised helper. I didn't want to  expand the risk of this PR by changing it here. At some point we should align all of our `z-index` values in this position, but probably with 'larger gaps' between them (like each 'new layer' would be 10 or 100 different from the last, then minor tweaks made in the digits between)

0xdevalias:
@sparkletown/tech-leads 🔨🍷 breaking glass to deploy as this is a required bug fix for an upcoming event, anno onene else is around to review/approve at the moment.

PR #1105, opened 2021-01-27
add venue admin page for all templates

denisdimitrov:
- New route that can be accessed by any venue template in order to change the admin banner
- Clean up and removal of Router components that are no longer necessary.
- Refactoring in some components to keep the consistency
- Add `AnnouncementMessage` component to the rest of the venue templates.
- Remove deprecated `Announcement` component

fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/192


src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isCurrentVenueRequestedSelector,
+  isCurrentVenueRequestingSelector,
+  venueSelector,
+} from "utils/selectors";
+
+import { IFRAME_TEMPLATES } from "settings";
+
+import { useSelector } from "hooks/useSelector";
+import { useUserIsVenueOwner } from "hooks/useUserIsVenueOwner";
+import useConnectCurrentVenue from "hooks/useConnectCurrentVenue";
```
0xdevalias, line 13:
```typescript
/**
 * @deprecated use useConnectCurrentVenueNG instead
 * @see useConnectCurrentVenueNG
 */
export const useConnectCurrentVenue = () => {
```

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  if (!isLoggedIn) {
+    return <div className="admin-page-title">You need to log in first.</div>;
+  }
+
+  if (!isVenueOwner) {
+    return (
+      <div className="admin-page-title">{`You don't have the permissions to access this page`}</div>
+    );
+  }
```
0xdevalias, line 50: This will provide an 'oracle' that would allow valid `venueId`'s to be determined. At the very least we probably want the `isLoggedIn` check to happen first.

And there's probably no point even loading the venue data/etc if they aren't already logged in.

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const isVideoVenue = IFRAME_TEMPLATES.includes(venue.template);
+
+  return (
+    <>
```
0xdevalias, line 55: Do we really want these elements to be output as is (which is what this fragment will do)? Or do we want them to be wrapped in a 'containing `div`'?

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return <div className="admin-page-title">This venue does not exist</div>;
+  }
+
+  if (!isLoggedIn) {
+    return <div className="admin-page-title">You need to log in first.</div>;
+  }
+
+  if (!isVenueOwner) {
+    return (
+      <div className="admin-page-title">{`You don't have the permissions to access this page`}</div>
```
0xdevalias, line 48:
```suggestion
      <div className="admin-page-title">You don't have the permissions to access this page</div>
```

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -1,19 +1,18 @@
...
 import { PlayaRouter } from "components/templates/Playa/Router";
 import { PartyMapRouter } from "components/templates/PartyMap/PartyMapRouter";
-
-import { FriendShipPage } from "pages/FriendShipPage";
+import { FireBarrel } from "components/templates/FireBarrel";
+import { Audience } from "components/templates/Audience/Audience";
 import { WithNavigationBar } from "components/organisms/WithNavigationBar";
-
-import { Venue, VenueTemplate } from "types/venues";
+import { AvatarGrid } from "components/templates/AvatarGrid";
```
0xdevalias, line 15: Import sorting seems off here?

Why did the `FireBarrel`/`Audience`/`AvatarGrid` `template/`s move to the bottom of the list? `FriendShipPage` is fine at the top since it's a `page/`

src/hooks/useUserIsVenueOwner.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const useUserIsVenueOwner = (): boolean => {
+  const { user } = useUser();
+  const currentVenue = useSelector(venueSelector);
+
+  if (!currentVenue || !user) return false;
+
+  const isVenueOwner = currentVenue.owners.includes(user.uid);
+
+  return isVenueOwner;
```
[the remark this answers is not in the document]
0xdevalias: Also.. we very rarely actually have venue owners set directly on most venues. This permission is usually inherited from the 'root parent venue' (which is being called `sovereignVenue` in https://github.com/sparkletown/sparkle/pull/1047, which implements `src/hooks/useSovereignVenueId.ts`)

So the logic here is going to lock most `owners` out of being able to use this feature as they would expect.

I'm not sure exactly where/how we handle this logic currently, but it does exist in production at the moment (even without the changes in https://github.com/sparkletown/sparkle/pull/1047). It would be good to find where, and implement the same logic (ideally using some form of common helper functions rather than copy/paste replicating it)

From memory, we should check if there is a `currentVenue.parentVenueId`, if so, load that and check `owners`. If still no match, check if there is a `parentVenueId` on that, if so, load it and check `owners`, etc (repeat until you reach the 'root of the tree', which is when there is no longer a `parentVenueId`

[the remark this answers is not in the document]
0xdevalias: I expect this check would be far simpler to implement as a direct `firestore` function call within this hook (not using the redux connect patterns), properly memoised to only need to re-run when the exact dependencies have changed (in this case, that is probably `venueId` and `userId` (not `venue`/`user`, as that is 'too wide of a scope')

It could potentially even be implemented in a recursive way (though if we want to be super robust, we would probably want to keep track of which `venueId`s we've already checked so that we can't get into an infinite loop.

src/types/venues.ts
```diff
@@ -178,6 +178,14 @@ export interface JazzbarVenue extends Venue {
...
     icon: string;
   };
+  owners: string[];
+  theme?: {
+    primaryColor: string;
+    backgroundColor?: string;
+  };
+  id: string;
+  rooms?: AnyRoom[];
+  mapBackgroundImageUrl?: string;
```
0xdevalias, line 188: These seem unrelated to this PR/any changes I've seen in it? Where did they come from?

src/types/venues.ts
```diff
@@ -46,7 +46,7 @@ export interface Venue_v2_Base {
   host: {
     icon: string;
   };
-  owners?: string[];
+  owners: string[];
```
0xdevalias, line 49: Similar concerns/caveats as other places in this PR where I have mentioned making this optional, and whether we can guarantee that it will be set for all venues (including currently existing data within firestore)

src/components/templates/FireBarrel/FireBarrel.tsx
```diff
@@ -1,25 +1,28 @@
...
+import { currentVenueSelector } from "utils/selectors";
 
-import * as S from "./FireBarrel.styled";
 import { useVideoState } from "./useVideo";
-import { useUser } from "hooks/useUser";
-import { usePartygoers, useUsersById } from "hooks/users";
-
-import VideoErrorModal from "components/organisms/Room/VideoErrorModal";
 import LocalParticipant from "../Playa/Video/LocalParticipant";
 import RemoteParticipant from "../Playa/Video/RemoteParticipant";
```
0xdevalias, line 17:
```suggestion
import LocalParticipant from "components/templates/Playa/Video/LocalParticipant";
import RemoteParticipant from "components/templates/Playa/Video/RemoteParticipant";
```

0xdevalias (dismissed):
Some comments/cleanup/simplifications that I think can be done here + ensuring we don't lose some of the code cleanup that was made in things that I assume were merged into this from `staging`.

src/api/admin.ts
```diff
@@ -92,7 +92,7 @@ export type VenueInput = AdvancedVenueInput &
     height?: number;
     bannerMessage?: string;
     parentId?: string;
-    owners?: string[];
+    owners: string[];
```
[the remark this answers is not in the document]
denisdimitrov: This is one of the faulty types. This field always exists. The moment you create a venue, this field is created as well with your user id in it.

src/pages/Admin/Venue/VenueAdminPage.tsx
denisdimitrov: The symbol `'` leaves a lint error and underlines the text saying that it should be escaped.

src/pages/VenuePage/TemplateWrapper.tsx
denisdimitrov: Maybe I don't remember the pattern correctly. I believe the components and scss were supposed to be at the bottom of the import list, no?

src/types/venues.ts
denisdimitrov: I think this is a rebase/merge thing. I will double check and remove it if it's not being used.

src/components/templates/PartyMap/PartyMapRouter.tsx
```diff
@@ -13,7 +11,6 @@ export const PartyMapRouter: FC = () => {
 
   return (
     <Switch>
-      <Route exact path={`${match.path}/admin`} component={BannerAdmin} />
       <Route path={`${match.path}/:roomTitle`} component={PartyMap} />
       <Route path={`${match.path}/`} component={PartyMap} />
     </Switch>
```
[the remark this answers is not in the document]
denisdimitrov: To be honest, I can't see this `/:roomTitle` anywhere in partymap. Clicking on a room show a modal and not a route. I think I can safely remove this one.

src/components/templates/Jazzbar/JazzbarRouter/JazzbarRouter.tsx
```diff
@@ -2,15 +2,13 @@ import React from "react";
...
 
 export const JazzbarRouter: React.FC = () => {
   const match = useRouteMatch();
 
   return (
     <Switch>
       <Route path={`${match.path}/band`} component={ReactionPage} />
-      <Route path={`${match.path}/admin`} component={VideoAdmin} />
       <Route path={`${match.path}/`} component={JazzBar} />
     </Switch>
```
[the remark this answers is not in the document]
denisdimitrov: `/band` doesn't seem to be used anywhere as well. It's probably a legacy thing.

src/components/templates/Audience/Audience.tsx
```diff
@@ -303,6 +304,7 @@ export const Audience: React.FunctionComponent = () => {
 
     const renderReactionsContainer = () => (
       <>
+        <AnnouncementMessage message={venue.bannerMessage} />
```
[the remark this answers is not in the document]
denisdimitrov: Nice catch.

src/components/templates/AvatarGrid/Announcement.tsx
```diff
@@ -1,45 +0,0 @@
-import React, { useEffect, useState } from "react";
```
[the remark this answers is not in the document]
denisdimitrov: Actually they both existed. This is just removed in favor to use `AnnouncementMessage`

src/components/templates/Playa/Router.tsx
```diff
@@ -13,7 +10,6 @@ export const PlayaRouter: React.FunctionComponent = () => {
 
   return (
     <Switch>
-      <Route exact path={`${match.path}/admin`} component={BannerAdmin} />
       <Route path={`${match.path}/:camp`} component={Playa} />
       <Route path={`${match.path}/`} component={Playa} />
     </Switch>
```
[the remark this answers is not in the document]
denisdimitrov: Not much. I'll do it.

src/components/templates/AvatarGrid/Router.tsx
```diff
@@ -1,20 +0,0 @@
...
-
-import useConnectCurrentVenue from "hooks/useConnectCurrentVenue";
-
-import { BannerAdmin } from "components/organisms/BannerAdmin";
-
-import AvatarGrid from ".";
-
-export const AvatarRouter: FC = () => {
-  const match = useRouteMatch();
-  useConnectCurrentVenue();
```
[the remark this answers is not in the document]
denisdimitrov: That's correct but the template is not being used. Should I even bother to fix this?

src/hooks/useUserIsVenueOwner.ts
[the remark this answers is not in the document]
denisdimitrov: I need this kind of logic to allow and prevent people from accessing the VenueAdminPage because only the owner should be able to edit the banner. And each venue has it's own banner and owners. I don't think `parentId` is relevant here. Even if the venue is a child of another venue it  should have it's own banner and different owners. Hypothetically the best way imo is to fetch all related venues merge all of the owners and check if you are there. Wdyt?

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -158,6 +158,7 @@ export const validationSchema = Yup.object()
       then: Yup.string().required("Radio station (stream) is required!"),
     }),
 
+    owners: Yup.array<string>().required(),
```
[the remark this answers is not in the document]
denisdimitrov: This has always been provided by the frontend. It was a false type.

src/types/venues.ts
denisdimitrov: False type, it's not going to be a problem.

src/pages/Admin/Venue/VenueAdminPage.tsx
0xdevalias: Cool, good reason to use this method then :)

src/components/templates/Jazzbar/JazzbarRouter/JazzbarRouter.tsx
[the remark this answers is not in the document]
0xdevalias: I haven't checked it out, but possibly it's a screen that the 'band' can watch to see the reactions without needing to be in the 'main venue'? Not sure if anyone uses that, could be something to check in with @sparkleed / the rest of the team.

This is definitely OOS of this PR though.

src/components/templates/AvatarGrid/Router.tsx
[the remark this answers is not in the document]
0xdevalias: Nah, probably don't worry about it. It will get deleted in https://github.com/sparkletown/sparkle/pull/833 anyway

src/components/templates/Playa/Router.tsx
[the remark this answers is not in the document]
0xdevalias: Just noticed, this will be deleted in https://github.com/sparkletown/sparkle/pull/833 anyway, so don't worry about this one.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: I have a plan to fix this in an automated/linted way across the codebase when I get a chance (see https://github.com/sparkletown/internal-sparkle-issues/issues/156) so don't worry too much about this.

The general pattern is 'most distant/big' thing (eg. 3rd party library) to 'most specific thing' (eg. the stylesheet that relates to this component directly). So pages/templates/organisms/molecules would be in that order in between those two 'extremes', etc.

src/pages/Admin/Venue/DetailsValidationSchema.ts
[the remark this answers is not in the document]
0xdevalias: Awesome. Good to know!

src/hooks/useUserIsVenueOwner.ts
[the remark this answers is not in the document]
0xdevalias: Nope, that's not true. I don't know where in the code this is implemented currently, but the 'owners' of a 'root parent venue' are effectively 'owners' of any of those 'child venues', even though they aren't specifically in that child venue's 'owners' data.

Or said simpler: If I can admin 'root parent venue', I can admin all of it's child venues.

This happened a while back, I think it may have been a feature @thecadams worked on (at least I don't remember personally implementing it), but was basically done to reduce the 'administrative load' for setting up/running events. Instead of needing to go and add people to each venue individually, they can just be added to the one 'main' venue (usually the partymap).

src/types/venues.ts
0xdevalias: Sounds good

src/components/molecules/AdminVideo/AdminVideo.tsx
```diff
@@ -26,14 +30,16 @@ const VideoAdmin: React.FC = () => {
 
   const saveIframeUrl = () => updateIframeUrl(iframeUrl);
 
+  if (!isVenueOwner) {
+    return <>{`You don't have the permissions to access this page`}</>;
```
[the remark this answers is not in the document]
0xdevalias: description in https://github.com/sparkletown/sparkle/pull/1105#discussion_r565901530

src/components/templates/PartyMap/PartyMapRouter.tsx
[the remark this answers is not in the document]
0xdevalias: I don't think it can be removed. I remember that we had a bug a while back and the fix was adding this route.

I believe this is used when it's a sub-venue like a jazz bar.

Clicking a room will open the modal, but entering a room (that isn't a link off-platform like zoom) will use this route.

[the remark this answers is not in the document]
0xdevalias: I think we can probably treat this question as OOS for this PR.

If we moved it to `AppRouter` as-is, then it will enable that route even for templates that don't support subvenues, which feels wrong.

It's not a huge issue still having these 'child routers' here for now.

src/api/admin.ts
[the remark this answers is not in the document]
0xdevalias: Awesome, sounds good!

src/hooks/useUserIsVenueOwner.ts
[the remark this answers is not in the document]
denisdimitrov: I stand corrected, will fix this.

src/pages/Admin/Venue/VenueAdminPage.tsx
denisdimitrov: Since there are no styles there is no point in rendering an extra element in the DOM. If someone needs to style these he can just replace the fragment with `div`

src/components/molecules/AdminVideo/AdminVideo.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const AdminVideo: React.FC<AdminVideoProps> = ({ venueId, venue }) => {
+  const [iframeUrl, setIframeUrl] = useState("");
+  const [error, setError] = useState<string | null>();
+
+  useEffect(() => {
+    setIframeUrl(venue?.iframeUrl || "");
+  }, [venue]);
+
+  const updateIframeUrl = (iframeUrl: string | null) => {
```
mike-lvov, line 21: [OOS] this should move to api/ folder

mike-lvov: When will it be `null?`

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -27,7 +27,7 @@ import "./AvatarGrid.scss";
 const DEFAULT_COLUMNS = 40;
 const DEFAULT_ROWS = 25;
 
-const AvatarGrid = () => {
+export const AvatarGrid = () => {
```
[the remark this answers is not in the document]
mike-lvov: Just for my info. Is this template should be maintained?

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -27,6 +27,7 @@ const partyMapVenueSelector = (state: RootState) =>
   state.firestore.ordered.currentVenue?.[0] as PartyMapVenue;
 
 export const PartyMap: React.FC = () => {
+  useConnectCurrentVenue();
```
mike-lvov, line 30: [OOS] This should use `useConnectCurrentVenueNG`.

src/hooks/useUserIsVenueOwner.ts
```diff
@@ -0,0 +1,19 @@
+import { parentVenueSelector } from "utils/selectors";
+import { useConnectCurrentVenueNG } from "./useConnectCurrentVenueNG";
+import { useSelector } from "./useSelector";
+import { useUser } from "./useUser";
+import { useVenueId } from "./useVenueId";
+
+export const useUserIsVenueOwner = (): boolean => {
```
mike-lvov, line 7: useIsUserVenueOwner? (to amplify that it is a boolean)

src/hooks/useUserIsVenueOwner.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { user } = useUser();
+
+  const currentVenueId = useVenueId();
+  const { currentVenue } = useConnectCurrentVenueNG(currentVenueId);
+  const parentVenue = useSelector(parentVenueSelector);
```
mike-lvov, line 12: [OOS] Should eventually become useParentVenue

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useUserIsVenueOwner } from "hooks/useUserIsVenueOwner";
+import { useUser } from "hooks/useUser";
+
+import { LoadingPage } from "components/molecules/LoadingPage";
+import { AdminVideo } from "components/molecules/AdminVideo";
+import { BannerAdmin } from "components/organisms/BannerAdmin";
+
+import "./VenueAdminPage.scss";
+import { useVenueId } from "hooks/useVenueId";
+import { useConnectCurrentVenueNG } from "hooks/useConnectCurrentVenueNG";
```
mike-lvov, line 20: not correct order

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const VenueAdminPage: React.FC = () => {
+  useConnectCurrentVenueNG();
```
mike-lvov, line 23: redundant

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const { profile, user } = useUser();
+  const venueId = useVenueId();
+  const { currentVenue: venue } = useConnectCurrentVenueNG(venueId);
+  console.log(venue);
```
mike-lvov, line 28: ?

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const venueRequestStatus = useSelector(isCurrentVenueNGRequestedSelector);
+  const venueRequestingStatus = useSelector(isCurrentVenueNGRequestingSelector);
+
+  const isVenueOwner = useUserIsVenueOwner();
+  const isLoadingVenue = venueRequestingStatus || !venueRequestStatus;
```
mike-lvov, line 33: isVenueLoading

src/utils/selectors.ts
```diff
@@ -155,9 +168,7 @@ export const experienceSelector = (state: RootState) =>
   state.firestore.data.experience;
 
 export const venueSelector = (state: RootState) =>
```
mike-lvov, line 170: currentVenueSelector?

src/components/templates/PartyMap/PartyMap.tsx
denisdimitrov: It should be the rest of the code is using the default one and I don't have the time to change it atm.

src/components/molecules/AdminVideo/AdminVideo.tsx
denisdimitrov: Good catch.

src/pages/Admin/Venue/VenueAdminPage.tsx
denisdimitrov: It's not, this is a new route now and needs the connect. otherwise the venue is undefined and not in the redux.

src/utils/selectors.ts
denisdimitrov: Should be changed in another PR there might be a lot of usage for this one.

denisdimitrov:
@mike-lvov All comments have been fixed, please take another look.

>Just for my info. Is this template should be maintained?

AvatarGrid is legacy template and is no longer being used.

mike-lvov (approved):
Taken the urgency, it is good enough to be merged in

src/components/templates/PartyMap/PartyMapRouter.tsx
```diff
@@ -1,18 +0,0 @@
...
-
-import PartyMap from "./PartyMap";
-
-export const PartyMapRouter: FC = () => {
-  const match = useRouteMatch();
-  useConnectCurrentVenue();
-
-  return (
-    <Switch>
-      <Route path={`${match.path}/:roomTitle`} component={PartyMap} />
```
0xdevalias, line 17: The detail I added on the original comment asking if we could remove this `PartyMapRouter` component said that this route is still required, for nested sub-venues on a PartyMap (eg. jazz bar), and explained some nuances around concerns it would raise if it was added to the main app router for venues/etc incapable of displaying sub-venues.

Were these taken into account before this got removed?

src/hooks/useIsUserVenueOwner.ts
```diff
@@ -0,0 +1,19 @@
...
+import { useSelector } from "./useSelector";
+import { useUser } from "./useUser";
+import { useVenueId } from "./useVenueId";
+
+export const useIsUserVenueOwner = (): boolean => {
+  const { user } = useUser();
+
+  const currentVenueId = useVenueId();
+  const { currentVenue } = useConnectCurrentVenueNG(currentVenueId);
+  const parentVenue = useSelector(parentVenueSelector);
```
0xdevalias, line 12: While we are using the selector here, are we ever 'connecting' it anywhere?

src/hooks/useIsUserVenueOwner.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  if (!currentVenue || !user) return false;
+
+  const venueOwners = parentVenue ? parentVenue.owners : currentVenue.owners;
+
+  return venueOwners.includes(user.uid);
```
0xdevalias, line 18: The logic here should be:

- check `currentVenue.owners.includes(user.uid)`
- if no match, and there is a `currentVenue.parentVenueId`, then load that as `parentVenue` and check `parentVenue.owners.includes(user.uid)`
- if no match, and there is a parentVenue.parentVenueId`, then repeat the above process of checking/loading the parent venue/etc until there is no more `parentVenueId` (at which point you have reached the 'root parent venue' aka 'sovereign venue')
- if there is no match on the 'root parent venue' aka 'sovereign venue', then we can finally consider that there is no match, and that the user doesn't have permission

Obviously we want to ensure we do this in an efficient way, so that it only completes the 're-checks' when required, and properly caches any data otherwise.

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -1,48 +1,47 @@
...
-export const VenueAdminPage: FC = () => {
-  useConnectCurrentVenue();
-
+export const VenueAdminPage: React.FC = () => {
   const { profile, user } = useUser();
-  const venue = useSelector(venueSelector);
-  const venueRequestStatus = useSelector(isCurrentVenueRequestedSelector);
-  const venueRequestingStatus = useSelector(isCurrentVenueRequestingSelector);
+  const venueId = useVenueId();
+  const { currentVenue: venue } = useConnectCurrentVenueNG(venueId);
```
0xdevalias, line 25: Probably would have renamed this variable throughout the code, as `currentVenue` is more descriptive of what it relates to.

src/pages/Admin/Venue/VenueAdminPage.tsx
0xdevalias: Makes sense

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -79,7 +80,10 @@ const TemplateWrapper: React.FC<Props> = ({ venue }) => {
   }
 
   return (
-    <WithNavigationBar fullscreen={fullscreen}>{template}</WithNavigationBar>
+    <WithNavigationBar fullscreen={fullscreen}>
+      <AnnouncementMessage message={venue?.bannerMessage} />
```
0xdevalias, line 84: Looking at the props, I don't think` venue` can ever be `undefined` here?

src/utils/selectors.ts
```diff
@@ -114,6 +114,14 @@ export const shouldRetainAttendanceSelector: SparkleSelector<boolean> = (
   state
 ) => state.attendance.retainAttendance;
 
+export const isCurrentVenueNGRequestedSelector: SparkleSelector<boolean> = makeIsRequestedSelector(
+  "currentVenueNG"
+);
+
+export const isCurrentVenueNGRequestingSelector: SparkleSelector<boolean> = makeIsRequestingSelector(
+  "currentVenueNG"
+);
```
0xdevalias, line 123: It would probably make sense to include return these from the `useConnectCurrentVenueNG` hook, rather than needing to use them individually/separately external to it.

src/components/organisms/BannerAdmin/BannerAdmin.tsx
```diff
@@ -2,14 +2,15 @@ import React, { useState, useCallback, useRef } from "react";
...
 import { makeUpdateBanner } from "api/bannerAdmin";
 
-import { currentVenueSelector } from "utils/selectors";
+import { AnyVenue } from "types/venues";
 
-import { useSelector } from "hooks/useSelector";
+interface BannerAdminProps {
+  venueId?: string;
+  venue: AnyVenue;
+}
```
0xdevalias, line 10: Since we're passing in `venue`, is there any need for `venueId`? If the intent here was that `AnyVenue` doesn't have a `venueId`, then we should make it `WithId<AnyVenue>`

src/components/organisms/BannerAdmin/BannerAdmin.tsx
```diff
@@ -2,14 +2,15 @@ import React, { useState, useCallback, useRef } from "react";
...
 
-export const BannerAdmin: React.FC = () => {
-  const currentVenue = useSelector(currentVenueSelector);
-  const venueId = currentVenue?.id;
-  const existingBannerMessage = currentVenue?.bannerMessage ?? "";
+export const BannerAdmin: React.FC<BannerAdminProps> = ({ venueId, venue }) => {
```
0xdevalias, line 12: We can likely genericise this so that it can be used for both `BannerAdmin` and `src/components/molecules/AdminVideo/AdminVideo.tsx` as they are basically the same thing.

This would probably involve:

- removing the `venue`/`venueId` props
- adding a `label` prop that would be used in place of the current "Banner Message" text in the `<label>`
- adding a `defaultValue` prop that would be used in place of the current `existingBannerMessage`
- adding an `onSubmit` prop that receives the input field data (`string`)
- genericizing the "Clear Banner"/"Save Banner" labels to something like "Clear"/"Save" or similar (potentially as props with defaults set if we want to make this more customizable)

0xdevalias (commented):
@denisdimitrov Retroactive review + future cleanup notes

src/components/templates/PartyMap/PartyMapRouter.tsx
0xdevalias: see https://github.com/sparkletown/sparkle/pull/1105#discussion_r565938578

src/components/templates/Jazzbar/JazzbarRouter/JazzbarRouter.tsx
```diff
@@ -1,20 +0,0 @@
...
-import { useRouteMatch, Switch, Route } from "react-router-dom";
-import JazzBar from "../Jazzbar";
-import ReactionPage from "pages/ReactionPage";
-
-export const JazzbarRouter: React.FC = () => {
-  const match = useRouteMatch();
-
-  return (
-    <Switch>
-      <Route path={`${match.path}/band`} component={ReactionPage} />
```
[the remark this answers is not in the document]
0xdevalias: see https://github.com/sparkletown/sparkle/pull/1105#discussion_r565905822 for my original comments on why we may not be able to remove this

PR #1106, opened 2021-01-27
remove userlist component from jazzbar template

denisdimitrov:
Removes the UserList component from the Jazzbar template only.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/235

### DIFF
--- a/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
+++ b/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
@@ -24,13 +24,11 @@ import CallOutMessageForm from "components/molecules/CallOutMessageForm/CallOutM
 import TableComponent from "components/molecules/TableComponent";
 import TableHeader from "components/molecules/TableHeader";
 import TablesUserList from "components/molecules/TablesUserList";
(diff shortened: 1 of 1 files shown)


0xdevalias (approved):
Tentatively approved, pending a satisfactory answer to the question raised in https://github.com/sparkletown/internal-sparkle-issues/issues/235#issuecomment-768685324

> Reason for this change: Essentially it's superfluous information, since you can already see who is there based on who is sitting at the tables/etc.
>
> Though this does raise a question for me.. what happens if a user is in the jazz bar but not sitting at a table? Will they be shown anywhere then? And if not, how are we going to convey this information if we remove `UserList` from here?

denisdimitrov:
@0xdevalias I left a comment in the issue as well. But since it's not an actual answer to the question you brought up, I suggest waiting for the daily before merging this.

0xdevalias:
https://github.com/sparkletown/internal-sparkle-issues/issues/235#issuecomment-769648581

> In our standup today, @sparkleed said that it's fine for us to not delete this for now, since we will be looking at redesigning this code soon anyway.

PR #1107, opened 2021-01-27
fix admin v2 entrance editing and redirect

denisdimitrov:
- Admin v2 redirection wasn't working when on save
- Fix bug where map background was being deleted after save
- Fix typo
- Pass prop to open inputs when already populated

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/237


functions/venue.js
```diff
@@ -648,8 +648,6 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
 
   if (data.mapBackgroundImageUrl) {
     updated.mapBackgroundImageUrl = data.mapBackgroundImageUrl;
-  } else {
-    updated.mapBackgroundImageUrl = "";
   }
```
0xdevalias, line 651: Do we have any features currently that want to be able to delete the `mapBackgroundImageUrl`? Or will it always just be replaced with a new url; never deleted/set back to default?

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
```diff
@@ -128,6 +128,7 @@ const EntranceExperience: React.FC<EntranceExperienceProps> = ({
               register={register}
               fieldName="entrance"
               showTitle={false}
+              editing={venue.entrance}
```
0xdevalias, line 131: Should this also have an `errors` key set like the other examples here?

0xdevalias (approved):
[CRO] Tentatively approved, assuming satisfactory answers to the questions raised in https://github.com/sparkletown/sparkle/pull/1107#pullrequestreview-577982976

functions/venue.js
denisdimitrov: We do not have any features that delete the mapBackgroundImageUrl. This update functions updates other venue information and should always persist the background image.

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
denisdimitrov: Good point. I'll add those too.

functions/venue.js
0xdevalias: Awesome, thanks for the context.

mike-lvov (approved):
LGTM

denisdimitrov:
@sparkletown/engineering The CI fails saying that prettier needs to be run. But when I run it locally it says this: `All matched files use Prettier code style!` Any ideas what the problem might be?

0xdevalias:
> The CI fails saying that prettier needs to be run. But when I run it locally it says this: `All matched files use Prettier code style!` Any ideas what the problem might be?

@denisdimitrov Have you ‘npm install’d on this branch since prettier got updated?

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
```diff
@@ -84,7 +87,9 @@ const EntranceExperience: React.FC<EntranceExperienceProps> = ({
     },
     [onSave, user, venue.name]
   );
+  console.log(venue.entrance);
 
+  console.log(getValues());
```
0xdevalias, line 92: These should be removed

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
```diff
@@ -128,13 +133,13 @@ const EntranceExperience: React.FC<EntranceExperienceProps> = ({
...
+              editing={venue.entrance}
+              errors={errors.entrance}
             />
           </S.ItemBody>
         </S.ItemWrapper>
 
-        <Button type="submit" disabled={!dirty}>
-          Save
-        </Button>
+        <Button type="submit">Save</Button>
```
0xdevalias, line 142: Why did we remove the `disabled` from this?

src/pages/Admin/Venue/EntranceInput.tsx
```diff
@@ -101,6 +114,9 @@ const EntranceInput: React.FC<EntranceInputProps> = ({
           />
           <Form.Label>Video URL</Form.Label>
           <Form.Control ref={register} name={videoUrl} custom />
+          {errors && errors[index].videoUrl && (
```
0xdevalias, line 117:
```suggestion
          {errors?.[index]?.videoUrl && (
```

0xdevalias (changes requested):

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
denisdimitrov: The fields have validation already. And the `dirty` state is not changing to true, after calling `clear` or `remove` from `useForm`

denisdimitrov:
@0xdevalias Comments have been fixed.

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
0xdevalias: Makes sense.

I'm not sure that our other forms have the 'disable the submit button' sort of logic on them anyway do they? If there is a pattern that we already have that handles this well that you can copy from, that would be cool. But if this was a one off/different from our standards, then no need to try and implement it.

0xdevalias:
@denisdimitrov Do you know why these commits are showing up as part of this PR? It's making the code diff hard to review since it's including unrelated things.

image

0xdevalias (approved):

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
denisdimitrov: Yeah, this disable pattern is new and should be removed imo, just to keep it consistent with the other ux flow.

denisdimitrov:
> Do you know why these commits are showing up as part of this PR? It's making the code diff hard to review since it's including unrelated things.

Yes, I accidentally rebased with staging instead of this PR. My bad.

src/pages/Admin/EntranceExperience/EntranceExperience.tsx
0xdevalias: Sounds good to me

0xdevalias:
> Yes, I accidentally rebased with staging instead of this PR. My bad.

@denisdimitrov Cool, makes sense. Just re-skimmed through the diff now that you've fixed it and it all looks good 👍🏻

PR #1108, opened 2021-01-27
Reworks user counting writes

mike-lvov:
Reworks lastSeen user's data writes


mike-lvov:
Don't bother yourselves on leaving comments, this is a draft. Will make it `ready-for-review`, when ready

src/utils/userLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+  userId,
+  venueName,
+  venueId,
+}: EnterVenueRoomProps) => {
+  trackLocationEntered({
+    userId,
+    locationName: venueName,
+  });
+  openUrl(venueInsideUrl(venueId));
+};
```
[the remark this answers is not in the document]
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/245

src/utils/userLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+type EnterExternalRoomProps = {
+  userId: string;
+  room: AnyRoom;
+};
+
+export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
+  trackLocationEntered({
+    userId,
+    locationName: `external/${room.title}`,
+  });
```
0xdevalias, line 79: This `external/` seems like a new concept being introduced here. Would be good to know what your thinking is around it/how it will work/be used/etc.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+) => {
+  const shouldUseInterval = user && roomName;
+
+  useInterval(
+    () => {
+      // Time spent is currently counted multiple time if multiple tabs are open
+      if (!user || !roomName) return;
+
+      const firestore = firebase.firestore();
+      const doc = `users/${user.uid}/visits/${roomName}`;
```
0xdevalias, line 106: Using potentially user controllable values (`roomName`) in a firebase 'document path' like this can likely lead to the equivalent of SQL injection.

We would probably be better off avoiding this 'path construction' method, and using the slightly more verbose code way of doing `firestore.collection('users').doc(user.id).collection('visits').doc(roomName)`

src/utils/userLocation.ts
```diff
@@ -0,0 +1,120 @@
...
+      const increment = firebase.firestore.FieldValue.increment(
+        LOCATION_INCREMENT_SECONDS
+      );
+
+      return firestore
+        .doc(doc)
+        .update({ timeSpent: increment })
+        .catch(() => {
+          firestore.doc(doc).set({ timeSpent: LOCATION_INCREMENT_SECONDS });
+        });
```
0xdevalias, line 116: We want to avoid this 'try to update, then catch the error and try to set it instead' pattern (see https://github.com/sparkletown/internal-sparkle-issues/issues/125)

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -195,25 +143,22 @@ const VenuePage: React.FC = () => {
     )
       return;
     if (!venue || !user) return;
-    updateProfileEnteredVenueIds(
-      profile?.enteredVenueIds,
-      user?.uid,
-      venue?.id
-    );
+    updateProfileEnteredVenueIds(profile?.enteredVenueIds, userId, venue?.id);
```
0xdevalias, line 146: `updateProfileEnteredVenueIds` is defined in  `src/utils/profile.ts`

This function could be refactored to not require `enteredVenueIds` to be passed into it, and to instead use our 'best practice' firestore update patterns:

- https://firebase.google.com/docs/firestore/manage-data/add-data#update_fields_in_nested_objects
  - `admin.firestore.FieldValue.arrayUnion`

From there, this `useEffect`'s deps can be simplified down to only being the `userId` and `venueId` (rather than the whole objects), which will cause the code to only ever run when the exact things we care about have changed.

This will simplify the code, make it more robust, and make it more performant.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,77 +93,47 @@ const VenuePage: React.FC = () => {
...
+      if (document.visibilityState === "visible") {
+        updateLocationData(userId, {
+          [venueName]: getCurrentTimeInMilliseconds(),
+        });
       }
-    };
-    window.addEventListener("beforeunload", leaveRoomBeforeUnload, false);
-    return () => {
-      window.removeEventListener("beforeunload", leaveRoomBeforeUnload, false);
-    };
```
0xdevalias, line 187: I wrote a bunch of deepdive notes on this topic (as there seems to be a lot of nuance, and a lot of potentially useful things we can use to do this well), captured in https://github.com/sparkletown/internal-sparkle-issues/issues/99

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,77 +93,47 @@ const VenuePage: React.FC = () => {
...
-    profile,
-    retainAttendance,
-    unmounted,
-    user,
-  ]);
+    if (!userId || !venueName) return;
+
+    updateLocationData(userId, {
+      [venueName]: getCurrentTimeInMilliseconds(),
+    });
```
0xdevalias, line 119: It would be good to have a function that removes this 'implementation detail' of having to lay out the location data structure, and just be callable as something like: `updateUserLocation(venueName)`, `updateUserLocation(venueName, roomName)`, etc

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -82,8 +75,8 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
         onMouseOver={triggerAttendance}
         onMouseOut={clearAttendance}
         className="btn btn-primary room-entry-button"
-        onClick={joinRoom}
-        id={`enter-room-in-ongoing-event-card-${room.title}`}
+        onClick={onRoomEnter}
+        id={`enter-room-in-ongoing-event-card-${roomTitle}`}
```
0xdevalias, line 79: In react, we pretty much never have a need for using `id`, so this can probably be removed altogether.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -29,41 +29,39 @@ interface RoomModalProps {
...
 
     return venues?.find((venue) => room.url.endsWith(`/${venue.id}`));
   }, [room, venues]);
 
-  const venueName = roomVenue?.name;
+  // const getRoomTitle = (room, venues) => {
+  //   const roomVenue = venues?.find((venue) => room.url.endsWith(`/${venue.id}`));
+  // }
 
-  // TODO: @debt refactor this to use openRoomWithCounting
```
0xdevalias, line 50: This is probably still a relevant note. When I implemented the refactor that cleaned up/standardised most of the usages to a single helper function in `src/utils/useLocationUpdateEffect.ts` there were a few places that diverged enough from the 'standard pattern' that I didn't have time to figure out how to make it work, which I labelled with this `@debt` sort of thing.

The intent was that in a future cleanup, we would centralise the remaining logic into a helper function/hook in the same sort of way, and none of this 'custom' stuff would live in any of the individual components.

I know this a WIP PR, but I just wanted to capture my intent/plan around this, so that you can make sure to fit that into the changes you're making here.

0xdevalias (commented):
Quick/light comments, mostly to call out patterns/capture intent of things that seem to be getting changed here, so that we can communicate early/often to ensure alignment, rather than waiting for the final PR review to find out something doesn't align with how we were wanting to do it.

src/utils/userLocation.ts
0xdevalias: Also backlogged https://github.com/sparkletown/internal-sparkle-issues/issues/248 to cleanup this potential security issue across the entire codebase.

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -53,11 +48,26 @@ export const EventDisplay: React.FC<EventDisplayProps> = ({ event, venue }) => {
           {event.description}
         </div>
         <div className="schedule-event-info-room">
-          <div onClick={enterEvent}>
-            {event.room ?? "Enter"} {venue && `- ${venue.name}`}
-          </div>
+          {event.room && room ? (
+            <EnterRoomButton room={room as Room}>{buttonText}</EnterRoomButton>
```
mike-lvov, line 52: I don't feel that this is the best way to solve this. I think it is better than before, but can be improved. I'm opened to suggestions

src/components/molecules/LiveSchedule/LiveEvent.tsx
```diff
@@ -43,11 +39,26 @@ export const LiveEvent: FC<LiveEventProps> = ({ venue, event }) => {
           {event.description}
         </div>
         <div className="schedule-event-info-room">
-          <div onClick={enterLiveEvent}>
-            {event.room ?? "Enter"} {venue && `- ${venue.name}`}
-          </div>
+          {event.room && room ? (
+            <EnterRoomButton room={room as Room}>{buttonText}</EnterRoomButton>
```
mike-lvov, line 43: I don't feel that this is the best way to solve this. I think it is better than before, but can be improved. I'm opened to suggestions

src/components/templates/AvatarGrid/AvatarGrid.tsx
```diff
@@ -13,7 +13,7 @@ import { useVenueId } from "hooks/useVenueId";
 import { useRecentVenueUsers } from "hooks/users";
 
 import { WithId } from "utils/id";
-import { openRoomWithCounting } from "utils/useLocationUpdateEffect";
+// import { openRoomWithCounting } from "utils/userLocation";
```
mike-lvov, line 16: legacy

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -24,53 +21,21 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
-      userLastSeenIn
-    );
-
-    if (roomVenue) {
-      openUrl(venueInsideUrl(roomVenue.id));
-      return;
-    }
-    openRoomUrl(room.url);
-  }, [userLastSeenIn, room, roomTitle, user, venueName, roomVenue]);
+  useMousetrap({
```
mike-lvov, line 33: Moved from useKeyboard hook

src/hooks/useKeyboardControls.ts
```diff
@@ -31,28 +30,10 @@ export const useKeyboardControls: ReactHook<UseKeyboardControlsProps, void> = ({
...
-   * enter
-   */
-  const enter = useCallback(() => {
-    if (!row || !column) return;
-
-    // TODO: implement: openRoomUrl(selectedRoom.url) ? can we do it in enterSelectedRoom?
-    enterSelectedRoom();
-  }, [row, column, enterSelectedRoom]);
-
-  useMousetrap({
```
mike-lvov, line 49: moved to RoomModal

src/hooks/users.ts
```diff
@@ -132,23 +132,3 @@ export const useRecentVenueUsers = () => {
     isRecentVenueUsersLoaded: isRecentLocationUsersLoaded,
   };
 };
-
-export const useRecentRoomUsers = (roomTitle?: string) => {
```
mike-lvov, line 136: moved to useRoom

src/hooks/useRoom.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Room } from "types/rooms";
+
+import { enterVenue, enterExternalRoom } from "utils/userLocation";
+import { orderedVenuesSelector } from "utils/selectors";
+
+import { useSelector } from "hooks/useSelector";
+import { useRecentLocationUsers } from "hooks/users";
+import { useUser } from "hooks/useUser";
+
+export const useRoom = (room: Room) => {
```
mike-lvov, line 12: Probably there is a better name for this

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -83,13 +83,15 @@ export const Map: React.FC<MapProps> = ({
 
   const takeSeat = useCallback(
     (row: number | null, column: number | null) => {
+      if (!userUid) return;
       makeUpdateUserGridLocation({
         venueId,
-        userUid: user?.uid,
+        userUid: userUid,
```
denisdimitrov, line 89: You can just do `userUid,`

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -148,6 +150,7 @@ export const Map: React.FC<MapProps> = ({
 
   const onSeatClick = useCallback(
     (row: number, column: number, seatedPartygoer?: WithId<User>) => {
+      console.log("on seat click", row, column, seatedPartygoer);
```
denisdimitrov, line 153: .

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +90,47 @@ const VenuePage: React.FC = () => {
...
-    window.addEventListener("beforeunload", leaveRoomBeforeUnload, false);
-    return () => {
-      window.removeEventListener("beforeunload", leaveRoomBeforeUnload, false);
-    };
-  }, [prevLocations, retainAttendance, user, venueName]);
+    if (!userId) return;
+
+    // NOTE: Clear user location on page close
+    window.addEventListener("beforeunload", () =>
+      updateLocationData(userId, {})
```
denisdimitrov, line 120: If I have 2 pages open and close one of them, isn't this going to clear both locations?

denisdimitrov: Actually looks like, you are keeping only 1 location open. Have you talked with Ed about this, because the very first implementation was like this and we had to change it to support multiple venues.

src/utils/useLocationUpdateEffect.ts
```diff
@@ -1,188 +0,0 @@
-import firebase, { UserInfo } from "firebase/app";
```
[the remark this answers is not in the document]
mike-lvov: Yes

src/pages/VenuePage/VenuePage.tsx
mike-lvov: Agreed. This is an edge case. But I think this implementation is better than the one we have now

denisdimitrov (approved):
LGTM!

src/pages/VenuePage/VenuePage.tsx
mike-lvov: I asked @sparkleed about this and he said it is OK

src/components/templates/AvatarGrid/AvatarGrid.tsx
0xdevalias: What do you mean by 'legacy' here?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Why was this moved from the `useKeyboard hook` to here?

src/hooks/useKeyboardControls.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1108/files/36acbe1b56bfcb0eaeb2dac543748428e999a1be#r567550279, why was this moved out of here to `RoomModal`?

src/pages/VenuePage/VenuePage.tsx
0xdevalias: The 'implementation detail' of clearing this data by updating it with an empty object doesn't belong in individual components. There should be a function in the same file as `updateLocationData` called something like `clearLocationData` that handles this.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
```
0xdevalias, line 128: You add the event listener, but you never clean it up again. This is a regression based on the code that was already here, and should be restored.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
+    );
+  }, [userId]);
 
   useEffect(() => {
     if (
       profile?.enteredVenueIds &&
```
0xdevalias, line 134: We can get rid of this line and just use `profile?.enteredVenueIds?.includes(venueId)` below

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
-    if (!venue || !user) return;
-    updateProfileEnteredVenueIds(
-      profile?.enteredVenueIds,
-      user?.uid,
-      venue?.id
-    );
-  }, [profile, user, venue]);
+    if (!venueId || !user || !profile) return;
+
+    updateProfileEnteredVenueIds(profile?.enteredVenueIds, user?.uid, venueId);
```
0xdevalias, line 141: This should be refactored to use proper firestore update/arrayUnion/etc logic, instead of needing to pass through the old 'enteredVenueIds' and overwriting the entire field in the database.

At that point, depending on whether we already have the `enteredVenueIds` in the frontend, we wouldn't even need to check if the current `venueId` is already there, as we could just always 'union' it, and then ensure that this function is only called when the `userId`/`venueId` changes.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
+  }, [profile, user, venueId]);
+
+  // NOTE: User's timespent updates
+
+  useUpdateTimespentPeriodically(user, venueName);
```
0xdevalias, line 146: This should probably be combined with 'location tracking' and maybe even the `enteredVenueIds` updates into a single 'useTrackUserLocation` or similarly named hook.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -211,10 +154,10 @@ const VenuePage: React.FC = () => {
   // @debt refactor this + related code so as not to rely on using a shadowed 'storeAs' key
   //   this should be something like `storeAs: "currentUserPrivateChats"` or similar
   useFirestoreConnect(
-    user?.uid
+    userId
       ? {
           collection: "privatechats",
```
0xdevalias, line 159: This should be refactored into a `usePrivateChats` or similar hook.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -32,59 +27,20 @@ export const PartyMap: React.FC = () => {
...
   const selectRoom = useCallback((room: Room) => {
     setSelectedRoom(room);
-    setRoomModalOpen(true);
   }, []);
 
-  // Note: we explicitly don't unset selectedRoom here as that would cause RoomModal to close abruptly
   const unselectRoom = useCallback(() => {
-    setRoomModalOpen(false);
+    setSelectedRoom(undefined);
   }, []);
```
0xdevalias, line 39: Why was this changed? It's a regression and should be reverted. (see other comments in this review for further explanation)

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -111,14 +67,9 @@ export const PartyMap: React.FC = () => {
...
         unselectRoom={unselectRoom}
-        enterSelectedRoom={enterSelectedRoom}
       />
 
-      <RoomModal
-        show={isRoomModalOpen}
-        room={selectedRoom}
-        onHide={unselectRoom}
-      />
+      {selectedRoom && <RoomModal room={selectedRoom} onHide={unselectRoom} />}
```
0xdevalias, line 72: Why was this changed? We explicitly want the `show` prop, and we never want to 'unrender' the `RoomModal` component. This is a regression and should be reverted. (see other comments in this review for further explanation)

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -32,59 +27,20 @@ export const PartyMap: React.FC = () => {
...
-  );
-
-  // Note: since we explicitly don't unset selectedRoom in unselectRoom, we need to check if
-  //   the RoomModal is open to know if we have a room selected
-  const enterSelectedRoom = useCallback(() => {
-    if (!selectedRoom || !isRoomModalOpen) return;
-
-    enterRoom(selectedRoom);
-  }, [enterRoom, isRoomModalOpen, selectedRoom]);
-
```
0xdevalias, line 85: Why was this removed/where was it refactored to?

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -32,59 +27,20 @@ export const PartyMap: React.FC = () => {
...
   // Find current room from url
   const { roomTitle } = useParams();
+
   const currentRoom = useMemo(() => {
```
0xdevalias, line 44: This seems similar if not exactly the same as the code in `useRoom`, so this can probably be refactored to use that.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+export const trackLocationEntered = ({
+  userId,
+  locationName,
+}: TrackLocationProps) => {
+  updateLocationData(userId, {
+    [locationName]: getCurrentTimeInMilliseconds(),
+  });
+};
+
+export const enterVenue = (venueId: string) => openUrl(venueInsideUrl(venueId));
```
0xdevalias, line 41: Should this have `trackLocationEntered` code within it as well? If not, this function feels like it would be better placed in `utils/url`

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+type EnterExternalRoomProps = {
+  userId: string;
+  room: AnyRoom;
+};
+
+export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
+  trackLocationEntered({
+    userId,
+    locationName: room.url,
```
0xdevalias, line 51: Why `room.url`? I don't think this is correct? Or at the very least I believe, this is a change in logic from what we were previously doing.

It should be using the `venueName/roomName` slug.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  openRoomUrl(room.url);
+};
+
+export const useUpdateTimespentPeriodically = (
+  user: UserInfo | undefined,
```
0xdevalias, line 58: this should use `user?: UserInfo`

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+  roomName: string
+) => {
+  const shouldUseInterval = user && roomName;
+
+  useInterval(
+    () => {
+      // Time spent is currently counted multiple time if multiple tabs are open
```
0xdevalias, line 65: This should probably have `@debt` on it

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      const firestore = firebase.firestore();
+      const doc = `users/${user.uid}/visits/${roomName}`;
+      const increment = firebase.firestore.FieldValue.increment(
+        LOCATION_INCREMENT_SECONDS
+      );
+
+      return firestore
+        .doc(doc)
+        .update({ timeSpent: increment })
```
0xdevalias, line 76: Since this field is only used once, it seems like it would be cleaner if it was just included directly in the `.update()` rather than split out as it's own `const`

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+        .catch(() => {
+          firestore.doc(doc).set({ timeSpent: LOCATION_INCREMENT_SECONDS });
```
0xdevalias, line 78: It isn't immediately obvious why we are doing this, so perhaps it is worth adding a comment to explain that when this block is run, it's (intended to be) because there was no document to update, and so we are defaulting it to `LOCATION_INCREMENT_SECONDS` since that's how often this code runs.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+import { openRoomUrl, openUrl, venueInsideUrl } from "./url";
+
+const LOCATION_INCREMENT_SECONDS = 10;
+const LOCATION_INCREMENT_MS = LOCATION_INCREMENT_SECONDS * 1000;
+
+type LocationData = Record<string, number>;
+
+export const updateLocationData = (
+  userId: string,
+  newLocationData: LocationData = {}
```
0xdevalias, line 18: Since this prop isn't optional, there is no reason to be setting a default on it here.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
+import firebase, { UserInfo } from "firebase/app";
+
+import { AnyRoom } from "types/rooms";
+
+import { updateUserProfile } from "pages/Account/helpers";
```
0xdevalias, line 5: This shouldn't be importing from `pages/`, that function (or seemingly everything in that file) should be refactored into `api/`.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+type LocationData = Record<string, number>;
+
+export const updateLocationData = (
+  userId: string,
+  newLocationData: LocationData = {}
+) => {
+  updateUserProfile(userId, {
+    lastSeenAt: getCurrentTimeInMilliseconds(),
+    lastSeenIn: newLocationData,
```
0xdevalias, line 22: This changes the logic from how this function used to work. Previously it had a bunch of logic to ensure it didn't remove data from this field that it shouldn't be updating, whereas now we are just wiping out the field entirely each time. While this may be the desired effect, it would be good to know your thought process, and why this way is better than what was there before.

src/utils/userLocation.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const trackLocationEntered = ({
+  userId,
+  locationName,
+}: TrackLocationProps) => {
+  updateLocationData(userId, {
+    [locationName]: getCurrentTimeInMilliseconds(),
+  });
+};
```
0xdevalias, line 39: Previously this function (while it was named `trackRoomEntered`) allowed an object of locations to be supplied, which I believe was normally used to track a user as being both in the venue, and the associated room. The way the code has changed here seems to not allow that anymore. What was your reasoning behind changing it like this?

src/utils/useLocationUpdateEffect.ts
```diff
@@ -1,188 +0,0 @@
...
-  lastSeenIn,
-}: TrackRoomEnteredNGProps) => {
-  if (!user) return;
-
-  trackRoomEntered(
-    user,
-    { [`${venue.name}/${room.title}`]: getCurrentTimeInMilliseconds() },
-    lastSeenIn
-  );
-};
```
0xdevalias, line 78: Why was this removed?

src/utils/useLocationUpdateEffect.ts
```diff
@@ -1,188 +0,0 @@
...
-  lastSeenIn,
-}: TrackVenueEnteredProps) => {
-  if (!user) return;
-
-  trackRoomEntered(
-    user,
-    { [venue.name]: getCurrentTimeInMilliseconds() },
-    lastSeenIn
-  );
-};
```
0xdevalias, line 98: Why was this removed?

src/utils/useLocationUpdateEffect.ts
```diff
@@ -1,188 +0,0 @@
...
-
-export const leaveRoom = (user: UserInfo) => {
-  updateUserProfile(user.uid, {
-    lastSeenAt: 0,
-    lastSeenIn: {},
-    room: null,
-  });
-};
```
0xdevalias, line 106: Why was this removed?

src/utils/useLocationUpdateEffect.ts
```diff
@@ -1,188 +0,0 @@
...
-  // Track room counting
-  trackRoomEnteredNG({
-    user,
-    venue,
-    room,
-    lastSeenIn: profile?.lastSeenIn,
-  });
-
-  openRoomUrl(room.url);
-};
```
0xdevalias, line 144: Why was this removed? The seemed equivalent function in the new code `enterVenue` doesn't seem to do any tracking. This feels like a regression.

src/components/molecules/EventDisplay/EventDisplay.tsx
mike-lvov: `as` assignment was used because we still have `Camp room` type or similar. Also, the way I separated this logic into a separate component doesn't seem to be perfect

0xdevalias: It might be worth adding an `@debt` comment or similar in these cases, that briefly explains what it is/why it was done that way/how/when it can be fixed up, eg.

```
// @debt using 'as' here because we still have multiple legacy Room types, clean up once we remove those
```

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -78,32 +71,11 @@ const VenuePage: React.FC = () => {
...
-      lastSeenIn: updatedLastSeenIn,
-      lastSeenAt: getCurrentTimeInMilliseconds(),
-      room: venueName,
-    });
-  }, [prevLocations, user, venueName]);
-
-  useInterval(() => {
-    updateUserLocation();
-  }, LOC_UPDATE_FREQ_MS);
-
```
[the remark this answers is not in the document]
mike-lvov: The changes that you selected are not related to `timespent`

`useUpdateTimespentPeriodically ` is not really related to user's counting. I would rather have a separate issue for that. Since, it didn't seem to work properly before this PR

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
+    trackLocationEntered({ userId, locationName: venueName });
+  }, [userId, venueName]);
 
-  const newLocations = {
-    ...prevLocations,
-    ...newLocation,
-  };
+  useInterval(() => {
+    updateUserLocationToCurrentVenue();
+  }, LOC_UPDATE_FREQ_MS);
```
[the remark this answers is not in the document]
mike-lvov: `useUpdateTimespentPeriodically(user, venueName)` is a completely separate entity. It updates `visits` collection in user's object

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -115,92 +87,63 @@ const VenuePage: React.FC = () => {
...
-    newLocation,
-    newLocations,
-    prevLocations,
-    profile,
-    retainAttendance,
-    unmounted,
-    user,
-  ]);
+    updateUserLocationToCurrentVenue();
+  }, [updateUserLocationToCurrentVenue]);
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/1108#discussion_r567806211

src/pages/VenuePage/VenuePage.tsx
mike-lvov: We have an issue for that. It's in progress and is related to chats https://github.com/sparkletown/internal-sparkle-issues/issues/109

src/components/templates/PartyMap/PartyMap.tsx
mike-lvov: I don't think we need `isRoomModalOpen` entity since it is a dupe of `selectRoom`

src/components/templates/PartyMap/PartyMap.tsx
mike-lvov: refactored into `useRoom`. Was passed to `Map` as far as I remember

src/components/templates/PartyMap/PartyMap.tsx
mike-lvov: How is it the same?

src/utils/userLocation.ts
mike-lvov: Yes, it is a change of logic. I don't understand the reasoning behind using `venueName/roomName` slug tbh. This slug amplifies the entry point rather than the location itself, which is not what we want, I guess

src/utils/userLocation.ts
mike-lvov: Fixed small things regarding this chunk of code, but I'd call it OOS for this PR since this is not related to counting

src/utils/userLocation.ts
mike-lvov: [OOS]

src/utils/userLocation.ts
mike-lvov: The reasoning behind this approach is that a person can be only in one location at any point in time. I was also thinking about removing `lastSeenAt` field, since `lastSeenIn`'s value is a dupe of `lastSeenAt` value. WDYT of it?

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1108#discussion_r567817241

src/utils/useLocationUpdateEffect.ts
mike-lvov: You don't need tracking for entering the venue inside the platform. We have `VenuePage` that has us covered there

src/utils/useLocationUpdateEffect.ts
mike-lvov: I don't know how to answer questions like this. It was not used anymore\the alternatives are used instead.

src/utils/useLocationUpdateEffect.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1108#discussion_r567819017

src/utils/useLocationUpdateEffect.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1108#discussion_r567819017

src/components/templates/PartyMap/PartyMap.tsx
mike-lvov: My initial reasoning was that I wanted to have `room` to be defined there. But I got your point. Will do it in another way

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
0xdevalias: Hrmm, fair enough. Makes sense. So long as we don't lose this.

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
0xdevalias: Makes sense. I guess the fact that all of these are updating things to do with being in venues/etc, how long we've been there, etc; it all feels somewhat related (even though the underlying collections aren't necessarily). So my thought/desire here was that we would just have a single hook 'track all the things about my location and how long i've been here' that wraps all of this up.

src/pages/VenuePage/VenuePage.tsx
0xdevalias: Awesome. Sounds good.

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: It's not a dupe (as explained elsewhere in the comments, though let me know if you still don't understand the nuance of why). We need to be able to hide/unhide the modal without unsetting the venue it has, to avoid 'unrendering' it, and to allow our 'smooth css transitions' to run.

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Makes sense.

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Looking at the full amount of it, perhaps it isn't in its entirety. I believe the part I was referring to was finding the room (which again, may not be the same, but while skimming and reviewing the code, it seemed similar):

```
currentVenue.rooms.find(
      (venueRoom) =>
        createUrlSafeName(venueRoom.title) === createUrlSafeName(roomTitle)
    );
```

src/utils/userLocation.ts
0xdevalias: One reason is that that is how it has always been, and we haven't designed or discussed changing it. There is nothing in this PR that handles migrating existing data. None of the related scripts/reporting/etc are updated for it. We can't just decide to change a core feature of the platform without raising it with the team, and discussing the potential impacts. And we DEFINITELY should not be changing that haphazardly hidden among a giant PR that doesn't mention anything about changing it. This is the sort of things that I raised in retro yesterday where there needs to be more up front planning/discussion/etc.

src/utils/userLocation.ts
0xdevalias: Since you renamed the file/etc, all I see is an entire file of changed diff, so I comment appropriately.

src/utils/userLocation.ts
0xdevalias: Then as per our discussions around OOS and review comments in the team meeting the other day, please also create the backlog issue for it and link it from here, so that we know that this information isn't going to be lost.

src/utils/userLocation.ts
0xdevalias: I don't believe we currently enforce them being only in a single location though, if they have 2 tabs open, etc. So i'm not sure that "can only be in one location at any point of time" is true right now (even though I know you have thoughts/plans to change that)

---

We can't remove `lastSeenAt` currently because it persists, whereas `lastSeenIn` gets wiped out whenever room changes/they leave the platform/etc.

If you want to remove it as part of a greater change to how `lastSeenIn` works in general (so that it doesn't ever lose that data), I am open to exploring it as part of that work, but I expect it may have a bunch of related code changes that will need to be taken into consideration as part of it.

src/utils/userLocation.ts
0xdevalias: https://github.com/sparkletown/sparkle/pull/1108#discussion_r568321282

src/utils/useLocationUpdateEffect.ts
0xdevalias: Makes sense.

src/utils/useLocationUpdateEffect.ts
0xdevalias: These functions served a purpose that I don't believe is handled by any of the ones we have now. Namely the thing I have called out in a few of your PRs/similar related to this, around encoding the `venueName/roomName` slug and similar from the 'base values (venue, room, etc), so that we aren't forced to construct the 'implementation detail' of the format that counting uses at every single place we want to update it.

You raised concerns around violating 'single purpose' functions at that point, at which I suggested that helper functions that could construct the appropriate 'slugs' would also satisfy my need. Yet I haven't seen that in any of the code I have reviewed around this feature/the changes that have been made to it recently.

These functions satisfied that need (though seemingly weren't consistently used everywhere).

src/utils/useLocationUpdateEffect.ts
0xdevalias: https://github.com/sparkletown/sparkle/pull/1108#discussion_r568322817

src/utils/useLocationUpdateEffect.ts
0xdevalias: https://github.com/sparkletown/sparkle/pull/1108#discussion_r568322817

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Sounds good.

PR #1109, opened 2021-01-28
add reporting dependencies to dependabot + fix reviewers

0xdevalias:
- the `reporting/` folder wasn't being tracked by dependabot, now it is
- we renamed `sparkletown/tech-leads` to `sparkletown/engineering-leads`, but this config was still using the old group name


denisdimitrov (approved):
Looks like a small change. LGTM!

PR #1115, opened 2021-01-28
merge staging into bigtop

denisdimitrov:


0xdevalias:
@denisdimitrov For future note, staging should only ever be merged into master. Then we push master to the other environment branches (like big top) to deploy it. We never open a PR to merge from staging -> otherEnvironment (or from master -> otherEnvironment)

denisdimitrov:
@0xdevalias bigtop is protected branch I can't push directly.

0xdevalias:
@denisdimitrov Interesting... we will probably have to change the branch protections we have on that then. They may not have been updated properly since we added this new environment pattern. Though the interim workaround is just disable the branch protection to make that 'deploy push'.

PR #1117, opened 2021-01-28
Profile fix

mike-lvov:
FIx urgent issue

### DIFF
--- a/src/components/templates/PartyMap/components/Map/Map.tsx
+++ b/src/components/templates/PartyMap/components/Map/Map.tsx
@@ -33,7 +33,7 @@ export const DEFAULT_ROWS = 25;
 
 interface MapProps {
   user: FirebaseReducer.AuthState;
-  profileData: UserExperienceData;
+  profileData?: UserExperienceData;
   venue: PartyMapVenue;
   partygoers: readonly WithId<User>[];
   selectRoom: (room: Room) => void;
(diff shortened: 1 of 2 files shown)


denisdimitrov (approved):
LGTM! Considering that empty object for profileData won't break the venue layout.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -33,7 +33,7 @@ export const DEFAULT_ROWS = 25;
 
 interface MapProps {
   user: FirebaseReducer.AuthState;
-  profileData: UserExperienceData;
+  profileData?: UserExperienceData;
```
0xdevalias, line 36: Looks like while this was a required prop here, the actually usages of it elsewhere in the code were treating it as though it could already be `undefined`.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -43,7 +43,7 @@ interface MapProps {
 
 export const Map: React.FC<MapProps> = ({
   user,
-  profileData,
+  profileData = {},
```
0xdevalias, line 46: Since the rest of the code that uses this already appears to be written to handle it being potentially `undefined`, we could probably avoid defaulting it here.

Alternately, we could change the other usages to always expect it to be defined.

0xdevalias: I think my preference would be to leave it as being potentially `undefined`, and remove the default here (as then we are able to easily check whether they have existing data or not); but i'm open to counter arguments why that wouldn't be the best way.

0xdevalias (commented):
A few retroactive/future cleanup comments.

PR #1122, opened 2021-01-29
fix partymap column validation and field creation

denisdimitrov:
Changes the partymap default columns to 5 to make the room container big enough to fix the screen.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/184


0xdevalias (commented):
This seems like more of a bugfix workaround than a longterm solution, is that right? (not a blocker, just want to make sure it's framed right in my mind)

How will this address existing data in firestore that may already fall outside of these? Do we have a way to fix it up? Or to add a workaround in the code that 'forces' the render to use a minimum of 5 (even if they only have 1 set)?

denisdimitrov:
> How will this address existing data in firestore that may already fall outside of these? Do we have a way to fix it up? Or to add a workaround in the code that 'forces' the render to use a minimum of 5 (even if they only have 1 set)?

@0xdevalias  All we have to do is go to the admin panel and change the number of columns. I don't think it's necessary to force the render to use minimum of 5, if we are catching the minimum that means we need to catch the maximum as well. There is always the human factor where someone can go to the firebase and change it to 100000. Only the venues created by admin v2 have the value of `columns: 1` and they can be easily changed manually. 

The problem is that we already have a check but it checks if the columns exist or not. That's why I'm going to commit a different code where instead of `columns:1` it won't push anything and that check will work as it should.

functions/venue.js
```diff
@@ -163,7 +163,7 @@ const createVenueData_v2 = (data, context) => ({
   },
   owners: [context.auth.token.user_id],
   showGrid: data.showGrid || false,
-  columns: data.columns || 1,
+  ...(showGrid && { columns: data.columns }),
```
mike-lvov, line 166: should we set `rows` as well?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -682,8 +682,8 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
           className="align-left"
           placeholder={`Number of grid columns`}
         />
-        {errors.name ? (
```
mike-lvov, line 685: What if we have name-related error?

functions/venue.js
denisdimitrov: Rows are automatically calculated in the Map.tsx (Party map component)

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: The name related error is already added at the top of this file next to the venue name, it was copy pasted here and it wasn't showing the correct error for the columns.

0xdevalias:
> All we have to do is go to the admin panel and change the number of columns. I don't think it's necessary to force the render to use minimum of 5, if we are catching the minimum that means we need to catch the maximum as well. There is always the human factor where someone can go to the firebase and change it to 100000. Only the venues created by admin v2 have the value of `columns: 1` and they can be easily changed manually.

My request above is in line with avoiding the need to make manual changes to existing data. We need to design/build things in a way that it will still work properly even when the data is from an 'old version'. I've detailed a lot more about this in the data migration patterns/etc backlog item in https://github.com/sparkletown/internal-sparkle-issues/issues/152, but for this specific PR, it seems like the easy solution is to set `Math.min` at the render point as well. If we have a maximum, then we should also set that at the render point. While we're obviously not there at the moment, we don't want there to be "the human factor", our code/platform should handle that (eg. eventually I think we should validate that a venue config matches the expected schema when we read it back out as well, not just when we save it. But that's definitely OOS of this task)

> The problem is that we already have a check but it checks if the columns exist or not. That's why I'm going to commit a different code where instead of `columns:1` it won't push anything and that check will work as it should.

That seems like a decent fix for the core of the issue. But I still feel like we need to handle the 'existing data' case in an automatic, no manual fix required way.

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -86,7 +86,10 @@ export const validationSchema = Yup.object()
     logoImageFile: createFileSchema("logoImageFile", false).notRequired(),
 
     showGrid: Yup.bool().notRequired(),
-    columns: Yup.number().notRequired().min(1).max(100),
+    columns: Yup.number().when("showGrid", {
+      is: true,
+      then: Yup.number().required("At least 5 columns are required").min(5),
+    }),
```
0xdevalias, line 92: Previously there was a maximum set in this validation, yet that doesn't seem to exist here anymore?

0xdevalias:
As additional/alternative approaches/ways of thinking about this, do we know why the styles/etc break when this is set so small/not set? It feels like there is probably a CSS styling element that probably should be involved, and that perhaps setting a `min-width`/`min-height`/similar might be viable?

0xdevalias (changes requested):
As per https://github.com/sparkletown/sparkle/pull/1122#issuecomment-770008040, we need to make sure that code changes like this properly account for data that already exists within the database, either directly, or by way of scripts/etc to backfill data migrations.

denisdimitrov:
> As additional/alternative approaches/ways of thinking about this, do we know why the styles/etc break when this is set so small/not set? It feels like there is probably a CSS styling element that probably should be involved, and that perhaps setting a min-width/min-height/similar might be viable?

This is a regression bug. Short answer, yes. It has to do something with the CSS.
This was working correctly last time I checked. I spent like an hour trying to figure out the problem but I couldn't even with the history changes. The columns will probably work fine for now, but at some point someone might want a map with a 2-3 columns.

denisdimitrov:
@0xdevalias You comment about the min/max handling of this has been taken in account and I've added a fix for this as well. Feel free to give this another review.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -55,7 +55,10 @@ export const Map: React.FC<MapProps> = ({
   const userUid = user?.uid;
   const showGrid = venue.showGrid;
 
-  const totalColumns = venue.columns ?? DEFAULT_COLUMNS;
+  const totalColumns = Math.max(
+    5,
+    Math.min(100, venue.columns ?? DEFAULT_COLUMNS)
+  );
```
0xdevalias, line 61: This took me a while to mentally process the logic, I think the main thing was that the 'maximum value' was 'before' the 'attempted value'. So my mind read it as MIN MAX venueColumns, rather than `MIN venueColumns MAX`, which feels more natural for a 'range'.

We should also extract the numbers out into `const`s, and probably use those same `const`s in the validation as well as here (which sounds like moving them all into `settings.ts` would be a decent place for them to live.

```suggestion
  const totalColumns = Math.max(
    MINIMUM_COLUMNS,
    Math.min(venue.columns ?? DEFAULT_COLUMNS, MAXIMUM_COLUMNS)
  );
```

src/pages/Admin/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -58,7 +58,10 @@ const validationSchema = Yup.object().shape<Venue_v2_AdvancedConfig>({
   showGrid: Yup.boolean().notRequired(),
   columns: Yup.number().when("showGrid", {
     is: true,
-    then: Yup.number().required("Columns are required!").min(1),
+    then: Yup.number()
+      .required("The columns need to be between 5 and 100.")
+      .min(5)
+      .max(100),
```
0xdevalias, line 64: As per other comment, using the shared `const`s for these would be good.

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -86,7 +86,13 @@ export const validationSchema = Yup.object()
...
 
     showGrid: Yup.bool().notRequired(),
-    columns: Yup.number().notRequired().min(1).max(100),
+    columns: Yup.number().when("showGrid", {
+      is: true,
+      then: Yup.number()
+        .required("The columns need to be between 5 and 100.")
+        .min(5)
+        .max(100),
+    }),
```
0xdevalias, line 95: As per other comment, using the shared `const`s for these would be good.

0xdevalias: This raises another question too, we seem to have similar (if not the same) validation rules shared between this file and `src/pages/Admin/AdvancedSettings/AdvancedSettings.tsx`. Are they the same? And if so, can we reduce the duplication of them and use the same definition for both?

0xdevalias (approved):
A few minor changes that would be good to make before this is merged.

Will give this an approval so you're not blocked, but would be good to fix them up before you do.

PR #1123, opened 2021-01-29
Performance improvement

mike-lvov:
Wraps MapCell into a memo

### DIFF
--- a/src/api/profile.ts
+++ b/src/api/profile.ts
@@ -14,22 +14,18 @@ export interface MakeUpdateUserGridLocationProps {
 export const makeUpdateUserGridLocation = ({
   venueId,
   userUid,
-  profileData,
 }: MakeUpdateUserGridLocationProps) => (
   row: number | null,
   column: number | null
 ) => {
-  if (!userUid || !venueId || !profileData) return;
+  if (!userUid || !venueId) return;
 
   const doc = `users/${userUid}`;
 
   const newData = {
-    data: {
-      ...profileData,
-      [venueId]: {
-        row,
-        column,
-      },
+    [`data.${venueId}`]: {
+      row,
+      column,
     },
   };
 
--- a/src/components/molecules/MapCell/MapCell.tsx
+++ b/src/components/molecules/MapCell/MapCell.tsx
@@ -3,7 +3,7 @@ import React, { useCallback } from "react";
 import { User } from "types/User";
 import { WithId } from "utils/id";
 
-interface MapRowProps {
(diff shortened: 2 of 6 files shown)


denisdimitrov (approved):
LGTM!

src/api/profile.ts
```diff
@@ -14,22 +14,18 @@ export interface MakeUpdateUserGridLocationProps {
...
-    data: {
-      ...profileData,
-      [venueId]: {
-        row,
-        column,
-      },
+    [`data.${venueId}`]: {
+      row,
+      column,
     },
```
0xdevalias, line 29: 👌🏻 🎉

src/components/molecules/MapCell/MapCell.tsx
```diff
@@ -53,3 +53,5 @@ export const MapRow: React.FC<MapRowProps> = ({
     </div>
   );
 };
+
+export const MapCell = React.memo(_MapCell);
```
0xdevalias, line 57: @mike-lvov what was the intended purpose of this change? My understanding is that `MapCell` (formerly `MapRow`) is only ever used in `src/components/templates/PartyMap/components/Map/hooks/useMapGrid.tsx`, within a section that is already `useMemo`'d. So it would seem that this change isn't actually going to do anything to improve the performance?

src/api/profile.ts
```diff
@@ -14,22 +14,18 @@ export interface MakeUpdateUserGridLocationProps {
 export const makeUpdateUserGridLocation = ({
   venueId,
   userUid,
-  profileData,
 }: MakeUpdateUserGridLocationProps) => (
```
0xdevalias, line 17: `MakeUpdateUserGridLocationProps` should also have been changed to remove this prop from it.

src/components/templates/PartyMap/components/Map/hooks/useMapGrid.tsx
```diff
@@ -5,7 +5,7 @@ import { ReactHook } from "types/utility";
 
 import { WithId } from "utils/id";
 
-import { MapRow } from "components/molecules/MapRow";
+import { MapCell } from "components/molecules/MapCell";
```
0xdevalias, line 8: 👌🏻 🎉  Nice change. I've been meaning/wanting to do similar for a while.

0xdevalias (commented):
Retroactive review.

@mike-lvov Some questions/comments + things that need to be cleaned up in a future PR.

PR #1129, opened 2021-02-01
Users count PR cleanup

mike-lvov:
This PR addresses cleanup around some of the changes made in https://github.com/sparkletown/sparkle/pull/1108

Or particular note are the review comments in: https://github.com/sparkletown/sparkle/pull/1108#pullrequestreview-580000514


src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -16,21 +17,26 @@ interface EventDisplayProps {
...
+  });
+
   return (
-    <div
-      key={event.name + Math.random().toString()}
-      className={`schedule-event-container ${
-        isLiveEvent && "schedule-event-container_live"
-      }`}
-    >
+    <div key={event.name} className={containerClasses}>
```
0xdevalias, line 39: This shouldn't need the `key` prop at all, as it's not within a loop.

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -16,21 +17,26 @@ interface EventDisplayProps {
...
   const buttonText = `${event.room ?? "Enter"} ${venue && `- ${venue.name}`}`;
 
   const isLiveEvent =
     event.start_utc_seconds < getCurrentTimeInUTCSeconds() &&
     event.start_utc_seconds + event.duration_minutes * 60 >
       getCurrentTimeInUTCSeconds();
 
+  const containerClasses = classNames("schedule-event-container", {
+    "schedule-event-container_live": isLiveEvent,
+  });
```
0xdevalias, line 36: Is this `_live` meant to be a 'variant' (aka modifier)? If so, i've been loosely following BEM in my recent styling cleanups, which would make this `--live`:

- http://getbem.com/naming/

src/components/molecules/LiveSchedule/LiveEvent.tsx
```diff
@@ -1,64 +0,0 @@
-import React from "react";
```
0xdevalias, line 1: What was this file refactored/renamed into, if anything?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,23 +20,17 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
 
 import "./RoomModal.scss";
 
-interface RoomModalProps {
+type RoomModalProps = {
```
0xdevalias, line 23:
```suggestion
export type RoomModalProps = {
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -81,7 +75,6 @@ export const RoomModal: React.FC<RoomModalProps> = ({ onHide, room }) => {
               </div>
               <div className="col">
                 <RoomModalOngoingEvent
-                  roomTitle={room.title}
```
0xdevalias, line 84: Guessing this was just never used/needed?

src/hooks/useKeyboardControls.ts
```diff
@@ -34,6 +35,15 @@ export const useKeyboardControls: ReactHook<UseKeyboardControlsProps, void> = ({
   const { profile } = useUser();
   const { row, column } = profile?.data?.[venueId] ?? {};
 
+  const { enterRoom } = useRoom(room);
```
0xdevalias, line 38: Where is `room` coming from for this?

Not sure if we would necessarily want this hook within this function, do we already use it in the component that calls `useKeyboardControls` or similar? Could we maybe just pass down `enterRoom` through the props or similar?

src/hooks/useRoom.ts
```diff
@@ -13,11 +14,13 @@ export const useRoom = (room: Room) => {
...
+
   const venues = useSelector(orderedVenuesSelector);
 
   const roomVenue = useMemo(
-    () => venues?.find((venue) => room.url.endsWith(`/${venue.id}`)),
-    [room, venues]
+    () => venues?.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
+    [roomUrl, venues]
   );
   const roomId = roomVenue ? roomVenue.name : room.url;
```
0xdevalias, line 25: Since this is used with `useRecentLocationUsers`, I still don't believe `room.url` is correct here. It should be the `venueName/roomName` slug.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -93,57 +94,49 @@ const VenuePage: React.FC = () => {
...
+      !venueId ||
+      !user ||
+      !profile ||
+      profile?.enteredVenueIds?.includes(venueId)
+    ) {
       return;
-    if (!venueId || !user || !profile) return;
+    }
 
     updateProfileEnteredVenueIds(profile?.enteredVenueIds, user?.uid, venueId);
```
0xdevalias, line 134: It seems we already have `userId` defined in this file, so can we narrow the scope of the checks/etc here to use that rather than the full user object?

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
 const LOCATION_INCREMENT_SECONDS = 10;
 const LOCATION_INCREMENT_MS = LOCATION_INCREMENT_SECONDS * 1000;
 
 type LocationData = Record<string, number>;
 
-export const updateLocationData = (
-  userId: string,
-  newLocationData: LocationData = {}
-) => {
+type UpdateLocationDataProps = {
```
0xdevalias, line 16:
```suggestion
export type UpdateLocationDataProps = {
```

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
+  newLocationData,
+}: UpdateLocationDataProps) => {
   updateUserProfile(userId, {
     lastSeenAt: getCurrentTimeInMilliseconds(),
     lastSeenIn: newLocationData,
   });
 };
 
-export interface TrackLocationProps {
+type SetNewLocationDataProps = {
```
0xdevalias, line 31:
```suggestion
export type SetNewLocationDataProps = {
```

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
 import { useInterval } from "hooks/useInterval";
 
 import { getCurrentTimeInMilliseconds } from "./time";
-import { openRoomUrl, openUrl, venueInsideUrl } from "./url";
+import { openRoomUrl } from "./url";
 
 const LOCATION_INCREMENT_SECONDS = 10;
 const LOCATION_INCREMENT_MS = LOCATION_INCREMENT_SECONDS * 1000;
 
 type LocationData = Record<string, number>;
```
0xdevalias, line 14:
```suggestion
export type LocationData = Record<string, number>;
```

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
 
 import { updateUserProfile } from "pages/Account/helpers";
 import { useInterval } from "hooks/useInterval";
 
 import { getCurrentTimeInMilliseconds } from "./time";
-import { openRoomUrl, openUrl, venueInsideUrl } from "./url";
+import { openRoomUrl } from "./url";
 
 const LOCATION_INCREMENT_SECONDS = 10;
 const LOCATION_INCREMENT_MS = LOCATION_INCREMENT_SECONDS * 1000;
```
0xdevalias, line 12: Might make sense for these to be moved to `settings.ts`?

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
     lastSeenAt: getCurrentTimeInMilliseconds(),
     lastSeenIn: newLocationData,
   });
 };
 
-export interface TrackLocationProps {
+type SetNewLocationDataProps = {
   userId: string;
   locationName: string;
   lastSeenIn?: LocationData;
```
0xdevalias, line 34: This doesn't seem to be used in the function anymore?

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
-}
+};
 
-export const trackLocationEntered = ({
+export const setNewLocationData = ({
```
0xdevalias, line 37: The naming of this feels weird.. almost reversed. I would expect that `setNewLocationData` would override everything, and `updateLocationData` would probably do the more selective update.

Would probably rename `setNewLocationData` to just `setLocationData` to match the same 'format' of `updateLocationData` as well.

(Doesn't necessarily have to follow the above, but the distinction/usages of them feels a bit confusing at the moment, so if we can improve that, that would be my aim here)

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
+  updateLocationData({
+    userId,
+    newLocationData: {
+      [locationName]: getCurrentTimeInMilliseconds(),
+    },
   });
 };
 
-export const enterVenue = (venueId: string) => openUrl(venueInsideUrl(venueId));
+type UpdateCurrentLocationDataProps = {
```
0xdevalias, line 49:
```suggestion
export type UpdateCurrentLocationDataProps = {
```

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
+  userId: string;
+  profileLocationData: LocationData;
+};
+
+export const updateCurrentLocationData = ({
+  userId,
+  profileLocationData,
+}: UpdateCurrentLocationDataProps) => {
+  const [locationName] = Object.keys(profileLocationData);
```
0xdevalias, line 58: This is just going to set `locationName` to be the first array item returned from `Object.keys` yeah? That seems, wrong? Maybe?

What's the intended effect here?

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
+
+  updateLocationData({
+    userId,
+    newLocationData: { [locationName]: getCurrentTimeInMilliseconds() },
+  });
+};
+
+type ClearLocationData = {
```
0xdevalias, line 66:
```suggestion
export type ClearLocationDataProps = {
```

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
+  userId: string;
+};
+
+export const clearLocationData = ({ userId }: ClearLocationData) => {
```
0xdevalias, line 70: Since we only have a single parameter, it's probably fine to just make it `userId: string` directly, and not wrap it in object destructuring. Which also means we can get rid of the `ClearLocationData` type above.

src/utils/userLocation.ts
```diff
@@ -1,52 +1,83 @@
...
 type EnterExternalRoomProps = {
   userId: string;
   room: AnyRoom;
 };
 
 export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
-  trackLocationEntered({
+  setNewLocationData({
     userId,
     locationName: room.url,
```
0xdevalias, line 82: This shouldn't be using `room.url`, rather it should be using `venueName/roomName` slug.

src/utils/userLocation.ts
```diff
@@ -55,26 +86,32 @@ export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
 };
 
 export const useUpdateTimespentPeriodically = (
-  user: UserInfo | undefined,
-  roomName: string
+  roomName: string,
+  userId?: string
 ) => {
```
0xdevalias, line 91: Why was the position of these parameters changed? Guessing because TypeScript wanted the optional param to be last? We should probably change this to use a 'props object' with object destructuring + named keys instead of positional parameters.

src/utils/userLocation.ts
```diff
@@ -55,26 +86,32 @@ export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
...
-  const shouldUseInterval = user && roomName;
+  const shouldUseInterval = userId && roomName;
 
   useInterval(
     () => {
-      // Time spent is currently counted multiple time if multiple tabs are open
-      if (!user || !roomName) return;
+      // @debt time spent is currently counted multiple time if multiple tabs are open
```
0xdevalias, line 96: [nit]
```suggestion
      // @debt time spent is currently counted multiple times if multiple tabs are open
```

src/utils/userLocation.ts
```diff
@@ -55,26 +86,32 @@ export const enterExternalRoom = ({ userId, room }: EnterExternalRoomProps) => {
...
+      if (!userId || !roomName) return;
 
       const firestore = firebase.firestore();
-      const doc = `users/${user.uid}/visits/${roomName}`;
-      const increment = firebase.firestore.FieldValue.increment(
-        LOCATION_INCREMENT_SECONDS
-      );
+      const doc = `users/${userId}/visits/${roomName}`;
```
0xdevalias, line 100: I still feel like we would be better to use the more verbose method of `firestore.doc(doc).collection('users').doc(userId)...etc` instead of this path, as I believe as it currently is, this would potentially allow a SQL injection/path traversal sort of issue.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -84,11 +84,13 @@ export const Map: React.FC<MapProps> = ({
...
     (row: number | null, column: number | null) => {
       if (!userUid) return;
+
       makeUpdateUserGridLocation({
         venueId,
         userUid,
       })(row, column);
-      trackLocationEntered({ userId: userUid, locationName: venueName });
+
+      setNewLocationData({ userId: userUid, locationName: venueName });
```
0xdevalias, line 93: I kind of feel like the old naming of this (`trackLocationEntered`) was better.

0xdevalias (commented):
Some early review notes to align before the final review.

src/components/molecules/LiveSchedule/LiveSchedule.tsx
```diff
@@ -42,12 +42,8 @@ const LiveSchedule: FC = () => {
...
     if (!hasEvents) return null;
 
-    return events.map((event, index) => (
-      <LiveEvent
-        key={`live-event-${index}`}
-        venue={relatedVenueFor(event)}
-        event={event}
-      />
+    return events.map((event) => (
+      <EventDisplay venue={relatedVenueFor(event)} event={event} />
```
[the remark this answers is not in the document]
mike-lvov: > If we're calculating the venue from the event anyway (with relatedVenueFor(event)), is it worth having it as a separate prop?

It is not the only place, where `EventDisplay is used`. Not every time you calculate venue, based on the event.

src/components/molecules/LiveSchedule/LiveEvent.tsx
mike-lvov: It was the copy of `EventDisplay` component

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: It was used in `tracking user's location` logic before. Have refactored that

src/hooks/useRoom.ts
mike-lvov: It is not related to slug/url lastSeenIn's logic

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: This naming aligns with the other function names `updateCurrentLocation`, `clearLocation`

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -487,7 +487,7 @@ $border-radius: 28px;
     &:nth-child(odd) {
       background-color: rgba($black, 0.3);
     }
-    &.schedule-event-container_live {
+    &.schedule-event-container--live {
```
mike-lvov, line 490: This is legacy style logic and I didn't want to mess with it

src/components/molecules/LiveSchedule/LiveSchedule.scss
```diff
@@ -45,7 +45,7 @@
     &:nth-child(odd) {
       background-color: rgba($black, 0.3);
     }
-    &.schedule-event-container_live {
+    &.schedule-event-container--live {
```
mike-lvov, line 48: This is legacy style logic and I didn't want to mess with it

src/utils/userLocation.ts
mike-lvov: The intended effect is to update the current location, without rewriting it. profileLocationData can only have 1 key at any point of time

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,15 +21,40 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
+export const RoomModalContent: React.FC<RoomModalContentProps> = ({
+  room,
+  currentVenueName,
+}) => {
   const venueEvents = useSelector(venueEventsSelector) ?? [];
 
-  const { enterRoom, recentRoomUsers } = useRoom(room);
+  const { enterRoom, recentRoomUsers } = useRoom(room, currentVenueName);
 
   useMousetrap({
```
mike-lvov, line 59: I would still prefer having it here instead of <Map> `useKeyboardControls`.

1) We want to only have a listener when we need it.
2) It will require hacks to have `enterRoom` passed there since to have `enterRoom` we need to be sure there is selected room. (One way to do that would be to have noop fn returned from useRoom, but I don't want to go that way as it seems a bit hacky)

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/509bf9e36436cddbbfb4c24212ca2e48f29b2a6c

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -82,11 +83,9 @@ const NavSearchBar = () => {
     <div className="nav-search-links">
       <div className="nav-search-icon" />
       <NavSearchBarInput value={searchQuery} onChange={setSearchQuery} />
-
```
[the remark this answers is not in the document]
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/templates/Audience/Audience.tsx
```diff
@@ -142,6 +142,7 @@ export const Audience: React.FunctionComponent = () => {
   const venue = useSelector(currentVenueSelectorData);
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const userId = user?.uid;
```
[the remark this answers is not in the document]
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/molecules/EventDisplay/EventDisplay.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/pages/VenuePage/VenuePage.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/molecules/LiveSchedule/LiveSchedule.tsx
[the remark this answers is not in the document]
0xdevalias: Cool, that makes sense. Fine for it to stay as is then.

src/components/molecules/LiveSchedule/LiveEvent.tsx
0xdevalias: Cool, thanks.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Makes sense.

src/hooks/useRoom.ts
0xdevalias: I believe it may be?

```
const roomId = roomVenue ? roomVenue.name : room.url;
const { recentLocationUsers } = useRecentLocationUsers(roomId);
```

If we are trying to get `recentLocationUsers` with `room.url`, that isn't using the `venueName/roomName` slug, so will fail to retrieve them. Or so I would think? (I haven't looked deeply into the logic of `useRecentLocationUsers` to verify this yet)

src/utils/userLocation.ts
0xdevalias: Right, that makes sense. Perhaps adding a comment saying something like that would be useful, as it wasn't immediately obvious to me.

src/components/molecules/NavBar/NavBar.scss
0xdevalias: Makes sense

src/components/molecules/LiveSchedule/LiveSchedule.scss
0xdevalias: Makes sense

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: From memory, the purpose of it within the `useKeyboardControls` was so that when you are on the map, and you come across a room, you can press enter to 'bring it up' (or maybe to enter it, or similar.. I'm not 100%, as this was keeping the existing legacy logic, even though I refactored the 'how' of it)

My desire for this not to be here is to avoid us 'spreading' keyboard control logic across the app.

A question to ask here is whether we specifically need this keyboard control at all, be it in `useKeyboardControls`, or in `RoomModal`. It somewhat feels like it doesn't do a lot for us. If our desire is that we can enter the room from the room modal when this dialog is open, it may just also be possible by using existing buttons/html features. eg. 'submitting a form', or focussing the 'enter' button so that if you press enter, that's the first thing that will be 'actioned'.

If we decide that we can't use basic html features, and that we need to keep it in `RoomModal`, then as a 'final resort', we will need to change it so that its not using a global binding, but instead is 'scoped' to only register within the containing `div`. We can get that ref with `useRef`, but I believe there is also a bug in `useMousetrap` currently (detailed either in `@debt` comments, or maybe in a backlog item; I can't remember specifically) where I don't believe it will work properly with a 'non-global' binding. So there will be a bugfix required for that as well.

src/hooks/useRoom.ts
```diff
@@ -2,38 +2,42 @@ import { useCallback, useMemo } from "react";
...
-    [room, venues]
+    () => venues?.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
+    [roomUrl, venues]
   );
-  const roomId = roomVenue ? roomVenue.name : room.url;
+
+  // @debt we should replace externalRoomSlug with preferrably room id
+  const roomId = roomVenue
+    ? roomVenue.name
+    : getExternalRoomSlug({ roomTitle: room.title, currentVenueName });
```
0xdevalias, line 30:
```suggestion
  const roomId = roomVenue
    ? roomVenue.name
    : getExternalRoomSlug({ venueName: currentVenueName, roomTitle: room.title });
```

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -93,57 +94,49 @@ const VenuePage: React.FC = () => {
...
-  }, [userId, venueName]);
-
   useInterval(() => {
-    updateUserLocationToCurrentVenue();
+    if (!userId || !profile?.lastSeenIn) return;
+
+    updateCurrentLocationData({
+      userId,
+      profileLocationData: profile.lastSeenIn,
+    });
```
0xdevalias, line 103: What will happen if they don't have `profile.lastSeenIn` (is that even possible?). Won't that make this never run? Or is this more about `profile` not being present? In which case, would we be better to just make it `!profile` instead?

Since `updateCurrentLocationData` is only taking the first entry from `profileLocationData` anyway, and then updating the time on that key, it feels weird that we need to pass the full thing through. Wouldn't it just be simpler to pass the name of the key we want to update through directly, and reduce all of this additional complexity?

Or alternatively, even though I believe you said that there can/should only ever be a single entry in `lastSeenIn`, since nothing actually enforces that currently, would it not be more robust to code `updateCurrentLocationData` to update every location key that is present? Or something to that effect?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -93,57 +94,49 @@ const VenuePage: React.FC = () => {
...
+      profile?.enteredVenueIds?.includes(venueId)
+    ) {
       return;
-    if (!venueId || !user || !profile) return;
+    }
 
-    updateProfileEnteredVenueIds(profile?.enteredVenueIds, user?.uid, venueId);
-  }, [profile, user, venueId]);
+    updateProfileEnteredVenueIds(profile?.enteredVenueIds, userId, venueId);
+  }, [profile, userId, venueId]);
```
0xdevalias, line 135: It would be better for this function to not have to pass in `enteredVenueIds` directly, but instead to use firebase's `arrayUnion` helper

It would also be good to use a props object with named keys since there are multiple parameters.

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
 import { getCurrentTimeInMilliseconds } from "./time";
-import { openRoomUrl, openUrl, venueInsideUrl } from "./url";
+import { openRoomUrl } from "./url";
 
-const LOCATION_INCREMENT_SECONDS = 10;
-const LOCATION_INCREMENT_MS = LOCATION_INCREMENT_SECONDS * 1000;
+export type LocationData = Record<string, number>;
 
-type LocationData = Record<string, number>;
+export type UpdateLocationDataProps = {
```
0xdevalias, line 13: As per https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074 this should use `interface` rather than `type`

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
+  newLocationData,
+}: UpdateLocationDataProps) => {
   updateUserProfile(userId, {
     lastSeenAt: getCurrentTimeInMilliseconds(),
     lastSeenIn: newLocationData,
   });
 };
 
-export interface TrackLocationProps {
+export type SetLocationDataProps = {
```
0xdevalias, line 28: As per https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074 this should stay as `interface`

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
+}: SetLocationDataProps) => {
+  updateLocationData({
+    userId,
+    newLocationData: {
+      [locationName]: getCurrentTimeInMilliseconds(),
+    },
+  });
+};
+
+export type UpdateCurrentLocationDataProps = {
```
0xdevalias, line 45: As per https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074 this should use `interface` rather than `type`

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
   });
 };
 
-export const enterVenue = (venueId: string) => openUrl(venueInsideUrl(venueId));
+export const clearLocationData = (userId: string) => {
+  updateLocationData({ userId, newLocationData: {} });
+};
 
-type EnterExternalRoomProps = {
+export type EnterExternalRoomProps = {
```
0xdevalias, line 66: As per https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074 this should use `interface` rather than `type`

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
   userId: string;
-  room: Room;
+  roomUrl: string;
+  roomId: string;
```
0xdevalias, line 69: Since we need both `id` and `url` from `Room`, what makes it better to pass them as individual props than just passing `Room`? Are there places we need to use this where we won't have a `Room` object available?

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
     userId,
-    locationName: room.url,
+    locationName: roomId,
   });
 
-  openRoomUrl(room.url);
+  openRoomUrl(roomUrl);
+};
+
+export type UseUpdateTimespentPeriodicallyProps = {
```
0xdevalias, line 85: As per https://github.com/sparkletown/sparkle/pull/1062#issuecomment-772978074 this should use `interface` rather than `type`

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
+  locationName: string;
+  userId?: string;
 };
 
-export const useUpdateTimespentPeriodically = (
-  user: UserInfo | undefined,
-  roomName: string
-) => {
-  const shouldUseInterval = user && roomName;
+// @dept I don't think this functinality works correctly, since we only log `internal venues` in this piece of code
```
0xdevalias, line 90:
```suggestion
// @debt I don't think this functionality works correctly, since we only log `internal venues` in this piece of code
```

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
+// Could be also beneficial to log external rooms' timespent
```
0xdevalias, line 91: One possible solution to this is described in https://github.com/sparkletown/internal-sparkle-issues/issues/281

src/utils/userLocation.ts
```diff
@@ -1,81 +1,129 @@
...
+          /*
+            NOTE: it's (intended to be) because there was no document to update,
+            and so we are defaulting it to LOCATION_INCREMENT_SECONDS
+            since that's how often this code runs
+          */
+          firestore
+            .collection("users")
+            .doc(userId)
+            .collection("visits")
+            .doc(locationName)
```
0xdevalias, line 125: If you set just this part as a `const` that will give you the 'ref', you can then use it for both the `.update` and the `.set` without having to duplicate the 'path'.
```suggestion
          userLocationRef
```

0xdevalias (changes requested):
[CRO] Review based on what is in the diff here, I haven't gone through the comments in https://github.com/sparkletown/sparkle/pull/1108#pullrequestreview-580000514 1-by-1 to ensure they have all been addressed here.

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: Makes sense

src/utils/userLocation.ts
0xdevalias: See also https://github.com/sparkletown/sparkle/pull/1129#discussion_r569931904

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,15 +21,40 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
+
+  return (
+    <Modal show={show} onHide={onHide}>
+      <RoomModalContent room={room} currentVenueName={currentVenue.name} />
+    </Modal>
+  );
+};
 
-export const RoomModal: React.FC<RoomModalProps> = ({ onHide, room }) => {
+export const RoomModalContent: React.FC<RoomModalContentProps> = ({
```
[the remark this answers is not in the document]
mike-lvov: Yes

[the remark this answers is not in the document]
0xdevalias: Cool cool, makes sense.

src/utils/userLocation.ts
mike-lvov: fixed in
https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/utils/userLocation.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/hooks/useRoom.ts
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/hooks/useKeyboardControls.ts
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/53f57c40e24983470b7552d4ebed6f1715ff74ba

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -10,27 +11,32 @@ import { useRoom } from "hooks/useRoom";
 
 import "./EventDisplay.scss";
 
-interface EventDisplayProps {
+export type EventDisplayProps = {
```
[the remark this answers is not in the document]
mike-lvov: Even though I disagree with this approach, I will change it since 2/3 team members agreed on this

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -61,13 +69,15 @@ export const EventDisplay: React.FC<EventDisplayProps> = ({ event, venue }) => {
 
 type EnterRoomButtonProps = {
```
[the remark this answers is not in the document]
mike-lvov: Even though I disagree with this approach, I will change it since 2/3 team members agreed on this

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,15 +21,40 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
 
 import "./RoomModal.scss";
 
-interface RoomModalProps {
+export type RoomModalProps = {
```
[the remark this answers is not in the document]
mike-lvov: Even though I disagree with this approach, I will change it since 2/3 team members agreed on this

src/utils/room.ts
```diff
@@ -0,0 +1,9 @@
+export type GetExternalRoomSlugProps = {
```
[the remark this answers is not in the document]
mike-lvov: According to the article you were referring to, they recommend using types, when creating functions(why only functions?). Honestly, this article looks so opinionated and unreasonable, but as I understood we are going to follow their approach?

image

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570165378

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570165378

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570165378

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570165378

src/utils/userLocation.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570165378

src/utils/room.ts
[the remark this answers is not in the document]
0xdevalias: The case called out here is for props, which is an object, and falls into the ‘interface’ recommendations

src/components/molecules/EventDisplay/EventDisplay.tsx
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/molecules/EventDisplay/EventDisplay.tsx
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/molecules/LiveSchedule/LiveSchedule.tsx
```diff
@@ -42,9 +42,9 @@ const LiveSchedule: FC = () => {
   const renderedEvents = useMemo(() => {
     if (!hasEvents) return null;
 
-    return events.map((event, index) => (
-      <LiveEvent
-        key={`live-event-${index}`}
+    return events.map((event) => (
+      <EventDisplay
+        key={event.id ?? event.name}
```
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/molecules/VenueInfoEvents/VenueInfoEvents.tsx
```diff
@@ -42,9 +42,9 @@ const VenueInfoEvents: React.FunctionComponent<PropsType> = ({
               {eventsNow.length > 0 ? (
                 <div className="events-list events-list_monday">
                   {eventsNow &&
-                    eventsNow.map((event, idx) => (
+                    eventsNow.map((event) => (
```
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -94,7 +94,7 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
     () =>
       orderedEvents[date]?.events.map((event) => (
```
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -94,7 +94,7 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
...
         <EventDisplay
-          key={event.id}
+          key={event.id ?? event.name}
```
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/molecules/VenueInfoEvents/VenueInfoEvents.tsx
```diff
@@ -42,9 +42,9 @@ const VenueInfoEvents: React.FunctionComponent<PropsType> = ({
...
                       <EventDisplay
-                        key={event.name + idx}
+                        key={event.id ?? event.name}
```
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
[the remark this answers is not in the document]
mike-lvov: changed in https://github.com/sparkletown/sparkle/pull/1129/commits/84027f4323263a9ef5e9e643914436010c1d2abe

src/pages/VenuePage/VenuePage.tsx
mike-lvov: [OOS] https://github.com/sparkletown/internal-sparkle-issues/issues/302

src/utils/userLocation.ts
mike-lvov: fixed

src/utils/userLocation.ts
mike-lvov: fixed

src/utils/userLocation.ts
mike-lvov: roomId is the slug, not the room.id

mike-lvov: changed `roomId` => `roomSlug`

src/hooks/useRoom.ts
mike-lvov: fixed

src/utils/room.ts
```diff
@@ -0,0 +1,9 @@
...
+  currentVenueName: string;
+  roomTitle: string;
+};
+
+export const getExternalRoomSlug = ({
+  currentVenueName,
+  roomTitle,
+}: GetExternalRoomSlugProps) => `${currentVenueName}/${roomTitle}`;
```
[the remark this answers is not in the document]
mike-lvov: roomSlug

src/hooks/useRoom.ts
```diff
@@ -2,38 +2,42 @@ import { useCallback, useMemo } from "react";
...
 import { orderedVenuesSelector } from "utils/selectors";
+import { enterVenue } from "utils/url";
+import { getExternalRoomSlug } from "utils/room";
 
 import { useSelector } from "hooks/useSelector";
 import { useRecentLocationUsers } from "hooks/users";
 import { useUser } from "hooks/useUser";
 
-export const useRoom = (room: Room) => {
+export const useRoom = (room: Room, currentVenueName: string) => {
```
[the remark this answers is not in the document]
mike-lvov: fixed

src/components/templates/PartyMap/components/RoomAttendance.tsx
```diff
@@ -18,7 +18,7 @@ export const RoomAttendance: React.FC<RoomAttendanceProps> = ({
   venue,
   room,
 }) => {
-  const { recentRoomUsers } = useRoom(room);
+  const { recentRoomUsers } = useRoom(room, venue.name);
```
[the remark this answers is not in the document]
mike-lvov: fixed

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -48,8 +54,10 @@ export const EventDisplay: React.FC<EventDisplayProps> = ({ event, venue }) => {
           {event.description}
         </div>
         <div className="schedule-event-info-room">
-          {event.room && room ? (
-            <EnterRoomButton room={room as Room}>{buttonText}</EnterRoomButton>
+          {event.room && room && venue ? (
+            <EnterRoomButton room={room as Room} currentVenue={venue}>
+              {buttonText}
+            </EnterRoomButton>
```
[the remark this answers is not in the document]
mike-lvov: fixed

src/components/templates/Audience/Audience.tsx
```diff
@@ -254,10 +255,11 @@ export const Audience: React.FunctionComponent = () => {
 
   return useMemo(() => {
     const takeSeat = (row: number | null, column: number | null) => {
```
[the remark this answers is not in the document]
mike-lvov: added the comment

src/pages/VenuePage/VenuePage.tsx
mike-lvov: if profile.lastSeenIn is empty, there is nothing to update, thus it will not update. We do set locations in a bunch of places. I don't know the way it is going to evolve, so I'm not sure it is efficient to think that ahead. If there is change in logic in the future, a changer should understand the way it works.

profileLocationData has only one key inside. I could pass this key, but why? What if we decide to change the logic in the future?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -37,91 +63,88 @@ export const RoomModal: React.FC<RoomModalProps> = ({ onHide, room }) => {
...
-                display: "flex",
-                flexDirection: "column",
-                flexWrap: "wrap",
-                marginTop: 10,
-              }}
-            >
-              <h2 className="room-modal-title">{room.title}</h2>
-              <div className="room-modal-subtitle">{room.subtitle}</div>
-            </div>
+    <div className="container room-modal-container">
```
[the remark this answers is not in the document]
mike-lvov: changed it here

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: Let's keep the modal-keyboard thing convo here.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,15 +21,40 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
+  room,
+  currentVenueName,
+}) => {
   const venueEvents = useSelector(venueEventsSelector) ?? [];
 
-  const { enterRoom, recentRoomUsers } = useRoom(room);
+  const { enterRoom, recentRoomUsers } = useRoom(room, currentVenueName);
 
   useMousetrap({
     keys: "enter",
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570203533

src/hooks/useKeyboardControls.ts
```diff
@@ -10,7 +10,6 @@ export const MovementKeys = {
   down: ["down", "s"],
   left: ["left", "a"],
   right: ["right", "d"],
-  enter: ["enter"],
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r570203533

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,15 +21,40 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
+  currentVenueName: string;
+};
+
+export const RoomModal: React.FC<RoomModalProps> = ({
+  onHide,
+  room,
+  show,
+  currentVenue,
+}) => {
+  if (!room || !currentVenue) return null;
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r569927212

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -30,6 +28,8 @@ export const PartyMap: React.FC = () => {
 
   const [selectedRoom, setSelectedRoom] = useState<Room | undefined>();
 
+  const hasSelectedRoom = !!selectedRoom;
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/1129#discussion_r569927212

mike-lvov:
Seems like I've addressed all of the comments. I reckon this was supposed to be a blocker for deploying master => production envs? If this is still the case, should only the main things get fixed and low-medium priorities skipped?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: https://github.com/sparkletown/internal-sparkle-issues/issues/303

src/utils/room.ts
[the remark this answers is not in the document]
0xdevalias: fixed

src/pages/VenuePage/VenuePage.tsx
0xdevalias: > profileLocationData has only one key inside. I could pass this key, but why? What if we decide to change the logic in the future?

My desire was to align what is passed into the function, with what the function is doing. I don't feel particularly strongly about changing the parameter, your reasoning seems to make sense RE: continuing to use `LocationData`

> I don't know the way it is going to evolve, so I'm not sure it is efficient to think that ahead. If there is change in logic in the future, a changer should understand the way it works.

Fair enough. I guess if you haven't already, can you add a comment or two in `updateCurrentLocationData` explaining why it's only taking the first param/updating that, not the entire collection/etc. That would satisfy my request here.

src/utils/userLocation.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1129#discussion_r570166887 I believe this is still a valid request, and aligns with the other changes made in this PR to align 'props' objects to using `interface`.

src/utils/userLocation.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1129#discussion_r570166887 I believe this is still a valid request, and aligns with the other changes made in this PR to align 'props' objects to using `interface`.

src/utils/userLocation.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1129#discussion_r570166887 I believe this is still a valid request, and aligns with the other changes made in this PR to align 'props' objects to using `interface`.

src/utils/userLocation.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1129#discussion_r570166887 I believe this is still a valid request, and aligns with the other changes made in this PR to align 'props' objects to using `interface`.

src/utils/userLocation.ts
0xdevalias: [nit] Since this is just being passed into the `locationName` field, which we use elsewhere to refer to the 'location slug', should we also name this param similarly? (eg. `locationName` instead of `roomSlug`?)

src/utils/userLocation.ts
0xdevalias: @mike-lvov can you please add a comment to the code here explaining what the intent/reasoning of this line is, so that future developers understand why we are only updating a single key here, and the assumptions/etc this was written with.

src/utils/userLocation.ts
```diff
@@ -1,81 +1,125 @@
...
+export const enterExternalRoom = ({
+  userId,
+  roomSlug,
+  roomUrl,
+}: EnterExternalRoomProps) => {
+  setLocationData({
     userId,
-    locationName: room.url,
+    locationName: roomSlug,
   });
```
0xdevalias, line 80: Since `roomSlug` is just passed into `locationName` (which seems to be our standard way of referring to 'location slugs' in the code here), and also represents the 'location slug', it would probably be useful to change this to `locationName` as well.

src/utils/userLocation.ts
0xdevalias: It seems like it may not have been? (or if it was, the code may not have been pushed?)

Since we will likely be searching for this in the codebase with `@debt` (rather than `@dept`), would be good to have it aligned.

```suggestion
// @debt I don't think this functionality works correctly, since we only log `internal venues` in this piece of code
```

0xdevalias (commented):
Thanks for fixing up the changes! Seems there may have been a couple of little bits missed.

The `interface/`type` comments are NB, though to keep consistency with the other changes made in this PR for the same 'pattern' (namely props objects used for the params in functions), it would be good to fix them up. I expect all of them would only take a minute or so tops, so shouldn't add much overhead.

src/utils/userLocation.ts
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/utils/userLocation.ts
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/utils/userLocation.ts
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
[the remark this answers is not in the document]
0xdevalias: Looks good. One more question around this in https://github.com/sparkletown/sparkle/pull/1129#discussion_r570716873 (but this comment seems resolved now, so i'll mark it as such)

src/utils/room.ts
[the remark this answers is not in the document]
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/hooks/useRoom.ts
```diff
@@ -2,38 +2,47 @@ import { useCallback, useMemo } from "react";
...
 import { orderedVenuesSelector } from "utils/selectors";
+import { enterVenue } from "utils/url";
+import { getExternalRoomSlug } from "utils/room";
 
 import { useSelector } from "hooks/useSelector";
 import { useRecentLocationUsers } from "hooks/users";
 import { useUser } from "hooks/useUser";
 
-export const useRoom = (room: Room) => {
+export type UseRoomProps = {
```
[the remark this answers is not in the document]
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/utils/userLocation.ts
0xdevalias: fixed in 200c94f53638642cd3f70418f66b2accb8f392fe

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -20,108 +20,121 @@ import { RoomModalOngoingEvent, ScheduleItem } from "..";
...
+  onHide,
+  room,
+  show,
+  venue,
+}) => (
+  <Modal show={show} onHide={onHide}>
+    <div className="container room-modal-container">
+      {room && venue && <RoomModalContent room={room} venueName={venue.name} />}
+    </div>
+  </Modal>
```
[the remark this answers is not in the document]
0xdevalias: It looks like we sort of keep the transition animation (so it seems it's applied to the container/similar), but because we're un-rendering the content within the modal, it looks like it quickly loses its height as part of the 'close modal' transition. We may want to give it a min-height, or do something else to 'solve this' in the future, but it doesn't seem like a big enough issue to block/delay merging this PR over. Would be good to backlog this for future fixing/similar.

0xdevalias (approved):
The only outstanding review comments seem to be:

- https://github.com/sparkletown/sparkle/pull/1129#discussion_r570728822
  - OOS/future cleanup
- https://github.com/sparkletown/sparkle/pull/1129#discussion_r570722366
  - @mike-lvov please add this explanation comment before merging
  
Based on that, i'm happy to approve this, with the caveat that the above comment is added before merging.

0xdevalias:
> Seems like I've addressed all of the comments. I reckon this was supposed to be a blocker for deploying master => production envs? If this is still the case, should only the main things get fixed and low-medium priorities skipped?

@mike-lvov I cleaned up a few of the little bits from my latest review quickly in https://github.com/sparkletown/sparkle/pull/1129/commits/200c94f53638642cd3f70418f66b2accb8f392fe

Based on that, the only outstanding thing needed before merging is adding the comment mentioned in https://github.com/sparkletown/sparkle/pull/1129#pullrequestreview-584033564

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/internal-sparkle-issues/issues/306

PR #1131, opened 2021-02-04
Improve reporting scripts + allow them to filter based on zoom 'event' tracking field

0xdevalias:
- convert `reporting/fetch-reports` to TypeScript + cleanup
- output the total number of zoom reports within the specified date range in `reporting/fetch-reports`
- adds the ability to filter reports to download by 'event' tracking field to `reporting/fetch-reports`
- output the number of reports downloaded in `reporting/fetch-reports`
- add `requirements.txt` for the dependencies of `makegraphs.py`
- update parts of the README
- maybe some other things

(partially) fixes https://github.com/sparkletown/internal-sparkle-issues/issues/249


reporting/fetch-reports.ts
```diff
@@ -0,0 +1,305 @@
+#!/usr/bin/env node -r esm -r ts-node/register
```
0xdevalias, line 1: **Note:** It's probably worth diffing this file against the previous file to see what has actually changed, rather than attempting to view/review it in its entirety.

I believe something like the following (while on this branch) should work:

```
git diff staging:reporting/fetch-reports.js HEAD:reporting/fetch-reports.ts
```

0xdevalias:
Note: I have marked this PR as ready for review (was previously in draft), though I haven't actually read through it in a long time to see what/if it is 100% complete.

reporting/fetch-reports.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  return numberOfReportsTotal;
+};
+
+// Log in to zoom, and download all participants reports from the above selected dates.
+(async () => {
+  const browser = await puppeteer.launch({ headless: false });
+  const page = await browser.newPage();
+
+  await page.setViewport({ width: 1400, height: 900 });
```
mike-lvov, line 143: Do we need to explicitly set the size of the viewport?

mike-lvov (approved):
LGTM

reporting/fetch-reports.ts
0xdevalias: That would have just been me reworking the existing code that was there, so.. not sure, but I’d rather leave it for now than spend any time looking into it.

I want to get rid of/change this method entirely when we can, to just use the APIs/similar.

reporting/README.md
```diff
@@ -9,37 +9,45 @@ The `makegraphs.py` script will create `analytics.pdf` based on Usage Reports fr
...
-$ pip3 install hv numpy matplotlib # there may be others, please update the docs!
+
+# Install required python packages
+$ pip3 install -r requirements.txt
+
+# If that doesn't work, try installing manually as follows
+# $ pip3 install holoviews matplotlib
+
+# Finally, run the script
 $ python3 makegraphs.py ~/Downloads/My\ Event\ CSVs/*.csv # python 3 compatible, unsure about python2
```
0xdevalias, line 51:
```suggestion
$ ./makegraphs.py ~/Downloads/My\ Event\ CSVs/*.csv # python 3 compatible, unsure about python2

# Or if that doesn't work for you, you can try:
#   python3 makegraphs.py ~/Downloads/My\ Event\ CSVs/*.csv
```

PR #1141, opened 2021-02-05
Fix/add uuuidv4

denisdimitrov:
add missing dep


0xdevalias:
@denisdimitrov Can you please add context for why this was thought to be needed, and why it was subsequently closed.

0xdevalias:
Looking further, https://github.com/sparkletown/sparkle/pull/1142 seems to have been opened in it's place.

PR #1015, opened 2020-12-15
WIP: [FEATURE] Adds access mode to venue landing page

denisdimitrov:
WIP of the access mode implementation for the venue landing page.

P.S. Should I leave `showSecretPassword` for backwards compatibility?

PR #1148, opened 2021-02-08
Venue Access - UI - add access venue options to venue landing page

denisdimitrov:
(**Note:** This is a re-creation of #1134, which was a re-creation of #1015)

Add's venue access UI and some fixes for the venue access logic that was previously introduced.

partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/218


src/components/organisms/AuthenticationModal/AuthenticationModal.tsx
```diff
@@ -13,16 +13,18 @@ export enum AuthOptions {
   initial = "initial",
 }
 
-interface PropsType {
+interface AuthenticationModalProps {
```
0xdevalias, line 16: [nit]
```suggestion
export interface AuthenticationModalProps {
```

src/components/organisms/AuthenticationModal/AuthenticationModal.tsx
```diff
@@ -13,16 +13,18 @@ export enum AuthOptions {
...
   show: boolean;
   onHide: () => void;
+  headerMessage?: string;
   afterUserIsLoggedIn?: () => void;
   showAuth: AuthOptions;
 }
 
-export const AuthenticationModal: React.FunctionComponent<PropsType> = ({
+export const AuthenticationModal: React.FunctionComponent<AuthenticationModalProps> = ({
```
0xdevalias, line 24: [nit]
```suggestion
export const AuthenticationModal: React.FC<AuthenticationModalProps> = ({
```

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useCallback } from "react";
+import { VenueAccessMode } from "types/VenueAcccess";
+import { AnyVenue } from "types/venues";
+import { getCurrentTimeInUTCSeconds, getTimeBeforeParty } from "utils/time";
+import { joinVenue } from "utils/url";
+import { SecretPasswordForm } from "../SecretPasswordForm";
+
+interface VenueJoinButtonProps {
```
0xdevalias, line 8: [nit]
```suggestion
export interface VenueJoinButtonProps {
```

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  venueId?: string;
+  venue: AnyVenue;
+  onPasswordSubmit: () => void;
+  onPasswordSuccess: () => void;
```
0xdevalias, line 12: Make these optional so that callers don't need to potentially pass in noop functions + make sure to properly check they are defined before calling them later on.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  venueId,
+  venue,
+  onPasswordSubmit,
+  onPasswordSuccess,
+}) => {
+  const handleJoinVenue = useCallback(() => {
+    joinVenue(venueId, venue?.entrance);
+  }, [venue, venueId]);
+
+  // @debt Handle emails and codes as well.
```
0xdevalias, line 25: Why is this `@debt`? Seems related to a core part of the PR/venue access feature?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+        />
+      </div>
+    );
+  }
+
+  return (
+    <button
+      className="btn btn-primary btn-block btn-centered"
+      onClick={handleJoinVenue}
+    >
```
0xdevalias, line 44: This will always show the secret password form if that mode is configured, so the JSX below won't be rendered. Is that the desired state/logic flow here?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    joinVenue(venueId, venue?.entrance);
+  }, [venue, venueId]);
+
+  // @debt Handle emails and codes as well.
+  if (venue.access === VenueAccessMode.Password) {
+    return (
+      <div className="secret-password-form-wrapper">
+        <SecretPasswordForm
+          buttonText={
+            venue.config?.landingPageConfig.joinButtonText ?? "Join the party"
```
0xdevalias, line 31: Should this be pulled out into a `const` and shared with the 'join the event' text below?

Also, do we want 'party' to be the default fallback? I think we were aiming to use generic terms such as 'event' across the platform, though can't remember.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <button
+      className="btn btn-primary btn-block btn-centered"
+      onClick={handleJoinVenue}
+    >
+      Join the event
+      {(venue?.start_utc_seconds ?? 0) > getCurrentTimeInUTCSeconds() && (
+        <span className="countdown">
+          Begins in {getTimeBeforeParty(venue?.start_utc_seconds)}
+        </span>
+      )}
```
0xdevalias, line 50: Do we want to show this countdown regardless of the Venue Access protection method?

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { FC, Fragment, useCallback } from "react";
+import { Modal } from "react-bootstrap";
+
+import { AnyVenue } from "types/venues";
+
+import { joinVenue } from "utils/url";
+
+import "./JoinVenueModal.scss";
+
+interface JoinVenueModalProps {
```
0xdevalias, line 10: [nit]
```suggestion
export interface JoinVenueModalProps {
```

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+  venueId?: string;
+  venue: AnyVenue;
+  show: boolean;
+  isLoggedIn: boolean;
+  onHide?: () => void;
+}
+
+export const JoinVenueModal: FC<JoinVenueModalProps> = ({
```
0xdevalias, line 18: [nit]
```suggestion
export const JoinVenueModal: React.FC<JoinVenueModalProps> = ({
```

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { FC, Fragment, useCallback } from "react";
```
0xdevalias, line 1: [nit]

```suggestion
import React, { useCallback } from "react";
```

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <Modal show={show} onHide={hideModal}>
+      <Modal.Body>
+        <Fragment>
+          <div className="join-venue-title">
+            {isLoggedIn
+              ? `You are now attending ${venue.name}`
+              : "You need an account to join this party"}
+          </div>
```
0xdevalias, line 41: As per https://github.com/sparkletown/sparkle/pull/1134#discussion_r571795309

> "You are now attending :venueName" doesn't seem like the right sort of wording here.
> 
> @sparkletown/design @sparkletown/product Any thoughts?

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const JoinVenueModal: FC<JoinVenueModalProps> = ({
+  show,
+  venueId,
+  venue,
+  isLoggedIn,
+  onHide,
+}) => {
+  const handleJoinVenue = useCallback(() => {
+    joinVenue(venueId, venue.entrance);
+  }, [venue.entrance, venueId]);
```
0xdevalias, line 27: Similar code is used in `src/components/molecules/VenueJoinButton/VenueJoinButton.tsx`

Under what cases are the modal used vs the button? Can the modal use the button?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -81,19 +87,44 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
+
+  const hideJoinVenueModal = useCallback(() => {
+    setJoinVenueModalOpen(false);
+  }, []);
+
+  const handlePasswordSubmit = () => {
+    if (user) return;
+
+    openAuthenticationModal();
+  };
```
0xdevalias, line 127: `useCallback`'s for each of these

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -375,11 +365,21 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
-        showAuth={AuthOptions.register}
+      {!isLoggedIn && (
+        <AuthenticationModal
+          show={isAuthenticationModalOpen}
+          onHide={closeAuthenticationModal}
+          headerMessage={"You need an account to join this party"}
+          afterUserIsLoggedIn={() => setShouldOpenPaymentModal(true)}
+          showAuth={AuthOptions.initial}
+        />
+      )}
```
0xdevalias, line 376: I don't believe we ever want to conditionally render/unrender modals. I believe the intent is that the `show` prop is used for this purpose. So probably 
```suggestion
      <AuthenticationModal
        show={!isLoggedIn && isAuthenticationModalOpen}
        onHide={closeAuthenticationModal}
        headerMessage={"You need an account to join this party"}
        afterUserIsLoggedIn={() => setShouldOpenPaymentModal(true)}
        showAuth={AuthOptions.initial}
        />
```

src/utils/url.ts
```diff
@@ -39,6 +42,21 @@ export const openRoomUrl = (url: string) => {
 
 export const enterVenue = (venueId: string) => openUrl(venueInsideUrl(venueId));
 
+export const joinVenue = (
+  venueId?: string,
+  venueEntrance?: EntranceStepConfig[]
+) => {
+  if (!venueId) return;
+
+  const hasVenueEntrance = venueEntrance ? !!venueEntrance.length : false;
```
0xdevalias, line 51: Do we want to pass the whole `venueEntrance` in here, just to check if it is empty? Or should we make the param a boolean instead, and push that logic outside of this helper function?

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: [OOS] We seem to use this `useState` with a boolean, then `useCallback`'d 'setTrue`, `setFalse` pattern a fair bit.

We could make a `useBooleanState` or `useToggleState`  or similar that wraps up this pattern for us. I would probably make it return an array of 3 things. The state, the setTrue, and the setFalse function (using an array so that people can choose whatever name they want for the functions in a similar way to how we currently use `useState`)

0xdevalias (changes requested):
As per requests in https://github.com/sparkletown/sparkle/pull/1148#pullrequestreview-586112493

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
```diff
@@ -1,34 +1,51 @@
...
 import { isTruthy } from "utils/types";
 
 import { useVenueId } from "hooks/useVenueId";
+import { useUser } from "hooks/useUser";
 
 import "./SecretPasswordForm.scss";
 
-const SecretPasswordForm = ({ buttonText = "Join the party" }) => {
-  const history = useHistory();
+interface SecretPasswordFormProps {
```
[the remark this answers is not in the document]
denisdimitrov: Is there a particular reason to export something that's not being used from the outside of this file?

src/components/organisms/AuthenticationModal/AuthenticationModal.tsx
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1148#discussion_r572755911

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1148#discussion_r572755911

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: When the PR was created I was given UI only for password. I don't have any information about `codes` and `emails`.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Yes, this was a `renderFunction` with switch case returning different components. Now it's a component with the intention to add `codes` and `emails` with different button or UI

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Yes, this should be a `const`, I'm going to change it.

I believe 'party' was the default state. Since most of the venues are parties and not events. We should bring this up on the daily tomorrow for more info and then add the default terms to `settings.ts`

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Good question, I don't have enough info to answer though. Ed is the one who knows more about this. I'll ask tomorrow on the daily.

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
denisdimitrov: Modal is used to show text and redirect button after you enter the password or force you to login if you are not logged in and try to enter the venue.

The button is used to check the password and show the modal.

src/utils/url.ts
denisdimitrov: Moved it outside as a prop.

denisdimitrov:
@0xdevalias  Most of the comments have been fixed. The rest of them, just answered. Ready for rereview.

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
denisdimitrov: @sparkletown/engineering 
I'd like to say that I prefer to use named fragments because of the ability to use `key`. Sometimes you need it and we should keep things consistent. I'd also like to say that I'd prefer if we directly import Fragment and FC instead of using the dot syntax `React.FC`. They are named exports and we should take advantage of that, we don't import hooks using `React.useState` why import FC with it?. Of course, that's just my personal preference. I believe it's up to @0xdevalias to decide if that makes sense or not.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: Mostly 'past frustration'/annoyance with libraries that 'lock away' their types, where I might want to use them. I feel like the type of the component's props forms part of its 'public contract', so it just "feels more right" that they are exported along with it.

At the end of the day, if they aren't, it's not going to break things, and at worst it may just require additional changes if something wants to use them in the future. It's more of a personal preference around 'doing things well', as I tend to think from the perspective of what I would expect/want to see from a public library I was consuming/etc.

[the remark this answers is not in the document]
0xdevalias: Actually, it is often also really useful when it comes to writing tests/etc, and being able to reuse things like that, without having to go and also change the source files when you are.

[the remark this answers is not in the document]
0xdevalias: (captured this in https://github.com/sparkletown/sparkle/pull/1062#issuecomment-776407944)

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
0xdevalias: I have no strong feelings on the shorthand fragment syntax vs the longhand. In the past I actively chose to use the longhand form, as it is more explicit/'self-documenting' to other users as to what it is (eg. it's hard to do a google search for `<>` and find meaningful results, but searching for 'react fragment' will give you much better insight)

I personally don't feel that 'absolute consistency' is required, if we decided that the shorthand is better for most cases, and then we only used the longhand form where it is required (which is pretty much only when we need to add `key` to it I believe) In many cases where we would want to add `key` to it though, I would question if Fragment was the right choice, as off the top of my head, I can't think of many use cases where 'smooshing' an array of things together without containing divs/similar to 'group' them feels 'right' (i'm sure there are cases I'm not immediately thinking of though)

My preference towards using `React.Fragment` and `React.FC` is that I feel like it provides more 'context'/etc to describe what it is. I'm a little bit either way on `React.Fragment` (my preference in the past was to use the `React.` prefix explicitly, I'm not sure if I feel particularly strongly about that anymore); but with `React.FC`, for the same reason that I prefer variable names to be 'descriptive', I feel like `FC` is too 'obscure' on it's own, and spelling out the full `FunctionalComponent` version of it feels overly verbose; so for me, `React.FC` seems to fit that middleground of terse, yet descriptive.

0xdevalias: (captured in https://github.com/sparkletown/sparkle/pull/1062#issuecomment-776410919)

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
0xdevalias: Makes sense. I guess the part I was (am?) confused apart is why both of them have the `handleJoinVenue` code, and whether there is a simple/quick way that we avoid that duplication.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Hrmm, fair enough.

That's probably something that we may need to get from @sparkletown/design and/or @sparkletown/product 

Can you please sync up on that, and/or create a backlog ticket to track adding these parts.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Cool, makes sense :)

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Agreed RE: chatting in standup + adding whatever the 'correct default term' is to `settings.ts`

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Sounds good.

functions/access.js
```diff
@@ -136,7 +136,7 @@ const createToken = async (venueId, uid, password, email, code) => {
 };
 
 exports.checkAccess = functions.https.onCall(async (data, context) => {
-  if (!data || !context) return { token: undefined };
+  if (!data || !context || !context.auth) return { token: undefined };
 
   if (
     context &&
```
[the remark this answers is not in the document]
0xdevalias: Should `context.auth.uid` also be getting checked in the first guard statement too?

functions/access.js
```diff
@@ -136,7 +136,7 @@ const createToken = async (venueId, uid, password, email, code) => {
 };
 
 exports.checkAccess = functions.https.onCall(async (data, context) => {
```
0xdevalias, line 138: Should we be using our generic `checkAuth` helper function at the start of this (and other endpoints in this file)?

- https://github.com/sparkletown/sparkle/blob/staging/functions/auth.js#L4-L12

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: @denisdimitrov This still seems to be outstanding?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Reminder to get back to this + make the const after standup discussion/etc.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <button
+      className="btn btn-primary btn-block btn-centered"
+      onClick={handleJoinVenue}
+    >
+      Join the event
```
0xdevalias, line 45: We should also use whatever default we add to `settings.ts` in place of 'event' here

src/components/organisms/AuthenticationModal/InitialForm/InitialForm.tsx
```diff
@@ -19,7 +21,7 @@ export const InitialForm: FC<InitialFormProps> = ({
       >
         Create your account
       </div>
-      <div className="buttons-separator">or</div>
+      {showSeparator && <div className="buttons-separator">or</div>}
```
0xdevalias, line 24: Not sure how this displays in the UI, so just want to confirm that this doesn't need any additional spacing around the "or" for it to display correctly (eg. don't want it showing up as "Create your accountorLog In")

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
0xdevalias: @denisdimitrov Would be good to check in on wording for this in standup tonight + also use whatever default terminology we decided on for 'party' loaded from `settings.ts`

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { SecretPasswordForm } from "../SecretPasswordForm";
+
+interface VenueJoinButtonProps {
+  venueId?: string;
+  venue: AnyVenue;
+  onPasswordSubmit: () => void;
+  onPasswordSuccess: () => void;
+}
+
+export const VenueJoinButton: React.FC<VenueJoinButtonProps> = ({
```
0xdevalias, line 15: [nit] We have `VenueJoinButton` and `JoinVenueModal`. Would be good to standardise/align the naming as `VenueJoin*` or `JoinVenue*`. I don't have a strong preference on which one.

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+            {isLoggedIn
+              ? `You are now attending ${venue.name}`
+              : "You need an account to join this party"}
+          </div>
+          <button
+            className="btn btn-primary btn-block btn-centered"
+            onClick={handleJoinVenue}
+          >
+            {`Let's go`}
+          </button>
```
0xdevalias, line 47: [NB] This part still feels like it should be what the `VenueJoinButton` component handles, and maybe the 'extra stuff' around checking things and/or popping up this modal/etc that are currently in it perhaps somewhere else?

Basically, having a tiny component that wraps up the button, the text we use for it, the 'enter venue' handler logic, etc, that gets reused wherever we need that button. As right now, this feels duplicated (which isn't a huge deal, but does mean there is a far higher chance of 'fixes'/changes being applied to only one of them, and then them getting 'out of sync' from one another)

0xdevalias: Also.. the logic above switches the text based on `isLoggedIn`, but this button will always be shown. Is that correct? Feels like I should be given some kind of 'go log in' button or similar? Or maybe just this modal never even popped up, and we go straight to the 'you need to login' or similar?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -376,10 +366,18 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
         />
       )}
       <AuthenticationModal
-        show={isAuthenticationModalOpen}
+        show={isLoggedIn && isAuthenticationModalOpen}
         onHide={closeAuthenticationModal}
+        headerMessage={"You need an account to join this party"}
```
0xdevalias, line 371: [nit]

```suggestion
        headerMessage="You need an account to join this party"
```

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -376,10 +366,18 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
         afterUserIsLoggedIn={() => setShouldOpenPaymentModal(true)}
```
0xdevalias, line 372: Should this also be wrapped in a `useCallback`'d handler rather than inline?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -376,10 +366,18 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
-        showAuth={AuthOptions.register}
+        showAuth={AuthOptions.initial}
```
0xdevalias, line 373: Was this change from `AuthOptions.register` to `AuthOptions.initial` intended? What was the intent/resoning?

src/utils/url.ts
```diff
@@ -39,6 +40,16 @@ export const openRoomUrl = (url: string) => {
 
 export const enterVenue = (venueId: string) => openUrl(venueInsideUrl(venueId));
 
+export const joinVenue = (venueId?: string, hasVenueEntrance?: boolean) => {
+  if (!venueId) return;
+
```
0xdevalias, line 45: Should we make `venueId` a required prop here, and push the 'if no venueId do nothing' logic back to the calling side? As-is, it feels like it could be a bit confusing.

Also, potentially use named props for the 2nd param/'options', so that instead of looking like `joinVenue('foo', true)` on the calling side, it can be more semantic/self documenting: `joinVenue('foo', { hasVenueEntrace: true })`

```suggestion
export interface JoinVenueOptions {
  hasVenueEntrance?: boolean
}

export const joinVenue = (venueId: string, { hasVenueEntrance }: JoinVenueOptions) => {
```

0xdevalias (commented):
Comments/questions + pointed out a few things that seem to be outstanding from the last review.

@denisdimitrov Can you also please make sure to 'resolve' any of the fixed comments/etc from past reviews (+ ideally a note about them being fixed/etc). It helps me to be able to quickly see what has been done, what is still outstanding, etc; without having to go and explicitly check each one individually.

functions/access.js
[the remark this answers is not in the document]
denisdimitrov: Makes sense.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
```diff
@@ -1,34 +1,54 @@
...
+  buttonText: string;
+  onPasswordSubmit?: () => void;
+  onPasswordSuccess?: () => void;
+}
+
+export const SecretPasswordForm: React.FC<SecretPasswordFormProps> = ({
+  buttonText = "Join the party",
```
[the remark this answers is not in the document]
denisdimitrov: Moving it to `settings.ts` and adding it in all of the usages.

functions/access.js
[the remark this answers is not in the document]
denisdimitrov: > Should context.auth.uid also be getting checked in the first guard statement too?

Yeah, that sounds like a reasonable change. I'll move it.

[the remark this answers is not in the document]
denisdimitrov: > Since context and context.auth are being checked above now, they shouldn't also need to be checked here too.

Agreed.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Move to `settings.ts` and used.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Components names should start with a noun. I will change it.

src/pages/VenueLandingPage/VenueLandingPage.tsx
denisdimitrov: Changed to const from `settings.ts`

src/pages/VenueLandingPage/VenueLandingPage.tsx
denisdimitrov: `initial` shows both register and login buttons and lets the user choose one of them instead of forcing them to register. Just because you are not logged in and you try to join a venue doesn't mean that you don't have an account. This was also in the design when I was writing this thing but it was in trello and it was from a screenshot.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
```diff
@@ -1,34 +1,51 @@
...
+      if (!venueId) return;
+
       e.preventDefault();
       setMessage("Checking password...");
+      onPasswordSubmit && onPasswordSubmit();
+
+      if (!user) {
+        setMessage("");
+        return;
+      }
```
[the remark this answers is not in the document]
denisdimitrov: We want `onPasswordSubmit` to run if there is `venueId` but regardless if the user exists or not.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
```diff
@@ -1,34 +1,51 @@
...
     setMessage("");
     setError(false);
-  }
+  };
 
   const passwordSubmitted = useCallback(
     async (e) => {
+      if (!venueId) return;
+
       e.preventDefault();
```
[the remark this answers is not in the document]
denisdimitrov: If venueId doesn't exist this component won't render at all, because of the guards in the higher hierarchy components. In this case I don't think it matters that much, but I guess we want to preventDefault only if there is some logic being executed. In this case only when venueId exists.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
denisdimitrov: Makes sense I'll export the interfaces as well. I'll consider this as a convention and do it from now on.

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: Sounds reasonable to me. Thanks for explaining :)

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: _nods_ fair enough.

What is `onPasswordSubmit` / `onPasswordSuccess` used for?

My 'concern' (unsure if it's a valid one) is that when `!user`, onPasswordSuccess will never be run. You know the usage/logic of the related parts better than me, so if that's not a problem as you see/understand it, then happy to go with your judgement there. Just wanted to make sure I spelled out my 'concern' explicitly.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Fixed.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Yes, will do.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: _nods_ cool, makes sense

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: Great, thank you for that :)

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Thankyou :)

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: Changed locally, will let you know when it's ready for re-review.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
denisdimitrov: This is supposed to be visible for all of the venue access types. I spoke with Ed about this yesterday.

functions/access.js
denisdimitrov: Yeah I'll add.

src/utils/url.ts
denisdimitrov: Sounds like a reasonable change, I like the idea.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
0xdevalias: Cool. Then will probably require some changes here to ensure that (as from memory, it was only shown on the 'final branch' of the conditional renders.

functions/access.js
0xdevalias: Thanks :)

src/components/organisms/AuthenticationModal/InitialForm/InitialForm.tsx
denisdimitrov: This is shown when you enter a venue without an account.

login

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
```diff
@@ -1,34 +1,54 @@
...
-    setError(false);
-  }
+  const changePassword = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      setPassword(e.target.value);
+      setMessage("");
+      setError(false);
+    },
+    []
+  );
```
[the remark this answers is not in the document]
denisdimitrov: looks and sounds interesting but isn't this an anti pattern? Directly manipulating the DOM is not one of the best practices and should be used only if necessary.

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
denisdimitrov: I think if I prop drill it, I can have only 1 handleJoinVenue function.

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
denisdimitrov: onPasswordSubmit shows the authentication modal if you are logged out. That's pretty much it.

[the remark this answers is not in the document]
denisdimitrov: It almost feels like one of the 2 components is unnecessary (Modal and Button). Maybe I should merge them into one.

src/components/organisms/JoinVenueModal/JoinVenueModal.tsx
denisdimitrov: VenueJoinButton should be something like VenueAccessMode or VenueAccessForm because it's responsible to render different button / UI depending on the access in that venue

src/components/organisms/AuthenticationModal/InitialForm/InitialForm.tsx
0xdevalias: Ah, cool. No issue here then :)

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: We're not directly manipulating the DOM, we are just getting a reference to the input element, and reading a value from it.

If you wanted to go a step further with it, perhaps we should be aligning all of our textbox/form/etc code to use our proper 'rect form handling' libraries, rather than 'doing it directly' with state/refs/etc. Which looking at our `package.json` is:

- https://react-hook-form.com/

src/components/molecules/SecretPasswordForm/SecretPasswordForm.tsx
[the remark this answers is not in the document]
0xdevalias: > It almost feels like one of the 2 components is unnecessary (Modal and Button). Maybe I should merge them into one.

Yeah, I think I agree. I'm not sure of the specifics to give further suggestion at this stage, but it just kind of 'feels wrong' at the moment. Interested to see what you come up with to 'fuse' these.

functions/access.js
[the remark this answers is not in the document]
0xdevalias: @denisdimitrov The changes in https://github.com/sparkletown/sparkle/pull/1148/commits/d3c30b88a45c6ac26105324dcad1176e2ea00402 seem at odds with this request, and seem like they may have regressed?

src/utils/url.ts
```diff
@@ -1,6 +1,11 @@
 import Bugsnag from "@bugsnag/js";
+
 import { VALID_URL_PROTOCOLS } from "settings";
 
+export interface JoinVenueOptions {
+  hasVenueEntrance?: boolean;
+}
```
0xdevalias, line 7: [NB] I'm not sure if I have strong feelings one way or the other about this, but usually I tend to keep the types close to the function/etc that they relate to. Though I believe other patterns/conventions keep them at the top of the file, as you've done here. Don't need to make changes here.. just wanted to point it out/bring it up as something we should probably consider/standardise on as a team.

0xdevalias (commented):
A few questions/comments/potential small changes, but I think this is looking pretty close to done.

src/utils/url.ts
denisdimitrov: I agree, we should discuss this and add it in the documentation PR if it's missing there.

src/components/organisms/VenueJoinModal/VenueJoinModal.scss
```diff
@@ -0,0 +1,4 @@
+.join-venue-title {
```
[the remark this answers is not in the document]
denisdimitrov: I'm totally up for BEM. If this is our final decision I'll start using it everywhere.

functions/access.js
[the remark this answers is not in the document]
denisdimitrov: I was testing some edge cases in the logic and forgot to revert it. I'll change it back.

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <button
+      className="btn btn-primary btn-block btn-centered"
+      onClick={handleJoinVenue}
+    >
+      {DEFAULT_PARTY_BUTTON_TEXT}
```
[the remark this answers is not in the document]
denisdimitrov: Makes sense

functions/access.js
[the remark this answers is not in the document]
denisdimitrov: Also I didn't use `checkAuth` because it's returning an error if you are not logged in. And we need this logic to run even if you are not logged in.

functions/access.js
```diff
@@ -139,7 +139,6 @@ exports.checkAccess = functions.https.onCall(async (data, context) => {
   if (!data || !context) return { token: undefined };
```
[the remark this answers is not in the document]
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1148#discussion_r574366106

functions/access.js
```diff
@@ -139,7 +139,6 @@ exports.checkAccess = functions.https.onCall(async (data, context) => {
...
 
   if (
-    context &&
     context.auth &&
     context.auth.uid &&
```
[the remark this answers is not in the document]
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1148#discussion_r574366106

denisdimitrov:
@0xdevalias Still not ready with the changes, hold up with the re-review for now.

0xdevalias (commented):
@denisdimitrov Just bouncing the 'waiting on' back to you as per https://github.com/sparkletown/sparkle/pull/1148#issuecomment-777630859

denisdimitrov:
@0xdevalias Ready for rereview.

functions/access.js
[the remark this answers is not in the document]
0xdevalias: Oh? What about it needs to run before logged in out of curiosity? Shouldn't we be logged in before we try and access a venue?

src/components/organisms/VenueJoinModal/VenueJoinModal.scss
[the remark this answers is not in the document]
0xdevalias: I'm a fan of it, and have been using it in my code of late. We should probably start a thread in `#tech` and get everyones thoughts/opinions/buy-in/objections on it.

src/utils/url.ts
0xdevalias: Sounds good, can you please raise this in `#tech`

src/components/molecules/VenueJoinButton/VenueJoinButton.scss
```diff
@@ -0,0 +1,4 @@
+.venue-join-modal__title {
```
0xdevalias, line 1: The filename is `VenueJoinButton`, but this is `.venue-join-modal`, we should align these to be the same.

0xdevalias: Also, following BEM, we should have a 'block' class here, as well as the 'element' (as defined).

I would tend to write the styles for this as:

```scss
.venue-join-modal {
  &__title {
    text-align: center;
    padding-bottom: 20px;
  }
}
```

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+                Begins in {getTimeBeforeParty(venue.start_utc_seconds)}
+              </span>
+            )}
+          </button>
+          <div className="form-group">{message}</div>
+          {errors.password && errors.password.type === "required" && (
+            <span className="input-error">Password is required</span>
+          )}
+        </form>
+      </div>
```
0xdevalias, line 162:
```suggestion
      </div>

```

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+                className="secret-password-input"
+                name="password"
+                ref={register({ required: true })}
+                placeholder="password"
+                autoFocus
+                onChange={changePassword}
+                id="password"
+              />
+            </>
+          )}
```
0xdevalias, line 145:
```suggestion
          )}

```

0xdevalias, line 145: Presumably this is going to need to change to support other forms of venue access? Should we be checking `venue.access === 'foo-password-mode-whatever-it-is'` here?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+                Got an invite? Join in with the secret password
+              </p>
+              <input
+                className="secret-password-input"
+                name="password"
+                ref={register({ required: true })}
+                placeholder="password"
+                autoFocus
+                onChange={changePassword}
+                id="password"
```
0xdevalias, line 142: Do we need an `id` field here at all?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+      <div className="secret-password-form-wrapper">
+        <form
+          className="secret-password-form"
+          onSubmit={handleSubmit(submitPassword)}
+        >
+          {venue.access && (
+            <>
+              <p className="small-text">
+                Got an invite? Join in with the secret password
+              </p>
```
0xdevalias, line 134: Should this be a `<label>` for the input field below?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+          <button
+            className="btn btn-primary btn-block btn-centered"
+            type="submit"
+          >
+            {buttonText}
+            {(venue.start_utc_seconds ?? 0) > getCurrentTimeInUTCSeconds() && (
+              <span className="countdown">
+                Begins in {getTimeBeforeParty(venue.start_utc_seconds)}
+              </span>
+            )}
```
0xdevalias, line 155: Is this going to join `buttonText` with `Begins in...` as "{buttonText}Begins in..." (eg. no spacing between them)

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+          </button>
+          <div className="form-group">{message}</div>
+          {errors.password && errors.password.type === "required" && (
```
0xdevalias, line 158:
```suggestion
          {errors.password?.type === "required" && (
```

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <span className="input-error">Password is required</span>
+          )}
+        </form>
+      </div>
+      <Modal show={isJoinVenueModalOpen} onHide={hideJoinVenueModal}>
+        <Modal.Body>
```
0xdevalias, line 164: Can classes be applied here? And if so, should we be applying the 'block' class of `venue-join-modal`?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }, [clearError]);
+
+  const submitPassword = useCallback(
+    async ({ password }) => {
+      if (!venueId) return;
+
+      if (!venue.access) {
+        handleJoinVenue();
+        return;
+      }
```
0xdevalias, line 84: Since this does more than submit the password, perhaps a better name for the `const`? Maybe `checkAccessOrJoinVenue` or similar?

0xdevalias: Also, should we be calling `handleJoinVenue` directly here? Or do we still want to be popping up the modal?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+    ]
+  );
+
+  return (
+    <>
+      <div className="secret-password-form-wrapper">
+        <form
+          className="secret-password-form"
+          onSubmit={handleSubmit(submitPassword)}
+        >
```
0xdevalias, line 129: Should this be outside of the `venue.access` conditional rendering below?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
+import React, { useCallback, useState } from "react";
```
0xdevalias, line 1: Is `VenueJoinButton` still an accurate name for this component?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+          )}
+        </form>
+      </div>
+      <Modal show={isJoinVenueModalOpen} onHide={hideJoinVenueModal}>
+        <Modal.Body>
+          <div className="venue-join-modal__title">
+            {isLoggedIn
+              ? `You are now attending ${venue.name}`
+              : DEFAULT_PARTY_JOIN_MESSAGE}
+          </div>
```
0xdevalias, line 169: 'You are now attending venue' seems like the wrong text to be showing here, given this ternary is only checking if we are logged in or not.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -81,19 +87,40 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
 
-  const { user } = useUser();
+  const openPaymentModal = useCallback(() => {
+    setShouldOpenPaymentModal(true);
+  }, []);
+
+  const closePaymentModal = useCallback(() => {
+    setPaymentModalOpen(false);
+  }, []);
+
```
0xdevalias, line 110: Should `openPaymentModal` be setting `setPaymentModalOpen(true)`, rather than `setShouldOpenPaymentModal`?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -105,7 +132,7 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
 
   useEffect(() => {
     if (shouldOpenPaymentModal && !isAuthenticationModalOpen) {
-      setIsPaymentModalOpen(true);
+      setPaymentModalOpen(true);
       setShouldOpenPaymentModal(false);
     }
   }, [shouldOpenPaymentModal, isAuthenticationModalOpen]);
```
0xdevalias, line 138: This block of code feels convoluted. Can we just roll the logic into the open/close `useCallback`s above instead?

src/components/molecules/VenueJoinButton/VenueJoinButton.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { getCurrentTimeInUTCSeconds, getTimeBeforeParty } from "utils/time";
+import { setLocalStorageToken } from "utils/localStorage";
+import { isTruthy } from "utils/types";
+import { joinVenue } from "utils/url";
+
+import "./VenueJoinButton.scss";
+
+export interface VenueJoinButtonProps {
+  venueId?: string;
+  venue: AnyVenue;
```
0xdevalias, line 26: Can we do this?

```suggestion
  venue: WithId<AnyVenue>;
```

0xdevalias: If so, there is a `withId` helper that can be used on the calling side to construct the appropriately typed object.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -50,6 +53,9 @@ export interface VenueLandingPageProps {
 
 export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> = () => {
   const venueId = useVenueId();
+  const { user } = useUser();
+
+  const isLoggedIn = !!user;
   useConnectCurrentVenue();
 
   const venue = useSelector(currentVenueSelectorData);
```
0xdevalias, line 61: Does this include the `venue.id`? If not, can we switch to a different selector that does, then probably wouldn't need to call `useVenueId` above.

0xdevalias (changes requested):
Questions/comments/cleanup/etc

0xdevalias:
@denisdimitrov It seems my last review on this was 23 days ago, can we please get back to this PR when possible.

src/pages/VenueLandingPage/VenueLandingPage.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1148/commits/5e9e39355ce983ca1db81c1af798117439fbca80

src/pages/VenueLandingPage/VenueLandingPage.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1148/commits/5e9e39355ce983ca1db81c1af798117439fbca80

src/pages/VenueLandingPage/VenueLandingPage.tsx
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1148/commits/6f64cb54d9d168ab6365655b2e1169886fc48b1e

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -56,9 +53,8 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
   const { user } = useUser();
 
   const isLoggedIn = !!user;
-  useConnectCurrentVenue();
 
-  const venue = useSelector(currentVenueSelectorData);
+  const { currentVenue: venue } = useConnectCurrentVenueNG(venueId);
   const venueRequestStatus = useSelector(
     (state) => state.firestore.status.requested.currentVenue
   );
```
0xdevalias, line 60: @mike-lvov Note that by changing to `useConnectCurrentVenueNG` here, the not commonly used `venueRequestStatus` pattern below it will likely be invalid now.

denisdimitrov:
@0xdevalias Can we close this PR? I think you or someone else already implemented something like this which is kind of working in our code base.

denisdimitrov:
Closing because the implementation is old and the approach is not 100% correct for this functionality. All of the patterns are legacy and this was never required.

PR #1150, opened 2021-02-10
[old branch] track open tabs and remove location only on last one

0xdevalias:
@sparkletown/engineering I noticed this sitting in our old branches when I was cleaning them up. It might not be directly useful anymore, but it also sounds like it could be a good idea, so wanted to bring it up here for discussion. If we don't need it, we can close this PR + delete the branch.

@denisdimitrov this seems to be one of your old branches

@mike-lvov you've been working on counting related things most recently

### DIFF
--- a/src/components/organisms/AppRouter/AppRouter.jsx
+++ b/src/components/organisms/AppRouter/AppRouter.jsx
@@ -33,6 +33,8 @@ import { venueLandingUrl } from "utils/url";
 import { RoomsForm } from "pages/Admin/Venue/Rooms/RoomsForm";
 import { SchedulePage } from "pages/Schedule/SchedulePage";
 
+const STORAGE_TABS_KEY = "TABS_OPEN";
+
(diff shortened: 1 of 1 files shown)


denisdimitrov:
After the new implementation of room counting and location tracking introduced by Mike. I don't think this is necessary.

0xdevalias:
> After the new implementation of room counting and location tracking introduced by Mike. I don't think this is necessary.

@denisdimitrov Sounds good.

@mike-lvov Can you please confirm? If so, we can close this + delete the branch.

0xdevalias:
> I don't think this particular PR is useful, but the problem that is addressed here exists even now. One of the ways to overcome it would be to enable only one instance of the app running at any point in time.

@mike-lvov That would be one way to resolve it. This would be another.

You said it still exists, is there an existing backlog issue tracking this by chance? And if so, are you able to link it here?

mike-lvov:
Unrelevant PR

PR #1151, opened 2021-02-10
[old branch] add video chat to partymap

0xdevalias:
@sparkletown/engineering I noticed this sitting in our old branches when I was cleaning them up. It might not be directly useful anymore, but it also sounds like it could be a good idea, so wanted to bring it up here for discussion. If we don't need it, we can close this PR + delete the branch.

@denisdimitrov this seems to be one of your branches

---

From a quick skim, my guess is that this might have been related to the 1-on-1 video chat feature that used to exist within the Playa template, and perhaps aiming to bring that to our current templates.

The code itself is likely outdated, but there may be aspects we can cherry pick/use with our current code to get this landed?

possibly fixes https://github.com/sparkletown/internal-sparkle-issues/issues/24

### DIFF
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -115,7 +115,7 @@ const NavBar: React.FunctionComponent<PropsType> = ({ redirectionUrl }) => {
     [radioStations]
   );
 
-  const { volume, setVolume } = useRadio(sound);
+  const { volume, setVolume } = useRadio(IS_BURN ? sound : undefined);
 
   const radioPopover = (
     <Popover id="radio-popover">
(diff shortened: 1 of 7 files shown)


denisdimitrov:
This was copy pasted from Playa and it wasn't working properly because of the implementation. Even if there is a small logic that can be reused here, I'd suggest not using this. I believe this is going to be a feature in the future but I'd write it from scratch.

0xdevalias:
> This was copy pasted from Playa and it wasn't working properly because of the implementation. Even if there is a small logic that can be reused here, I'd suggest not using this. I believe this is going to be a feature in the future but I'd write it from scratch.

@denisdimitrov Fair enough, makes sense. Lets close this one and remove the branch then. We can always refer back to it during the re-implementation if we need to.

mike-lvov:
Agreed with @denisdimitrov

PR #1156, opened 2021-02-11
Fixes announcement bar styles

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/330
closes https://github.com/sparkletown/internal-sparkle-issues/issues/329

### DIFF
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -6,11 +6,12 @@
   position: absolute;
   background: $primary;
   top: 66px;
-  left: 50%;
+  // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
+  left: 40%;
   width: 500px;
   margin: 0 auto 0 -250px;
 
-  padding: 20px 35px 40px 35px;
+  padding: 15px;
   border-radius: 0 0 50px 50px;
   text-align: center;
 
@@ -18,7 +19,7 @@
     cursor: pointer;
     position: absolute;
     top: 10px;
-    right: 10px;
+    right: 15px;
     :hover {
       color: var(--primary-color);
       text-decoration: underline;
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -7,7 +7,7 @@
   background: $primary;
   top: 66px;
   // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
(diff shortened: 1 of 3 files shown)


src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -6,19 +6,20 @@
   position: absolute;
   background: $primary;
   top: 66px;
-  left: 50%;
+  // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
+  left: 40%;
```
0xdevalias, line 10: Is it possible to set a variable for the sidebar/map %'s, and then calculate the value here by using that variable divide 2?

mike-lvov: I think yes, should take not much time, but the sidebar's width/logic will change in https://github.com/sparkletown/sparkle/pull/1074

0xdevalias (approved):
My previous comment isn't really blocking, so while it would be nice to have, not worth holding approval over.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -6,19 +6,20 @@
...
   width: 500px;
```
denisdimitrov, line 11: You say it takes 20%, how about this width value?

denisdimitrov: Looks like it has 500px width and -250 margin.

mike-lvov: I believe this is a technique to make the bar centered. This width + margin styles set the `moving point` of the bar to the center of the bar

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1156/commits/d46487a0e6597ec71501e136b1822c8586897a0b

denisdimitrov (approved):
LGTM!

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -6,19 +6,20 @@
   position: absolute;
   background: $primary;
   top: 66px;
-  left: 50%;
+  // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
```
0xdevalias, line 9: @mike-lvov What does this '40%' here refer to? It doesn't seem to align with the code? Is it outdated/can it be deleted?

PR #1159, opened 2021-02-11
fix account creation when venue access is missing

denisdimitrov:
The guards were checking if result exists and were preventing the logic from finishing. Result is missing when venue access is not set and this is expected. Token won't be set if the venue access is missing (open) because everyone can join the party/event. Password on the other hand is tricky and needs further discussion. Because when you are creating an account, you don't have an input field for venue password and this might be a problem. The UX is the following:
Create account -> /in/ -> No Access -> /v/ -> Type password

We should probably redirect people to /v/ instead of /in/ when `venue.access` exists.

Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/338

### DIFF
--- a/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
+++ b/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
@@ -3,15 +3,12 @@ import firebase from "firebase/app";
 import { useForm } from "react-hook-form";
 import { useHistory } from "react-router-dom";
 
-import { checkAccess } from "api/auth";
-
 import { SPARKLE_TERMS_AND_CONDITIONS_URL } from "settings";
 
 import { VenueAccessMode } from "types/VenueAcccess";
 
 import { venueSelector } from "utils/selectors";
 import { isTruthy } from "utils/types";
-import { setLocalStorageToken } from "utils/localStorage";
 
 import { useSelector } from "hooks/useSelector";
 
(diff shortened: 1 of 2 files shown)


src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -114,13 +114,13 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
           break;
       }
 
-      if (!result) return;
-
-      if (result.data === false) {
+      if (result?.data === false) {
```
mike-lvov, line 117: What if it results in `undefined`? The check will not pass and it will not throw an error

denisdimitrov: It won't because of the typescript optional parameter `result?.data` if result is `null` or `undefined` this statement will fallback to `false`

mike-lvov: It will not fallback, since you have `===` without an implicit type conversion

denisdimitrov: Actually I decided to remove the logic because this is the not correct place for it. Login and register don't have the required fields for the access check and this should be done in the venue landing page only.

mike-lvov (approved):
LGTM

PR #1160, opened 2021-02-11
Adds an unlocking clearing function on leave

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/328

### DIFF
--- a/src/components/molecules/TableHeader/TableHeader.tsx
+++ b/src/components/molecules/TableHeader/TableHeader.tsx
@@ -43,9 +43,9 @@ const TableHeader: React.FC<TableHeaderProps> = ({
   const numberOfUsersAtCurrentTable = usersAtCurrentTable.length;
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const firestoreUpdate = (doc: string, update: any) => {
+  const firestoreUpdate = async (doc: string, update: any) => {
     const firestore = firebase.firestore();
-    firestore
+    await firestore
       .doc(doc)
       .update(update)
       .catch(() => {
@@ -79,6 +79,12 @@ const TableHeader: React.FC<TableHeaderProps> = ({
 
   const leaveSeat = useCallback(async () => {
     if (!user || !profile) return;
+
+    // NOTE: Unlock the table, if you are the last leaving this table
+    if (numberOfUsersAtCurrentTable === 1) {
+      onLockedChanged(seatedAtTable, false);
+    }
+
     const doc = `users/${user.uid}`;
     const existingData = profile.data;
     const update = {
@@ -91,11 +97,6 @@ const TableHeader: React.FC<TableHeaderProps> = ({
       },
     };
     await firestoreUpdate(doc, update);
-
-    // NOTE: Unlock the table, if you are the last leaving this table
-    if (numberOfUsersAtCurrentTable === 0) {
-      onLockedChanged(seatedAtTable, false);
-    }
 
     setSeatedAtTable("");
   }, [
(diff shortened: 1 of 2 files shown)


0xdevalias:
Since this runs in the user’s frontend, what happens if they are the last user on the table, and then their browser crashes/Internet is disconnected/etc?

Not sure how the security rules are on this to allow updates, but perhaps the logic could check if anyone is on a table, and if not unlock it/let people in?

The ideal situation would probably be for this to be a backed triggered function or similar I think, but we don’t currently have patterns for that.

denisdimitrov:
By a quick look at issue, it looks like the problem is that when you enter a room it's locked. But when you leave it it's unlocked. Which indicates that there is a logic that unlocks it when you leave but also logic that locks it when you enter.

denisdimitrov:
What Glenn said makes sense, we should probably check if you are the first to go in and unlock the table. And people should be able to enter empty locked tables.

mike-lvov:
@0xdevalias @denisdimitrov 
People can come into the empty locked tables now

```
  const tableLocked = (table: string) => {
    // Empty tables are never locked
    if (
      recentVenueUsers.filter(
        (user: User) => user.data?.[venueName]?.table === table
      ).length === 0
    ) {
      return false;
    }
    // Locked state is in the experience record
    return allTables?.[table]?.locked;
  };
```

You leave your seat on `beforeunload` event

```
  useEffect(() => {
    window.addEventListener("beforeunload", leaveSeat);
    return () => {
      window.removeEventListener("beforeunload", leaveSeat, false);
    };
  }, [leaveSeat]);
```

denisdimitrov:
> Since this runs in the user’s frontend, what happens if they are the last user on the table, and then their browser crashes/Internet is disconnected/etc?

How about this case?

mike-lvov:
Yeah, I do agree that it works like shit and doesn't handle many usecases. I'm just wondering about what I can do to fix it so it works `good enough` for the upcoming event and demo. The best way to handle it would be the `backendish approach` by Glenn, but I don't think I'll be able to do it in a short period of time. I did my best to handle the most common usecase

@0xdevalias @denisdimitrov

mike-lvov:
Ah, I see. There is a suggestion to unlock it when you enter. Okay, will do it this way then

src/components/molecules/TableHeader/TableHeader.tsx
```diff
@@ -149,10 +168,8 @@ const TableHeader: React.FC<TableHeaderProps> = ({
           <label className="switch">
             <input
               type="checkbox"
-              checked={!!tableLocked(seatedAtTable)}
-              onChange={() =>
-                onLockedChanged(seatedAtTable, !tableLocked(seatedAtTable))
-              }
+              checked={!!isCurrentTableLocked}
+              onChange={() => setIsCurrentTableLocked(!isCurrentTableLocked)}
```
denisdimitrov, line 172: You can call `onChange={setIsCurrentTableLocked}` directly and use `isCurrentTableLocked` without passing it as a parameter because it's defined at the top of the component.

denisdimitrov (changes requested):

src/components/molecules/TableHeader/TableHeader.tsx
mike-lvov: I don't think I can, since `setIsCurrentTableLocked` is not a toggle fn. It is used as `setIsCurrentTableLocked(false)` as well

denisdimitrov: Just noticed that you need it for the `useEfffect` but in this case `onChange` won't memo the function because of the arrow syntax. Maybe you can make the parameter optional.

denisdimitrov (approved):
LGTM!

PR #1165, opened 2021-02-12
Admin - Fix for venue edit page in legacy admin

denisdimitrov:
The form validation was not passing because owners was a required field in the validation and it was not submitting the form.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/343

### DIFF
--- a/src/pages/Admin/Venue/DetailsForm.tsx
+++ b/src/pages/Admin/Venue/DetailsForm.tsx
@@ -81,21 +81,27 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
   const queryParams = useQuery();
   const parentIdQuery = queryParams.get("parentId");
 
-  const { watch, formState, register, setValue, ...rest } = useForm<FormValues>(
-    {
-      mode: "onSubmit",
-      reValidateMode: "onChange",
-      validationSchema: validationSchema,
-      validationContext: {
-        template: state.templatePage?.template,
-        editing: !!venueId,
-      },
-      defaultValues: {
-        ...defaultValues,
-        parentId: parentIdQuery ?? defaultValues?.parentId ?? "",
-      },
-    }
-  );
+  const {
+    watch,
+    formState,
+    register,
+    setValue,
+    control,
+    handleSubmit,
+    errors,
+  } = useForm<FormValues>({
+    mode: "onSubmit",
+    reValidateMode: "onChange",
+    validationSchema: validationSchema,
+    validationContext: {
+      template: state.templatePage?.template,
+      editing: !!venueId,
+    },
+    defaultValues: {
+      ...defaultValues,
+      parentId: parentIdQuery ?? defaultValues?.parentId ?? "",
+    },
+  });
   const { user } = useUser();
   const history = useHistory();
   const { isSubmitting } = formState;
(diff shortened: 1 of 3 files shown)


src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -204,10 +203,10 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
               isSubmitting={isSubmitting}
               register={register}
               watch={watch}
-              {...rest}
-              onSubmit={onFormSubmit}
+              onSubmit={onSubmit}
               editing={!!venueId}
               formError={formError}
+              {...rest}
```
mike-lvov, line 209: You do aware of `rest` overriding all of that, right? Will it cause any problems?

mike-lvov (approved):
If tested, looks good

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: I'll check if I can remove it, it seemed unnecessary on first glance.

0xdevalias: (for context, when it was at the start of the props, the ones set afterwards would override it, but now that it’s moved to the end, it will override all the other props)

0xdevalias (commented):
Do we know why there were no validation errors being shown before? (At least that I remember seeing), and does this fix that aspect too? (doesn’t look like it will from a quick skim of the diff)

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -204,10 +209,12 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
...
               watch={watch}
-              {...rest}
-              onSubmit={onFormSubmit}
+              onSubmit={onSubmit}
               editing={!!venueId}
               formError={formError}
+              control={control}
+              handleSubmit={handleSubmit}
+              errors={errors}
             />
```
mike-lvov, line 218: Now that you have removed `...rest` at all, are we sure that we pass every needed prop there? Did you go one-by-one to check what `rest` object contained?

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: > (for context, when it was at the start of the props, the ones set afterwards would override it, but now that it’s moved to the end, it will override all the other props)

Yup, it was kinda in the middle, overriding only half of them

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: I didn't check the rest object, I just checked the expected Props in the component and passed only the ones being used/necessary.

denisdimitrov:
> Do we know why there were no validation errors being shown before? (At least that I remember seeing), and does this fix that aspect too? (doesn’t look like it will from a quick skim of the diff)

It was a change of mine, I changed owners to be no longer optional in the VenueTypes, but I also changed it in the validation which is wrong.

src/pages/Admin/Venue/DetailsForm.tsx
mike-lvov: Is it possible that rest had some `optional` props that were used and now we are not passing them?

denisdimitrov: Nope, I checked the interface of the component and every prop is being passed. If the `rest` object had any other fields that means they were not being used by the component.

0xdevalias:
> There was no validation error because there is no owners field in that form.

That makes sense, though I guess I wonder how we can have these errors show up in a ‘generic’ field or somewhere else (eg. bugsnag?) or otherwise, rather than failing silently like it was.

PR #1166, opened 2021-02-12
render rooms only when they are enabled

denisdimitrov:
The rooms were always render without the isEnabled check in place. How this should work correctly.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/346

### DIFF
--- a/src/components/templates/PartyMap/components/Map/Map.tsx
+++ b/src/components/templates/PartyMap/components/Map/Map.tsx
@@ -212,20 +212,17 @@ export const Map: React.FC<MapProps> = ({
     setSelectedUserProfile,
   });
 
-  const roomOverlay = useMemo(
-    () =>
-      venue.rooms?.map((room) =>
-        room.isEnabled ? (
-          <MapRoom
-            key={room.title}
-            venue={venue}
-            room={room}
-            selectRoom={() => selectRoom(room)}
-          />
-        ) : null
-      ),
-    [selectRoom, venue]
-  );
+  const roomOverlay = useMemo(() => {
+    const enabledRooms = venue?.rooms?.filter((room) => room.isEnabled) ?? [];
+    return enabledRooms.map((room) => (
+      <MapRoom
+        key={room.title}
+        venue={venue}
+        room={room}
+        selectRoom={() => selectRoom(room)}
+      />
+    ));
+  }, [selectRoom, venue]);
 
   const gridContainerStyles = useMemo(
     () => ({
--- a/src/components/templates/PartyMap/components/Map/Map.tsx
+++ b/src/components/templates/PartyMap/components/Map/Map.tsx
@@ -14,7 +14,7 @@ import { PartyMapVenue } from "types/venues";
 import { makeUpdateUserGridLocation } from "api/profile";
 
 import { hasElements, isTruthy } from "utils/types";
-import { makeRoomHitFilter } from "utils/filter";
+import { filterEnabledRooms, makeRoomHitFilter } from "utils/filter";
 import { WithId } from "utils/id";
 import { setLocationData } from "utils/userLocation";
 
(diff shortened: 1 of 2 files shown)


src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -214,14 +214,16 @@ export const Map: React.FC<MapProps> = ({
...
     () =>
-      venue.rooms?.map((room) => (
-        <MapRoom
-          key={room.title}
-          venue={venue}
-          room={room}
-          selectRoom={() => selectRoom(room)}
-        />
-      )),
+      venue.rooms?.map((room) =>
```
mike-lvov, line 217: I think a better approach would be to filter them first, based on `isEnabled` prop and then to map without a ternary operator

mike-lvov (changes requested):

0xdevalias:
I’m not sure where/when in the code it changed, but definitely at some point in the past the rooms were able to be disabled, as this gets used fairly regularly. I’m pretty sure it was on the party map, but it’s possible it was on theme camp.

src/components/templates/PartyMap/components/Map/Map.tsx
denisdimitrov: Usually I'd be against this extra iteration, but this array is small and has no impact on the performance.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -212,18 +212,17 @@ export const Map: React.FC<MapProps> = ({
...
-      )),
-    [selectRoom, venue]
-  );
+  const roomOverlay = useMemo(() => {
+    const enabledRooms = venue?.rooms?.filter((room) => room.isEnabled) ?? [];
+    return enabledRooms.map((room) => (
+      <MapRoom
+        key={room.title}
+        venue={venue}
+        room={room}
```
mike-lvov, line 221:
```
return venue?.rooms?
.filter(...)
.map(...)
```

denisdimitrov: Can we avoid chaining please? This won't save any lines of code or anything. Unless you are both up for this I'd rather stay with the named constant.

mike-lvov: [nit], you can merge it now, but this way it looks cleaner and you will actually save a few lines of code by removing `return` + will reduce the cognitive load + we will not create an unnecessary `const`

mike-lvov (approved):
A few nit updates could be done after merging

src/components/templates/PartyMap/components/Map/Map.tsx
denisdimitrov: Okey, I've changed my mind. You can take a look.

src/utils/filter.ts
```diff
@@ -76,20 +64,4 @@ export const makeRoomHitFilter = ({
   return isWithinBounds(checkPercent, roomBounds);
 };
 
-export const filterUnreadPrivateChats = (
```
denisdimitrov, line 79: This is not being used.

denisdimitrov: was*

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: Thanks

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: > Usually I'd be against this extra iteration, but this array is small and has no impact on the performance. +1

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: [/2c] I'm a fan of extracting `const`s when it will make the statement simpler to read/cognitively process/etc, but in this case, agree that it is probably 'cleaner' chained.

PR #1167, opened 2021-02-12
[scripts/get-badges] Get visits based on venue name

thecadams:
There were several users on the OG prod environment (emails available upon request please ask me) who were missing the `enteredVenueIds` key from their user record in Firestore.

This caused a recent event's analytics to misleadingly and incorrectly show as being attended by 10% of the believed attendance.

This change requires you to specify the venue name, but it will filter ALL users and ALL visits for those venue names.

With this, we were able to see a lot more data.

This is part of the solution that resolves https://github.com/sparkletown/internal-sparkle-issues/issues/294

### DIFF
--- a/scripts/get-badges.ts
+++ b/scripts/get-badges.ts
@@ -9,6 +9,7 @@ import { UserVisit } from "../src/types/Firestore";
 import { User } from "../src/types/User";
 
 import { WithId, withId } from "../src/utils/id";
+import { formatSecondsAsDuration } from "../src/utils/time";
 
 import { initFirebaseAdminApp, makeScriptUsage } from "./lib/helpers";
 
@@ -112,7 +113,13 @@ interface UsersWithVisitsResult {
 
   // Display CSV headings
   console.log(
-    ["Email", "Party Name", "Venue Visited", "Time Spent"]
+    [
+      "Email",
+      "Party Name",
+      "Venue Visited",
+      "Time Spent (seconds)",
+      "Time Spent (formatted)",
+    ]
       .map((heading) => `"${heading}"`)
       .join(",")
   );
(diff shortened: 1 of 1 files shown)


scripts/get-badges.ts
```diff
@@ -56,7 +48,6 @@ interface UsersWithVisitsResult {
     await admin
       .firestore()
       .collection("users")
-      .where("enteredVenueIds", "array-contains-any", venueIdsArray)
```
0xdevalias, line 59: Not sure that this is a change we want to be doing here. The edgecase the changes in this PR was required for is definitely the exception, not the norm. So to have to process so much extra data every time is suboptimal.

This should probably only run when we need this 'special mode' or similar.

0xdevalias (changes requested):
Haven't deeply reviewed all of the code here, but wanted to make sure that this blocker was captured in case someone else is the one that reviews it.

scripts/get-badges.ts
```diff
@@ -99,34 +90,41 @@ interface UsersWithVisitsResult {
...
 
   // TODO: filter enteredVenueIds and visitsTimeSpent so that they only contain related venues?
-  const result = usersWithVisits.map((userWithVisits) => {
-    const { user, visits } = userWithVisits;
-    const { id, partyName, enteredVenueIds = [] } = user;
-    const { email } = authUsersById[id] ?? {};
-
-    const visitsTimeSpent = visits.map(
-      (visit) => `${visit.id} (${formatSecondsAsDuration(visit.timeSpent)})`
-    );
```
0xdevalias, line 109: We may want to keep this formatted version of the time spent. Or alternately, include another column that includes this, as it's quite user-friendly data to include.

0xdevalias: https://github.com/sparkletown/sparkle/pull/1167/commits/84b4c1a4a6bc5a9bd4dfd083287e9fac85ccd10f

thecadams:
Guessing this is stale, just doing a cleanup pass on my open PRs across all projects. Going to close but feel free to reopen if needed.

PR #1168, opened 2021-02-12
add conditional banner centering

denisdimitrov:
More info can been seen in the issue about the problem. The fix is pretty straightforward.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/348

### DIFF
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -11,6 +11,10 @@
   width: 500px;
   margin: 0 auto 0 -250px;
 
+  &.center {
+    left: 50%;
+  }
+
   padding: 15px;
   border-radius: 0 0 50px 50px;
   text-align: center;
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
@@ -1,17 +1,22 @@
 import React, { useCallback, useEffect, useState } from "react";
+import classNames from "classnames";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faTimesCircle } from "@fortawesome/free-solid-svg-icons";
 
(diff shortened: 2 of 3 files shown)


mike-lvov (approved):
Would be nice to change namings, but general approach is good. Given the time restrictions, let's fix that later

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -11,6 +11,10 @@
   width: 500px;
   margin: 0 auto 0 -250px;
 
+  &.center {
+    left: 50%;
+  }
```
0xdevalias, line 16: Why this instead of more standard methods such as `margin` `auto`/etc?

- https://css-tricks.com/centering-css-complete-guide/

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
```diff
@@ -27,8 +32,14 @@ export const AnnouncementMessage: React.FC<AnnouncementMessageProps> = ({
...
 
+  const isBannerCentered = template !== VenueTemplate.partymap;
+
   return (
-    <div className="announcement-container">
+    <div
+      className={classNames("announcement-container", {
+        center: isBannerCentered,
+      })}
+    >
```
0xdevalias, line 42: This seems kind of hacky.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
0xdevalias: Which it appears is already being used on line 12, but there is an arbitrary `-250px` on it as well.

My assumption is that that `250px` at one stage represented the width of the sidebar (though doesn't seem to anymore, since we use `20%` in the variable now)

0xdevalias (commented):
[TECH DEBT] Some comments on the implementation here, and how it can be cleaned up.

cc // @denisdimitrov @mike-lvov 

---

Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/377

PR #1170, opened 2021-02-12
Fixes modal image sizing

mike-lvov:
Fixes modal image sizing

### DIFF
--- a/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
+++ b/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
@@ -28,8 +28,8 @@ $spacing: 20px;
     align-items: center;
 
     > img {
-      width: 100%;
-      height: 100%;
+      width: 60%;
+      height: auto;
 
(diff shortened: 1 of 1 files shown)


src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -28,8 +28,8 @@ $spacing: 20px;
     align-items: center;
 
     > img {
-      width: 100%;
-      height: 100%;
+      width: 60%;
+      height: auto;
```
0xdevalias, line 32: Instead of setting an arbitrary fixed % here, it would be good to see if the methods described in the following (or similar) would work, and if so, clean it up:

- https://github.com/sparkletown/internal-sparkle-issues/issues/344#issuecomment-778412961
- https://github.com/sparkletown/internal-sparkle-issues/issues/344#issuecomment-778419761

0xdevalias (commented):
Potential future cleanup.

PR #1164, opened 2021-02-12
refactor scss z-index's to use z() helper + fix user avatar 'sidebar' layering issues

0xdevalias:
The user avatar image was showing above the various 'sidebar-like' components. 

- refactored all scss `z-index` to use our `z()` helper (to centralise this logic, and reduce future layering issues)
- add `@debt` comments to `z-index` locations within `styled-components` (until we can 💥  all of the `styled-components` in favour of plain scss)
- update all of our 'sidebar-like' layers to use `z-index` `15`
  - fixes https://github.com/sparkletown/internal-sparkle-issues/issues/339

## Screenshots of bug before fix

image
image

PR #1172, opened 2021-02-12
fix z-index layering issues with NavBar, LiveSchedule, and SideBar

0xdevalias:
This is a regression based on the changes I made to fix the z-index layering for the avatars in sparkletown/sparkle#1164:

- When searching for people in the `NavBar`, the search results list is 'behind' the `Sidebar`, so you can't see your results
  - fixes https://github.com/sparkletown/internal-sparkle-issues/issues/352
- When viewing the Live Schedule, it is shown 'behind' the `Sidebar`, so the details get cut off
  - fixes https://github.com/sparkletown/internal-sparkle-issues/issues/357
- When viewing the Live Schedule, the user's avatar was incorrectly shown on a layer 'above' it

The fixes here are, generally speaking, to ensure that each element is on an appropriate `z-index` layer, and that they are in the correct 'stacking context'. It's also worth noting that a child component can never 'exceed' its parent component's `z-index` layer, which was one of the bugs here (even though it had a higher `z-index`, its parent was 'keeping it down')

Further reading:

- https://www.freecodecamp.org/news/4-reasons-your-z-index-isnt-working-and-how-to-fix-it-coder-coder-6bc05f103e6c/

### DIFF
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -344,14 +344,14 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
         </div>
       </header>
 
-      <SchedulePageModal isVisible={isEventScheduleVisible} />
-
       <div
         className={`schedule-dropdown-backdrop ${
           isEventScheduleVisible ? "show" : ""
         }`}
         onClick={hideEventSchedule}
-      />
+      >
+        <SchedulePageModal isVisible={isEventScheduleVisible} />
+      </div>
 
       {venue?.parentId && parentVenue?.name && (
         <div className="back-map-btn">
--- a/src/components/molecules/NavBar/playa.scss
+++ b/src/components/molecules/NavBar/playa.scss
@@ -16,9 +16,10 @@ $page-max-width: 1240px;
 
 $sand: #937c63;
 
+// @debt Refactor the relevant parts of this file into ./NavBar.scss
 .navbar.navbar_playa {
   position: fixed;
-  z-index: z(header);
+  z-index: z(navbar);
   top: 0;
   left: 0;
   width: 100%;
--- a/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
+++ b/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
@@ -82,7 +82,10 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
         <li
           key={formatDate(day.dateDay.getTime())}
           className={`button ${idx === date ? "active" : ""}`}
-          onClick={() => setDate(idx)}
+          onClick={(e) => {
+            e.stopPropagation();
+            setDate(idx);
+          }}
         >
           {formatDateToWeekday(day.dateDay.getTime() / 1000)}
         </li>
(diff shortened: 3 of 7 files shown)


src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -344,14 +344,14 @@ const NavBar: React.FC<NavBarPropsType> = ({ redirectionUrl }) => {
...
-
       <div
         className={`schedule-dropdown-backdrop ${
           isEventScheduleVisible ? "show" : ""
         }`}
         onClick={hideEventSchedule}
-      />
+      >
+        <SchedulePageModal isVisible={isEventScheduleVisible} />
+      </div>
```
0xdevalias, line 354: [FYI] I nested these so that they aren't in the same 'stacking context', which simplifies our `z-index` layering.

[Future Cleanup] We can probably remove the explicit `z-index` layer on the `SchedulePageModal` component now I think, since I believe it will inherit from `.schedule-dropdown-backdrop`'s layer now.

src/components/molecules/NavBar/playa.scss
```diff
@@ -16,9 +16,10 @@ $page-max-width: 1240px;
 
 $sand: #937c63;
 
+// @debt Refactor the relevant parts of this file into ./NavBar.scss
```
0xdevalias, line 19: [Future Cleanup] `@debt` Refactor the relevant parts of this file into ./NavBar.scss

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -82,7 +82,10 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
         <li
           key={formatDate(day.dateDay.getTime())}
           className={`button ${idx === date ? "active" : ""}`}
-          onClick={() => setDate(idx)}
+          onClick={(e) => {
+            e.stopPropagation();
+            setDate(idx);
+          }}
```
0xdevalias, line 88: [FYI] `e.stopPropagation` is required here to stop the click event 'bubbling higher', and triggering the `hideEventSchedule` `onClick` on the `.schedule-dropdown-backdrop` `div` in `src/components/molecules/NavBar/NavBar.tsx`

src/scss/constants.scss
```diff
@@ -47,6 +47,8 @@ $room-info-bg: #005ee5;
 //   pages/Admin/Room/Modal/RoomModal.styles.ts (z-index: 101)
 //   components/molecules/UserProfilePicture/UserProfilePicture.styles.ts (z-index: 10)
 //   components/atoms/Legend/Legend.styles.ts (z-index: 5)
+$z-layer-navbar: 30;
+$z-layer-live-schedule: 20;
 $z-layer-sidebar: 15;
```
0xdevalias, line 52: [FYI] Starting to build up our layers. Ideally we would have them separated by at least increments of `10` for our 'main layers', and then we can use the digits in between for 'micro adjustments' as required.

[Future Cleanup] Eventually we can give these a better generic name based on the 'type of layer' it is, and probably reduce the number of 'specific names' we have in our `z()` helper. I expect we could also reduce the number of places where we even need to explicitly set a `z-layer` as we refactor/cleanup our code. In most situations I find that needing to use `z-index` is actually a bit of a hack.

0xdevalias (commented):
A few notes.

0xdevalias:
@sparkletown/engineering-leads 🔨 🍷 breaking glass to deploy as this is impacting a current event and there are no other engineers around to review/approve.

PR #1176, opened 2021-02-15
add proper validation to room edit and creation

denisdimitrov:
Bugs fixed:
- You can no longer create a room without logo.
- You can no longer edit a room and leave empty name or url.
- You can no longer edit a room without a logo.
- Newly created rooms are now enabled by deafult.
- Room stays enabled after edit

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/374
Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/375


0xdevalias:
> Newly created rooms are now enabled by default

@denisdimitrov Might want to check with @sparkletown/non-tech-team about that, and/or look at the code history that made it that way. I believe that may have been a deliberate choice that we made to have them not enabled by default.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -32,7 +32,8 @@ export interface RoomSchemaShape {
   title: string;
   venueName?: string;
   url?: string;
-  useUrl: boolean;
+  useUrl?: boolean;
+  image_url: string;
```
0xdevalias, line 36: Why not using `lowerCamelCase` for `image_url`?

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -160,6 +161,17 @@ export const roomCreateSchema = Yup.object().shape<RoomSchemaShape>({
...
+
+export const roomEditSchema = Yup.object().shape<RoomSchemaShape>({
+  title: Yup.string()
+    .required("Room name is required")
+    .min(3, ({ min }) => `Name must be at least ${min} characters`),
+  url: Yup.string()
+    .required("Url is required!")
+    .min(3, ({ min }) => `Url must be at least ${min} characters`),
+  image_url: Yup.string().required("Room image is required"),
 });
```
0xdevalias, line 175: Is there a reason why we can't reuse the same schema stuff that is used on creation? What differs between the two? And why does it differ? Feels like they should both validate with the same code?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -32,13 +33,16 @@ const EditRoom: React.FC<EditRoomProps> = ({
     watch,
     handleSubmit,
     setValue,
+    errors,
     formState: { isSubmitting, dirty },
   } = useForm({
+    validationSchema: roomEditSchema,
```
0xdevalias, line 39: As mentioned elsewhere RE: using the same validation as `roomCreateSchema`

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -32,13 +33,16 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
     defaultValues: {
       title: room.title,
       url: room.url,
       description: room.description,
       tempalte: room.template,
```
0xdevalias, line 44: This seems like a bug/typo?
```suggestion
      template: room.template,
```

src/pages/Admin/Room/Modal/Item/Item.tsx
```diff
@@ -63,6 +64,7 @@ const RoomModalItem: React.FC<RoomModalItemProps> = ({
       const valuesWithTemplate = {
         ...values,
         template,
+        isEnabled: true,
```
0xdevalias, line 67: I'm assuming this is the part I called out in https://github.com/sparkletown/sparkle/pull/1176#issuecomment-779632114 that required further checking?

src/pages/Admin/Room/Modal/Item/Item.tsx
```diff
@@ -97,7 +99,7 @@ const RoomModalItem: React.FC<RoomModalItemProps> = ({
     }
   }, [editValues, onSubmitHandler, template, useUrl, user, values, venueId]);
 
-  const handleOnChange = (val: string) => setValue("image_url", val, false);
+  const handleOnChange = (val: string) => setValue("image_url", val);
```
0xdevalias, line 102: What does the `false` here do? It's hard to tell given we aren't using named parameters on `setValue`

0xdevalias (commented):
Questions/comments

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: This is legacy code in the firebase, if I change it here. I'll have to change it in another 20-30 places. I don't think this should be done in this PR.

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: When creating a room there is an option to select either new venue and room or simple room with url. That validation fails in  the edit because that field doesn't exist anywhere in the firebase.

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Yup, agreed OOS for here. Just wanted to know the 'why'

src/pages/Admin/Room/Modal/Item/Item.tsx
denisdimitrov: Correct.

src/pages/Admin/Room/Modal/Item/Item.tsx
denisdimitrov: The third parameter was `triggerValidation` and it was set on `false`. And when you upload a new image the validation was not triggering.

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Fair enough. I'm not deeply familiar with yup, but is there a way we can share the same partial validation rules between the 2? It looks like it's just an object, so perhaps we can move the object outside of the `.shape()` call, and then use the common bits from that to build the 2 separate validations.

src/pages/Admin/Room/Modal/Item/Item.tsx
0xdevalias: Cool. Probably worth leaving this open until we have a resolution one way or the other.

src/pages/Admin/Room/Modal/Item/Item.tsx
0xdevalias: Makes sense.

src/pages/Admin/Room/Edit/RoomEdit.tsx
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1176#discussion_r576619116

denisdimitrov:
@0xdevalias 
> @denisdimitrov Might want to check with @sparkletown/non-tech-team about that, and/or look at the code history that made it that way. I believe that may have been a deliberate choice that we made to have them not enabled by default.

I just got verification from Ed that rooms are supposed to be enabled. That's how it is in the legacy code.

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: I created some constants that are being shared between them.

src/pages/Admin/Room/Modal/Item/Item.tsx
denisdimitrov: https://github.com/sparkletown/sparkle/pull/1176#issuecomment-779703122

jess-sparkles:
To answer the above question, having rooms enabled right after creating is great! We asked for this back in December so thanks for fixing it!

0xdevalias:
> We asked for this back in December so thanks for fixing it!

@jess-sparkles Interesting. I don't believe that ever flowed through to me (or I forgot about it since if it did). Good to know that this will solve that for you though :)

Is that a universal want/need? Or is it something that it would be useful to have a toggle option on this form (that defaults to show)? (as my memory was that we made them hidden by default by explicit request, as per https://github.com/sparkletown/sparkle/pull/1176#issuecomment-779632114)

jess-sparkles:
Seems like on Slack @Sparkle-T wants them to be turned off by default (so no change). @sofisparkle and I wanted them turned on by default as we didn't think it was obvious that the rooms were turned off until you go back to the admin page (after making the room) and we thought this would be confusing for users.

0xdevalias:
> Seems like on Slack @Sparkle-T wants them to be turned off by default (so no change). @sofisparkle and I wanted them turned on by default as we didn't think it was obvious that the rooms were turned off until you go back to the admin page (after making the room) and we thought this would be confusing for users.

@jess-sparkles By my read of his @Sparkle-T's [comments on slack](https://doesitsparkle.slack.com/archives/C01B7MSMA3W/p1613554268235800?thread_ts=1613524973.230900&cid=C01B7MSMA3W) it sounded like he was agreeing with you both, that there is no reason to have them default to off since most events are created/set up in advance (though noted that having the ability to configure that in future might be a nice feature, for those 'edge cases' where someone may want to change the default)

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -156,10 +163,15 @@ export const roomCreateSchema = Yup.object().shape<RoomSchemaShape>({
...
-      .required("Url is required!")
-      .min(3, ({ min }) => `Url must be at least ${min} characters`),
+    then: roomUrlSchema,
   }),
+  image_url: roomImageUrlSchema,
+});
+
+export const roomEditSchema = Yup.object().shape<RoomSchemaShape>({
+  title: roomTitleSchema,
+  url: roomUrlSchema,
```
0xdevalias, line 173: In `roomCreateSchema`, this is only present within the `.when("useUrl")`. Is that logic also required here?

0xdevalias (approved):
[CRO] One small question/possible refinement, but aside from that I think this looks good to go (assuming it's been properly clicked through/tested/etc)

---

**Edit:** Actually @denisdimitrov, 1 more question, I never saw a reply to https://github.com/sparkletown/internal-sparkle-issues/issues/375#issuecomment-778905257 confirming whether it is even actually a bug, or if it's 'feature as desired/expected'. Would be good to confirm this before merging:

> @0xdevalias:
> I'm not 100% sure that this is a bug per-se. Didn't we want the ability to be able to not have to explicitly upload an image on all of these sorts of things, and if we didn't, that we would use a default image instead? (cc // @sparkleed @sparkletown/product)

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: It is not because that field is being used only to determine if it should validate url or venue name, when editing a room you are always validating the url.

0xdevalias: Cool. That works for me!

PR #1178, opened 2021-02-15
fix room preview save UX and aspect ratio

denisdimitrov:
- Lock image aspect ratio when resizing.
- Persist the room size when dragging
- Show indicator for saving the rooms
- Disabled rooms when saving them.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/371


src/components/atoms/Button/Button.styles.ts
```diff
@@ -10,8 +10,8 @@ export const Button = styled.button<ButtonProps>`
   border-radius: 1.375em;
   border: none;
   background-color: #361f6e;
-  background-image: ${({ hasGradient }) =>
-    hasGradient
+  background-image: ${({ hasGradient, disabled }) =>
+    hasGradient && !disabled
```
0xdevalias, line 14: [NB] The sooner we can remove `styled-components` the happier I will be

src/pages/Account/Venue/VenueMapEdition/DraggableSubvenue.tsx
```diff
@@ -66,35 +76,39 @@ export const DraggableSubvenue: React.FC<PropsType> = (props) => {
...
-            }}
-          />
-          <div
-            style={{
-              ...styles.resizeTab,
-              bottom: 0,
-              right: 0,
-            }}
-          />
+        <div ref={!isLoading ? drag : null} style={styles.dragContainer}>
```
0xdevalias, line 79: What are the changes to the `ref` here intended to do?

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -171,12 +174,13 @@ const MapPreview: React.FC<MapPreviewProps> = ({
               height: 100,
             }}
             otherIconsStyle={{ opacity: 0.4 }}
-            // lockAspectRatio
+            lockAspectRatio
```
0xdevalias, line 177: [OOS] Huh.. neat that this already existed!

That means that it would be fairly trivial (in the future) to implement a concept of being able to change the aspect ratio when holding a 'modifier key', as described in https://github.com/sparkletown/internal-sparkle-issues/issues/371#issuecomment-778673339

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -171,12 +174,13 @@ const MapPreview: React.FC<MapPreviewProps> = ({
...
+            isLoading={isLoading}
           />
         )}
 
-        <S.EditButton onClick={handleEditButton}>
-          {isEditing ? "Save layout" : "Edit layout"}
+        <S.EditButton disabled={isLoading} onClick={handleEditButton}>
+          {isLoading ? "Saving..." : isEditing ? "Save layout" : "Edit layout"}
```
0xdevalias, line 183: Perhaps instead of `isLoading` this could be called `isSaving` or similar? Would make the code easier to follow.

The nested ternary here is.. interesting. I don't necessarily hate it.. but I feel like I should for some reason.

0xdevalias (commented):
Most of my comments are NB/etc, there are a couple of questions that depending on the answer may warrant changes.

While this is more or less ready for an approve, I will hold off on giving it for now to see what the answers are.

src/pages/Admin/MapPreview/MapPreview.tsx
0xdevalias: (Might be worth creating a new issue to backlog this part/suggestion?)

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: Yeah, I feel the same way about the nesting here. I'll figure something out to make it more readable.

src/pages/Account/Venue/VenueMapEdition/DraggableSubvenue.tsx
denisdimitrov: When the room positions are being saved, they ref will die and the drag and drop will be disabled. I looked in the library on how to disabled the drag with a prop or ref but looks like it can't be disabled.

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/392

src/pages/Account/Venue/VenueMapEdition/Container.tsx
```diff
@@ -73,9 +74,15 @@ export const Container: React.FC<PropsType> = (props) => {
...
     lockAspectRatio,
+    isSaving,
   } = props;
   const [boxes, setBoxes] = useState<SubVenueIconMap>(iconsMap);
   const [imageDims, setImageDims] = useState<Dimensions>();
+  const [dragBoxId, setDragBoxId] = useState<number>(0);
+
+  const setDragItemId = useCallback((id: number) => {
+    setDragBoxId(id);
+  }, []);
```
0xdevalias, line 85: What does this function add that `setDragBoxId` doesn't already do by itself? Seems like a layer of indirection that adds no value?

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -175,12 +177,13 @@ const MapPreview: React.FC<MapPreviewProps> = ({
...
-            isLoading={isLoading}
+            isSaving={isSaving}
           />
         )}
 
-        <S.EditButton disabled={isLoading} onClick={handleEditButton}>
-          {isLoading ? "Saving..." : isEditing ? "Save layout" : "Edit layout"}
+        <S.EditButton disabled={isSaving} onClick={handleEditButton}>
+          {isSaving && <div>Saving...</div>}
+          {!isSaving && <div>{editButtonText}</div>}
```
0xdevalias, line 186: [NB] This could stay as a ternary now that you've separated the `editButtonText` part out of it.

0xdevalias (approved):
[CRO] A couple of minor comments, probably worth cleaning up the function one, the ternary one isn't as important.

But from there, this seems good to go (assuming it's been tested/clicked through/etc)

src/pages/Account/Venue/VenueMapEdition/Container.tsx
denisdimitrov: It's using `useCallback` that's it.

0xdevalias: `setDragBoxId` comes from `useState`, which is already a stable function reference. The only time wrapping it with `useCallback` makes sense is when we're 'closing around the state' to make a 'no arguments' function. But since we still have the `id` arg in the `useCallback` that is just passed through, there's no benefit to having this here.

We can remove this chunk of code and just call/pass `setDragBoxId` to the appropriate places.

PR #1184, opened 2021-02-16
Fix banner text and button overlap

malaikaaiyar:
This closes sparkletown/internal-sparkle-issues#358

Before: 
Screen Shot 2021-02-16 at 8 58 25 PM

After: 
Screen Shot 2021-02-16 at 8 59 20 PM


src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -1,5 +1,4 @@
 @import "scss/constants.scss";
-
```
0xdevalias, line 2: Probably good to keep this space (this same comment goes for all of the lines where whitespace was removed)

0xdevalias (commented):
It's a minor/nit type thing, but would be good to get the whitespace that was lost here restored before merging this.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
malaikaaiyar: I'll fix it, but I didn't actually delete the whitespace? Do you know how that could have happened?

0xdevalias: It's possible that your code editor/IDE is configured to automatically reformat things in a certain style. That's usually where these issues tend to come from.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,7 +15,7 @@
     left: 50%;
   }
 
-  padding: 15px;
+  padding: 15px 30px 15px 20px;
```
0xdevalias, line 18: It would be good to use some sass `$` variables here, and pass that to the `.close-button` `right` attribute so that we can guarantee they will stay 'in sync'.

Looking at the screenshot in https://github.com/sparkletown/sparkle/pull/1184#issue-574472150 it feels almost 'too spacey'. The overly rounded corners we use throughout our styling at the moment make this somewhat challenging to make look good IMO, but perhaps we could reduce the spacing somewhat as well?

Playing around locally:

- top: `10px` (shared variable with `.close-button`)
- bottom: `10px` (can probably use the same variable, but doesn't need to be applied to `.close-button`'s bottom padding
- (i forget if there were other bits and pieces here as I switched to playing with the below)

If it was me doing things, I would probably even go as far as changing the `border-radius` from `50px` down to like `20px` or so. (you would want to check with @sparkletown/product and/or @sparkletown/design for that), which would then allow the styles to be refined even better:

- `.close-button` could have top `10px`, right `7.5px`
- `.announcement-container` would have `padding: 10px 0px`
- we'd add a new `div` that wraps the text (but not the close icon?) and add styles to set margin left/right `30px`
  - (honestly not sure why 30px.. with the close icon positioned `10px` from the right, I would have expected a right margin of `15px` would give sufficient space, but if you look at the orange areas in the screenshot below (margins) when set to `30px` it seems to only just be enough space.)
    - image
  - (It might have to do with the width of the close icon, since that seems to be `20px`, so adding the `7.5px` right position, that would be `27.5px`; so `30px` ensures we have at least `2.5px` of space between the edge of the text, and the close icon)
    - image

The logic in `src/utils/getLinkFromText.tsx` that wraps every word in a `span` makes it hard to play around with this in the Chrome DevTools. That could be changed to use `React.Fragment` (on line 22), which would then unclutter things and make it easier to play with.

Further Reading:

- https://sass-lang.com/documentation/variables
- https://sass-lang.com/documentation/operators/numeric
- https://sass-lang.com/documentation/syntax/special-functions#calc-clamp-element-progid-and-expression
  - I think the numeric operators above should be sufficient, but if not, this reference may be useful
  
And for my reference, since I always forget the order/which it refers to:

- https://www.w3schools.com/css/css_padding.asp
  - 4: top, right, bottom, left
  - 3: top, right/left, bottom
  - 2: top/bottom, right/left
  - 1: top/bottom/right/left

0xdevalias: (which was way more detail/thought than I intended to put into that.. take the specifics with a grain of salt, but the main intent here was to reduce the 'overly padded' feeling a bit, and to use variables to share various widths/etc used to calculate the numbers we use (so that it's clear how they were calculated + will auto update themselves if we change any of the 'parts of the calculation' in future)

0xdevalias (commented):
Some thoughts/comments (that got a little rabbit-holey there at one point)

Delegating to @mike-lvov for the final calls on this (taking into account what i've shared here, as I feel like there would be some good cleanup bits based on what I said, but perhaps they could be considered OOS)

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
malaikaaiyar: Should I also change the class names to follow the BEM approach?

malaikaaiyar: 10px on the top and bottom looks really awkward when the banner is only 1 line, so I'm leaving it at 15px. 

10px:
Screen Shot 2021-02-19 at 10 05 00 AM

15px:
Screen Shot 2021-02-19 at 10 05 46 AM

malaikaaiyar: I think 15px looks good even when it's two lines:
Screen Shot 2021-02-19 at 10 06 32 AM

0xdevalias: I’m ok with 15px

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +15,23 @@
     left: 50%;
   }
 
-  padding: 15px;
-  border-radius: 0 0 50px 50px;
+  padding: $announcement-container-padding;
```
0xdevalias, line 18: If this padding is specifically just for this file (unlikely to be shared elsewhere) it’s fine for it to be defined at the top of the file, rather than in the ‘global’ constants file

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +15,23 @@
...
     position: absolute;
-    top: 10px;
-    right: 15px;
+    top: 15px;
+    right: 7.5px;
     :hover {
       color: var(--primary-color);
       text-decoration: underline;
     }
+    margin-right: 7.5px;
```
0xdevalias, line 31: Does this have any effect, since it’s positioned absolutely?

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +15,23 @@
...
   }
 }
+
+.announcement-container__text {
```
0xdevalias, line 35: I’d probably call this announcement-message (rather than container) to align with the component name. I find myself not using the word container much/at all when I’m writing BEM, as it feels like adds nothing.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
malaikaaiyar: Yes
Without: 
Screen Shot 2021-02-19 at 6 26 46 PM
(the x is a little bit too far to the right)

malaikaaiyar: But instead of having a margin, it looks the same if I make ```right```  15px

0xdevalias: Let’s do the right 15px, as that’s more obvious/clear

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -1,4 +1,5 @@
 @import "scss/constants.scss";
+$announcement-container-padding: 15px 0px;
```
0xdevalias, line 2: [nit]
```suggestion

$announcement-container-padding: 15px 0px;
```

0xdevalias: I would probably also make this just `15px` here in the variable, and maybe just call it `$spacing` or something else generic/short like that. Then we can reuse it in all of the 'spacing' related places within this file that need to be the same space.

```suggestion

$spacing: 15px;
```

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +16,22 @@
     left: 50%;
   }
 
-  padding: 15px;
-  border-radius: 0 0 50px 50px;
+  padding: $announcement-container-padding;
```
0xdevalias, line 19:
```suggestion
  padding: $spacing 0;
```

or

```suggestion
  padding-top: $spacing;
  padding-bottom: $spacing;
```

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +16,22 @@
...
+  border-radius: 0 0 20px 20px;
   text-align: center;
 
   .close-button {
     cursor: pointer;
     position: absolute;
-    top: 10px;
+    top: 15px;
     right: 15px;
```
0xdevalias, line 27:
```suggestion
    top: $spacing;
    right: $spacing;
```

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -15,18 +16,22 @@
...
     :hover {
       color: var(--primary-color);
       text-decoration: underline;
     }
   }
 }
+
+.announcement-message {
+  margin: 0px 30px;
+}
```
0xdevalias, line 37: At the top of the file with the other variables I would probably add something like:

```scss
$close-icon-width = 20px;
$message-margin = $spacing + $close-icon-width + 5px;
```

Then here, we could have:

```suggestion
.announcement-message {
  margin: 0 $message-margin;
}
```

Further reading:

- see reference links for SCSS math/calculations/etc in https://github.com/sparkletown/sparkle/pull/1184#discussion_r578884512

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
0xdevalias: Might also be worth adding `width` here explicitly: `width: $close-icon-width`

0xdevalias (changes requested):
A few more little tweaks, this is getting reall close!

0xdevalias (approved):
Looks good! LGTM! :shipit: 

image

image

PR #1187, opened 2021-02-17
add admin venues cards to the admin v2 page

denisdimitrov:
Adds the venue cards to the admin v2 default UI. Sidebar is shown only when you select a venue to manage with the ability to go back via the new button. This sidebar will be removed in the next couple of PR's for redesign.

W.I.P: Handle UI when there are no venues on the page

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/271

adminv2
backbutton


src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -0,0 +1,99 @@
+@import "scss/constants.scss";
+
+$primary: #6f43ff;
```
denisdimitrov, line 3: If I understand correctly this should be the primary color in the constants which is purplish instead of the blue current one. @mcflyok Is this statement correct?

src/pages/Admin/Admin.styles.ts
```diff
@@ -1,24 +1,20 @@
...
-
   scrollbar-width: thin;
 
-  .no-venue-selected {
-    display: flex;
-    justify-self: center;
-    align-self: center;
-
-    font-size: 3rem;
-  }
```
0xdevalias, line 20: The deleted styles here aren't fully represented in the styles that had `hasSelectedVenue` added to them. Just want to confirm that this is intended/as expected.

src/pages/Admin/Admin_v2.tsx
```diff
@@ -3,30 +3,30 @@ import "firebase/storage";
...
-import { useAdminVenues } from "hooks/useAdminVenues";
 import EntranceExperience from "./EntranceExperience";
 import AdvancedSettings from "./AdvancedSettings";
+import VenueDetails from "./Venue/Details";
 // import TicketingAndAccess from "./TicketingAndAccess";
 
+import "./Admin.scss";
 import * as S from "./Admin.styles";
-import BasicInfo from "./BasicInfo";
+import { LoadingPage } from "components/molecules/LoadingPage";
```
0xdevalias, line 29: Should be sorted with `components/` imports

src/pages/Admin/Admin_v2.tsx
```diff
@@ -3,30 +3,30 @@ import "firebase/storage";
...
+import AuthenticationModal from "components/organisms/AuthenticationModal";
+import BasicInfo from "./BasicInfo";
 import AdminSidebar from "./Sidebar/Sidebar";
-import { useVenueId } from "hooks/useVenueId";
-import { orderedVenuesSelector } from "utils/selectors";
-import { useAdminVenues } from "hooks/useAdminVenues";
 import EntranceExperience from "./EntranceExperience";
 import AdvancedSettings from "./AdvancedSettings";
+import VenueDetails from "./Venue/Details";
 // import TicketingAndAccess from "./TicketingAndAccess";
```
0xdevalias, line 25: Is this needed here anymore? Can it be removed?

src/pages/Admin/Admin_v2.tsx
```diff
@@ -140,21 +140,20 @@ const Admin_v2: React.FC = () => {
...
   return (
     <>
-      <S.Wrapper>
-        <AdminSidebar
-          sidebarOptions={sidebarOptions}
-          selected={selectedOption}
-          onClick={setSelectedOption}
-        />
+      <S.Wrapper
+        className="no-venue-selected"
```
0xdevalias, line 144: `.no-venue-selected` was removed from `src/pages/Admin/Admin.styles.ts` yet is still used here. Is that correct? If so, where is the class coming from now?

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Button } from "react-bootstrap";
+import { Link } from "react-router-dom";
+
+import { AnyVenue, VenueTemplate } from "types/venues";
+
+import { WithId } from "utils/id";
+
+import "./AdminVenues.scss";
+
+interface AdminVenuesProps {
```
0xdevalias, line 11:
```suggestion
export interface AdminVenuesProps {
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+  venues: WithId<AnyVenue>[];
+}
+
+export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
+  const partyVenues = useMemo(() => {
+    return venues?.filter((venue) => venue.template === VenueTemplate.partymap);
+  }, [venues]);
+
+  const hasVenues: boolean = !!partyVenues.length;
```
0xdevalias, line 20:
```suggestion
  const hasVenues = partyVenues.length > 0;
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <div className="admin-venue">
+      <div className="admin-venue__create-button">
+        <h3>Admin Dashboard</h3>
+        <Button as={Link} to="/admin_v2/venue/creation">
+          Create a new space
+        </Button>
+      </div>
```
0xdevalias, line 29:
```suggestion
      </div>

```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+                    style={{ backgroundImage: `url(${venue.host?.icon})` }}
+                  ></div>
+                  <h3>{venue.name}</h3>
+                </div>
+                <Link className="card__button" to={`/admin_v2/${venue.id}`}>
+                  Manage Party
+                </Link>
+              </div>
+            );
+          })}
```
0xdevalias, line 60: We'll want to `useMemo` this `.map`.

We don't ever want to use `index` as a key as it's unstable/causes rendering issues. I believe it should be possible to use `venue.id` here?

It would also make sense to extract a `AdminVenueCard` or similar component that encapsulates this chunk here (not including the `.map` part)

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+                  className="card__bg"
+                  style={{
+                    backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
+                    backgroundSize: "cover",
+                  }}
+                ></div>
+                <div className="card__info">
+                  <div
+                    className="card__logo"
+                    style={{ backgroundImage: `url(${venue.host?.icon})` }}
```
0xdevalias, line 51: This will render as `url(undefined)` or similar when `venue.host?.icon` isn't defined. We should have a different/better option for this.

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <div className="admin-venue__empty-venues">
+          <h3>Welcome!</h3>
+          <h3>Create your first Sparkle space</h3>
+        </div>
+      )}
+      {hasVenues && (
+        <div className="admin-venue__cards">
+          {partyVenues.map((venue, index) => {
+            return (
+              <div key={index} className="card">
```
0xdevalias, line 40: As part of extracting this into its own sub-component, it would make sense to change this `className` to something more specific, such as `venue-card` or similar (+the appropriate related BEM changes once updated)

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+                <div
+                  className="card__bg"
+                  style={{
+                    backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
+                    backgroundSize: "cover",
```
0xdevalias, line 45: Can this be added to the `card__bg` class? If not, why does it need to be an inline style here?

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div className="admin-venue__create-button">
+        <h3>Admin Dashboard</h3>
+        <Button as={Link} to="/admin_v2/venue/creation">
+          Create a new space
+        </Button>
+      </div>
+      {!hasVenues && (
+        <div className="admin-venue__empty-venues">
+          <h3>Welcome!</h3>
+          <h3>Create your first Sparkle space</h3>
```
0xdevalias, line 33: This feels like we're abusing the `h3` tag for its size/etc, but breaking it's 'semantic guarantee' of being a heading (which has impacts on accessibility/screenreaders/similar)

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div className="admin-venue">
+      <div className="admin-venue__create-button">
+        <h3>Admin Dashboard</h3>
+        <Button as={Link} to="/admin_v2/venue/creation">
+          Create a new space
+        </Button>
+      </div>
+      {!hasVenues && (
+        <div className="admin-venue__empty-venues">
```
0xdevalias, line 31: Potentially this would still be `admin-venue__cards` but with a `--empty` modifier or similar? As it seems to be the same 'slot' in the UI.

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,65 @@
...
+import "./AdminVenues.scss";
+
+interface AdminVenuesProps {
+  venues: WithId<AnyVenue>[];
+}
+
+export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
+  const partyVenues = useMemo(() => {
+    return venues?.filter((venue) => venue.template === VenueTemplate.partymap);
+  }, [venues]);
```
0xdevalias, line 18:
```suggestion
  const partyVenues = useMemo(() => venues?.filter((venue) => venue.template === VenueTemplate.partymap) ?? [], [venues]);
```

0xdevalias (changes requested):
Some comments/changes/etc.

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -0,0 +1,112 @@
...
+
+.admin-venue {
+  width: 100%;
+  height: 100%;
+
+  .admin-venue__empty-venues {
```
0xdevalias, line 9:
```suggestion
  &__empty-venues {
```

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -0,0 +1,112 @@
...
+    display: flex;
+    height: 100%;
+    justify-content: center;
+    align-items: center;
+    flex-direction: column;
+  }
+
+  .admin-venue__create-button {
```
0xdevalias, line 17:
```suggestion
  &__create-button {
```

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -0,0 +1,112 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+  }
+
+  .admin-venue__cards {
```
0xdevalias, line 23:
```suggestion
  &__cards {
```

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -0,0 +1,112 @@
...
+        text-decoration: none;
+
+        &:hover {
+          color: $white;
+          background-color: darken($primary, 5%);
+          box-shadow: 0 2px 30px rgba(124, 70, 251, 0.6);
+        }
+      }
+    }
+  }
```
0xdevalias, line 111: These styles should move to a new file/sub-component, and be named something like `.venue-card` or similar.

Each BEM 'element' should be at the 'root', rather than nested within their 'parent structures'.

So it would look something like:

```scss
.venue-card {
  // ..snip..
  
  &__background { /* ..snip.. */ }

  &__info { /* ..snip.. */ }

  &__logo { /* ..snip.. */ }

  &__button { /* ..snip.. */ } // Potentially give this a more descriptive name?

  // ..snip..
}
```

0xdevalias (changes requested):
Forgot to review these styles the first time.

src/pages/Admin/Admin.styles.ts
denisdimitrov: Yup, they were unused

src/pages/Admin/Admin_v2.tsx
denisdimitrov: Some venues have tickets and this will be added in the future when we have the whole picture.

src/pages/Admin/Admin_v2.tsx
denisdimitrov: This is not being used and I'm removing it.

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: > We'll want to useMemo this .map

Why? This will rerender only if there is a change in `partyVenues` which will change only when `venues` changes. It will have no benefit only using resources for caching this.

denisdimitrov: Sometimes we are overusing the memo/cache logic, I think this is one of those cases.

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: This is a required field when creating a venue. It will fallback to undefined if someone removes the icon from the database directly. That being said I'm still going handle it but we need to think about the other fields as well.

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: Venues is required array field. No need for this.

denisdimitrov: The inline syntax looks good though.

src/components/organisms/AdminVenues/AdminVenues.scss
mcflyok: Yes! Sorry I've just seen this now !

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: This is a separate element on purpose. If venues exist you I render element A, if the venues don't exist I render element B

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: Isn't this like the only reason to use it? For it's styling and heading purpose? Otherwise we'd be styling everything in the project and have no h1,h2....etc.

src/pages/Admin/Admin.styles.ts
0xdevalias: Cool :)

src/pages/Admin/Admin_v2.tsx
0xdevalias: Makes sense. It might be useful to add a `// @debt` comment above it explaining that. Otherwise I'll end up asking about it in every PR I see.

src/pages/Admin/Admin_v2.tsx
0xdevalias: Sounds good.

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: > This will rerender only if there is a change in partyVenues which will change only when venues changes.

I don't believe that is correct.

This `.map()` will run every single time this component is re-rendered for any reason (which includes if it's own hooks/dependencies change, if any of its props change, or if literally any parent of it is re-rendered for any reason), even when `partyVenues` is exactly the same.

If you believe otherwise, I'd love for you to point to reference material that shows why that isn't the case; but this is my understanding of how React works, particular with regards to functional components as we use.

You could avoid a 'double useMemo' situation by just moving the `.map` into the existing useMemo, after the `.filter`, and then rendering that `const` directly in the JSX. The only logic here that seems to rely on it is the `.map` (that we would be moving into it anyway), and the `.length` (which would still work after the suggestion here)

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: I guess that raises a bigger question about what we should 'allow'/how we should write our types/etc. Eg. is 'someone broke it in the DB' a situation that our types should account for? Currently, since we don't validate venue config/enforce it when edited in the DB, I would opt for the more cautious side of things (eg. where we assume the worst and handle it accordingly)

If we were to add something that validated and enforced any DB changes to ensure they matched our standards (regardless of where those edits are made from), then we would be able to trust that a venue is in a 'valid state', and could opt for more lenient typing. This would be my preferred option of how to handle things, but obviously that requires a bunch of work that is definitely OOS of this PR.

0xdevalias: (backlogged the above in https://github.com/sparkletown/internal-sparkle-issues/issues/432)

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: Having 2 `h3` headings one after the other isn't using it for it's 'heading purpose'. We also already use styling explicitly to set text size/boldness/etc in many places within the app.

The heading tags are designed to convey the semantic structure of a document's sections, as well as it's hierarchical flow through subsections. The fact that they are styled a certain way is a side effect (that is often abused by web developers), rather than their main intent/purpose.

---

- https://www.w3.org/WAI/tutorials/page-structure/headings/
  - >  Headings with an equal or higher rank start a new section, headings with a lower rank start new subsections that are part of the higher ranked section.
- https://usability.yale.edu/web-accessibility/articles/headin
  - > Do not select heading levels based on their appearance. Select the appropriate heading rank in your hierarchy.
  - > Headings are often improperly used in cases where there are no dedicated semantic HTML elements(link is external). Headings should not be used as subheadings, subtitles, alternative titles, taglines, or slogans. In such cases, it is best to use a <p> tag.
 
---

It's the same reason that bootstrap separates its 'header-like styles' from the semantic h1-6 tags:

- https://getbootstrap.com/docs/4.0/content/typography/
  - > .h1 through .h6 classes are also available, for when you want to match the font styling of a heading but cannot use the associated HTML element.

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: I understand what/how this is being used here, i'm just suggesting that the distinction/separate element for it is probably not needed, as per my above.

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: Suggestion was based on the fact that it was using `venues?.`, implying it could be undefined.

0xdevalias:
@denisdimitrov Guessing you rebased this PR? As discussed in the past, please don't rebase when PR's are in review, as it breaks all of the tooling GitHub provides to assist with the review process, and makes it harder for me to see what has changed since my last review, etc.

image

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const AdminVenueCard: React.FC<AdminVenueCardProps> = ({ venue }) => {
+  return (
+    <div className="admin-venue-card">
+      <div
+        className="admin-venue-card__bg"
+        style={{
+          backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
+        }}
+      ></div>
```
0xdevalias, line 24: Since the div has no content, use the 'single tag' format:

```suggestion
      <div
        className="admin-venue-card__bg"
        style={{
          backgroundImage: `url(${venue.mapBackgroundImageUrl})`,
        }}
      />
```

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <div className="admin-venue-card__info">
+        <div
+          className="admin-venue-card__logo"
+          style={{
+            backgroundImage: `url(${venue.host?.icon ?? DEFAULT_VENUE_LOGO})`,
+          }}
+        ></div>
```
0xdevalias, line 31:
```suggestion
        <div
          className="admin-venue-card__logo"
          style={{
            backgroundImage: `url(${venue.host?.icon ?? DEFAULT_VENUE_LOGO})`,
          }}
        />
```

src/pages/Admin/Admin_v2.tsx
0xdevalias: Makes sense. It might be useful to add a `// @debt` comment above it explaining that. Otherwise I'll end up asking about it in every PR I see.

0xdevalias (changes requested):
A few comments/changes/etc,

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: I understand what you are saying but in this case we have a single prop `venues` and nothing else. The parent doesn't have a state that can retrigger a render and this will always rely on `venues` If there is a change in `venues` the memo will trigger because of the dep array and the `.map` logic will go again.

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: In some of those links there are examples with multiple headings one after another and I think it makes perfect sense. Why would it be semantically incorrect to have multiple h2 or h3 one after another. In my case I have 2 heading texts and I'm taking advantage of the h3. Yes, there are no children which makes it slightly confusing. And there are no parent elements h1 and h2. But I don't think using multiple headings one after another is a wrong thing to do.

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
```diff
@@ -0,0 +1,79 @@
+@import "scss/constants.scss";
+
+$primary: #6f43ff;
```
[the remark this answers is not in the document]
denisdimitrov: Our current primary color is different than the one in the designs, I'm going to change it to this one. But I'm not sure how this will affect our layout/theme across the website. I'll test it out and see how it goes.

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
```diff
@@ -0,0 +1,79 @@
...
+
+.admin-venue-card {
+  position: relative;
+  overflow: hidden;
+  z-index: 1;
```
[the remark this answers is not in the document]
denisdimitrov: This one is unnecessary.

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
```diff
@@ -0,0 +1,79 @@
...
+  display: flex;
+  justify-content: flex-end;
+  flex-direction: column;
+  border: none;
+  padding: 1rem;
+  border-radius: 20px;
+
+  &__bg {
+    position: absolute;
+    z-index: -1;
```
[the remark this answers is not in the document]
denisdimitrov: This renders the background linear background behind the components on the avatar card.

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: Those examples omit the content for brevity, but the intent is that they are each a unique section containing data. The way it’s used here is for the ‘same’ heading, or perhaps as a ‘subheading’.

@mcflyok May be outside the technical realm of ‘design’, but since you’re pretty up to scratch on your HTML demo skills, do you have any thoughts here?

@mike-lvov Potentially also?

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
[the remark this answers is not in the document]
0xdevalias: Even if we can’t change the global primary colour, you can just rename the variable when you include it there, like $primary-admin-venue-card and put it on the line after $primary.

that way it’s moved to the constants and reshareable, but won’t impact beyond the scope of this PR. Will make it a simple job to clean up in future once we know what/how we want to align all of our colours/etc.

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: Are you saying that no parent/grandparent/etc component literally anywhere in the ‘parent chain’ between here and the very root of the app will ever render more than once under any circumstance? Because if not, then the issue I stated above remains true.

Unless the map is in the useMemo, it will run every time this component function is called. React can and does call component functions many times, even when ultimately it results in the same output. In those cases, it doesn’t modify the DOM, but it still had to run the entire component function to check that. This is a standard case of react component optimisation, and there is a lot written on it across the net.

An alternative would be to use React.memo on the entire component, at which point it may react how it sounds like you’re thinking it should react currently. If it’s purely driven by its props, has no internal state/connections/side effects that would change other than through those props, and contains no links to any child components that have any internal state/etc that would need to be triggered for any case other than these props changing, then that would be a fine approach. But if we do that, it entirely removes everything beneath this ‘node’ in the ‘render tree’ from being rendered, it’s a shortcut circuit. So if there was a child/grandchild/etc component somewhere in its hierarchy that was meant to listen/update based on redux state/similar, it wouldn’t anymore, which can lead to weird/hard to track down bugs. React.memo is a fairly nuclear option, and really only makes sense IMO where you can guarantee ‘pure’ ‘dumb components’ the whole way down, which IMO we rarely if ever can given the state of our platform code/design at the moment.

That’s why we tend to favour useMemo, and why I’m saying here (and other places where we map in the JSX) to do so.

It’s a simple matter of moving the .map inside the existing useMemo.

This is a blocker from me. If this gets merged without it, I will have to go and chase down the tech debt created and raise a PR to implement it. I don’t feel like this should be a contentious issue, and I’m kind of confused why you seem to be pushing back so strongly about it.

(cc // @mike-lvov perhaps you know of some good resources that can explain better than I have here? Or alternately that ‘disprove’ my claims here?)

If this is still an issue for you, please start a thread on slack, referring this comment thread, and let’s discuss.

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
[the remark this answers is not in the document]
0xdevalias: @denisdimitrov Looks like this still needs to be done.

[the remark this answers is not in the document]
0xdevalias: Actually, since you've updated the main `$primary` definition now, I think this just needs to be removed here.

src/components/molecules/EventDisplay/EventDisplay.scss
```diff
@@ -1,5 +1,5 @@
 $side-padding: 30px;
-$primary: #005ee5;
+$primary: #6f43ff;
 $black: #000000;
 $white: #fff;
```
0xdevalias, line 4: Since this is the same as the main definition in constants now, I think we can just delete this (and import constants if required)

```suggestion
```

src/components/molecules/NavBar/playa.scss
```diff
@@ -1,7 +1,6 @@
 @import "scss/constants.scss";
 
 $side-padding: 30px;
-$primary: #005ee5;
 $black: #000000;
 $white: #fff;
```
0xdevalias, line 5:
```suggestion
```

src/components/organisms/AdminVenues/AdminVenues.tsx
0xdevalias: This still needs to be done.

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+      {!hasVenues && (
+        <div className="admin-venue__cards--empty">
+          <h3>Welcome!</h3>
+          <h3>Create your first Sparkle space</h3>
+        </div>
+      )}
+
+      {hasVenues && (
+        <div className="admin-venue__cards">
```
0xdevalias, line 42: Since both this and the above use the same div/class now, we can keep the div statically rendered, adjust the classes on it using our `classNames` helper, and only change the inner content as needed.

src/components/organisms/SchedulePageModal/SchedulePageModal.scss
```diff
@@ -1,5 +1,5 @@
 $side-padding: 30px;
-$primary: #005ee5;
+$primary: #6f43ff;
 $black: #000000;
 $white: #fff;
```
0xdevalias, line 4: Import the main constants file if required, and then:

```suggestion
```

0xdevalias (changes requested):
Cleanup on a few of the unneeded scss variables + the few remaining outstanding fixes from discussions from previous reviews (commented on in this review), including (which I was unable to comment on within this review for some reason):

- https://github.com/sparkletown/sparkle/pull/1187#discussion_r582284091

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: My intention is not to push this back and I'm sorry if it feels that I'm strongly trying to do so, I'm going to change it with my next commit, I don't want to force you to go back and clean my mess.

But like I already said we are overusing the memos and callbacks and there are many blogs stating that this should not be overused because sometimes the resources it takes to cache the data is more than it would take to re-render the component or run through the function again, usually in React you don't think about performance until you have a problem, I believe I saw this somewhere written by Dan Abramov or maybe in the React website itself. React is one of the fastest frameworks and you can re-render a component 10-15 times and the user won't notice because that's less than 50 milliseconds. I'm not staying stop using memo and callback, but maybe turn it down a notch and use it only when we have complex and heavy logic or big components that re-rendering them would slow don't the page. I believe the only place where this performance boost was hugely noticeable was when you refactored the grid to use callback and memo and the animation was smooth and pleasant because it was caching 5-10k div elements on the screen.

denisdimitrov:
@sparkletown/engineering All of the change requests have been applied.

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,35 +17,39 @@ export interface AdminVenuesProps {
 
 export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
   const partyVenues = useMemo(
-    () => venues?.filter((venue) => venue.template === VenueTemplate.partymap),
+    () =>
+      venues
+        ?.filter((venue) => venue.template === VenueTemplate.partymap)
```
0xdevalias, line 22: `src/types/venues.ts:291`

```suggestion
        ?.filter(isPartyMapVenue)
```

Does this actually care if it's a Party Map venue specifically? Or only if it's a venue with rooms?

```suggestion
        ?.filter(isVenueWithRooms)
```

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -4,10 +4,15 @@
...
+  &__header {
     display: flex;
     align-items: center;
     justify-content: space-between;
+
+    &--title {
+      font-size: 2em;
+      font-weight: bolder;
+    }
   }
```
0xdevalias: The title isn't a 'modifier', it's just more just another 'element':

- http://getbem.com/naming/

```suggestion
  &__header {
    display: flex;
    align-items: center;
    justify-content: space-between;
  }
  
  &__title {
    font-size: 2em;
    font-weight: bolder;
  }
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,35 +17,39 @@ export interface AdminVenuesProps {
...
   );
 
   const hasVenues = partyVenues.length > 0;
 
   return (
     <div className="admin-venue">
-      <div className="admin-venue__create-button">
-        <h3>Admin Dashboard</h3>
+      <div className="admin-venue__header">
+        <div className="admin-venue__header--title">Admin Dashboard</div>
```
0xdevalias, line 32:
```suggestion
        <div className="admin-venue__title">Admin Dashboard</div>
```

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -25,6 +30,11 @@
       flex-direction: column;
     }
 
+    &--welcome-title {
+      font-size: 2em;
+      font-weight: bolder;
+    }
+
```
0xdevalias, line 37: This isn't needed as it's exactly the same as `&__title` now

```suggestion
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,35 +17,39 @@ export interface AdminVenuesProps {
...
-        </div>
-      )}
+      <div
+        className={classNames("admin-venue__cards", {
+          "admin-venue__cards--empty": !hasVenues,
+        })}
+      >
+        {!hasVenues && (
+          <>
+            <div className="admin-venue__cards--welcome-title">Welcome!</div>
```
0xdevalias, line 45:
```suggestion
            <div className="admin-venue__title">Welcome!</div>
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,35 +17,39 @@ export interface AdminVenuesProps {
...
+            <div className="admin-venue__cards--welcome-title">
```
0xdevalias, line 46:
```suggestion
            <div className="admin-venue__title">
```

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,35 +17,39 @@ export interface AdminVenuesProps {
...
+              Create your first Sparkle space
+            </div>
+          </>
+        )}
+        {hasVenues && partyVenues}
```
0xdevalias, line 51: This line is sort of confusing at the moment. Maybe this would make it easier to read?

```suggestion
        {hasVenues && renderedPartyVenues}
```

0xdevalias (approved):
[CRO] Review based on the diff between my last review and now, didn't re-look through all of the changes, assuming they are fine.

@denisdimitrov A few comments/suggestions to be fixed up before this can be merged. I'll give the approval since they are relatively minor, but please make sure to do so before merging.

src/components/organisms/AdminVenues/AdminVenues.tsx
denisdimitrov: These should be partymap venues only. Because admin v2 creates a party map by default and all of the cards should be party map venues only. This is probably going to change after we add the worlds concept.

0xdevalias: Sounds good.

PR #1191, opened 2021-02-18
Fix admin edit buttons

malaikaaiyar:
This fixes sparkletown/internal-sparkle-issues#393

Before
Screen Shot 2021-02-18 at 4 52 10 PM

After
Screen Shot 2021-02-18 at 4 52 31 PM

### DIFF
--- a/src/pages/Admin/Room/Edit/RoomEdit.styles.ts
+++ b/src/pages/Admin/Room/Edit/RoomEdit.styles.ts
@@ -2,6 +2,11 @@ import styled from "styled-components";
 
 export const Wrapper = styled.div``;
 
+export const ButtonsWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+`;
+
 export const InputWrapper = styled.div`
   display: flex;
   width: 100%;
--- a/src/pages/Admin/Room/Edit/RoomEdit.tsx
+++ b/src/pages/Admin/Room/Edit/RoomEdit.tsx
@@ -143,13 +143,15 @@ const EditRoom: React.FC<EditRoomProps> = ({
 
           {renderTemplateSelect()}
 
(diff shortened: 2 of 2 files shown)


0xdevalias (approved):
Looks good! Clean implementation that follows the existing code pattern in this file. LGTM! :shipit: 

---

[OOS] `styled-components` needs to 🔥

@malaikaaiyar If you wanted to try your hand at converting a `styled-components` pattern (which is what is used here, but is a pattern we want to remove from our codebase), and convert it to our standard scss pattern, this could be a good place to start/get a feel for it.

Our 'standard scss pattern' is a single `.scss` file, named the same as the component file (so in this case `RoomEdit.scss`), imported into the component file (you'll see how this is done in most places outside of admin v2), and following the BEM naming pattern for the CSS/SCSS classes, that would be cool.

Given how few styles there are for this component, and that they are all static (`styled-components` allows you to pass variables into the styles, so in some cases converting it to scss will be slightly more tricky, but not here) it should be relatively straightforward.

Further Reading:

- BEM (block, element, modifier):
  - A CSS naming style/convention that makes styles cleaner/easier to understand, and 'scopes' things such that it avoids 'naming clashes' that can cause weird/unexpected/hard to debug styling issues
  - http://getbem.com/
  - http://getbem.com/naming/
  - 
- Sass / SCSS: https://sass-lang.com/
  - a CSS preprocessor that adds more powerful features for writing clean/reusable styles
  - Follows the same basic syntax/style as normal CSS, so easy to pick up, but adds way more powerful/nice features on top!
  
(**Edit:** Backlogged this section in https://github.com/sparkletown/internal-sparkle-issues/issues/412)

malaikaaiyar:
@0xdevalias, Mike did help me with this one, but it's good that it works well! I can work on the the scss conversion; I'll comment on the issue you created.

PR #1192, opened 2021-02-18
Make the bottom bar appear on scrolling, not as an overlay always visible 

jigar-jstechno:
This fixes https://github.com/sparkletown/internal-sparkle-issues/issues/233.

Also add padding:10px


0xdevalias (changes requested):
Looking at how this renders on the preview branch, this looks like it's all sorts of broken at the moment.

- https://co-reality-staging--preview-pr-1192-qov7b6x7.web.app/in/devaliashacksville

image

Can you please look into deeper into this, fix up any issues, test that it looks correct, and then mark this ready for review again.

There are a number of places (eg. `Sidebar`) that currently 'make space' for the the sticky footer in their styling that will also need to be changed to 'fill the gap'.

jigar-jstechno:
image
After setting the sidebar it looks like this.Can I change the z-index of footer behind sidebar or it looks fine

0xdevalias:
> After setting the sidebar it looks like this.Can I change the z-index of footer behind sidebar or it looks fine

As part of being 'non-sticky', I believe the footer should be sitting 'below' the sidebar in the page. So their z-indexes should never need to interact.

jigar-jstechno:
> > After setting the sidebar it looks like this.Can I change the z-index of footer behind sidebar or it looks fine
> 
> As part of being 'non-sticky', I believe the footer should be sitting 'below' the sidebar in the page (eg. needing to scroll to be seen). So their z-indexes should never need to interact.

Sidebar is also fixed so when scrool down footer overlaps sidebar

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -11,7 +11,7 @@ $header-height: 45px;
   top: 66px;
   right: 0;
   width: 360px;
-  height: calc(100vh - 96px);
+  height: calc(100vh - 66px);
```
0xdevalias, line 14: What does this relate to? What did the `96px` represent, and what does the `66px` now represent? This seems like a place where we should potentially be using variables from `scss/constants`

I think maybe it is what is described in https://github.com/sparkletown/internal-sparkle-issues/issues/233#issuecomment-776502846 ?

0xdevalias (changes requested):
A couple of little comments.

---

RE: https://github.com/sparkletown/sparkle/pull/1192#issuecomment-783146431

Yeah, I can see the issue there now.

Honestly I'm not sure how to easily fix this off the top of my head, and given I believe we may want to remove the footer entirely (in favour of moving the things listed in it somewhere else, possibly in a dropdown from the `NavBar`, i'm not sure that it's worth the time/effort trying to resolve it now (and personally, I kind of fundamentally disagree that the 'issue' raised in https://github.com/sparkletown/internal-sparkle-issues/issues/233 is even something that matters/should be changed in the first place)

---

@sparkletown/engineering We can make this scrollable/offscreen if we do the following (though not sure if that is a good idea/worth it/what else it might impact elsewhere)

- `src/components/organisms/WithNavigationBar/WithNavigationBar.tsx`
  - always apply the `.fullscreen` class (or more likely, move those styles into the main class in `src/components/organisms/WithNavigationBar/WithNavigationBar.scss` + remove the `fullscreen` prop)
    - There are a LOT of places that render this component currently.. not sure if/how this would impact those
  - also remove the `overflow: hidden` styles from that class (which then seems to allow the footer to render 'below the fold', and allows a slight scroll to get to it)
    - With this method, it seems there will be a slight 'on render' bug where the footer is displayed much higher up, and looks broken. I believe that would be resolved by properly setting the height of the map image, even before it's loaded (or similar)
- Revert the change in this PR that makes the footer have a higher `z-index` than the sidebar, and ensure it has a lower z-index, or none at all (realistically, we may not need to set any `z-index` on the footer anymore)

Do we want to go ahead with that sort of thing to make this work? Or is it fine to just leave the sticky footer there (my preference), as I personally don't feel like it's a 'big deal' anymore since it was reduced in size, and I actually think it's better being there than not being there (which I know is an unpopular opinion)

0xdevalias:
From https://github.com/sparkletown/internal-sparkle-issues/issues/233#issuecomment-783963615:

> I think we leave the issue as is - it is reduced in size and "not a problem"

This was discussed/decided in our meeting today.

---

Given the above, I'll close this PR now.

PR #1193, opened 2021-02-19
Change admin_v2 edit room modal from styled components to scss

malaikaaiyar:
Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/412


src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
       <Modal.Body>
         <Form onSubmit={handleSubmit(onSubmit)}>
-          {renderNameInput()}
-          {renderUrlInput()}
-          {renderDescriptionInput()}
-          {renderLogoInput()}
-
-          {roomTemplate?.customInputs?.map((input) => renderCustomInput(input))}
-
-          {renderTemplateSelect()}
```
0xdevalias, line 144: Technically all of this (+related) refactoring seems OOS for this PR as it's title/description/linked issue are concerned. What was the reasoning/thought process behind this? Does it relate to a separate issue? It's generally best to keep PRs focussed on one thing, as it make it easier to review + easier to see where the changes happened/why in future.

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+                forwardRef={register}
+                small
+                nameWithUnderscore
+                imgUrl={room.image_url}
+              />
+            </div>
+          </Form.Row>
+
+          {roomTemplate?.customInputs?.map((input) => (
+            <div className="input-wrapper" key={input.name}>
```
0xdevalias, line 113: Is `input.name` guaranteed to be unique here?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+                name="image"
+                forwardRef={register}
+                small
+                nameWithUnderscore
+                imgUrl={room.image_url}
+              />
+            </div>
+          </Form.Row>
+
+          {roomTemplate?.customInputs?.map((input) => (
```
0xdevalias, line 112: Potentially this `.map` should be `useMemo`'d to prevent needing to be run on every render of this component.

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+            <div className="input-wrapper">
+              <Form.Label>Change room template</Form.Label>
+              <Form.Control
+                as="select"
+                custom
+                name="template"
+                ref={register}
+                onChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
+                  setSelectedTemplate(e.target.value)
+                }
```
0xdevalias, line 129: Potentially this `onChange` should be `useCallback`'d to prevent re-rendering (inline functions will change their 'reference' on every render.

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+              >
+                {ROOM_TEMPLATES.map((template) => (
+                  <option key={template.name}>{template.template}</option>
+                ))}
```
0xdevalias, line 133: Potentially this `.map` should be `useMemo`'d to prevent needing to be run on every render of this component.

Since it's operating on a constant, it doesn't even need to happen within this component. So you could have it happen once outside of this component, then just render the result here.

0xdevalias: Is `template.name` unique? We'd want to use the most unique `id` type value for the `key` here.

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -38,92 +37,13 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
       url: room.url,
       description: room.description,
-      tempalte: room.template,
+      template: room.template,
     },
   });
 
   const values = watch();
 
   const handleImageChange = (val: string) => setValue("image_url", val, false);
```
0xdevalias, line 46: Potentially `useCallback` this?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+              <Form.Label>The room url</Form.Label>
+              <Form.Control
+                type="text"
+                ref={register}
+                name="url"
+                placeholder="Room url"
+                custom
+              />
+            </div>
+          </Form.Row>
```
0xdevalias, line 81: This block seems to differ from how it was previously defined in `renderUrlInput`. What was the reasoning for the differences?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+              <Form.Label>The room description (optional)</Form.Label>
+              <Form.Control
+                type="text"
+                ref={register}
+                name="description"
+                placeholder="Description"
+                custom
+              />
+            </div>
+          </Form.Row>
```
0xdevalias, line 94: This block seems to differ from how it was previously defined in `renderDescriptionInput`. What was the reasoning for the differences?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+              <ImageInput
+                onChange={handleImageChange}
+                name="image"
+                forwardRef={register}
+                small
+                nameWithUnderscore
+                imgUrl={room.image_url}
+              />
+            </div>
+          </Form.Row>
```
0xdevalias, line 110: This block seems to differ from how it was previously defined in `renderLogoInput`. What was the reasoning for the differences?

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -1,14 +1,13 @@
...
-import { Button } from "react-bootstrap";
+import { useForm } from "react-hook-form";
+import { Form, Modal, Button } from "react-bootstrap";
+import { RoomTemplate, ROOM_TEMPLATES } from "settings";
 import { EditRoomProps } from "./RoomEdit.types";
 
-const EditRoom: React.FC<EditRoomProps> = ({
+import "./RoomEdit.scss";
+
+export const EditRoom: React.FC<EditRoomProps> = ({
```
0xdevalias, line 10: We should align the filename and the main component's name here. These should both be one of `RoomEdit` or `EditRoom` (not both)

Since this seems to be a Modal, it may also be beneficial to suffix this with `Modal`, making it one of `RoomEditModal` or `EditRoomModal`.

I feel like @denisdimitrov made some naming changes related to this (or a similar pattern) recently. I think he started with a noun from memory.. so making this `RoomEditModal` would probably be the preference.

src/pages/Admin/Room/Edit/RoomEdit.scss
```diff
@@ -1,17 +1,15 @@
...
 
-export const Wrapper = styled.div``;
-
-export const ButtonsWrapper = styled.div`
-  display: flex;
-  justify-content: space-between;
-`;
-
-export const InputWrapper = styled.div`
+.input-wrapper {
```
0xdevalias, line 3: Following BEM we should probably have a 'root' class such as `.room-edit-modal` (which would be applied to the 'outermost container element', so probably the `Modal.Body` or similar, if it can have classes applied to it?), and then each of these would be an 'element' under that, so `.room-edit-modal__input` and `.room-edit-modal__buttons`.

Using scss, we can write/nest this as:

```scss
.room-edit-modal {
  &__input { /* add the scss here */ }

  &__buttons { /* add the scss here */ }
}
```

Further Reading:

- http://getbem.com/naming/
- https://css-tricks.com/the-sass-ampersand/
- https://sass-lang.com/documentation/style-rules/parent-selector

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: As an example, there was no `Form.Row` wrapping this before.

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: As an example, there was no `Form.Row` wrapping this before.

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: As an example, there was no `Form.Row` wrapping this before.

src/pages/Admin/Room/Edit/RoomEdit.tsx
```diff
@@ -134,24 +54,96 @@ const EditRoom: React.FC<EditRoomProps> = ({
...
+                onChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
+                  setSelectedTemplate(e.target.value)
+                }
+              >
+                {ROOM_TEMPLATES.map((template) => (
+                  <option key={template.name}>{template.template}</option>
+                ))}
+              </Form.Control>
+            </div>
+          </Form.Row>
```
0xdevalias, line 136: This block seems to differ from how it was previously defined in `renderTemplateSelect`. What was the reasoning for the differences?

As an example, there was no `Form.Row` wrapping this before.

0xdevalias (changes requested):
Some suggestions/comments/etc, including better alignment with our patterns.

It might also be good to see a before/after screenshot of how this renders, just to make it easier to 'visually see' that the changes still 'render the same', without needing to click through on the old/new code to verify this.

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: I just thought that it's easier to read/write scss if it doesn't use the "renderSomeElement()" functions. For now, should I leave it or put everything back to how it was?

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: At first, only the room name field used ```Form.Row``` and all the other blocks that are ```Form.Something```, so I added it to everything. The styling looked strange as different elements had a different amount of space around them. This seems to have fixed that (this probably should have been a new PR)

Before:
Screen Shot 2021-02-22 at 9 33 06 PM

After:
Screen Shot 2021-02-22 at 9 32 38 PM

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: Changed everything to RoomEditModal. Question: should I rename the ```EditRoomProps``` ```RoomEditModalProps``` or something else because that sounds a little too long?

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: This is the same reasoning I used when I added ```Form.Row``` etc to the all the other things that you commented.

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: Happy for you to leave it for now. We want to get rid of the 'render function' pattern in general anyway.

My comments were more related to understanding the 'why' of it, and to teach about keeping our PR's focussed. It turned what would have been a fairly quick tick/flick code review into something where I had to 'unpick' the changes, reverse engineer the intent, etc; to then be able to review it and see/make comments. Takes a lot longer/more effort/etc that way.

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: Cool, makes sense and sounds good. Thanks for explaining.

---

> this probably should have been a new PR

In future, that sort of thing probably would have been better as it's own standalone PR, 'focussed' on the intended changes.

As it is though, happy for you to leave it here now.

src/pages/Admin/Room/Edit/RoomEdit.tsx
0xdevalias: I would keep at as `RoomEditModalProps`, as that's the pattern we're following. While it is kind of long, I don't think it matters that much. In all likelihood it's only ever going to get used in 1 place within this file anyway.

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: Yes, because input.name is the type of input field (right?)

src/pages/Admin/Room/Edit/RoomEdit.tsx
malaikaaiyar: I'm trying to fix this and I get this warning which is preventing me from committing to this branch.
```  Line 62:5:  React Hook useMemo has a missing dependency: 'register'. Either include it or remove the dependency array      react-hooks/exhaustive-deps```

the code looks like this:

```
const customInputFields = useMemo(() =>
    roomTemplate?.customInputs?.map((input) => (
      <div className="room-edit-modal__input" key={input.name}>
        <span>{input.title}</span>
        <input type="text" name={input.name} ref={register} />
      </div>
    )
  ),
    [roomTemplate]
  );
```

I only half understand how hooks work. If you have any good articles/pages to read, could you share?

Comment (malaikaaiyar) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx:112]: Got a similar error for the useCallback one too:

```  Line 50:6:  React Hook useCallback has a missing dependency: 'setValue'. Either include it or remove the dependency array  react-hooks/exhaustive-deps```

with the code 

```
const handleImageChange = useCallback((val: string) => {
    setValue("image_url", val, false);
}, [room.image_url]);
```

Comment (0xdevalias) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx:112]: For general info on hooks, I recommend starting at https://reactjs.org/docs/hooks-intro.html and reading through the 8 or so related items. You can skip/skim through bits if it's not immediately relevant.

Those warnings relate to the 'rules of hooks':

- https://reactjs.org/docs/hooks-rules.html

It relates to the 'deps' (dependencies) array at the end of the hook. While that exists on a few different hooks, this section seems to explain it for the `useEffect` hook (it works the same way on all of them):

- https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

Basically, while the value in the 'deps' remain the same, the hook will only run once. When any of the 'deps' change, it will 'break' this 'memoisation' ('break memo'), and run again.

The warning/error you're seeing here is basically saying that there is some variable/thing used within the 'body' of the hook, that isn't included within that 'deps' array. Which means that the hook won't properly re-run  all the times its meant to (which can lead to weird/hard to find/debug issues)

Comment (0xdevalias) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx:112]: Generally I don't actually type the values into there manually, as my IDE allows me to automatically apply the fix, and fill the array for me.

eg. 
image

When I click on the blue ESLint 'fix', it will do it for me. I believe you can probably also apply the fix via the command line, but most IDE's should have similar functionality, and doing it in the IDE is way quicker/easier.

File: src/pages/Admin/Room/Edit/RoomEdit.tsx (hunk 2 above)
Comment (0xdevalias) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx:113]: That seems logical/probably correct (I didn't deeply read into the code)

File: src/pages/Admin/Room/Edit/RoomEdit.tsx (hunk 3 above)
Comment (malaikaaiyar) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx]: Good to know, thanks! I think I've fixed everything up now.

Comment (0xdevalias) [reply src/pages/Admin/Room/Edit/RoomEdit.tsx]: Awesome :)

File: src/pages/Admin/Room/Edit/RoomEditModal.tsx [hunk 13]
```diff
@@ -0,0 +1,173 @@
...
+  );
+
+  const [roomTemplate, setRoomTemplate] = useState<RoomTemplate | null>(null);
+
+  useEffect(() => {
+    const template = ROOM_TEMPLATES.find(
+      (i) => i.template === selectedTemplate
+    );
+    setRoomTemplate(template!);
+  }, [selectedTemplate]);
```
0xdevalias, line 29: Removing this typescript 'force resolution' hack, as it's not safe. This may cause other type issues to show up, my guess is that it will be potentially undefined, since that's what this function returns when it doesn't find anything:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Which will probably cause type issues with the `| null` above. I expect that can probably just become `| undefined` + fix up the relevant areas that consume it.

```suggestion
  useEffect(() => {
    const template = ROOM_TEMPLATES.find(
      (i) => i.template === selectedTemplate
    );
    setRoomTemplate(template);
  }, [selectedTemplate]);
```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+export const RoomEditModal: React.FC<RoomEditModalProps> = ({
+  isVisible,
+  onClickOutsideHandler,
+  room,
+  submitHandler,
+  deleteHandler,
+}) => {
+  const [selectedTemplate, setSelectedTemplate] = useState<string>(
+    room.template!
```
0xdevalias, line 19: Using TypeScript's 'non-null assertion operator' is an unsafe antipattern in most places we use it currently. It's probably better to ban it from our codebase in general.

- https://dev.to/tmaximini/typescript-bang-operator-considered-harmful-3hhi

```suggestion
    room.template
```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const handleImageChange = useCallback(
+    (val: string) => {
+      setValue("image_url", val, false);
+    },
+    [setValue]
+  );
+
+  function onSubmit() {
+    return submitHandler(values, room.roomIndex!);
+  }
```
0xdevalias, line 57: Change this to a `const onSubmit = () =>` format to match our style elsewhere + `useCallback`

src/pages/Admin/Room/Edit/RoomEditModal.types.ts
```diff
@@ -1,6 +1,6 @@
 import { RoomData_v2 } from "types/rooms";
 
-export interface EditRoomProps {
+export interface RoomEditModalProps {
   isVisible: boolean;
```
0xdevalias, line 4: Can we please move this type definition into the `src/pages/Admin/Room/Edit/RoomEditModal.tsx` file (and remove this file entirely after that), placed just above the component definition like our standard patterns. This separate `foo.types.ts` is a pattern that we want to remove from our codebase.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+              Save room
+            </Button>
+          </div>
+        </Form>
+      </Modal.Body>
+    </Modal>
+  );
+};
+
+export default RoomEditModal;
```
0xdevalias, line 173: Prefer explicitly named exports instead of `default`

```suggestion
```

src/pages/Admin/Room/Edit/index.ts
```diff
@@ -1 +1 @@
-export { default } from "./RoomEdit";
+export { default } from "./RoomEditModal";
```
0xdevalias, line 1: Prefer explicitly named exports instead of `default`

```suggestion
export { RoomEditModal } from "./RoomEditModal";
```

(this is also going to require changing how it is imported in the files that use this, pretty much just by wrapping it in `{` and `}` (you'll see other examples of how this is done)

0xdevalias (changes requested):
A few additional little things I noticed (in the old code) after renaming the file (as it all showed as changed, so I looked a little closer)

0xdevalias:
@malaikaaiyar Can you please make sure you resolve old review comments that have been completed + ideally also react to them/comment on them in a way to show that they have been done (in particular on this issue, the earlier review comments)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -2,30 +2,40 @@ import React, { useEffect, useState, useCallback, useMemo } from "react";
...
   onClickOutsideHandler,
   room,
   submitHandler,
   deleteHandler,
 }) => {
-  const [selectedTemplate, setSelectedTemplate] = useState<string>(
-    room.template!
+  const [selectedTemplate, setSelectedTemplate] = useState<string | undefined>(
+    room.template
   );
```
0xdevalias: I know I originally said to add the `| undefined` here, but apparently `useState` does that for us by default! So we can actually simplify this to just:

```suggestion
  const [selectedTemplate, setSelectedTemplate] = useState<string>(
    room.template
  );
```

Ref:

image

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -2,30 +2,40 @@ import React, { useEffect, useState, useCallback, useMemo } from "react";
...
 
-  const [roomTemplate, setRoomTemplate] = useState<RoomTemplate | null>(null);
+  const [roomTemplate, setRoomTemplate] = useState<RoomTemplate | undefined>(
+    undefined
+  );
```
0xdevalias, line 32: Looks like we can simplify this one too:

```suggestion
  const [roomTemplate, setRoomTemplate] = useState<RoomTemplate>();
```

image

0xdevalias (approved):
[CRO] Reviewed just the changes in the diff since my last review. Looks good!

Just need to merge staging into this branch (+fix any conflicts.. hopefully there aren't too many) then this should be good to go! (if there are significant changes/conflicts when merging and you're not certain, feel free to re-request review and I will quickly check it over again)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
malaikaaiyar: I implemented these changes, and at first it was fine, but after resolving conflicts it doesn't work unless I put the `| undefined` in there (same goes for the other useState).

This is the error:
```node
TypeScript error in /Users/vinayramani/Documents/GitHub/sparkle/src/pages/Admin/Room/Edit/RoomEditModal.tsx(28,5):
Argument of type 'string | undefined' is not assignable to parameter of type 'string | (() => string)'.
  Type 'undefined' is not assignable to type 'string | (() => string)'.  TS23451
```

0xdevalias: What is on `src/pages/Admin/Room/Edit/RoomEditModal.tsx(28,5)` (line 28, column 5) in your branch here?

Looking at the diff, it would seem to be `room.template`:

- https://github.com/sparkletown/sparkle/pull/1193/files#diff-7e274eb670b2bc8208899a93075c06d4f3df2d06bc5964ecf95cd2c20438c66fR28

So what is the type of `room.template`?

0xdevalias: Actually, disregard the above, looking at both of the overloads for `useState`:

image

The `| undefined` part only exists on the 'overload' when we don't pass any default state to it. But because we pass it `room.template`, it will use the other form, which requires `initialState: S | (() => S)`.

So basically, we need to keep it as `useState<string | undefined>`; I was wrong  with my original suggestion.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+              <Form.Label>The room description (optional)</Form.Label>
+              <Form.Control
+                type="text"
+                ref={register}
+                name="description"
+                placeholder="Description"
+                custom
+              />
+            </div>
+          </Form.Row>
```
0xdevalias, line 139: Should this row also have an error section on it? (my thoughts are probably yes, even if we don't currently have an explicit validation setting an error on it, having that consistent 'wiring' in place makes sense to me)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+                onChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
+                  setSelectedTemplate(e.target.value)
+                }
+              >
+                {ROOM_TEMPLATES.map((template) => (
+                  <option key={template.name}>{template.template}</option>
+                ))}
+              </Form.Control>
+            </div>
+          </Form.Row>
```
0xdevalias, line 179: Was this removed in the stuff in `staging`? I believe we wanted to remove it in general, but not sure if we got around to that yet.

Probably [OOS] if we haven't removed it already. But mostly want to make sure we aren't 'resurrecting' anything here that we didn't want to keep (so probably do a 'visual spot check' between the `RoomEdit` file in `staging` currently + this one to ensure they're aligned.

0xdevalias (approved):
[CRO] Approving this again, the changes since last approval look fine to me, though a couple of questions to check before clicking merge. If you're satisfied with those, feel free to merge.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
malaikaaiyar: In the staging `RoomEdit.tsx` file, it wasn't there so I didn't add it, but I will now

src/pages/Admin/Room/Edit/RoomEditModal.tsx
malaikaaiyar: It wasn't removed in staging (I can open a new pr after this to remove it)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
malaikaaiyar: I added an error section but at the moment it shouldn't validate anything (as there isn't really anything a description shouldn't be; maybe a maximum length could be implemented?). In my opinion, adding the error validation for this is OOS for this PR.

0xdevalias: Agreed that the validation itself is OOS, just wanted to make sure the 'slots' were in place for it.

PR #1196, opened 2021-02-19
Fixes chat message bubble styles

mike-lvov:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/419

### DIFF
--- a/src/components/atoms/ChatMessage/ChatMessage.scss
+++ b/src/components/atoms/ChatMessage/ChatMessage.scss
@@ -6,6 +6,7 @@ $my-text-border-radius: 28px 28px 4px 28px;
 .chat-message {
   align-self: flex-start;
   display: flex;
+  align-items: flex-start;
   flex-direction: column;
   margin-bottom: 16px;
 
@@ -56,6 +57,7 @@ $my-text-border-radius: 28px 28px 4px 28px;
 
   &--me {
     align-self: flex-end;
+    align-items: flex-end;
 
     .chat-message__text {
       text-align: right;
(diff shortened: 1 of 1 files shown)


denisdimitrov (approved):
LGTM!

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -6,6 +6,7 @@ $my-text-border-radius: 28px 28px 4px 28px;
 .chat-message {
   align-self: flex-start;
   display: flex;
+  align-items: flex-start;
```
0xdevalias, line 9: @mike-lvov How does adding `align-items` here differ from `align-self`?

- https://www.w3schools.com/cssref/css3_pr_align-self.asp
  - > The align-self property specifies the alignment for the selected item inside the flexible container.
  - > Note: The align-self property overrides the flexible container's align-items property.
  
Is `align-self` still needed here?

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -56,6 +57,7 @@ $my-text-border-radius: 28px 28px 4px 28px;
 
   &--me {
     align-self: flex-end;
+    align-items: flex-end;
```
0xdevalias, line 60: As above, how does `align-self` and `align-items` differ here?

0xdevalias (commented):
Retroactive review comments.

src/components/atoms/ChatMessage/ChatMessage.scss
mike-lvov: They target different items. Align-self targets the actual class. align-items targets its children

0xdevalias: Right, that makes sense.

PR #1197, opened 2021-02-19
Allows message deletion for an admin

mike-lvov:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/420

### DIFF
--- a/src/hooks/useVenueChat.ts
+++ b/src/hooks/useVenueChat.ts
@@ -15,7 +15,7 @@ import { useFirestoreConnect } from "./useFirestoreConnect";
 import { useVenueId } from "./useVenueId";
 import { useUser } from "./useUser";
 import { useWorldUsersById } from "./users";
-import { useConnectCurrentVenueNG } from "./useConnectCurrentVenueNG";
+import { useRoles } from "./useRoles";
 
 export const useConnectVenueChatMessages = (venueId?: string) => {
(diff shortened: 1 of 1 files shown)


denisdimitrov (approved):
LGTM!

0xdevalias (commented):
We also have `useIsAdminUser` and related hooks in `src/hooks/roles.ts` that would have resolved this is a potentially cleaner/simpler way.

0xdevalias (commented):
@mike-lvov Actually.. further to my last review.. the logic here is actually wrong.. and introduces a security issue.

To be allowed to delete a message, I need to have both the admin role AND be an owner of the current venue, or one of the related parent venues.

The logic as implemented here means that anyone who can create any venue within this environment can delete anyone's messages, even on venues they have no access to.

mike-lvov:
>  introduces a security issue.
> 
> To be allowed to delete a message, I need to have both the admin role AND be an owner of the current venue, or one of the related parent venues.
> 
> The logic as implemented here means that anyone who can create any venue within this environment can delete anyone's messages, even on venues they have no access to.

It doesn't introduce it. It just keeps it :). The previous implementation :

` const allowDelete =
    ((userRoles && userRoles.includes("admin")) ||
      (user && venue?.owners?.includes(user.uid))) ??
    false;`

PR #1202, opened 2021-02-19
Fixes/minor scss fixes

margulies:
Two minor issues I noticed — you might already be on top of these, but thought to send a PR nevertheless

### DIFF
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -7,7 +7,7 @@
   background: $primary;
   top: 66px;
   // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
-  left: calc((100% - #{$map-sidebar-width}) / 2);
+  left: 50%; // calc((100% - #{$map-sidebar-width}) / 2);
   width: 500px;
   margin: 0 auto 0 -250px;
   transition: left 0.1s linear;
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -6,8 +6,7 @@
   position: absolute;
   background: $primary;
   top: 66px;
-  // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
-  left: 50%; // calc((100% - #{$map-sidebar-width}) / 2);
+  left: 50%;
   width: 500px;
   margin: 0 auto 0 -250px;
   transition: left 0.1s linear;
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -286,7 +286,7 @@ $border-radius: 28px;
 
 .schedule-dropdown-body {
   position: fixed;
-  z-index: z(schedule);
+  z-index: z(navbar-schedule);
   top: 0;
   left: 0;
   width: 100%;
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -109,7 +109,6 @@ $z-layers: (
 @function z($layer) {
   @if not map-has-key($z-layers, $layer) {
     @error "No layer found for `#{$layer}` in $z-layers map.";
-    @return 1;
   }
 
   @return map-get($z-layers, $layer);
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -108,7 +108,7 @@ $z-layers: (
 
 @function z($layer) {
   @if not map-has-key($z-layers, $layer) {
-    @warn "No layer found for `#{$layer}` in $z-layers map. Property omitted. Falling back to 1.";
+    @error "No layer found for `#{$layer}` in $z-layers map.";
     @return 1;
   }
 


src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -7,7 +7,7 @@
   background: $primary;
   top: 66px;
   // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
-  left: calc((100% - #{$map-sidebar-width}) / 2);
+  left: 50%; // calc((100% - #{$map-sidebar-width}) / 2);
```
0xdevalias, line 10: @margulies can you please explain what the intent of this change is?

My understanding is that we did this to ensure that the sidebar is centered with the regards to the map (though my guess is this may have changed recently with some of @mike-lvov's chat refactor work in https://github.com/sparkletown/sparkle/pull/1179 related followups)

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -286,7 +286,7 @@ $border-radius: 28px;
 
 .schedule-dropdown-body {
   position: fixed;
-  z-index: z(schedule);
+  z-index: z(navbar-schedule);
```
0xdevalias, line 289: Nice catch!

I expected our `z()` helper to error-out when there isn't a matching key, but seems we only emit a warning, then default it:

- https://github.com/sparkletown/sparkle/blob/staging/src/scss/constants.scss#L109-L116

Could I please get you to also make a change to the `z()` helper to swap `@warn`/the default value for `@error`:

- https://sass-lang.com/documentation/at-rules/error

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
0xdevalias: If that's the case, and we no longer need to take the width into account here, then we should also cleanup the commented bit of code here too.

0xdevalias (changes requested):
A couple of questions and a small request.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
margulies: The issue this is addressing is that with the refactored chat (#1179), the announcement banner now shifts its position depending on whether the chat sidebar is open or closed. I found the user experience better to keep the banner in a stable position. 

If that sounds good, just let me know and I'll clean up the comment.

0xdevalias:
@margulies Reminder to click the 're-request review' button once you make changes so that I know that I need to look at it again.

image

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -6,8 +6,7 @@
   position: absolute;
   background: $primary;
   top: 66px;
-  // NOTE: Currently, the sidebar takes 20% of the screen, thus 40% is half of map's occupation
-  left: calc((100% - #{$map-sidebar-width}) / 2);
+  left: 50%;
```
0xdevalias, line 9: @mike-lvov Can you please confirm that this is in line with your thoughts/expectations around the recent chat refactors.

src/scss/constants.scss
```diff
@@ -108,7 +108,7 @@ $z-layers: (
 
 @function z($layer) {
   @if not map-has-key($z-layers, $layer) {
-    @warn "No layer found for `#{$layer}` in $z-layers map. Property omitted. Falling back to 1.";
+    @error "No layer found for `#{$layer}` in $z-layers map.";
     @return 1;
```
0xdevalias, line 112:
```suggestion
    @error "No layer found for `#{$layer}` in $z-layers map.";
```

0xdevalias: With this `@error` change, if you try to use a `z(doesnt-exist)` or similar, how does the error show? Hopefully in an obvious way.

0xdevalias (changes requested):
One tiny change, and seeking another set of 👀 from @mike-lvov to confirm the fix here.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
mike-lvov: @0xdevalias it's more a question to @mcflyok if we need to center the sidebar, based on the chat's sidebar opened\closed state

mcflyok: @mike-lvov @0xdevalias @margulies Sounds good to make the announcement banner centered counting the presence of the chat sidebar ! Nice one !

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
mike-lvov: https://github.com/sparkletown/sparkle/pull/1202#discussion_r580994391

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
0xdevalias: @mcflyok I think what @margulies is saying is that he found it felt better when the announcement bar was stable, not moving with the chat sidebar.

While I haven’t played with both options, at least in its current state, I think I would tend to agree. The chat sidebar is currently a ‘slide out draw’ that sits above/outside the ‘flow’ of everything else. Therefore for it to ‘push’ the announcement banner when it opens would feel really weird to me.

IMO the sidebar should ‘push’ everything (map included), or push nothing (which is kind of how it is being proposed/implemented in this PR currently). My personal vote is for the latter.

mcflyok: @0xdevalias @margulies ooops sorry guys didnt get it, I'm down to keep it stable and always centered as it is kinda part of the topbar. sorry for the misunderstanding

0xdevalias: No worries. Thanks for your input! :)

src/components/molecules/NavBar/NavBar.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1202/commits/5ae3d61bad5f17cd0190f10b24913231d6ad7bf1

src/scss/constants.scss
0xdevalias: Just checked this, looks good, and blocks the app from starting, so nice and obvious!

image

---

Removed the extra `@return` in https://github.com/sparkletown/sparkle/pull/1202/commits/f19c6c45d102faaface23b02282aec6b2db32c5e

0xdevalias (approved):
This looks good to go now! Thanks for the contribution @margulies :) :shipit:

PR #1204, opened 2021-02-22
Can still hear emojis even though they are muted 

jigar-jstechno:
Updated JazzTab with if isAudioEffectDisabled is true it will mute emoji audio.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/310

### DIFF
--- a/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
+++ b/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
@@ -170,7 +170,9 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
                       key={reaction.name}
                       className="reaction"
                       onClick={() =>
-                        user && reactionClicked(user, reaction.type)
+                        user &&
+                        !isAudioEffectDisabled &&
+                        reactionClicked(user, reaction.type)
                       }
                       id={`send-reaction-${reaction.type}`}
                     >


0xdevalias (changes requested):
This code as it currently exists will prevent a user from sending any reaction when they are muted, which isn't the desired effect here.

It also won't stop any reactions sent from other users from causing sounds to play.

---

You can see the definition of `reactionClicked` below, it basically just dispatches the `addReaction` Redux action:

- https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx#L80-L88

Searching for this, we can see that it is used in 3 files (the current file, and 2 more):

- https://github.com/sparkletown/sparkle/search?q=addReaction
  - https://github.com/sparkletown/sparkle/blob/staging/src/store/actions/Reactions.ts
    - Defines the action as a redux thunk, basically just updates the firestore DB
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Audience/Audience.tsx#L172-L183
    - Another component that also has the ability to send reactions
    
In normal circumstances I would have expected that to show me more useful files that relate to the actually 'playing' of the reactions, but since it is 'disconnected' from the redux action used to send it, we have to look deeper.

The data is stored in `experiences/${venueId}/reactions` within the firestore DB:

- https://github.com/sparkletown/sparkle/blob/staging/src/store/actions/Reactions.ts#L22

Looking at our core `Firestore` types, we can see that `Experience` is defined, and includes `reactions`:

- https://github.com/sparkletown/sparkle/blob/staging/src/types/Firestore.ts#L15-L18

Searching around the codebase.. I'm not even sure if that is used anywhere.. the closes I could find is the `useReactions` hook, which doesn't actually store anything in redux:

- https://github.com/sparkletown/sparkle/blob/staging/src/hooks/useReactions.ts

Which for some reason is included within `UserProfilePicture`:

- https://github.com/sparkletown/sparkle/search?q=useReactions
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/UserProfilePicture/UserProfilePicture.tsx#L68

We can see that the code that renders the reactions (and plays the audio) is rather hacky/suboptimal (it should use a similar pattern for playing audio as was used in https://github.com/sparkletown/sparkle/pull/1180 (not conditionally rendering the `<audio>` element like it currently is), but it seems as though it should already be muting, based on `!muteReactions && !isAudioEffectDisabled`:

- https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/UserProfilePicture/UserProfilePicture.tsx#L107-L129

`muteReactions` is set by `useSelector((state) => state.room.mute)` (which should be extracted into a selector function within `src/utils/selectors.ts`

`isAudioEffectDisabled` is passed into this component as a prop, so we'll need to search for usages across the app to figure out what that is set to/controlled by. While it is set in a bunch of other components, these are the ones that actually set it using state (rather than statically):

- https://github.com/sparkletown/sparkle/search?q=isAudioEffectDisabled
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Audience/Audience.tsx#L321-L329
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx#L182-L191
  
In both of these cases, it's controlled by local `useState` within that component.

Since the `UserProfilePicture` component is used in tons of places all across the application, and most of them aren't likely to be passing in the optional but not defaulted `isAudioEffectDisabled` prop, that is probably the root cause of this issue.

The quick fix here would probably be to add a default value to that prop, ideally setting it to `true` so that audio is muted in all of those locations, except where we explicitly request it.

jigar-jstechno:
> reactionClicked

Can we add withAudio : Boolean in utils/reactions.ts and every time saving to Firestore DB we save this value also so when we get reaction to other users it play sount if true else mute

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1204#issuecomment-783871588
> Can we add withAudio : Boolean in utils/reactions.ts and every time saving to Firestore DB we save this value also so when we get reaction to other users it play sound if true else mute

@jigar-jstechno Muting the sound is a thing that users will do locally. So it shouldn't be sent to firestore/saved there at all.

---

I explained the seeming root cause of the issue, and the probable fix, in my review in https://github.com/sparkletown/sparkle/pull/1204#pullrequestreview-595986840

jigar-jstechno:
> RE: [#1204 (comment)](https://github.com/sparkletown/sparkle/pull/1204#issuecomment-783871588)
> 
> > Can we add withAudio : Boolean in utils/reactions.ts and every time saving to Firestore DB we save this value also so when we get reaction to other users it play sound if true else mute
> 
> @jigar-jstechno Muting the sound is a thing that users will do locally. So it shouldn't be sent to firestore/saved there at all.
> 
> @jigar-jstechno I explained the seemed root cause of the issue, and the probable fix, in my review in [#1204 (review)](https://github.com/sparkletown/sparkle/pull/1204#pullrequestreview-595986840)

that is the solution for the same user but for other user audio will come... is it fine?

0xdevalias:
> that is the solution for the same user but for other user audio will come... is it fine?

As per the first comment I made in https://github.com/sparkletown/sparkle/pull/1204#issuecomment-783875041:

> Muting the sound is a thing that users will do locally. So it shouldn't be sent to firestore/saved there at all.

If a user who isn't sending the reaction wants to mute themselves, then they can just click the mute button, which will store that in their local state. The existing logic will then prevent the sound from being played for them. It's not a 'per reaction' mute, it's 'mute all reactions that I receive'.

PR #1208, opened 2021-02-23
add optional maxVisible prop to RoomAttendance

0xdevalias:
Something I was playing around with as part of a different PR/code review, figured it was worth committing.

While it's not currently used, since it is an optional prop that is defaulted, it also shouldn't cause any issues.

### DIFF
--- a/src/components/templates/PartyMap/components/RoomAttendance.tsx
+++ b/src/components/templates/PartyMap/components/RoomAttendance.tsx
@@ -1,10 +1,11 @@
 import React, { useMemo } from "react";
 
-import { useRoom } from "hooks/useRoom";
 import { DEFAULT_ROOM_ATTENDANCE_LIMIT } from "settings";
 
 import { Room } from "types/rooms";
 import { PartyMapVenue } from "types/venues";
+
+import { useRoom } from "hooks/useRoom";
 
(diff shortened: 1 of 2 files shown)


src/components/templates/PartyMap/components/RoomAttendance.tsx
```diff
@@ -10,35 +10,36 @@ import "./RoomAttendance.scss";
 type RoomAttendanceProps = {
   venue: PartyMapVenue;
   room: Room;
+  maxVisible?: number;
 };
 
 const MAX_AVATARS_VISIBLE = 2;
```
denisdimitrov, line 16: [NB] This can be moved to `settings.ts`

denisdimitrov (approved):

src/components/templates/PartyMap/components/RoomAttendance.tsx
0xdevalias: Good choice, I agree.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1208/commits/39516c38d1632bac156701a29bcd7644b789de24

PR #1209, opened 2021-02-23
use React.Fragment instead of span in getLinkFromText

0xdevalias:
use `React.Fragment` instead of `span` in `getLinkFromText`, as previously it would create some really ugly HTML output that was hard to play around with when tweaking styling/etc in devtools.

## Before

image

## After

image

### DIFF
--- a/src/utils/getLinkFromText.tsx
+++ b/src/utils/getLinkFromText.tsx
@@ -19,6 +19,6 @@ export const getLinkFromText = (text: string) => {
           {word}{" "}
         </a>
       );
-    else return <span key={index}>{word} </span>;
+    else return <React.Fragment key={index}>{word} </React.Fragment>;
   });
 };


denisdimitrov (approved):
LGTM!

src/utils/getLinkFromText.tsx
```diff
@@ -19,6 +19,6 @@ export const getLinkFromText = (text: string) => {
           {word}{" "}
         </a>
       );
-    else return <span key={index}>{word} </span>;
+    else return <React.Fragment key={index}>{word} </React.Fragment>;
```
mike-lvov, line 22: Do we want to have `<React.Fragment>` or plain `<></>` is enough?

0xdevalias: @mike-lvov I believe `React.Fragment` is required here, since we're using an array, and it's the only fragment form that supports the `key` prop:

- https://reactjs.org/docs/fragments.html#keyed-fragments

PR #1213, opened 2021-02-23
fix admin room edit urls

denisdimitrov:
The same api call is being used for both update and create of the room and since `useUrl` is being passed only on creation, it was missing when updating and the check was falling back to `window.origin + venueInsideUrl(input.venueName!)` which was always domain name + /undefined

Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/428#issuecomment-783598157


0xdevalias (approved):
[CRO] I'll admit that this is a bit of a blind review/approve.. I tried to dig into the logic to confirm this would fix it, but it is SUPER hacky and hard to follow. Definitely needs some future cleanup/simplification IMO.

If you've tested this and are happy that it will resolve the issue, happy for you to merge.

0xdevalias:
@denisdimitrov This PR seems to have been approved for almost a week now. Can you please get it merged in when you're able.

PR #1215, opened 2021-02-24
restore ReactionPage + some related cleanup to reactions/similar

0xdevalias:
The `ReactionPage` was lost in https://github.com/sparkletown/sparkle/pull/1105

This PR restores the page, as well as cleaning it + a bunch of little related things up in the process.

See the [commits](https://github.com/sparkletown/sparkle/pull/1215/commits) in this PR for a full list of changes, they are broken down into contextual/descriptive chunks.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/442

Note: The general look/feel/styling/etc of the `ReactionPage` is out of scope for this PR. It can have a separate design pass in future to make it look a lot better. This PR is more about restoring the core functionality.


src/components/templates/ReactionPage/ReactionPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { ReactionList } from "./ReactionList";
+
+import "./ReactionPage.scss";
+
+const wantedReactionsSelector = SHOW_EMOJI_IN_REACTION_PAGE
+  ? reactionsSelector
+  : messagesToTheBandSelector;
+
+export const ReactionPage = () => {
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1215/commits/2d17f3a6d604cfd1d7567c7687e5887fc566c574

src/hooks/users.ts
```diff
@@ -69,6 +70,19 @@ export const useWorldUsersById = () => {
...
+export const useWorldUsersByIdWorkaround = () => {
+  const {
+    worldUsersById: _worldUsersById,
+    isWorldUsersLoaded,
+  } = useWorldUsersById();
+
+  const worldUsersById: Partial<Record<string, User>> = _worldUsersById;
+
+  return { worldUsersById, isWorldUsersLoaded };
+};
```
0xdevalias, line 84: @mike-lvov FYI. Temporary 'centralised' workaround until we refactor all the related code to work well with potentially undefined results.

src/store/actions/Room.ts
```diff
@@ -1,20 +1,18 @@
...
+export interface PreviewRoomAction {
   type: typeof PREVIEW_ROOM;
   room: string;
 }
 
-interface ToggleMuteAction {
-  type: typeof ToggleButtonGroup;
+export interface ToggleMuteAction {
+  type: typeof TOGGLE_MUTE_REACTIONS;
 }
```
0xdevalias, line 16: [note] I'm not sure this was ever a valid type for this to have. In any case, this part of the code doesn't even seem to be used at the moment, so making this change shouldn't affect anything. I didn't delete it outright because moving to more of a Redux-based pattern for muting is potentially an ideal direction to go in.

src/types/utility.ts
```diff
@@ -22,3 +22,8 @@ export type Bounds = Point & Dimensions;
 export type Matrix<T> = T[][];
 
 export type ReactHook<T, U> = (props: T) => U;
+
+/**
+ * @deprecated use Partial<Record<K, T>> directly
+ */
+export type PartialRecord<K extends keyof never, T> = Partial<Record<K, T>>;
```
0xdevalias, line 29: [note] This is here as a reminder/hint as to how to properly create a partial record.

src/components/templates/ReactionPage/ReactionPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+      ? {
+          collection: "experiences",
+          doc: currentVenue.id,
+          subcollections: [{ collection: "reactions" }],
+          orderBy: ["created_at", "desc"],
+          storeAs: "reactions",
+        }
+      : undefined
+  );
+  const reactions = useSelector(wantedReactionsSelector) ?? [];
```
0xdevalias, line 42: [note] While this would probably be good to do at some stage, I didn't want to deal with the specifics/edge cases of useReactions, and how that may impact other areas of the app unknowingly. The risk didn't seem worth the reward in this PR. So this part is OOS.

0xdevalias (commented):
A few notes/guide to some of the changes in this PR.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -64,10 +78,18 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
         </p>
       );
       break;
+
+    // Note: This is the template that is used for the Auditorium
     case VenueTemplate.audience:
-      template = <Audience />;
+      template = (
+        <Switch>
+          <Route path={`${match.path}/reactions`} component={ReactionPage} />
```
denisdimitrov, line 86: What's the purpose of this screen? I've never heard of this screen being used by anyone and that's why I removed it in my PR.

0xdevalias: You removed the routes to it, the components still existed.

This displays all of the reactions/shoutouts/etc that have been shared in the respective venue, in a way that allows 'the band' or equivalent to easily see them on a screen, not just few the couple of seconds that they flash up within the venue itself.

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
```diff
@@ -1,13 +1,15 @@
 import { User } from "types/User";
 import { WithId } from "utils/id";
 
+// @debt follow our standard patterns and move this back into ./UserProfilePicture.tsx
```
mike-lvov, line 4: Is it possible to move it in this PR, since it's basically a one-minute thing?

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+import { useWorldUsersByIdWorkaround } from "hooks/users";
+
+import UserProfileModal from "components/organisms/UserProfileModal";
+import UserProfilePicture from "components/molecules/UserProfilePicture";
+import { UserAvatar } from "components/atoms/UserAvatar";
+
+export interface ReactionListProps {
+  reactions: Reaction[];
+  chats: ChatMessage[];
```
mike-lvov, line 24: should this be chatMessages/similar?

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -0,0 +1,107 @@
...
+  small?: boolean;
```
mike-lvov, line 25: ?should we refactor this into `size`?

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -0,0 +1,107 @@
...
+    WithId<User>
+  >();
+
+  const allReactions = useMemo(() => {
+    const chatsAsBandMessages = chats?.map(chatMessageAsMessageToTheBand) ?? [];
+
+    const allReactionsSorted = [
+      ...(reactions ?? []),
+      ...chatsAsBandMessages,
+    ].sort((a, b) => b.created_at - a.created_at);
```
mike-lvov, line 46: could we have `chatSort` here?

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -0,0 +1,107 @@
...
+}) => {
+  // @debt see comments in useWorldUsersByIdWorkaround
+  const { worldUsersById } = useWorldUsersByIdWorkaround();
+
+  const [selectedUserProfile, setSelectedUserProfile] = useState<
+    WithId<User>
+  >();
+
+  const allReactions = useMemo(() => {
+    const chatsAsBandMessages = chats?.map(chatMessageAsMessageToTheBand) ?? [];
```
mike-lvov, line 41: wrap into useMemo?

src/types/utility.ts
mike-lvov: Why do we need a partial Record, if we can have `{ [key: type]: type }`?

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
0xdevalias: I didn’t want to as I assumed that would touch a bunch of files that could import this. Didn’t actually check if there were other files that did. Was trying to stick to the ‘focussed PR’ thing (aka not scope creeping), while also pointing out issues as I came across them.

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: Could be, happy to change it. This is just how it was defined in the existing component.

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: If we did, how would you envisage it looking? Presumably wouldn’t be a Boolean anymore if we did that?

0xdevalias: I think since there’s only 1 option with it, it’s probably fine as is.

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: I did think about that, but then didn’t go and look if it would work for this. I believe these are technically ‘messages to the band’ format, so it may not work for it

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: Wrap what into useMemo? There’s already a useMemo on this line?

src/types/utility.ts
0xdevalias: What does that type do, and how would it support this usecase? How does it compare/contrast to a Record?

Record<K,T> says it will contain mappings from K to T.

Partial says they may be undefined.

mike-lvov: off top of my head `{ [key: K]: T }` states that it is an object and if you index it with `K` it will return `T or undefined`

src/components/templates/ReactionPage/ReactionList.tsx
mike-lvov: missed that^

src/components/templates/ReactionPage/ReactionList.tsx
mike-lvov: I guess since it uses the same `Message` type, it should

src/components/templates/ReactionPage/ReactionList.tsx
mike-lvov: `If we did, how would you envisage it looking? Presumably wouldn’t be a Boolean anymore if we did that?`

size would be `ReactionSize`. I didn't dig into the logic, but I guess it only is used in styles, so dynamic classNames would do the trick here

src/pages/VenuePage/TemplateWrapper.tsx
mike-lvov: But I guess since no one navigates to `/reactions`( or does anyone?), this ReactionPage component is never rendered

0xdevalias: This used to be /band, but I renamed it in this PR when I restored it. No one navigates to it currently because they can’t. This is the same kind of feature as how the banner admin page works.

There is zero harm in having it restored, and given it’s a feature that is done/exists/etc already, it feels pointless to spend additional time questioning/pondering/etc over it.

src/types/utility.ts
0xdevalias: Looking at the TypeScript docs, I think you're referring to 'mapped types':

- https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types

It basically seems like a 'type metaprogramming' method for transforming a type based on a set of 'rules'.

As examples of things that can be implemented using mapped types, they directly mention `Partial` (and some others)

We can see that `Partial` is included as a default utility type (presumably implemented on top of the basic 'mapped types' concept):

- https://www.typescriptlang.org/docs/handbook/utility-types.html

If I click through to see the implementation of `Partial` in my IDE, it looks like the following, confirming that it is just a 'wrapper' around the 'mapped types' concept:

`node_modules/typescript/lib/lib.es5.d.ts:1444`
```
/**
 * Make all properties in T optional
 */
type Partial<T> = {
    [P in keyof T]?: T[P];
};
```

---

So, long story short, `Partial` is already the canonical implementation/abstraction of what you mentioned above.

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: It doesn't though, which is what I meant by the following above:

> I believe these are technically ‘messages to the band’ format,

image

image

So the types are `Reaction` and `MessageToTheBandReaction`, which since `Reaction` is the generic 'grouping' type, will be what they both become.

This is defined in `src/utils/reactions.ts`:

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/reactions.ts#L86-L99

and seems to use `created_at`, which is a `number`.

`chatSort` works on `ts_utc`, which is a `firebase.firestore.Timestamp`:

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/chat.ts#L15-L18

---

@mike-lvov So, roundabout path to a researched answer, no we can't use `chatSort` here.

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
0xdevalias: IDE says that it's only used in the 1 place, so yeah, happy to move it in this PR:

image

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1215/commits/e65dcb96777870bc51317ace331c43f2c5ce4b0f

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1215/commits/46b090f79f949a09c4a68254188b947da2b058a6

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: Right, so an enum then. Yeah, if we had more than one option, that would definitely be the way to go, but since we don't, I don't think it matters a whole lot here. Feel free to unresolve this if you feel otherwise.

src/types/utility.ts
0xdevalias: Resolving this based on the above, if you still feel that there is a distinction/reason we should be doing it differently, feel free to re-open.

mike-lvov (approved):
Good to go

PR #1218, opened 2021-02-24
refactor UserProfileModal state into redux + render component in a single canonical location

igachmovmelon:
The UserProfileModal component show and hide logic is moved to redux.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/425


0xdevalias (changes requested):
Only had a very quick skim through this so far, so may have more detailed comments later, but it looks as though there are still a lot of cases where local useState is used in components; that shouldn’t be required at all. The components should consume the state from redux, and ‘manipulate’ the state only through dispatching the redux action (for open/close).

It also looks as though we still have the modal component spread across the app, whereas ideally we want it only in a single canonical location. For the user-facing side of things, this will most likely be in the TemplateWrapper component or similar (though I’m open to other suggestions). I don’t believe TemplateWrapper would work for the admin side of things, so need to check/confirm where the modal needs to be triggered from and ensure it exists in that ‘tree’ accordingly. It may be that we have it added in a couple of locations (eg. TemplateWrapper and something for the admin side of things). The intent is basically that it would be ‘high enough’ in the ‘component tree’ that it will be ‘available’ in all of the current places that are trying to trigger it, rather than added individually in each of those locations.

it might also be useful to wrap the useDispatch, useSelector, and some useCallback’d helper methods in a custom hook, and call it something like useProfileModalControls (or something similarly ‘semantically named’), we would then call that in each component that needs to open the modal.

src/store/reducers/UserProfile.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  UserProfileActions,
+  UserProfileActionTypes,
+} from "store/actions/UserProfile";
+
+export type UserProfileState = {
+  isUserProfileVisible: boolean;
```
0xdevalias, line 7: I would probably just simplify this to `isVisible` as well. We already know that it's in the context of the `UserProfileState`, so adding that again in the name here feels redundant.

0xdevalias (commented):
Extra comment on top of my review this morning.

src/components/molecules/OnlineStats/OnlineStats.tsx
```diff
@@ -78,6 +84,17 @@ const OnlineStats: React.FC = () => {
...
+  const isUserProfileModalVisible = useSelector(
+    userProfileModalVisibilitySelector
+  );
+
+  const setUserProfileModalVisible = useCallback(() => {
+    if (isUserProfileModalVisible) {
+      setSelectedUserProfile(undefined);
+    }
+    dispatch(setUserProfileModalVisibility(!isUserProfileModalVisible));
+  }, [dispatch, isUserProfileModalVisible]);
```
mike-lvov, line 97: I would see this moved into `useProfileModalHook`, which would be quite similar to useChatSidebarControls. It would return the state, turn off/ turn on functions/ maybe a toggler?

mike-lvov (changes requested):
Good start there, would love to see these changes implemented for further discussion!

src/components/molecules/OnlineStats/OnlineStats.tsx
0xdevalias: As per my comments in https://github.com/sparkletown/sparkle/pull/1218#pullrequestreview-597998714 most of this is likely not required. I spelled out the custom hook/etc stuff as well.

igachmovmelon:
@mike-lvov 
@0xdevalias 

I have made the suggested changes for this report.

src/components/molecules/OnlineStats/OnlineStats.tsx
```diff
@@ -304,7 +303,9 @@ const OnlineStats: React.FC = () => {
                         <div
                           key={index}
                           className="user-row"
-                          onClick={() => setSelectedUserProfile(user)}
+                          onClick={() => {
+                            setUserProfile(user);
+                          }}
```
mike-lvov, line 308: I don't think you need to wrap this into { }

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -82,21 +76,16 @@ export const ChatSidebar: React.FC = () => {
         </div>
         <div className="chat-sidebar__tab-content">
           {chatSettings.openedChatType === ChatTypes.VENUE_CHAT && (
-            <VenueChat onAvatarClick={setSelectedProfile} />
+            <VenueChat onAvatarClick={setUserProfile} />
```
mike-lvov, line 79: I don't think you need to prop-drill it. You can just use the hook inside the components

src/components/organisms/Room/Participant.tsx
```diff
@@ -134,16 +136,13 @@ const Participant: React.FC<React.PropsWithChildren<ParticipantProps>> = ({
         <div className="profile-icon">
           <UserProfilePicture
             user={{ ...profileData, id: participant.identity }}
-            setSelectedUserProfile={() => setShowProfile(true)}
+            setSelectedUserProfile={() =>
+              setUserProfile({ ...profileData, id: participant.identity })
```
mike-lvov, line 140: we have `withId` function that does this. Try using that

src/hooks/useProfileModal.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from "react";
+
+import { userProfileSelector } from "utils/selectors";
+
+import { setUserProfileData } from "store/actions/UserProfile";
+
+import { useSelector } from "./useSelector";
+import { useDispatch } from "./useDispatch";
+import { User } from "types/User";
+import { WithId } from "utils/id";
```
mike-lvov, line 10: please, sort these according to `VenuePage.tsx`

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -15,13 +15,16 @@ import { WithNavigationBar } from "components/organisms/WithNavigationBar";
...
+import UserProfileModal from "components/organisms/UserProfileModal";
+import { useProfileModal } from "hooks/useProfileModal";
 
 type TemplateWrapperProps = {
   venue: Venue;
 };
 
 const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
   const history = useHistory();
+  const { selectedUserProfile, setUserProfile } = useProfileModal();
```
mike-lvov, line 27: You can also have `unsetUserProfile` inside this hook. This way you will not need to expose the logic of how you do it below.

src/store/reducers/UserProfile.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "store/actions/UserProfile";
+import { User } from "types/User";
+import { WithId } from "utils/id";
+
+export type UserProfileState = {
+  userProfile: WithId<User> | undefined;
+};
+
+const initialChatState: UserProfileState = {
+  userProfile: undefined,
```
mike-lvov, line 13: You can just have an empty object here

mike-lvov (commented):
Overall good. We are almost there. I have pointed out a few more things, please make sure you find all instances of those and fix it everywhere

src/hooks/useProfileModal.ts
0xdevalias: Roughly speaking the logic of the order goes from external/generic down to internal/specific; which roug equates to the following order:

- react
- other 3rd party libs
- [space]
- 'root imports' (eg. settings, etc)
- [space]
- imports from `types/`
- [space]
- imports from `utils/`
- [space]
- imports from `hooks/`
- [space]
- imports from `pages/`
- [optional space, sort of depends how many there are]
- imports from `components/templates`
- [optional space, sort of depends how many there are]
- imports from `components/organisms`
- [optional space, sort of depends how many there are]
- imports from `components/molecules`
- [optional space, sort of depends how many there are]
- imports from `components/atoms`
- [space]
- imports from `../` (if any.. and mostly only ever allowed to go up by '1 folder', but we always preference using the full import path for the above rather than going up 1 dir to it
- [space]
- imports from `./`
- [space]
- scss/other asset imports from `./`

This might not be 100% complete.. but that's generally how I go about doing it. (we have plans to enforce this with linting when I get the chance to implement it)

src/store/reducers/UserProfile.ts
0xdevalias: @mike-lvov Why do you suggest that rather than `undefined`? I'm not a fan of 'shadowing' 'empty' types with things like this (empty object, empty string, etc) where we can more correctly use `undefined`/etc to represent them (and then gain access to all of the related typing/etc tools that allow us to easily assert when things are 'empty in a direct way.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -155,12 +153,6 @@ const NavSearchBar = () => {
...
 
-      {/* @debt use only one UserProfileModal instance with state controlled with redux  */}
-      <UserProfileModal
-        userProfile={selectedUserProfile}
-        show={selectedUserProfile !== undefined}
-        onHide={() => setSelectedUserProfile(undefined)}
-      />
       {/* @debt use only one RoomModal instance with state controlled with redux */}
       <RoomModal
         show={hasSelectedRoom}
```
0xdevalias, line 158: [OOS] But perhaps this can be the next task we take on? As it follows the same pattern as this one.

src/components/molecules/OnlineStats/OnlineStats.tsx
```diff
@@ -345,8 +347,7 @@ const OnlineStats: React.FC = () => {
         <OverlayTrigger
           trigger="click"
           placement="bottom-end"
-          overlay={popover}
-          rootClose={!selectedUserProfile} // allows modal inside popover
+          overlay={popover} // allows modal inside popover
```
0xdevalias, line 350: Are these changes correct? Since removing the `UserProfileModal` from here are these needed anymore?

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -60,7 +54,9 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
             <div
               className="row result-user"
               key={user.id}
-              onClick={() => onSelect(user)}
+              onClick={() => {
+                setUserProfile(user);
+              }}
```
0xdevalias, line 59: Where does `onSelect` come from? And why are we not using it here anymore? Should we be removing other code related to `onSelect`? This feels like it may not be correct?

Also, don't need to wrap in the extra `{` `}`'s

src/components/molecules/UserSearchBar/UserSearchBar.tsx
```diff
@@ -75,13 +71,7 @@ const UserSearchBar: FC<UserSearchBarProps> = ({ onSelect }) => {
...
       )}
-      <UserProfileModal
-        userProfile={selectedUserProfile}
-        show={!!selectedUserProfile}
-        onHide={clearSelectedUserProfile}
-      />
     </div>
   );
 };
-
```
0xdevalias, line 86: Shouldn't delete this line.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
0xdevalias: 👍🏻

src/components/templates/Audience/Audience.tsx
```diff
@@ -421,9 +419,7 @@ export const Audience: React.FunctionComponent = () => {
                                     isOnRight ? "left" : "right"
                                   }
                                   avatarClassName={"profile-avatar"}
-                                  setSelectedUserProfile={
-                                    setSelectedUserProfile
-                                  }
+                                  setSelectedUserProfile={setUserProfile}
```
0xdevalias, line 422: I expect that this prop can probably be removed entirely, and just use the hook directly within the component.

src/components/templates/Jazzbar/components/ReactionList.tsx
```diff
@@ -63,7 +61,7 @@ const ReactionList: React.FC<ReactionListProps> = ({
             <img
               onClick={() =>
                 worldUsersById[message.created_by] &&
-                setSelectedUserProfile({
+                setUserProfile({
                   ...worldUsersById[message.created_by],
                   id: message.created_by,
                 })
```
0xdevalias, line 67: `withId` helper

src/components/templates/Jazzbar/components/ReactionList.tsx
```diff
@@ -52,6 +49,7 @@ const ReactionList: React.FC<ReactionListProps> = ({
   const profileImageSize = small
     ? REACTION_PROFILE_IMAGE_SIZE_SMALL
     : REACTION_PROFILE_IMAGE_SIZE_LARGE;
+
   return (
     <>
       <div className={`reaction-list ${small && "small"}`}>
```
0xdevalias, line 55: In all cases where there is only going to be a single root `<div>`/etc element now that we've removed the modal, we'll no longer need to wrap it with this fragment (`<>`)

src/components/templates/Playa/Playa.tsx
```diff
@@ -55,7 +55,8 @@ import { useFirestoreConnect } from "hooks/useFirestoreConnect";
 
 import { DustStorm } from "components/organisms/DustStorm/DustStorm";
 import { SchedulePageModal } from "components/organisms/SchedulePageModal/SchedulePageModal";
-import UserProfileModal from "components/organisms/UserProfileModal";
+
+import { useProfileModal } from "hooks/useProfileModal";
```
0xdevalias, line 59: Sort this with the other `hooks/` imports

src/components/templates/Playa/Playa.tsx
```diff
@@ -803,7 +804,7 @@ const Playa = () => {
         movingLeft={movingLeft}
         movingRight={movingRight}
         setMyLocation={setMyLocation}
-        setSelectedUserProfile={setSelectedUserProfile}
+        setSelectedUserProfile={setUserProfile}
```
0xdevalias, line 807: Technically we should probably just remove this prop entirely + use the hook within this component. but `Playa` is a legacy template that is in the process of being deleted, and `AvatarLayer` seems to only be called by it, so no need to worry about that here.

src/components/templates/Playa/Playa.tsx
```diff
@@ -1043,7 +1044,7 @@ const Playa = () => {
           className={`playa-videochat ${inVideoChat ? "show" : ""}`}
           style={{ height: videoChatHeight }}
         >
-          <VideoChatLayer setSelectedUserProfile={setSelectedUserProfile} />
+          <VideoChatLayer setSelectedUserProfile={setUserProfile} />
```
0xdevalias, line 1047: Technically we should probably just remove this prop entirely + use the hook within this component. but `Playa` is a legacy template that is in the process of being deleted, and `VideoChatLayer` seems to only be called by it, so no need to worry about that here.

src/hooks/useProfileModal.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from "react";
```
0xdevalias, line 1: Can we call this file + exported hook `useProfileModalControls` or similar?

src/hooks/useProfileModal.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useDispatch } from "./useDispatch";
+import { User } from "types/User";
+import { WithId } from "utils/id";
+
+export const useProfileModal = () => {
+  const dispatch = useDispatch();
+  const selectedUserProfile = useSelector(userProfileSelector);
+
+  const setUserProfile = useCallback(
+    (userProfile: WithId<User> | undefined) => {
```
0xdevalias, line 17: We use the `?` shorthand rather than explicit `| undefined` in params like this:
```suggestion
  const setUserProfile = useCallback(
    (userProfile?: WithId<User>) => {
```

0xdevalias: It might be worth including a `useCallback`'d `unsetUserProfile` here as well.

0xdevalias: Which it seems @mike-lvov already mentioned in https://github.com/sparkletown/sparkle/pull/1218/files#r584710069

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: And it should be wrapped in `useCallback` so that it has a stable function reference when passed as props to `onHide`/etc, to prevent unnecessary re-renders.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -90,6 +93,11 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
...
   return (
     <WithNavigationBar fullscreen={fullscreen}>
+      <UserProfileModal
+        userProfile={selectedUserProfile}
+        show={selectedUserProfile !== undefined}
+        onHide={() => setUserProfile(undefined)}
+      />
       <AnnouncementMessage message={venue?.bannerMessage} />
       {template}
       <ChatSidebar />
```
0xdevalias, line 103: [nit] I tend to put modals at the bottom of things, both to 'keep them out of the way' and because 'last items' in render order tend to appear on top of earlier items (not that that would make a difference here)

```suggestion
      <AnnouncementMessage message={venue?.bannerMessage} />
      {template}
      <ChatSidebar />
      <UserProfileModal
        userProfile={selectedUserProfile}
        show={selectedUserProfile !== undefined}
        onHide={() => setUserProfile(undefined)}
      />
```

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,22 @@
...
+  UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  { userProfile: WithId<User> | undefined }
+>;
+
+export const setUserProfileData = (
+  userProfile: WithId<User> | undefined
+): SetUseProfileDataAction => ({
+  type: UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  payload: { userProfile },
+});
```
0xdevalias, line 20: Following the Redux styleguide, we shouldn't write our actions as 'setters' like this:

- https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters

(I realise this is how most of the existing code is written, but it's legacy and we want to change/remove that, so best not to add to the tech debt when we can help it)

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export type SetUseProfileDataAction = ReduxAction<
+  UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  { userProfile: WithId<User> | undefined }
+>;
+
+export const setUserProfileData = (
+  userProfile: WithId<User> | undefined
+): SetUseProfileDataAction => ({
```
0xdevalias, line 17:
```suggestion
export const setUserProfileData = (
  userProfile?: WithId<User>
): SetUseProfileDataAction => ({
```

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,22 @@
...
+import { User } from "types/User";
+import { WithId } from "utils/id";
+
+export enum UserProfileActionTypes {
+  SET_USER_PROFILE_DATA = "SET_USER_PROFILE_DATA",
+}
+
+export type SetUseProfileDataAction = ReduxAction<
+  UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  { userProfile: WithId<User> | undefined }
```
0xdevalias, line 12:
```suggestion
  { userProfile?: WithId<User> }
```

src/store/reducers/UserProfile.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  UserProfileActions,
+  UserProfileActionTypes,
+} from "store/actions/UserProfile";
+import { User } from "types/User";
+import { WithId } from "utils/id";
+
+export type UserProfileState = {
+  userProfile: WithId<User> | undefined;
```
0xdevalias, line 9:
```suggestion
  userProfile?: WithId<User>;
```

src/store/reducers/UserProfile.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const userProfileReducer = (
+  state = initialChatState,
+  action: UserProfileActions
+): UserProfileState => {
+  switch (action.type) {
+    case UserProfileActionTypes.SET_USER_PROFILE_DATA:
+      const { userProfile } = action.payload;
+      return { ...state, userProfile };
+    default:
```
0xdevalias, line 24:
```suggestion
      return { ...state, userProfile };

    default:
```

src/store/reducers/index.ts
```diff
@@ -13,6 +14,7 @@ export const VenueTemplateReducers: { [key: string]: Reducer } = {};
 export const MiscReducers = {
   room: roomReducer,
   chat: chatReducer,
+  userProfile: userProfileReducer,
   location: locationReducer,
   attendance: attendanceReducer,
   sovereignVenue: sovereignVenueReducer,
```
0xdevalias, line 20: While it's not true currently, it would be good to keep these in alphabetic order.

0xdevalias (changes requested):
Additional comments/suggestions on top of @mike-lvov's review.

src/store/reducers/UserProfile.ts
mike-lvov: You already have the type of this state a few lines above. I don't think of this as shadowing since you have direct access to the type of this state. But nvm, I don't feel strongly about this

0xdevalias: Yeah, but the type above is `WithId<User>` or `undefined`, and you're suggesting we default it to `{}` which is neither of those things. That's what i'm saying is the issue here.

0xdevalias:
@IvanBuljovcic Please also make sure to resolve any previous review comments once you have actioned them, and ideally comment in that thread that they have been fixed, and what commit it was done in (as a point of reference to help the re-reviewer)

src/components/molecules/OnlineStats/OnlineStats.tsx
igachmovmelon: I can't tell, because this component OnlineStats is not used anywhere in the project. But it's OnlineStats.scss is used.

src/components/molecules/UserSearchBar/UserSearchBar.tsx
igachmovmelon: Same as OnlineStats it is not used anywhere. Again I am not sure.

src/components/molecules/OnlineStats/OnlineStats.tsx
0xdevalias: It looks like `OnlineStats` and the corresponding `.scss` file are related to the Playa template, which is legacy code and on the way to being deleted. So it looks like any concerns related to that aren't relevant here.

src/components/molecules/UserSearchBar/UserSearchBar.tsx
0xdevalias: From a quick search, it seems like `UserSearchBar` and related aren't actually used by anything at the moment, so I assume they also relate to legacy code.

As a general rule though, since there is still an `onSelect` property being passed to the component, I would consider removing it (particularly when you don't know what it does/why) to be an unrelated change; so in this instance, I would probably expect it to still be called within this `onClick` handler. If we had better information (eg. this component was actually used somewhere) then we could make a better decision as to whether we should remove that prop entirely or not. But since we can't, I would say add the `onSelect(user)` call back here, as well as the `setUserProfile(user)` call.

igachmovmelon:
@0xdevalias @mike-lvov

0xdevalias:
@igachmovmelon Did you rebase this branch, or push the wrong branch to it or something? It looks like maybe you merged master into it?

3782BE30-AC56-4EE6-B722-9E0AABD1CDA4

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -1,35 +1,29 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react";
+
+// Typings
```
0xdevalias, line 3: Let’s get rid of these comments. This is a pattern that went against our standards that we want to remove across the codebase, definitely not add more of.

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
```diff
@@ -0,0 +1,12 @@
+import { User } from "types/User";
```
0xdevalias, line 1: This file was removed in a previous PR, now it seems to be back? I expect this may be because of something to do with the wrong branch being merged in/etc, as mentioned in another comment.

0xdevalias (changes requested):
I think this might be messed up, see my comment above (https://github.com/sparkletown/sparkle/pull/1218#issuecomment-790585171) RE: it seeming like you merged the wrong branch into this or similar. Lots of unrelated commits here.

src/components/molecules/UserProfilePicture/UserProfilePicture.types.ts
igachmovmelon: Well I merged with origin/master

0xdevalias: *nods* I figured you may have. Our default branch is staging, so we never merge master into PR’s like this.

0xdevalias:
> I needed to force push this branch to the stage where it wasn't merged with the origin/master branch. I am very sorry for this it won't happen again.

@igachmovmelon That's fine. Force push was the only real way to 'undo' that anyway.

No worries, we all make mistakes at times :)

src/components/templates/Playa/Video/Participant.tsx
```diff
@@ -154,16 +149,11 @@ const Participant: React.FC<React.PropsWithChildren<ParticipantProps>> = ({
     ? " (host)"
     : "";
 
-  const onNameClick = useCallback(() => setSelectedUserProfile(user), [
-    setSelectedUserProfile,
-    user,
-  ]);
-
```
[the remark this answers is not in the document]
0xdevalias: probably name it something better though, such as `selectUser` or `openUserProfile`

src/components/templates/Playa/VideoChatLayer.tsx
```diff
@@ -5,20 +5,13 @@ import { useUser } from "hooks/useUser";
...
-type PropsType = {
-  setSelectedUserProfile: (user: WithId<User>) => void;
-};
-
 const ROOM_PREFIX = PROJECT_ID + "-";
 
-const VideoChatLayer: React.FunctionComponent<PropsType> = ({
-  setSelectedUserProfile,
-}) => {
+const VideoChatLayer: React.FunctionComponent = () => {
```
0xdevalias, line 14:
```suggestion
const VideoChatLayer: React.FC = () => {
```

0xdevalias (changes requested):
Comments/suggestions/changes.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -7,36 +7,26 @@ import {
...
-import { WithId } from "utils/id";
 
 import "./ChatSidebar.scss";
 
 export const ChatSidebar: React.FC = () => {
-  const [selectedProfile, setSelectedProfile] = useState<WithId<User>>();
-  const unselectProfile = useCallback(() => setSelectedProfile(undefined), [
-    setSelectedProfile,
-  ]);
-  const hasSelectedProfile = selectedProfile !== undefined;
```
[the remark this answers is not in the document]
0xdevalias: > we can you the state if there is user or not.

@igachmovmelon What is that meant to say?

---

We tend to build useful APIs for our hooks that encapsulate their 'inner workings', and provide 'useful exports'.

In the 'return object' from the controls hook, add `hasSelectedProfile: selectedProfile !== undefined`, that way it's included in this 'API' and available to use anywhere we need it, without exposing/relying on knowing the 'inner workings' of the hook.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -1,18 +1,21 @@
...
 import { WithId } from "utils/id";
 import { currentVenueSelectorData, venueEventsSelector } from "utils/selectors";
 import { isTruthy } from "utils/types";
+
+import { Room } from "types/rooms";
+import { VenueEvent } from "types/venues";
+import { User } from "types/User";
+
 import "./NavSearchBar.scss";
-import { NavSearchBarInput } from "./NavSearchBarInput";
```
0xdevalias, line 15: [nit] Import sorting order, based on what is here:

- `types/*`
- `utils/*`
- `hooks/*`
- `components/templates/*`
- `components/organisms/*`
- `./*`
- `./*.scss`

src/components/organisms/ChatSidebar/components/VenueChat/VenueChat.tsx
```diff
@@ -21,7 +17,7 @@ export const VenueChat: React.FC<VenueChatProps> = ({ onAvatarClick }) => {
         messages={messagesToDisplay}
         sendMessage={sendMessage}
         deleteMessage={deleteMessage}
-        onAvatarClick={onAvatarClick}
+        onAvatarClick={openUserProfileModal}
```
0xdevalias, line 20: [NB] We could potentially also remove `onAvatarclick` from `ChatBox` and just have it use `useProfileModalControls` directly? Haven't looked where else this component is used and if this prop is used in a different way in those locations.

src/components/templates/Audience/Audience.tsx
```diff
@@ -461,7 +459,7 @@ export const Audience: React.FunctionComponent = () => {
                               seat && seatedPartygoer === null
                                 ? takeSeat(row, column)
                                 : seatedPartygoer !== null
-                                ? setSelectedUserProfile(seatedPartygoer)
+                                ? openUserProfileModal(seatedPartygoer)
                                 : null
                             }
```
0xdevalias, line 464: [NB, potentially OOS] I'm not 100%, but I feel like this is sort of legacy/leftover from ancient code, and may not even be required anymore. Theoretically, if I click on a seat that has a user in it, then the `UserProfilePicture` should be registering the click, and handling opening the modal for me. So I don't think we should also need to have explicit logic here to handle it?

We probably want to make sure we don't try and take the seat if there is already another user there still (which eventually should be something we handle/enforce on the backend probably, rather than just the frontend.

This whole section of the code is actually fairly legacy, as I know I refactored some almost exactly similar code that was used on the `PartyMap` to use `usePartygoersBySeat`/`usePartygoersOverlay` and related to handle this sort of stuff:

- https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/PartyMap/components/Map/Map.tsx
- https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/PartyMap/components/Map/hooks/usePartygoersBySeat.ts
- https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/PartyMap/components/Map/hooks/usePartygoersOverlay.tsx

src/components/templates/Jazzbar/components/JazzBarTableComponent/JazzBarTableComponent.tsx
```diff
@@ -40,7 +41,7 @@ const TableComponent: React.FunctionComponent<TableComponentPropsType> = ({
           usersSeatedAtTable.length >= 0 &&
           usersSeatedAtTable.map((user) => (
```
0xdevalias, line 42: [OOS] This should really be `useMemo`'d

src/hooks/useProfileModalControls.ts
```diff
@@ -0,0 +1,33 @@
...
+export const useProfileModalControls = () => {
+  const dispatch = useDispatch();
+  const selectedUserProfile = useSelector(userProfileSelector);
+
+  const openUserProfileModal = useCallback(
+    (userProfile?: WithId<User>) => {
+      dispatch(setUserProfileData(userProfile));
+    },
+    [dispatch]
+  );
```
0xdevalias, line 22: [NB] If we make this a required prop, what are the 'flow on effects' of that at the 'call sites' where it is used? I expect this is often potentially `undefined` where it is used, which is why we probably implemented it like this here.

I feel like the UX would be strange if for some reason it was undefined, and we clicked on something expecting the profile to open, particularly if the modal was already open, as this would sort of result in it being closed (by setting it to `undefined`). This is likely a rare edgecase though.. so may not really matter in reality.

If there aren't many 'flow on effects' from making this non-optional, then it would be good to change it:

```suggestion
  const openUserProfileModal = useCallback(
    (userProfile: WithId<User>) => {
      dispatch(setUserProfileData(userProfile));
    },
    [dispatch]
  );
```

If there would be a lot of flow on effects, then perhaps we can make the code here more obvious that it's not intended to be able to unset the profile, with something like:

```suggestion
  const openUserProfileModal = useCallback(
    (userProfile?: WithId<User>) => {
      // We can only open the modal when we actually have a userProfile
      if (!userProfile) return
      
      dispatch(setUserProfileData(userProfile));
    },
    [dispatch]
  );
```

src/hooks/useProfileModalControls.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const closeUserProfileModal = useCallback(() => {
+    dispatch(setUserProfileData(undefined));
+  }, [dispatch]);
+
+  return {
+    selectedUserProfile,
+    openUserProfileModal,
+    closeUserProfileModal,
+  };
```
0xdevalias, line 32: You gave https://github.com/sparkletown/sparkle/pull/1218#discussion_r592918721 a 👍🏻 and yet don't seem to have actually implemented it here at all? Usually a 👍🏻 lie that would imply also making the suggested change, and if you're not going to make it, then comments/further discussion as to why not.

```suggestion
  return {
    selectedUserProfile,
    hasSelectedUserProfile = selectedUserProfile !== undefined,
    openUserProfileModal,
    closeUserProfileModal,
  };
```

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -5,6 +5,8 @@ import { AnyVenue, VenueTemplate } from "types/venues";
 
 import { FriendShipPage } from "pages/FriendShipPage";
 
+import { useProfileModalControls } from "hooks/useProfileModalControls";
+
```
0xdevalias, line 9: [nit] Import sort order. `pages/*` should really be the 'top-level' of the 'atomic design methodology' structure (pages, templates, organisms, molecules, atoms)

```suggestion
import { useProfileModalControls } from "hooks/useProfileModalControls";

import { FriendShipPage } from "pages/FriendShipPage";
```


Further reading:

- https://github.com/sparkletown/internal-sparkle-issues/issues/414
  - https://atomicdesign.bradfrost.com/chapter-2/

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -126,6 +134,11 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
       <AnnouncementMessage message={venue?.bannerMessage} />
       {template}
       <ChatSidebar />
+      <UserProfileModal
+        userProfile={selectedUserProfile}
+        show={selectedUserProfile !== undefined}
+        onHide={closeUserProfileModal}
+      />
```
0xdevalias, line 141: Looks like https://github.com/sparkletown/sparkle/pull/1218#discussion_r590989276 is still outstanding? Are we planning on doing anything about it?

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -16,6 +18,7 @@ import { PlayaRouter } from "components/templates/Playa/Router";
 import { ReactionPage } from "components/templates/ReactionPage";
 
 import { ChatSidebar } from "components/organisms/ChatSidebar";
+import UserProfileModal from "components/organisms/UserProfileModal";
```
0xdevalias, line 21: It would be good to use a named export and remove the default export from `UserProfileModal` (default exports are a legacy pattern we're trying to eradicate from the codebase).

Should be a fairly minimal change now that it should only be imported in one or a couple of locations.

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type SetUserProfileDataAction = ReduxAction<
+  UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  { userProfile?: WithId<User> }
+>;
+
+export const setUserProfileData = (userProfile?: WithId<User>) => ({
+  type: UserProfileActionTypes.SET_USER_PROFILE_DATA,
+  payload: { userProfile },
+});
```
0xdevalias, line 18: My previous review comments in https://github.com/sparkletown/sparkle/pull/1218#discussion_r585248664 were marked as resolved, yet seemingly nothing ever actually done about them? Why were they marked as resolved?

0xdevalias (changes requested):
A few outstanding questions/etc, + need to resolve the merge conflicts with `src/components/molecules/NavSearchBar/NavSearchBar.tsx`.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -126,6 +129,11 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
       <AnnouncementMessage message={venue?.bannerMessage} />
       {template}
       <ChatSidebar />
+      <UserProfileModal
+        userProfile={selectedUserProfile}
+        show={selectedUserProfile !== undefined}
+        onHide={unsetUserProfile}
+      />
```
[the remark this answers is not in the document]
mike-lvov: Yep, I'm for having them inside the component.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -53,50 +43,41 @@ export const ChatSidebar: React.FC = () => {
...
-      />
-    </>
+      <div className="chat-sidebar__tab-content">
+        {chatSettings.openedChatType === ChatTypes.VENUE_CHAT && <VenueChat />}
+        {chatSettings.openedChatType === ChatTypes.PRIVATE_CHAT && (
+          <PrivateChats
+            recipientId={chatSettings.recipientId}
+            onAvatarClick={setUserProfile}
+          />
+        )}
```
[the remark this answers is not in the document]
mike-lvov: There should be no `onAvatarClick`. The hook should be used directly in the component

src/components/organisms/ChatSidebar/components/VenueChat/VenueChat.tsx
```diff
@@ -21,7 +17,7 @@ export const VenueChat: React.FC<VenueChatProps> = ({ onAvatarClick }) => {
         messages={messagesToDisplay}
         sendMessage={sendMessage}
         deleteMessage={deleteMessage}
-        onAvatarClick={onAvatarClick}
+        onAvatarClick={setUserProfile}
       />
```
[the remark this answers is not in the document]
mike-lvov: Yes

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -53,50 +43,41 @@ export const ChatSidebar: React.FC = () => {
...
-        show={hasSelectedProfile}
-        onHide={unselectProfile}
-      />
-    </>
+      <div className="chat-sidebar__tab-content">
+        {chatSettings.openedChatType === ChatTypes.VENUE_CHAT && <VenueChat />}
+        {chatSettings.openedChatType === ChatTypes.PRIVATE_CHAT && (
+          <PrivateChats
+            recipientId={chatSettings.recipientId}
+            onAvatarClick={openUserProfileModal}
```
mike-lvov, line 77: I think we still should remove `onAvatarClick` as a prop here and use openUserProfile directly in `PrivateChats` component

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -98,11 +90,7 @@ export const ReactionList: React.FC<ReactionListProps> = ({
         {allReactions}
       </div>
 
-      <UserProfileModal
-        userProfile={selectedUserProfile}
-        show={selectedUserProfile !== undefined}
-        onHide={() => setSelectedUserProfile(undefined)}
-      />
+      <UserProfileModal />
```
mike-lvov, line 93: Do you still need it here?

Isn't it covered by the one declared on the `TemplateWrapper` level?

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,15 @@
+import { User } from "types/User";
+import { WithId } from "utils/id";
+export const UPDATE_USER_PROFILE_DATA: string = "UPDATE_USER_PROFILE_DATA";
+
+interface UpdateProfileDataAction {
+  type: typeof UPDATE_USER_PROFILE_DATA;
+  payload?: WithId<User>;
+}
```
mike-lvov, line 8: could you please take actions/Chat as an example and do it accordingly?

```
export type SetChatsSidebarVisibilityAction = ReduxAction<
  ChatActionTypes.SET_CHAT_SIDEBAR_VISIBILITY,
  { isVisible: boolean }
>;

export type SetVenueChatTabOpenedAction = ReduxAction<
  ChatActionTypes.SET_VENUE_CHAT_TAB_OPENED,
  VenueChatSettings
>;
```

src/store/reducers/UserProfile.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  UserProfileActions,
+  UPDATE_USER_PROFILE_DATA,
+} from "../actions/UserProfile";
```
mike-lvov, line 4:
```suggestion
} from "store/actions/UserProfile";
```

mike-lvov (changes requested):
A few comments/suggestions

mike-lvov (approved):
LGTM

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -33,8 +34,14 @@ export const ChatMessage: React.FC<ChatProps> = ({
...
+        <UserAvatar
+          onClick={() => openUserProfileModal(author)}
+          avatarSrc={author.pictureUrl}
+        />
+        <span
+          onClick={() => openUserProfileModal(author)}
+          className="chat-message__author"
+        >
           {author.partyName}
         </span>
```
0xdevalias, line 46: Instead of having the same `onClick` handler on both `UserAvatar` and the `span`, it might be better to move it up to the parent `div`, and just have it in a single location?

We should also `useCallback` the handler to prevent re-renders from the inline function call changing on every render.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -26,48 +27,54 @@ import Button from "components/atoms/Button";
 import "./UserProfileModal.scss";
 
 type PropTypes = {
```
0xdevalias, line 29: Prefer `interface` for props:

```suggestion
export interface UserProfileModalProps {
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -26,48 +27,54 @@ import Button from "components/atoms/Button";
...
-  show: boolean;
-  onHide: () => void;
   zIndex?: number;
-  userProfile?: WithId<User>;
 };
 
-const UserProfileModal: React.FunctionComponent<PropTypes> = ({
-  show,
-  onHide,
+export const UserProfileModal: React.FunctionComponent<PropTypes> = ({
```
0xdevalias, line 33:
```suggestion
export const UserProfileModal: React.FC<UserProfileModalProps> = ({
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -26,48 +27,54 @@ import Button from "components/atoms/Button";
...
     return <></>;
   }
 
   // REVISIT: remove the hack to cast to any below
   return (
-    <Modal show={show} onHide={onHide} style={zIndex && { zIndex }}>
+    <Modal
+      show={hasSelectedProfile}
+      onHide={closeUserProfileModal}
+      style={zIndex && { zIndex }}
```
0xdevalias, line 67: [NB] I think we don't need the `zIndex &&` check here, as if `zIndex` is `undefined` then it won't 'output' any styles, and if the styles object is empty, it won't output any styles. So I believe this can just be:

```suggestion
      style={{ zIndex }}
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -26,48 +27,54 @@ import Button from "components/atoms/Button";
...
       <Modal.Body>
         <div className="modal-container modal-container_profile">
           <div className="profile-information-container">
             <div className="profile-basics">
               <div className="profile-pic">
                 <img
-                  src={userProfile.pictureUrl || "/default-profile-pic.png"}
+                  src={
+                    selectedUserProfile.pictureUrl || "/default-profile-pic.png"
+                  }
```
0xdevalias, line 77: `"/default-profile-pic.png"` should be moved to a const in `src/settings.ts` (as something like `DEFAULT_PROFILE_PIC`) and then referenced here.

(I feel like we already made this change somewhere.. maybe it's in one of @denisdimitrov's outstanding branches or similar?)

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -83,7 +90,7 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
               </div>
               <div className="profile-text">
                 <h2 className="italic">
-                  {userProfile.partyName || "Captain Party"}
+                  {selectedUserProfile.partyName || "Captain Party"}
                 </h2>
```
0xdevalias, line 94: `"Captain Party"` should be moved to a const in `src/settings.ts` (as something like `DEFAULT_PARTY_NAME`) and then referenced here.

(I feel like we already made this change somewhere.. maybe it's in one of @denisdimitrov's outstanding branches or similar?)

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -95,7 +102,7 @@ const UserProfileModal: React.FunctionComponent<PropTypes> = ({
                     {/*
                     // @debt typing - need to support known User interface with unknown question keys
                     // @ts-ignore */}
-                    {userProfile[question.name] || //@debt typing - look at the changelog, was this a bug?
+                    {selectedUserProfile[question.name] || //@debt typing - look at the changelog, was this a bug?
                       "I haven't edited my profile to tell you yet"}
```
0xdevalias, line 106: This default text should possibly also be moved to a const in `src/settings.ts` and then referenced here.

src/components/templates/Audience/Audience.tsx
```diff
@@ -12,6 +12,12 @@ import { addReaction } from "store/actions/Reactions";
 
 import { makeUpdateUserGridLocation } from "api/profile";
 
+import { useDispatch } from "hooks/useDispatch";
+import { useRecentVenueUsers } from "hooks/users";
+import { useSelector } from "hooks/useSelector";
+import { useUser } from "hooks/useUser";
+import { useVenueId } from "hooks/useVenueId";
+
```
0xdevalias, line 20: Why were these moved from where they were? `hooks/*` live below `utils/*` in the import order groupings.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -253,11 +241,7 @@ export const Map: React.FC<MapProps> = ({
           {roomOverlay}
         </div>
 
-        <UserProfileModal
-          userProfile={selectedUserProfile}
-          show={isUserProfileSelected}
-          onHide={deselectUserProfile}
-        />
+        <UserProfileModal />
```
0xdevalias, line 244: I don't think we need to be rendering `UserProfileModal` in this component anymore do we? It should only be rendered in a single location.

src/hooks/useProfileModalControls.ts
```diff
@@ -0,0 +1,38 @@
...
+  const openUserProfileModal = useCallback(
+    (userProfile?: WithId<User>) => {
+      // We can only open the modal when we actually have a userProfile
+      if (!userProfile) return;
+
+      dispatch(updateUserProfileData(userProfile));
+    },
+    [dispatch]
+  );
+  const closeUserProfileModal = useCallback(() => {
```
0xdevalias, line 28:
```suggestion
  );

  const closeUserProfileModal = useCallback(() => {
```

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -3,7 +3,11 @@ import { Route, Switch, useHistory, useRouteMatch } from "react-router-dom";
 
 import { AnyVenue, VenueTemplate } from "types/venues";
 
-import { FriendShipPage } from "pages/FriendShipPage";
+import { AnnouncementMessage } from "components/molecules/AnnouncementMessage";
+
+import { ChatSidebar } from "components/organisms/ChatSidebar";
+import { UserProfileModal } from "components/organisms/UserProfileModal";
+import { WithNavigationBar } from "components/organisms/WithNavigationBar";
```
0xdevalias, line 10: The way the order of these imports was changed is backwards. From top to bottom it should be:

- `pages/*`
- `components/templates/*`
- `components/organsims/*`
- `components/molecules/*`
- `components/atoms/*`

As was detailed in my original review comment about this in https://github.com/sparkletown/sparkle/pull/1218#discussion_r594002774

src/store/actions/UserProfile.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type UpdateUserProfileDataAction = ReduxAction<
+  UserProfileActionTypes.UPDATE_USER_PROFILE_DATA,
+  { userProfile?: WithId<User> }
+>;
+
+export const updateUserProfileData = (userProfile?: WithId<User>) => ({
+  type: UserProfileActionTypes.UPDATE_USER_PROFILE_DATA,
+  payload: { userProfile },
+});
```
0xdevalias, line 18: I believe this should be something like this?

```suggestion
export const updateUserProfileData: UpdateUserProfileDataAction = (userProfile) => ({
  type: UserProfileActionTypes.UPDATE_USER_PROFILE_DATA,
  payload: { userProfile },
});
```

0xdevalias (changes requested):
Comments/suggestions/changes.

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -1,42 +1,45 @@
...
 
   const timestamp = ts_utc.toMillis();
 
   const containerStyles = classNames("chat-message", {
     "chat-message--me": isMine,
   });
 
+  const onOpenUserProfile = useCallback(() => {
+    openUserProfileModal(author);
+  }, [openUserProfileModal, author]);
```
0xdevalias, line 35:
```suggestion
  const openAuthorProfile = useCallback(() => {
    openUserProfileModal(author);
  }, [openUserProfileModal, author]);
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -25,49 +32,53 @@ import Button from "components/atoms/Button";
 
 import "./UserProfileModal.scss";
 
-type PropTypes = {
-  show: boolean;
-  onHide: () => void;
+export interface UserProfileModalProps {
   zIndex?: number;
```
0xdevalias, line 36: [NB] Do we actually use this prop anywhere anymore?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -25,49 +32,53 @@ import Button from "components/atoms/Button";
...
     // NOTE: Hide the modal, after the chat is opened;
-    onHide();
-  }, [selectRecipientChat, onHide, chosenUserId]);
+    closeUserProfileModal();
+  }, [selectRecipientChat, closeUserProfileModal, chosenUserId]);
 
-  if (!userProfile || !chosenUserId || !user) {
+  if (!selectedUserProfile || !chosenUserId || !user) {
     return <></>;
   }
```
0xdevalias, line 66:
```suggestion
  if (!selectedUserProfile || !chosenUserId || !user) {
    return null;
  }
```

**Further Reading:**

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -25,49 +32,53 @@ import Button from "components/atoms/Button";
...
 
   // REVISIT: remove the hack to cast to any below
```
0xdevalias, line 68: Does this hack still exist? If not, let's remove this comment.

0xdevalias (approved):
A couple of tiny changes/questions that it would be good to fix up before merging this PR.

Then, LGTM!

src/components/organisms/UserProfileModal/UserProfileModal.tsx
igachmovmelon: @0xdevalias I am not sure what is the hack and if it is still there.

0xdevalias: No worries. Thanks for checking in any case :)

PR #1221, opened 2021-03-01
add new firebase environment

0xdevalias:
partially closes https://github.com/sparkletown/internal-sparkle-issues/issues/454


.circleci/config.yml
```diff
@@ -248,13 +251,19 @@ jobs:
...
               TARGET=sparkle-kotr
-              RELEASE_STAGE=kotr
+              RELEASE_STAGE=env/kotr
             fi
             if [ $CIRCLE_BRANCH = "env/memrise" ]; then
               PREFIX=MEMRISE_
               ENV=sparkle-memrise
               TARGET=sparkle-memrise
-              RELEASE_STAGE=memrise
+              RELEASE_STAGE=env/memrise
```
0xdevalias, line 260: [note] This was set incorrectly previously.

.circleci/config.yml
```diff
@@ -248,13 +251,19 @@ jobs:
               PREFIX=KOTR_
               ENV=sparkle-kotr
               TARGET=sparkle-kotr
-              RELEASE_STAGE=kotr
+              RELEASE_STAGE=env/kotr
```
0xdevalias, line 254: [note] This was set incorrectly previously.

0xdevalias (commented):
Notes

src/index.tsx
```diff
@@ -147,6 +147,7 @@ if (BUGSNAG_API_KEY) {
     "deloitte",
```
mike-lvov, line 147: should deloitte be `env/deloitte?`

mike-lvov (approved):
There is some inconsistency with env naming I guess?

src/index.tsx
0xdevalias: The `env/` pattern is something I only implemented in the last round of environments. Eventually it would be good to align them all to following this pattern, but that is OOS of this PR.

There will be some code changes, and possibly some nuance (eg. in how we map it to existing bugsnag data/etc) if/when we decide to do this.

0xdevalias:
@mike-lvov / @sparkletown/engineering For this to be usable, I will need to be able to deploy it through to `master`. Is there anything in `staging` at the moment that we aren't confident about that would be an issue if I deployed it?

PR #1224, opened 2021-03-01
fix: make external banner links open in new tab

malaikaaiyar:
This fixes sparkletown/internal-sparkle-issues#471

Before: link in banner opened in same tab as sparkle venue
Now: link opens in new tab


0xdevalias (changes requested):
The changes made here will make every link open in a new window, which isn't the desired state.

---

See my comments on the original issue for how to approach this:

> From a technical standpoint, this should just be a change to the helper function that the  component uses to make links clickable, and then using the helpers in utils/url to correctly handle the external/not external stuff.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/471#issuecomment-786940486_

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/url.ts
  - There are various helpers here such as `isExternalUrl`, `openUrl`, `externalUrlAdditionalProps`, `getExtraLinkProps`, etc
  - You probably want to use `getExtraLinkProps`, probably in combination with `isExternalUrl`

src/utils/getLinkFromText.tsx
```diff
@@ -1,4 +1,5 @@
 import React from "react";
+import { isExternalUrl, externalUrlAdditionalProps } from "./url";
```
0xdevalias, line 2: [nit]

```suggestion

import { isExternalUrl, externalUrlAdditionalProps } from "./url";
```

src/utils/getLinkFromText.tsx
```diff
@@ -15,6 +16,10 @@ export const getLinkFromText = (text: string) => {
               : `https://${word}`
           }
           key={index}
+          target={
+            isExternalUrl(word) ? externalUrlAdditionalProps.target : undefined
+          }
+          rel={isExternalUrl(word) ? externalUrlAdditionalProps.rel : undefined}
```
0xdevalias, line 22: You can do this on a single line between ln5-6, outside of the JSX. Just assign it to a `const`. Then you can use the react 'spread props' syntax here.

```suggestion
          {...fooConstNameYouPicked}
```

Further Reading:

- https://reactjs.org/docs/jsx-in-depth.html#spread-attributes

0xdevalias (changes requested):
Suggestions/comments

0xdevalias (approved):
[CRO] Changes look good here. If you’ve clicked through and tested it works, LGTM!

PR #1225, opened 2021-03-01
add unclickable room type

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/387

By leveraging the existing 'room placement' functionality, we can create an 'unclickable room' that will allow us to position arbitrary images on the PartyMap (including gifs), for adding some extra 'flavour' to a map.


src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -13,21 +13,25 @@ $label-visible: flex;
 
   transition: transform 600ms cubic-bezier(0.23, 1, 0.32, 1);
 
-  &:hover {
-    z-index: z(map-room-hovered);
-    transform: scale(1.1);
+  &--active {
```
0xdevalias, line 16: Perhaps could call this portal or similar? Or maybe leave it not as a variant since it'll likely be the main/default type?

The specific reason I wanted to comment here though was that the styles for the other more specific variants such as `--always-show-label` will likely only make sense on this 'default type' as well, as the 'unclickable' type definitely won't need them, etc.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -3,7 +3,7 @@ import classNames from "classnames";
```
0xdevalias, line 3: I would potentially do more in the actual logic of this file when the type is unclickable, to prevent any unneeded things running/etc when that type is set.

0xdevalias: Probably would prevent `handleRoomHovered`/`handleRoomUnhovered` from doing anything. Or just outright don't set `onClick`/`onMouseEnter`/`onMouseLeave` (can set them to undefined or something? Maybe a ternary. Maybe in and object and just spread them into the props when not an unclickable type?)

0xdevalias: Should make the `selectRoom` prop optional as well.

0xdevalias: I'd possibly even just outright conditionally render the whole `maproom__label` block rather than hiding it with CSS/etc.

src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -37,6 +39,22 @@ $label-visible: flex;
...
 
+  &--unclickable {
+    cursor: default;
+
+    &:hover {
+      transform: none;
+
+      .maproom__label {
+        display: $label-hidden;
+      }
```
0xdevalias, line 50: This won't be needed if you conditionally render that section, as mentioned in another comment in this review.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -31,6 +31,8 @@ export const PartyMap: React.FC = () => {
   const hasSelectedRoom = !!selectedRoom;
 
   const selectRoom = useCallback((room: Room) => {
+    if (room.type === RoomTypes.unclickable) return;
+
```
0xdevalias, line 35: This logic shouldn't be handled here, only the room component should care about it IMO. See my other comment in this PR about not using this + the other callbacks when room is unclickable.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -38,6 +38,7 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   }, [dispatch]);
 
   const containerClasses = classNames("maproom", {
+    "maproom--unclickable": room.type === RoomTypes.unclickable,
     "maproom--always-show-label":
       venue.roomVisibility === RoomVisibility.nameCount ||
       (venue.roomVisibility === RoomVisibility.count && hasRecentRoomUsers),
```
0xdevalias, line 44: Maybe only set this when room.type != unclickable

0xdevalias: Same with `titleClasses`?

src/types/rooms.ts
```diff
@@ -1,20 +1,24 @@
...
-  attendanceBoost?: number;
   height_percent: number;
   image_url: string;
   isEnabled: boolean;
   width_percent: number;
   x_percent: number;
   y_percent: number;
+  attendanceBoost?: number;
+  type?: RoomTypes;
 }
```
0xdevalias, line 19: Copy/paste this block over the diff here, sorts them better to match the v2 styles/etc.

```typescript
export interface Room {
  type?: RoomTypes;
  title: string;
  subtitle: string;
  url: string;
  about: string;
  x_percent: number;
  y_percent: number;
  width_percent: number;
  height_percent: number;
  isEnabled: boolean;
  image_url: string;
  // Legacy?
  attendanceBoost?: number;
}
```

src/types/rooms.ts
```diff
@@ -28,4 +32,5 @@ export interface RoomData_v2 {
   template?: string;
   image_url?: string;
   roomIndex?: number;
+  type?: RoomTypes;
```
0xdevalias, line 35: Maybe sort this at the top like my other suggestion

src/types/rooms.ts
```diff
@@ -1,20 +1,24 @@
-export interface BaseRoom {
+export enum RoomTypes {
+  unclickable = "UNCLICKABLE",
+}
```
0xdevalias, line 3: This is what I was wanting 👍🏻 :)

May consider it OOS, but https://github.com/sparkletown/internal-sparkle-issues/issues/505 will also basically want to add a type here too (required for the Monday event)

0xdevalias (changes requested):
So far looking good, matches the approach I was wanting with this. A few comments/tweaks/changes.

src/components/templates/PartyMap/PartyMap.tsx
mike-lvov: @0xdevalias we have keyboard controls, which are out of the Room component.

0xdevalias: That's fair, can keep this.

But the comments I made about the logic changes within the Room component should still be done as well.

src/types/rooms.ts
0xdevalias: > May consider it OOS, but sparkletown/internal-sparkle-issues#505 will also basically want to add a type here too (required for the Monday event)

That comment was wrong, as it was a new venue template we needed there, not a new room type.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -59,9 +63,12 @@ const NavSearchBar = () => {
...
 
-    const roomsResults: Room[] = (venue?.rooms?.filter((room) =>
-      room.title.toLowerCase().includes(normalizedSearchQuery)
-    ) ?? []) as Room[]; // @debt Clean up this hack properly once old templates are deleted
+    const roomsResults: Room[] =
+      venue?.rooms
+        ?.filter((room) =>
+          room.title.toLowerCase().includes(normalizedSearchQuery)
+        )
+        .filter((room) => room.type !== RoomTypes.unclickable) ?? [];
```
0xdevalias, line 71: We can handle both of these `.filter` cases in a single pass, rather than iterating through the collection twice.

We could also likely remove this at the query level, rather than fetching them all then filtering on the frontend.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -1,9 +1,10 @@
 import React, { useCallback, useMemo } from "react";
+import noop from "lodash/noop";
```
0xdevalias, line 2: We don't need to import from a library for this.. the whole implementation is `() => undefined`

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -61,16 +67,18 @@ export const MapRoom: React.FC<MapRoomProps> = ({
     <div
       className={containerClasses}
       style={roomPositionStyles}
-      onClick={selectRoom}
-      onMouseEnter={handleRoomHovered}
-      onMouseLeave={handleRoomUnhovered}
+      onClick={isUnclickable ? noop : selectRoom}
+      onMouseEnter={isUnclickable ? noop : handleRoomHovered}
+      onMouseLeave={isUnclickable ? noop : handleRoomUnhovered}
```
0xdevalias, line 72: We could handle this in 'one hit' by spreading these props in only when relevant.

0xdevalias (changes requested):
A few small tweaks/comments

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1225/commits/03ee9c9b63c7abd63465d57951e08c49b50653d9

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1225/commits/6a8bb74a884fc0df9e55f819a976c8214b771741

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: This works though.. going to call this wontfix.

0xdevalias:
We may want/need to implement some z-index stuff for this room type. Currently the user's avatar shows above it, but the grid shows below it. Given it's intended use case, I expect we want to show the grid over the top of unclickable rooms. They should be basically 1 step above the map background.

image

---

Semi-related, https://github.com/sparkletown/internal-sparkle-issues/issues/387#issuecomment-791587960 suggests that we may want to allow the `z-index` to be configurable from firestore, at least until we have a 'proper' 'room sorting'/layering method implemented.

0xdevalias:
@mike-lvov Did you see this comment on the issue?

> We can sort the gifs if the unclickable room function is working, but we will need a reordering function (@Sparkle-T  says- z index) to make it not a time sucker.
> 
> _Originally posted by @sofisparkle in https://github.com/sparkletown/internal-sparkle-issues/issues/387#issuecomment-791587960_

I've asked for further clarity in https://github.com/sparkletown/internal-sparkle-issues/issues/387#issuecomment-796404112

src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -41,6 +41,7 @@ $label-visible: flex;
 
   &--unclickable {
     cursor: default;
+    z-index: revert;
```
0xdevalias, line 44: Is `revert` even a valid value here? I can't see anything about it when googling?

- https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

src/scss/constants.scss
```diff
@@ -90,6 +90,7 @@ $z-layers: (
   // Map
   map-room-hovered: 3,
   map-room: 2,
+  map-seat: 1,
```
0xdevalias, line 93: The way I envisioned this was that we would have the `map-background` at the lowest level (which may not need an explicit `z-index`, not sure), then the `map-room-unclickable`, then `map-grid`, then `map-room` and finally `map-hovered`. Or something similar to that.

As currently defined here, it seems like we're explicitly moving the grid below the rooms. But unsure how that will impact the rest of the 'layering' as set out in my 'bug report' in https://github.com/sparkletown/internal-sparkle-issues/issues/387#issuecomment-792289425

0xdevalias (changes requested):
Comments/questions.

See also, required features described in https://github.com/sparkletown/sparkle/pull/1225#issuecomment-796404316

src/scss/constants.scss
0xdevalias: Further to my questions in https://github.com/sparkletown/sparkle/pull/1225#issuecomment-796404316 I have just looked at the prod environment/venue for this event, and it seems that there is a need for the unclickable room gifs to be able to be 'layered' on top of existing rooms as well.

So the changes here probably aren't going to be desirable.

0xdevalias: See screenshot/link in https://github.com/sparkletown/internal-sparkle-issues/issues/387#issuecomment-796474043

src/components/templates/PartyMap/components/Map/MapRoom.scss
mike-lvov: Not sure. It was suggested by the IDE. The intended outcome is to turn the z-index to a default

src/scss/constants.scss
mike-lvov: So, it should be over the grid? I don't see other way to have it over the other rooms. So basically, we should revert the latest changes. Is that what you mean?

0xdevalias: I replied on slack before seeing this, copying here for reference:

> @0xdevalias:
>
> The request is the ability to arbitrarily assign z-indexes to each unclickable room individually, so that their layering can be precisely controlled individually; not as an entire ‘block of things’ on a single layer.
> 
> Some unclickable rooms want to be at the ‘map background image’ layer (which I feel should probably be our default), some want to be sitting ‘above’ every other layer (eg. the birds flying around the top of one of the island room images)
> 
> Eventually we would want to expose this ability to ‘reorder room layers’ in the admin UI tooling directly, but until then, it’s probably sufficient to follow a similar pattern as we did for the ‘iframe venue’, and allow the z-index to be set within the database on the room objects, then used as inline styles.
> <sub>[View in Slack](https://doesitsparkle.slack.com/archives/C01JHLMSFND/p1615499612333000)</sub>

src/components/templates/PartyMap/components/Map/MapRoom.scss
0xdevalias: By my read of the above link, setting it to `auto` should probably do that.

---

Though searching for `revert` on it's own lead me to this:

- https://developer.mozilla.org/en-US/docs/Web/CSS/revert
  - > The revert CSS keyword reverts the cascaded value of the property from its current value to the value the property would have had if no changes had been made by the current style origin to the current element. Thus, it resets the property to its inherited value if it inherits from its parent or to the default value established by the user agent's stylesheet (or by user styles, if any exist). It can be applied to any CSS property, including the CSS shorthand all.
  - > The `revert` keyword works exactly the same as `unset` in many cases. The only difference is for properties that have values set by the browser or by custom stylesheets created by users (set on the browser side).
  - > The `revert` keyword is different from and should not be confused with `initial`, which uses the initial value defined on a per-property basis by the CSS specifications. In contrast, user-agent stylesheets set default values on the basis of CSS selectors.
  
If we weren't using BEM (as I don't technically think we have a parent in this selector path, as `&--unclickable` will just turn into `.maproom--unclickable` rather than `.maproom.maproom--unclickable`), then this would probably apply:

- https://developer.mozilla.org/en-US/docs/Web/CSS/revert#revert_on_a_parent
  - > Reverting effectively removes the value for the element you select with some rule and only for that element.

I think if we moved the z-index `revert` part to a new block that looked something like this, then it would probably work in this 'revert on a parent' way:

```scss
&&--unclickable {
  z-index: revert;
}
```

Which should then output the `.maproom.maproom--unclickable` selecter I believe.

---
  
(sidenote: also opened https://github.com/mdn/content/issues/3047 to fix the `z-index` docs on this)

mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1225/commits/7db0a2f720d51d4000e5f94ebcb251924cec7345

0xdevalias: Didn't want to follow the final part of my suggestion there then?

> I think if we moved the z-index revert part to a new block that looked something like this, then it would probably work in this 'revert on a parent' way:
> 
> ```scss
> &&--unclickable {
>   z-index: revert;
> }
> ```

> Which should then output the .maproom.maproom--unclickable selector with this revert I believe.

mike-lvov: The implementation with the `auto` seems more easy-to-understand and obvious IMO

0xdevalias: Mm, fair enough. I don't believe it will always strictly be accurate. But it's probably 'good enough'.

0xdevalias: I believe the edge case would be, as an example, would be if a parent was set to say `z-index: 100`, the room was set to `z-index: 150`, and the unclickable room is set to `z-index: auto`.

The expectation would be that it 'falls back to `100`, but `auto` will set it to (essentially) 'nothing'.

I think.

src/scss/constants.scss
0xdevalias: Can you rename this `map-grid` or something similar please.

src/components/templates/PartyMap/components/Map/MapRoom.scss
mike-lvov: I don't think I would expect it to be 100. When I used it in my case, my intention was to disable `z-index`, not to lower the value/set the parent's value.

src/scss/constants.scss
0xdevalias: Can you rename this `map-grid` or something similar please.

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -58,6 +58,7 @@ $border-radius: 28px;
   .seat-column {
     width: 100%;
     height: 100%;
+    z-index: z(map-seat);
```
0xdevalias, line 61: Is there any reason this needs to be applied to the seat? I would have expected to apply it to the grid itself?

mike-lvov: There is no grid wrapper

image

src/scss/constants.scss
mike-lvov: did in https://github.com/sparkletown/sparkle/pull/1225/commits/c780464ddb4431dccdb1bef05448301e56ee1c9a

src/components/templates/PartyMap/components/Map/MapRoom.scss
0xdevalias: Fair. Either way, good enough for this PR.

src/components/templates/PartyMap/components/Map/Map.scss
0xdevalias: Fair, that makes sense. No changes required here then.

0xdevalias (approved):
[CRO] Changes look good! If the clickthrough works well, then LGTM! :shipit:

PR #1228, opened 2021-03-02
fix venue edit by venue id when name is changed

denisdimitrov:
Passes the venueId from the params and uses it for venue update in admin v1. This is not a bug per-say but it was an edge case introduced by directly manipulating data in the database. The venue name was changed by someone and it was different than the venue id which was returning an error when trying to find the venue by it's name instead of ID. Which is a big no no and should never be done.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/495

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -410,7 +410,7 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
 });
 
 exports.updateVenue = functions.https.onCall(async (data, context) => {
-  const venueId = getVenueId(data.name);
+  const venueId = data.id || getVenueId(data.name);
   checkAuth(context);
 
   await checkUserIsOwner(venueId, context.auth.token.user_id);
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -10,6 +10,7 @@ import {
 } from "types/venues";
 import { RoomData_v2 } from "types/rooms";
 import { venueInsideUrl } from "utils/url";
+import { WithId } from "utils/id";
 
 export interface EventInput {
   name: string;
@@ -312,7 +313,10 @@ export const createVenue_v2 = async (input: VenueInput_v2, user: UserInfo) => {
   );
 };
 
-export const updateVenue = async (input: VenueInput, user: UserInfo) => {
+export const updateVenue = async (
+  input: WithId<VenueInput>,
+  user: UserInfo
+) => {
   const firestoreVenueInput = await createFirestoreVenueInput(input, user);
 
   return await firebase.functions().httpsCallable("venue-updateVenue")(
--- a/src/pages/Admin/Venue/DetailsForm.tsx
+++ b/src/pages/Admin/Venue/DetailsForm.tsx
@@ -131,11 +131,12 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
       if (!user) return;
       try {
         // unfortunately the typing is off for react-hook-forms.
-        if (!!venueId) await updateVenue(vals as VenueInput, user);
+        if (!!venueId)
+          await updateVenue({ ...(vals as VenueInput), id: venueId }, user);
         else await createVenue(vals as VenueInput, user);
 
         vals.name
-          ? history.push(`/admin/${createUrlSafeName(vals.name)}`)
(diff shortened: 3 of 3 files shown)


mike-lvov (approved):

0xdevalias:
@denisdimitrov the venueId was changed when the venue was cloned across environments, but presumably I didn’t also change the name.

The fact that the name (which IMO should only be a human facing label) is used to derive the id forever seems super broken to me. It should be used to optionally derive the id once in the beginning, but then we should just be using the id directly from that point onwards.

(Obviously this is OOS for this PR, but just capturing my thoughts here as they arose)

functions/venue.js
```diff
@@ -410,7 +410,7 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
 });
 
 exports.updateVenue = functions.https.onCall(async (data, context) => {
-  const venueId = getVenueId(data.name);
+  const venueId = data.id || getVenueId(data.name);
```
0xdevalias, line 413: [Future Cleanup] We should remove `getVenueId(data.name)` from this entirely (and any other backend calls like this).

As per my comments in https://github.com/sparkletown/sparkle/pull/1228#issuecomment-789199882, we shouldn't use a 'name' as an 'id' (even in this case where it's passed through a function to derive it)

> The fact that the name (which IMO should only be a human facing label) is used to derive the id forever seems super broken to me. It should be used to optionally derive the id once in the beginning, but then we should just be using the id directly from that point onwards.

src/api/admin.ts
```diff
@@ -312,7 +313,10 @@ export const createVenue_v2 = async (input: VenueInput_v2, user: UserInfo) => {
   );
 };
 
-export const updateVenue = async (input: VenueInput, user: UserInfo) => {
+export const updateVenue = async (
+  input: WithId<VenueInput>,
+  user: UserInfo
+) => {
```
0xdevalias, line 319: This fixes it for admin v1, but will the same issue not also exist in admin v2 with the `updateVenue_v2` function below? (and possibly any other functions that rely on this same 'derive id from name' behaviour)

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -131,11 +131,12 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
       if (!user) return;
       try {
         // unfortunately the typing is off for react-hook-forms.
-        if (!!venueId) await updateVenue(vals as VenueInput, user);
+        if (!!venueId)
+          await updateVenue({ ...(vals as VenueInput), id: venueId }, user);
         else await createVenue(vals as VenueInput, user);
```
0xdevalias, line 136: Probably would have been better to use `{` + `}` on the if/else blocks since it flows over more than 1 line. We can use the `withId` helper to simplify this line.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -131,11 +131,12 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
       if (!user) return;
       try {
         // unfortunately the typing is off for react-hook-forms.
```
0xdevalias, line 133: [OOS] We can probably resolve that.. which would be better than explicitly `as` casting below, which is going to hide any potential typing issues we would otherwise run into/be protected from.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -131,11 +131,12 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
...
-        if (!!venueId) await updateVenue(vals as VenueInput, user);
+        if (!!venueId)
+          await updateVenue({ ...(vals as VenueInput), id: venueId }, user);
         else await createVenue(vals as VenueInput, user);
 
         vals.name
-          ? history.push(`/admin/${createUrlSafeName(vals.name)}`)
+          ? history.push(`/admin/${createUrlSafeName(venueId ?? vals.name)}`)
           : history.push(`/admin`);
```
0xdevalias, line 140: Should this be `venueId || vals.name ?` given the change made here?

0xdevalias (commented):
@denisdimitrov Retroactive review + cleanup comments

PR #1229, opened 2021-03-02
fix edit background in admin v2

denisdimitrov:
The edit background button sets the map background to an empty string `""` but the check wasn't passing because the statement is false.

The default background is also removed because if the map background url is missing and it defaults to the image then the select options will never render.


0xdevalias:
@denisdimitrov Please remember to assign the appropriate labels to your PRs.

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -29,6 +29,7 @@ const MapPreview: React.FC<MapPreviewProps> = ({
   const [mapRooms, setMapRooms] = useState<RoomData_v2[]>([]);
   const [isSaving, setSaving] = useState<boolean>(false);
 
+  console.log(mapBackground);
```
0xdevalias, line 32: Needs to be removed

src/pages/Admin/Admin_v2.tsx
```diff
@@ -99,12 +99,7 @@ const Admin_v2: React.FC = () => {
   const renderVenueView = () => {
     switch (selectedOption) {
       case SidebarOptions.dashboard:
-        return (
-          <VenueDetails
-            venue={selectedVenue as Venue_v2}
-            onSave={() => setSelectedOption(sidebarOptions[0].id)}
-          />
```
0xdevalias, line 106: Why was the `onSave` removed here? Should other aspects related to `setSelectedOption` be removed as well? This feels like it might not be related to the stated purpose of this PR?

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -92,10 +93,10 @@ const MapPreview: React.FC<MapPreviewProps> = ({
     setIsEditing(false);
   };
 
-  const handleBackgroundRemove = () => {
+  const handleBackgroundRemove = async () => {
     if (!user) return;
 
-    return updateVenue_v2(
+    await updateVenue_v2(
```
0xdevalias, line 99: Should the `Promise` still be getting returned here? That's generally how these should work, even if the end result is `Promise<void>`

src/pages/Admin/MapPreview/MapPreview.tsx
```diff
@@ -123,7 +124,7 @@ const MapPreview: React.FC<MapPreviewProps> = ({
           <Legend
             text="Edit background"
             position="right"
-            onClick={() => handleBackgroundRemove()}
+            onClick={handleBackgroundRemove}
```
0xdevalias, line 127: This should probably be done in an async thunk or updating some local state on error or similar; rather than just calling it and 'forgetting' about it. As is, we have no way of knowing if it worked/failed/etc, and reacting appropriately.

src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -44,13 +40,7 @@ type EditRoomType = RoomData_v2 & {
...
 
 const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
-  const {
-    name,
-    owners,
-    id: venueId,
-    rooms,
-    mapBackgroundImageUrl = DEFAULT_MAP_BACKGROUND,
-  } = venue;
+  const { name, owners, id: venueId, rooms, mapBackgroundImageUrl } = venue;
```
0xdevalias, line 43: Since you removed the default on `mapBackgroundImageUrl`, does anything else here need to change now since it may no longer be defined/provided?

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -243,7 +243,7 @@ export const Map: React.FC<MapProps> = ({
         <img
           width="100%"
           className="party-map-background"
-          src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
+          src={venue.mapBackgroundImageUrl}
```
0xdevalias, line 246: Since DEFAULT_MAP_BACKGROUND was removed here, I would have expected there to also be no need to import it anymore, but it doesn't appear that is the case? Where else is it used in this file, and for what purpose? It seems like it should probably be removed from everywhere..?

functions/venue.js
```diff
@@ -650,7 +650,7 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
     updated.radioStations = [data.radioStations];
   }
 
-  if (data.mapBackgroundImageUrl) {
+  if (typeof data.mapBackgroundImageUrl === "string") {
```
0xdevalias, line 653: What's the intent of this change here? I'm guessing because `""` is falsy, and that's what you said the 'default' is set to currently, so this wouldn't change it previously?

0xdevalias (changes requested):
Questions/comments/etc + aspects detailed in https://github.com/sparkletown/sparkle/pull/1229#issuecomment-789474804

functions/venue.js
denisdimitrov: Yes, that's correct.

src/pages/Admin/Admin_v2.tsx
denisdimitrov: VenueDetails expects only one prop `venue`, that's why I'm removing the `onSave` it's not being used.

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: This function is being used by a `div` with `onClick`. Having `return` here miss-leads (at least me) that the result is going to be used but it isn't

src/pages/Admin/Venue/Details/VenueDetails.tsx
denisdimitrov: Nope, if it's undefined or not selected the choose background options will appear and let you choose one.

src/components/templates/PartyMap/components/Map/Map.tsx
denisdimitrov: The default map dimensions are being used to calculate the grid when the map background is missing. I'm not sure if this is the accurate flow, but that's why this is still being used.

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: I can just wrap the promise inside with try/catch and use some state. How does that sound?

0xdevalias: Yeah, that would be fine for now, that's a similar pattern to what we have elsewhere.

src/pages/Admin/Admin_v2.tsx
0xdevalias: Makes sense

src/pages/Admin/MapPreview/MapPreview.tsx
0xdevalias: Fair enough. What is the type on this function at the moment? If it's implicitly `Promise<void>`, then that's fine/nothing required here.

src/pages/Admin/Venue/Details/VenueDetails.tsx
0xdevalias: Sounds good.

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: Without looking too deeply into it, I'd say that's probably not a desirable flow.

Looking at the `useEffect` code where it's used, it seems to only be used for calculating the aspect ratio of the map image (which is then used to calculate the number of rows to use). I feel like we can just as easily set that to a default (I think we even have default rows already maybe?) I expect it will be wrong and need to re-run currently anyway, so no need to do keep it around for that reason.

https://github.com/sparkletown/sparkle/blob/7d5773daa6eadd6a1faf46bafb9b521365d36a84/src/components/templates/PartyMap/components/Map/Map.tsx#L73-L87

Since we load `mapBackgroundImageUrl` into an `` tag later as well (from an entirely different piece of code), it's possible that this will actually result in us attempting to fetch the map background image twice (assuming caching/etc doesn't intervene and 'save us')

I haven't tested this, but I believe we can probably handle this same 'requirement' using a ref (or a callback ref perhaps?) on the `` tag. I would be looking to follow the same sort of pattern as I used in the `Audience` component to call my 'focus on first load' code.

https://github.com/sparkletown/sparkle/blob/707ab63e1bbdc2f592213be7b71a62c378a66541/src/components/templates/Audience/Audience.tsx#L160-L174

https://github.com/sparkletown/sparkle/blob/707ab63e1bbdc2f592213be7b71a62c378a66541/src/components/templates/Audience/Audience.tsx#L414-L418

0xdevalias: Actually.. even simpler than that, since we are using an `` we can just call it from `onLoad`:

- https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onload
  - > The `onload` property of the `GlobalEventHandlers` mixin is an `EventHandler` that processes load events on a `Window`, `XMLHttpRequest`, **`` element**, etc.
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img
- https://www.javascriptstuff.com/detect-image-load/

functions/venue.js
0xdevalias: Makes sense.

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: Yes, just checked this. It's a void function, no need for the return but the error needs to be handled as you already commented here: https://github.com/sparkletown/sparkle/pull/1229#discussion_r586153454

On it atm.

src/pages/Admin/MapPreview/MapPreview.tsx
denisdimitrov: I'm adding this as we agreed on it, but it feels weird and not correct. I think we should have a more global and generic way of handling these. maybe notifications or some kind of pop ups.

src/components/templates/PartyMap/components/Map/Map.tsx
denisdimitrov: The only reason this logic is in `useEffect` instead of `onLoad` is because it needs to be retriggered if someone changes the number of grid rows in the admin. The `useEffect` will trigger again and recalculate the columns because of the dep array, but the `onLoad` runs only once, when rendered. I searched for a way to 'retrigger' `onLoad` but I couldn't find anything.

denisdimitrov: That being said, the constant `DEFAULT_MAP_BACKGROUND` can be removed because it's not needed for the calculation.

denisdimitrov: In my future answers I will try to be more descriptive on how and why things are written in the way they are. It will probably save you some time writing these suggestions. I understand that sometimes it's hard to understand the whole thing just from 10 lines of code.

src/pages/Admin/MapPreview/MapPreview.tsx
0xdevalias: https://github.com/sparkletown/internal-sparkle-issues/issues/630

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: Makes sense.

---

> In my future answers I will try to be more descriptive on how and why things are written in the way they are. It will probably save you some time writing these suggestions. I understand that sometimes it's hard to understand the whole thing just from 10 lines of code.

Please do! This would be immensely helpful.

I've been asking people to add proper descriptions to their PRs for a while (not just linking it to a backlog issue, but actually explaining the what/why/rationale of the changes in the PR where it makes sense to), as it helps give the reviewer the required context to assess it without needing to 'brute force' or 'figure it all out from scratch'

denisdimitrov:
This is outdated and the functionality has been already refactored and changed as per the new design in https://github.com/sparkletown/sparkle/pull/1583

PR #1231, opened 2021-03-03
fix Auditorium video container to be properly centred + some related cleanup

0xdevalias:
The 'center' video in the Auditorium would 'break out' of its 'container' and overlap seats. It seems its position was set based on the viewport of the browser, rather than it's containing parent.

It also seemed that when there were more seats than would fit on screen at one time, we weren't able to scroll around them.

And to top it off, the 'carved out section' in the middle always felt way too big/space wasting for the size of the video that was normally embedded within it.

I fixed all of the above + cleaned up a few things along the way. Since it's now possible for the 'seats' to extend outside of the window (requiring scrolling), I also auto-focus the central video container when the component is first rendered, so that the user starts off seeing that, rather than the top-left corner of the auditorium seats.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/326

## Notes

- I purposefully didn't clean up the majority of this component/styles, including BEM naming patterns, to minimise the changes is this PR. The focus was on fixing the bug in the most minimal way.
- We can theoretically tweak the overall 'size feeling' of the auditorium seats by changing the `SEAT_SIZE` variable in `Audience.tsx` and `$seat-size` in `Audience.scss` (they must be set to the same value); though I haven't tested this and it will almost certainly require tweaks to other styles to do so.

## Before

image

image

## After

image

image

### DIFF
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -46,6 +46,9 @@ interface ChatOutDataType {
   text: string;
 }
 
+// If you change this, make sure to also change it in Audience.scss's $seat-size
+const SEAT_SIZE = "4vh";
+
 // The seat grid is designed so we can dynamically add rows and columns around the outside when occupancy gets too high.
 // That way we never run out of digital seats.
 // How the seat grid works:
@@ -153,6 +156,22 @@ export const Audience: React.FunctionComponent = () => {
 
   const [iframeUrl, setIframeUrl] = useState<string>("");
 
+  const [hasAlreadyFocussed, setAlreadyFocussed] = useState(false);
+  const focusElementOnLoad = useCallback(
+    (ref: HTMLDivElement | null) => {
+      if (ref && !hasAlreadyFocussed) {
+        ref.scrollIntoView({
+          behavior: "auto",
+          block: "center",
+          inline: "center",
+        });
+
+        setAlreadyFocussed(true);
+      }
+    },
+    [hasAlreadyFocussed]
+  );
+
   useEffect(() => {
     firebase
       .firestore()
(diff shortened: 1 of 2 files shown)


src/components/templates/Audience/Audience.tsx
```diff
@@ -153,6 +152,22 @@ export const Audience: React.FunctionComponent = () => {
 
   const [iframeUrl, setIframeUrl] = useState<string>("");
 
+  const [hasAlreadyFocussed, setAlreadyFocussed] = useState(false);
```
mike-lvov, line 155: Do we need Already here? Maybe we should shorten this to hasFocus/setFocus?

mike-lvov:
Good job on this. Looks like it fixes this one 
https://github.com/sparkletown/internal-sparkle-issues/issues/326

src/components/templates/Audience/Audience.tsx
0xdevalias: The usecase of this state is to be a flag that is set once, and then never again. Ideally I would have just used a `useEffect` that ran once on page load, but refs aren't guaranteed to be defined at that point, and `useEffects` don't memo over refs, so this was my closest workaround.

You could make it `hasFocussed` or similar if you needed to shorten it, but most of the time I feel that shorter variables for the sake of shortness, when it loses explicitness/context/etc isn't really worthwhile. Particularly when they're going to be transpiled into a couple of random characters in the production assets anyway, so size/etc isn't an issue.

src/components/templates/Audience/Audience.tsx
```diff
@@ -235,22 +250,51 @@ export const Audience: React.FunctionComponent = () => {
...
+  const carvedOutHeightInSeats = Math.ceil(carvedOutWidthInSeats * (9 / 16));
+
+  // Calculate the position/size for the central video container
+  const videoContainerWidthInSeats = carvedOutWidthInSeats * 2 + 1;
+  const videoContainerHeightInSeats = carvedOutHeightInSeats * 2 + 1;
+  const videoContainerTopOffsetInSeats = Math.floor(
+    (rowsForSizedAuditorium - carvedOutHeightInSeats * 2) / 2
+  );
+  const videoContainerLeftOffsetInSeats = Math.floor(
+    (columnsForSizedAuditorium - carvedOutWidthInSeats * 2) / 2
```
mike-lvov, line 266: [OOS] Not sure how I feel about this(magical numbers). I do appreciate the comments about these numbers, but they still appear to me somewhat magical. Maybe we should move them into consts?

mike-lvov (approved):
Let's merge it in after @malaikaaiyar approves it

mike-lvov:
@malaikaaiyar said it's good to go. Merging it in

src/components/templates/Audience/Audience.tsx
0xdevalias: I'm not necessarily against consts for them, though figured it would make it overly verbose, particularly in the case where it is just doubling/halving a thing.

I see this is a temporary stop-gap, I feel like the refactor should be able to come up with a much cleaner implementation than this, as it feels overly convoluted/hacky to me still (even though it seemed to work)

PR #1233, opened 2021-03-03
Changes Jazzbar table styles

mike-lvov:
Changes Jazzbar table styles


0xdevalias (approved):
[CRO] Only looked at the diff, didn't do a clickthrough. But if this is tested and works as expected, looks fine to merge.

---

Also, please remember to either link PRs to an existing issue with closing keywords, or directly add it to the current sprint board, so that we can track visibility.

PR #1234, opened 2021-03-04
Revert "fix Auditorium video container to be properly centred + some related cleanup"

mike-lvov:
Reverts sparkletown/sparkle#1231

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -2,42 +2,39 @@
 
 $light: #ffffff;
 
-$spacing: 30px;
-
-// If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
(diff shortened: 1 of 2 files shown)


denisdimitrov (approved):
LGTM!

mike-lvov:
Will get back to this, when we fix all the issues. I need to unblock the staging

0xdevalias:
@mike-lvov I wonder if this could have just been deployed as well? I’m pretty sure that it’s no more broken than the existing Auditorium, even though it does still have a slight bug. In any case, need to make sure this gets in-reverted and not lost.

mike-lvov:
@0xdevalias inreverted here https://github.com/sparkletown/sparkle/pull/1237

0xdevalias:
@mike-lvov Yeah, that's fair enough.

PR #1237, opened 2021-03-04
Fix Auditorium video container to be properly centred + some related cleanup

mike-lvov:
_(Originally created in https://github.com/sparkletown/sparkle/pull/1231, reverted in https://github.com/sparkletown/sparkle/pull/1234, and restored in this PR)_

Fix Auditorium video container to be properly centered + some related cleanup

Fixed related big screen issues and Emoji container styles

closes https://github.com/sparkletown/internal-sparkle-issues/issues/326

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -64,7 +64,7 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
         text-align: center;
 
         width: 90%;
-        min-height: 90px;
+        min-height: 105px;
         padding: 15px;
 
         border-bottom-left-radius: 20px;
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -18,11 +18,14 @@ $seat-size: 4vh;
   padding: $spacing;
   min-width: fit-content;
 
-  .video-container-wrapper {
+  .audience-overlay {
     position: absolute;
     width: 100%;
+    height: 100%;
     display: flex;
     justify-content: center;
+    align-items: center;
+    top: 1.5vh;
 
     .video-container {
       display: flex;
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -174,7 +174,7 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
       min-width: var(--seat-size-min);
       min-height: var(--seat-size-min);
 
-      font-size: calc(min(#{$seat-size}, #{$seat-size-min}) * 0.75);
+      font-size: calc(0.75 * min(var(--seat-size), var(--seat-size-min)));
 
       border-radius: 100%;
 
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -50,11 +50,6 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
           border-radius: 15px;
           box-shadow: 2px 2px 20px $light;
         }
-
-        .aspect-16-9 {
-          max-width: 88.5%;
-          margin-left: 6.5%;
-        }
       }
 
       .reaction-container {
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -2,12 +2,11 @@
 
 $light: #ffffff;
 
-$spacing: 30px;
-$video-spacing: 20px;
-
 // If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
 $seat-size: 4vh;
-$seat-size-min: 20px;
+$seat-size-min: 25px;
+
+$spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
 
 .audience-container {
   --seat-size: #{$seat-size};
@@ -24,7 +23,6 @@ $seat-size-min: 20px;
 
   .audience-overlay {
     position: absolute;
-    top: 1.5vh;
 
     width: 100%;
     height: 100%;
@@ -34,10 +32,12 @@ $seat-size-min: 20px;
     align-items: center;
 
     .video-container {
+      margin-top: calc(1 * min(var(--seat-size), var(--seat-size-min)));
+      padding: calc(1.5 * min(var(--seat-size), var(--seat-size-min)));
+
       display: flex;
       flex-direction: column;
       align-items: center;
-      padding: $video-spacing;
 
       .video {
         width: 100%;
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -10,133 +10,138 @@ $seat-size: 4vh;
 .audience-container {
   background-attachment: fixed;
   background-position: center 46%;
+  padding: $spacing;
 
   position: relative;
   display: flex;
+  justify-content: center;
   flex: 1;
 
   overflow: auto;
   width: 100%;
   height: 100%;
 
-  .video-container {
+  .video-container-wrapper {
     position: absolute;
-    // note: this is positioned using inline styles within the Audience component
-
-    margin: $spacing;
-    padding: $spacing;
-
+    z-index: 0;
+    width: 100%;
     display: flex;
-    flex-direction: column;
-    align-items: center;
+    justify-content: center;
 
-    .video {
-      width: 100%;
-      height: 100%;
-
-      .frame {
-        width: 100%;
-        height: 100%;
-        border: 1px black;
-        border-radius: 15px;
-        box-shadow: 2px 2px 20px $light;
-      }
-
-      .aspect-16-9 {
-        max-width: 88.5%;
-        margin-left: 6.5%;
-      }
-    }
-
-    .reaction-container {
+    .video-container {
       display: flex;
       flex-direction: column;
-      justify-content: space-around;
-      text-align: center;
+      align-items: center;
+      padding: 10px;
 
-      // these are fairly arbitrary figures
-      // min-width: 495px;
-      // max-width: 60%;
-      width: 90%;
-      min-height: 105px;
-      padding: 15px;
+      .video {
+        width: 100%;
+        height: 100%;
 
-      border-bottom-left-radius: 20px;
-      border-bottom-right-radius: 20px;
-      box-shadow: 2px 2px 20px $light;
-      background-color: #1a1d24;
-
-      .emoji-container {
-        display: flex;
-        flex: 1;
-        justify-content: space-around;
-        margin: 5px 0;
-
-        .reaction {
-          background-color: transparent;
-          border-color: transparent;
-          font-size: 20px;
-          width: 50px;
-          animation: sway 2200ms 1200ms ease-in-out;
+        .frame {
+          width: 100%;
+          height: 100%;
+          border: 1px black;
+          border-radius: 15px;
+          box-shadow: 2px 2px 20px $light;
         }
 
-        .mute-button {
-          display: flex;
-          align-items: center;
-          font-size: 20px;
-        }
-
-        .leave-seat-button {
-          display: inline-block;
-          font-weight: 700;
-          text-decoration: none;
-          border-radius: 8px;
-          text-align: center;
-          font-size: 0.7em;
-          transition: all 400ms cubic-bezier(0.23, 1, 0.32, 1);
-          line-height: 1.5;
-          border: 1px solid transparent;
-          transform: translateY(0);
-          cursor: pointer;
-          padding: 4px 10px;
-          background: rgb(0, 0, 0);
-          color: $white;
-
-          &:hover {
-            background-color: rgba(255, 255, 255, 0.5);
-          }
+        .aspect-16-9 {
+          max-width: 88.5%;
+          margin-left: 6.5%;
         }
       }
 
-      .shout-container {
-        margin: 0 8px 4px 8px;
-        padding-bottom: 4px;
+      .reaction-container {
+        display: flex;
+        flex-direction: column;
+        justify-content: space-around;
+        text-align: center;
 
-        .shout-form {
+        // these are fairly arbitrary figures
+        // min-width: 495px;
+        // max-width: 60%;
+        width: 90%;
+        min-height: 105px;
+        padding: 15px;
+
+        border-bottom-left-radius: 20px;
+        border-bottom-right-radius: 20px;
+        box-shadow: 2px 2px 20px $light;
+        background-color: #1a1d24;
+
+        .emoji-container {
           display: flex;
+          flex: 1;
+          justify-content: space-around;
+          margin: 5px 0;
 
-          input {
-            margin: 0 8px;
-            border-radius: 8px;
-            font-size: 1em;
-            font-weight: 500;
-            text-shadow: 0 1px 2px rgba(0, 0, 0, 0.3);
-            padding: 10px;
+          .reaction {
+            background-color: transparent;
+            border-color: transparent;
+            font-size: 20px;
+            width: 50px;
+            animation: sway 2200ms 1200ms ease-in-out;
           }
 
-          .text {
-            flex: 4;
+          .mute-button {
+            display: flex;
+            align-items: center;
+            font-size: 20px;
+          }
 
-            &:disabled {
-              cursor: not-allowed;
+          .leave-seat-button {
+            display: inline-block;
+            font-weight: 700;
+            text-decoration: none;
+            border-radius: 8px;
+            text-align: center;
+            font-size: 0.7em;
+            transition: all 400ms cubic-bezier(0.23, 1, 0.32, 1);
+            line-height: 1.5;
+            border: 1px solid transparent;
+            transform: translateY(0);
+            cursor: pointer;
+            padding: 4px 10px;
+            background: rgb(0, 0, 0);
+            color: $white;
+
+            &:hover {
+              background-color: rgba(255, 255, 255, 0.5);
             }
           }
+        }
 
-          .shout-button {
-            flex: 1;
+        .shout-container {
+          margin: 0 8px 4px 8px;
+          padding-bottom: 4px;
 
-            &:disabled {
-              cursor: not-allowed;
+          .shout-form {
+            display: flex;
+
+            input {
+              margin: 0 8px;
+              border-radius: 8px;
+              font-size: 1em;
+              font-weight: 500;
+              text-shadow: 0 1px 2px rgba(0, 0, 0, 0.3);
+              padding: 10px;
+            }
+
+            .text {
+              flex: 4;
+
+              &:disabled {
+                cursor: not-allowed;
+              }
+            }
+
+            .shout-button {
+              flex: 1;
+
+              &:disabled {
+                cursor: not-allowed;
+              }
             }
           }
         }
@@ -149,15 +154,13 @@ $seat-size: 4vh;
     flex: 1;
     flex-direction: column;
     justify-content: space-between;
-
-    padding: $spacing;
-
-    width: fit-content;
+    position: relative;
+    width: 100%;
     height: fit-content;
 
     .seat-row {
       display: flex;
-      // justify-content: space-between;
+      justify-content: center;
     }
 
     .seat {
@@ -165,11 +168,14 @@ $seat-size: 4vh;
       border-radius: 100%;
       width: $seat-size;
       height: $seat-size;
+      min-width: $seat-size;
+      min-height: $seat-size;
       justify-content: center;
       align-content: center;
       background-color: rgba(255, 255, 255, 0.2);
       color: white;
       user-select: none;
+      z-index: 1;
 
       &:hover {
         background-color: rgba(255, 255, 255, 0.4);
@@ -192,6 +198,8 @@ $seat-size: 4vh;
       border-radius: 100%;
       width: $seat-size;
       height: $seat-size;
+      min-width: $seat-size;
+      min-height: $seat-size;
       justify-content: center;
       align-content: center;
       background-color: rgba(0, 0, 0, 0);
@@ -200,7 +208,7 @@ $seat-size: 4vh;
 
     .add-participant-button {
       color: black;
-      font-size: 3vh;
+      font-size: calc(#{$seat-size} * 0.75);
       align-self: center;
     }
   }
@@ -215,6 +223,8 @@ $seat-size: 4vh;
   .user {
     width: $seat-size;
     height: $seat-size;
+    min-width: $seat-size;
+    min-height: $seat-size;
   }
 
   .profile-avatar {
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -7,8 +7,12 @@ $video-spacing: 20px;
 
 // If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
 $seat-size: 4vh;
+$seat-size-min: 20px;
 
 .audience-container {
+  --seat-size: #{$seat-size};
+  --seat-size-min: #{$seat-size-min};
+
   background-attachment: fixed;
   background-position: center 46%;
 
@@ -161,16 +165,32 @@ $seat-size: 4vh;
       justify-content: center;
     }
 
-    .seat {
+    .seat,
+    .not-seat,
+    .profile-avatar {
       display: flex;
-      border-radius: 100%;
-      width: $seat-size;
-      height: $seat-size;
       justify-content: center;
-      align-content: center;
-      background-color: rgba(255, 255, 255, 0.2);
-      color: white;
+      align-items: center;
+
+      width: var(--seat-size);
+      height: var(--seat-size);
+      min-width: var(--seat-size-min);
+      min-height: var(--seat-size-min);
+
+      font-size: calc(min(#{$seat-size}, #{$seat-size-min}) * 0.75);
+
+      border-radius: 100%;
+
+      color: $black;
       user-select: none;
+    }
+
+    .seat {
+      background-color: rgba(255, 255, 255, 0.2);
+
+      .add-participant-button {
+        z-index: z(audience-participant-add-button);
+      }
 
       &:hover {
         background-color: rgba(255, 255, 255, 0.4);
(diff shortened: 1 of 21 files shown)


src/components/templates/Audience/Audience.tsx
```diff
@@ -251,34 +251,37 @@ export const Audience: React.FunctionComponent = () => {
...
+    Math.ceil(columnsForSizedAuditorium / 5),
+    8
+  );
 
   // Keep a 16:9 ratio
-  const carvedOutHeightInSeats = Math.ceil(carvedOutWidthInSeats * (9 / 16));
+  const carvedOutHeightInSeats = Math.max(
+    Math.ceil(carvedOutWidthInSeats * (9 / 16)),
+    5
+  );
```
0xdevalias: Seems the comments are out of date from the code here now?

What is the `/ 5` used here for? Why does that change from what it was previously? (`3 * 2`)

It's probably worth pulling out the other `5` and the `8` into constants, as it was super confusing what they were used for. `MAX_CENTER_ROWS` / `MAX_CENTER_COLUMNS` perhaps? Or something like that?

src/components/templates/Audience/Audience.tsx
```diff
@@ -251,34 +251,37 @@ export const Audience: React.FunctionComponent = () => {
...
     (rowsForSizedAuditorium - carvedOutHeightInSeats * 2) / 2
   );
-  const videoContainerLeftOffsetInSeats = Math.floor(
-    (columnsForSizedAuditorium - carvedOutWidthInSeats * 2) / 2
+
+  const videoWrapperStyles = useMemo(
+    () => ({
+      top: `calc(${videoContainerTopOffsetInSeats} * ${SEAT_SIZE})`,
+    }),
+    [videoContainerTopOffsetInSeats]
```
0xdevalias, line 276: No `left` offset needed here?

src/components/templates/Audience/Audience.tsx
```diff
@@ -410,31 +413,35 @@ export const Audience: React.FunctionComponent = () => {
...
+                    allowFullScreen
+                  />
                 </div>
-              )}
+
+                {venue.showReactions && (
+                  <div
+                    className={`reaction-container ${
+                      userSeated ? "seated" : ""
+                    }`}
```
0xdevalias, line 437: Would be better to use the `classNames` helper here.

src/pages/Account/Account.scss
```diff
@@ -121,6 +121,7 @@ $page-max-width: 1240px;
   justify-content: center;
   flex: 1;
   background-color: #1a1d24;
+  height: 100%;
```
0xdevalias, line 124: Is `100%` fine here? Should it be `100vh` so that it is always 100% of the viewport (browser) height, rather than relying on the size of it's nearest parent?

src/scss/constants.scss
```diff
@@ -66,6 +66,8 @@ $z-layers: (
   // Audience
   audience-emoji-reaction: 1000,
   audience-video: 2,
+  audience-participant-add-button: 1,
+  audience-video-container: 0,
```
0xdevalias, line 70: I'd suspect that most of these `z-index`s aren't required.

src/scss/global.scss
```diff
@@ -7,6 +7,13 @@
 @import "node_modules/bootswatch/dist/darkly/bootswatch";
 @import "./constants.scss";
 
+html,
+body,
+#root {
+  height: 100%;
+  width: 100%;
+}
+
```
0xdevalias, line 16: Why adding this? I don't think we need to do it, and I think it may even be an antipattern (at least so it seemed in some of the things I was reading yesterday)

0xdevalias (changes requested):
Questions/comments/suggestions

src/components/templates/Audience/Audience.scss
```diff
@@ -2,142 +2,143 @@
...
-        max-width: 88.5%;
-        margin-left: 6.5%;
-      }
-    }
+  .video-container-wrapper {
+    position: absolute;
+    z-index: z(audience-video-container);
+    width: 100%;
+    display: flex;
+    justify-content: center;
```
0xdevalias, line 25: Playing with the styling a bit more, this looks like it will work, and should remove the direct `top` styles needed in react.

```suggestion
  .audience-overlay {
    position: absolute;
    top: 15px; /* Not sure why this is needed for alignment.. */
    left: 0;
    
    width: 100%;
    height: 100%;
    
    display: flex;
    justify-content: center;
    align-items: center;
```

Open to why this might be a worse idea than the current method though.

src/components/templates/Audience/Audience.scss
```diff
@@ -142,22 +151,23 @@ $seat-size: 4vh;
...
+      justify-content: center;
     }
 
     .seat {
       display: flex;
       border-radius: 100%;
       width: $seat-size;
       height: $seat-size;
+      min-width: $seat-size;
+      min-height: $seat-size;
```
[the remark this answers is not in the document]
mike-lvov: min-width/min-height ensures that the width is going to be $seat-size, even if their parent is smaller. I guess the reason I added it is because the seats shrank when you decreased the width

https://dev.to/ziizium/the-difference-between-css-width-min-width-and-max-width-property-49ah

src/components/templates/Audience/Audience.scss
```diff
@@ -193,8 +205,9 @@ $seat-size: 4vh;
 
     .add-participant-button {
       color: black;
-      font-size: 3vh;
+      font-size: calc(#{$seat-size} * 0.75);
       align-self: center;
+      z-index: z(audience-participant-add-button);
```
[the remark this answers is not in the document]
mike-lvov: The buttons are overlayed by video-overlay and were not accessible

src/components/templates/Audience/Audience.tsx
mike-lvov: It is centered with flexbox instead

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: Based on a quick skim through that link you provided, it doesn't sound like setting `width` and `min-width` to the same thing is going to have any difference in effect?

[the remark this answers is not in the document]
0xdevalias: I believe the main 'issue' with the sizing here is that we size the seats with `4vh`, so it's 4% of the viewport height (aka browser window height). That's why when the height is small, and the width is large, we get weird display issues.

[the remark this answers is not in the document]
0xdevalias: I believe what we actually want to do here is to use a fixed minimum width/height, probably in `px`. That would allow us to force a minimum size.

We would then also want to use that same minimum when calculating the position/size/etc of the video container within the middle 'cut out area'. If we ensure that that middle area always has enough size that the emoji area will fit, then we should be fine, regardless of screen size/weird dimensions/etc.

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: Shouldn't the video overlay never sit over them though? When there are no `z-index`s set, it uses the 'element order' in the code, and the div for the overlay is before all of the seats. So the seats should be 'on top'.

src/components/templates/Audience/Audience.tsx
0xdevalias: Makes sense

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: I don't think so. You can try removing this

src/components/templates/Audience/Audience.scss
```diff
@@ -12,124 +12,133 @@ $seat-size: 4vh;
...
 
-  .video-container {
+  .video-container-wrapper {
     position: absolute;
-    // note: this is positioned using inline styles within the Audience component
-
-    margin: $spacing;
-    padding: $spacing;
-
+    z-index: z(audience-video-container);
```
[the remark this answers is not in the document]
mike-lvov: removed in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.scss
```diff
@@ -12,124 +12,133 @@ $seat-size: 4vh;
...
+    justify-content: center;
 
-    .video {
-      width: 100%;
-      height: 100%;
+    .video-container {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      padding: 10px;
```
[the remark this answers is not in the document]
mike-lvov: added in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.scss
```diff
@@ -12,124 +12,133 @@ $seat-size: 4vh;
...
+        text-align: center;
 
-      // these are fairly arbitrary figures
-      min-width: 495px;
-      max-width: 60%;
-      min-height: 105px;
-      padding: 15px;
+        // these are fairly arbitrary figures
+        // min-width: 495px;
+        // max-width: 60%;
```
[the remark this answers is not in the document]
mike-lvov: removed in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: removed in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.scss
```diff
@@ -185,6 +195,8 @@ $seat-size: 4vh;
       border-radius: 100%;
       width: $seat-size;
       height: $seat-size;
+      min-width: $seat-size;
+      min-height: $seat-size;
```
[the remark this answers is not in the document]
mike-lvov: removed in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.scss
```diff
@@ -208,6 +221,8 @@ $seat-size: 4vh;
   .user {
     width: $seat-size;
     height: $seat-size;
+    min-width: $seat-size;
+    min-height: $seat-size;
```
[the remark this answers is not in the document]
mike-lvov: removed in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.tsx
mike-lvov: Fixed/Added in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/components/templates/Audience/Audience.tsx
mike-lvov: [OOS] but addded in https://github.com/sparkletown/sparkle/pull/1237/commits/6ff3b5876858d4e306936f59a383e5f0e9c734aa

src/scss/global.scss
mike-lvov: Try removing this. It breaks the scrolling capabilities.

src/scss/constants.scss
mike-lvov: Deleted the unneded one

src/pages/Account/Account.scss
mike-lvov: 100% results in taking 100% of the viewport

src/components/templates/Audience/Audience.scss
```diff
@@ -142,22 +151,23 @@ $seat-size: 4vh;
     flex: 1;
     flex-direction: column;
     justify-content: space-between;
-
-    padding: $spacing;
-
-    width: fit-content;
```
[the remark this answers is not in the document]
mike-lvov: I played with it and it worked(removing the width)

src/components/templates/Audience/Audience.scss
mike-lvov: Added here, but replaced pixels with vh
https://github.com/sparkletown/sparkle/pull/1237/commits/4f768f4803498e25ab6ed9b5c963139ab4536de2

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: @mike-lvov Did you read the other comments I added here? As they are directly related to ensuring the emoji container doesn't break out/etc.

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: Yup, seems you are correct. I expect it's because it's `position: absolute`, so sits outside of the normal flow I was referring to.

A better solution than `z-index` for this case is to use `pointer-events: none` on `audience-overlay` though:

- https://css-tricks.com/almanac/properties/p/pointer-events/

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: > We would then also want to use that same minimum when calculating the position/size/etc of the video container within the middle 'cut out area'. If we ensure that that middle area always has enough size that the emoji area will fit, then we should be fine, regardless of screen size/weird dimensions/etc.

I do agree that this is the best approach to handle all of the use cases, but I don't think it is time-efficient to implement this since I'm not really aware of the logic that is used for `cut out area` size. It will take a substantial amount of my time to get into this(to only handle a case with really wide and short screens)

src/pages/Account/Account.scss
0xdevalias: Not technically. 100% is 100% relevant to its parent container. `100vh` always represents 100% of the viewport.

- https://www.lullabot.com/articles/unexpected-power-of-viewport-units-in-css

mike-lvov: Technically, you are correct, but practically, 100%, in this case, results in 100% viewport. I don't think we have time for this discussion, so changed in 
https://github.com/sparkletown/sparkle/pull/1237/commits/33d029f0825d052343fabc5be8c570e31b3bcabe

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: The logic remains the same, it calculates the rows/columns/etc. The difference is that we need to update the inline style applied to positioning our container so that it also uses the same minimum.

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: Fair enough. I think by default things get set to 100% width of their parent anyway, so perhaps that is why.

src/components/templates/Audience/Audience.scss
0xdevalias: Is using `vh` here correct though? I believe `15px` remained perfectly aligned regardless of the size of the browser, whereas `vh` is going to change based on the height. While I couldn't see why it was required, I feel like it is going to be a `15px` `margin`/`padding` somewhere, which is why I set it in `px`.

0xdevalias: Also, the order of the keys was explicitly laid out as I wanted it, with position/top/left being together, and all of the flex stuff being together, etc. Yet that didn't translate through when you made the changes.

src/pages/Account/Account.scss
0xdevalias: I believe it's only practically that way because of the 100% you set on the body/etc, which I believe is incorrect.

Thankyou for changing it.

src/scss/global.scss
0xdevalias: Just pulled down your latest changes + toggled them off. Scrolling seems to work fine in the auditorium/etc; and looking at the calculated height/width of each element in the 'chain' going down seems like they are correct?

Where are you seeing it not work/in what situations?

0xdevalias: Using this size for all of the below cases:
image

---

Using these styles:
image

The computed size is:
image

---

Using these styles:
image

The computed size is:
image

---

Unsetting these styles completely:
image

The computed size is:
image

src/components/templates/Audience/Audience.scss
mike-lvov: `perfectly aligned regardless of the size of the browser` it didn't though. If you make the height of the screen small enough, it broke out a little. But I don't think that it has very huge difference either case

src/scss/global.scss
mike-lvov: It doesn't allow to scroll to the left side

Uploading Screen Recording 2021-03-05 at 12.01.11.mov…

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: Do you know exactly what to do? If so, could you apply those changes, please?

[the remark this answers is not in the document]
0xdevalias: I believe I do, was just playing with it in devtools..

I believe this is the calculation that would work for the video container part (obviously using the correct row/col vals from the code):

```
element.style {
    width: calc(20 * min(20px, 4vh));
    height: calc(15 * min(20px, 4vh));
}
```

And this is what we need to apply to `seat`, `not-seat`, and `avatar`:

```css
.audience-container .audience .seat {
    // ..snip..
    width: 4vh;
    height: 4vh;
    min-width: 20px;
    min-height: 20px;
    // ..snip..
}
```

Assuming `20px` is the value we want to use as a minimum. Basically, whatever we pick should allow enough space for the emoji reactions section even if we don't change it to make it responsive.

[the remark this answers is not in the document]
0xdevalias: Got some changes in my working copy, will commit them in a minute.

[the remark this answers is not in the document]
0xdevalias: Added in https://github.com/sparkletown/sparkle/pull/1237/commits/0beac219a5a4474071574e883b6a7b828e7af916 it's still not 100% perfect, I believe to do so we would just need to pick a minimum `px` size that would allow the number of 'carved out rows' worth of sizing to account for the non-responsive reaction container.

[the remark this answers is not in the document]
mike-lvov: @0xdevalias this has done the trick, I believe
https://github.com/sparkletown/sparkle/pull/1237/commits/958c41f091f15e3cc0b9627a6dfd86e76216578f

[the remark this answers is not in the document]
0xdevalias: See the changes I made in https://github.com/sparkletown/sparkle/pull/1237/commits/88a23c5ae42316b2bbb9d3e46b47b5fc186ba259, which I believe may make your change unneeded.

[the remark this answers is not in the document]
0xdevalias: Even at ridiculous dimensions, it all stays within the middle container now:

image

image

---

iPhone in landscape:
image

iPad:
image

Hi DPI Laptop:
image

Medium DPI Laptop:
image

src/scss/global.scss
0xdevalias: FYI: you can attach videos directly to GitHub issues/comments/etc now, just drag/drop/upload it.

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: @0xdevalias Reverted

src/scss/global.scss
mike-lvov: https://user-images.githubusercontent.com/63194656/110104473-df0f4200-7daf-11eb-96fc-20bc7c191da5.mov

0xdevalias: Yup.. I see the issue. Had a quick play/look around and not sure why tbh.

Also had a quick google and couldn't find any sources backing up my "it's an antipattern" thoughts from whatever I was reading yesterday either.

So based on that, I say let's keep it! 👍🏻

mike-lvov: It lacks some perspective ^, but basically, I can scroll to the right side(to the end), but can't do it to the left side

src/components/templates/Audience/Audience.scss
```diff
@@ -2,142 +2,149 @@
 
 $light: #ffffff;
 
+// If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
+$seat-size: 4vh;
+$seat-size-min: 25px;
```
0xdevalias, line 7: [NB] This comment might not be needed anymore since we use the `--seat-size`/etc variables in the component now too. Probably doesn't hurt to leave it here though..

src/components/templates/Audience/Audience.scss
```diff
@@ -2,142 +2,149 @@
...
-      }
+  overflow: auto;
+  height: 100%;
+  padding: $spacing;
+  min-width: fit-content;
 
-      .aspect-16-9 {
-        max-width: 88.5%;
-        margin-left: 6.5%;
-      }
```
0xdevalias, line 45: We should probably remove any references of this class from the component if we haven't already.

src/components/templates/Audience/Audience.scss
```diff
@@ -148,28 +155,44 @@ $light: #ffffff;
...
-      background-color: rgba(255, 255, 255, 0.2);
-      color: white;
+      align-items: center;
+
+      width: var(--seat-size);
+      height: var(--seat-size);
+      min-width: var(--seat-size-min);
+      min-height: var(--seat-size-min);
+
+      font-size: calc(min(#{$seat-size}, #{$seat-size-min}) * 0.75);
```
0xdevalias, line 182:
```suggestion
      font-size: calc(0.75 * min(var(--seat-size), var(--seat-size-min)));
```

src/components/templates/Audience/Audience.scss
```diff
@@ -148,28 +155,44 @@ $light: #ffffff;
...
+
+      color: $black;
       user-select: none;
+    }
+
+    .seat {
+      background-color: rgba(255, 255, 255, 0.2);
+
+      .add-participant-button {
+        z-index: z(audience-participant-add-button);
```
0xdevalias, line 194: [note] I tried removing this + tried a bunch of alternatives, still had issues, not sure why. tl;dr: Worth keeping it for now.

src/components/templates/Audience/Audience.tsx
```diff
@@ -46,6 +43,14 @@ interface ChatOutDataType {
   text: string;
 }
 
+// If you change this, make sure to also change it in Audience.scss's $seat-size
+const SEAT_SIZE = "var(--seat-size)";
+const SEAT_SIZE_MIN = "var(--seat-size-min)";
```
0xdevalias, line 48: [NB] As per my other comment in this review, if we think there is better wording we could use here, feel free to make it better.

src/components/templates/Audience/Audience.tsx
```diff
@@ -46,6 +43,14 @@ interface ChatOutDataType {
...
+
+const VIDEO_MIN_WIDTH_IN_SEATS = 8;
+// We should keep the 16/9 ratio
+const VIDEO_MIN_HEIGHT_IN_SEATS = (VIDEO_MIN_WIDTH_IN_SEATS * 9) / 16;
```
0xdevalias, line 52: I think this makes the intent of it slightly clearer.

```suggestion
const VIDEO_MIN_HEIGHT_IN_SEATS = VIDEO_MIN_WIDTH_IN_SEATS * (9 / 16);
```

src/components/templates/Audience/Audience.tsx
```diff
@@ -235,22 +256,46 @@ export const Audience: React.FunctionComponent = () => {
   const rowsForSizedAuditorium = minRows + auditoriumSize * 2;
   const columnsForSizedAuditorium = minColumns + auditoriumSize * 2;
 
+  // 3 because 1/3 of the sie of the auditorium, * 2 because we're calculating in halves due to using cartesian coordinates + Math.abs
```
0xdevalias, line 259:
```suggestion
  // We use 3 because 1/3 of the size of the auditorium, and * 2 because we're calculating in halves due to using cartesian coordinates + Math.abs
```

src/components/templates/Audience/Audience.tsx
```diff
@@ -365,27 +414,34 @@ export const Audience: React.FunctionComponent = () => {
...
+                  />
+                </div>
+
+                {venue.showReactions && (
+                  <div className={reactionContainerClassnames}>
+                    {userSeated
+                      ? renderReactionsContainer()
+                      : renderInstructions()}
+                  </div>
+                )}
```
0xdevalias, line 441: It might be worth checking how this looks when `venue.showReactions` is turned off too..

src/scss/constants.scss
```diff
@@ -66,6 +66,7 @@ $z-layers: (
   // Audience
   audience-emoji-reaction: 1000,
   audience-video: 2,
+  audience-participant-add-button: 1,
```
0xdevalias, line 69: [NB] I wonder if instead of setting this z-index, if we set `-1` on the overlay.. would that fix our problems here..?

src/scss/global.scss
0xdevalias: _nods_ yeah, I saw that. And while it frustrates me that I can't explain why that is. Your change fixes it. So :shipit:

0xdevalias (approved):
A few little comments/potential cleanup/tweaks (that if you agree, maybe do before merging), but overall, I say this is good to go! LGT(Finally)M! :shipit:

src/scss/constants.scss
mike-lvov: @0xdevalias I tried doing that. You are not able to click the buttons on the reaction bar, if you do that

src/components/templates/Audience/Audience.tsx
mike-lvov: Looks good
image

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
0xdevalias: Awesome, sounds good!

src/components/templates/Audience/Audience.tsx
0xdevalias: Yay!

src/components/templates/Audience/Audience.scss
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1237/commits/48743b83a4ffdd69726fa4e1b101f40cc67a636b

src/scss/constants.scss
0xdevalias: _nods_ fair.

PR #1239, opened 2021-03-05
Minor tweaks to Auditorium styles + related cleanup

0xdevalias:
Follow on to https://github.com/sparkletown/sparkle/pull/1237 that cleans up a few more little bits and pieces.

@mike-lvov It seems this will allow the seats and everything to still be clickable, without us needing to use an explicit `z-index`/etc.

Related reading:

- https://css-tricks.com/almanac/properties/p/pointer-events/
- https://caniuse.com/pointer-events
- https://stackoverflow.com/questions/3680429/click-through-div-to-underlying-elements/26799885#26799885

### DIFF
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -63,10 +63,6 @@ $z-layers: (
   admin-placement: 10,
   admin-venue-header-after: 2,
   admin-venue-header: -1,
-  // Audience
-  audience-emoji-reaction: 1000,
-  audience-video: 2,
-  audience-participant-add-button: 1,
   // JazzBar
   jazzbar-participant-profile-icon: 1,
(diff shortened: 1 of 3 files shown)


src/scss/constants.scss
```diff
@@ -63,10 +63,6 @@ $z-layers: (
   admin-placement: 10,
   admin-venue-header-after: 2,
   admin-venue-header: -1,
-  // Audience
-  audience-emoji-reaction: 1000,
-  audience-video: 2,
-  audience-participant-add-button: 1,
```
0xdevalias, line 69: The other 2 entries here were unused.

mike-lvov (approved):
LGTM

PR #1247, opened 2021-03-08
Fix sidebar chat to streamline navigating back to the list of all DMs

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/439

> As a user having a direct chat, I should be able to navigate easily to other DMs.


denisdimitrov (approved):
LGTM!

0xdevalias (approved):
LGTM!

---

@igachmovmelon Please make sure to include a descriptive/useful title + relevant context from the original issue so that anyone who view this public/open-source repo can understand the changes made here (as they won't have access to our private issues board). I've noticed this as an issue on a few of your PRs now, so want to make sure to call it out directly so you can improve in that area.

PR #1248, opened 2021-03-08
Filter online users in ChatSidebar PrivateChats to exclude those with an existing DM

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/478

chat

## Questions

After filtering the online users should the number of connected users remain "X other connected people" or "X other online people"  ?


src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -34,6 +34,10 @@ export const PrivateChats: React.FC<PrivateChatsProps> = ({
   const { selectRecipientChat } = useChatSidebarControls();
 
   const numberOfOnline = onlineUsers.length;
+  const privateChatUsers: String[] = [];
+  privateChatPreviews.forEach((user) =>
+    privateChatUsers.push(user.counterPartyUser.id)
```
denisdimitrov, line 39: `.map` creates a new array. Use it instead of `forEach` and `push`

```tsx
const privateChatUsers = privateChatPreviews.map(chatPreview => chatPreview.counterPartyUser.id)
```

denisdimitrov (changes requested):

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
mike-lvov: I also suggest renaming `privateChatUsers` into for example `contactedChatUsers`

0xdevalias: +1 to both @denisdimitrov and @mike-lvov 's feedback here

0xdevalias (changes requested):
To answer your original question:

> After filtering the online users should the number of connected users remain 4 because they are 3 + 1 (the person with who you have direct message) or the number needs to be 3 ?

I would change the label here to something like "X other connected people" or "X other online people" or similar.

Then I would make it use `renderedOnlineUsers.length`, as the number shown here should match the number of users we show below the label.

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
0xdevalias: > I also suggest renaming privateChatUsers into for example contactedChatUsers

Eh.. maybe. But if you're going to do that, then you would probably also want/need to change `privateChatPreviews` to some more meaningful/useful name. Feels a little irrelevant/OOS to me. If anything, I feel like `privateChatUsers` is potentially a good name, and `privateChatPreviews` could change. But at the end of the day, not sure that either need to.

0xdevalias (dismissed):
It looks like part of my request in https://github.com/sparkletown/sparkle/pull/1248#pullrequestreview-606912710 was implemented, but not the rest of it?

> I would change the label here to something like "X other connected people" or "X other online people" or similar.

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -33,7 +33,9 @@ export const PrivateChats: React.FC<PrivateChatsProps> = ({
   const onlineUsers = useOnlineUsersToDisplay();
   const { selectRecipientChat } = useChatSidebarControls();
 
-  const numberOfOnline = onlineUsers.length;
+  const privateChatUsers = privateChatPreviews.map(
+    (chatPreview) => chatPreview.counterPartyUser.id
```
mike-lvov, line 37: `privateChatUserIds`

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -83,6 +87,7 @@ export const PrivateChats: React.FC<PrivateChatsProps> = ({
 
   const numberOfSearchResults = renderedSearchResults.length;
   const hasChatPreviews = renderedPrivateChatPreviews.length > 0;
+  const numberOfOnline = renderedOnlineUsers.length;
```
mike-lvov, line 90: `numberOfOnlineUsers`

mike-lvov: And please, change the text of the label as suggested by @0xdevalias to `${numberOfTherOnlineUsers} other online users`

mike-lvov (approved):
LGTM

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -33,7 +33,9 @@ export const PrivateChats: React.FC<PrivateChatsProps> = ({
   const onlineUsers = useOnlineUsersToDisplay();
   const { selectRecipientChat } = useChatSidebarControls();
 
-  const numberOfOnline = onlineUsers.length;
+  const privateChatUsers = privateChatPreviews.map(
+    (chatPreview) => chatPreview.counterPartyUser.id
+  );
```
mike-lvov: Could you also wrap it into `useMemo`, please?

PR #1249, opened 2021-03-08
fix audience unclickable video controls

denisdimitrov:
The video was inheriting the `pointer-events: none;` from the parent styles and the video was unclickable because of it.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/556

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -41,6 +41,8 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
       flex-direction: column;
       align-items: center;
 
+      pointer-events: all;
+
       .video {
         width: 100%;
         height: 100%;


denisdimitrov:
@0xdevalias I'm self approving this because it's urgent. Not sure if this is a quick fix or the real thing.

src/components/templates/Audience/Audience.scss
```diff
@@ -41,6 +41,8 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
       flex-direction: column;
       align-items: center;
 
+      pointer-events: all;
```
0xdevalias, line 44: - https://css-tricks.com/almanac/properties/p/pointer-events/
  - While the pointer-events property takes eleven possible values, all but three of them are reserved for use with SVG. The three valid values for any HTMl element are:
    - `none`: prevents all click, state and cursor options on the specified HTML element
    - `auto`: restores the default functionality (useful for use on child elements of an element with `pointer-events: none;` specified
    - `inherit` will use the pointer-events value of the element’s parent
    
---

Based on the above, it looks like we want to use `pointer-events: auto;` here, rather than `all`

0xdevalias (commented):
@denisdimitrov Retroactive review, looks like we want to change this for the 'proper fix'.

src/components/templates/Audience/Audience.scss
0xdevalias: Backlogged in https://github.com/sparkletown/internal-sparkle-issues/issues/578 to ensure this isn't missed/lost.

PR #1252, opened 2021-03-09
[tests] add cypress tests

malaikaaiyar:
Basic 'v1' tests:
Logs in and out, enters and goes to all the tables in the jazzbar, goes into auditorium.

For future: interact with the elements in the venues (make sure video works, click around, react, send chat messages, etc)


cypress/integration/login-test.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-undef */
-const EMAIL = 'malaika.aiyar@gmail.com';
-const PASSWORD = 'WER$wer4';
```
0xdevalias: @malaikaaiyar If you haven't already, since this password is exposed now, you should change it.

cypress/integration/login-test.js
```diff
@@ -0,0 +1,95 @@
...
+    cy.clearCookies();
+
+    cy.get(".login-button")
+      .click();
+
+    cy.get(".form > .input-group:nth-child(1) > input")
+      .type('malaika.aiyar@gmail.com');
+
+    cy.get(".form > .input-group:nth-child(2) > input")
+      .type('QWE#qwe3');
```
0xdevalias: @malaikaaiyar If you haven't already, since this password is exposed now, you should change it.

cypress/integration/login-test.js
malaikaaiyar: I have

cypress/integration/login-test.js
```diff
@@ -0,0 +1,130 @@
...
+  cy.get(".form > .input-group:nth-child(2) > input")
+    .type(PASSWORD);
+  cy.get('.form > .btn')
+    .click();
+}
+
+const logOut = () => {
+  cy.get('.navbar-link-profile')
+    .click();
+  cy.get(".popover-body > .profile-modal-container > input:nth-child(7)")
```
yarikoptic, line 19: Didn't check actual HTML for elements, but might be desired to assign unique `id` to elements to avoid hardcoding full paths where avoidable --  so tests would become more "semantically driven" and less fragile to code changes.

cypress/integration/login-test.js
```diff
@@ -0,0 +1,130 @@
...
+    .click();
+}
+
+describe("Test staging.sparkle.space/in/devaliashacksville login, logout, jazzbar and auditorium", function () {
+
+  it("can log in and out", () => {
+    cy.visit("https://staging.sparkle.space/in/devaliashacksville");
```
yarikoptic, line 26: define top url globally and avoid duplication of the value?

cypress/integration/login-test.js
```diff
@@ -0,0 +1,130 @@
+/* eslint-disable no-undef */
+const EMAIL = '';
+const PASSWORD = '';
```
yarikoptic, line 3: to possibly come from env var or some other config specification for automation etc

yarikoptic:
Seems not yet included to be ran within CI somewhere, right?

cypress/integration/login-test.js
0xdevalias: Strong disagree on using id’s. Super nom-react’y way of approaching things, often causes more problems than it solves.

Our newer patterns of class naming (following BEM patterns) gives us ‘component name paths’ that can be used in most cases.

I haven’t looked at the specific paths here, but they may also be able to be optimised further.

cypress/integration/login-test.js
0xdevalias: 100% should be injected from ENV

0xdevalias:
> Seems not yet included to be ran within CI somewhere, right?

I haven’t looked specifically deeper at this PR, but there was previously cypress CI integration hooked up. It’s commented out I believe (look for smoke test in the circleci config files), but I think it wouldn’t take all that much to ‘bring back.

My preference would be to switch to doing this on GitHub actions though, as that’s more the direction we want to move with most of our CI, and keep things more ‘in this platform/ecosystem’. There’s a bunch of good things on the marketplace to handle this without us needing to reinvent the wheel.

mobilevinay:
migrated to zenhub, check!

yarikoptic:
is that zenhub public? could there be a back reference placed along with the announcement of a "migration"?

0xdevalias:
@yarikoptic Zenhub itself isn’t public, but it’s just a layer that largely sits over GitHub issues and adds some nicer sprint/planning/etc tools (eg. Epics, etc) It still uses GitHub issues as it’s ‘base’.

these comments were probably rather confusing for @mobilevinay tl have plastered across our open source issues without further context/explanation, so I’m not surprised that you found it somewhat confusing.

PR #1253, opened 2021-03-10
admin v3 start screen - add create logic in view screen and add new route

denisdimitrov:
pending on #1190


denisdimitrov:
https://github.com/sparkletown/sparkle/pull/1190#discussion_r592244657 Leaving this here to avoid getting it lost.

src/pages/Admin/Details/Form/DetailsForm.styles.ts
```diff
@@ -29,13 +29,82 @@ export const InputContainer = styled.div`
...
+  }
 
     &:focus {
       outline: none;
     }
+
+  }
+`;
+
+export const BannerContainer = styled.div`
```
mike-lvov, line 66: If this file is going to be used in V3, we should move the styles into scss

mike-lvov (commented):
This PR is in a draft mode and @denisdimitrov said it needs polish, will make a final pass review after those are done

denisdimitrov:
Closing the PR because it's outdated and both the structure of admin v3 and the design has changed drastically. The effort to resolve the conflicts and redesign it would be more than just re-writing the logic.

PR #1256, opened 2021-03-12
ensure unclickable rooms maintain their z-index onHover

0xdevalias:
Unclickable rooms were added in https://github.com/sparkletown/sparkle/pull/1225, but it seems we missed an edgecase. When hovering over rooms, the `.maproom:hover` `z-index` was being applied.

Also cleaned up a duplicate style/etc.


mike-lvov (approved):
LGTM

PR #1262, opened 2021-03-15
add notification sound when receiving new private message

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/179

Compare if there is a new message, if there is any looking if the new message is from another user and playing the sound. Also comparing if the user is in current chat and receiving new messages in this chat to ignore the sound.


src/hooks/privateChats.ts
```diff
@@ -44,6 +49,60 @@ export const usePrivateChatMessages = () => {
...
+        play();
+      }
+    }
+  }, [
+    privateChatMessages,
+    chatWithUser,
+    play,
+    privateChatMessagesState,
+    userId,
+  ]);
```
0xdevalias, line 105: The relevant code related to the notification sounds should be wrapped up in a custom hook (eg. `src/hooks/useNotificationSound.ts`), similar to how the previous PR was attempting to go (though properly using `useSound` and similar as per this PR):

- https://github.com/sparkletown/sparkle/pull/1180/files#diff-a22d769c39c4171c61bd19fa96454692ce900a294f3507e3c841a23086b17e47

src/settings.ts
```diff
@@ -39,6 +39,7 @@ export const SPARKLE_ICON = "/sparkle-icon.png";
 export const DEFAULT_MAP_BACKGROUND = "/maps/Sparkle_Field_Background.jpg";
 export const DEFAULT_VENUE_BANNER = "/assets/Sparkle_Banner_Default.jpg";
 export const DEFAULT_VENUE_LOGO = "/assets/Sparkle_SquareLogo_Default.jpg";
+export const NEW_MESSAGE_SOUND = "/sounds/newmessage.m4a";
```
0xdevalias, line 42: We don't want to use 'magic strings' to import this asset. Please use proper webpack imports as per the previous PR:

- https://github.com/sparkletown/sparkle/pull/1180/files#diff-433083e9966b3a808a0b2dce3e47b0b74152c61cb7d5b6a4ad8aad2592dacd81
- https://github.com/sparkletown/sparkle/pull/1180/files#diff-a22d769c39c4171c61bd19fa96454692ce900a294f3507e3c841a23086b17e47

src/hooks/privateChats.ts
```diff
@@ -44,6 +49,60 @@ export const usePrivateChatMessages = () => {
...
+    if (
+      privateChatMessages !== undefined &&
+      privateChatMessagesState !== undefined &&
+      newMessage !== undefined &&
+      newMessage.length !== 0 &&
+      privateChatMessages.length !== privateChatMessagesState.length &&
+      !newMessage[0].isRead &&
+      newMessage[0].from !== userId
+    ) {
+      setPrivateChatMessagesState(privateChatMessages);
```
0xdevalias, line 91: I can't be certain off the top of my head as I'm having trouble following the logic here, but this seems way more complex than I would have expected this code needed to be. Is there any way we can simplify it? Or if not, what are the reasons/edge cases that require it to be so verbose?

0xdevalias (changes requested):
Comments/suggestions/etc, mostly in bringing this back into alignment with the patterns from the previous PR as we want them to remain.

---

I can't comment directly on the file, but `public/sounds/newmessage.m4a` should be moved to `src/assets/sounds/notification.m4a`

src/hooks/privateChats.ts
igachmovmelon: @0xdevalias the logic is first to find which is the new message that the user receives. After that we check if the arrays are empty the first array is the one that comes from the firebase server the other one is the old one that was before the new message. We are also looking for if the new message is also empty and defined. Now we are looking if there is a change between the old array of messages or new array if there is we look if the new message is read if it isn't read we need to see if the message that came from the server is from us or not. In that way we can find out that the user has received message and it isn't from him.
There are 2 more ifs in this one that help us found if the user is in the chat or not, because if he is in the chat there should not be notification playing. So to sum up this if helps us find if there is a new message and if the user is in the chat or not.

mike-lvov:
Looking at the code now and I'm thinking that we should reuse useCustomSounds here. @0xdevalias WDYT? I know its implementation is all about the rooms, but at least storing the sounds & accessing them should be possible

0xdevalias:
> Looking at the code now and I think that we should reuse useCustomSounds here. @0xdevalias WDYT? I know its implementation is all about the rooms, but at least storing the sounds & accessing them should be possible

@mike-lvov Nah, my intent was to only use that for custom custom things, where we would eventually intend it to be controlled by the admin UI/etc.

For these more 'standard platform sounds' I would still be intending to do the 'import the sound file using webpack' sort of thing, which from memory is what is being done here already (haven't looked at the code again on my mobile). It would still be passed to the standard useSound hook (from the lib), and that should be wrapped up in a custom hook for this usecase (as I believe I described in previous review comments somewhere, similar to the pattern from Jigar's PR).

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+import notificationSound from "assets/sounds/newmessage.m4a";
+
+import { PrivateChatMessage, ChatTypes } from "types/chat";
+
+import { useChatSidebarControls } from "hooks/chatSidebar";
+
+import { useUser } from "./useUser";
+import { WithId } from "utils/id";
+
+export const useNotificationSound = (
```
0xdevalias, line 13: Given this is explicitly tied to private chats now, it might be better to rename it (both hook and file) to something like `usePrivateChatNotification`/`useNewPrivateChatNotification`/similar

0xdevalias: Part of me feels like it almost belongs in the `src/hooks/privateChats.ts` file too, instead of it's own file.

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+import useSound from "use-sound";
+
+import notificationSound from "assets/sounds/newmessage.m4a";
+
+import { PrivateChatMessage, ChatTypes } from "types/chat";
+
+import { useChatSidebarControls } from "hooks/chatSidebar";
+
+import { useUser } from "./useUser";
+import { WithId } from "utils/id";
```
0xdevalias, line 11: [nit] `utils/*` imports should be sorted after `types/*` imports

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+  const chatWithUser =
+    chatSettings.openedChatType === ChatTypes.PRIVATE_CHAT
+      ? chatSettings.recipientId
+      : undefined;
+
+  useEffect(() => {
+    const newMessage = privateChatMessages?.filter(
+      ({ id: id1 }) =>
+        !privateChatMessagesState?.some(({ id: id2 }) => id2 === id1)
+    );
```
0xdevalias, line 40: Instead of `id1` `id2` can we give these semantically meaningful names please.

Maybe something like `existingChatMessageId` and `newPrivateChatMessageId` or similar.

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    if (
+      privateChatMessages !== undefined &&
+      privateChatMessagesState !== undefined &&
+      newMessage !== undefined &&
+      newMessage.length !== 0 &&
+      privateChatMessages.length !== privateChatMessagesState.length &&
+      !newMessage[0].isRead &&
+      newMessage[0].from !== userId
+    ) {
```
0xdevalias, line 54: We're using TypeScript, so we can just use the `?.` 'optional chaining' navigator to reduce the verbosity here:

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

```suggestion
    if (
      privateChatMessages?.length !== privateChatMessagesState?.length &&
      newMessage?.length > 0
      !newMessage[0].isRead &&
      newMessage[0].from !== userId
    ) {
```

src/hooks/useNotificationSound.ts
0xdevalias: Looking at the code later on, you only ever seem to be accessing `newMessage` as `newMessage[0]`, implying that you're only ever checking the first message returned, which makes either that logic wrong, or it pointless to use `.filter` here (instead of `.some` or similar)

src/hooks/useNotificationSound.ts
0xdevalias: [note] I last asked about the logic here in https://github.com/sparkletown/sparkle/pull/1262#discussion_r594125910 and it was explained in https://github.com/sparkletown/sparkle/pull/1262#discussion_r597416377

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+      setPrivateChatMessagesState(privateChatMessages);
+
+      if (chatWithUser === undefined) {
+        play();
+      } else if (chatWithUser !== newMessage[0].from) {
+        play();
+      }
+    }
```
0xdevalias, line 62: This can be simplified to something like this:

```suggestion
      if (chatWithUser === undefined || chatWithUser !== newMessage[0].from) {
        play();
      }
    }
```

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const [play] = useSound(notificationSound, {
+    // `interrupt` ensures that if the sound starts again before it's
+    // ended, it will truncate it. Otherwise, the sound can overlap.
+    interrupt: true,
+  });
+
+  const [privateChatMessagesState, setPrivateChatMessagesState] = useState(
+    privateChatMessages
+  );
```
0xdevalias, line 29: I'm guessing that we're using `privateChatMessagesState` to hold the previous value of `privateChatMessages` here, so that we can then compare the messages later, to determine if we need to play the sound?

If so, since we don't need to force re-renders when we update the data, it might be better to use `useRef` here rather than `useState`:

That's that basic pattern that utility hooks such as the following (which we don't currently have in our dependencies) use:

- https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state
  - based on this pattern
- https://github.com/streamich/react-use/blob/master/docs/usePrevious.md
- https://github.com/streamich/react-use/blob/master/docs/usePreviousDistinct.md

src/hooks/useNotificationSound.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const chatWithUser =
+    chatSettings.openedChatType === ChatTypes.PRIVATE_CHAT
+      ? chatSettings.recipientId
+      : undefined;
```
0xdevalias, line 34: Maybe rename this something more informative such as `currentChatUserId` or similar

src/hooks/useNotificationSound.ts
0xdevalias: I'm not sure we want to keep an entire copy of all of the private chat messages in memory just to determine if there is a new one though. Could we just store the last updated time/similar, and then filter/process only messages received after that instead?

0xdevalias: `PrivateChatMessage` extends from `BaseChatMessage` which has a `ts_utc` (`firebase.firestore.Timestamp`) field:

- https://github.com/sparkletown/sparkle/blob/staging/src/types/chat.ts#L4-L14

The first time this hook is run, we can sort `privateChatMessages` by `ts_utc` (see `chatSort` in `src/utils/chat.ts`), take the newest, and store that message in the `useRef`.

Then, the next time the hook is run, we can check that we have a value stored in the `useRef`, we can filter `privateChatMessages` to exclude anything older than that stored message.

If there are no newer messages, we can probably bail out at this point.

If there are newer messages, we want to:

- sort them by `ts_utc` (see `chatSort` in `src/utils/chat.ts`), get the newest timestamp, and then store that in our `useRef` (updating the `useRef` might be the last thing we do)
- find at least one value ([`.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)) that matches our rules for 'should play a sound'
  - eg. message isn't read, message isn't sent by us, user isn't in chat with the person the message is from
  - If there is a matching value, play the sound
  
---

This should probably be in a `useEffect` or similar, but this is a rough snippet of how it will probably end up looking:

```typescript
const shouldPlaySound = privateChatMessages.some((message) => !message.isRead && ![userId, currentChatUserId].includes(message.from))

if (shouldPlaySound) {
  play()
}
```

src/hooks/useNotificationSound.ts
0xdevalias: [FYI] My suggested changes in https://github.com/sparkletown/sparkle/pull/1262#discussion_r604633362 likely override my comments/suggestions made above.

src/hooks/useNotificationSound.ts
0xdevalias: [FYI] My suggested changes in https://github.com/sparkletown/sparkle/pull/1262#discussion_r604633362 likely override my comments/suggestions made above.

src/hooks/useNotificationSound.ts
0xdevalias: [FYI] My suggested changes in https://github.com/sparkletown/sparkle/pull/1262#discussion_r604633362 likely override my comments/suggestions made above.

0xdevalias:
@igachmovmelon Please make sure you mark review comment threads that have been actioned as resolved ('resolve conversation'), so it's easier to see what has been done vs what is still outstanding.

src/hooks/privateChats.ts
```diff
@@ -215,3 +225,61 @@ export const useRecipientChat = (recipientId: string) => {
...
+
+  const { chatSettings } = useChatSidebarControls();
+
+  const [play] = useSound(newMessageSound, {
+    // `interrupt` ensures that if the sound starts again before it's
+    // ended, it will truncate it. Otherwise, the sound can overlap.
+    interrupt: true,
+  });
+
+  const privateChateMessagesSorted = useMemo(
```
0xdevalias, line 251:
```suggestion
  const privateChatMessagesSorted = useMemo(
```

src/hooks/privateChats.ts
```diff
@@ -215,3 +225,61 @@ export const useRecipientChat = (recipientId: string) => {
...
+    () => Object.values(privateChatMessages).sort(chatSort),
+    [privateChatMessages]
+  );
+
+  const prevPrivateChatMessages = usePrevious(privateChateMessagesSorted[0]);
```
0xdevalias, line 256: Since you're only storing a single message here + better semantic naming:

```suggestion
  const previousNewestMessage = usePrevious(privateChateMessagesSorted[0]);
```

0xdevalias: After sorting with `chatSort`, is `privateChateMessagesSorted[0]` the most recent message, or the oldest message?

src/hooks/privateChats.ts
```diff
@@ -215,3 +225,61 @@ export const useRecipientChat = (recipientId: string) => {
...
+    if (shouldPlaySound) {
+      play();
+    }
+  }, [
+    privateChateMessagesSorted,
+    currentChatUserId,
+    play,
+    prevPrivateChatMessages,
+    userId,
+  ]);
```
0xdevalias, line 284: Why is this using `useMemo` here? And what are we returning (I don't think anything)?

In my last review comments describing this solution, I talked about using `useEffect` here? (https://github.com/sparkletown/sparkle/pull/1262#discussion_r604633362) Why didn't we do that?

I expect that's what we should be doing here + probably removing the `return`, as I don't think we need it do we?

src/hooks/privateChats.ts
```diff
@@ -215,3 +225,61 @@ export const useRecipientChat = (recipientId: string) => {
...
+    const shouldPlaySound = privateChateMessagesSorted
+      .filter((message) => {
+        return (
+          message?.ts_utc.seconds > prevPrivateChatMessages?.ts_utc.seconds
+        );
+      })
+      .some(
+        (message) =>
+          !message.isRead && ![userId, currentChatUserId].includes(message.from)
+      );
```
0xdevalias, line 273: I think we can probably combine the `.filter()` and `.some()` together here, since we're not actually storing the filtered results/using them elsewhere.

Something like:

```suggestion
      .some(
        (message) => {
          const isNewMessage = message?.ts_utc.seconds > previousNewestMessage?.ts_utc.seconds;
          const notRead = !message.isRead;
          const notFromMeOrCurrentChat = ![userId, currentChatUserId].includes(message.from);
          
          return isNewMessage && notRead && notFromMeOrCurrentChat;
      });
```

src/hooks/privateChats.ts
0xdevalias: I believe we only ever actually end up using the timestamp from this message, right? If so, we probably don't even need to save the whole message. We can probably just save the timestamp itself.

Something like:

```suggestion
  const lastMessageTimestampSec = usePrevious(privateChateMessagesSorted[0]?.ts_utc.seconds);
```

0xdevalias (changes requested):
Comments/Changes/Etc

src/hooks/privateChats.ts
```diff
@@ -248,38 +241,39 @@ export const usePrivateChatNotification = (
     interrupt: true,
   });
 
-  const privateChateMessagesSorted = useMemo(
+  const privateChatMessagesSorted = useMemo(
     () => Object.values(privateChatMessages).sort(chatSort),
     [privateChatMessages]
   );
```
0xdevalias: Why do we need `Object.values` here? Isn't `privateChatMessages` already an array? Would this work?

```suggestion
    () => privateChatMessages.sort(chatSort),
    [privateChatMessages]
  );
```

0xdevalias (changes requested):
[CRO] Quick/high level pass, a couple of tiny tweaks, but aside from that didn't see anything big/blocking.

I haven't had a chance to click through/test this, so won't approve just now, but once those changes are done, I think this would be ready to go.

src/hooks/privateChats.ts
igachmovmelon: If we don't use `Object.values` it gives an error `Cannot assign to read only property '0' of object '[object Array]' in typescript`, but we can use `() => [...privateChatMessages].sort(chatSort),` If you think is better,

0xdevalias: I think the spread syntax you suggested is more obvious in intent, that we basically want to clone the array before sorting it.

Though that raises another question for me, do we need to sort it at all? Are they already sorted correctly when they are queried from firestore?

src/hooks/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from "react";
+
+export const usePrevious = <T>(value: T) => {
+  const ref = useRef<T>(value);
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
```
0xdevalias, line 9: We can just use `usePrevious` from `react-hooks` now that we have that dependency:

- https://github.com/streamich/react-use/blob/master/docs/usePrevious.md

0xdevalias (changes requested):

PR #1263, opened 2021-03-15
fix the ratio of the image in .profile-pic

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/549

Added property to the css `object-fit: cover` so that the image can be resized correctly.


src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -9,6 +9,10 @@
     height: 100%;
     width: 100%;
     border-radius: 50%;
+    object-fit: cover;
+    background-position: center;
+    background-size: cover;
+    background-repeat: no-repeat;
```
0xdevalias, line 15: It's probably also worth adding a background color here as a fallback in case there are problems loading the image.

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -21,11 +21,12 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
...
   return (
     <div className={containerStyles}>
-      <img
+      <div
         onClick={onClick}
         className="user-avatar__image"
-        src={avatarSrc}
-        alt="user avatar"
+        style={{
+          backgroundImage: `url(${avatarSrc})`,
```
0xdevalias, line 28: Since `avatarSrc` can be potentially `undefined` here, we probably want to do something like the following (so it doesn't ender up rendering the styles as `url(undefined)`:

```suggestion
          backgroundImage: avatarSrc ? `url(${avatarSrc})` : undefined,
```

0xdevalias (changes requested):
[CRO] A couple of tiny tweaks.

It would also be good if you could show some screenshots of what this looks like.

igachmovmelon:
There is some problem with using `backgroundImage`, sometimes the url is there but the image is not visible, so I returned it to the original way with the `` element. For the scss I am using `object-fit: cover;` because it is covering the resizing of the image you can check it more in here [https://www.w3schools.com/css/css3_object-fit.asp](url)

0xdevalias:
> The `background-position: center` from the Avatar needed to be removed and remain only the `background-size: cover` That way the image is scaled correctly.
> 
> [musing] @igachmovmelon @mike-lvov @mcflyok Would using `contain` work here? Or do we want to keep it as `cover`?
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1264#issuecomment-820861169_

igachmovmelon:
I added ` object-position: top` for better view when the image is vertical. That way looks similar to the UserProfileModal:

image

image

mike-lvov:
@goran-peoski-work , please take over this PR

goran-peoski-work:
> closes sparkletown/internal-sparkle-issues#549
👆
This was not the case.

The profile picture in the profile modal doesn't have `<UserAvatar>` component (subsequently fixed by  since e2f971d5a7c3835b58d43c66acb789ecfdd65847 ), but a plain `` under `div.profile-pic`  to which the fix was applied.

alexey-petrov (approved):
LGTM

PR #1264, opened 2021-03-15
fix avatar alignment in auditorium seat

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/560

Added `object-fit` and `overflow` for scaling the image in the auditorium


0xdevalias:
@IvanBuljovcic Can you please add a screenshot showing how this looks now. Ideally a before and after would be great (and make it much easier to assess that it fixes what was needed here in review)

0xdevalias (changes requested):
Marking as 'changes requested' based on https://github.com/sparkletown/sparkle/pull/1264#issuecomment-811767398

igachmovmelon:
The problem wasn't fixed with the `object-fit` property this only works with `` elements. The `background-position: center` from the Avatar needed to be removed and remain only the `background-size: cover` That way the image is scaled correctly.

### Before
image

### After
image

0xdevalias:
> The `background-position: center` from the Avatar needed to be removed and remain only the `background-size: cover` That way the image is scaled correctly.

@igachmovmelon @mike-lvov @mcflyok Would using `contain` work here? Or do we want to keep it as `cover`?

igachmovmelon:
@0xdevalias if we use `contain` this is the effect:

image

mike-lvov:
@goran-peoski-work , please, take over this PR

goran-peoski-work:
This PR is no longer relevant. As per e2f971d5a7c3835b58d43c66acb789ecfdd65847 the `` element was replaced with `<UserAvatar>` which fixes the alignment issue by itself.

mike-lvov:
closing as per https://github.com/sparkletown/sparkle/pull/1264#issuecomment-889779386

PR #1267, opened 2021-03-15
Increases seat size

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/564

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -3,8 +3,8 @@
 $light: #ffffff;
 
 // If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
-$seat-size: 4vh;
-$seat-size-min: 25px;
+$seat-size: max(4vh, 3vw);
+$seat-size-min: 30px;
 
 $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
 
@@ -34,7 +34,6 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
     pointer-events: none;
 
     .video-container {
-      margin-top: calc(1 * min(var(--seat-size), var(--seat-size-min)));
       padding: calc(1.5 * min(var(--seat-size), var(--seat-size-min)));
 
       display: flex;
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -4,7 +4,7 @@ $light: #ffffff;
 
 // If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
 $seat-size: max(4vh, 3vw);
-$seat-size-min: 30px;
+$seat-size-min: 25px;
 
 $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
 
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -46,7 +46,7 @@ interface ChatOutDataType {
 const SEAT_SIZE = "var(--seat-size)";
 const SEAT_SIZE_MIN = "var(--seat-size-min)";
 
-const VIDEO_MIN_WIDTH_IN_SEATS = 10;
+const VIDEO_MIN_WIDTH_IN_SEATS = 8;
 // We should keep the 16/9 ratio
 const VIDEO_MIN_HEIGHT_IN_SEATS = VIDEO_MIN_WIDTH_IN_SEATS * (9 / 16);
 
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -388,6 +388,7 @@ export const Audience: React.FunctionComponent = () => {
               placeholder="Shout out to the crowd"
               ref={register({ required: true })}
               disabled={isShoutSent}
+              autoComplete="off"
             />
             <input
               className={`shout-button ${isShoutSent ? "btn-success" : ""} `}
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -46,7 +46,7 @@ interface ChatOutDataType {
 const SEAT_SIZE = "var(--seat-size)";
 const SEAT_SIZE_MIN = "var(--seat-size-min)";
 
-const VIDEO_MIN_WIDTH_IN_SEATS = 8;
+const VIDEO_MIN_WIDTH_IN_SEATS = 10;
 // We should keep the 16/9 ratio
 const VIDEO_MIN_HEIGHT_IN_SEATS = VIDEO_MIN_WIDTH_IN_SEATS * (9 / 16);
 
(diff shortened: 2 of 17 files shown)


src/components/templates/Audience/Audience.scss
```diff
@@ -34,7 +34,6 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
     pointer-events: none;
 
     .video-container {
-      margin-top: calc(1 * min(var(--seat-size), var(--seat-size-min)));
```
mike-lvov, line 37: What was the point of adding the top margin?
It looks not broken without it(probably one of the reasons the video overflowed the seats?)

0xdevalias: Off the top of my head, I feel like it was sitting 'too high' without it, but I don't think I could ever figure out why that was the case.

It could be that some changes we made after I originally added it 'fixed the need for it' too. If it looks fine without it, then it's probably fine to remove it.

It always felt kind of hacky to me.. particularly after we switched to using the flexbox to center it (off the top of my head I was originally positioning it by exact column sizes/etc, so maybe this was leftover from that)

src/components/templates/Audience/Audience.tsx
```diff
@@ -46,7 +46,7 @@ interface ChatOutDataType {
 const SEAT_SIZE = "var(--seat-size)";
 const SEAT_SIZE_MIN = "var(--seat-size-min)";
 
-const VIDEO_MIN_WIDTH_IN_SEATS = 8;
+const VIDEO_MIN_WIDTH_IN_SEATS = 10;
```
0xdevalias, line 49: Why did we decide to make this `10`? I didn't see any discussion/etc related to it on the original issue, nor explanation here explaining it.

Have we considered what the flow on effects of this change are going to be, and whether we will need to change any of our other defaults/min/max values/etc? (including in admin v1/v2, validations, etc)

0xdevalias: Before this change:
image

After this change:
image

src/components/templates/Audience/Audience.scss
0xdevalias: Off the top of my head, I feel like it was sitting 'too high' without it, but I don't think I could ever figure out why that was the case.

It could be that some changes we made after I originally added it 'fixed the need for it' too. If it looks fine without it, then it's probably fine to remove it.

It always felt kind of hacky to me.. particularly after we switched to using the flexbox to center it (off the top of my head I was originally positioning it by exact column sizes/etc, so maybe this was leftover from that)

0xdevalias (commented):
Questions/comments.

src/components/templates/Audience/Audience.scss
```diff
@@ -3,8 +3,8 @@
 $light: #ffffff;
 
 // If you change this, make sure to also change it in Audience.tsx's SEAT_SIZE
-$seat-size: 4vh;
-$seat-size-min: 25px;
+$seat-size: max(4vh, 3vw);
+$seat-size-min: 30px;
```
0xdevalias, line 7: Also, why did we change `$seat-min-size` to `30px`? I didn't see any discussions/etc anywhere about that.

src/components/templates/Audience/Audience.tsx
mike-lvov: The video being tiny was mentioned by @jess-sparkles in the original issue
 `The video is also super tiny and hard to see/ watch. Is there a way to make iFrame in the center larger and the seats fill the screen more?`

src/components/templates/Audience/Audience.scss
mike-lvov: I don't have an educated answer to this one. I guess my intention was to make seats `larger` so that they would take more space on the small screens.

jess-sparkles:
I feel like making the video larger would automatically make the seats fill up more of the screen because they would be pushed outwards away from the video

src/components/templates/Audience/Audience.tsx
0xdevalias: The iframe was tiny because the whole ‘seats block’ was only using a small chunk of the screen. So the changes to make it bigger/use the whole of the screen (by increasing the seat size) already make the iframe bigger.

IMO the outer rows of the seats look too ‘squished’ now, it felt way more balanced as was.

0xdevalias:
> I feel like making the video larger would automatically make the seats fill up more of the screen because they would be pushed outwards away from the video

@jess-sparkles Thats not how it works. From a ‘square of seats’ a central area of them is ‘carved out’ for the video block.

By making the seats use a percentage of the larger of the browser width/height for their size, we get a more balanced ‘use more of the screen’ feel.

I believe I showed a screenshot of what this looks like (without the additional ‘widening of the central video’ in the original issue thread, which from memory you were fine with.

jess-sparkles:
Got it, as long as the result is that the video is larger, idc how it gets done ;)

src/components/templates/Audience/Audience.scss
0xdevalias: I don't think that's a thing we need to with this. It's only going to apply when `4vh` or `3vw` is smaller than `30px`, which isn't going to be the case on most devices, except for probably mobiles, where I expect we won't want the `30px` as we will need them to be smaller.

---

Not sure why exactly, but this change also seems to make the video window smaller, at least on the current production code:

`--seat-size-min: 25px`:
image

`--seat-size-min: 30px`:
image

---

Personally I would just stick to the original plan/investigation, and revert this aspect of this PR. I don't think it adds any benefit, and it seems to detriment things.

src/components/templates/Audience/Audience.scss
0xdevalias: I locally changed `VIDEO_MIN_WIDTH_IN_SEATS` back to `8` and `--seat-size-min` back to `25px` for these screenshots.

With reactions enabled:
image

With reactions disabled:
image

I even tried changing it back to a vimeo embed like the old venue, but it still seems fine:
image
image

---

So basically.. no clue why it was there originally/etc, but it seems like it's fine to remove it now.

0xdevalias (changes requested):
Would be good to restore these to how they originally were, as I don't think they add any benefit, and they seem to detract from the UX.

src/components/templates/Audience/Audience.scss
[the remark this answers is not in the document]
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1267/commits/3ce92bcfe8cf07d9ad37c72f1fde352e53555bd9

mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1267/commits/3ce92bcfe8cf07d9ad37c72f1fde352e53555bd9

src/components/templates/Audience/Audience.tsx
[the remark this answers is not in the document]
mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1267/commits/3ce92bcfe8cf07d9ad37c72f1fde352e53555bd9

mike-lvov: fixed in https://github.com/sparkletown/sparkle/pull/1267/commits/3ce92bcfe8cf07d9ad37c72f1fde352e53555bd9

0xdevalias (approved):
Looks good, LGTM!

PR #1268, opened 2021-03-15
Fix reactions and shoutouts going behind avatars (or emojis)

malaikaaiyar:
closes sparkletown/internal-sparkle-issues#551

Before (all have z-index 10 or auto):
Screen Shot 2021-03-15 at 9 21 57 PM

After:
Screen Shot 2021-03-15 at 9 22 10 PM


0xdevalias (approved):
[CRO] Screenshots/code look good. LGTM!

---

~While the changes here look fine on their own, having more and more `z-index`'s 'locked away' within `styled-components` like this is almost certainly going to lead to future bugs/headaches.~

~Since `Container` seems to just be static styling~ (**Edit:** actually, it refers to the other styles such as `Avatar`/etc, which then have dynamic dependencies), ~can I please get you to convert it across to scss, and then use our `z()` layer helper for the 2 z-indexes (adding the new values to the 'map' in `constants.scss`.~

~I believe the `${Avatar}`/etc stuff is just referencing the css class for avatars/etc, which should be easy enough to look up.~

---

~[NB] If you felt up for it, it wouldn't take a huge amount of extra work to move the `Reaction` styles to scss as well (using the BEM naming patterns (perhaps something like `reaction--right` and `reaction--left`, and our `classNames` helper for choosing when to apply the `rectionPosition` modifiers.~

**Edit:** While these changes would be good to make sooner rather than later, they may make this issue somewhat more complex to resolve, so maybe they should be OOS here.

malaikaaiyar:
In that case, I'll merge it as is. Should I make an issue for these changes?

0xdevalias:
> In that case, I'll merge it as is. Should I make an issue for these changes?

@malaikaaiyar Yeah, that would be good. Thanks.

PR #1269, opened 2021-03-16
add useCustomSound + configurable custom sound effects on Room entry

0xdevalias:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/388

> - When users click on a room, a sound effect will play.
> - These sound effects should play when you click on and enter rooms.

---

Sounds can be a simple audio file, or a 'sound sprite':

- https://github.com/goldfire/howler.js#define-and-play-a-sound-sprite
- https://github.com/pixijs/pixi-sound/wiki/Generating-AudioSprites-with-Audacity

Audio format recommendations:

- https://github.com/goldfire/howler.js/#format-recommendations

We use `react-sound` (which uses Howler) to play audio effects from a hook:

- https://github.com/joshwcomeau/use-sound
- https://github.com/goldfire/howler.js

While you could choose to store/load the audio assets from anywhere, I am using firebase storage. Which means we also needed to ensure the correct CORS headers were set for the files:

- https://cloud.google.com/storage/docs/configuring-cors

We store these `SoundConfig`s within the `sounds` collection in firestore, with each referencing a `url`, and optionally defining the `sprites` contained within it.

## Testing

**Edit:** ~Note, none of these test rooms are available anymore due to https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1615899809010700~ I have re-created them all again now.

- https://co-reality-staging--preview-pr-1269-odmj9e15.web.app/in/devaliashacksville
  - `devaliasconversationspace` is configured with a sound sprite
  - `devaliasburnbarrel` is configured with a different sound sprite
  - `devaliasartpiece` is configured with a non-existent sound sprite
  - `devaliasjazzbar` is configured with a non-existent `soundId`
  - none of the other rooms have `enterSound`s defined on them (at time of writing)
- Note: If you see an error such as `fetchSoundConfigs FirebaseError: Missing or insufficient permissions` it might be because someone has deployed to `staging` and overwritten the rule I deployed
  - The rules code is in this PR, and you can re-deploy it to `staging` with: `npx firebase --project co-reality-staging deploy --only firestore:rules`
- Note: I think you may get a CORS error from running it on this preview URL at the moment when trying to load the sounds, see https://github.com/sparkletown/internal-sparkle-issues/issues/388#issuecomment-799875109
  - The error looks like: `Access to XMLHttpRequest at 'XXX' from origin 'https://co-reality-staging--preview-pr-1269-odmj9e15.web.app' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header has a value 'http://localhost:3000' that is not equal to the supplied origin.`
  - As a workaround, you could clone this branch and run it locally.

## TODO

- [x] Add `enterSound` (`SoundConfigReference`) to `Room` type
- [x] Wire up `useCustomSound` in `src/components/templates/PartyMap/components/Map/MapRoom.tsx`
- [x] Wire up `useCustomSound` in `src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx`
- [x] Fix TODOs in https://github.com/sparkletown/sparkle/pull/1269#pullrequestreview-612863433

## OOS

- Admin tooling
  - for uploading/managing sound configurations (this can be done manually within firebase for now)
  - for adding sounds to rooms (this can be done manually within firestore for now)
  - etc


src/api/sounds.ts
```diff
@@ -0,0 +1,38 @@
...
+      .collection("sounds")
+      .withConverter(soundConfigConverter)
+      .get();
+
+    return soundConfigsSnapshot.docs
+      .map((docSnapshot) => withId(docSnapshot.data(), docSnapshot.id))
+      .reduce(itemsToObjectByIdReducer, {});
+  } catch (err) {
+    // TODO: properly log this with bugsnag or similar?
+    console.error("fetchSoundConfigs", err);
```
0xdevalias, line 21: [TODO] Log error with Bugsnag

src/api/sounds.ts
```diff
@@ -0,0 +1,38 @@
...
+export const soundConfigConverter: firebase.firestore.FirestoreDataConverter<SoundConfig> = {
+  toFirestore: (soundConfig: SoundConfig): firebase.firestore.DocumentData => {
+    return SoundConfigSchema.validateSync(soundConfig);
+  },
+  fromFirestore: (
+    snapshot: firebase.firestore.QueryDocumentSnapshot
+  ): SoundConfig => {
+    return SoundConfigSchema.validateSync(snapshot.data());
+  },
+};
```
0xdevalias, line 38: [note] An example of using `FirestoreDataConverter` + proactively validating data when it is retrieved from firestore to ensure that it matches what we expect (aka: no one can break it by hacking the database without us knowing, thus avoiding weird/hard to debug/find bugs)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -53,6 +54,11 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
 
   const { enterRoom, recentRoomUsers } = useRoom({ room, venueName });
 
+  const [enterRoomWithSound] = useCustomSound(room.enterSound, {
+    interrupt: true,
+    onend: enterRoom,
+  });
```
0xdevalias, line 60: [TODO] Fix bug where we can't enter the room anymore if there is no `enterSound` defined for this room.

---

[note] We use `onend` here so that we wait until the sound has finished playing before entering the room.

src/hooks/sounds.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  // Fetch all of our loaded sound configs
+  const { soundConfigs } = useCustomSoundsContext();
+
+  // Try to access just the one we want here
+  const soundConfig = soundConfigs[soundId];
+
+  // TODO: throw some kind of error/tracking/etc when we don't find a valid soundConfig
+  // TODO: throw some kind of error/tracking/etc when we don't find soundRef.spriteName in soundConfig
+  // TODO: throw some kind of error/tracking/etc when we hasSprites but not wantsSprites, or when we wantsSprites but not hasSprites
```
0xdevalias, line 127: [TODO] Log errors in Bugsnag or similar.

```
// TODO: throw some kind of error/tracking/etc when we don't find a valid soundConfig
// TODO: throw some kind of error/tracking/etc when we don't find soundRef.spriteName in soundConfig
// TODO: throw some kind of error/tracking/etc when we hasSprites but not wantsSprites, or when we wantsSprites but not hasSprites
```

src/api/sounds.ts
```diff
@@ -0,0 +1,38 @@
...
+import { withId } from "utils/id";
+import { itemsToObjectByIdReducer } from "utils/reducers";
+
+export const fetchSoundConfigs = async (): Promise<SoundConfigMap> => {
+  try {
+    const soundConfigsSnapshot = await firebase
+      .firestore()
+      .collection("sounds")
+      .withConverter(soundConfigConverter)
+      .get();
```
0xdevalias, line 14: [OOS, future] Currently this will load everything in the `sounds` collection, which IMO is fine for this MVP version of the feature.

Once we flesh it out/polish it a bit more (eg. adding admin tooling, etc) it would also make sense to change this part too, perhaps to load sounds from the 'world' object (once it exists), or similar?

src/types/sounds.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @see SoundConfig
+ */
+export const SoundConfigSchema: Yup.ObjectSchema<SoundConfig> = Yup.object()
+  .shape<SoundConfig>({
+    url: Yup.string().required(),
+    sprites: SpriteMapSchema,
+  })
+  .noUnknown()
+  .required();
```
0xdevalias, line 49: [note] The schema used for proactively validating data when it is retrieved from firestore to ensure that it matches what we expect (aka: no one can break it by hacking the database without us knowing, thus avoiding weird/hard to debug/find bugs)

0xdevalias (commented):
Notes and TODOs

0xdevalias:
@sparkletown/engineering While there are still some final polish bits and pieces to do (noted in https://github.com/sparkletown/sparkle/pull/1269#pullrequestreview-612863433), this PR should be mostly ready to go/for review.

Please refer to the 'Testing' section in the initial description for further details.

src/api/sounds.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+};
+
+/**
+ * Convert SoundConfig objects between the app/firestore formats, including validation.
+ */
+export const soundConfigConverter: firebase.firestore.FirestoreDataConverter<SoundConfig> = {
+  toFirestore: (soundConfig: SoundConfig): firebase.firestore.DocumentData => {
+    return SoundConfigSchema.validateSync(soundConfig);
+  },
```
mike-lvov, line 32: [neat] 
```suggestion
  toFirestore: (soundConfig: SoundConfig): firebase.firestore.DocumentData => SoundConfigSchema.validateSync(soundConfig);,
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: `onend => onEnd`?

src/hooks/sounds.tsx
```diff
@@ -0,0 +1,164 @@
...
+    }),
+    [soundConfigs]
+  );
+
+  return (
+    <CustomSoundsContext.Provider value={providerData}>
+      {children}
+    </CustomSoundsContext.Provider>
+  );
+};
```
mike-lvov, line 68: Not against it, but just want to understand. What is the benefit of doing it this way, instead of our `default` useFirestoreConnect?. Even though I do understand that we only need to fetch it one time, it seems to be very verbose, compared to useFirestoreConnect

src/hooks/sounds.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+      // This case won't actually allow us to play any sounds, and is only here to
+      // work around the inability to conditionally render hooks
+      case "undefined":
+        return { soundId: "", spriteName: undefined };
+
+      default:
+        return soundRef;
+    }
+  })();
```
mike-lvov, line 117: It took me a while to actually understand what is going on here. Is it possible to reduce the complexity. Maybe just use a ternary operator, if you don't plan on expanding the data types of `soundRef`.

src/hooks/sounds.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const hasSoundConfig = soundConfig !== undefined;
+  const hasSprites = soundConfig?.sprites !== undefined;
+  const wantsSprites = spriteName !== undefined;
+
+  // We must both haveSprites && wantSprites or not have either for the config to be valid
+  const soundUrl =
+    (hasSprites && wantsSprites) || (!hasSprites && !wantsSprites)
+      ? soundConfig?.url ?? USE_SOUND_DISABLED_URL
+      : USE_SOUND_DISABLED_URL;
```
mike-lvov, line 137: maybe we should drop a warning/error at this point?

mike-lvov (commented):
It was a bit hard to go through the code. It's a bit different from what we have in the codebase(though a bit similar to how the custom loaders were implemented). Does it need to be different though?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Nope. For whatever reason that’s what howler chose to use as their API here.

src/hooks/sounds.tsx
0xdevalias: There are 3 distinct cases that need to be handled here, how would you see that being simplified as a ternary?

This is a pretty standard usage of a switch/case statement. The IIFE (immediately invoked function expression) is also fairly standard (though perhaps less commonly used these days) An alternative to that would be to pull it out into a separate function, but that felt overkill given its’s only ever useful/needed in this one place.

0xdevalias: What do you mean by “don’t plan on expanding the data types of sound ref”?

src/hooks/sounds.tsx
0xdevalias: See the TODO’s noted in https://github.com/sparkletown/sparkle/pull/1269#discussion_r594874508

src/hooks/sounds.tsx
0xdevalias: We want to move away from useFirestoreConnect as it’s more or less abandoned.

This context pattern means that we are actually literally only calling the code once, and providing that where it’s needed. This is the singleton pattern I described that we should have used for a lot of our current ‘needs to be called once’ hooks. I would say a lot of our current ‘use once’ hooks only sort of work by crude side effect rather than good patterns (because useFirestoreConnect /the firestore SDK attempts to de-duper/reuse configs/caches/etc so as not to make extra unnneded connections; but we’ve seen bugs in this before)

Context’s are a very common, standard, and recommended way of building clean react code (despite the opinions of one of our former engineers who ripped some of the previous ones out)

The ‘verbosity’ of the pattern is a few extra lines of boilerplate during setup, and is on par with the equivalent requirements of a redux setup (eg initialState, etc)

The other alternative would be to move it all to redux, which would have just as much boilerplate (if not more), and would still require us to ‘connect it’ at some point, which would end up looking very similar. I expect at the end of the day the difference between implementation would be that the data is stored in redux and we use a selector to retrieve it, vs us using a useState and already having the sata ‘selected’.

0xdevalias: Also, if this is a pattern we start to use more, I can absolutely abstract this boilerplate away into a nicer 'API' using helpers/etc; but given the urgency of the timeline around both the loaders and the sounds features, I decided it wasn't worth 'refining' to that level at this stage, particularly as it would likely only save a few lines of boilerplate anyway.

src/hooks/sounds.tsx
0xdevalias: [FYI] As an aside, the pattern I've used here is sort of a hack/workaround that allows us to do things in a similar way to the ECMAScript Pattern Matching proposal (though not as powerful as it), that is currently in 'Stage 1' of the tc39 system:

> Stage 1 proposals represent problems that the committee is interested in spending time exploring solutions to.

- https://github.com/tc39/proposal-pattern-matching

There is also a PR for a babel plugin (handles our transpilation) that implements this, but it seems to still be outstanding, pending expected changes to the design of the spec:

- https://github.com/babel/babel/pull/9318

0xdevalias:
> It was a bit hard to go through the code. It's a bit different from what we have in the codebase (though a bit similar to how the custom loaders were implemented). Does it need to be different though?

@mike-lvov When you say "a bit different from what we have in the codebase", can you please expand on that?

Inferring from your more specific comments (eg. https://github.com/sparkletown/sparkle/pull/1269#discussion_r595428447) I'm guessing you're referring to using the React context pattern here? If so, does my response in https://github.com/sparkletown/sparkle/pull/1269#discussion_r595532378 satisfy your concerns? And if not (or if they relate to something different), can you please expand on the specifics that concern you.

src/api/sounds.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1269/commits/0d496de6156ad43094c293764036304f458d8477

src/hooks/sounds.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1269/commits/0d496de6156ad43094c293764036304f458d8477

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1269/commits/ce530ff4e0e30ad24826b7e12174790e71307670

src/hooks/sounds.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1269/commits/0d496de6156ad43094c293764036304f458d8477

0xdevalias:
@mike-lvov This should be good for review testing again. See notes in https://github.com/sparkletown/sparkle/pull/1269#issue-593553961

mike-lvov (approved):
LGTM. Tested in the preview link

PR #1240, opened 2021-03-05
[ENV/KOTR] custom fonts, colors and loaders implementation

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/386
closes https://github.com/sparkletown/internal-sparkle-issues/issues/389

PR #1272, opened 2021-03-16
Fix RoomModal image styling

malaikaaiyar:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/437

Before:
Screen Shot 2021-03-16 at 10 18 19 PM

After:
Screen Shot 2021-03-16 at 10 18 32 PM


0xdevalias:
> RE: https://github.com/sparkletown/internal-sparkle-issues/issues/437#issuecomment-800650229
> 
> > If the image isn't square it leaves part of the square black
> > 
> > this is with contain
> 
> @malaikaaiyar Did you just set  for that screenshot, not any other of the related properties (see below)?
> 
> Contrasting against your 'before' screenshot in https://github.com/sparkletown/sparkle/pull/1272#issue-594273266 it looks like it isn't getting cut off at all, the full image is being shown.
> 
>  will 'letterbox' the image so that it fits though, and it looks like this 'letterboxed' version isn't centered within the constrained image square. I believe the 'letterbox' part will use the colour from , which is presumably currently set to black.
> 
> For the recent 'loading screen' changes I made ([ref](https://github.com/sparkletown/sparkle/pull/1240/files#diff-61299cd4c35db28ff9e3f3bd94203ce824abbcdc4b377729c49490334c4b3638R67-R70)) I used the following:
> 
> ```scss
> // Fallback to showing the main container's background in case we don't load an image
> background-color: transparent;
> 
> background-image: $loading-page-background-image;
> background-position: center;
> background-repeat: no-repeat;
> background-size: contain;
> ```
> 
> It might be worth setting these or similar on it as well?
> 
> **Further Reading:**
> 
> - https://developer.mozilla.org/en-US/docs/Web/CSS/background-position
>   - > The  CSS property sets the initial position for each background image. The position is relative to the position layer set by .
> - https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat
>   - > The  CSS property sets how background images are repeated. A background image can be repeated along the horizontal and vertical axes, or not repeated at all.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/437#issuecomment-800768749_

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -87,6 +87,10 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
 
   const hasRoomEvents = renderedRoomEvents?.length > 0;
 
+  const iconStyle = {
+    backgroundImage: `url(${room.image_url})`,
+  };
```
0xdevalias, line 92: [nit]

```suggestion
  const iconStyles = {
    backgroundImage: `url(${room.image_url})`,
  };
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -24,18 +24,11 @@ $spacing: 20px;
...
 
-    > img {
-      width: 60%;
-      height: auto;
-
-      border-radius: $border-radius;
-    }
-
     > * {
-      flex: 1;
```
0xdevalias, line 38: What was the intent behind removing this here?

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -45,6 +38,16 @@ $spacing: 20px;
...
+  &__icon {
+    flex-shrink: 0;
+    height: 180px;
+    width: 180px;
+    border-radius: 18px;
+    background-color: #000;
+    background-size: cover;
+    background-repeat: no-repeat;
+  }
+
```
0xdevalias, line 50: This should probably be using variables such as `$border-radius` (as per the styles on `img` removed in this PR), `$black` (or whatever colour we want to be used as the fallback/letterbox), etc.

Also probably want to include `background-position: center;` here as per https://github.com/sparkletown/sparkle/pull/1272#issuecomment-800768978 (and it would be good to see what a comparison screenshot looks like with `background-size: contain;` once this is added too)

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -24,18 +24,11 @@ $spacing: 20px;
   &__main {
     display: flex;
     flex-direction: row;
+    flex: auto 1;
```
0xdevalias, line 27: What was the intent behind adding this here?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -97,7 +101,7 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
 
       <div className="room-modal__main">
         {room.image_url ? (
-          <img src={room.image_url} alt={room.title} />
+          <div className="room-modal__icon" style={iconStyle}></div>
         ) : (
           <span>{room.title}</span>
         )}
```
0xdevalias, line 107: I think the `span` here will have styling issues/look strange if it ever gets rendered. You could check this by hacking the code to render that code path (eg. making the check `!room.image_url` temporarily) to see what it looks like. Might be worth including a screenshot of that here as well to help with the review.

I know it's existing code, but it's probably irrelevant to render the `room.title` here again since it's already being rendered just above.

We may just want to always render the `room-modal__icon` `div`, and have it fall back to the `background-color` when it has an invalid URL (which is how the CSS works by default)

0xdevalias (changes requested):

See also my comments/requests in https://github.com/sparkletown/sparkle/pull/1272#issuecomment-800768978

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
malaikaaiyar: Hm, I thought it did something but it seems like it doesn't- I'll delete it

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
malaikaaiyar: Looks like this with the span
Screen Shot 2021-03-25 at 6 20 36 PM

Like this with no image but just the div (better imo)
Screen Shot 2021-03-25 at 6 21 55 PM

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
malaikaaiyar: It looks very strange with `background-size: contain`
Screen Shot 2021-03-25 at 6 23 18 PM

But I agree with the `background-position: center;`, so I'll add that

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
malaikaaiyar: Not sure! It actually looks a bit better with it, so I'll put it back

malaikaaiyar:
So the modal looks like this with these styles:

```
background-color: transparent;
background-size: cover;
background-repeat: no-repeat;
background-position: center;
```
Screen Shot 2021-03-25 at 8 55 26 PM

---

With the same styles except `background-size: contain`, it looks like this

for long images:
Screen Shot 2021-03-25 at 9 00 18 PM

for wide images:
Screen Shot 2021-03-25 at 9 00 13 PM

for square images:
Screen Shot 2021-03-25 at 9 00 25 PM

---

We have two options now:
- use `cover` and keep the images square
- use `contain` and get rid of the `border-radius` so that no images have rounded corners

0xdevalias:
> and get rid of the `border-radius` so that no images have rounded corners

@malaikaaiyar why would we need to do this? Just because it wouldn’t apply in the edge cases of an extremely long or wide image doesn’t seem like a reason we would remove it for all cases. (cc // @sparkletown/design)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Agreed. Let’s just keep the ‘div’ for both cases and drop the ‘spam’ entirely.

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
0xdevalias: The Center with contain should centre the image so that the ‘letterboxing’ appears on both sides. In your screenshot above it looks like you may have only had contain but not centre set. We want both at the same time for it.

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
0xdevalias: The intent of this style is to add flex: 1 to all of the immediate child components, which I believe would tell them to space evenly, and grow/shrink together, etc.

Then I believe we explicitly override the image style elsewhere so it doesn’t actually grow/shrink.

mcflyok:
@malaikaaiyar @0xdevalias I’d go for contain+center to show the full image contained inside the same size square, with the background color and border radius, as I believe most of our room images will be transparent pngs it would look like this :
modal--room

malaikaaiyar:
> @malaikaaiyar why would we need to do this? Just because it wouldn’t apply in the edge cases of an extremely long or wide image doesn’t seem like a reason we would remove it for all cases.
> 
> Particularly if the background colour style of the ‘image section’ is set to something other than transparent, then even when getting the letterbox effect from ‘contain’, you would still see the rounded corners.

I assumed that we would have a lot of long/wide images, but since that's not really the case, I'll leave it as is.

malaikaaiyar:
They look like this now:
Screen Shot 2021-03-30 at 6 31 38 PM
Screen Shot 2021-03-30 at 6 32 12 PM

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -45,6 +38,17 @@ $spacing: 20px;
     }
   }
 
+  &__icon {
+    flex-shrink: 0;
+    height: 180px;
+    width: 180px;
```
0xdevalias, line 44: Should these be `min-width` and `min-height` or similar?

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -45,6 +38,17 @@ $spacing: 20px;
...
+    border-radius: $border-radius;
+    background-color: transparent;
```
0xdevalias, line 46: I believe @mcflyok's suggestion was to use a different explicit colour for this rather than `transparent`?

- https://github.com/sparkletown/sparkle/pull/1272#issuecomment-809135362
  - > I’d go for contain+center to show the full image contained inside the same size square, **with the background color** and border radius, as I believe most of our room images will be transparent pngs
  
I'm not sure what the best color to use here is (@mcflyok thoughts?), but perhaps `$black` is fine?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -93,6 +93,10 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
 
   const hasRoomEvents = renderedRoomEvents?.length > 0;
 
+  const iconStyles = {
+    backgroundImage: `url(${room.image_url})`,
```
0xdevalias, line 97: Something like this/similar is probably better so we don't render the style as `url(undefined)`:

```suggestion
    backgroundImage: room.image_url ? `url(${room.image_url})` : undefined,
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -102,11 +106,7 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
       )}
 
       <div className="room-modal__main">
-        {room.image_url ? (
-          <img src={room.image_url} alt={room.title} />
-        ) : (
-          <span>{room.title}</span>
-        )}
+        <div className="room-modal__icon" style={iconStyle}></div>
```
0xdevalias, line 109: When there is no 'body' to a `div`/etc, it's better to use the shorthand format rather than having an explicit open/closing tag:

```suggestion
        <div className="room-modal__icon" style={iconStyle} />
```

0xdevalias (changes requested):
A few more changes needed here, then this should be good to go.

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
malaikaaiyar: No, we tried it and it doesn't actually look better

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
malaikaaiyar: his designs use black so I'll switch it to that

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
0xdevalias: I guess since we set the flex area to never change size it probably doesn't matter. My thoughts here was that this will never responsively resize. But you'd probably have to get down to mobile browser size for that to matter at this stage I would guess, and we likely have a ton of other visual/other issues at that point anyway. 👍🏻

0xdevalias:
@malaikaaiyar I'll assume this is ready for re-review based on your `#daily` update, but please remember to click the `re-request review` button on the PR so that I get notified that it is ready/etc.

0xdevalias (approved):
Looks good, LGTM!

PR #1273, opened 2021-03-16
Disable autocomplete suggestions when writing a chat message, shoutout or other input

malaikaaiyar:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/571

> When I send chat messages, I get the 'suggestions' that I would get when filling out a form for the second/third time, but there aren't many cases when I would want to send the same text message multiple times (and it looks ugly)

---

> I believe the fix for this will just be adding  to the text field:
> 
> - https://css-tricks.com/snippets/html/autocomplete-off/
> 
> And if we want to 'scope creep' this issue a little bit, it might be nice to think about/enumerate all of the other places within our codebase where we would also not want these suggestions, and proactively add it to those as well.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/571#issuecomment-797129346_

### DIFF
--- a/src/components/molecules/Chatbox/Chatbox.tsx
+++ b/src/components/molecules/Chatbox/Chatbox.tsx
@@ -72,6 +72,7 @@ export const Chatbox: React.FC<ChatboxProps> = ({
           ref={register({ required: true })}
           name="message"
           placeholder="Write your message..."
+          autoComplete="off"
         ></input>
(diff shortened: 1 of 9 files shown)


0xdevalias (approved):
Looks good, LGTM!

PR #1274, opened 2021-03-17
fix updateVenue function to properly update iframeUrl + simplify implementation

0xdevalias:
When attempting to edit an `embeddable`/`firebarrel` venue, the backend `updateVenue` function wasn't correctly updating the `iframeUrl` value.

This was likely due to the disparate locations that needed to be updated in our backend functions when adding new functionality. I did a minimal cleanup refactor on some of this to reduce the likelihood of this sort of issue occurring in future; + added some extra related comments/debt comments/etc.

As a workaround, it was still possible to update the `iframeUrl` via the `/in/:venueId/admin` endpoint.

- fixes https://github.com/sparkletown/internal-sparkle-issues/issues/580
- fixes https://github.com/sparkletown/internal-sparkle-issues/issues/206

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -553,6 +553,7 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.radioStations = [data.radioStations];
   }
 
+  // @debt this would currently allow any value to be set in this field, not just booleans
   updated.requiresDateOfBirth = data.requiresDateOfBirth || false;
 
   switch (updated.template) {
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -560,6 +560,8 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     case VenueTemplate.performancevenue:
     case VenueTemplate.artpiece:
     case VenueTemplate.audience:
+    case VenueTemplate.embeddable:
+    case VenueTemplate.firebarrel:
       if (data.iframeUrl) {
         updated.iframeUrl = data.iframeUrl;
       }
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -28,7 +28,7 @@ const VenueTemplate = {
 };
 
 const DEFAULT_PRIMARY_COLOR = "#bc271a";
-const VALID_TEMPLATES = [
+const VALID_CREATE_TEMPLATES = [
   VenueTemplate.jazzbar,
   VenueTemplate.friendship,
   VenueTemplate.partymap,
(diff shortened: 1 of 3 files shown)


functions/venue.js
```diff
@@ -120,28 +142,22 @@ const createVenueData = (data, context) => {
...
-    case VenueTemplate.firebarrel:
-      venueData.iframeUrl = data.iframeUrl;
-      break;
+  if (IFRAME_TEMPLATES.includes(data.template)) {
+    venueData.iframeUrl = data.iframeUrl;
+  }
+
+  if (ZOOM_URL_TEMPLATES.includes(data.template)) {
+    venueData.zoomUrl = data.zoomUrl;
+  }
```
mike-lvov, line 151: But what was the problem? venueData.iframeUrl was set in the prev implementation as well

mike-lvov (approved):
I couldn't spot the root of the issue

functions/venue.js
0xdevalias: Here in `exports.createVenueData` it was. The problem was in `exports.updateVenue` where it wasn't included. That's why it worked for create, but not for edit/update.

The additional changes here unify both locations to be 'driven' by the same array, so that there is only a single canonical location to update in future, and thus it will be impossible to make this mistake/forget to update both anymore.

This was also the same issue for a long outstanding `firebarrel` bug, implying that it was a relatively easy mistake to make.

functions/venue.js
```diff
@@ -553,26 +569,15 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
...
-      break;
+  if (IFRAME_TEMPLATES.includes(updated.template) && data.iframeUrl) {
+    updated.iframeUrl = data.iframeUrl;
+  }
 
-    default:
-      break;
+  if (ZOOM_URL_TEMPLATES.includes(updated.template) && data.zoomUrl) {
+    updated.zoomUrl = data.zoomUrl;
   }
```
0xdevalias, line 581: RE: https://github.com/sparkletown/sparkle/pull/1274#pullrequestreview-613992793

> I couldn't spot the root of the issue

@mike-lvov The root of the issue was that `iframeUrl` is never set for `VenueTemplate.embedded` / `VenueTemplate.firebarrel` here in `exports.updateVenue`

image

0xdevalias:
> I couldn't spot the root of the issue

@mike-lvov see explanation in https://github.com/sparkletown/sparkle/pull/1274/files/54e096c1dcfe3eee8288e86c548a0f467e8c76f0#r595776640

functions/venue.js
0xdevalias: see explanation in https://github.com/sparkletown/sparkle/pull/1274#discussion_r595776640

PR #1286, opened 2021-03-18
Fixes hyphens and sound effects

mike-lvov:
Fixes hyphens and sound effects


0xdevalias (commented):
[Retroactive Review] Changes here look fine, though confusing why the ordering of `AuthIsLoaded` and `CustomSoundsProvider` would change/fix an audio issues experienced here.

PR #1290, opened 2021-03-25
Fix emojis not muting

malaikaaiyar:
Note: this is the fix that @0xdevalias suggested in the comments of https://github.com/sparkletown/internal-sparkle-issues/issues/310 (to add a default state to the `isAudioEffectDisabled` prop, but it's not working - either I've implemented this wrong or there is a worse logic issue. 

Normally (and after this line was added) the auditorium emoji reaction muting works fine, but the jazzbar emojis don't mute.
With this change, the auditorium emojis still work the same but the jazzbar emojis don't make any noise at all.

closes https://github.com/sparkletown/internal-sparkle-issues/issues/310


0xdevalias:
> Note: this is the fix that @0xdevalias suggested in the comments of [sparkletown/internal-sparkle-issues#310](https://github.com/sparkletown/internal-sparkle-issues/issues/310) (to add a default state to the `isAudioEffectDisabled` prop, but it's not working - either I've implemented this wrong or there is a worse logic issue.
> 
> Normally (and after this line was added) the auditorium emoji reaction muting works fine, but the jazzbar emojis don't mute.
> With this change, the auditorium emojis still work the same but the jazzbar emojis don't make any noise at all.

@malaikaaiyar I can’t remember where I mentioned this previously, but it sounds like the mute button isn’t actually wired up in the jazz bar. I assume the prop isn’t being passed to the appropriate component at all. So you would need to add the useState/etc to the jazz bar in a similar way to how it currently exists in the auditorium. 

That should then fix the remaining aspect of this issue I would expect.

0xdevalias (changes requested):
See my previous comments/requests for changes in https://github.com/sparkletown/sparkle/pull/1290#issuecomment-807769190

0xdevalias:
I've added further deep dive/high level plan of the solution in https://github.com/sparkletown/internal-sparkle-issues/issues/310#issuecomment-811682151 + https://github.com/sparkletown/internal-sparkle-issues/issues/310#issuecomment-811683137

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -169,6 +169,7 @@ const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
 UserProfilePicture.defaultProps = {
   avatarClassName: "profile-icon",
   miniAvatars: false,
+  isAudioEffectDisabled: true,
 };
```
0xdevalias, line 173: [NB] It would be nice to refactor this to our standard pattern of providing defaults at the top of the component, rather than this `ComponentName.defaultProps` pattern.

0xdevalias (approved):
[CRO] Code changes look good here.

I'm assuming that you've tested this and that it seems to work correctly in all the places now?

Did you end up checking the ❓❓❓ items listed in https://github.com/sparkletown/internal-sparkle-issues/issues/310#issuecomment-811682151 to see if they also need changes?

---

It would be good to get the `defaultProps` changed before merging this PR as well.

Assuming all of the above are properly checked/tested + those changes made, then I say LGTM!

malaikaaiyar:
I tested it in the jazz bar and it works there. 

For the other places, they don't have a mute reaction button or use `isAudioEffectDisabled` anywhere, so should I add the state and pass it to `Room.tsx` (in `organisms/Room`)

0xdevalias:
> For the other places, they don't have a mute reaction button or use `isAudioEffectDisabled` anywhere, so should I add the state and pass it to `Room.tsx` (in `organisms/Room`)

Nah, if they don’t have a mute button/the reaction feature then we shouldn’t need to do anything.

That’s what the default value I described as the original fix for this will resolve.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -34,8 +34,8 @@ export interface UserProfilePictureProp {
 // @debt the UserAvatar component serves a very similar purpose to this, we should unify them as much as possible
 const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
   isAudioEffectDisabled,
```
0xdevalias, line 36: You seem to have lost the default on this that was the original fix in this PR. That’s still required

PR #1291, opened 2021-03-25
Validate room name length

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/476 
(only partly fixes because the existing over 20 character room names in the database need to be changed/handled)

How it looks:
Screen Shot 2021-03-25 at 9 40 47 PM

### DIFF
--- a/src/pages/Admin/Venue/DetailsValidationSchema.ts
+++ b/src/pages/Admin/Venue/DetailsValidationSchema.ts
@@ -59,6 +59,7 @@ export const validationSchema = Yup.object()
     name: Yup.string()
       .required("Required")
       .min(1, "Required")
+      .max(20, ({ max }) => `Name must be less than ${max} characters`)
       .when(
         "$editing",
         (editing: boolean, schema: Yup.StringSchema) =>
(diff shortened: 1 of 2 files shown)


src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -59,6 +59,7 @@ export const validationSchema = Yup.object()
     name: Yup.string()
       .required("Required")
```
TroySK, line 60: Unrelated to your change but would be great if these texts are replaced with more descriptive ones like in `src/pages/Admin/Details/ValidationSchema.ts`

TroySK (approved):
lgtm otherwise

mike-lvov (approved):
lgtm

src/pages/Admin/Venue/DetailsValidationSchema.ts
malaikaaiyar: I can make a new PR for this

0xdevalias: See https://github.com/sparkletown/sparkle/pull/1297

0xdevalias (commented):
[Retroactive Review] Approved

PR #1293, opened 2021-03-29
Fix retroactive review comments from sparkletown/sparkle#1249 RE: pointer-events in Audience component

yurii-lubynets:
update Audience component styles

closes https://github.com/sparkletown/internal-sparkle-issues/issues/578


mike-lvov (approved):
lgtm

PR #1294, opened 2021-03-29
[BIGTOP ENV] Announcement banner color change

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/599

### DIFF
--- a/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
+++ b/src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
@@ -8,7 +8,7 @@ $message-margin: $spacing + $close-icon-width + 5px;
   display: block;
   z-index: z(announcement);
   position: absolute;
-  background: $primary;
+  background: #213565;
   top: 66px;
   left: 50%;
   width: 500px;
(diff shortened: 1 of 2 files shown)


0xdevalias (commented):
While the explicit ask was the announcement banner, would it not be better to update the `$primary` variable colour in general? I would assume that is more likely to be the intent of the ask, even if they didn't know to describe it in as many words.

I also raised this on slack, as I'm not sure this sort of thing is something we should even be doing on a request by request basis:

- https://doesitsparkle.slack.com/archives/C01700P4QCT/p1617058152002600

0xdevalias:
> Regarding the request itself, I think we should create UI for color scheme customizations because it looks like there will be a lot of requests of this kind in the future

@mike-lvov Agreed RE: this part, and I believe it is in our plans to do so (eg. see https://github.com/sparkletown/internal-sparkle-issues/issues/602 + all of the related 'design system'/'styleguide' issues such as https://github.com/sparkletown/internal-sparkle-issues/issues/4, https://github.com/sparkletown/internal-sparkle-issues/issues/242, the designs/ around colour customisation in admin v3, etc)

My specific 'maybe we should push back on doing these sorts of custom tweaks' is based on the current state of the codebase, and the fact that it takes extra manual work/tech debt to do it currently; not a desire for us to never have that capability.

If @sparkletown/product decides that it's an important enough feature to prioritise building the tooling around self-serve customisation of these sorts of things sooner rather than later then I'm happy to roll with that.

0xdevalias (commented):
[Retroactive Review] Approved

PR #1296, opened 2021-03-30
Properly validate room urls to ensure they are valid, then remove our workaround

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/221
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/9

### DIFF
--- a/src/pages/Admin/Details/ValidationSchema.ts
+++ b/src/pages/Admin/Details/ValidationSchema.ts
@@ -126,9 +126,9 @@ const roomTitleSchema = Yup.string()
   .required("Room name is required")
   .min(3, ({ min }) => `Name must be at least ${min} characters`);
 const roomUrlSchema = Yup.string()
-  .test("url validation", "Enter correct url", isValidUrl)
   .required("Url is required!")
-  .min(3, ({ min }) => `Url must be at least ${min} characters`);
+  .min(3, ({ min }) => `Url must be at least ${min} characters`)
+  .test("url validation", "Enter correct url", isValidUrl);
 const roomImageUrlSchema = Yup.string().required("Room image is required");
 
 export const roomCreateSchema = Yup.object().shape<RoomSchemaShape>({
--- a/src/pages/Admin/Details/ValidationSchema.ts
+++ b/src/pages/Admin/Details/ValidationSchema.ts
@@ -125,10 +125,12 @@ export const validationSchema_v2 = Yup.object()
 const roomTitleSchema = Yup.string()
   .required("Room name is required")
   .min(3, ({ min }) => `Name must be at least ${min} characters`);
-const roomUrlSchema = Yup.string()
+
+export const roomUrlSchema = Yup.string()
   .required("Url is required!")
   .min(3, ({ min }) => `Url must be at least ${min} characters`)
-  .test("url validation", "Enter correct url", isValidUrl);
+  .test("url validation", "Please enter a valid URL", isValidUrl);
+
 const roomImageUrlSchema = Yup.string().required("Room image is required");
 
 export const roomCreateSchema = Yup.object().shape<RoomSchemaShape>({
--- a/src/pages/Admin/Details/ValidationSchema.ts
+++ b/src/pages/Admin/Details/ValidationSchema.ts
@@ -10,6 +10,7 @@ import {
   PLAYA_WIDTH,
   PLAYA_HEIGHT,
 } from "settings";
+import { isValidUrl } from "utils/url";
 
 const initialMapIconPlacement: VenueInput["placement"] = {
   x: (PLAYA_WIDTH - PLAYA_VENUE_SIZE) / 2,
@@ -125,6 +126,7 @@ const roomTitleSchema = Yup.string()
   .required("Room name is required")
   .min(3, ({ min }) => `Name must be at least ${min} characters`);
 const roomUrlSchema = Yup.string()
+  .test("url validation", "Enter correct url", isValidUrl)
   .required("Url is required!")
   .min(3, ({ min }) => `Url must be at least ${min} characters`);
 const roomImageUrlSchema = Yup.string().required("Room image is required");
--- a/src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
+++ b/src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
@@ -26,7 +26,12 @@ export const validationSchema = Yup.object()
     title: Yup.string().required("Required"),
     subtitle: Yup.string().required("Required"),
     about: Yup.string().required("Required"),
-    url: Yup.string().required("Required"),
+    url: Yup.string()
+      .matches(
+        /((https?):\/\/)?(www.)?[a-z0-9]+(\.[a-z]{2,}){1,3}(#?\/?[a-zA-Z0-9#]+)*\/?(\?[a-zA-Z0-9-_]+=[a-zA-Z0-9-%]+&?)?$/,
+        "Enter correct url!"
+      )
+      .required("Required"),
     x_percent: Yup.number()
       .default(INITIAL_PERCENTAGE_POS)
       .required("Required")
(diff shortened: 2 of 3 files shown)


src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -126,7 +127,8 @@ const roomTitleSchema = Yup.string()
   .min(3, ({ min }) => `Name must be at least ${min} characters`);
 const roomUrlSchema = Yup.string()
   .required("Url is required!")
-  .min(3, ({ min }) => `Url must be at least ${min} characters`);
+  .min(3, ({ min }) => `Url must be at least ${min} characters`)
+  .test("url validation", "Enter correct url", isValidUrl);
```
0xdevalias, line 131: Maybe change "Enter correct url" to something like "Please enter a valid URL" or "Entered URL is invalid" or similar (i'm not 100% how this shows up in the app/etc.

Or maybe even something that explains the format the URL needs to be in, as that would be most helpful to the user?

---

Ref: https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema

src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
```diff
@@ -26,7 +27,9 @@ export const validationSchema = Yup.object()
     title: Yup.string().required("Required"),
     subtitle: Yup.string().required("Required"),
     about: Yup.string().required("Required"),
-    url: Yup.string().required("Required"),
+    url: Yup.string()
+      .required("Required")
+      .test("url validation", "Enter correct url", isValidUrl),
```
0xdevalias, line 32: Can we export `roomUrlSchema` from `src/pages/Admin/Details/ValidationSchema.ts` and then reuse it directly here?

```suggestion
    url: roomUrlSchema,
```

src/utils/url.ts
```diff
@@ -40,20 +40,6 @@ export const openRoomUrl = (url: string) => {
...
-        `Invalid URL ${url} on page ${window.location.href}; allowing for now (workaround)`
-      ),
-      (event) => {
-        event.addMetadata("utils/url::openUrl", { url });
-      }
-    );
-    // @debt keep the checking in place so we can debug further, but don't block attempts to open
-    // return;
-  }
-
```
0xdevalias, line 56: I believe we want to keep this (particularly the Bugsnag error reporting bit) so that we know if/when data is invalid (eg. all of the existing data in the database that may not match the correct URL format)

But we likely want to get rid of the `@debt` part and restore the `return` that 'bails out early' when it isn't a valid URL.

src/utils/url.ts
```diff
@@ -63,8 +49,18 @@ export const openUrl = (url: string) => {
...
   try {
-    return VALID_URL_PROTOCOLS.includes(new URL(url).protocol);
+    if (new RegExp(relativeLinksRegex).test(url)) {
+      return checkUrl(window.location.origin + url);
+    }
+    if (new RegExp(shortUrlRegex).test(url)) {
+      return checkUrl(window.location.protocol + url);
+    }
+    return checkUrl(url);
   } catch (e) {
```
0xdevalias, line 64: The changes here look quite convoluted for what I believe should be a relatively simple fix.

---

Regex explanations (where `*` in the below descriptions denotes anything):

- `relativeLinksRegex`: matches `/in/*` and `/v/*` urls
- `shortUrlRegex`: matches `//*` urls

---

Looking at the docs for `URL`, it seems that by including the `base` param, it can handle relative urls:

- https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters
  - If `url` is a relative URL, `base` is required, and will be used as the base URL. If `url` is an absolute URL, a given `base` will be ignored.
  - https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#examples

Playing around with that:

```javascript
console.log(window.location.origin)
// https://developer.mozilla.org

new URL('//', window.location.origin).toString()
// Uncaught TypeError: Failed to construct 'URL': Invalid URL

new URL('//foo', window.location.origin).toString()
// "https://foo/"

new URL('//foo.com', window.location.origin).toString()
// "https://foo.com/"

new URL('/in/aaa', window.location.origin).toString()
// "https://developer.mozilla.org/in/aaa"

new URL('/v/bbb', window.location.origin).toString()
// "https://developer.mozilla.org/v/bbb"
```
  
So I believe that removing all of the new code here, and just changing line 67 to something like the following might be sufficient:

```typescript
return VALID_URL_PROTOCOLS.includes(new URL(url, window.location.origin).protocol);
```

Note: I haven't checked through all of the reported errors to see what other URL patterns exist, so it may need to be slightly more complex than this

0xdevalias (changes requested):
Comments/suggestions

0xdevalias:
@yurii-lubynets FYI, we don't need to add the PR's to the sprint project boards directly when we link them to their respective issues using 'closing keywords' (as you have done here), as the issues themselves will already be present in the project board (and they will display their linked PRs)

src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
yurii-lubynets: fixed

src/utils/url.ts
yurii-lubynets: restored the Bugsnag error reporting and fixed return statement

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Let's just keep it as `'Please enter a valid URL'` for now, we can worry about coming up with a better/more descriptive message in a future cleanup if we need to.

src/utils/url.ts
yurii-lubynets: you are right. It's the most suitable way to validate inputting URLs. Have checked the error cases from the Bugsnag

0xdevalias: > Have checked the error cases from the Bugsnag

When you say that, do you mean that you've checked against the error cases and this is sufficient to handle them? Or do you still need to do something else to improve this check further?

0xdevalias:
@yurii-lubynets Is this ready for re-review, or does it need more work done to it still first?

When it's ready for re-review, you can signal that by clicking the 're-request review' button in the top right. That's how i'll know it's ready to go again:

image

src/utils/url.ts
```diff
@@ -50,10 +50,8 @@ export const openUrl = (url: string) => {
         event.addMetadata("utils/url::openUrl", { url });
       }
     );
-    // @debt keep the checking in place so we can debug further, but don't block attempts to open
-    // return;
+    return;
   }
-
```
0xdevalias, line 56: [nit] We want to keep this whitespace as it provides clean separation from the if statement below, etc.

0xdevalias (changes requested):
A couple of little things, then this should be good to go!

src/utils/url.ts
yurii-lubynets: no, it's fine, have already checked the error cases

src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
```diff
@@ -26,7 +27,7 @@ export const validationSchema = Yup.object()
     title: Yup.string().required("Required"),
     subtitle: Yup.string().required("Required"),
     about: Yup.string().required("Required"),
-    url: Yup.string().required("Required"),
+    url: roomUrlSchema,
```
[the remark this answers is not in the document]
yurii-lubynets: Yes, I have tested it locally and confirmed this pattern is worked.

src/utils/url.ts
```diff
@@ -50,10 +50,8 @@ export const openUrl = (url: string) => {
         event.addMetadata("utils/url::openUrl", { url });
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/utils/url.ts
yurii-lubynets: fixed

src/utils/url.ts
0xdevalias: Awesome :)

src/pages/Admin/Venue/Rooms/RoomsValidationSchema.ts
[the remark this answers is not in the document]
0xdevalias: Awesome!

0xdevalias (approved):
Looks good, LGTM!

PR #1300, opened 2021-04-01
Fix Room data (eg. enterSound) being overwritten when rooms are edited

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/388


src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -67,8 +67,12 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
   );
 
   const onSubmit = useCallback(() => {
-    return submitHandler(values, room.roomIndex!);
-  }, [room.roomIndex, submitHandler, values]);
+    const roomValues = {
```
0xdevalias, line 70: Just to make it more explicit/obvious:

```suggestion
    const roomValues: RoomData_v2 = {
```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -67,8 +67,12 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+      ...room,
+      ...values,
+    };
+    return submitHandler(roomValues, room.roomIndex!);
```
0xdevalias, line 74: We want to eliminate this `!` TypeScript operator from our codebase. It disables type checking and says that there can never be an `undefined` value here, which has lead to weird/hard to find bugs:

```suggestion
    return submitHandler(roomValues, room.roomIndex);
```

This is likely going to need some other related checks/type changes as well once this change is made.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -137,7 +137,11 @@ const RoomInnerForm: React.FC<RoomInnerForm> = (props) => {
     async (vals: Partial<FormValues>) => {
       if (!user) return;
       try {
-        await upsertRoom(vals as RoomInput, venueId, user, editingRoomIndex);
+        const roomValues = {
+          ...editingRoom,
+          ...vals,
+        } as RoomInput;
```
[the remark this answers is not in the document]
0xdevalias: There will likely be some type issues that will need fixing up when you do this. Seemingly possibly just for `title`:

image

0xdevalias (changes requested):
Based on the changes made here, the 'root' of this issue appears to be that we're not passing the full data through to the `upsertRoom` API function.

Searching for usages of this, there appear to be more locations that may also need fixes to be made:

- https://github.com/sparkletown/sparkle/search?q=upsertRoom
  - https://github.com/sparkletown/sparkle/blob/staging/src/pages/Admin/AdminVenueRoomDetails.tsx#L62-L67

Similarly, there seems to be an almost identical function defined called `updateRoom`, so the fix will likely need to be applied to these areas as well:

- https://github.com/sparkletown/sparkle/search?q=updateRoom+%22api%2Fadmin%22
  - `src/pages/Admin/AdminVenueRoomDetails.tsx`
  - `src/pages/Admin/Room/Card/RoomCard.tsx`
  - `src/pages/Admin/MapPreview/MapPreview.tsx`
  - `src/pages/Admin/Venue/Details/VenueDetails.tsx`

Can you please check and ensure that the fixes required here are applied in all locations.

---

While I was diving deeper into this, I found the following, which might also be good to cleanup as part of this PR, to fix the backend function to be more robust/less likely to cause similar issues to this in the future:

The frontend API calls (`upsertRoom`, `updateRoom`) are defined in:

- https://github.com/sparkletown/sparkle/blob/staging/src/api/admin.ts#L435-L471

Which call:

- `createFirestoreRoomInput`: https://github.com/sparkletown/sparkle/blob/staging/src/api/admin.ts#L336-L381
- `createFirestoreRoomInput_v2`: https://github.com/sparkletown/sparkle/blob/staging/src/api/admin.ts#L383-L433

**Note:** These functions are somewhat long/convoluted (seemingly mostly related to uploading files/rewriting the data based on that), and could almost certainly be cleaned up/simplified, at the very least to use a shared helper method for the uploads) (Though this is likely OOS of this PR)

`upsertRoom`, `updateRoom` then call:

- `firebase.functions().httpsCallable("venue-upsertRoom")`

---

The backend function `upsertRoom` is defined in:

- https://github.com/sparkletown/sparkle/blob/staging/functions/venue.js#L338-L356

This basically gets the existing venue document, does some checks, then extracts it's data (as `docData`), decides whether to append the `room` or overwrite the existing `room`, and then writes the updated data back to the firestore database.

The way this is written, it will actually overwrite the entire venue config object, and since it isn't done within a transaction, this introduces a potential race condition bug.

Since we only actually care about updating the `rooms` key, we should be doing something like this:

```diff
-  if (typeof roomIndex !== "number") {
-    docData.rooms = [...docData.rooms, room];
-  } else {
-    docData.rooms[roomIndex] = room;
-  }
-
- admin.firestore().collection("venues").doc(venueId).update(docData);

+ const newRoomsData = typeof roomIndex !== "number" ? [...docData.rooms, room] : docData.rooms.splice(roomIndex, 1, room);
+ admin.firestore().collection("venues").doc(venueId).update({ rooms: newRoomsData });
```

This should tell it to only update the `room` array, rather than overwriting the entire venue config each time. If we wanted to be super safe/sure about this, we would also still wrap the get/update within a transaction to ensure there can never be a race condition here.

**Note:** This seems to also be an issue in `deleteRoom` and likely other similar backend functions.

**Further reading:**

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice
- https://firebase.google.com/docs/firestore/manage-data/add-data#update-data
- https://firebase.google.com/docs/firestore/manage-data/add-data#update_fields_in_nested_objects
  - > Dot notation allows you to update a single nested field without overwriting other nested field. If you update a nested field without dot notation, you will overwrite the entire map field
  - Note: I don't think this will help us in this instance, as we care about maintaining the room index/ordering/etc. Including here just as a point of learning around that may help with firebase things in future.
- https://firebase.google.com/docs/firestore/manage-data/transactions#transactions

yurii-lubynets:
@0xdevalias 
Thank you a lot for your huge help!

Firstly, I should mention there is currently no way to update a single array element (or add/remove a single element) in Cloud Firestore (If your document contains an array field, you can use arrayUnion() and arrayRemove() to add and remove elements based on their docs). But I assume these changes could be extremely huge and hard to cover all cases.
Secondly,

> This should tell it to only update the room array, rather than overwriting the entire venue config each time

Firestore rewrites the room array but doesn't update the item (just rewrite it). So here is the main problem I've faced.
Thirdly, I suppose this issue could be fixed by ensuring that the fixes required here are applied in all locations. Checked, covered, and tested it.

0xdevalias:
@yurii-lubynets please remember to use the re-request review button in the top-right corner of the GitHub PR page. That signals that this PR is ready to be reviewed again, and ensures I get notified of it, etc.

Further Reading:

- https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/
- https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review
  - > After your pull request is reviewed and you've made the necessary changes, you can ask a reviewer to re-review your pull request. Navigate to Reviewers in the right sidebar and click  next to the reviewer's name whose review you'd like.

functions/venue.js
```diff
@@ -345,14 +345,15 @@ exports.upsertRoom = functions.https.onCall(async (data, context) => {
...
   if (typeof roomIndex !== "number") {
-    docData.rooms = [...docData.rooms, room];
+    rooms = [...rooms, room];
   } else {
-    docData.rooms[roomIndex] = room;
+    rooms[roomIndex] = room;
   }
 
-  admin.firestore().collection("venues").doc(venueId).update(docData);
+  admin.firestore().collection("venues").doc(venueId).update({ rooms });
```
0xdevalias, line 356: The solution described in https://github.com/sparkletown/sparkle/pull/1300#pullrequestreview-628399237 is a better way to solve this:

> Since we only actually care about updating the `rooms` key, we should be doing something like this:
> 
> ```diff
> -  if (typeof roomIndex !== "number") {
> -    docData.rooms = [...docData.rooms, room];
> -  } else {
> -    docData.rooms[roomIndex] = room;
> -  }
> -
> - admin.firestore().collection("venues").doc(venueId).update(docData);
> 
> + const newRoomsData = typeof roomIndex !== "number" ? [...docData.rooms, room] : docData.rooms.splice(roomIndex, 1, room);
> + admin.firestore().collection("venues").doc(venueId).update({ rooms: newRoomsData });
> ```
> 
> This should tell it to only update the `room` array, rather than overwriting the entire venue config each time. If we wanted to be super safe/sure about this, we would also still wrap the get/update within a transaction to ensure there can never be a race condition here.

0xdevalias: It would also be good to wrap this whole section of code (from the initial DB read, to this update) in a transaction:

- https://firebase.google.com/docs/firestore/manage-data/transactions

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -67,8 +67,13 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
   );
 
   const onSubmit = useCallback(() => {
-    return submitHandler(values, room.roomIndex!);
-  }, [room.roomIndex, submitHandler, values]);
+    if (typeof room.roomIndex !== "number") return;
```
0xdevalias, line 70: In what cases will this not be a `number`? Is it only when it's `undefined`? Can it ever be the case? What happens when we 'silently return' here?

I feel like we should be signalling an error in some way or another if this isn't handled some other way, either directly to the client UI, through `Bugsnag.notify`, or potentially both (depending on what the 'solution' to the error is)

src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -154,6 +154,7 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
 
   const handleEditRoomSave = async (values: RoomData_v2, index: number) => {
     const newData = {
+      ...editingRoom,
       ...values,
       x_percent: editingRoom?.x_percent,
       y_percent: editingRoom?.y_percent,
```
0xdevalias, line 160: If we're spreading `...editingRoom` here now, is there any reason to keep the explicit `x_percent`/etc keys below?

Should this be happening before `...values`, or after it?

0xdevalias (changes requested):
Questions/Comments/Changes

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1300#issuecomment-814380763

> Thank you a lot for your huge help!

No worries :) This is how I tend to review things, to ensure the codebase quality stays high, and our patterns/best practices are followed.

---

> Firstly, I should mention there is currently no way to update a single array element (or add/remove a single element) in Cloud Firestore (If your document contains an array field, you can use arrayUnion() and arrayRemove() to add and remove elements based on their docs). But I assume these changes could be extremely huge and hard to cover all cases.

Yeah, that's fair enough. That's why I went with the proposal outlined in https://github.com/sparkletown/sparkle/pull/1300#discussion_r608284534, which uses `array.splice()` to create a new array, replacing only the element we want to update, without modifying the existing array:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

I feel like `arrayUnion()`/`arrayRemove()` are more useful when using primitive types (eg. `string`), I wouldn't 'trust' using them for more complex types like this.

---

> Firestore rewrites the room array but doesn't update the item (just rewrite it). So here is the main problem I've faced.

Yup, that's why I use the `array.splice()` to replace the element within the array that we want to 'update' (though obviously this still relies on the frontend passing us the correct data. We could technically handle the 'merge' of old/new data on the backend, but that's more complex with the nuanced edgecases it might bring up than I would want to get into for this PR)

---

> Thirdly, I suppose this issue could be fixed by ensuring that the fixes required here are applied in all locations. Checked, covered, and tested it.

Yeah, ensuring all the frontend areas pass the right data through will definitely fix the core of the issue as it exists for now.

Some of the other little tweaks/changes I described in https://github.com/sparkletown/sparkle/pull/1300#pullrequestreview-628399237 will also fix/avoid the potential for other bugs/etc at the same time, which is why I mentioned them too.

functions/venue.js
yurii-lubynets: `const newRoomsData = typeof roomIndex !== "number" ? [...docData.rooms, room] : docData.rooms.splice(roomIndex, 1, room);`

Here the method `splice()` returns not the modified array in `newRoomsData` but the modified element only. 
We should keep `docData.rooms` array in variable `rooms` and then use `splice()` and after that `rooms`  will have a modified array.

0xdevalias: > Here the method splice() returns not the modified array in newRoomsData but the modified element only.

Yup, you're correct about that 🤦🏻 

I think I read this as 'returns the array with the replaced elements' or something. Brainfail.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#return_value
  - > An array containing the deleted elements.
  
> We should keep `docData.rooms` array in variable rooms and then use `splice()` and after that rooms will have a modified array.

Sounds like a good plan 👌🏻

src/pages/Admin/Room/Edit/RoomEditModal.tsx
yurii-lubynets: Well, (property) RoomData_v2.roomIndex?: number | undefined and we should check if it isn't a number.
 Actually, it's just a check for TS cause this case (`roomIndex` is undefined) can't be alive (when you edit room - you will definitely have roomIndex)

src/pages/Admin/Venue/Details/VenueDetails.tsx
yurii-lubynets: We could avoid this in a new way. Just to update the only properties from edit form

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: In that case, we could probably just do `if (room.roomIndex === undefined) return;` to make it more explicit.

src/pages/Admin/Venue/Details/VenueDetails.tsx
0xdevalias: > Just to update the only properties from edit form

Not sure what you mean by this, can you please elaborate?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
yurii-lubynets: fixed

src/pages/Admin/Venue/Details/VenueDetails.tsx
yurii-lubynets: We are spreading the `editingRoom` and are merging the `values` into one obj `newData`. It should work smoothly

0xdevalias: Cool cool.

Can `values` ever container `x_percent`/etc in a way that we don't want to keep it/allow it to be overwritten from what was already in `editingRoom`? That's about the only potential edgecase I can see with this change.

0xdevalias: `values` appears to be of type `RoomData_v2`, so at least in theory could have these keys set.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -67,8 +67,13 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
   );
 
   const onSubmit = useCallback(() => {
-    return submitHandler(values, room.roomIndex!);
-  }, [room.roomIndex, submitHandler, values]);
+    if (room.roomIndex === undefined) return;
```
0xdevalias, line 70: We tend to keep a space after these sorts of 'guard statements' to make them stand out more:
```suggestion
    if (room.roomIndex === undefined) return;

```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -67,8 +67,13 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+    const roomValues: RoomData_v2 = {
+      ...room,
+      ...values,
+    };
+    return submitHandler(roomValues, room.roomIndex);
```
0xdevalias, line 75: [nit]
```suggestion
    };

    return submitHandler(roomValues, room.roomIndex);
```

src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -153,15 +153,10 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
...
-    const newData = {
+    const newData: RoomData_v2 = {
+      ...editingRoom,
       ...values,
-      x_percent: editingRoom?.x_percent,
-      y_percent: editingRoom?.y_percent,
-      width_percent: editingRoom?.width_percent,
-      height_percent: editingRoom?.height_percent,
-      isEnabled: editingRoom?.isEnabled,
     };
```
0xdevalias, line 159: [nit]
```suggestion
    };

```

src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -153,15 +153,10 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
...
   const handleEditRoomSave = async (values: RoomData_v2, index: number) => {
-    const newData = {
+    const newData: RoomData_v2 = {
+      ...editingRoom,
       ...values,
-      x_percent: editingRoom?.x_percent,
-      y_percent: editingRoom?.y_percent,
-      width_percent: editingRoom?.width_percent,
-      height_percent: editingRoom?.height_percent,
-      isEnabled: editingRoom?.isEnabled,
```
0xdevalias, line 162: Pending satisfactory resolution of investigating https://github.com/sparkletown/sparkle/pull/1300#discussion_r609102083

0xdevalias (approved):
Pending a satisfactory resolution of investigating https://github.com/sparkletown/sparkle/pull/1300#discussion_r609102083

+ ideally making the few other small changes mentioned in this review.

Then this should be good to go!

PR #1301, opened 2021-04-02
Run audit fix to reduce the number of high vulnerabilities

sunny-viktoryia:
I faced +30 high vulnerabilities while installing the packages.
So here are the updated versions after running the `npm audit fix`.


0xdevalias:
@sunny-viktoryia FYI: I've specifically avoided doing this as it will change a LOT of underlying packages all at once, and that has the high likelihood of breaking something we aren't going to catch due to not having any automated testing at the moment.

We use dependabot, and ideally we would check and merge those updates individually as we have the time to, then it will open a few new PR's, etc. Repeat until all of our dependencies are up to date.

IMO this is a much more methodical and safer way to approach it, particularly given our current state of lack of automated testing.

(cc // @jonboutelle @mike-lvov FYI. Can be overruled on this if that is the mass consensus, but I want to make sure we de-risk ourselves as much as possible, and IMO mass updates like this are a way higher risk to us than the couple of security issues in dependencies that likely aren't even exposed/exploitable/etc in our app/configuration)

0xdevalias (changes requested):
Formally marking this as a review, see comments/concerns in https://github.com/sparkletown/sparkle/pull/1301#issuecomment-812500687

sunny-viktoryia:
closing this one in order to fix package vulnerabilities one by one in a safer way

PR #1305, opened 2021-04-05
add un-clickable room type setting for admin v1 and v2

TroySK:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/612

We have the setting at the backend as a string while usage is binary. Switching to boolean now, might make sense to prevent pain in the future. There maybe more planned usage for the field in the future which I am not aware of.


0xdevalias:
@TroySK FYI, please use 'closing keywords' to link to the related issue:

- https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

image

0xdevalias:
> We have the setting at the backend as a string while usage is binary. Switching to boolean now, might make sense to prevent pain in the future. There maybe more planned usage for the field in the future which I am not aware of.

@TroySK While it is a `string` on the backend, it is actually an `enum`, as can be seen at:

- https://github.com/sparkletown/sparkle/blob/staging/src/types/rooms.ts#L3-L5

It was explicitly designed this way rather than as a `boolean` so that we can add other room types in future if/as required.

functions/venue.js
```diff
@@ -68,6 +68,8 @@ const PlacementState = {
   Hidden: "HIDDEN",
 };
 
+const UNCLICKABLE_ROOM_TYPE = "UNCLICKABLE";
```
0xdevalias, line 71: This should be implemented as an 'enum' in a similar war to `VenueTemplate`/etc.

See also: https://github.com/sparkletown/sparkle/pull/1305#issuecomment-813780291


```suggestion
// These represent all of our 'special' room types (they should remain alphabetically sorted) If no type is set, then it is just a 'normal room'.
// @debt unify this with RoomTypes in src/types/rooms.ts + share the same code between frontend/backend
const RoomType = {
  unclickable = "UNCLICKABLE",
}
```

(It would also be good to rename the existing `RoomTypes` enum in the frontend code to `RoomType`, as enums shouldn't be pluralised)

functions/venue.js
```diff
@@ -113,6 +115,7 @@ const createVenueData = (data, context) => {
     profile_questions: data.profile_questions,
     placement: { ...data.placement, state: PlacementState.SelfPlaced },
     showLiveSchedule: data.showLiveSchedule ? data.showLiveSchedule : false,
+    type: data.type === UNCLICKABLE_ROOM_TYPE,
```
0xdevalias, line 118: What is the intent of this change here? I don't think it is correct.. but it would be good to understand what you were meaning to do with it/why.

src/pages/Admin/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -284,6 +286,13 @@ const AdvancedSettings: React.FC<AdvancedSettingsProps> = ({
 
         {renderShowGridToggle()}
 
+        <ToggleElement
+          forwardRef={register}
+          isChecked={values.type}
+          name="type"
+          title="Set room un-clickable"
+        />
```
0xdevalias, line 294: The unclickable value isn't part of the `Venue`, it should be part of the `Room` (and is already defined on that type):

- https://github.com/sparkletown/sparkle/blob/staging/src/types/rooms.ts#L9
- https://github.com/sparkletown/sparkle/blob/staging/src/types/rooms.ts#L28

This shouldn't be a `boolean`, it is an enum of `RoomTypes`.

However, we could choose to 'hide' that nuance in the frontend for now if we want by making it a 'checkbox' that will cause the enum value to be set. This might be simpler UI/UX for now, as we only have a single `type` at the moment.

src/pages/Admin/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -99,6 +100,7 @@ const AdvancedSettings: React.FC<AdvancedSettingsProps> = ({
       columns: venue.columns,
       radioStations: venue.radioStations ? venue.radioStations[0] : "",
       requiresDateOfBirth: venue.requiresDateOfBirth,
+      type: venue.type,
```
0xdevalias, line 103: `type` is defined on `Room`, not `Venue`.

```suggestion
```

src/pages/Admin/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -72,6 +72,7 @@ const validationSchema = Yup.object().shape<Venue_v2_AdvancedConfig>({
     then: Yup.string().required("Radio stream is required!"),
   }),
   requiresDateOfBirth: Yup.bool().notRequired(),
+  type: Yup.bool().notRequired(),
```
0xdevalias, line 75: `type` is defined on `Room`, not `Venue`. And it is an optional `enum` of `RoomTypes`, not a `boolean`.

```suggestion
```

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -551,6 +551,16 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </div>
   );
 
+  const renderRoomTypeToggle = () => (
+    <div className="toggle-room">
+      <h4 className="italic input-header">set room un-clickable</h4>
+      <label id={"type"} className="switch">
+        <input type="checkbox" id={"type"} name={"type"} ref={register} />
+        <span className="slider round"></span>
+      </label>
```
0xdevalias, line 560: When nesting labels like this we shouldn't need to set the `id`, and in React, we almost never want to use an `id`, as there is no need to.

Similarly, with basic string props, there is no need to wrap them in `{` `}`.

```suggestion
      <label className="switch">
        <input type="checkbox" name="type" ref={register} />
        <span className="slider round"></span>
      </label>
```

0xdevalias: As mentioned elsewhere, we don't want `type` to be added to venue config, so this may need to be reworked.

src/types/venues.ts
```diff
@@ -83,6 +83,7 @@ export interface Venue_v2_AdvancedConfig {
   radioStations?: string | string[]; // single string on form, array in DB
   requiresDateOfBirth?: boolean;
   roomVisibility?: RoomVisibility;
+  type?: boolean;
```
0xdevalias, line 86: `type` is defined on `Room`, not `Venue`.

```suggestion
```

src/types/venues.ts
```diff
@@ -137,6 +138,7 @@ export interface BaseVenue {
   profileAvatars?: boolean;
   hideVideo?: boolean;
   showLiveSchedule?: boolean;
+  type?: boolean;
```
0xdevalias, line 141: `type` is defined on `Room`, not `Venue`.

```suggestion
```

functions/venue.js
```diff
@@ -514,6 +517,10 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.showLiveSchedule = data.showLiveSchedule;
   }
 
+  if (typeof data.type === "boolean") {
+    updated.type = data.type ? UNCLICKABLE_ROOM_TYPE : null;
+  }
+
```
0xdevalias, line 523: `type` is defined on `Room`, not `Venue`.

```suggestion
```

functions/venue.js
```diff
@@ -630,6 +637,10 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
...
```
0xdevalias, line 643: `type` is defined on `Room`, not `Venue`.

```suggestion
```

functions/venue.js
0xdevalias: From reading the rest of the code in this PR I can now understand that you chose to add a new boolean field to `Venue`, which isn't the solution here. `type` exists on `Room`, and is an enum of `RoomTypes`

```suggestion
```

0xdevalias (changes requested):
Comments/changes/etc.

src/types/venues.ts
TroySK: 🤦‍♂️

functions/venue.js
TroySK: Resolving this and the rest of them as the change-set has been reverted.

0xdevalias:
@TroySK it looks like you may have rebased this onto master/similar, we use staging as our default branch. That’s why there are now 50 odd unrelated commits here. Can you please fix that up.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -183,6 +183,24 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+          <Form.Row>
+            <div className="room-edit-modal__input">
+              <Form.Label>Change room type</Form.Label>
+              <Form.Control
+                as="select"
+                custom
+                name="type"
+                ref={register}
+                defaultValue={room.type}
+              >
```
0xdevalias, line 195: Why are we using `custom` here? (not saying it's wrong, just curious what it's used/needed for)

It seems like this may relate to the following?

- https://react-bootstrap.github.io/components/forms/#forms-custom-select

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -183,6 +183,24 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+                <option />
```
0xdevalias, line 196: I wonder if we should add a label like 'Default' or similar here, rather than just being an empty option? Might be better UX?

```suggestion
                <option value="">Default</option>
```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: Looking at https://react-bootstrap.github.io/components/forms/ it sounds like we should potentially be wrapping `Form.Label` and `Form.Control` within a `Form.Group` as well? I know that probably doesn't seem to be the existing pattern here, but how does it look when we do that? If it still looks fine, probably good to do (even if we only add it to this part and consider the others OOS for a future cleanup or similar)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -183,6 +183,24 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+                {Object.values(RoomTypes).map((roomType) => (
+                  <option key={roomType}>{roomType}</option>
+                ))}
```
0xdevalias, line 199: Do we also need to set the `selected` prop on one of these? Or does `Form.Control` handle that for us?

**Further Reading:**

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#attributes

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -183,6 +183,24 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
+              <Form.Label>Change room type</Form.Label>
+              <Form.Control
+                as="select"
+                custom
+                name="type"
+                ref={register}
+                defaultValue={room.type}
+              >
+                <option />
+                {Object.values(RoomTypes).map((roomType) => (
```
0xdevalias, line 197: This might be a useful thing to extract as a little helper in types/rooms. I can imagine we may want to do similar in other places in the app in future.

```typescript
export const allRoomTypes: string[] = () => Object.values(RoomTypes)
```

0xdevalias: I wonder, do we want to change the way the label of these is shown so it looks 'nicer' to a human viewing it? Either all lowercase, or even better maybe: Titlecase

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -313,6 +314,22 @@ const RoomInnerForm: React.FC<RoomInnerForm> = (props) => {
                       <span className="slider round"></span>
                     </label>
                   </div>
+                  <div className="toggle-room">
+                    <div className="input-title">Room Type</div>
+                    <Form.Control
```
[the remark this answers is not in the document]
0xdevalias: Instead of copy/paste duplicating the code between here and `src/pages/Admin/Room/Edit/RoomEditModal.tsx`, it would be better to extract it into a new 'mini component', and then use that in both places.

src/scss/global.scss
```diff
@@ -296,6 +296,13 @@ textarea {
   }
 }
 
+.custom-select {
+  background-color: rgba($white, 0.12);
+  color: $white;
+  margin-bottom: 0.8em;
+  border-radius: 22px;
+}
```
0xdevalias, line 304: I expect this relates to the `custom` prop on `Form.Control` in `src/pages/Admin/Room/Edit/RoomEditModal.tsx`/etc?

What defines this class name? Are we able to change it?

While this seems like a reasonable use case for including in `global.scss`, we're trying to minimize/reduce what is included here to make things more modular/understandable/etc.

It might be nice to add a comment explaining it's intent/where it applies/etc (eg. to `react-bootstrap`'s `Form.Control` when the `custom` prop is set) to make it easier for future engineers to know why it exists/etc.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: Also, I'm not sure how this will translate into the data we send to the backend, and ultimately how it gets stored. Currently though, the 'default case' in the backend should be that the room type doesn't exist (which will translate to `undefined` in the app code). We don't want to set it to an empty string, etc. So we might need to tweak some things to ensure that works as expected.

0xdevalias (changes requested):
Questions/Comments/Suggestions.

Please also make sure to fix the rebasing issue mentioned in https://github.com/sparkletown/sparkle/pull/1305#issuecomment-815276438

---

Are any changes required to the `api/` layer of functions (or similar, even if they aren't in that folder) to support adding these new fields? Or to the validations? Or to the backend functions? Etc?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: > Why are we using `custom` here? (not saying it's wrong, just curious what it's used/needed for)
> 
> It seems like this may relate to the following?
> 
> * https://react-bootstrap.github.io/components/forms/#forms-custom-select

Using the class `custom-select` to customize the look of the select. Styled it to look more like the input with a grey background. I can remove it if it requires design approval.

TroySK: > Form.Group

It looks different, in the attached image the lower one is nested in a Form.Group element. I would prefer to have the same convention across the file and do the markup cleanup in a different exercise. 
image

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: > I wonder if we should add a label like 'Default' or similar here, rather than just being an empty option? Might be better UX?

Using `<option>Default</option>` so that value is set to `undefined`.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
TroySK: Do we plan to keep Admin as well as Admin_v2?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: We should agree on a better way to handle this. If we follow this pattern, editing fields are going to be a pain. Would setting `null` work? Trying to figure out a convention which is easy to maintain.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: The component handles the setting of the selected prop.

src/scss/global.scss
TroySK: Yeah, makes sense. I'll add the comment.

The `custom` class name can be changed using the `bsCustomPrefix` prop. By default it uses Bootstrap's custom class name for select.
https://react-bootstrap.github.io/components/forms/#form-control-props

TroySK:
> @TroySK it looks like you may have rebased this onto master/similar, we use staging as our default branch. That’s why there are now 50 odd unrelated commits here now. Can you please fix that up.
> 
> (also, for future reference, we generally don’t rebase/force push in PRs once the review process has been started, as it makes it harder to see the changes since the last review was made. In this case, to fix it the unrelated commits, obviously a rebase/etc will be required though; that’s ok)
> 
> image



> @TroySK it looks like you may have rebased this onto master/similar, we use staging as our default branch. That’s why there are now 50 odd unrelated commits here now. Can you please fix that up.
> 
> (also, for future reference, we generally don’t rebase/force push in PRs once the review process has been started, as it makes it harder to see the changes since the last review was made. In this case, to fix it the unrelated commits, obviously a rebase/etc will be required though; that’s ok)
> 
> image

Sorry about this, I guess my muscle memory kicked in. Will fix one review is done.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
0xdevalias: Long term, no; but right now as we are transitioning and they don’t both support everything, no.

That said, duplication hurts us in the interim, and extracting sub-components is a better pattern than ending up with the 100’s+ line files our admin code currently has.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: Seemingly based on a comment you made elsewhere, removing the value=“” is all that’s required?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: Nice

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
TroySK: Sounds good. I'll refactor it.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: That is what I had expected but wasn't the case when I tested after posting the comment. I deleted the comment thereafter.

0xdevalias: Instead of just deleting the comment, you should correct what you said so that you’re communicating that new insight with others. That would have saved this back and forth; and me operating on incorrect information.

0xdevalias: > We should agree on a better way to handle this. If we follow this pattern, editing fields are going to be a pain. Would setting null work? Trying to figure out a convention which is easy to maintain.

Why are editing fields going to be a pain/etc if we do this? Would seem it would just be a couple of tiny changes to transform the data?

TroySK: The comment was deleted within seconds of posting it. My workflow does not involve reading review comments off Github. I realise that isn't the case with everyone here, will post updates going forward.

TroySK: > > We should agree on a better way to handle this. If we follow this pattern, editing fields are going to be a pain. Would setting null work? Trying to figure out a convention which is easy to maintain.
> 
> Why are editing fields going to be a pain/etc if we do this? Would seem it would just be a couple of tiny changes to transform the data?

Yes, the changes are trivial but they might introduce bugs later. The reasoning is; this field behaves differently than other fields or how developers perceive an app to work.
Editing a field, updates a value but changing that pattern here, we delete the key as well. This is an exception which may be overlooked by future devs. We may add comments and tests to have a guard against such bugs but it will be one more thing developers have to look throughout codebase for such exceptions.
Even on the access side, we will now have to check whether the key is defined before checking for match.

In SQL world, this would be equivalent to running a drop column every time we delete the value of a column. But we are in NoSQL land so maybe this is how it works.

IRL, will a room loose it's `type` if it does not have a `type` or does the room `type` become default? Answer to that question may help think about a solution too.

Happy to roll with a convention we pick which we can carry forward.

0xdevalias: The current default is that the ‘type’ key doesn’t exist, thus it is ‘undefined’ within the code when ‘read back’. The pattern that we tend to follow is that defaults aren’t defined in the database, they’re defined in the code. Handling this is a standard pattern, and quite trivial with TypeScripts ‘safe access’ checks (foo?.bar) and defaults (foo?.bar ?? “baz”)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: > I would prefer to have the same convention across the file and do the markup cleanup in a different exercise.

I'm ok with this, so long as we make sure to do it. Can you please create a backlog issue detailing that change to be made.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: > Yes, the changes are trivial but they might introduce bugs later. The reasoning is; this field behaves differently than other fields or how developers perceive an app to work.
Editing a field, updates a value but changing that pattern here, we delete the key as well. This is an exception which may be overlooked by future devs.

Further to my comments in https://github.com/sparkletown/sparkle/pull/1305#discussion_r611932752, we 'delete the key' only because firebase doesn't let us explicitly set `undefined` as a key that gets stored on the backend. This is a pattern we already use in other locations within the app.

src/scss/global.scss
0xdevalias: Makes sense. Are we intending to have this change applied to all `select`s across the app? Or is it intended for the specific changes being made here/in a generic control?

If the latter, it'd probably be good to give it a more 'localised' name, aligned to BEM patterns on that component:

- http://getbem.com/naming/

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: If this is a pattern which is being followed, happy to roll with it. Maybe we should document these nuances somewhere more explicitly so other developers joining in are aware of the conventions.

My personal preference is to use `null` and try and avoid safe checks as they add additional conditional checks leading slower benchmarks. More like pre-optimization; and we all know what that leads to! 😃

0xdevalias: There is a PR on the public repo titled documentation that captures a lot of these nuances. There are also backlogged issues about documentation, as well as the internal documentation repo itself. These are all a WIP, but if you find an area lacking, you’re more than welcome to create a new backlog issue for it and/or write up some of the existing ‘snapshots’ of how we do things more formally. We eventually want to get to this, but time is limited, and there have been higher priorities.

0xdevalias: The massive increase in engineer efficiency trumps the negligible small gains we would get by avoiding using safe accessors/etc.

Let’s not hyperfocus on ‘code gold’ optimisation that is unlikely to have any real world impact on users. Particularly when that reduces engineer efficiency and/or code maintainability.

TroySK: @0xdevalias Issues aren't the best place to document these as new engineer aren't sure what to search for in the issues. Wiki/Google Docs might be a better option.

Totally agree on the `code golf` optimisation doesn't make sense. My primary concern was around the pattern which has been addresses as thats the convention here.

0xdevalias: As I said above, they aren’t permanently documented in issues, the issues are there to capture the insights to be written up into proper documentation within our ‘internal docs’ repo, as per our standards.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
TroySK: @0xdevalias I am not sure where do I put the refactored component as I do not see it being reused anywhere once the legacy admin code is removed. Should this go in atoms or in as a component in `pages/Admin/Room`.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: I am removing the custom styling as it is out of scope for this change.
a682ab592733936af071015f3dae38aaf57cecb3

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: @0xdevalias I couldn't lookup an example where we remove keys in `/api`. Help me locate an example which I can follow.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: Not going with this change for now as React Bootstrap select doesn't use the value property in the traditional sense and uses keys and state to manage the values. It was resulting in option label being set as option value.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: The docs links above should show you how to do it. I’m not sure if removing the key is still in an ‘in review’ PR, probably is. Maybe look at the auditorium v2 one? Might be there.

You can probably also just search for Field Value or similar to find similar examples.

But basically, as I believe the linked docs describe, within an update, the ‘value’ is set to the special FieldValue entry that says to delete it.

This might be something that gets added on the backend function even, depending where you think it makes more sense to have it, though I believe that might require changes in our frontend validation if we do.

TroySK: @0xdevalias Please share a piece of code which shows the convention. I am not familiar enough with the codebase to dig it out.

0xdevalias: Does searching the codebase for the literal “FieldValue” (without the quotes; as mentioned above) not find it for you?

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
0xdevalias: @denisdimitrov @mike-lvov Do you have thoughts/opinions on @TroySK's question in https://github.com/sparkletown/sparkle/pull/1305#discussion_r612637836 ?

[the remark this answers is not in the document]
denisdimitrov: I think it would be fine to put the component in the atoms or molecules directory. IMO we should do this to all of the other components that are in pages/Admin but in another PR.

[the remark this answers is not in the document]
mike-lvov: > I think it would be fine to put the component in the atoms or molecules directory. IMO we should do this to all of the other components that are in pages/Admin but in another PR.

I would still expect that there are cases when pages/venues have their own components, which are not shared because they are quite specific. For example `ChatSidebar` has `PrivateChatPreview` component, which I don't think can be reused anywhere else. And if in the future there is a case, where we would actually want to reuse it, it's quite easy to just drag that component into `shared` components.

Thus, I would say that we should not just `move all of the components` into `atoms/molecules`.

@denisdimitrov do you think all of those components will be reused anywhere else except for Admin page?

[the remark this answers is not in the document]
0xdevalias: I’m still not convinced that the molecules/atoms/etc pattern is cleaner/better than the ‘feature’ pattern. Maybe for generic things, but it kind of feels like it takes more cognitive effort and ends up being somewhat arbitrary anyway.

In past projects I’ve stuck more to a ‘feature’ based folder structure, and then within that would sometimes break it up further by components/etc folders. We’ve vaguely used this pattern in a couple of places already for these more ‘private’ components.

—-

I definitely think we should be splitting up the admin stuff way more than it is (particularly the files that have tons of components in them), but I’d be inclined to think of it more in this ‘private component folder’ layout. It’d also be nice to have an explicit distinction between admin v1 and v2 in those folders. IMO that would logically lend itself more towards how we (or at least iI) tend to think about the app and it’s logical dependencies; and sort of aligns closer to the various routes/‘areas’ of the app.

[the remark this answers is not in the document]
0xdevalias: (This stuff is probably OOS to be discussing on this PR thread, and should probably be copied into a new backlog issue for further discussion/refinement/to assess our patterns/etc)

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: No, I didn't. Most of them were array operations, none of them were about deleting a key from an object before adding back to an array.

0xdevalias: @mike-lvov Are you able to help out here by chance? Not sure I’m going to be able to explain better than I already have + the docs/references I’ve already included. Perhaps you’re able to do so better..

mike-lvov: @TroySK 

`export const setGridData = async ({
  venueId,
  userId,
  gridData,
}: SetGridDataProps): Promise<void> => {
  const userProfileRef = firebase.firestore().collection("users").doc(userId);

  const newGridData = {
    [`data.${venueId}`]: gridData ?? firebase.firestore.FieldValue.delete(),
  };

  return userProfileRef.update(newGridData).catch((err) => {
    Bugsnag.notify(err, (event) => {
      event.addMetadata("context", {
        location: "api/profile::setGridData",
        venueId,
        userId,
        gridData,
      });
    });
  });
};
`

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
TroySK: Okay, I will refactor it for atoms for now and file a new issue to look into it at a later stage

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: Thanks @mike-lvov! This is close but not exactly what I was looking for. Please review https://github.com/sparkletown/sparkle/pull/1305/commits/2c3bf9a8656a0c591ab4774e50ad1a9e8fb16e31. This is what I wanted to avoid as this kind of special cases lead to bugs in the future.

functions/venue.js
```diff
@@ -268,6 +268,19 @@ const removeAdmin = async (adminId) => {
...
+ *
+ * @param {Room} room
+ */
+const buildRoom = (room) => {
+  const { type, ...newRoom } = room;
+  if (type) {
+    return type ? room : newRoom;
+  } else {
+    return newRoom;
+  }
```
0xdevalias, line 281: Why is there an `if` statement and a ternary that seem to be doing the exact same thing here?

0xdevalias: And what is the purpose of this function in general? The description doesn't really explain what it's doing/why, and i'm struggling to decipher the intent/need for it from the logic as well.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { useEffect, useState, useCallback, useMemo } from "react";
+import { startCase } from "lodash";
```
0xdevalias, line 2:
```suggestion
import startCase from "lodash/startCase";
```

src/pages/Admin/Room/Edit/RoomEditModal.tsx
```diff
@@ -183,6 +184,25 @@ export const RoomEditModal: React.FC<RoomEditModalProps> = ({
...
           </Form.Row>
 
+          <Form.Row>
+            <div className="room-edit-modal__input">
+              <Form.Label>Change room type</Form.Label>
+              <Form.Control
+                as="select"
+                name="type"
+                ref={register}
+                defaultValue={room.type}
```
0xdevalias, line 194: Will this correctly select the default value when `room.type === undefined`? Or should we have `?? ""` here?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: @TroySK You 👍🏻 'd my suggestion in https://github.com/sparkletown/sparkle/pull/1305/files#r609141461 yet don't seem to have made any code changes for it?

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useState, useMemo, useCallback } from "react";
 import Bugsnag from "@bugsnag/js";
+import { startCase } from "lodash";
```
0xdevalias, line 3:
```suggestion
import startCase from "lodash/startCase";
```

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
0xdevalias: @TroySK This still seems to be outstanding?

0xdevalias (changes requested):
Still seem to be outstanding changes required from previous reviews/etc?

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: @TroySK I don't understand the purpose/intent of the code in the commit you linked, and how that relates to the above? What @mike-lvov shared was what I was talking about in my original suggestions here.

TroySK: The pattern @mike-lvov attached doesn't work for what I am trying to solve. It could be my knowledge of TypeScript is lacking. `room`s are array and inserted using an upsert. The `type` key on the room object needs to be removed before adding it back to the array if it is null. Not sure how null coalescing can be used here. Are you suggesting I use `??` and use `delete room.type`? I selected this alternative as it doesn't mutate the object.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: Will be doing the change once I am able to implement https://github.com/sparkletown/sparkle/pull/1305#discussion_r617310471 correctly as I plan to use the method there.

functions/venue.js
TroySK: Just ternary operator should work. Missed cleaning up the if construct.

The function is named `buildRoom` as it builds the `room` object. It removes the `type` attribute if it is empty. I can add a comment as to why we are doing this to make it clear for posterity.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
TroySK: Yes, if value is null or undefined, the default option is selected.

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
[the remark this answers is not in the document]
TroySK: Yes, will make the change once https://github.com/sparkletown/sparkle/pull/1305#discussion_r617310471 gets done. Hoping to use that to generate the options.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: Cool, makes sense.

functions/venue.js
0xdevalias: > It removes the type attribute if it is empty.

Fair enough, this makes sense. The naming 'buildRoom' seems strange given the intent seems to be just to remove the type attribute when empty.

src/pages/Admin/Room/Edit/RoomEditModal.tsx
0xdevalias: > Are you suggesting I use `??` and use `delete room.type`? I selected this alternative as it doesn't mutate the object.

Assuming I am understanding what you're saying here correctly, that was what I was suggesting, yeah.

From @mike-lvov's screenshot above, this bit:

image

Use the data if we have data, or if it's `undefined` then we tell firebase to delete the field.

PR #1306, opened 2021-04-06
add generic input component (with support for icons)

sunny-viktoryia:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/615

`InputField` component is used now as:
* search input in the private chat
* nav search input

Please, note that the view of the search button on navigation panes is a bit changed as we're going to change the styling of this input anyway based on the https://github.com/sparkletown/internal-sparkle-issues/issues/451:
1. search icon is always visible ( [not] hover / [not] focus)
2. left padding is increased to give search icon more space
3. text-align: left

(but if needed I can add more styling to nav search input to make it look as it was before)


src/components/atoms/IconInput/IconInput.scss
```diff
@@ -0,0 +1,41 @@
+@import "scss/constants";
+
+.icon-input {
+  position: relative;
+
+  &__input {
+    height: 100%;
+
+    &--icon-from-left {
```
0xdevalias, line 9:
```suggestion
    &--icon-left {
```

src/components/atoms/IconInput/IconInput.scss
```diff
@@ -0,0 +1,41 @@
...
+      padding-left: 2.5rem;
+    }
+
+    &--icon-from-right {
```
0xdevalias, line 13:
```suggestion
    &--icon-right {
```

src/components/atoms/IconInput/IconInput.scss
0xdevalias: It might be better to move these variants onto the base class itself, so that they are:

- `.icon-input--icon-left`
- `.icon-input--icon-right`

or similar.

src/components/atoms/IconInput/IconInput.scss
0xdevalias: It might be better to move these variants onto the base class itself, so that they are:

- `.icon-input--icon-left`
- `.icon-input--icon-right`

or similar.

src/components/atoms/IconInput/IconInput.scss
```diff
@@ -0,0 +1,41 @@
...
+    z-index: z(input-icon);
+    opacity: 0.2;
+
+    &--left {
+      left: 0;
+    }
+
+    &--right {
+      right: 0;
+    }
```
0xdevalias, line 39: We could potentially move these variants onto the base class itself, so that they are:

- `.icon-input--icon-left`
- `.icon-input--icon-right`

or similar. Not sure if that is better or worse than how it is now though (@mike-lvov any thoughts?)

Then these would end up being written as something like (in the 'root' of the main class):

```scss
&--icon-left {
  .icon-input__icon {
    left: 0;
  }
}

&--icon-right {
  .icon-input__icon {
    right: 0;
  }
}
```

This would also allow any of the other style changes related to the variant to be 'grouped together' in this 'single location', and then we would only need a single check in the `classNames` helper to set the variant on the root component's classes.

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  autoComplete?: string;
+  placeholder?: string;
+  iconPosition?: "right" | "left";
+}
+
+const ICON_INPUT_CLASS_NAME = "icon-input";
+const ICON_CLASS_NAME = `${ICON_INPUT_CLASS_NAME}__icon`;
+const INPUT_CLASS_NAME = `${ICON_INPUT_CLASS_NAME}__input`;
+
+export const IconInput: React.FC<IconInputProps> = ({
```
0xdevalias, line 21: I think it might be better to name this something like `Input` or `InputField` or similar rather than `IconInput`. The fact that it can support icons doesn't feel like it needs to be in the component name itself, just having props that support that should be sufficient.

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  value,
+  autoComplete = "off",
+  placeholder,
+  iconPosition = "left",
+}) => {
+  const wrapperClassNames = classNames(ICON_INPUT_CLASS_NAME, className);
+  const iconClassNames = classNames(
+    ICON_CLASS_NAME,
+    iconPosition === "right" && `${ICON_CLASS_NAME}--right`
+  );
```
0xdevalias, line 34: What about `--left`?

```suggestion
  const iconClassNames = classNames(
    ICON_CLASS_NAME,
    {
      `${ICON_CLASS_NAME}--right`: iconPosition === "right",
      `${ICON_CLASS_NAME}--left`: iconPosition === "left",
    }
  );
```

**Further Reading:**

- https://github.com/JedWatson/classnames#usage

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onChange?: (e?: ChangeEvent<HTMLInputElement>) => void;
+  value?: string;
+  autoComplete?: string;
+  placeholder?: string;
+  iconPosition?: "right" | "left";
+}
+
+const ICON_INPUT_CLASS_NAME = "icon-input";
+const ICON_CLASS_NAME = `${ICON_INPUT_CLASS_NAME}__icon`;
+const INPUT_CLASS_NAME = `${ICON_INPUT_CLASS_NAME}__input`;
```
0xdevalias, line 19: While I understand and see what you were going for here, I feel like this extra 'indirection' here makes it the code harder to read (and search for) than just manually constructing the strings below. Since they'll only really be passed into the couple of `classNames` calls, I don't feel like the duplication/etc will be a big deal. It's not like we're scattering them across the codebase.

That will likely also make it easier/more likely for IDE/etc static code checkers to correctly be able to hint on these classes/allow to click through to their definitions, etc.

```suggestion
```

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { ChangeEvent } from "react";
```
0xdevalias, line 1: @mike-lvov @mcflyok etc: Do you know if this counts as an `atom`? Or should it be a `molecule` or something else? (i'm still not perfectly sure what should be where)

- `src/components/atoms/IconInput/IconInput.tsx`

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  const wrapperClassNames = classNames(ICON_INPUT_CLASS_NAME, className);
+  const iconClassNames = classNames(
+    ICON_CLASS_NAME,
+    iconPosition === "right" && `${ICON_CLASS_NAME}--right`
+  );
+  const inputClassNames = classNames(
+    INPUT_CLASS_NAME,
+    `${INPUT_CLASS_NAME}--icon-from-${iconPosition}`
+  );
```
0xdevalias, line 38: Since `iconPosition` is allowed to be `undefined`, this won't work correctly. This should though:

```suggestion
  const inputClassNames = classNames(
    INPUT_CLASS_NAME,
    {
      `${INPUT_CLASS_NAME}--icon-right`: iconPosition === "right",
      `${INPUT_CLASS_NAME}--icon-left`: iconPosition === "left",
    }
  );
```

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <div className={wrapperClassNames}>
+      <input
+        className={inputClassNames}
+        placeholder={placeholder}
+        value={value}
+        onChange={onChange}
+        autoComplete={autoComplete}
+      />
```
0xdevalias, line 48: We may want to be able to pass `ref` to this input so that we can use it with `react-hook-form` / `useForm`s `register` function and similar.

This might be useful reading for this:

- https://reactjs.org/docs/forwarding-refs.html

0xdevalias: It might also be nice to be able to pass any `extraProps` through to this without them necessarily being explicitly defined by name in `IconInputProps`. We can use the `{...extraInputProps}` syntax to add that here (and would also need a `...extraInputProps` as the last value in the function props definition) Not sure what the exact typing will be to represent that in `IconInputProps`, but it should be relatively easy to lookup (or if there is a specific type for 'all the things that can be on an input field', that would be a useful type to use there maybe)

That'll save us having to explicitly add every single prop we would ever want to be able to pass through to this component individually.

src/components/atoms/IconInput/IconInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface IconInputProps {
+  className?: string;
+  icon?: IconProp;
+  onChange?: (e?: ChangeEvent<HTMLInputElement>) => void;
+  value?: string;
+  autoComplete?: string;
+  placeholder?: string;
+  iconPosition?: "right" | "left";
+}
```
0xdevalias, line 15: We might want to split this up into `BaseIconInputProps` and then 2 variants (one for when there is an icon supplied, one for when no icon supplied) so that we can properly hint through the types that `iconPosition` is only going to be useful when `icon` is set, etc.

src/components/atoms/IconInput/IconInput.scss
```diff
@@ -0,0 +1,41 @@
...
+      opacity: 0.6;
+    }
+  }
+
+  &__icon {
+    display: flex;
+    position: absolute;
+    top: 0;
+    height: 100%;
+    width: 40px;
```
0xdevalias, line 27: How will this work with responsive sizing? Do we always want this to be `40px`?

src/components/atoms/IconInput/IconInput.scss
0xdevalias: We potentially could even drop the word 'icon' from this, and just have the variants as `--left` and `--right`?

0xdevalias (commented):
[Partial Review] Only looked at:

- `src/components/atoms/IconInput/IconInput.scss`
- `src/components/atoms/IconInput/IconInput.tsx`

src/components/atoms/IconInput/IconInput.tsx
mike-lvov: I think basic components: Button/Input/Text should be considered `atoms`, even if they have some additional logic inside

src/components/atoms/IconInput/IconInput.tsx
mike-lvov: Alternatively, and I think a better way to handle this would be to have 2 props:
`startIcon/endIcon`similar to what material-ui lib has: 
https://material-ui.com/components/text-fields/

src/components/atoms/IconInput/IconInput.tsx
0xdevalias: _nods_ cool, works for me.

src/components/atoms/IconInput/IconInput.tsx
0xdevalias: I'm ok with that. Avoids the potential runtime issues + is a cognitively simpler concept to grasp.

src/components/atoms/IconInput/IconInput.tsx
sunny-viktoryia: > this extra 'indirection' here makes it the code harder to read (and search for)

yeah, you're right. I'll rewrite that.

<hr />

btw, if we want to use BEM class names, we can consider this package which generates class names in the needed format
https://www.npmjs.com/package/b_

or do you think it'll be better / easier to build them manually?

0xdevalias: Since that link seemed to break GitHub's autolinker: [https://github.com/azproduction/b_](https://github.com/azproduction/b_)

Interesting concept.. i've never ever thought to use something like that before tbh.

Also came across https://github.com/ybiquitous/bem-ts which seems to have been updated more recently/might be better maintained.

---

I think these still probably don't add a huge amount of gain/productivity, and might break IDE features/make searching for the classes harder, etc. Given we'd also need to separately write the BEM parts in the SCSS too, we're probably still going to have to think about the 'structure' of it, etc.

There appear to be some eslint/stylelint related plugins for detecting/enforcing BEM patterns, my plans were to enable them at some point to make it a less manual process of 'ensuring we get it right'.

So i'd say probably not worth adding these libs here.

src/components/atoms/IconInput/IconInput.tsx
sunny-viktoryia: @mike-lvov 
so `startIcon` / `endIcon` would be booleans and stand for position of the `icon`?
or
you suggest to replace `icon` prop by these 2 props and specify icon src there?

mike-lvov: They should accept both iconProp + rendered icon(to cover the case, when you want to add an onClick handler to an icon)

mike-lvov: For example this case should be covered: 
image

sunny-viktoryia: so potentially, we can have both startIcon and endIcon for the input, right?

mike-lvov: @sunny-viktoryia yes. Plus, a developer should be able to pass an icon component from `@fortawesome/free-solid-svg-icons`. For ex. `faSearch` 

or he can pass a rendered Icon 
              ```
endIcon={<FontAwesomeIcon
                  className="chat-sidebar__controller__second-icon"
                  icon={faCommentDots}
                  size="lg"
                />}|
```

File: src/components/atoms/IconInput/IconInput.scss (hunk 2 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.scss]: renamed the classNames, As we decided to use `iconStart` and `iconEnd` prop names, I assumed that it'd be more logical to give class names similar names:
* `input-field--icon-start`
* `input-field--icon-end`

commit: https://github.com/sparkletown/sparkle/pull/1306/commits/4522f27ae790a9bc30e71c2e4e06433f7e2b9141

File: src/components/atoms/IconInput/IconInput.scss (hunk 3 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.scss]: added `--icon-star` and `--icon-end` modificators to the base `input-field` class. 

commit: https://github.com/sparkletown/sparkle/pull/1306/commits/4522f27ae790a9bc30e71c2e4e06433f7e2b9141

File: src/components/atoms/IconInput/IconInput.tsx (hunk 4 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.tsx]: renamed: IconInput -> InputField

commit: https://github.com/sparkletown/sparkle/pull/1306/commits/c0acf340a2ba41cb9b7fa7425726ace7838eccf2

File: src/components/atoms/IconInput/IconInput.tsx (hunk 9 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.tsx]: forwarding Ref: https://github.com/sparkletown/sparkle/pull/1306/commits/c137aec001cc7119da4274f039102d1d65e2c96b
applying extra props: https://github.com/sparkletown/sparkle/pull/1306/commits/4522f27ae790a9bc30e71c2e4e06433f7e2b9141
and renaming inputProps -> extraInputProps: https://github.com/sparkletown/sparkle/pull/1306/commits/2d98fa288fa22220801a9a7f3765e88ec04a549c

File: src/components/atoms/IconInput/IconInput.tsx (hunk 10 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.tsx]: added support for both side icons (`IconProp` or `JSX.Element`): https://github.com/sparkletown/sparkle/pull/1306/commits/4522f27ae790a9bc30e71c2e4e06433f7e2b9141

File: src/components/atoms/IconInput/IconInput.scss (hunk 11 above)
Comment (sunny-viktoryia) [reply src/components/atoms/IconInput/IconInput.scss]: a good question.
I think we need @mathieud here. I can show how it looks like right now.  There are 2 input fields where I'm using my `InputField` component: 

image

the width of the icon (as well as the value of input padding ) is 2.5rem for both cases

In my opinion, if there is another size needed you can always add your custom class to `InputField` component and assign the width you need for a particular input.

What do you think? Is that okay with you?

File: src/components/atoms/InputField/InputField.scss [hunk 12]
```diff
@@ -0,0 +1,51 @@
...
+      padding-right: $icon-width;
+    }
+  }
+
+  &__input {
+    height: 100%;
+
+    &:focus + .input-field__icon > * {
+      opacity: 0.6;
+    }
```
mike-lvov, line 25: I don't really understand the logic of this.
But is it possible to not use the wildcard selector(> *)?
Given that we assign opacity, you can assign it to a parent(and it will affect he children)

src/components/atoms/InputField/InputField.scss
```diff
@@ -0,0 +1,51 @@
...
+
+    &--end {
+      right: 0;
+    }
+
+    & > * {
+      opacity: 0.2;
+      width: 100%;
+    }
```
mike-lvov, line 49: same as above ^

src/components/atoms/InputField/InputField.scss
```diff
@@ -0,0 +1,51 @@
...
+
+$icon-width: 2.5rem;
+
+.input-field {
+  position: relative;
+
+  &--icon-start {
+    input {
+      padding-left: $icon-width;
+    }
```
mike-lvov, line 11: can you please assign styles to classes, not tags? It makes the behavior of the component more predictable.

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { forwardRef } from "react";
+import classNames from "classnames";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+import { isTruthy } from "utils/types";
+import "./InputField.scss";
```
mike-lvov, line 6:
```suggestion
import React, { forwardRef } from "react";
import classNames from "classnames";
import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
import { IconProp } from "@fortawesome/fontawesome-svg-core";

import { isTruthy } from "utils/types";

import "./InputField.scss";
```

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export interface InputFieldProps extends React.HTMLProps<HTMLInputElement> {
+  iconStart?: IconProp | JSX.Element | false;
+  iconEnd?: IconProp | JSX.Element | false;
+}
```
mike-lvov, line 11:
```suggestion
export interface InputFieldProps extends React.HTMLProps<HTMLInputElement> {
  iconStart?: IconProp | JSX.Element;
  iconEnd?: IconProp | JSX.Element;
}
```

Why do we need to have false here?

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const isJsxElement = (
+  component: IconProp | JSX.Element
+): component is JSX.Element => {
+  return isTruthy(component?.hasOwnProperty("props"));
+};
+
+const renderIcon = (
+  icon: IconProp | JSX.Element,
+  position: string
```
mike-lvov, line 24: Since you are only using `position` as a way to generate a class, wouldn't it be better just to pass a classname?

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -99,18 +99,13 @@ export const PrivateChats: React.FC<PrivateChatsProps> = ({
...
-          <FontAwesomeIcon icon={faSearch} size="1x" />
-        </div>
-      </div>
+      <InputField
+        className="private-chats__search"
+        placeholder="Search for online people"
+        value={userSearchQuery}
+        onChange={onInputChange}
+        iconStart={faSearch}
+      />
```
mike-lvov, line 108:
```suggestion
      <InputField
        className="private-chats__search"
        placeholder="Search for online people"
        value={userSearchQuery}
        onChange={onInputChange}
        iconStart={faSearch}
        autoComplete="off"
      />
```

mike-lvov (commented):
[OOS] Could you please create an issue to replace all of the inputs across the codebase with this new one?

src/components/atoms/InputField/InputField.tsx
sunny-viktoryia: I included `false` when I started to use `InputField` in `NavSearchBar`. 

```tsx
<InputField
...
  iconEnd={
    isTruthy(searchQuery) && (
      <img
        className="nav__clear-search"
        src="/icons/nav-dropdown-close.png"
        alt="close button"
        onClick={clearSearchQuery}
      />
    )
  }
/>  
```

I supposed that it'd be helpful to have `false` for such cases. If we remove it from icon type alias we'll have to write something like this: 

```tsx
<InputField
    iconEnd={
      isTruthy(searchQuery) ?
        <img
          className="nav__clear-search"
          src="/icons/nav-dropdown-close.png"
          alt="close button"
          onClick={clearSearchQuery}
        />
      : undefined
    }
  />
```

But sure, we can go without it, if you think that it's not needed. or maybe you have another idea how to pass this props in a better way?

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: I can. :ok_hand: 
I did that because I wanted the selector to be less specified so that it could be easier to overwrite if needed. 
`className + tagName` has lower specificity than `className + className`.

But since `className + tagName` can't be overwritten by a single `className`, and we'll need a `className` or a `tagName` in addition anyway, it seems it doesn't make much sense to think about it here.

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: I used wildcard selector for custom icon elements which we can pass as `iconStart`/`iconEnd`. Let's say we use img-tag as `iconStart`:

```tsx
iconStart={<img
  className="nav__clear-search"
  src="/icons/nav-dropdown-close.png"
  alt="close button"
  onClick={clearSearchQuery}
/>}
```

so we can now overwrite opacity for this icon and add its own styles using `.nav__clear-search` selector. But if we assign opacity to parent element instead, we'll have to go inside `InputField` component searching for the className (`.input-field__icon`) that we need to overwrite. I thought that it may be uncomfortable to do that every time and also, it produces selector `.input-field__icon` which can't be found directly on the page where we use `InputField`, so someone can accidentally remove it thinking that it's not needed here.

so what do you think? should still I remove wildcard selector?

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: please, check my comment above. If we decide to remove `*` for opacity, I'll do that here as well.

hmm :thinking:, and about `width`. I think we need `*` for width property as we don't know anything about custom icons which we can assign to `iconStart`/`iconEnd` (nor className, nor tag), but we know that its width shouldn't be bigger than input-padding. Although, I should better use `max-width` property instead here.

src/components/atoms/InputField/InputField.tsx
sunny-viktoryia: fixed: https://github.com/sparkletown/sparkle/pull/1306/commits/2778ea8bf7b45b4016fea94b850bae5e0a449c92

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: fixed: https://github.com/sparkletown/sparkle/pull/1306/commits/70022e4294d880248ca128366bff475348bf4ab8

src/components/atoms/InputField/InputField.tsx
sunny-viktoryia: updated: https://github.com/sparkletown/sparkle/pull/1306/commits/2778ea8bf7b45b4016fea94b850bae5e0a449c92

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
sunny-viktoryia: fixed: https://github.com/sparkletown/sparkle/pull/1306/commits/42720db4f38ca39001ce9d5fbedaebc7a2d35df2

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: change proprty from `width` to `max-width`: https://github.com/sparkletown/sparkle/pull/1306/commits/70022e4294d880248ca128366bff475348bf4ab8

src/components/atoms/InputField/InputField.tsx
mike-lvov: I see.
Let's stick to the second option `isTruthy ? icon : undefined`. it looks a bit more explicit

src/components/atoms/InputField/InputField.scss
mike-lvov: I see. Makes sense, thanks for explaining

src/components/atoms/InputField/InputField.scss
mike-lvov: I got the logic. Makes sense

src/components/atoms/InputField/InputField.tsx
sunny-viktoryia: updated here: https://github.com/sparkletown/sparkle/pull/1306/commits/13040f8da655ed6b3a8ed7e889e8411f1d9da6f3

mike-lvov (approved):
lgtm

src/components/atoms/InputField/InputField.tsx
0xdevalias: Is JSX.Element the correct ‘catch all’ type for ‘things that can be rendered’, or is there a better one? I remember there was one that handles everything, but can’t remember it’s name.

src/components/atoms/InputField/InputField.tsx
0xdevalias: @mike-lvov Why do you say that? The fact that it’s used in the class names feels like an irrelevant internal detail that shouldn’t need to be exposed in the ‘public API’ of this component.

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div ref={ref} className={wrapperClassNames}>
+      <input className="input-field__input" {...extraInputProps} />
+
+      {iconStart && renderIcon(iconStart, "input-field__icon--start")}
+      {iconEnd && renderIcon(iconEnd, "input-field__icon--end")}
+    </div>
+  );
+};
+
+export default forwardRef(InputField);
```
0xdevalias, line 61: I thought I commented on this somewhere, but can't find it, so perhaps I thought about doing so but forgot.

We prefer using 'named exports' rather than '`default` exports'.

I can see that you 'renamed' it as a 'named export' in `src/components/atoms/InputField/index.ts`, but it would be preferable to also use named exports here.

We could either inline the `forwardRef` call directly where we assign the `const`, or alternatively could prefix the 'raw component const' with `_` (eg. `_ InputField`), then export the 'wrapped const' as `InputField`

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+
+import { isTruthy } from "utils/types";
+
+import "./InputField.scss";
+
+export interface InputFieldProps extends React.HTMLProps<HTMLInputElement> {
+  iconStart?: IconProp | JSX.Element;
+  iconEnd?: IconProp | JSX.Element;
+}
```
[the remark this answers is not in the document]
0xdevalias: Also, I think we need to add typing to support passing through the `extraProps` here.

I believe this is called 'variadic' paramaters.

**Further Reading:**

- https://stackoverflow.com/questions/12699058/typescript-types-for-variadic-functions
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types
- https://www.infoq.com/news/2020/08/typescript-4-variadic-tuple-type/

---

Looking at the types we can pass to the `input` itself:

Screen Shot 2021-04-12 at 1 08 42 pm

It's defined in `node_modules/@types/react/index.d.ts:3021` as:

```typescript
React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>;
```

With the most relevant aspect seeming to be defined in `node_modules/@types/react/index.d.ts:2067`

```typescript
interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
  // ..snip..
}
```

So I believe the type we would want to use for this would be: `React.InputHTMLAttributes<HTMLInputElement>`

I'm not sure if that is possible using the variadic parameters.. so perhaps we would just need to make it an explicitly named parameter (eg. `extraInputProps`) in the types/etc as well.

src/components/atoms/InputField/InputField.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div className={classNames("input-field__icon", modifiedClassName)}>
+      {iconComponent}
+    </div>
+  );
+};
+
+const InputField: React.ForwardRefRenderFunction<
+  HTMLDivElement,
+  InputFieldProps
+> = ({ className, iconStart, iconEnd, ...extraInputProps }, ref) => {
```
0xdevalias, line 41: As per the other comment in this review, I would probably just have added `forwardRef` directly on this line as well, rather than including it as a `default` export below.

src/components/molecules/NavSearchBar/NavSearchBar.scss
```diff
@@ -4,69 +4,26 @@
...
-    background-image: url(/icons/nav-link-search-icon.png);
-    background-size: 20px;
-    opacity: 0.4;
-  }
-
-  .nav-search-close-icon {
-    position: absolute;
-    align-self: center;
-    z-index: z(navbar-nav-search-close-icon);
+  .nav__clear-search {
```
0xdevalias, line 7: For the scss classes with names changed here, did we search across the entire codebase to ensure they aren't used/overridden anywhere else (and also update them there)?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -14,9 +14,9 @@ import { useProfileModalControls } from "hooks/useProfileModalControls";
 
 import { RoomModal } from "components/templates/PartyMap/components";
 
-import { NavSearchBarInput } from "./NavSearchBarInput";
-
 import "./NavSearchBar.scss";
+import { InputField } from "components/atoms/InputField";
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
```
0xdevalias, line 19: These imports should be sorted differently (3rd party import at the top, `components/atoms/*` below `components/templates/*`

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -26,6 +26,10 @@ interface SearchResult {
 
 const NavSearchBar = () => {
   const [searchQuery, setSearchQuery] = useState("");
+  const onSearchInputChange = useCallback(
+    (e) => setSearchQuery(e.target.value),
+    []
+  );
```
0xdevalias, line 32: It might be useful to add a debounce or similar here, rather than having it update the state on every single keypress.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -84,14 +88,26 @@ const NavSearchBar = () => {
...
   }, []);
 
+  const clearSearchIcon = (
+    <img
+      className="nav__clear-search"
+      src="/icons/nav-dropdown-close.png"
+      alt="close button"
+      onClick={clearSearchQuery}
+    />
+  );
```
0xdevalias, line 98: I would write this as a sub-component that takes no props, rather than just a `const`, that way it can be used within the JSX as normal.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: Also, why is `e` not typed here anymore?

In `src/components/molecules/NavSearchBar/NavSearchBarInput.tsx` it was typed as `e: ChangeEvent<HTMLInputElement>`

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.scss
```diff
@@ -7,40 +7,19 @@
...
+    & input {
+      background-color: $black;
+      text-align: left;
+      font-size: 0.875rem;
+
+      &::placeholder {
+        color: $white;
+        opacity: 0.2;
+      }
     }
```
0xdevalias, line 22: I believe we/BEM prefer using class names for styling, rather than directly styling html elements.

src/scss/constants.scss
```diff
@@ -102,7 +100,7 @@ $z-layers: (
   footer: $z-layer-sidebar,
   global-profile-icon: 1,
   user-avatar-status-dot: 1,
-  private-chat-input-search-icon: 1,
+  input-icon: 1,
```
0xdevalias, line 103: This should probably be `input-field-icon` or similar, and be included under a `// InputField` comment or similar.

0xdevalias (commented):
[Retroactive Review] Comments/changes/etc

src/components/atoms/InputField/InputField.scss
```diff
@@ -0,0 +1,51 @@
...
+    width: $icon-width;
+    z-index: z(input-icon);
+
+    &--start {
+      left: 0;
+    }
+
+    &--end {
+      right: 0;
+    }
```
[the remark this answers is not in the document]
sunny-viktoryia: I'm not sure that would work. 
The logic is the following: if input-field has start/end icon prop, it achieves `input-field--icon-start` / `input-field--icon-end` className accordingly.

```scss
.input-field {
  &--icon-start {
    .input-field__icon {
      left: 0;
    }
  }

  &--icon-end {
    .input-field__icon {
      right: 0;
    }
  }
}
```

As we have the ability to add both icons (start and end), it means that parent div will have both classes, and the property `left: 0` will be ignored, as the icon can't be on the left and right sides simultaneously and both icons will appear on the right side.

or did you mean it differently?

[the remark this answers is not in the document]
0xdevalias: Yup, that’s a fair point. The way we are using the start/end icons now aren’t really —variants anymore, they are more like 2 distinct __elements; so that’s why my method here wouldn’t work.

Perhaps we should change the __element part so that they distinctly reference start/end icons.

PR #1307, opened 2021-04-06
Set 12 tables as default at Jazzbar

sunny-viktoryia:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/625

### DIFF
--- a/src/components/templates/Jazzbar/JazzTab/constants.ts
+++ b/src/components/templates/Jazzbar/JazzTab/constants.ts
@@ -1,74 +1,13 @@
 import { Table } from "types/Table";
 
-export const JAZZBAR_TABLES: Table[] = [
-  {
-    capacity: 6,
-    title: "Table 1",
-    reference: "jazzbar-table-1",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 2",
-    reference: "jazzbar-table-2",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 3",
-    reference: "jazzbar-table-3",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 4",
-    reference: "jazzbar-table-4",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 5",
-    reference: "jazzbar-table-5",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 6",
-    reference: "jazzbar-table-6",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 7",
-    reference: "jazzbar-table-7",
-    rows: 2,
-    columns: 3,
-  },
-  {
-    capacity: 6,
-    title: "Table 8",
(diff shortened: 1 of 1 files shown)


mike-lvov (approved):
LGTM

src/components/templates/Jazzbar/JazzTab/constants.ts
```diff
@@ -1,74 +1,13 @@
...
+const jazzBarTable = (index: number): Table => ({
+  capacity: 6,
+  title: `Table ${index}`,
+  reference: `jazzbar-table-${index}`,
+  rows: 2,
+  columns: 3,
+});
+
+export const JAZZBAR_TABLES: Table[] = new Array(12)
+  .fill(0)
```
0xdevalias, line 12: Why do we need the .fill(0) here if we’re only using the indexes?

mike-lvov: Array(number) will create an array with empty elements, which are not mappable

0xdevalias: Yup, fair enough.

I was thinking of how we do it in `src/components/templates/PartyMap/components/Map/Map.tsx`, but seemingly that is slightly different again:

```typescript
Array.from(Array(totalRows))
```

0xdevalias: Not that it matters for performance here, but I was curious:

https://jsbench.me/

image

The method used here (`.fill(0)`) is ~70% slower than using Array.from(Array(12))

sunny-viktoryia: @0xdevalias 
thanks for the service! I've never heard about it.

I got curious too. :) By the way, do you know why we can have different results? :see_no_evil: 
image

0xdevalias: > thanks for the service! I've never heard about it.

No worries. It's neat/useful when i'm trying to figure between a few choices of ways to implement something, take the guesswork out of it and lets me base the decision on the data.

> By the way, do you know why we can have different results? 🙈

😮 huh.. that's interesting. What browser/OS version are you using out of curiosity? I wonder if there is an optimisation or something in mine that isn't in yours, or vice versa.

I'm using `Google Chrome Version 89.0.4389.90 (Official Build) (x86_64)` on `macOS Big Sur 11.2.3 (20D91)`

I tried testing it in Safari, but that benchmark site seemed to not work for me.

This one did though, and suggested your code was faster: https://jsben.ch/RSVLT

image

Whereas that same test on Chrome suggests mine was:

image

So yeah, must just come down to the different implementations in different browsers and such. For something like this, the 'real world' performance difference for a 'run once' thing like this that it's irrelevant for anything more than intellectual curiosity. But still, it's interesting to see such a simple case show such a 'dramatic' difference between browsers.

sunny-viktoryia: I tested that on `Mozilla Firefox for Fedora (version 85.0)`.
Yeah, seems it depends on browser mainly. 

Running your test https://jsben.ch/RSVLT

Firefox:
image

Chrome: 
image

yeah, of course. Would be useful to choose a faster way for the features that are going to be frequently used.
thanks again for the tools. It's interesting to compare the performance.

0xdevalias: No worries :)

PR #1309, opened 2021-04-07
Fix useCustomSound errors

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/606
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/591

- I assume the whole app should not be wrapped by the secure provider `<CustomSoundsProvider/>` so far as we watch a lot of errors in Bugsnap for not logged users

- `useCustomSound`  is used only on secured pages: 
> src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
src/components/templates/PartyMap/components/Map/MapRoom.tsx

That's why in humble opinion (or need a tip about the better place for this stuff) the `<TemplateWrapper/>` component ought to be wrapped by the `<CustomSoundsProvider/>`

- `fetchSoundConfigs` is an async operation and we need definitely know that is fulfilled (I've added the boolean flag `isLoaded` for this purpose). I've added an additional check on the `<PartyMap/>` component to make sure we render it when the user, its profile and sound are loaded.


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -230,7 +231,11 @@ const VenuePage: React.FC = () => {
     history.push(`/account/profile?venueId=${venueId}`);
   }
 
-  return <TemplateWrapper venue={venue} />;
+  return (
```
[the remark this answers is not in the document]
yurii-lubynets: @0xdevalias @mike-lvov 
please provide the best way to fix this issue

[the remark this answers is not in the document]
0xdevalias: You can just ignore this here.

0xdevalias:
> I assume the whole app should not be wrapped by the secure provider `<CustomSoundsProvider/>` so far as we watch a lot of errors in Bugsnap for not logged users

From https://github.com/sparkletown/internal-sparkle-issues/issues/591:

> A small fix was made in sparkletown/sparkle#1279 to ensure that we don't try and fetch the sounds till after AuthIsLoaded has finished running, as the security rules we have configured require a user to be authenticated before the can fetch the sounds data.

@yurii-lubynets Curious why you chose to move `CustomSoundsProvider` into `src/pages/VenuePage/VenuePage.tsx` in this PR, rather than the method used in #1279?

src/index.tsx
```diff
@@ -257,11 +256,9 @@ render(
...
           <Provider store={store}>
             <ReactReduxFirebaseProvider {...rrfProps}>
-              <CustomSoundsProvider>
-                <AuthIsLoaded>
-                  <AppRouter />
-                </AuthIsLoaded>
-              </CustomSoundsProvider>
+              <AuthIsLoaded>
+                <AppRouter />
+              </AuthIsLoaded>
```
0xdevalias, line 261: > I assume the whole app should not be wrapped by the secure provider `<CustomSoundsProvider/>` so far as we watch a lot of errors in Bugsnap for not logged users

From https://github.com/sparkletown/internal-sparkle-issues/issues/591:

> A small fix was made in sparkletown/sparkle#1279 to ensure that we don't try and fetch the sounds till after AuthIsLoaded has finished running, as the security rules we have configured require a user to be authenticated before the can fetch the sounds data.

@yurii-lubynets Curious why you chose to move `CustomSoundsProvider` into `src/pages/VenuePage/VenuePage.tsx` in this PR, rather than the method used in #1279?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -230,7 +231,11 @@ const VenuePage: React.FC = () => {
...
+    <CustomSoundsProvider>
+      <TemplateWrapper venue={venue} />
+    </CustomSoundsProvider>
```
0xdevalias, line 237: Having this here will only allow us to use custom sounds for templates in the client-side. Given right now we barely make use of the feature, and the only place we do is on `Rooms` on the `Map` this may not be super relevant.

But I can imagine situations when we would also want to be able to access these sounds from the admin panel (eg. to preview what they sound like, etc). In which case, this wouldn't make sense to have here.

As mentioned in another comment (and as followed originally in #1279), I think keeping this in `src/index.tsx` may be a better location.

Note that in #1279 the 'nesting order' is changed so that `CustomSoundsProvider` is a child of `AuthIsLoaded`, which will prevent the errors we are seeing.

---

Note also that for an unknown reason, #1279 seemed to break sounds in some cases. I expect it may be due to the loading/etc stuff you included in this PR or similar, but we didn't dig deeply into it at the time.

This is summarised on https://github.com/sparkletown/internal-sparkle-issues/issues/591:

> A small fix was made in https://github.com/sparkletown/sparkle/pull/1279 to ensure that we don't try and fetch the sounds till after `AuthIsLoaded` has finished running, as the security rules we have configured require a user to be authenticated before the can fetch the sounds data. This was merged directly into `env/kotr`.
> 
> Seemingly this caused issues with the sounds working within the `env/kotr` environment ([slack thread](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1616080529025000)), where they would work when you clicked on 'enter', but wouldn't work when you initially click on the room from the map ([slack ref](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1616083018026800?thread_ts=1616080529.025000&cid=C01ER4M61TL))
> 
> While he couldn't figure out the 'why' of it, @mike-lvov fixed the sound issues in https://github.com/sparkletown/sparkle/pull/1286 by essentially reverting the changes made by https://github.com/sparkletown/sparkle/pull/1279, which was deployed directly to the `env/kotr` environment.
> 
> This seemed to fix the issue in the `env/kotr` environment ([slack thread](https://doesitsparkle.slack.com/archives/C01K319QN5T/p1616086739085500)), but we still aren't really sure why those changes 'caused issues' in the first place.
>

src/hooks/sounds.tsx
```diff
@@ -35,33 +35,40 @@ export interface ExposedDataWithPlay extends ExposedData {
...
+        setSoundConfigs(soundConfigs);
+        setIsLoaded(true);
+      })
+      .catch(() => setIsLoaded(true));
   }, []);
 
   const providerData = useMemo(
     () => ({
       soundConfigs,
+      isLoaded,
```
0xdevalias, line 69: It'd probably be useful to use this new `isLoaded` bool in the `useCustomSound` hook below.

I haven't deeply looked through it/where it would be most useful.. but one place that immediately came to mind was on line 153 within the `useEffect` that has the check `if (hasSoundRef && !hasSoundConfig) {`. We probably don't need to notify Bugsnag here if we haven't even loaded the data yet.

There are almost certainly a few other cases where it would make sense to use within this hook.

src/hooks/sounds.tsx
```diff
@@ -35,33 +35,40 @@ export interface ExposedDataWithPlay extends ExposedData {
...
   useEffect(() => {
-    fetchSoundConfigs().then((soundConfigs) => {
-      setSoundConfigs(soundConfigs);
-    });
+    fetchSoundConfigs()
+      .then((soundConfigs) => {
+        setSoundConfigs(soundConfigs);
+        setIsLoaded(true);
+      })
+      .catch(() => setIsLoaded(true));
```
0xdevalias, line 63: Do we want to be setting `isLoaded` to `true` when we have an error? I guess this depends on how we intend to use this state.

If it's just meant to track that we 'currently are in the progress of loading data', then this makes sense (though would potentially be better set in a `.finally()` rather than in both the `.then()`/`.catch()`. If that's the case, I wonder if 'flipping' it to be `isLoading` would make more sense? Then we could derive `isLoaded` or similar from that if needs be.

If it's meant to represent that we successfully loaded the data from firestore, then this usage seems invalid.

Personally, I feel like having an indication that we successfully loaded the data from firestore seems like a useful thing to have here.

0xdevalias: In the case of an error, depending on what the error reason is, I wonder if there is something better we can do here to 'gracefully recover' (eg. perhaps retrying to fetch the data later or similar maybe? Not sure..)

src/hooks/sounds.tsx
```diff
@@ -35,33 +35,40 @@ export interface ExposedDataWithPlay extends ExposedData {
...
 );
 
 export const CustomSoundsProvider: React.FC = ({ children }) => {
+  const [isLoaded, setIsLoaded] = useState<boolean>(initialValue.isLoaded);
   const [soundConfigs, setSoundConfigs] = useState<SoundConfigMap>(
     initialValue.soundConfigs
   );
 
   // Fetch the sound configs data on first load
   useEffect(() => {
```
0xdevalias, line 57: While this will only currently run on first render, since it's dependent on a user being logged in, it might make sense to add that to it's dependencies.

We could probably use `user` from `src/hooks/useUser.ts` for this.

---

Alternatively, we could follow the default pattern described in the firebase docs, though it sort of deviates from our existing patterns a little (though isn't a bad pattern necessarily):

- https://firebase.google.com/docs/auth/web/manage-users#get_the_currently_signed-in_user

To do that, we would probably have something like this:

```typescript
const [isLoggedIn, setLoggedIn] = useState(false);

useEffect(() => {
  const unsubscribeListener = firebase.auth().onAuthStateChanged((user) => {
    if (user) {
      // User is signed in.
      setLoggedIn(true);
    } else {
      // No user is signed in.
      setLoggedIn(false);
    }
  });

  return () => {
    unsubscribeListener()
  }
}, [])
```

We could then use `isLoggedIn` as the dependency that controls whether we try to fetch the data or not. This should make it work even if this provided is positioned 'outside' of the `AuthIsLoaded` component.

If we do this, it would probably make sense to wrap it up as it's own custom hook, then call it here.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -23,6 +24,7 @@ export const PartyMap: React.FC = () => {
   useConnectCurrentVenue();
   const { user, profile } = useUser();
   const { recentVenueUsers } = useRecentVenueUsers();
+  const { isLoaded } = useCustomSoundsContext();
```
0xdevalias, line 27: To make this variable more meaningful in this context:

```suggestion
  const { isLoaded: isCustomSoundsLoaded } = useCustomSoundsContext();
```

Though it raises the question, do we want to block the entire venue from being accessible just because the sounds aren't loaded yet? I feel like maybe we shouldn't be doing that, as it'll still be largely usable even before the sounds exist.

By checking `isLoaded` at various points in the `useCustomSounds` hook, we can suppress any of the Bugsnag error reporting that are false positives because the data hasn't loaded yet. I expect that will make the need to use this here irrelevant.

So we can probably just remove it outright:

```suggestion
```

We also probably don't want to be using the context directly in components like this. It's more of an 'implementation detail' of the 'private API' that the `useCustomSounds` hook uses.

0xdevalias (changes requested):
Comments/changes/etc

0xdevalias:
For https://github.com/sparkletown/internal-sparkle-issues/issues/606, have you explicitly checked/confirmed that the changes here fix these aspects that @mike-lvov identified?

> I reproduced it by clearing up the cache and reentering the venue.
> 
> _Originally posted by @mike-lvov in https://github.com/sparkletown/internal-sparkle-issues/issues/606#issuecomment-811121299_

> Some info on top of your findings: 
> - people get sounds, when they reload/change the venue
> - clear all the cache & log in to reproduce
> ..snip.. (the other parts here are probably unrelated)
> 
> _Originally posted by @mike-lvov in https://github.com/sparkletown/internal-sparkle-issues/issues/606#issuecomment-811277493_

src/hooks/sounds.tsx
yurii-lubynets: got it

src/components/templates/PartyMap/PartyMap.tsx
yurii-lubynets: got it, removed from `useCustomSoundsContext` and used `user` from `src/hooks/useUser.ts`, we shouldn't block the main page if the sounds is loading.

src/hooks/sounds.tsx
yurii-lubynets: I prefer using `user` from `src/hooks/useUser.ts`

src/index.tsx
yurii-lubynets: Yeah, I strongly agree with it. Like it's better to keep just inside `<AuthIsLoaded/>`

src/pages/VenuePage/VenuePage.tsx
yurii-lubynets: working on it. try to understand why it happens

src/hooks/sounds.tsx
yurii-lubynets: removed `isLoaded` try to find the better solution

yurii-lubynets:
> For [sparkletown/internal-sparkle-issues#606](https://github.com/sparkletown/internal-sparkle-issues/issues/606), have you explicitly checked/confirmed that the changes here fix these aspects that @mike-lvov identified?
> 
> > I reproduced it by clearing up the cache and reentering the venue.
> > _Originally posted by @mike-lvov in [sparkletown/internal-sparkle-issues#606 (comment)](https://github.com/sparkletown/internal-sparkle-issues/issues/606#issuecomment-811121299)_
> 
> > Some info on top of your findings:
> > 
> > * people get sounds, when they reload/change the venue
> > * clear all the cache & log in to reproduce
> >   ..snip.. (the other parts here are probably unrelated)
> > 
> > _Originally posted by @mike-lvov in [sparkletown/internal-sparkle-issues#606 (comment)](https://github.com/sparkletown/internal-sparkle-issues/issues/606#issuecomment-811277493)_

Yes, definitely! reproduced it and working on it

src/hooks/sounds.tsx
```diff
@@ -49,13 +51,15 @@ export const CustomSoundsProvider: React.FC = ({ children }) => {
...
     initialValue.soundConfigs
   );
+  const { user } = useUser();
 
   // Fetch the sound configs data on first load
   useEffect(() => {
+    if (!user?.uid) return;
     fetchSoundConfigs().then((soundConfigs) => {
       setSoundConfigs(soundConfigs);
     });
```
mike-lvov, line 61:
```suggestion
  const { user } = useUser();
  const userId = user?.uid;

  // Fetch the sound configs data on first load
  useEffect(() => {
    if (!userId) return;
    
    fetchSoundConfigs().then((soundConfigs) => {
      setSoundConfigs(soundConfigs);
    });
```

mike-lvov: this way we will prevent useEffect from firing when there is change unrelated to id

src/hooks/sounds.tsx
```diff
@@ -49,13 +51,17 @@ export const CustomSoundsProvider: React.FC = ({ children }) => {
...
 
   // Fetch the sound configs data on first load
   useEffect(() => {
+    if (!userId) return;
+    
     fetchSoundConfigs().then((soundConfigs) => {
       setSoundConfigs(soundConfigs);
     });
-  }, []);
+  }, [user]);
```
mike-lvov, line 64:
```suggestion
  }, [userId]);
```

yurii-lubynets:
@mike-lvov @0xdevalias

I strongly believe we shouldn't call `useCustomSound` until the `soundConfigs` are not loaded because right now we synchronously use it (`useCustomSound`) in `<MapRoom/>` and `<RoomModal/>`. 
The swipe of providers (AuthIsLoaded and CustomSoundsProvider) helps to prevent unlogged user errors.
But when the user is logged in, we just display him `<PartyMap/>` with empty (still is loading) `soundConfigs`.

All in all, I can't find the right solution to fix the issue without blocking the main page `<PartyMap/>` until `soundConfigs` is loaded. Feel free to propose a better way to wait until it's loaded.

0xdevalias:
> All in all, I can't find the right solution to fix the issue without blocking the main page `<PartyMap/>` until `soundConfigs` is loaded.

As I believe I mentioned in one of my earlier comments, sounds should ‘become available’ once they’re loaded, at which point they would apply to the rooms/etc.

So in my mind, the question is whether it’s a worse UX to make the user wait on a loading screen longer, or for them to potentially not have the sounds available for a few seconds when they first load the map. Personally I feel like the latter is the less disruptive experience, because, at least in this ‘sounds on rooms’ experience, I would expect them to not instantly want to click on a room. And then if they do and the sounds aren’t loaded, they aren’t ‘functionally blocked’ from using the app, they just won’t hear the sound.

I’m open to others thoughts/opinions on what is better UX/why though.

yurii-lubynets:
> > All in all, I can't find the right solution to fix the issue without blocking the main page `<PartyMap/>` until `soundConfigs` is loaded.
> 
> @yurii-lubynets @mike-lvov As I believe I mentioned in one of my earlier comments, sounds should ‘automatically become available’ once they’re loaded, at which point they would apply to the rooms/etc.
> 
> So in my mind, the question is whether it’s a worse UX to make the user wait on a loading screen longer, or for them to potentially not have the sounds available for a few seconds when they first load the map. Personally I feel like the latter is the less disruptive experience, because, at least in this ‘sounds on rooms’ experience, I would expect them to not instantly want to click on a room. And then if they do and the sounds aren’t loaded, they aren’t ‘functionally blocked’ from using the app, they just won’t hear the sound.
> 
> I’m open to others thoughts/opinions on what is better UX/why though.

I strongly agree: the best solution to fix useCustomSound errors is to wait until sounds are loaded
And only after that to call our hook `useCustomSound`. But I'm not sure it could be done in the current implementation. 

My point is how to wait until sounds are loaded and only after that to call `useCustomSound`?
(it looks like a conditional call of React hook - which is unavailable)

0xdevalias:
From Slack, RE: https://github.com/sparkletown/sparkle/pull/1309#issuecomment-818192965

> @jonboutelle:
>
> I would vote for not having the sounds available for a few seconds after loading the map. But defer to @mobilevinay 
> <sub>[View in Slack](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1618349075031300)</sub>

mike-lvov:
@0xdevalias I had a call with @yurii-lubynets and played/debugged the sounds. The context/data load thing works as expected. It seems like `"use-sound"` is the problem. I didn't deep dived into the lib itself, but its hook doesn't react to a `url` change. If the hook was provided with an URL(1), and then URL(2) was provided, URL(2) doesn't work. My uneducated guess is that it happens because of the way the `howler` instance is created(???)

0xdevalias:
> I didn't deep dived into the lib itself, but its hook doesn't react to a `url` change. If the hook was provided with an URL(1), and then URL(2) was provided, URL(2) doesn't work. My uneducated guess is that it happens because of the way the `howler` instance is created(???)

@mike-lvov Interesting (and annoying) if that’s the case. Just to be clear, did you trace that change through all of my logic in useCustomSound where I make decisions about if we should play a sound or not? Just want to make sure it’s definitely an issue within useSound and not within our logic there.

mike-lvov:
@0xdevalias 
> @mike-lvov Interesting (and annoying) if that’s the case. Just to be clear, did you trace that change through all of my logic in useCustomSound where I make decisions about if we should play a sound or not? Just want to make sure it’s definitely an issue within useSound and not within our logic there.

Yes, I tracked the data passed to useSound hook. It's valid. One of the proving points that my theory is correct is that the sounds actually work, when you click `Enter`. The reason for that is that the new `useCustomSound` instance is created(with the correct data passed on init).

mike-lvov:
@0xdevalias 
I think the solution, in this case, would be to load the sounds before initiating the map. The other option is to change the lib, which I'm not really fond of.

0xdevalias:
> I think the solution, in this case, would be to load the sounds before initiating the map. The other option is to change the lib, which I'm not really fond of

@mike-lvov Yeah, for now I suppose ensuring they are loaded before the map loads is probably the quickest/easiest solution.

In the future may have to re-assess our choice in using that lib though. The last project I worked on I used howler directly with some minimal custom hooks I wrote to wrap around it. I don’t think this hook actually does a whole lot, so it probably wouldn’t be all that hard to do similar here as well. 

But that would likely be something I would be better placed to do, and I don’t have the capacity for it right now.

yurii-lubynets:
@mike-lvov @0xdevalias I'm really grateful to you.

> @mike-lvov Yeah, for now I suppose ensuring they are loaded before the map loads is probably the quickest/easiest solution.

Actually, I'm preparing this solution today and hope to deal with it this week

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -23,6 +24,7 @@ export const PartyMap: React.FC = () => {
   useConnectCurrentVenue();
   const { user, profile } = useUser();
   const { recentVenueUsers } = useRecentVenueUsers();
+  const [, , isCustomSoundsLoaded] = useCustomSound();
```
0xdevalias, line 27: Since we're returning multiple objects here now, it would make sense for us to move from returning an array, to returning an object of named keys like we do in most of our other custom hooks.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -15,6 +15,7 @@ import { Map, RoomModal } from "./components";
 import SparkleFairiesPopUp from "components/molecules/SparkleFairiesPopUp/SparkleFairiesPopUp";
 
 import "./PartyMap.scss";
+import { useCustomSound } from "hooks/sounds";
```
0xdevalias, line 18: Sort this with the other `hooks/*` imports.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -40,7 +42,7 @@ export const PartyMap: React.FC = () => {
     setSelectedRoom(undefined);
   }, []);
 
-  if (!user || !profile) return <>Loading..</>;
+  if (!user || !profile || !isCustomSoundsLoaded) return <>Loading..</>;
```
0xdevalias, line 45: Since we don't actually use the custom sounds directly within this component (but within a subcomponent) it feels weirdly 'disconnected' to have this code here, and feels like it will be quite brittle/easy to miss when making changes due to it being 'spread out' across multiple components.

I think a better way would be to create a new wrapping component for this, probably that works similarly to how `AuthIsLoaded` (in `src/index.tsx`) works. Basically, waits until the sound config is loaded, and then renders the `children`.

This can probably be done directly within the `CustomSoundsProvider` component, though probably should be controlled by a new prop (eg. `waitTillConfigLoaded: boolean`, or something similar but better named)

I expect we probably want to use the `<LoadingPage />` component here. We can probably 'inject' that via another prop, something like: `loadingComponent: ReactNode` or similar.

Then I would expect the 'provider chain' in `src/index.tsx` to end up looking something like:

```jsx
// ..snip..
  <ReactReduxFirebaseProvider {...rrfProps}>
    <AuthIsLoaded>
      <CustomSoundsProvider loadingComponent={LoadingPage} waitTillConfigLoaded>
        <AppRouter />
      </CustomSoundsProvider>
    </AuthIsLoaded>
  </ReactReduxFirebaseProvider>
// ..snip..
```

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Though probably can be removed entirely, see other comments below.

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Even with the changes described below (about waiting for the loading in the `CustomSoundsProvider` component), it would probably be good to make these changes still + have this variable exposed from the hook (even if we don't end up needing to use it here in our code)

src/hooks/sounds.tsx
```diff
@@ -49,19 +53,29 @@ export const CustomSoundsProvider: React.FC = ({ children }) => {
...
-    });
-  }, []);
+    if (!userId) return;
+
+    fetchSoundConfigs()
+      .then((soundConfigs) => {
+        setSoundConfigs(soundConfigs);
+      })
+      .finally(() => setIsCustomSoundsLoaded(true));
+  }, [userId]);
```
0xdevalias, line 71: The `.finally()` here will run regardless of whether the `Promise` succeeds or fails. Is that intended?

I guess if we're blocking the app from loading based on this, then we probably need to signal to 'allow it to keep loading' even when there is an error.

src/hooks/sounds.tsx
```diff
@@ -35,10 +37,12 @@ export interface ExposedDataWithPlay extends ExposedData {
 
 export interface CustomSoundsState {
   soundConfigs: SoundConfigMap;
+  isCustomSoundsLoaded: boolean;
 }
 
 export const initialValue: CustomSoundsState = {
   soundConfigs: {},
+  isCustomSoundsLoaded: false,
```
0xdevalias, line 45: We can probably just keep this simple, as the context of how we access this variable (eg. from the hooks/etc) will imply that it relates to the custom sounds:

```suggestion
  isLoaded: false,
```

src/hooks/sounds.tsx
```diff
@@ -260,5 +274,5 @@ export const useCustomSound = (
     [play, spriteName, hasSoundConfig, hasValidSpriteConfig, onend]
   );
 
-  return [playSprite, { ...exposedData, play }];
+  return [playSprite, { ...exposedData, play }, isCustomSoundsLoaded];
```
0xdevalias, line 277: As mentioned in an earlier comment, it would be better to return named keys in an object here now, rather than a positional array like this.

0xdevalias (changes requested):
A few more comments/suggestions/refinements.

src/components/templates/PartyMap/PartyMap.tsx
yurii-lubynets: Strongly agree, added 
```
type UseCustomSoundType = {
  play: PlaySpriteFunction,
  data: ExposedDataWithPlay,
  isLoaded: boolean,
};
```
as returned type from `useCustomSound`

src/hooks/sounds.tsx
yurii-lubynets: Yes, I'm using the`.finally()` statement to be sure it's not a blocker for the app.

src/components/templates/PartyMap/PartyMap.tsx
yurii-lubynets: That's looking better than I've expected. Thank you!

src/hooks/sounds.tsx
0xdevalias: Makes sense

src/hooks/sounds.tsx
```diff
@@ -35,35 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
 
 export const CustomSoundsContext = createContext<CustomSoundsState>(
   initialValue
 );
 
-export const CustomSoundsProvider: React.FC = ({ children }) => {
+export type CustomSoundsType = {
+  loadingComponent: React.ReactNode;
+  waitTillConfigLoaded: boolean;
+};
```
0xdevalias, line 55: We use `interface` for props, and name it with the component name + `Props` suffix.

We also want these props to be optional rather than required:

```suggestion
export interface CustomSoundsProviderProps = {
  waitTillConfigLoaded?: boolean;
  loadingComponent?: React.ReactNode;
};
```

src/hooks/sounds.tsx
```diff
@@ -35,35 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
+
+export const CustomSoundsProvider: React.FC<CustomSoundsType> = ({
+  loadingComponent,
+  waitTillConfigLoaded,
+  children,
+}) => {
```
0xdevalias, line 61: Setting defaults for the now-optional parameters. Not sure if this will work exactly as is, but something like this:

```suggestion
export const CustomSoundsProvider: React.FC<CustomSoundsProviderProps> = ({
  loadingComponent = 'Loading...',
  waitTillConfigLoaded = false,
  children,
}) => {
```

src/hooks/sounds.tsx
```diff
@@ -35,35 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
+  const { user } = useUser();
+  const userId = user?.uid;
 
   // Fetch the sound configs data on first load
   useEffect(() => {
-    fetchSoundConfigs().then((soundConfigs) => {
-      setSoundConfigs(soundConfigs);
-    });
-  }, []);
+    if (!userId) return;
```
0xdevalias, line 71: Add a newline of space after these 'guard' statements.
```suggestion
    if (!userId) return;

```

src/hooks/sounds.tsx
```diff
@@ -35,35 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
+    setIsLoading(true);
+    fetchSoundConfigs()
+      .then((soundConfigs) => setSoundConfigs(soundConfigs))
+      .finally(() => setIsLoading(false));
```
0xdevalias, line 75: [NB] I think you could just use `setSoundConfigs` as the 'handler' in the `.then()` directly:

```suggestion
    fetchSoundConfigs()
      .then(setSoundConfigs)
      .finally(() => setIsLoading(false));
```

src/hooks/sounds.tsx
```diff
@@ -35,35 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
       soundConfigs,
+      isLoading,
     }),
-    [soundConfigs]
+    [soundConfigs, isLoading]
   );
 
+  if (isLoading && waitTillConfigLoaded) {
+    return <>{loadingComponent}</>;
+  }
```
0xdevalias, line 88: I think this might be better 'inside' the `CustomSoundsContext.Provider` below. Possibly just as a ternary?

Something like:

```typescript
return (
    <CustomSoundsContext.Provider value={providerData}>
      {waitTillConfigLoaded && isLoading ? loadingComponent : children}
    </CustomSoundsContext.Provider>
  );
```

0xdevalias (changes requested):
[CRO] A couple of tiny tweaks, then I think this should be good to go from a code perspective (I haven't clicked through/tested the implementation to ensure it fixes all the issues properly/doesn't break anything/etc)

src/hooks/sounds.tsx
yurii-lubynets: fixed

src/hooks/sounds.tsx
```diff
@@ -22,6 +22,8 @@ import { fetchSoundConfigs } from "api/sounds";
 
 import { SoundConfigMap, SoundConfigReference } from "types/sounds";
 
+import { useUser } from "hooks/useUser";
+
 import { isDefined } from "utils/types";
```
0xdevalias, line 27:
```suggestion
import { isDefined } from "utils/types";

import { useUser } from "hooks/useUser";
```

0xdevalias, line 27:
```suggestion
import { isDefined } from "utils/types";

import { useUser } from "hooks/useUser";
```

src/hooks/sounds.tsx
```diff
@@ -35,38 +37,56 @@ export interface ExposedDataWithPlay extends ExposedData {
...
+export const CustomSoundsProvider: React.FC<CustomSoundsProviderProps> = ({
+  loadingComponent = "Loading...",
+  waitTillConfigLoaded = false,
+  children,
+}) => {
   const [soundConfigs, setSoundConfigs] = useState<SoundConfigMap>(
     initialValue.soundConfigs
   );
+  const [isLoading, setIsLoading] = useState<boolean>(initialValue.isLoading);
+  const { user } = useUser();
```
0xdevalias, line 66:
```suggestion

  const { user } = useUser();
```

0xdevalias (approved):
[CRO] Couple of tiny styling tweaks, but aside from that, the code looks good to go! I haven't clicked through/tested this out specifically, so if we're happy that it works and doesn't break anything, LGTM!

PR #1310, opened 2021-04-08
add firebase performance monitoring + react interaction tracing

0xdevalias:
Most likely follow up PRs to come to add more custom code traces/etc, but this should be sufficient to start capturing some metrics, so it would be good to get it out sooner rather than later.

- added firebase performance including default pageload analytics + trace on `loadStripe` (`createPerformanceTrace` in `src/utils/performance.ts`)
  - https://firebase.google.com/products/performance
- added experimental React scheduler interaction tracing (`traceReactScheduler`/`wrapReactScheduler` in `src/utils/performance.ts`)
  - https://reactjs.org/docs/profiler.html
  - https://fb.me/react-interaction-tracing
- etc

---

partially closes https://github.com/sparkletown/internal-sparkle-issues/issues/77

---

### Aggregate

image

### Over Time

image


0xdevalias (commented):
Best reviewed with 'Hide whitespace changes' enabled.

src/index.tsx
```diff
@@ -250,26 +267,28 @@ const AuthIsLoaded: React.FunctionComponent<React.PropsWithChildren<{}>> = ({
...
+                </CustomSoundsProvider>
+              </ReactReduxFirebaseProvider>
+            </Provider>
+          </DndProvider>
+        </Elements>
+      </ThemeProvider>
+    </BugsnagErrorBoundary>,
+    document.getElementById("root")
+  );
+});
```
mike-lvov, line 291: I wonder if measuring the performance can affect the performance in the prod

src/utils/performance.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @see https://firebase.google.com/docs/perf-mon/custom-code-traces?platform=web
+ * @see https://firebase.google.com/docs/reference/js/firebase.performance.Performance#trace
+ * @see https://firebase.google.com/docs/reference/js/firebase.performance.Trace
+ */
+export const createPerformanceTrace = (
+  traceName: PerformanceTrace,
+  options?: CreatePerformanceTraceOptions
+) => {
+  const { startNow = false, attributes, metrics } = options ?? {};
```
mike-lvov, line 87:
```suggestion
  options?: CreatePerformanceTraceOptions = {}
) => {
  const { startNow = false, attributes, metrics } = options;
```

It seems to be a more widespread approach

mike-lvov: Also, what is `startNow` and why do you have it `false` by default?

mike-lvov (approved):

src/utils/performance.ts
0xdevalias: It’s used further down to handle starting the created trace. It’s false by default as I wanted to ensure people are explicit about using it, as otherwise it could add unrelated ‘latency’.

The usual way is to call .start() and .stop() manually.

src/index.tsx
0xdevalias: Technically yes, slightly, since we have profiling enabled in production (if I remember correctly)

Though IMO getting the data is far more useful than the tiny performance hit it will add.

PR #1311, opened 2021-04-08
Ensure ChatSidebar is always visible (aka: 'pinned open') on large screens

yurii-lubynets:
- closes: https://github.com/sparkletown/internal-sparkle-issues/issues/622


mike-lvov:
@yurii-lubynets  We need to apply this to every Template. I guess the best place to do it is `TemplateWrapper`

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -18,17 +18,28 @@ import { ReactionPage } from "components/templates/ReactionPage";
...
+  const dispatch = useDispatch();
   const history = useHistory();
   const match = useRouteMatch();
 
+  useEffect(() => {
+    if (width > LARGE_SCREEN_WIDTH) {
+      dispatch(setChatSidebarVisibility(true));
+    }
+  }, [dispatch, width]);
+
```
mike-lvov, line 42: this logic should be moved into a hook. Probably into chats hook.

src/hooks/useWindowDimensions.ts
```diff
@@ -0,0 +1,26 @@
...
+  );
+
+  useEffect(() => {
+    function handleResize() {
+      setWindowDimensions(getWindowDimensions());
+    }
+
+    window.addEventListener("resize", handleResize);
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
```
mike-lvov, line 23: I think this will lead to the creation of multiple `resize` listeners, which makes this hook not reusable

src/components/templates/PartyMap/PartyMap.scss
```diff
@@ -23,6 +23,10 @@
   .userlist-container {
     margin: 0 80px;
   }
+
+  @media (min-width: $large-laptop-width) {
+    width: calc(100% - #{$sidebar-width});
+  }
```
mike-lvov, line 29: This will only handle it for `PartyMap` template. We need to handle it for every template

mike-lvov (changes requested):

src/hooks/useWindowDimensions.ts
yurii-lubynets: yes, but the `return` statement will clean up all of them, and every time we are having the current window size

src/pages/VenuePage/TemplateWrapper.tsx
yurii-lubynets: got it

src/hooks/useWindowDimensions.ts
mike-lvov: Yes, but imagine using this hook in 5 places simultaneously. It will create 5 listeners

src/pages/VenuePage/TemplateWrapper.tsx
yurii-lubynets: done

src/hooks/useWindowDimensions.ts
yurii-lubynets: make sense, will handle it as well

src/components/templates/PartyMap/PartyMap.scss
yurii-lubynets: fixed

src/hooks/useWindowDimensions.ts
0xdevalias: Instead of rolling our own, can use something like https://github.com/streamich/react-use/blob/master/docs/useWindowSize.md

I want to bring this lib in anyway as it has tons of useful util functions like this

0xdevalias: (Probably not needed here, but for future knowledge, see also: https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API)

yurii-lubynets: IMHO `window.innerWidth` deals with it effective.

src/hooks/chatSidebar.ts
```diff
@@ -87,11 +86,10 @@ export const useChatSidebarInfo = () => {
...
-  const { width } = useWindowDimensions();
   const dispatch = useDispatch();
   useEffect(() => {
-    if (width > LARGE_SCREEN_WIDTH) {
+    if (window.innerWidth > LARGE_SCREEN_WIDTH) {
       dispatch(setChatSidebarVisibility(true));
     }
-  }, [dispatch, width]);
+  }, [dispatch]);
 };
```
0xdevalias: In removing the `useWindowDimensions` hook, this code will never be triggered automatically on a window resize. Which means it's only going to be 'properly checked' and fix the rendering whenever this screen ends up being re-rendered by something else. Which most likely is going to lead to display issues.

0xdevalias (changes requested):

0xdevalias (commented):
Currently this PR seems to be needing to add the media query to a LOT of individual components.. I wonder if there is a better way that we can solve this in a single place 'above' those components, and then have them 'just work', without needing to individually know about/do anything specific to cater for this sidebar. My gut feel is that this would happen in `src/pages/VenuePage/TemplateWrapper.tsx` or similar.

yurii-lubynets:
> Currently this PR seems to be needing to add the media query to a LOT of individual components. I wonder if there is a better way that we can solve this in a single place 'above' those components, and then have them 'just work', without needing to individually know about/do anything specific to cater for this sidebar. My gut feel is that this would happen in `src/pages/VenuePage/TemplateWrapper.tsx` or similar.

It was my first idea to update the TemplateWrapper to support this feature 
but checking deeply I clearly see that every template has its own layout and it's hard to predict how these changes could be applied for them. I strongly believe that these changes should be inside every component to make sure it's rendering correctly

src/hooks/chatSidebar.ts
yurii-lubynets: I was trying to install this package `react-use` and had lost a fight with TS:
```
/Applications/projects/sparkle/sparkle/node_modules/react-use/lib/useGetSet.d.ts
TypeScript error in /Applications/projects/sparkle/sparkle/node_modules/react-use/lib/useGetSet.d.ts(3,82):
',' expected.  TS1005
```
Can't find the way to deal with this `width` checking. Open for any tips/advise

src/hooks/useWindowDimensions.ts
0xdevalias: See issues with this method mentioned in https://github.com/sparkletown/sparkle/pull/1311#discussion_r611312150

0xdevalias:
> It was my first idea to update the `TemplateWrapper` to support this feature
> but checking deeply I clearly see that every template has its own layout and it's hard to predict how these changes could be applied for them. I strongly believe that these changes should be inside every component to make sure it's rendering correctly

@yurii-lubynets While they do have their own layouts, these are all largely (if not literally) constrained to the 'box' that they are rendered in; and they should be able to react to that box being smaller and 'just work'. If they can't currently, that would be the better fix to make here (and would likely just involve changing the units used in sizing I would guess, eg. if they currently use `vw`/`vh`/etc rather than `%`)

The `ChatSidebar` component is rendered within `TemplateWrapper`, and that is where we want to handle this 'concern'. It may be possible to handle it within the `ChatSidebar` component itself, which would likely be an even more ideal situation, as it would keep that feature even more 'encapsulated'.

Essentially, we want to limit the 'sprawl' of 'unrelated bits' as much as possible, as that makes code much harder to maintain, and increases the likelihood of introducing bugs. So if we can 'focus' the implementation within the `TemplateWrapper`/`ChatSidebar`/similar, and have that 'flow down' to all of the child components, that's a better solution (even if it means we have to spend a little bit of extra up-front time checking/testing/tweaking those components so that they work well with this pattern)

src/hooks/chatSidebar.ts
0xdevalias: Yup, I just tried locally: `npm i react-use`, then tried adding the following minimal code to a component to test:

```typescript
```

Which gave this error:

```
Failed to compile
/Users/devalias/dev/sparkle/sparkle/node_modules/react-use/lib/useGetSet.d.ts
TypeScript error in /Users/devalias/dev/sparkle/sparkle/node_modules/react-use/lib/useGetSet.d.ts(3,82):
',' expected.  TS1005

    1 | import { Dispatch } from 'react';
    2 | import { IHookStateInitAction, IHookStateSetAction } from './misc/hookState';
  > 3 | export default function useGetSet<S>(initialState: IHookStateInitAction<S>): [get: () => S, set: Dispatch<IHookStateSetAction<S>>];
      |                                                                                  ^
    4 | 
```

Looking at https://stackoverflow.com/questions/46398916/typescript-error-ts1005-expected-ii/51936515 it seems to suggest that this error tends to come up when TypeScript doesn't understand the syntax being used, which supports my theory that we need to updated to a newer version.

This seems to be confirmed in this closed issue on the repo:

- https://github.com/streamich/react-use/issues/1879
  - > Use TS version 4+

We're currently using `"typescript": "^3.9.3"`:

- https://github.com/sparkletown/sparkle/blob/staging/package.json#L156

I bumped this to the latest version (`4.2.4`) locally to test if it would resolve this issue:

- https://github.com/microsoft/TypeScript/releases/tag/v4.2.4
  - https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#breaking-changes
  - https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#breaking-changes
  - https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#breaking-changes
  - https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/#breaking-changes

There were a couple of typing issues this uncovered that needed to be fixed in legacy code, but it fixed the error that was coming up based on `react-use`.

Since upgrading TypeScript is unrelated to this PR, I will handle doing this in a new PR. Shall link it here once i've pushed the changes up.

0xdevalias: See https://github.com/sparkletown/internal-sparkle-issues/issues/160

0xdevalias: See https://github.com/sparkletown/sparkle/pull/1321 for adding `react-use` as a dependency

yurii-lubynets: Thank you a lot for your help. 
I've used your PR https://github.com/sparkletown/internal-sparkle-issues/issues/160 in my own changes to install `react-use`

0xdevalias: Sounds good. Just make sure those other 2 PRs get merged before this one does so it keeps everything properly segregated :)

src/hooks/useWindowDimensions.ts
yurii-lubynets: Using `useWindowSize` from `react-use` package

src/components/organisms/WithNavigationBar/WithNavigationBar.scss
```diff
@@ -3,6 +3,11 @@
 .navbar-margin {
   padding-top: $navbar-height;
   padding-bottom: $footer-height;
+
+  @media (min-width: $large-laptop-width) {
+    display: flex;
+    justify-content: space-between;
+  }
```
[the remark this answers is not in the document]
0xdevalias: It looks like every case within `src/pages/VenuePage/TemplateWrapper.tsx` renders a single component for it's template, so presumably the spacing provided by these `flex`/etc would be fine for all of those cases by default.

src/components/templates/FireBarrel/FireBarrel.styled.ts
```diff
@@ -6,6 +6,12 @@ export const Wrapper = styled.div`
   margin: 6rem auto;
   grid-template-columns: repeat(4, 1fr);
   grid-gap: 1rem;
+
+  @media (min-width: 1200px) {
+    width: calc(100% - 360px);
+    margin: 0;
+    padding: 3rem;
+  }
```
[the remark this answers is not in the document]
0xdevalias: Given `src/components/organisms/WithNavigationBar/WithNavigationBar.scss` sets `flex` styles, i'm curious why we need to do anything 'special' here? Shouldn't this 'inherit' the width it can use based on that? There's a possibility we might need to use `min-width` and/or `max-width`, to make that work.. but I expect we shouldn't need to be doing anything 'special' to calculate the size here ourselves.

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -10,7 +10,8 @@ $header-height: 45px;
   flex-flow: column;
   top: 66px;
   right: 0;
-  width: 360px;
+  width: $sidebar-width;
+  min-width: $sidebar-width;
   height: calc(100vh - 96px);
```
0xdevalias, line 15: What does this `96px` 'magic number` represent here, and do we have an SCSS variable for it yet? Is it the header height or similar?

Also, is it possible to change the `100vh` here to `100%` and still have it look the same?

0xdevalias (changes requested):
Questions/comments/changes.

package-lock.json
```diff
@@ -3759,6 +3759,11 @@
         }
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

package.json
```diff
@@ -61,6 +61,7 @@
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
     "react-switch": "^5.0.1",
+    "react-use": "^17.2.3",
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

package.json
```diff
@@ -153,7 +154,7 @@
     "gifwrap": "^0.9.2",
     "phin": "^3.5.0",
     "ts-node": "^9.1.1",
-    "typescript": "^3.9.3",
+    "typescript": "^4.2.4",
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/components/templates/Playa/Playa.tsx
```diff
@@ -376,14 +376,14 @@ const Playa = () => {
...
   useEffect(() => {
     if (camp) {
       const campVenue = venues?.find((venue) => venue.id === camp);
       if (campVenue && !PLAYA_TEMPLATES.includes(campVenue.template)) {
-        if (camp.placement) {
-          setCenterX(camp.placement.x);
-          setCenterY(camp.placement.y);
+        if (campVenue.placement !== undefined) {
+          setCenterX(campVenue.placement.x);
+          setCenterY(campVenue.placement.y);
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -20,6 +21,10 @@ $header-height: 45px;
     transition: transform 0.3s;
     background-color: rgba($saturated-black, 0.9);
     box-shadow: -10px 0 40px rgba(0, 0, 0, 0.5);
+    @media (min-width: $large-laptop-width) {
+      position: relative;
+      top: 0;
+    }
```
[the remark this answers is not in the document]
yurii-lubynets: when we change position to relative -> it starts its calculation from container top and it should be `top: 0`

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -18,14 +18,15 @@ import { ReactionPage } from "components/templates/ReactionPage";
 import { ChatSidebar } from "components/organisms/ChatSidebar";
 import { UserProfileModal } from "components/organisms/UserProfileModal";
 import { WithNavigationBar } from "components/organisms/WithNavigationBar";
-
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -18,14 +18,15 @@ import { ReactionPage } from "components/templates/ReactionPage";
...
 import { AnnouncementMessage } from "components/molecules/AnnouncementMessage";
+import { useChatVisibility } from "hooks/chatSidebar";
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -18,14 +18,15 @@ import { ReactionPage } from "components/templates/ReactionPage";
...
 
 export interface TemplateWrapperProps {
   venue: AnyVenue;
 }
 
 const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
+  useChatVisibility();
```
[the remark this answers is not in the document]
yurii-lubynets: moved to `ChatSidebar `

src/components/organisms/ChatSidebar/ChatSidebar.scss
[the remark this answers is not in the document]
0xdevalias: Makes sense. Can you maybe add a comment explaining that we don't need to account for the `NavBar` anymore here because of what you described above.

src/hooks/chatSidebar.ts
```diff
@@ -16,6 +16,7 @@ import { useDispatch } from "./useDispatch";
 import { useNumberOfUnreadChats } from "./privateChats";
 import { useVenueId } from "./useVenueId";
 import { useConnectCurrentVenueNG } from "./useConnectCurrentVenueNG";
+import { LARGE_SCREEN_WIDTH } from "settings";
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/hooks/chatSidebar.ts
```diff
@@ -83,3 +84,14 @@ export const useChatSidebarInfo = () => {
...
+export const useChatVisibility = () => {
+  const dispatch = useDispatch();
+  const { width } = useWindowSize();
+
+  useEffect(() => {
+    if (width > LARGE_SCREEN_WIDTH) {
+      dispatch(setChatSidebarVisibility(true));
+    }
+  }, [dispatch, width]);
+};
```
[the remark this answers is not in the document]
yurii-lubynets: fixed

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -68,9 +73,6 @@ $header-height: 45px;
     background-color: $intermediate-grey;
     border-radius: 16px 0 0 16px;
     cursor: pointer;
-    display: flex;
-    align-items: center;
-    justify-content: center;
```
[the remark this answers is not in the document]
yurii-lubynets: duplicate, see the file above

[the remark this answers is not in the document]
0xdevalias: Which 'file above' are you specifically referring to?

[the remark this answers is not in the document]
0xdevalias: Oh, I see. In this same class. Makes sense.

src/scss/constants.scss
```diff
@@ -37,6 +37,7 @@ $danger-gradient: linear-gradient(
 $admin-gradient: linear-gradient(141deg, #5a35ad 0%, #26144f 100%);
 
 $map-sidebar-width: 20%;
+$sidebar-width: 360px;
```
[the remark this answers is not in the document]
yurii-lubynets: `$map-sidebar-width` is used only in `.party-map-map-sidebar` and this style is not used at all. removed `$map-sidebar-width`

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -83,6 +85,10 @@ $header-height: 45px;
     &__second-icon {
       margin-left: 5px;
     }
+
+    @media (min-width: $large-laptop-width) {
+      display: none;
+    }
```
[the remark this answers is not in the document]
Sparkle-T: What I see looks good AS AN OPTION, and a good OPTION is to lock it open - but is it possible to hide / minimise this by default? 

In practice, only some people want to have a chat function - and others feel it takes away from the "focus of the experience."  Most people are not on Sparkle to have detailed text based chats with others - but seeing other peoples comments in spaces has a part.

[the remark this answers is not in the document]
yurii-lubynets: imho when we are facing the problem to handle logic related to keeping the sidebar 'locked open' - we will update this .css 
But right now it's fine solution I suppose

src/components/organisms/WithNavigationBar/WithNavigationBar.scss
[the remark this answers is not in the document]
yurii-lubynets: got it, fix

src/components/templates/FireBarrel/FireBarrel.styled.ts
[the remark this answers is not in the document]
yurii-lubynets: fixed by `LARGE_SCREEN_WIDTH `

src/components/organisms/ChatSidebar/ChatSidebar.scss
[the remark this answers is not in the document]
0xdevalias: @yurii-lubynets The changes in this PR are basically a version of the 'locked open' stuff

@Sparkle-T While I hear your concerns, they aren't aligned with the issue described in https://github.com/sparkletown/internal-sparkle-issues/issues/622, which is what this PR is working on. Until @sparkletown/product says otherwise, we are continuing as planned.

If you feel strongly about this (which it seems you do), then please chase up https://github.com/sparkletown/internal-sparkle-issues/issues/622#issuecomment-814613821 and get a decision made.

Until then, @yurii-lubynets please disregard the @Sparkle-T 's comments above and continue as planned.

(@jonboutelle @mobilevinay FYI RE: 'too many cooks spoil the stew')

src/components/organisms/ChatSidebar/ChatSidebar.scss
yurii-lubynets: well, `navbar-margin` styles contains 
```
padding-top: 66px;
padding-bottom: 30px;
```
That's why here is using 96px. Changing to `100%` will cut the chat for small resolution

src/components/templates/Audience/Audience.scss
```diff
@@ -203,4 +203,8 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
       background-color: rgba(0, 0, 0, 0);
     }
   }
+
+  @media (min-width: $large-laptop-width) {
+    min-width: initial;
+  }
```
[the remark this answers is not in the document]
yurii-lubynets: I suppose it's the only case for `.audience-container` which has property: `min-width: fit-content;` and it'd be overrided on large screens

src/components/organisms/ChatSidebar/ChatSidebar.scss
0xdevalias: I meant literally replacing the 100vh with 100%, yet still keeping the calc/etc logic. Does that work?

Do we have an scss variable for that 96px of navbar height set anywhere? Can we use that here?

0xdevalias: Looking at `src/components/organisms/WithNavigationBar/WithNavigationBar.scss`:

```scss
.navbar-margin {
  padding-top: $navbar-height;
  padding-bottom: $footer-height;
}
```

Looking at `src/scss/constants.scss`:

```scss
$navbar-height: 66px;
$footer-height: 30px;
```

So we have variables for the `NavBar` height, as I guessed.

---

Therefore, within `src/components/organisms/ChatSidebar/ChatSidebar.scss` I'm wondering if we can set the styles like this:

```diff
- top: 66px;
+ top: $navbar-height;
  right: 0;
  width: $sidebar-width;
  min-width: $sidebar-width;
- height: calc(100vh - 96px);
+ height: calc(100% - $navbar-height - $footer-height);
```

I just tested in DevTools and it still seems to look correct to me when using 100% (since when floating it's going to be 100% of I believe the `body` or `html` or similar tags, which are set in a way that they should be more or less equivalent to `100vh` (which means '100% of the viewport height'). But by using the `%` instead of `vh`, if we embed this within another container and/or when we make it 'non-floating', then it should end up working better. In the 'non-floating' mode we'll likely want/need to set the height to just `100%`, as we'll likely no longer need to account for the header/footer explicitly.

yurii-lubynets: Yes, it look better now, thanks

0xdevalias:
On a large screen currently, it looks like the whole map/sidebar/etc doesn't take up the full height of the screen:
image

This appears to be from the `navbar-margin` class in the `src/components/organisms/WithNavigationBar/WithNavigationBar.tsx` component.

If I add the `fullscreen` class to this as well, it looks correct again though: 
image

I remember reviewing some code recently (I think it might have been @mike-lvov's if I remember correctly?) where this was removed as it was thought that it had no use/impact anymore, though as we can see in this PR, that doesn't seem to be true (that code may not have been merged yet though)

I believe the fix for this will be to pass the `fullscreen` prop to the `WithNavigationBar` component in `src/pages/VenuePage/TemplateWrapper.tsx` for this template type. It's probably also worth checking the other templates as well, and seeing whether they look correct without this parameter. I expect we will probably want to be setting it for every template; which if we do, we can probably just pass it statically in the JSX rather than having it as a variable that is changed for each template.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -9,13 +9,18 @@ import {
...
+  useChatSidebarControls,
+  useChatSidebarInfo,
+  useChatSidebarLockedOpen,
+} from "hooks/chatSidebar";
 
 import { ChatTypes } from "types/chat";
 
 import "./ChatSidebar.scss";
 
 export const ChatSidebar: React.FC = () => {
```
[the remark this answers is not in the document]
0xdevalias: This looks ok again once the sidebar returns to 'floating mode':

image

Except notice how the footer is now 'cutting off' the sidebar'? That shouldn't be happening:

image

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -8,10 +8,12 @@ $header-height: 45px;
...
   flex-flow: column;
-  top: 66px;
+  top: $navbar-height;
   right: 0;
-  width: 360px;
-  height: calc(100vh - 96px);
+  bottom: 0;
+  width: $sidebar-width;
+  min-width: $sidebar-width;
+  min-height: calc(100% - #{$navbar-height + $footer-height});
```
0xdevalias, line 16: Since we're already using `calc` here, we may as well show our full 'calculations'. Will make the 'why' of it more obvious when viewed in devtools/etc
```suggestion
  min-height: calc(100% - #{$navbar-height} - #{$footer-height});
```

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -20,6 +22,12 @@ $header-height: 45px;
     transition: transform 0.3s;
     background-color: rgba($saturated-black, 0.9);
     box-shadow: -10px 0 40px rgba(0, 0, 0, 0.5);
+
+    // change position to relative -> it starts its calculation from container top and it should be top: 0
+    @media (min-width: $large-laptop-width) {
```
0xdevalias, line 27: I'm wondering whether instead of these media queries/breakpoints, if we should just define these overrides as a sidebar variant (eg. `&--pinned` or similar). That way we would have more control as to when we enable/disable it (eg. in venue config, or as a button that the user can click on the sidebar itself to 'pin it open', etc) That might satisfy @Sparkle-T's concerns + also allow us to put more control over the UX into the end user's hands.

(cc // @mike-lvov @mcflyok FYI: what do you think?)

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -9,13 +9,18 @@ import {
...
+  useChatSidebarLockedOpen();
```
0xdevalias, line 23: This might be a better name for it?

```suggestion
  useChatSidebarPinned();
```

src/scss/constants.scss
```diff
@@ -36,7 +36,7 @@ $danger-gradient: linear-gradient(
 );
 $admin-gradient: linear-gradient(141deg, #5a35ad 0%, #26144f 100%);
 
-$map-sidebar-width: 20%;
+$sidebar-width: 360px;
```
0xdevalias, line 39: As mentioned in another comment in this review, perhaps instead of a static `px` width here, we might want to make it responsive? Probably with a minimum width as well.

(cc // @mike-lvov @mcflyok FYI: Thoughts?)

0xdevalias (changes requested):
Comments/suggestions/etc.

See also the screenshots/display issue described in https://github.com/sparkletown/sparkle/pull/1311#issuecomment-820848738

src/components/organisms/ChatSidebar/ChatSidebar.scss
0xdevalias: Awesome :)

0xdevalias:
**FYI:** I have also set the `❌ do-not-merge` flag on this PR so that we can be certain we have addressed/aligned on the concerns that @Sparkle-T brought up (pending @sparkletown/product chiming in on them) before merging it.

src/components/organisms/ChatSidebar/ChatSidebar.scss
yurii-lubynets: made as `pinned` variant.

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -8,10 +8,12 @@ $header-height: 45px;
...
   display: flex;
   flex-flow: column;
-  top: 66px;
+  top: $navbar-height;
   right: 0;
-  width: 360px;
-  height: calc(100vh - 96px);
+  bottom: 0;
+  width: $sidebar-width;
+  min-width: $sidebar-width;
```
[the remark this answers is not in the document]
yurii-lubynets: Got it, what should be the expected behavior? Like I can't imagine the flexible width chat.
Right now this chat space works as a youtube video with chat

src/components/organisms/ChatSidebar/ChatSidebar.tsx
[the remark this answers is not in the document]
yurii-lubynets: fixed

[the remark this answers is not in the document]
yurii-lubynets: @0xdevalias 
min width for `audience-container` at `Audience` component is 1270px (based on layout with seat-size) so it's not responsible if width is less.
That's why I'm facing an issue with the "pinned" Chatsidebar (width: 360px) for screens from 1200px to 1630px (they are cut). 
IMHO the easiest solution here is making custom media query. Any ideas?

src/scss/constants.scss
yurii-lubynets: left as fixed value, updated styles for Chatsidebar

src/components/organisms/ChatSidebar/ChatSidebar.scss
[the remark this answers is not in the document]
0xdevalias: @mike-lvov @mcflyok Any thoughts on the above? Particularly RE: making the chat sidebar responsive rather than having a constant fixed `px` width?

src/components/organisms/ChatSidebar/ChatSidebar.tsx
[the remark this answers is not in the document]
0xdevalias: @mike-lvov @mcflyok Any thoughts/inputs on this?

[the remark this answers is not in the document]
mike-lvov: @yurii-lubynets @0xdevalias I don't think there is a default, explicit min-width for audience-container. It's calculated based on the number of seats(which can/will vary). Thus, I don't think queries can help here and I think  for the Audiecne template, we should just set it to the floating mode(even on large screens)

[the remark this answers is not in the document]
yurii-lubynets: sounds like a plan, could implement it after the poll feature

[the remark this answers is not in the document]
0xdevalias: This'll make @Sparkle-T happy.

src/components/templates/Audience/Audience.scss
```diff
@@ -20,6 +20,7 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
   height: 100%;
   padding: $spacing;
   min-width: fit-content;
+  width: 100%;
```
0xdevalias, line 23: Does this need to be explicitly added here? Will the `flex` in `src/components/organisms/WithNavigationBar/WithNavigationBar.scss`/similar not handle this already?

src/hooks/chatSidebar.ts
```diff
@@ -83,3 +89,24 @@ export const useChatSidebarInfo = () => {
     venueChatTabTitle: `${chatTitle} Chat`,
   };
 };
+
+export const useChatSidebarPinned = () => {
+  const dispatch = useDispatch();
+  const { width } = useWindowSize();
+  const venue = useSelector(currentVenueSelectorData);
```
0xdevalias, line 96: We probably don't want to be using this 'raw selector' here like this.

Since `src/pages/VenuePage/TemplateWrapper.tsx` already has the `venue` in-scope, and it renders `src/components/organisms/ChatSidebar/ChatSidebar.tsx`, which then calls this `useChatSidebarPinned` hook; perhaps we could pass the `venue` as a prop to both `ChatSidebar` and then `useChatSidebarPinned`?

src/hooks/chatSidebar.ts
```diff
@@ -83,3 +89,24 @@ export const useChatSidebarInfo = () => {
...
+  const isPinned = useMemo(
+    () =>
+      width > LARGE_SCREEN_WIDTH && venue?.template !== VenueTemplate.audience,
```
0xdevalias, line 99: It would probably be good to extract this into a `const`, maybe at the top of this file, maybe in `src/settings.ts` (obviously come up with a better name than `FOO_GOOD_CONST_NAME`):

```suggestion
      width > LARGE_SCREEN_WIDTH && !FOO_GOOD_CONST_NAME.includes(venue?.template),
```

src/hooks/chatSidebar.ts
```diff
@@ -83,3 +89,24 @@ export const useChatSidebarInfo = () => {
...
+    [width, venue]
+  );
+
+  useEffect(() => {
+    if (isPinned) {
+      dispatch(setChatSidebarVisibility(true));
+    }
+  }, [dispatch, isPinned, width]);
```
0xdevalias, line 107: I don't think you want to be including `width` in the dependencies here?:

```suggestion
  useEffect(() => {
    if (isPinned) {
      dispatch(setChatSidebarVisibility(true));
    }
  }, [dispatch, isPinned]);
```

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -32,7 +32,6 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
   let template;
   // @debt remove backButton from Navbar
   let hasBackButton = true;
-  let fullscreen = false;
```
0xdevalias, line 35: What's the intent/justification for removing this? Have we tested that it looks correct on all of the other templates that didn't previously set this to `true`? eg.:

- `VenueTemplate.jazzbar`
- `VenueTemplate.friendship`
- `VenueTemplate.partymap` / `VenueTemplate.themecamp`
- `VenueTemplate.artpiece`
- `VenueTemplate.conversationspace`
- `VenueTemplate.firebarrel`
- etc

0xdevalias (changes requested):
[CRO] A few comments/suggestions/questions.

src/hooks/chatSidebar.ts
mike-lvov: @yurii-lubynets that const will be an array with all(currently it's just `audience`, but will be more in the future) venues that need `chatsidebar` to be always pinned

src/components/templates/Audience/Audience.scss
yurii-lubynets: Yes, we need, `.audience-container` is aligned width only for its content but not all parent container. Or use `flex: 1`

src/hooks/chatSidebar.ts
yurii-lubynets: fixed. Looks better

src/hooks/chatSidebar.ts
yurii-lubynets: fixed, thank you a lot

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -51,6 +51,7 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
     case VenueTemplate.partymap:
     case VenueTemplate.themecamp:
       template = <PartyMap />;
+      fullscreen = true;
```
yurii-lubynets, line 54: to make sure the chat is aligned full height

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -119,14 +120,15 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
     default:
       // Technically TypeScript should prevent us missing a case here, but just in case, we work around it with an explicit cast to be able to render this
       template = <div>Unknown Template: ${(venue as AnyVenue).template}</div>;
+      fullscreen = true;
```
yurii-lubynets, line 123: to make sure the chat is aligned full height

src/pages/VenuePage/TemplateWrapper.tsx
yurii-lubynets: have checked it and reverted. Fixed for some cases

src/components/templates/Audience/Audience.scss
0xdevalias: Would `flex: 1` make it more versatile?

0xdevalias (commented):
[CRO] A couple of minor comments/questions.

---

@mike-lvov Would be good to get your thoughts on these.

Aside from that/once they are handled, this seems good to me from a code review perspective. I haven't done a clickthrough on the preview URL, so that should be done before approving to make sure it all looks/works as expected. Once that's done and you're happy with things, feel free to approve/get this merged.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: Sounds good.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -43,20 +54,22 @@ export const ChatSidebar: React.FC = () => {
...
-              <FontAwesomeIcon
-                className="chat-sidebar__controller__second-icon"
-                icon={faCommentDots}
-                size="lg"
-              />
-            </>
-          )}
-        </div>
+        {!isPinned && (
+          <div className="chat-sidebar__controller" onClick={toggleSidebar}>
```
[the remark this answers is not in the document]
mike-lvov: @0xdevalias 
I used the same approach in `UserAvatar` atom component. 
image

IMO handling it via JS is more explicit, thus should be used

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -28,6 +38,7 @@ export const ChatSidebar: React.FC = () => {
 
   const containerStyles = classNames("chat-sidebar", {
     "chat-sidebar--expanded": isExpanded,
+    "chat-sidebar--pinned": isPinned,
   });
```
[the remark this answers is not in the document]
mike-lvov: agreed

src/components/organisms/ChatSidebar/ChatSidebar.tsx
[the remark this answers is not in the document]
0xdevalias: @mike-lvov Yeah.. I see that you did.. but not sure if that's a good choice there either.. (eg. in that case, it works when there is only a single status dot, but as I called out in the original PR, by properly using BEM variants for it we would be able to cleanly support all sorts of status' with the same general code; though that was deemed OOS)

Conditional rendering causes React to have to re-process things, and also makes it harder to inspect the source and tweak the variants on things/etc.

Not conditional rendering (depending what code needs rendering) may lead to excess things being run/processed/rendered..

I know we already follow this pattern in some places, for or example in `src/components/templates/PartyMap/components/Map/MapRoom.tsx` (before we added `isUnclickable`) the label's visibility was controlled by the scss + BEM variant applied, not by conditional rendering in the JSX.

_shrugs_ Seems there are pros and cons to each.. not worth blocking this PR over.

src/components/templates/Audience/Audience.scss
yurii-lubynets: updated to `flex: 1`

src/components/organisms/ChatSidebar/ChatSidebar.tsx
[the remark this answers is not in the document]
yurii-lubynets: left as JSX

src/components/organisms/ChatSidebar/ChatSidebar.tsx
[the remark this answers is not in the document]
yurii-lubynets: looks as rock

0xdevalias (commented):
[CRO] Code diff looks good to go. Haven't clicked through the preview to check things out yet/test functionality/etc.

0xdevalias:
Looks good on the main `partymap` when both pinned/not pinned

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliashacksville

---

Looks like it squishes the JazzBar when pinned (eg. center content is no longer `16:9`, etc), also doesn't seem to fill 100% of the height (notice the chat box isn't 'touching' the footer like it does on the map, etc. Pinned view also seems to look super broken when sitting at a table.

When not pinned it seems to look ok.

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasjazzbar
  -  image
  - image

---

Auditorium is never pinned, so it looks fine as normal:

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasauditorium

---

Conversation Space when pinned, the 'send message' part of the chat sidebar isn't full height/touching the footer as it should be.

When sitting at a table when pinned things look broken too.

Not pinned seems to look more or less fine.

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasconversationspace
  - image
  - image

---

Art Piece looks ok'ish when pinned.. not perfect though? Looks like it might need more space around it? The chat sidebar 'send message' part seems to be way off screen as well, and needs to be scrolled to be seen?

Seems to look fine when not pinned.

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliashacksville
  - image

---

Burn Barrel when pinned has a lot of space between the 'main content' and the sidebar, and the sidebar doesn't extend the full height to the footer.

Looks fine when not pinned.

- https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasburnbarrel
  - image

0xdevalias (changes requested):
Based on the clickthrough issues outlined in https://github.com/sparkletown/sparkle/pull/1311#issuecomment-828895800 there are more changes required here before this is ready to go.

Please make sure to test how it looks in all venue types when both pinned/not pinned to make sure it all looks correct before putting this up for re-review again.

yurii-lubynets:
@0xdevalias thanks a lot for your attention. I've tested venue types and fix your comments. 

- I've made `jazzbar` as the second exception for CHATBAR_NEVER_PINNED_TEMPLATES

 https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasjazzbar

(IMO the main experience here is the video and it's squished when "pinned" Chatbar is). Open for discussion

- I'm not sure about the spacing and position of the Chatbar footer for burn barrel (like all people has a widescreen display)
https://co-reality-staging--preview-pr-1311-2eubmcc0.web.app/in/devaliasburnbarrel

@Sparkle-T please check above

yarikoptic:
FWIW I think the scaling for the content on the "home" page doesn't account for the pinned taskbar width and thus introduces the scrollbar to it (undesired side-effect)

<details>
<summary>screenshot</summary>  

image

</details>

yurii-lubynets:
closed based on @ermiaSparkle input

PR #1314, opened 2021-04-09
Fix invalid URL

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/9

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -19,7 +19,7 @@ export const venueEntranceUrl = (venueId: string, step?: number) => {
 
 export const isExternalUrl = (url: string) => {
   try {
-    return new URL(url).host !== window.location.host;
+    return new URL(url, window.location.origin).host !== window.location.host;
   } catch (error) {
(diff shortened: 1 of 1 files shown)


src/utils/url.ts
```diff
@@ -19,7 +19,7 @@ export const venueEntranceUrl = (venueId: string, step?: number) => {
 
 export const isExternalUrl = (url: string) => {
   try {
-    return new URL(url).host !== window.location.host;
+    return new URL(url, window.location.origin).host !== window.location.host;
```
mike-lvov, line 22: Could you please describe what is the difference here?
Why do we need to add it?

yurii-lubynets: We use this approach in other method, discussed here: https://github.com/sparkletown/sparkle/pull/1296#discussion_r604609316

mike-lvov (approved):

0xdevalias (commented):
[Retroactive Review] Fix looks good!

PR #1315, opened 2021-04-09
Increases the max char count

mike-lvov:
Increases the max char count

closes https://github.com/sparkletown/internal-sparkle-issues/issues/644


0xdevalias (commented):
[Retroactive Review] Approved.

Though it would be better to refactor these 3 disparate validation locations to use the same shared 'validation chunk' for this field, rather than having it manually duplicated like this.

PR #1317, opened 2021-04-12
Fix user profile + reactions z-index layering issue

mike-lvov:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/551

### DIFF
--- a/src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
+++ b/src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
@@ -3,7 +3,7 @@ import React, { useMemo } from "react";
 import { WithId } from "utils/id";
 import { User } from "types/User";
 
-import UserProfilePicture from "components/molecules/UserProfilePicture";
+import { UserProfilePicture } from "components/molecules/UserProfilePicture";
 
 interface MapPartygoerOverlayProps {
   partygoer: WithId<User>;
@@ -43,6 +43,7 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
       top: `${avatarHeight * (currentRow - 1)}%`,
       left: `${avatarWidth * (currentCol - 1)}%`,
       justifyContent: "center",
+      zIndex: "12",
     }),
     [avatarHeight, avatarWidth, currentCol, currentRow]
   );
--- a/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
+++ b/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
@@ -15,7 +15,13 @@ export const Avatar = styled.div<AvatarProps>`
     backgroundImage ?? DEFAULT_PROFILE_IMAGE});
 `;
 
-// --- Reaction
+// --- Reaction Emoji
+type ReactionProps = {
+  reactionPosition?: "right" | "left" | undefined;
+};
+
+const reactionZIndex = 12;
+
 const reactionOffset = "-20px";
 const reactionLeft = css`
   left: ${reactionOffset};
@@ -24,9 +30,6 @@ const reactionRight = css`
   right: ${reactionOffset};
 `;
 
-type ReactionProps = {
-  reactionPosition?: "right" | "left" | undefined;
-};
 export const Reaction = styled.div<ReactionProps>`
   width: 50px;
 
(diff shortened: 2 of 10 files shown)


src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -35,36 +38,14 @@ export const Reaction = styled.div<ReactionProps>`
     reactionPosition === "right" ? reactionRight : reactionLeft};
   top: -25px;
   // @debt convert this to scss then use our z-index layer helper here
-  z-index: 10;
+  z-index: ${reactionZIndex};
```
0xdevalias, line 41: I believe emoji reactions and shout-outs explicitly had different layers by choice. This change seems to change/break that.

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -121,9 +102,29 @@ export const ShoutOutMessage = styled.div<ReactionProps>`
 
   background-color: rgba(255, 255, 255, 1);
   // @debt convert this to scss then use our z-index layer helper here
-  z-index: 12;
+  z-index: ${reactionZIndex};
```
0xdevalias, line 105: I believe emoji reactions and shout-outs explicitly had different layers by choice. This change seems to change/break that.

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -35,36 +38,14 @@ export const Reaction = styled.div<ReactionProps>`
...
-    border-radius: 10rem;
-  }
-
-  .on-right {
-    ${Reaction} {
-      right: unset;
-      left: -20px;
-    }
-  }
-`;
```
0xdevalias, line 65: Why was this moved? Presumably it worked fine where it was. Adds extra churn to this diff for little to no reason IMO.

src/components/templates/PartyMap/components/Map/Map.scss
```diff
@@ -117,6 +117,7 @@ $border-radius: 28px;
       width: 100%;
       height: 100%;
       border-radius: 100%;
+      z-index: 12;
```
0xdevalias, line 120: Given this is in a scss file, it can use the `z()` helper with a properly named layer.

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -43,6 +43,7 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
       top: `${avatarHeight * (currentRow - 1)}%`,
       left: `${avatarWidth * (currentCol - 1)}%`,
       justifyContent: "center",
+      zIndex: "12",
```
0xdevalias, line 46: Is the change added in `src/components/templates/PartyMap/components/Map/Map.scss` sufficient for this? If not, is there any reason we can't follow a similar pattern of setting the `z-index` layer within an scss class?

src/scss/constants.scss
```diff
@@ -53,7 +53,7 @@ $small-laptop-width: 980px;
 
 // @debt the following locations are using styled-components, so can't use our helper function yet
 //   pages/Admin/Room/Modal/RoomModal.styles.ts (z-index: 101)
-//   components/molecules/UserProfilePicture/UserProfilePicture.styles.ts (z-index: 10)
+//   components/molecules/UserProfilePicture/UserProfilePicture.styles.ts (z-index: 10, 12)
```
0xdevalias, line 56: What is the change here for? It seems if anything the `10` should be replaced with `12`, based on what I can see in the changes in `src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts`?

0xdevalias (changes requested):
Comments/suggestions/etc.

Not sure if the approach being taken here is necessarily the best way to resolve this issue.

See my notes on this issue in https://github.com/sparkletown/internal-sparkle-issues/issues/551#issuecomment-817428522

0xdevalias (changes requested):
Note that https://github.com/sparkletown/sparkle/pull/1318 fixes the core issue that this PR originally aimed to solve. It may still be worth keeping some of the other changes made in this PR though.

mike-lvov:
Given the layering issue is temporarily solved in https://github.com/sparkletown/sparkle/pull/1324 + we have a better refactor happening in https://github.com/sparkletown/sparkle/pull/1323, I don't think this PR has any value

PR #1323, opened 2021-04-13
refactor UserProfilePicture / Reactions / related to clean up tech-debt + improve performance

0xdevalias:
Fixes the following issues related to tech-debt:

- fixes https://github.com/sparkletown/internal-sparkle-issues/issues/459
  - > Refactor UserProfilePicture to simplify/separate unrelated concerns (eg. reactions)

Also fixes the following bugs:

- fixes https://github.com/sparkletown/internal-sparkle-issues/issues/655
  - > In the auditorium, avatars jump seats when clicking emojis or shouting out to the crowd

Also potentially/partially fixes the following issues which are most likely due to performance issues/the browser getting 'overwhelmed':

- partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/562
  - > Reactions could be a performance issue, needs investigation
- potentially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/563
  - > After staying in the auditorium for a long time, users get an error message that the "Page is nonresponsive" when they try to click the "Back to" button
- potentially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/608
  - > The emoji bar randomly disappears after turning it on
- potentially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/609
  - > Some web browsers automatically react to changes in the admin panel and others do NOT

And probably a bunch of miscellaneous other tech debt/performance/stylistic/etc things along the way.

## TODO

- [x] Refactor `src/hooks/useReactions.ts` to use a shared context/state, rather than individually subscribing to listeners every time it is called
- [x] Check the fixes made in https://github.com/sparkletown/sparkle/pull/1324 (particularly the `z-index` edge case mentioned in https://github.com/sparkletown/sparkle/pull/1324#discussion_r612776203) to ensure we have all of the same fixes here still
- [x] Cleanup/action all of the TODO comments/etc


src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
 @import "scss/constants.scss";
 
 $fadedWhite: rgba(255, 255, 255, 0.8);
-$borderWidth: 5px;
 
-.profile-picture-container {
-  position: relative;
+// TODO: this was copied from Audience.scss (do we want to move the variable to constants?)
+$avatarSize: max(4vh, 3vw);
```
0xdevalias, line 6: [TODO] this was copied from Audience.scss (do we want to move the variable to constants?)

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
...
+
+// TODO: should we move these to using CSS --variables rather than these SCSS vars
+$reactionContainerOffset: calc(-1 * #{$avatarSize});
+$reactionSize: calc(#{$avatarSize / 1.5});
+$reactionOffset: calc(-1 * #{$reactionSize} / 2);
+$shoutBounceTranslateXOffset: calc(#{$avatarSize} / 2);
```
0xdevalias, line 12: [TODO] should we move these to using CSS `--variables` rather than these SCSS vars

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
...
+    }
+  }
+
+  &__avatar {
+    width: $avatarSize;
+    height: $avatarSize;
+
+    background-color: transparent;
+    // TODO: how can we use a default image here if none set? Can we use CSS --variables here? I think we could..
+    //background-image: url(${({ backgroundImage }) => backgroundImage ?? DEFAULT_PROFILE_IMAGE});
```
0xdevalias, line 51: [TODO] how can we use a default image here if none set? Can we use CSS --variables here? I think we could..

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
...
+    //top: 0;
+    //left: 0;
+    width: $avatarSize;
+    height: $avatarSize;
+    //width: 100%;
+    //height: 100%;
+
+    // TODO: remove this
+    background-color: rgba(255, 0, 0, 0.2);
+    border: 1px solid #f00;
```
0xdevalias, line 73: [TODO] Cleanup

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
...
 
-    position: absolute;
-    top: 0;
-    left: 0;
-
-    background: $fadedWhite;
+    // TODO: this is how the old styled-components had it
+    //background-color: rgba(255, 255, 255, 1);
+    // TODO: this is how the old (potentially unused) scss had it
+    background-color: $fadedWhite;
```
0xdevalias, line 97: [TODO] Pick what colour we want these messages to be

@mcflyok Do you have input on this? Which is the correct/better choice here?

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
```diff
@@ -1,83 +1,152 @@
...
+  100% {
+    transform: scale(1);
+  }
+}
 
-    animation: pulse 4s ease-in-out infinite;
+// TODO: can we de-duplicate these with SCSS helper functions/etc?
+@keyframes shout-bounce--right {
+  0%,
+  100% {
```
0xdevalias, line 122: [TODO] can we de-duplicate these with SCSS helper functions/etc?

Currently the `--left` and `--right` variants are practically identical, it would be nice to not have to duplicate them.

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -1,138 +0,0 @@
-import styled, { css, keyframes } from "styled-components";
```
0xdevalias, line 1: 💥👋🏻

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+  "/avatars/" +
+  RANDOM_AVATARS[Math.floor(id?.charCodeAt(0) % RANDOM_AVATARS.length)];
+
+// TODO: refactor this to accept named props instead of positional args
+const avatarUrl = (
+  id: string,
+  anonMode?: boolean,
+  pictureUrl?: string,
+  miniAvatars?: boolean
+): string => {
```
0xdevalias, line 32: [TODO] refactor this to accept named props instead of positional args

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+  }
+
+  if (!miniAvatars && pictureUrl) {
+    return pictureUrl;
+  }
+
+  // TODO: how is this intended to be used? Why doesn't it use the profile image? It seems to be something set on venue config..
+  if (miniAvatars) {
+    return randomAvatarUrl(id);
+  }
```
0xdevalias, line 44: [TODO]: how is `miniAvatars` intended to be used? Why doesn't it use the profile image? It seems to be something set on venue config..

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
   miniAvatars?: boolean;
   avatarClassName?: string;
   avatarStyle?: object;
   containerStyle?: object;
-  reactionPosition?: "right" | "left" | undefined;
+  reactionPosition?: "left" | "right";
 }
 
 // @debt This component should be divided into a few with simpler logic. Also, remove `styled components`
 // @debt the UserAvatar component serves a very similar purpose to this, we should unify them as much as possible
```
0xdevalias, line 60: [TODO]
- This component should be divided into a few with simpler logic. Also, remove `styled components`
- the UserAvatar component serves a very similar purpose to this, we should unify them as much as possible

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+export const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
   isAudioEffectDisabled = true,
   miniAvatars = false,
-  avatarClassName = "profile-icon",
-  avatarStyle,
-  containerStyle,
-  reactionPosition,
+  avatarClassName = "UserProfilePicture__avatar",
+  avatarStyle, // TODO: do we need this prop? Can we remove it? Or at least rename it? Only seems to be used in MapPartygoersOverlay
+  containerStyle, // TODO: do we need this prop? Can we remove it? Or at least rename it? Only seems to be used in MapPartygoersOverlay
```
0xdevalias, line 66: [TODO] do we need this prop? Can we remove it? Or at least rename it? Only seems to be used in `MapPartygoersOverlay`

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+  // TODO: extract this logic into a generic custom reactions component + improve it?
   const reactions = useReactions(venueId);
-
+  // @debt some of the redux patterns exist for this, but I don't believe anything actually uses them/calls this at the moment. Used in MapPartygoersOverlay
+  const muteReactions = useSelector((state) => state.room.mute);
   const typedReaction = reactions ?? [];
-
   const messagesToBand = typedReaction.find(
     (r) => r.reaction === "messageToTheBand" && r.created_by === user.id
   ) as MessageToTheBandReaction | undefined;
```
0xdevalias, line 81: [TODO] extract this logic into a generic custom reactions component + improve it?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
-    [user.id]
+  // TODO: I believe we only need this state to support the imageErrorHandler functionality.. can we just remove it?
+  const [pictureUrl, setPictureUrl] = useState(
+    avatarUrl(user.id, user.anonMode, user.pictureUrl, miniAvatars)
   );
+  useEffect(() => {
+    setPictureUrl(
+      avatarUrl(user.id, user.anonMode, user.pictureUrl, miniAvatars)
+    );
+  }, [miniAvatars, user.anonMode, user.id, user.pictureUrl]);
```
0xdevalias, line 91: [TODO] I believe we only need this state to support the `imageErrorHandler` functionality.. can we just remove it?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+  //   (event: React.SyntheticEvent<HTMLImageElement, Event>) => {
+  //     const randomAvatar = randomAvatarUrl(user.id);
+  //     setPictureUrl(randomAvatar);
+  //
+  //     // TODO: this isn't very reacty.. is it even needed..?
+  //     // event.currentTarget.onerror = null;
+  //     // event.currentTarget.src = randomAvatar;
+  //   },
+  //   [user.id]
+  // );
```
0xdevalias, line 109: [TODO] extract this logic into a generic custom component + improve it?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+    <div className={containerClasses} style={containerStyle}>
+      {/* TODO: extract this logic into a generic custom component + improve it? Or just remove it entirely? */}
+      {/* Hidden image, used to handle error if image is not loaded */}
+      {/*<img*/}
+      {/*  src={pictureUrl}*/}
+      {/*  onError={(event) => imageErrorHandler(event)}*/}
+      {/*  hidden*/}
+      {/*  style={{ display: "none" }}*/}
+      {/*  alt={user.anonMode ? DEFAULT_PARTY_NAME : user.partyName}*/}
+      {/*/>*/}
```
0xdevalias, line 141: [TODO] extract this logic into a generic custom component + improve it? Or just remove it entirely?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
+
+      {/* TODO: can we use src/components/atoms/UserAvatar/UserAvatar.tsx here? Should we? */}
+      <div
+        role="img"
+        aria-label={`${userDisplayName}'s avatar`}
+        className={avatarClasses}
+        style={avatarStyles}
+        onClick={openProfileModal}
+      />
```
0xdevalias, line 150: [TODO] can we use `src/components/atoms/UserAvatar/UserAvatar.tsx` here? Should we?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
               (r) => r.created_by === user.id && r.reaction === reaction.type
             ) && (
-              <div key={index} className="reaction-container">
-                <S.Reaction
+              // TODO: don't use index as a key
+              // TODO: should we extract reactions into their own sub-components?
+              // TODO: is reaction-container even defined in our styles here..?
+              <React.Fragment key={index}>
+                <div
+                  className="UserProfilePicture__reaction"
```
0xdevalias, line 164: [TODO] refactor this to be more performant and potentially memo it if required

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
                   {reaction.text}
-                </S.Reaction>
+                </div>
 
+                {/* TODO: replace this with useSound or similar */}
                 {!muteReactions && !isAudioEffectDisabled && (
                   <audio autoPlay loop>
                     <source src={reaction.audioPath} />
                   </audio>
                 )}
```
0xdevalias, line 176: [TODO] replace this with `useSound` or similar

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +17,176 @@ import {
...
-  ]);
+      </div>
+    </div>
+  );
 };
 
+/**
+ * @deprecated Prefer using named exports instead of default exports
+ */
 export default UserProfilePicture;
```
0xdevalias, line 192: [TODO] Remove this + update all locations across the app that use it.

0xdevalias (commented):
[TODO] Calling out the various TODO notes/etc in this PR so far.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -190,13 +181,11 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
               {seatedAtTable && (
                 <div className="actions-container">
                   <div className="emoji-container">
-                    {Reactions.map((reaction) => (
+                    {EmojiReactions.map((reaction) => (
```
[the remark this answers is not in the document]
0xdevalias: This is a valid finding, the code is duplicated, but resolving that is OOS of this PR.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -89,15 +72,23 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
...
-    dispatch(
-      addReaction({
-        venueId,
-        reaction: createReaction({ reaction }, user),
-      })
-    );
-    setTimeout(() => (document.activeElement as HTMLElement).blur(), 1000);
-  };
+  // @debt de-duplicate this with version in src/components/templates/Audience/Audience.tsx
+  const reactionClicked = useCallback(
```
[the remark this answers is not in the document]
0xdevalias: This is a valid finding, the code is duplicated, but resolving that is OOS of this PR.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/2dc683db970609e234f3abb99b1a70d4485af6e8

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/57a6edbbb517f53372760da5a653d9c9471137ad + https://github.com/sparkletown/sparkle/pull/1323/commits/d198f74f1b3c0ff127e8f811fd9694431157e83e + https://github.com/sparkletown/sparkle/pull/1323/commits/c7de18da0c43f2fa81a524a3a4328fbeccc6b21f

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: It looks like this is only used in ~5 files currently.. so wouldn't create too much noise to do it.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/14f3867baa069138b1e6510c45f6e09f9f08b7d4

0xdevalias:
@sparkletown/engineering While this still needs some cleanup/polish/final testing/etc, the bulk of the 'main big bits' are done now, so could probably get a first pass of review.

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: Calling this OOS

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/f8b0fcbaf3ffd67f4a1226baea574719e9b22d65

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/ca403f5f39207837166c05ffa2b4c288e3d9a96c

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/9ae02ba722c2f12213e8670471155b083d1b2c51

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Potentially OOS to change this

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: See also https://github.com/sparkletown/sparkle/pull/1323/commits/db15bd47191b6c10594877933b651616c3375b67 + https://github.com/sparkletown/sparkle/pull/1323/commits/c3a2f885a4b5391c78e677009d55ba47238ab702 + https://github.com/sparkletown/sparkle/pull/1323/commits/f1469b499cbbb22ffcc4acb88790de65a3e1ddd8

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/db15bd47191b6c10594877933b651616c3375b67 + https://github.com/sparkletown/sparkle/pull/1323/commits/c3a2f885a4b5391c78e677009d55ba47238ab702 + https://github.com/sparkletown/sparkle/pull/1323/commits/f1469b499cbbb22ffcc4acb88790de65a3e1ddd8

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/db15bd47191b6c10594877933b651616c3375b67 + https://github.com/sparkletown/sparkle/pull/1323/commits/c3a2f885a4b5391c78e677009d55ba47238ab702 + https://github.com/sparkletown/sparkle/pull/1323/commits/f1469b499cbbb22ffcc4acb88790de65a3e1ddd8

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/db15bd47191b6c10594877933b651616c3375b67 + https://github.com/sparkletown/sparkle/pull/1323/commits/c3a2f885a4b5391c78e677009d55ba47238ab702 + https://github.com/sparkletown/sparkle/pull/1323/commits/f1469b499cbbb22ffcc4acb88790de65a3e1ddd8

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Calling this OOS, converted it to an `@debt` for future

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: OOS, marked this as `@debt` https://github.com/sparkletown/sparkle/pull/1323/commits/af6335be218415776a2481a7ad2a1aebb7d343b0

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Potentially OOS

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Cleaned up comments in https://github.com/sparkletown/sparkle/pull/1323/commits/e7d9b0644fa1364a170b627dae905206a6da7121

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: Potentially OOS

0xdevalias:
As per https://github.com/sparkletown/sparkle/pull/1323#issuecomment-820211942

> @sparkletown/engineering While this still needs some cleanup/polish/final testing/etc, the bulk of the 'main big bits' are done now, so could probably get a first pass of review.

src/components/molecules/CallOutMessageForm/CallOutMessageForm.tsx
```diff
@@ -10,6 +10,7 @@ interface PropsType {
   isMessageToTheBandSent: boolean;
 }
 
+// @debt This file doesn't appear to be used anywhere anymore. We can probably remove it in our next tech-debt cleanup pass
```
sunny-viktoryia, line 13: you can also check 
`src/components/molecules/UserSearchBar/...`
seems the files in the directory are also not used anywhere.

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
mike-lvov: Yes, I would say that moving that to constant make sense, since we already reuse it in at least 2 places. (will be 3 with new auditorium)

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +14,108 @@ import {
 
 import { User } from "types/User";
 
-// @debt remove styled-components in favour of using our standard scss patterns
-import * as S from "./UserProfilePicture.styles";
+import { UserReactions } from "./UserReactions";
+
 import "./UserProfilePicture.scss";
 
+const randomAvatarUrl = (id: string) =>
```
mike-lvov, line 21: generate/getRandomAvatarUrl?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +14,108 @@ import {
...
-// @debt the UserAvatar component serves a very similar purpose to this, we should unify them as much as possible
-const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
+export const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
   isAudioEffectDisabled = true,
   miniAvatars = false,
-  avatarClassName = "profile-icon",
-  avatarStyle,
-  containerStyle,
-  reactionPosition,
+  avatarClassName = "UserProfilePicture__avatar",
```
mike-lvov, line 62: I guess this should not have a default since `UserProfilePicture__avatar` is already used in avatarClasses

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +14,108 @@ import {
...
+const randomAvatarUrl = (id: string) =>
+  "/avatars/" +
+  RANDOM_AVATARS[Math.floor(id?.charCodeAt(0) % RANDOM_AVATARS.length)];
+
+export interface AvatarUrlProps {
+  user: WithId<User>;
+  miniAvatars?: boolean;
+}
+
+const avatarUrl = ({ user, miniAvatars }: AvatarUrlProps): string => {
```
mike-lvov, line 30:
```suggestion
const getAvatarUrl = ({ user, miniAvatars }: AvatarUrlProps): string => {
```

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -16,154 +14,108 @@ import {
...
+          reactionPosition={reactionPosition}
+        >
+          {/* @debt can we use src/components/atoms/UserAvatar/UserAvatar.tsx here? Should we? */}
+          <div
+            role="img"
+            aria-label={`${userDisplayName}'s avatar`}
+            className={avatarClasses}
+            style={avatarStyles}
+            onClick={openProfileModal}
+          />
```
mike-lvov, line 116: My opinion is that we should reuse it here. But Maybe we could change the name of `UserProfilePicture`, because it seems to be a little bit more than that. I don't have any suggestions yet though

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
+@import "scss/constants";
+
+// TODO: should we move these to using CSS --variables rather than these SCSS vars
+$reactionSize: max(2.5vh, 2vw);
```
mike-lvov, line 4: I think that we probably can get `$reactionSize`, based on `$avatarSize` since they are used in tandem. Wdyt?

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
...
+
+  &--reaction-left {
+    .UserReactions__reaction {
+      left: $reactionOffset;
+    }
+
+    .UserReactions__shout {
+      right: 50%;
+      transform-origin: right center;
+      animation: shout-bounce--left 5s ease;
```
mike-lvov, line 23: move into a var?

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
...
+
+  &--reaction-right {
+    .UserReactions__reaction {
+      right: $reactionOffset;
+    }
+
+    .UserReactions__shout {
+      left: 50%;
+      transform-origin: left center;
+      animation: shout-bounce--right 5s ease;
```
mike-lvov, line 35: move into a var?

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
...
+  &__reaction {
+    position: absolute;
+    top: $reactionOffset;
+
+    z-index: z(user-reactions-emoji);
+
+    // Note: emoji reactions are sized with font-size
+    font-size: $reactionSize;
+
+    animation: reaction-pulse 4s ease-in-out infinite;
```
mike-lvov, line 48: move into a var?

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
...
+    position: absolute;
+    bottom: 0;
+
+    z-index: z(user-reactions-shout);
+
+    width: max-content;
+    max-width: 20em;
+    padding: 6px 10px;
+
+    background-color: rgba(255, 255, 255, 1);
```
mike-lvov, line 61: $white?

src/components/molecules/UserProfilePicture/UserReactions.scss
```diff
@@ -0,0 +1,115 @@
...
+    color: $black;
+    font-size: 20px;
+    border-radius: 10px;
```
mike-lvov, line 64: I think that we should come up with a convention on how we approach border-radiuses/font-sizes/paddings/margins units. px/rem/em/??? + maybe we can create a styleguide with predefined values of those? Like the golden ratio for margins/padding. @mcflyok wdyt?

src/components/molecules/UserProfilePicture/UserReactions.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    const renderedEmojiReactions = Array.from(
+      userUniqueEmojiReactions.values()
+    ).map((emojiReaction) => (
+      <DisplayEmojiReaction
+        key={emojiReaction.type}
+        emojiReaction={emojiReaction}
+        isMuted={isMuted}
+      />
+    ));
```
mike-lvov, line 53: Why do we not create an array in the reduce?

src/components/templates/Audience/Audience.tsx
```diff
@@ -138,14 +132,20 @@ const requiredAuditoriumSize = (
...
 // Note: This is the component that is used for the Auditorium
 export const Audience: React.FunctionComponent = () => {
   const venueId = useVenueId();
-  const { user, profile } = useUser();
+  const { userWithId, profile } = useUser();
   const venue = useSelector(currentVenueSelectorData);
   const { recentVenueUsers } = useRecentVenueUsers();
 
-  const userUid = user?.uid;
+  const userUid = userWithId?.id;
```
mike-lvov, line 148: I've seen so many times that we tend to only import user to retrieve his id. Is it worth to just return userId var from the hook?

src/components/templates/Audience/Audience.tsx
```diff
@@ -313,27 +314,22 @@ export const Audience: React.FunctionComponent = () => {
...
-            ),
-          })
-        );
+
+      dispatch(
+        addReaction({
+          venueId,
+          reaction: createTextReaction(data.text, userWithId),
+        })
+      );
```
mike-lvov, line 326: Should this be moved into the hook with dispatch wrapping in the hook? So that we can use a clean `addTextReaction` fn here & everywhere else?

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -89,15 +72,23 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
...
+
+      dispatch(
+        addReaction({
+          venueId,
+          reaction: createEmojiReaction(emojiReaction, userWithId),
+        })
+      );
+
+      // @debt Why do we have this here..? We probably shouldn't have it/need it? It's not a very Reacty thing to do..
+      setTimeout(() => (document.activeElement as HTMLElement).blur(), 1000);
```
mike-lvov, line 88: 😲😲😲😲. It's not healthy to review things just before going to bed.

[offtop], feel free to resolve 😄

src/components/templates/ReactionPage/ReactionList.tsx
```diff
@@ -71,11 +72,11 @@ export const ReactionList: React.FC<ReactionListProps> = ({
...
           <div
             className={classNames("message-bubble", {
-              emoji: message.reaction !== "messageToTheBand",
+              emoji: isEmojiReaction(message),
             })}
           >
-            {message.reaction !== "messageToTheBand"
-              ? ReactionsTextMap[message.reaction]
+            {isEmojiReaction(message)
+              ? EmojiReactionsMap.get(message.reaction)?.text ?? null
```
mike-lvov, line 79: Why do we need `?? null`?

src/hooks/useReactions.ts
```diff
@@ -1,41 +1,74 @@
...
+  user?: WithId<User>;
+}
 
 // @debt refactor this to use useConnect like in src/components/templates/ReactionPage/ReactionPage.tsx ?
-export const useReactions = (venueId?: string) => {
+export const useReactions = ({ venueId, user }: UseReactionsProps) => {
   const [reactions, setReactions] = useState<Reaction[]>([]);
   const firebase = useFirebase();
 
+  const userId = user?.id;
```
mike-lvov, line 21: I guess we only use userId here? Then, it makes sense to only pass id

mike-lvov (changes requested):
A few suggestions/questions/ideas. It's for sure a good chunk of work done here.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Agreed. This naming was legacy, but didn't sit right with me, just hadn't thought of something better.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Agreed. Not sure why I did it this way originally, might have been intending to allow users to override it completely, then changed my mind and figured that wasn't a good idea.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Yeah, I was definitely leaning towards reuse it as well.. but then realised it would need some refactors to support the use case here (which isn't too hard), and then also realised that we kept it as a 'dump component' so that the status would also need to be passed into it (which again, probably isn't a big concern, as we can make that prop optional if it isn't already)

I'll see how much time is left in this sprint after I rework the `useReactions` logic for performance and try and fit this in. If not, we can handle it in a followup PR.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: I actually had the code that way originally, but then sort of changed my mind on it to remove the 'dependency' on avatar size. I think we could pull this (and likely avatar size) back out into our main constants, at which point it would make sense to me to use that way. Then we can 'configure them all in the same place'.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Which aspect would you see as being useful to have in a var here out of curiosity?

In testing, I think that the `5s` here corresponds with `REACTION_TIMEOUT` from `src/settings.ts`, and I expect they probably need to be kept in sync. Given that, it might make sense to make this a `--var` and pass it through from the JS side of things.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1323#discussion_r617972735

Whatever we decide for that will apply here also.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1323#discussion_r617972735

Whatever we decide for that might apply here also.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: I was originally keeping it this way to match how it was, but given it's 100% white, makes sense to just go with the basic `#fff` rather than `rgba`, so yeah, `$white` will work here.

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Definitely agree RE: standardising this + common helpers in `scss/constants.scss` or similar rather than defining it arbitrarily all over the codebase. I believe we already have some variables for this (though it's not always clear when they should be applied, so often I feel like we end up sticking with the legacy definitions by default)

@sunny-viktoryia Made some changes in https://github.com/sparkletown/sparkle/pull/1328 to introduce `$border-radius--small`, etc

- https://github.com/sparkletown/sparkle/blob/bec9ea233a7130a836079592bff2df7aae3a3734/src/scss/constants.scss#L49-L50

Also added SCSS mixins for things like scrollbar styling, since it was repeated all over the codebase:

- https://github.com/sparkletown/sparkle/blob/bec9ea233a7130a836079592bff2df7aae3a3734/src/scss/constants.scss#L128-L148

I'd probably like us to use SCSS variables/mixins/functions/etc where it makes sense to. CSS `--variables` where it makes sense to as well.

My main 'desire' is to keep styling changes to styling; i'm generally not really a fan of 'utility CSS classes' that end up 'infecting' the structural layer of the app code/JS, rather than including it in a specific CSS class 'tied' to that feature. That said, I am open to hearing/being convinced otherwise. I think my main 'concern' here is that if we are applying a specific class to a thing, that specific class should have all of the relevant styles. It shouldn't have to also be combined with additional utility classes in the JSX. That's where mixins/etc become useful IMO. Though in cases where an element doesn't 'need' a specific class for it, and the only thing we want to do is something 'generic', then utility classes might make sense there.

0xdevalias: (captured the above in https://github.com/sparkletown/internal-sparkle-issues/issues/4#issuecomment-824454619)

src/components/molecules/UserProfilePicture/UserReactions.tsx
0xdevalias: I believe I'm using the `Map` as the way of making it unique values. If they were primitive values then I could have used `Set`, but I don't believe that will work as well given they're complex objects.

Given this is the only place that uses this/needs it as an array, it didn't seem like something that was worth further abstracting into another util function/etc.

src/components/templates/Audience/Audience.tsx
0xdevalias: Yeah, that seems like a reasonable change to make. Might be OOS of this PR though, as to do it properly we should clean up all the places in the codebase that are using this pattern at the same time.

src/components/templates/Audience/Audience.tsx
0xdevalias: This is sort of OOS of this PR. IMO it belongs within the scope of @yurii-lubynets's https://github.com/sparkletown/internal-sparkle-issues/issues/637 / https://github.com/sparkletown/sparkle/pull/1339

0xdevalias: (referenced this in https://github.com/sparkletown/internal-sparkle-issues/issues/637#issuecomment-824456706)

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: I believe I was basing it on the following:

I wasn't sure that `undefined` was a valid usage for this:

- https://github.com/facebook/react/issues/1647#issuecomment-64447594

But it seems like it's fine:

- https://stackoverflow.com/questions/42083181/is-it-possible-to-return-empty-in-react-render-function/42083262#42083262
  - https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

So we can simplify the code here.

src/hooks/useReactions.ts
0xdevalias: As the prop? Yeah, that makes sense. I only just changed the code to extract this last night, before that I think I assumed I was going to be needing more than just the `userId` here.

src/components/molecules/CallOutMessageForm/CallOutMessageForm.tsx
0xdevalias: @sunny-viktoryia Commented on your PR, probably makes sense for you to remove it there.

0xdevalias:
After refactoring `useReactions` to use React context + subscribe once (rather than subscribe once per user) in https://github.com/sparkletown/sparkle/pull/1323/commits/4bee204f440ffa4a6f22cd0d04276ff3c70f2fa3 + https://github.com/sparkletown/sparkle/pull/1323/commits/76e5874ffc3c667527cb84cf7855cf3364fa14da, I figured I would take some naive stats to "prove it's going to help".

In each of the following tests, I basically sent 2-3 reactions + a shoutout while recording with the React profiler.

I also manually hacked the code to render my user in every seat, to slightly simulate a packed auditorium. I don't believe this 'hack test' is equivalent to having that many users all send a unique reaction at the same time, but it should hopefully at least assist in demonstrating the issue.

In the legacy code, each individual user rendered here would set up their own connection to firebase, listen for new reactions, store that data in their own internal state, etc. In the new code, this is all handled once in the provider + some of the methods used within that were made more efficient as well.

image

---

## Before Refactor (`useReactionsLegacy`)

Notice there are 1699 'render snapshots', which is more or less equivalent to 1699 re-renders/repaints/screen updates. This will take a LOT of work for the browser to handle, and will be noticeably slow/laggy for users.

Screen Shot 2021-04-22 at 4 06 21 pm

In this video, notice how the reactions seem to 'cascade', and they don't all change at once; as well as everything feeling quite slow/laggy:

https://user-images.githubusercontent.com/753891/115665364-8eb36a00-a386-11eb-8712-028eafee00eb.mov

## After Refactor (`useReactionsLegacy`)

Notice that for a more or less equivalent test (a couple of emoji reactions + a text shoutout) we now only have 19 'render snapshots'. A dramatic improvement!

Screen Shot 2021-04-22 at 4 05 18 pm

In this video, notice how all of the reactions seem to happen at the same time; and things seem a lot more responsive:

https://user-images.githubusercontent.com/753891/115665601-dc2fd700-a386-11eb-99b0-7c21e74d0fb3.mov

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.scss
```diff
@@ -0,0 +1,31 @@
...
+
+$gridTotalRows: var(--map-partygoer-overlay-total-rows, 1);
+$gridTotalColumns: var(--map-partygoer-overlay-total-columns, 1);
+$row: var(--map-partygoer-overlay-row, 0);
+$column: var(--map-partygoer-overlay-column, 0);
+
+$avatarWidth: calc(100% / #{$gridTotalRows});
+$avatarHeight: calc(100% / #{$gridTotalColumns});
+$avatarTop: calc((#{$row} - 1) * #{$avatarWidth});
+$avatarLeft: calc((#{$column} - 1) * #{$avatarHeight});
```
0xdevalias, line 11: [FYI] Calling out this pattern of using CSS `--variables` to pass dynamic content from the JS into SCSS, so that we can keep our styles in SCSS rather than inline styles.

(cc // @sparkletown/engineering FYI: you'll likely see more of this pattern moving forward)

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -26,47 +31,28 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
...
-      borderRadius: "100%",
-      alignSelf: "center",
-      backgroundImage: `url(${pictureUrl})`,
-      backgroundSize: "cover",
-    }),
-    [pictureUrl]
+  const containerClasses = classNames(
+    "MapPartygoerOverlay__avatar-container",
+    containerVars
   );
```
0xdevalias, line 47: [FYI] Calling out this pattern of using CSS `--variables` to pass dynamic content from the JS into SCSS, so that we can keep our styles in SCSS rather than inline styles.

(cc // @sparkletown/engineering FYI: you'll likely see more of this pattern moving forward)

0xdevalias (commented):
[FYI] Calling out/showing one of our newer'ish patterns + what it allows us to do.

src/components/molecules/CallOutMessageForm/CallOutMessageForm.tsx
sunny-viktoryia: the files are removed: https://github.com/sparkletown/sparkle/commit/2574d407862199372cf770036bc25c76283e698d

0xdevalias: Cool, thanks :)

TroySK:
Events which were cascading have lowered but the major issue around hooks updating remains. This is good progress, hopefully more interactions can be supported by browsers now.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/134a48d8abe7e0a93cf6e9045b5c6bc2d57f10a5

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Going to call this OOS for now in the interest of landing the main fixes in the PR (otherwise will spend too much time on micro polish)

We can follow up this PR in future to use standard values across the codebase + update all the locations at once.

src/components/molecules/UserProfilePicture/UserReactions.tsx
0xdevalias: Marking this resolved based on the above.

src/components/templates/Audience/Audience.tsx
0xdevalias: Added to `useUser` in https://github.com/sparkletown/sparkle/pull/1323/commits/e4fae8398c59a12504f0f0322c1d20bf0ae5895a

Fixed in `Audience` in https://github.com/sparkletown/sparkle/pull/1323/commits/7f0afea8265473ef7ec84c8cb83041973281a3c1

Refactoring the rest of the codebase in this PR is OOS

src/components/templates/ReactionPage/ReactionList.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/e8528100e2e3e41174a0e2af043797a858dbf317

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/249f26615b9b3232440656652a9042c514d3ccea

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/5efb685f8665173e76d562c8896cae5b5d31cb12

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/5efb685f8665173e76d562c8896cae5b5d31cb12

src/hooks/useReactions.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/commit/119bfae9a3770b7a2f8f14f28b06fcb212f55d97

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: See @mike-lvov's comments about this in https://github.com/sparkletown/sparkle/pull/1323#discussion_r617811235

Marking this thread resolved.

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/f0d70b78566230b1b6a33be0386e68aa45219450

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/a6a5eed78c56d517a3f5c00380a5ee775f37e97f + https://github.com/sparkletown/sparkle/pull/1323/commits/134a48d8abe7e0a93cf6e9045b5c6bc2d57f10a5 + https://github.com/sparkletown/sparkle/pull/1323/commits/0e71fc5bc49b5bbc0ecbaed763eb0d884c640374 + https://github.com/sparkletown/sparkle/pull/1323/commits/59073433104b9051729d8a4b6d83957c43fa4f76

src/components/molecules/UserProfilePicture/UserReactions.scss
mike-lvov: Oh, I think at first I thought they were identical, but then I spotted --left/--right modifiers. I guess time of the reaction can be reused

0xdevalias: Makes sense. Can probably pull that out into a var (or even a `--var` as mentioned above maybe)

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -1,169 +1,115 @@
...
+//   }
+//
+//   // @debt how is this intended to be used? Why doesn't it use the profile image? It seems to be something set on venue config..
+//   if (miniAvatars) {
+//     return generateRandomAvatarUrl(userId);
+//   }
+//
+//   return DEFAULT_PROFILE_IMAGE;
+// };
+
```
mike-lvov, line 43: Is this going to be deleted? I think it needs some kind of comment

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -1,169 +1,115 @@
...
-  avatarClassName = "profile-icon",
-  avatarStyle,
-  containerStyle,
-  reactionPosition,
+export const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
   user,
+  isAudioEffectDisabled = true,
+  containerClassName,
+  reactionPosition = "right",
+  // miniAvatars = false,
```
mike-lvov, line 61: Deleted/comment?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -1,169 +1,115 @@
...
+  // const { loadedImageUrl: pictureUrl } = useImage({
+  //   src: avatarUrl({ user, miniAvatars }),
+  //   fallbackSrc: () => randomAvatarUrl(user.id),
+  // });
+
+  // const containerVars = useCss({
+  //   "--user-profile-picture-avatar-url": pictureUrl
+  //     ? `url(${pictureUrl})`
+  //     : `url(${DEFAULT_PROFILE_IMAGE})`,
+  // });
```
mike-lvov, line 85: Deleted/comment?

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -1,169 +1,115 @@
...
-        {/* Hidden image, used to handle error if image is not loaded */}
-        <img
-          src={pictureUrl}
-          onError={(event) => imageErrorHandler(event)}
-          hidden
-          style={{ display: "none" }}
-          alt={user.anonMode ? DEFAULT_PARTY_NAME : user.partyName}
+  // const userDisplayName: string = user?.anonMode
+  //   ? DEFAULT_PARTY_NAME
+  //   : user?.partyName ?? DEFAULT_PARTY_NAME;
```
mike-lvov, line 95: Deleted/comment?

src/components/templates/Audience/Audience.tsx
```diff
@@ -300,11 +300,11 @@ export const Audience: React.FunctionComponent = () => {
   // @debt this return useMemo antipattern should be rewritten
   return useMemo(() => {
     const takeSeat = (row: number | null, column: number | null) => {
-      if (!venueId || !userUid) return;
+      if (!venueId || !userId) return;
 
       makeUpdateUserGridLocation({
         venueId,
-        userUid,
+        userUid: userId,
```
mike-lvov, line 307: [OOS] Do we agree on using userId everywhere with deprecating userUid?

mike-lvov (approved):
Valid after deleted/addressing the commented code. 🎉 🎉 🎉

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Unsure if deleted or debt comment yet. It seemed to work well at first, then stopped working well later. Not sure what happened, but it seemed to end up having to reload the image whenever a user moved/etc. At the very least will add the comment.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: There’s a comment in the type definition about it already. Does that suffice?

src/components/templates/Audience/Audience.tsx
0xdevalias: Personally I think we should. We only ever used userUid to align with what firebase’s underlying auth calls it, but that’s not a good reason to use it IMO; and our far more common patter n has been userId

0xdevalias: Added `@debt` comments to explain the commented code blocks in https://github.com/sparkletown/sparkle/pull/1323/commits/5f2acbd92b4f96c175bf39ab1ffc0c0bb7d92624

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Added `@debt` comments to explain the commented code blocks in https://github.com/sparkletown/sparkle/pull/1323/commits/5f2acbd92b4f96c175bf39ab1ffc0c0bb7d92624

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Added `@debt` comments to explain the commented code blocks in https://github.com/sparkletown/sparkle/pull/1323/commits/5f2acbd92b4f96c175bf39ab1ffc0c0bb7d92624

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Added `@debt` comments to explain the commented code blocks in https://github.com/sparkletown/sparkle/pull/1323/commits/5f2acbd92b4f96c175bf39ab1ffc0c0bb7d92624

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: Added `@debt` comments to explain the commented code blocks in https://github.com/sparkletown/sparkle/pull/1323/commits/5f2acbd92b4f96c175bf39ab1ffc0c0bb7d92624

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/be40ca41c3a1a4d06cc9e5dbc42f854c7d3c76d1 + https://github.com/sparkletown/sparkle/pull/1323/commits/d4946deea7fa550f4c6b2158a2b73f66a76dc06a

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/10de65755feef5989496b42a13c6b741df68e6d7

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/10de65755feef5989496b42a13c6b741df68e6d7

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/696b5e95e6d8c37a098b875e076d99e2bd934dbb

src/components/molecules/UserProfilePicture/UserReactions.scss
0xdevalias: Decided not to directly define them in terms of one another, but refactored both into `scss/constants` in https://github.com/sparkletown/sparkle/pull/1323/commits/3c6adcbfde6c7fc244a209afeac7de7460f79a05

src/components/molecules/UserProfilePicture/UserProfilePicture.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1323/commits/3c6adcbfde6c7fc244a209afeac7de7460f79a05

src/components/templates/Audience/Audience.tsx
mike-lvov: Sounds good 👍

PR #1324, opened 2021-04-13
Fix avatar position ‘jumping around’ when reactions used in Auditorium (v1)

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/655

### DIFF
--- a/src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
+++ b/src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
@@ -43,6 +43,7 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
       top: `${avatarHeight * (currentRow - 1)}%`,
       left: `${avatarWidth * (currentCol - 1)}%`,
       justifyContent: "center",
+      zIndex: "11",
     }),
     [avatarHeight, avatarWidth, currentCol, currentRow]
   );
--- a/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
+++ b/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
@@ -38,7 +38,7 @@ export const Reaction = styled.div<ReactionProps>`
 
   ${({ reactionPosition }) =>
     reactionPosition === "right" ? reactionRight : reactionLeft};
-  top: -25px;
+  bottom: 0;
 
   font-size: 50px;
 
@@ -118,6 +118,7 @@ const messageRight = css`
 export const ShoutOutMessage = styled.div<ReactionProps>`
   bottom: 0;
   position: absolute;
+  z-index: 1;
 
   width: max-content;
   max-width: 20em;
--- a/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
+++ b/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
@@ -33,8 +33,6 @@ type ReactionProps = {
 
 export const Reaction = styled.div<ReactionProps>`
   position: absolute;
-  // @debt convert this to scss then use our z-index layer helper here
-  z-index: 10;
 
   width: 50px;
 
@@ -120,9 +118,6 @@ const messageRight = css`
 export const ShoutOutMessage = styled.div<ReactionProps>`
   bottom: 0;
   position: absolute;
-
-  // @debt convert this to scss then use our z-index layer helper here
-  z-index: 12;
 
   width: max-content;
   max-width: 20em;
--- a/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
+++ b/src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
@@ -32,7 +32,7 @@ type ReactionProps = {
 };
 
 export const Reaction = styled.div<ReactionProps>`
-  position: relative;
+  position: absolute;
   // @debt convert this to scss then use our z-index layer helper here
   z-index: 10;
 
@@ -49,13 +49,14 @@ export const Reaction = styled.div<ReactionProps>`
 
 // --- Reaction Container
 export const Container = styled.div`
-  // @debt convert this to scss then use our z-index layer helper here
-  z-index: 11;
-
   height: 100%;
 
   background-position: center;
   background-size: cover;
+
+  .reaction-container {
+    position: relative;
+  }
 
   ${Avatar} {
     border-radius: 10rem;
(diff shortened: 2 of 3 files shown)


yurii-lubynets (approved):

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
```diff
@@ -43,6 +43,7 @@ export const MapPartygoerOverlay: React.FC<MapPartygoerOverlayProps> = ({
       top: `${avatarHeight * (currentRow - 1)}%`,
       left: `${avatarWidth * (currentCol - 1)}%`,
       justifyContent: "center",
+      zIndex: "11",
```
0xdevalias, line 46: What is this needed for?

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -49,14 +49,15 @@ export const Reaction = styled.div<ReactionProps>`
 
 // --- Reaction Container
 export const Container = styled.div`
-  // @debt convert this to scss then use our z-index layer helper here
-  z-index: 11;
```
0xdevalias, line 53: I believe all of the other s-indexes should be removed from here as well

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -49,14 +49,15 @@ export const Reaction = styled.div<ReactionProps>`
...
-
   height: 100%;
 
   background-position: center;
   background-size: cover;
 
+  .reaction-container {
```
0xdevalias, line 57: There will be 2 instances of this rendered currently, one on shouts, one on emoji reactions. Unless you’re also doing things to position this over the top of the avatar, I don’t believe this will work correctly

0xdevalias:
@mike-lvov Have you tested this to ensure it fixes the original issue as well? There should be 3 distinct layers:

- avatars
- reactions
- shoutouts

There should be no ‘breaking out’ from these layers; eg. An avatar must never be above a reaction/emoji; a reaction/emoji must never be above a shoutout, etc.

src/components/molecules/MapPartygoerOverlay/MapPartygoerOverlay.tsx
mike-lvov: That's because of the way party map components are layered. Cells, rooms are absolutely positioned with explicit zIndices, so we need it for user avatars as well

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
mike-lvov: It works as expected or what do you mean?

image

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
mike-lvov: You are right.

Removed in https://github.com/sparkletown/sparkle/pull/1324/commits/6d727cdede703fa438334cfb513f48665d1dbe71

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
0xdevalias: Like in my screenshots on the original bug fix, when there are multiple reactions happening at the same time, the original ‘basic fix’ ended up having some of the avatars overlaying the emoji/etc; whereas in my latest fix, we ensure that they always stay on heir appropriate layers as described above.

Simplest way to check this is to get userWithId from useUser in Audience, then change the seat code to render that user in every seat.

Then send a reaction, it should be obvious of the layering is working correctly or not (would be good to share a screenshot here to show that)

0xdevalias: Actually, the specific thing I meant here, which ‘direction’ are the reactions meant to appear. I believe top/right. Yet in your screenshot here it’s showing more bottom/right.

If you render a background/border/similar on the reaction-container I expect you’ll find it’s sitting beneath the user avatar, rather than in top of it.

mike-lvov:
image

src/components/molecules/UserProfilePicture/UserProfilePicture.styles.ts
```diff
@@ -118,6 +118,7 @@ const messageRight = css`
 export const ShoutOutMessage = styled.div<ReactionProps>`
   bottom: 0;
   position: absolute;
+  z-index: 1;
```
0xdevalias, line 121: Is this needed here? Based on the natural stack ordering I believe it shouldn’t be, as the shoutout is rendered after the reactions?

I don’t think I ended up with any Z indexes in my WIP refactor PR in the end (though perhaps there is an edgecase I missed that’s still broken there)

mike-lvov: Without this, other peoples' shoutouts are behind the emojis(probably because they are in different relative containers).

mike-lvov: @0xdevalias shall I merge this into the prod to unblock our sales?
It's probably a hacky way, but it works. Given that the PR to have it clean will be merged in the nearest days

mike-lvov: One the edge cases is to have the user(1) shoutout over user(2)emojis, since the natural stack layering doesn't seem to work in this case

0xdevalias: Makes sense, I’ll have to check mine closer.

Yeah, if this is tested and resolves things, definitely get it shipped out across all the prod envs.

src/components/molecules/UserProfilePicture/UserProfilePicture.tsx
```diff
@@ -114,40 +114,40 @@ const UserProfilePicture: React.FC<UserProfilePictureProp> = ({
...
-        )}
-        {messagesToBand && (
-          <div className="reaction-container">
+        <div className="reaction-container">
+          {Reactions.map(
+            (reaction, index) =>
+              reactions.find(
+                (r) => r.created_by === user.id && r.reaction === reaction.type
+              ) && (
+                <>
```
0xdevalias, line 123: @mike-lvov FYI: This should have used the named `<Fragment>` component, which accepts a `key` prop:

- https://reactjs.org/docs/fragments.html#keyed-fragments

0xdevalias (commented):
[Retroactive Review] Looks good.

[Note] We are refactoring this component even further in https://github.com/sparkletown/sparkle/pull/1323

PR #1325, opened 2021-04-13
Deploy staging -> production

mike-lvov:
Deploys:
- Fix avatar position ‘jumping around’ when reactions


0xdevalias:
@mike-lvov Once again, please ensure you correctly copy the commit messages into the description _including_ the `#1234` link so that the 'audit log' on all issues is correctly updated with a reference to this deployment, and _also_ ensure that you add deployment issues to the current sprint's project board.

I have to remind you of this every single time you make a deployment. Please don't make me remind you again/take further steps to enforce this such as blocking deploys without review/etc.

PR #1326, opened 2021-04-13
Refactor code to re-use our generateTables helper

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/430

(**Edit by @0xdevalias:** Does this actually close this issue? Seems unrelated to me.. see questions in https://github.com/sparkletown/sparkle/pull/1326#pullrequestreview-635177059)

### DIFF
--- a/scripts/lib/helpers.ts
+++ b/scripts/lib/helpers.ts
@@ -4,8 +4,6 @@ import { relative, resolve } from "path";
 import admin from "firebase-admin";
 import formatDate from "date-fns/format/index.js";
 
-import { Table } from "../../src/types/Table";
-
 export interface CredentialFile {
   type?: string;
   project_id?: string;
@@ -194,47 +192,3 @@ export const findUserByEmail = (app: admin.app.App) => (
     .auth()
     .getUserByEmail(email)
     .catch(() => undefined);
-
-/**
- * Generate an array of Table configs that can be used with Jazz Bar/similar.
- *
- * @param num number of tables to create
- * @param capacity how many people can sit at each table
- * @param rows how many rows will the seats be displayed across for each table
- * @param columns how many columns will the seats be displayed across for each table
- * @param titlePrefix what should the tables be called (will have the table number appended to it)
- * @param appendTableNumber whether to append the table number to the title or not
- * @param startFrom what number should we start from when generating table numbers in the title
- */
-export const generateTables: (props: {
-  num: number;
-  capacity: number;
-  rows?: number;
-  columns?: number;
-  titlePrefix?: string;
-  appendTableNumber?: boolean;
-  startFrom?: number;
-}) => Table[] = ({
-  num,
-  capacity,
-  rows = 2,
-  columns = 3,
-  titlePrefix = "Table",
-  appendTableNumber = true,
-  startFrom = 1,
-}) =>
-  Array.from(Array(num)).map((_, idx) => {
-    const tableNumber = startFrom + idx;
-
-    const title = appendTableNumber
-      ? `${titlePrefix} ${tableNumber}`
-      : titlePrefix;
-
-    return {
-      title,
-      reference: title,
-      capacity,
-      rows,
-      columns,
-    };
-  });
--- a/src/components/molecules/TablesUserList/TablesUserList.tsx
+++ b/src/components/molecules/TablesUserList/TablesUserList.tsx
@@ -166,7 +166,7 @@ const TablesUserList: React.FunctionComponent<PropsType> = ({
         <>
           {tables.map((table: Table, i: number) => (
             <TableComponent
-              key={table.title}
+              key={table.reference}
               experienceName={venueName}
               users={recentVenueUsers}
               table={table}
--- a/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
+++ b/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
@@ -41,6 +41,39 @@ import { JAZZBAR_TABLES } from "./constants";
 
 import "./JazzTab.scss";
 
+export const generateTables: (props: {
+  num: number;
+  capacity: number;
+  rows?: number;
+  columns?: number;
+  titlePrefix?: string;
+  appendTableNumber?: boolean;
+  startFrom?: number;
+}) => {} = ({
+  num,
+  capacity,
+  rows = 2,
+  columns = 3,
+  titlePrefix = "Table",
+  appendTableNumber = true,
+  startFrom = 1,
+}) =>
+  Array.from(Array(num)).map((_, idx) => {
+    const tableNumber = startFrom + idx;
+
+    const title = appendTableNumber
+      ? `${titlePrefix} ${tableNumber}`
+      : titlePrefix;
+
+    return {
+      title,
+      reference: title,
+      capacity,
+      rows,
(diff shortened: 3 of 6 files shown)


scripts/lib/helpers.ts
```diff
@@ -194,47 +192,3 @@ export const findUserByEmail = (app: admin.app.App) => (
...
-      : titlePrefix;
-
-    return {
-      title,
-      reference: title,
-      capacity,
-      rows,
-      columns,
-    };
-  });
```
0xdevalias, line 240: [note] Presumably this was moved verbatim into `src/utils/table.ts`

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -166,7 +166,7 @@ const TablesUserList: React.FunctionComponent<PropsType> = ({
         <>
           {tables.map((table: Table, i: number) => (
             <TableComponent
-              key={table.title}
+              key={table.reference}
```
0xdevalias, line 169: What's the intent/reasoning for this change? Is it to fix anything specifically? Or just seemed like a better choice?

src/components/templates/Jazzbar/JazzTab/constants.ts
```diff
@@ -1,13 +1,5 @@
...
-  reference: `jazzbar-table-${index}`,
-  rows: 2,
-  columns: 3,
-});
+import { generateTables } from "utils/table";
 
-export const JAZZBAR_TABLES: Table[] = new Array(12)
-  .fill(0)
-  .map((_, index) => jazzBarTable(index + 1));
+export const JAZZBAR_TABLES: Table[] = generateTables({ num: 12, capacity: 6 });
```
0xdevalias, line 5: In the new code we have different logic used for the `reference` key. How is that used by the app, and does it matter? Will the new way work fine?

src/utils/table.ts
```diff
@@ -0,0 +1,45 @@
...
+      : titlePrefix;
+
+    return {
+      title,
+      reference: title,
+      capacity,
+      rows,
+      columns,
+    };
+  });
```
0xdevalias, line 45: [note] Presumably this was moved verbatim from `scripts/lib/helpers.ts`

0xdevalias (approved):
[CRO] Changes here look fine, though in your PR description you say that this closes https://github.com/sparkletown/internal-sparkle-issues/issues/430

Reading that, it doesn't seem to relate to the changes made here? At the very least not in a way that would fix the described bug. This only seems to get rid of some of the tech debt of duplicated helper functions used for table generation.

If this actually does close the issue as you suggest, can you please explain more about how/why that is the code; or alternatively, what the motivation/intent/etc of this PR is.

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -41,6 +41,39 @@ import { JAZZBAR_TABLES } from "./constants";
 
 import "./JazzTab.scss";
 
+export const generateTables: (props: {
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1326/commits/e6d4161730de145834e06753b1f0dc55605b74f2

src/utils/table.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param num number of tables to create
+ * @param capacity how many people can sit at each table
+ * @param rows how many rows will the seats be displayed across for each table
+ * @param columns how many columns will the seats be displayed across for each table
+ * @param titlePrefix what should the tables be called (will have the table number appended to it)
+ * @param appendTableNumber whether to append the table number to the title or not
+ * @param startFrom what number should we start from when generating table numbers in the title
+ */
+export const generateTables: (props: {
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1326/commits/e6d4161730de145834e06753b1f0dc55605b74f2

src/components/molecules/TablesUserList/TablesUserList.tsx
mike-lvov: This change exactly is because it seemed to be a better choice and titles can potentially be modified thus same

src/components/templates/Jazzbar/JazzTab/constants.ts
mike-lvov: That was the exact reason people got kicked out of tables. Reference is like an id.

image


You can find more in src/components/molecules/TablesUserList/TablesUserList.tsx

src/utils/table.ts
mike-lvov: Yes

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: Makes sense.

src/components/templates/Jazzbar/JazzTab/constants.ts
mike-lvov: So, tables generated by script had different IDs

src/utils/table.ts
0xdevalias: Cool

src/components/templates/Jazzbar/JazzTab/constants.ts
0xdevalias: Huh, interesting!

0xdevalias: Is this reference used for anything else? Eg. Anything related to twilio/video sessions/etc?

I feel like there possibly was a reason we didn’t use titles for it before (or perhaps we used to and then changed the code so that we no longer did?)

0xdevalias: @mike-lvov I’ve asked you this a few times in the past, but can you please make sure you include an ‘actual’ description in PRs that talks about what you did/why/etc. Linking to the related issue is insufficient.

It would have saved both of us time here, as it would have set the right context/described things up front.

mike-lvov: 1) Yeah, table.reference is used in getting videoToken 
image

2) I don't think using titles is/was a good idea (the same reason I described above)

0xdevalias: Dupes are particularly harmful with the table reference thing too, as the way they are used will end up with anyone using the same firebase account, with the same name, ending up in the same video session I believe.

0xdevalias: Agreed that name is a bad choice to use here. We probably want something like the firebase environment id + venueId + tableId or similar to ensure its unique/isolated/etc.

mike-lvov: VenueId is indeed used in creation of roomName it's venueId + table reference. We have an issue, which will be implemented during this sprint, which requires adding new functionality to tables. I will have to refactor the table component to be able to add new things and will implement it there

0xdevalias: Cool cool, sounds good.

PR #1330, opened 2021-04-15
Fixes room edition

mike-lvov:
Fixes room edition

PR #1338, opened 2021-04-19
Unify types/values passed to upsertRoom/updateRoom in admin components

yurii-lubynets:
connected to https://github.com/sparkletown/sparkle/pull/1330 (see review comments in https://github.com/sparkletown/sparkle/pull/1330#pullrequestreview-637305519)

### DIFF
--- a/src/pages/Admin/Venue/Details/VenueDetails.tsx
+++ b/src/pages/Admin/Venue/Details/VenueDetails.tsx
@@ -158,7 +158,7 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
       ...values,
     };
 
-    await updateRoom(roomValues, venueId!, user, index);
+    await updateRoom(roomValues, venueId, user, index);
     closeEditingModal();
   };
 
--- a/src/pages/Admin/Venue/Details/VenueDetails.tsx
+++ b/src/pages/Admin/Venue/Details/VenueDetails.tsx
@@ -153,12 +153,12 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
   if (!user) return null;
 
   const handleEditRoomSave = async (values: RoomData_v2, index: number) => {
-    const newData: RoomData_v2 = {
+    const roomValues: RoomData_v2 = {
       ...editingRoom,
       ...values,
     };
 
-    await updateRoom(newData, venueId!, user, index);
+    await updateRoom(roomValues, venueId!, user, index);
     closeEditingModal();
   };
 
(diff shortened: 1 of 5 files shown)


src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -153,12 +153,12 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
...
 
   const handleEditRoomSave = async (values: RoomData_v2, index: number) => {
-    const newData: RoomData_v2 = {
+    const roomValues: RoomData_v2 = {
       ...editingRoom,
       ...values,
     };
 
-    await updateRoom(newData, venueId!, user, index);
+    await updateRoom(roomValues, venueId!, user, index);
```
0xdevalias, line 161: We want to remove this TypeScript `!` from the codebase as it basically disables type checking.

```suggestion
    await updateRoom(roomValues, venueId, user, index);
```

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -101,7 +101,7 @@ interface RoomInnerForm {
   editingRoomIndex?: number;
 }
 
-export type FormValues = Partial<Yup.InferType<typeof validationSchema>>;
+export type FormValues = Yup.InferType<typeof validationSchema>;
```
0xdevalias, line 104: Why did we change the type from using `Partial` here? Is that correct? Will it break anything? Are the form values always guaranteed to be set in the `onSubmit`? Or does this just end up making our types 'lie' about the underlying data at 'compile time', which will then lead to hard to debug/identify runtime issues.

0xdevalias (commented):
Comments/Questions/etc

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
yurii-lubynets: All fields in the `RoomsForm` are required and there is no sense to make them partial.
And it fixed the type checking `onSubmit` with `RoomInput`. Guess now it's more secure

0xdevalias: So long as everything still works as expected if I only partially fill out the form, then sounds good.

0xdevalias (approved):
[CRO] Code review looks good, I haven't clicked through/tested this, so if we're sure it works/won't break things (and https://github.com/sparkletown/sparkle/pull/1338#discussion_r617090844 is satisfied), then LGTM!

PR #1334, opened 2021-04-16
Refactor validation for room name length

malaikaaiyar:
Simple change that lets you change the room name length limits from one place in `src/pages/Admin/Details/ValidationSchema.tsx`. Also sets max length to 50 as discussed in https://github.com/sparkletown/internal-sparkle-issues/issues/476.

closes https://github.com/sparkletown/internal-sparkle-issues/issues/476

(@0xdevalias, not sure if this is what you had in mind when he asked me to refactor it, let me know if it's not)

PR #1341, opened 2021-04-21
clean up + unify some duplicated aspects of our validations

malaikaaiyar:
changes from @0xdevalias's retroactive review of https://github.com/sparkletown/sparkle/pull/1334#pullrequestreview-638719745 (+ a few other small related things)


src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -151,16 +142,7 @@ export const roomCreateSchema = Yup.object().shape<RoomSchemaShape>({
...
-        .required("Venue name is required")
-        .min(
-          VENUE_NAME_MIN_CHAR_COUNT,
-          ({ min }) => `Name must be at least ${min} characters`
-        )
-        .max(
-          VENUE_NAME_MAX_CHAR_COUNT,
-          ({ max }) => `Name must be less than ${max} characters`
-        ),
+      then: roomTitleSchema,
```
0xdevalias, line 145: This used to say "Venue name is required", it seems that `roomTitleSchema` says "Room name is required"

Is it still valid now? Do both of these actually refer to the same concept? If so, is 'venue' or 'room' more correct here? If not, can we make the wording more generic and just use 'Name is required' or similar?

0xdevalias (changes requested):
There were 4 changes called out in https://github.com/sparkletown/sparkle/pull/1334#pullrequestreview-638719745 across 2 files. I think you've covered most (if not all) of the ones related to `src/pages/Admin/Details/ValidationSchema.ts`, but the comment related to `src/pages/Admin/Venue/DetailsValidationSchema.ts` (https://github.com/sparkletown/sparkle/pull/1334#discussion_r615739372) seems to be outstanding still. I would probably `export` the `roomTitleSchema` definition and reuse it there as well.

---

As an aside (which may be OOS of this PR to resolve), what's the difference between these two (very similarly named, to the point that my brain got confused by them for a while) files?

- `src/pages/Admin/Details/ValidationSchema.ts`
- `src/pages/Admin/Venue/DetailsValidationSchema.ts`

Are there any other duplications within them that we can 'break apart' into reusable chunks/simplify/reuse/reduce duplication of?

src/pages/Admin/Details/ValidationSchema.ts
malaikaaiyar: I was wondering what I should do about this. I'll change it to 'Name is required'

malaikaaiyar:
> As an aside (which may be OOS of this PR to resolve), what's the difference between these two (very similarly named, to the point that my brain got confused by them for a while) files?
> 
> * `src/pages/Admin/Details/ValidationSchema.ts`
> * `src/pages/Admin/Venue/DetailsValidationSchema.ts`

I don't really know, but they are used in different places - `DetailsValidationSchema` is only used in `DetailsForm.tsx`, while `ValidationSchema` is used in many different files.

From a quick skim the things that can be shared are:
- `urlIfNoFileValidation`
- `roomTitleSchema`
- `validationSchema_v2`  (in ValidationSchema) and `validationSchema` (in DetailsValidationSchema) seem pretty similar

Should I make a PR after this gets approved to break those things apart and reuse them or put them into this PR?

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Sounds good.

0xdevalias:
> Should I make a PR after this gets approved to break those things apart and reuse them or put them into this PR?

@malaikaaiyar Probably fine to put in this PR since it's about cleaning up validation stuff in general. Though make sure to update the title/description to reflect that it's not just cleaning up the room title validation anymore.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -73,43 +73,41 @@ const urlIfNoFileValidation = (fieldName: string) =>
 const mustBeMinimum = (fieldName: string, min: number) =>
   `${fieldName} must be at least ${min} characters`;
 
+const mustBeMaximum = (fieldName: string, max: number) =>
+  `${fieldName} must be at least ${max} characters`;
+
+export const roomTitleSchema = Yup.string()
+  .required("Room name is required")
```
0xdevalias, line 80:
```suggestion
  .required("Name is required")
```

0xdevalias: Looks like this wasn't done from https://github.com/sparkletown/sparkle/pull/1341#discussion_r618730811 ?

0xdevalias (changes requested):
Passing this back to you, as looks like the changes you said you would make weren't done yet.

Also, see https://github.com/sparkletown/sparkle/pull/1341#issuecomment-825266282

0xdevalias:
> I found that validationSchema-v2 and validationSchema are pretty different so I didn't share it across both files.

@malaikaaiyar Yup, that's ok. That's a piece of tech debt we need to clean up to unify the types (and validations) that diverged in admin v2

(@denisdimitrov @mike-lvov FYI: i'm sure you're already aware, but for explicitness)

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -185,10 +168,10 @@ export const validationSchema = Yup.object()
     showReactions: Yup.bool().notRequired(),
     auditoriumColumns: Yup.number()
       .notRequired()
-      .min(5, "Columns must be at least 5"),
+      .min(5, "There must be at least 5 columns"),
     auditoriumRows: Yup.number()
       .notRequired()
-      .min(5, "Rows must be at least 5"),
+      .min(5, "There must be at least 5 rows"),
```
0xdevalias, line 174: Is this enforced/handled somehow in the auditorium (`Audience` component) at all? And if so, is it using any `const`s extracted into `src/settings.ts`?

We want to extract the `5`'s here into `src/settings.ts`, and use the same `const`s used in `Audience`.

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -46,53 +48,34 @@ const createFileSchema = (name: string, required: boolean) =>
...
     );
 
-const urlIfNoFileValidation = (fieldName: string) =>
-  Yup.string().when(
-    fieldName,
-    (file: FileList | undefined, schema: Yup.MixedSchema<FileList>) =>
-      file && file.length > 0
-        ? schema.notRequired()
-        : schema.required("Required")
-  );
```
0xdevalias, line 56: [note] This is exactly the same as the version in `src/pages/Admin/Details/ValidationSchema.ts` that is now imported here.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -131,16 +129,9 @@ export const validationSchema_v2 = Yup.object()
   })
   .required();
 
-const roomTitleSchema = Yup.string()
-  .required("Room name is required")
-  .min(
-    VENUE_NAME_MIN_CHAR_COUNT,
-    ({ min }) => `Name must be at least ${min} characters`
-  );
-
```
0xdevalias, line 140: [note] Moved above + refactored slightly.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -70,46 +70,44 @@ const urlIfNoFileValidation = (fieldName: string) =>
         : schema.required("Required")
   );
 
-const mustBeMinimum = (fieldName: string, min: number) =>
+export const mustBeMinimum = (fieldName: string, min: number) =>
   `${fieldName} must be at least ${min} characters`;
 
+export const mustBeMaximum = (fieldName: string, max: number) =>
+  `${fieldName} must be at least ${max} characters`;
```
0xdevalias, line 77: This seems wrong:
```suggestion
export const mustBeMaximum = (fieldName: string, max: number) =>
  `${fieldName} must be less than ${max} characters`;
```

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -46,53 +48,34 @@ const createFileSchema = (name: string, required: boolean) =>
...
-      .required("Venue name is required")
-      .min(
-        VENUE_NAME_MIN_CHAR_COUNT,
-        ({ min }) => `Name must be at least ${min} characters`
-      )
-      .max(
-        VENUE_NAME_MAX_CHAR_COUNT,
-        ({ max }) => `Name must be less than ${max} characters`
-      )
-      .when(
```
0xdevalias, line 71: [note] This is exactly the same as the version in `src/pages/Admin/Details/ValidationSchema.ts` that is now imported here.

0xdevalias (changes requested):
A couple of little tweaks, then this looks good to go.

src/pages/Admin/Venue/DetailsValidationSchema.ts
malaikaaiyar: I don't think so. There isn't a variable in `settings.ts` that relates to this and in `Audience` the minimum values are hard coded as 25 and 19 (so we should change it to either 5x5 or 25x19). Should I implement this?

0xdevalias: @malaikaaiyar If I remember correctly, I think the values set in `Audience` might be the defaults, but not the minimum allowed. If that's the case, then this should just be extracted as new `const`s.

---

@mike-lvov @denisdimitrov Do you know what we should be setting here?

0xdevalias (dismissed):
@malaikaaiyar The changes in https://github.com/sparkletown/sparkle/pull/1341#discussion_r631402820 are still pending, needs to be extracted to `src/settings`. Needs input from @mike-lvov / @denisdimitrov

src/pages/Admin/Venue/DetailsValidationSchema.ts
denisdimitrov: >  If I remember correctly, I think the values set in Audience might be the defaults, but not the minimum allowed. If that's the case, then this should just be extracted as new consts.


That's correct. Those values are the default ones. 

> Do you know what we should be setting here?

Nope. We've never talked about the min/max values for this venue template. Maybe we can set it to 5x5 or just ask product team for some input.

0xdevalias: @sparkletown/product Do you have any thoughts for the minimum size the auditorium should be allowed to be set to? Currently `5`.

mike-lvov: @0xdevalias there are no tech minimums neither for rows nor for columns.

malaikaaiyar: There is now a settings variable called MINIMUM_COLUMNS and MAXIMUM_COLUMNS. Can I use this for both the rows and columns?

malaikaaiyar: fixed in 2f465aa7a6c887c40d3e054ac5059ab66d0e2a33

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -185,10 +168,24 @@ export const validationSchema = Yup.object()
...
-      .min(5, "Rows must be at least 5"),
+      .min(
+        MINIMUM_COLUMNS,
+        `The rows need to be between ${MINIMUM_COLUMNS} and ${MAXIMUM_COLUMNS}.`
+      )
+      .max(
+        MAXIMUM_COLUMNS,
+        `The rows need to be between ${MINIMUM_COLUMNS} and ${MAXIMUM_COLUMNS}.`
+      ),
   })
```
mike-lvov, line 189: should be rows, not columns

mike-lvov (changes requested):

src/pages/Admin/Venue/DetailsValidationSchema.ts
malaikaaiyar: fixed in 5d2a552

mike-lvov: @malaikaaiyar they are about the partymap, correct? Would be good to rename those 4 to `MINIMUM_PARTYMAP_COLUMNS_COUNT`

malaikaaiyar: @mike-lvov ah I didn't see that. I will rename those variables and make new ones for the auditorium validation

mike-lvov: @malaikaaiyar we have the variable for the auditorium validation 
image

malaikaaiyar: fixed in 1416c33c

mike-lvov (approved):
LGTM

PR #1342, opened 2021-04-22
fix React warning about import of Loading component

yurii-lubynets:
fix React warning about import of Loading component


0xdevalias (approved):
LGTM!

PR #1344, opened 2021-04-23
Configures PA environment

mike-lvov:
related to https://github.com/sparkletown/internal-sparkle-issues/issues/666

Code changes accroding to https://github.com/sparkletown/internal-sparkle-issues/issues/186#issuecomment-781318065


0xdevalias (approved):
Code changes here look good based on the diff, and from memory this should be all the places (4 files seems right I think)

If you're confident that's all of the places that need to be updated for this, LGTM!

PR #1348, opened 2021-04-28
Fix input value clearing on x-icon click

sunny-viktoryia:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/678

There are 2 variables for search query: 
1. `searchInputValue` which is synchronized with the text that is written in the input field.
2. `searchQuery` which is changed with a delay (debounced `searchInputValue`) to reduce the number of search requests.

### DIFF
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -33,6 +33,7 @@ const emptyEventsArray: VenueEvent[] = [];
 const DEBOUNCE_TIME = 200; // ms
 
 const NavSearchBar = () => {
+  const [searchInputValue, setSearchInputValue] = useState("");
   const [searchQuery, setSearchQuery] = useState("");
 
   const debouncedSearch = useMemo(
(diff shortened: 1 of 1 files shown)


yurii-lubynets (approved):

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -170,6 +173,7 @@ const NavSearchBar = () => {
       </div>
 
       <InputField
+        value={searchInputValue}
         inputClassName="NavSearchBar__search-input"
         onChange={onSearchInputChange}
         placeholder="Search for people, rooms, events..."
```
mike-lvov, line 179: I don't like the fact that we have 2 states(UI/functional) of the same thing. It doesn't seem right to me. searchQuery should derive & be dependent from searchInputValue. We have react-hook-form to handle the basic input state management. It has nice API, which includes `reset ` and `watch`. Then we can move setting search query into useEffect, dependent on that watch value `src/components/molecules/Chatbox/Chatbox.tsx`

mike-lvov (changes requested):

src/components/molecules/NavSearchBar/NavSearchBar.tsx
mike-lvov: I gues you don't need handelSubmit, maybe there is a useful `mode` that suits us that can be found in https://react-hook-form.com/

mike-lvov (approved):

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1348/commits/270c5e913961b67432b9e5cc8d2096c756b6203c

PR #1350, opened 2021-04-28
fix NavSearchBar search results to show under the NavBar

sunny-viktoryia:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/679


yurii-lubynets (approved):

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -86,92 +87,6 @@ $border-radius: 28px;
...
-
-  .profile-icon-container {
-    cursor: pointer;
-  }
-
-  .notification-card {
-    background: $red;
-    color: $white;
-    position: absolute;
-    bottom: 0;
```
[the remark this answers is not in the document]
sunny-viktoryia: sure, if you look carefully at the file, you'll notice that this is a copy-paste of the lines 1-89.
I checked all the rule sets separately before the deletion.

mike-lvov (approved):

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -7,6 +7,7 @@ $border-radius: 28px;
   display: flex;
   justify-content: space-between;
   align-items: center;
+  background-color: inherit;
```
0xdevalias, line 10: @sunny-viktoryia How is this intended to differ from the `background: $black` set on line 6? I believe both of these properties are going to try and set the background color, so one of them is presumably wrong/should be removed.

src/components/molecules/NavBar/playa.scss
```diff
@@ -37,7 +37,6 @@ $sand: #937c63;
     height: 100%;
     margin-left: auto;
     margin-right: auto;
-    background-color: transparent;
 
     .navbar-logo_container {
       display: flex;
```
0xdevalias, line 42: Potentially/probably OOS for this issue specifically, but we should really be cleaning up/removing all of these extra styles from this `playa.scss` file. Eg. `.navbar-container` should entirely be defined within `src/components/molecules/NavBar/NavBar.scss`, so it shouldn't really exist in `src/components/molecules/NavBar/playa.scss`, etc.

src/components/molecules/NavBar/playa.scss
```diff
@@ -60,12 +59,13 @@ $sand: #937c63;
     }
 
     .navbar-links {
-      padding: 12px 20px;
+      padding: 11px 20px;
```
0xdevalias, line 62: Why the `1px` change here? And how are we calculating the 'magic number' `11`/etc? If it relates to some other sizes/etc elsewhere, can we use variables and 'show the workings' here?

src/components/molecules/NavBar/playa.scss
```diff
@@ -60,12 +59,13 @@ $sand: #937c63;
...
       display: flex;
       flex-wrap: wrap;
       align-items: center;
       align-content: space-between;
       justify-content: flex-end;
+      background-color: inherit;
```
0xdevalias, line 68: Why do we need to change this here? Presumably the `.navbar-links` are rendered within the `.navbar-container`, which explicitly sets a background color. The default is 'transparent', so that it will show the same background as the element behind it. Is that not sufficient here?

- https://developer.mozilla.org/en-US/docs/Web/CSS/background-color#formal_definition

src/components/molecules/NavBar/NavBar.scss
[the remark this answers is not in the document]
0xdevalias: @sunny-viktoryia Makes sense. Though as @mike-lvov said, it's good practice to proactively 'call out'/explain things like this up front to make the reviewers life easier. That way we save a 'round of back and forth', and speed up the review process.

src/scss/constants.scss
```diff
@@ -98,6 +98,7 @@ $z-layers: (
   navbar-schedule-backdrop: $z-layer-live-schedule,
   navbar-schedule: $z-layer-live-schedule,
   navbar: $z-layer-navbar,
+  navbar__dropdown: -1,
```
0xdevalias, line 101: Why is this set to `-1`? I'm not sure that's what we want to be doing here? I believe what we actually want to do is something similar to what we did with `$z-layer-live-schedule`, and define a new 'layer' for these 'navbar drawers' to sit on. My gut feel/guess is we can probably reuse `$z-layer-live-schedule` here since it's basically just another 'drawer' (though double check that + click around to ensure it still looks right/operates well with everything else). If we do reuse it, it's probably good to rename `$z-layer-live-schedule` to something more generic; perhaps `$z-layer-navbar-drawer`

0xdevalias:
```suggestion
  navbar__drawer: $z-layer-navbar-drawer,
```

src/components/molecules/NavSearchBar/NavSearchBar.scss
```diff
@@ -4,6 +4,7 @@
 .NavSearchBar {
   position: relative;
   width: $search-dropdown-width;
+  background-color: inherit;
```
0xdevalias, line 7: As asked in other comments in this review, why do we need this `inherit` here? Is the default 'not set' state of 'transparent' (thus showing the element behind it) not sufficient?

0xdevalias (changes requested):
[CRO] Questions/comments/changes

---

@mike-lvov Once these are satisfactorily addressed, if it's outside of my timezone hours, feel free to dismiss this review to get this merged.

src/components/molecules/NavBar/NavBar.scss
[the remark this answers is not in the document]
sunny-viktoryia: sure, will do next time. :upside_down_face: Thanks!

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: Let me explain the logic of the fix.

Here is the html structure.
```html
<div class="navbar">                        <--- main navbar container z-index: 30
  <div class="nav-container">           <--- background: black;
    <div class="navbar-links">
      <div class="NavSearchBar">
        <div class="NavSearchBar__nav-dropdown" />
        <div class="input-field" />
      </div>
    </div>
  </div>
</div>

...
<div class="schedule-dropdown-backdrop" />
```

The fix consists of 2 parts: 
1. `background-color: inherit`
initially divs have transparent color by default, that's why even being behind those divs, `NavSearchBar__nav-dropdown` was visible anyway. (see the screenshot of its appearing from the top)

image

specifying `inherit` background-color we tell `navbar-links` to take the background-color of its parent (e.g. black), so it's not transparent anymore and elements can be hidden behind that.

image

2. `z-index: -1`
`navbar` creates a stacking context because it has `z-index`. 

* https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
image


it means that if we apply `z-index` to its children, the meaning of that "child" z-index will be within "parent" z-index only.
So, if we apply `z-index: 20` to `NavSearchBar__nav-dropdown` which is lower value than `navbar` has (30)
it won't hide `NavSearchBar__nav-dropdown` behind `navbar` because its z-index works within parent stacking context.

That's why I use `z-index: -1` to have it behind all the elements within `navbar`.

That also explains why we can't use same z-index as `schedule-dropdown-backdrop` has.
You can notice that `navbar` and `schedule-dropdown-backdrop` don't have child-parent relationships. That's why their z-indexes work as expected.

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: when the color to `.navbar-links` became not transparent, it turned out that it hides border-bottom of the `.navbar` - see the yellow line, so I just decreased it a bit.
image

src/components/molecules/NavBar/NavBar.scss
sunny-viktoryia: ooops, was so happy that my fix worked, that forgot to check where the property is not needed. :see_no_evil:  
fixed: https://github.com/sparkletown/sparkle/pull/1350/commits/b6e481c7aad3433f43e567cdd0244c43fccfc78f

src/components/molecules/NavSearchBar/NavSearchBar.scss
sunny-viktoryia: fixed: https://github.com/sparkletown/sparkle/pull/1350/commits/b6e481c7aad3433f43e567cdd0244c43fccfc78f

src/scss/constants.scss
sunny-viktoryia: unfortunately, we can't reuse `$z-layer-live-schedule` (= 20).
Please, check the explanation in my comment : https://github.com/sparkletown/sparkle/pull/1350#discussion_r622827805

but yeah, I can rename it.

src/components/molecules/NavBar/playa.scss
0xdevalias: > that's why even being behind those divs, NavSearchBar__nav-dropdown was visible anyway

@sunny-viktoryia In that screenshot it doesn't look like it's behind the NavBar (eg. `z-index`) though?

image

---

> So, if we apply z-index: 20 to NavSearchBar__nav-dropdown which is lower value than navbar has (30) it won't hide NavSearchBar__nav-dropdown behind navbar because its z-index works within parent stacking context.

@sunny-viktoryia Is the 'drawer' not positioned `absolute`/similar that removes it from the parent stacking context?

---

> That also explains why we can't use same z-index as schedule-dropdown-backdrop has.
> You can notice that navbar and schedule-dropdown-backdrop don't have child-parent relationships. > That's why their z-indexes work as expected.

@sunny-viktoryia Makes sense. What would it take to remove the child-parent relationship here so that it can work in the same way? As that seems like it would likely be the better overall fix here?

---

@mike-lvov Thoughts?

src/components/molecules/NavBar/playa.scss
0xdevalias: Makes sense. Is it possible for this element to be 'properly container' within the NavBar, rather than being able to 'break out of it' like this?

src/components/molecules/NavBar/NavBar.scss
0xdevalias: Haha, I know that feel.

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: > In that screenshot it doesn't look like it's behind the NavBar (eg. z-index) though?

you're right. It's not behind NavBar (as it's in its stacking context), but it's behind transparent `.navbar-links`. And the 2nd 
screenshot was made when background-color of `.navbar-links` was set to `inherit` (not transparent, but black) - that was the main idea of the fix.



> What would it take to remove the child-parent relationship here so that it can work in the same way? As that seems like it would likely be the better overall fix here?

that what I tried to clarify in the #design slack channel yesterday. 
As far as I see it, we'll have to change the structure of the navbar components. 
It seems that we'll have to leave input-field as the child of the NavBar component and move search-results dropdown out of it.

... looking into the code to understand how the schedule opens ... 

NavBar component (`src/components/molecules/NavBar/NavBar.tsx`) is pretty big, and it has not only `.navbar` but `SchedulePageModal` there as well. So we can move schedule-results there as a sibling of `.navbar` and `shedule`. and using callback from NavSearchBar (on search query change) we can pass props to 'schedule-results'.

(note) 1 possible thing we need to think about is about positioning. When input and search-results are at the same component it's not difficult to place them vertically one under another , but as there may (or not) be radio (and maybe other components, which I don't know about) the positioning can differ.

src/scss/constants.scss
sunny-viktoryia: renamed: https://github.com/sparkletown/sparkle/pull/1350/commits/ea46715895e680fd894657bc141cafd1ce17202b

src/components/molecules/NavBar/playa.scss
mike-lvov: @0xdevalias @sunny-viktoryia 
I think the perfect way to do it would be to get search options out of the navbar stacking context and show them behind the navbar itself, but since the fix works as expected & doesn't have any drawbacks(or I just can't think of any?), let's leave this fix as is. I don't mind having it as OOS for every navbar element.

Let's focus on delivering OHBM features

0xdevalias: @mike-lvov @sunny-viktoryia  I'm ok with that, but can we we please add `@debt` comments to the places where this 'hack fix' was added that call this out so that we have context for future refactors to fix it properly?

0xdevalias (approved):
Code changes here look fine, but as per https://github.com/sparkletown/sparkle/pull/1350#discussion_r623539542 can we add some `@debt` comments to call things out for when we refactor/fix this for the NavBar drawers in future.

Once that is done, LGTM!

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: added the comments: https://github.com/sparkletown/sparkle/pull/1350/commits/d2e5fe9ef30c6cf6333f10fe908a91edb376ffc5

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: yeah, agree, there is a `@debt` comment in the file about that.
https://github.com/sparkletown/sparkle/blob/2ab1f73ef4f7eab79f07c2788a32525761310736/src/components/molecules/NavBar/playa.scss#L18

src/components/molecules/NavBar/playa.scss
sunny-viktoryia: I can't see what is the best way to do that right now.

In fact, it might be worth removing that padding fully. Not sure what was the original purpose of it.

with the padding:
image

without the padding: 
image

removing padding, profile icon comes closer to the right side. 
And looking at the design http://sapiens.chat/sparkle/styleguide/map.html it seem to be closer to the desired result.

@mathieud do you think, we need to remove the padding?

sunny-viktoryia: agreed with @mike-lvov to remove paddings. 

it broke the view in test env, for some reason.
image

https://github.com/sparkletown/sparkle/pull/1350/commits/d5c1e7f70916dcf9d54798b9f636a1523eb8785a
https://github.com/sparkletown/sparkle/pull/1350/commits/ff3ef23aa0fda2796f98e07ea004a9170eb80ef6

0xdevalias: @sunny-viktoryia FYI, @mcflyok is Mathieu

0xdevalias: > it broke the view in test env, for some reason.

@sunny-viktoryia Is this still an outstanding issue? Or is it resolved now?

src/components/molecules/NavBar/playa.scss
0xdevalias: Sounds good

PR #1336, opened 2021-04-19
Increases the char limit for venue name

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/476

Increases char limit for a venue name from 25 to 50

PR #1351, opened 2021-04-28
add create venue modal in admin v2

denisdimitrov:
Adds an additional modal as a step for the creation of a new venue. The newly created venue is intentionally created with no information (only venue name). The rest of the information can be edited from the "Manage venue" button.

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -170,16 +170,16 @@ const createVenueData_v2 = (data, context) => ({
   name: data.name,
   config: {
     landingPageConfig: {
-      coverImageUrl: data.bannerImageUrl,
-      subtitle: data.subtitle,
-      description: data.description,
+      coverImageUrl: data.bannerImageUrl || "",
+      subtitle: data.subtitle || "",
+      description: data.description || "",
     },
   },
   theme: {
     primaryColor: data.primaryColor || DEFAULT_PRIMARY_COLOR,
   },
   host: {
-    icon: data.logoImageUrl,
+    icon: data.logoImageUrl || "",
   },
   owners: [context.auth.token.user_id],
   showGrid: data.showGrid || false,
--- a/src/components/molecules/CreateVenueModal/CreateVenueModal.scss
+++ b/src/components/molecules/CreateVenueModal/CreateVenueModal.scss
@@ -1,0 +1,81 @@
+@import "scss/constants.scss";
+
+.modal-backdrop {
+  opacity: 0.9 !important;
+  background-color: black !important;
+}
+
+.modal-content {
+  background-color: $saturated-black !important;
+}
+
+.create-venue-modal {
+  background-color: $saturated-black;
+  padding: 20px;
+
+  &__content {
+    padding: 10px 20px;
+  }
+
+  &__title {
+    text-align: center;
+    padding-bottom: 1rem;
+    font-size: 1.5rem;
+    font-weight: bold;
+  }
+
+  &__url-info {
+    padding: 1rem 0;
+    font-size: 1rem;
+    overflow-wrap: break-word;
+    word-wrap: break-word;
+    word-break: break-word;
+  }
+
+  &__url {
+    padding: 1rem 0;
+    font-size: 1rem;
+    font-weight: bold;
+    overflow-wrap: break-word;
+    word-wrap: break-word;
+    word-break: break-word;
+  }
+
+  &__buttons {
+    display: flex;
+    flex-direction: row;
+
+    &--back {
+      display: flex;
+      flex: 1;
+      justify-content: center;
+      background-color: lighten($color: $saturated-black, $amount: 10);
+      color: $white;
+      padding: 16px 18px;
+      border-radius: 22px;
+      border-width: 0px;
+      outline: none !important;
+
+      &:hover {
+        background-color: lighten($color: $saturated-black, $amount: 20);
+      }
+    }
+
+    &--next {
+      display: flex;
+      flex: 4;
+      justify-content: center;
+      background-color: $primary;
+      color: $white;
+      margin-left: 10px;
+      padding: 16px 18px;
+      border-radius: 22px;
+      border-width: 0px;
+      outline: none !important;
+
+      &:hover {
+        background-color: lighten($color: $primary, $amount: 10);
+      }
+    }
+  }
+}
--- a/src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
+++ b/src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
@@ -1,0 +1,123 @@
+import React, { useCallback, useState } from "react";
+import { Form, Modal } from "react-bootstrap";
+import { useForm } from "react-hook-form";
+
+import { createUrlSafeName, createVenue_v2 } from "api/admin";
+
+import { useUser } from "hooks/useUser";
+import { useVenueId } from "hooks/useVenueId";
+
+import { venueLandingUrl } from "utils/url";
+
+import { VenueTemplate } from "types/venues";
+
+import { newVenueSchema } from "pages/Admin/Details/ValidationSchema";
+
+import "./CreateVenueModal.scss";
+
+interface CreateVenueModalProps {
+  isVisible: boolean;
+  onHide?: () => void;
+}
+
+export interface FormValues {
+  name: string;
+}
+
+export const CreateVenueModal: React.FC<CreateVenueModalProps> = ({
+  isVisible,
+  onHide,
+}) => {
+  const [isLoading, setLoading] = useState<boolean>(false);
+  const venueId = useVenueId();
+  const { user } = useUser();
+
+  const hide = useCallback(() => {
+    onHide && onHide();
+  }, [onHide]);
+
+  const onSubmit = useCallback(
+    async (vals: FormValues) => {
+      if (!user || isLoading) return;
+      setLoading(true);
+
+      try {
+        const venue = {
+          name: vals.name,
+          template: VenueTemplate.partymap,
+        };
+        await createVenue_v2(venue, user);
+        hide();
+      } catch (e) {
+        console.error(e);
+      } finally {
+        setLoading(false);
+      }
+    },
+    [hide, isLoading, user]
+  );
+
+  const { watch, register, errors, handleSubmit } = useForm<FormValues>({
+    mode: "onSubmit",
+    reValidateMode: "onSubmit",
+    validationSchema: newVenueSchema,
+    validationContext: {
+      editing: !!venueId,
+    },
+  });
+
+  const values = watch();
+
+  const urlSafeName = values.name
+    ? `${window.location.host}${venueLandingUrl(
+        createUrlSafeName(values.name)
+      )}`
+    : undefined;
+
+  return (
+    <Modal
+      show={isVisible}
+      onHide={hide}
+      dialogClassName="modal-dialog modal-dialog-centered"
+    >
+      <Modal.Body className="create-venue-modal">
+        <Form
+          onSubmit={handleSubmit(onSubmit)}
+          className="create-venue-modal__content"
+        >
+          <div className="create-venue-modal__title">Name your new space</div>
+          <Form.Control
+            name="name"
+            placeholder="Enter the name of your new space"
+            ref={register}
+            custom
+            disabled={isLoading}
+          />
+          {errors.name && (
+            <span className="input-error">{errors.name.message}</span>
+          )}
+          <div className="create-venue-modal__url-info">
+            The url of your party will be:{" "}
+            <span className="create-venue-modal__url">{urlSafeName}</span>
+          </div>
+          <div className="create-venue-modal__buttons">
+            <button
+              className="create-venue-modal__buttons--back"
+              onClick={hide}
+              disabled={isLoading}
+            >
+              Cancel
+            </button>
+            <button
+              className="create-venue-modal__buttons--next"
+              type="submit"
+              disabled={isLoading}
+            >
+              Next
+            </button>
+          </div>
+        </Form>
+      </Modal.Body>
+    </Modal>
+  );
+};
--- a/src/components/organisms/AdminVenues/AdminVenues.tsx
+++ b/src/components/organisms/AdminVenues/AdminVenues.tsx
@@ -7,9 +7,9 @@ import { AnyVenue, isPartyMapVenue } from "types/venues";
 import { WithId } from "utils/id";
 
 import { AdminVenueCard } from "components/molecules/AdminVenueCard";
+import { CreateVenueModal } from "components/molecules/CreateVenueModal";
 
 import "./AdminVenues.scss";
-import { CreateVenueModal } from "components/molecules/CreateVenueModal";
 
 export interface AdminVenuesProps {
   venues: WithId<AnyVenue>[];
--- a/src/components/organisms/AdminVenues/AdminVenues.tsx
+++ b/src/components/organisms/AdminVenues/AdminVenues.tsx
@@ -1,6 +1,5 @@
-import React, { useMemo } from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import { Button } from "react-bootstrap";
-import { Link } from "react-router-dom";
 import classNames from "classnames";
 
 import { AnyVenue, isPartyMapVenue } from "types/venues";
@@ -10,12 +9,23 @@ import { WithId } from "utils/id";
 import { AdminVenueCard } from "components/molecules/AdminVenueCard";
 
 import "./AdminVenues.scss";
+import { CreateVenueModal } from "components/molecules/CreateVenueModal";
 
 export interface AdminVenuesProps {
   venues: WithId<AnyVenue>[];
 }
 
 export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
+  const [showCreateModal, setShowCreateModal] = useState<boolean>(false);
+
+  const showModal = useCallback(() => {
+    setShowCreateModal(true);
+  }, []);
+
+  const hideModal = useCallback(() => {
+    setShowCreateModal(false);
+  }, []);
+
   const renderedPartyVenues = useMemo(
     () =>
       venues
@@ -30,9 +40,7 @@ export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
     <div className="admin-venue">
       <div className="admin-venue__header">
         <div className="admin-venue__title">Admin Dashboard</div>
-        <Button as={Link} to="/admin_v2/venue/creation">
-          Create a new space
-        </Button>
+        <Button onClick={showModal}>Create a new space</Button>
       </div>
       <div
         className={classNames("admin-venue__cards", {
@@ -49,6 +57,7 @@ export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
         )}
         {hasVenues && renderedPartyVenues}
       </div>
+      <CreateVenueModal isVisible={showCreateModal} onHide={hideModal} />
     </div>
   );
 };
--- a/src/pages/Admin/Details/ValidationSchema.ts
+++ b/src/pages/Admin/Details/ValidationSchema.ts
@@ -32,6 +32,10 @@ export interface SchemaShape {
   logoImageUrl: string;
 }
 
+export interface NewVenueSchema {
+  name: string;
+}
+
 export interface RoomSchemaShape {
   title: string;
   venueName?: string;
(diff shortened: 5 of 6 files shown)


functions/venue.js
```diff
@@ -170,16 +170,16 @@ const createVenueData_v2 = (data, context) => ({
...
+      subtitle: data.subtitle || "",
+      description: data.description || "",
     },
   },
   theme: {
     primaryColor: data.primaryColor || DEFAULT_PRIMARY_COLOR,
   },
   host: {
-    icon: data.logoImageUrl,
+    icon: data.logoImageUrl || "",
```
0xdevalias, line 182: Not sure that setting these to a default empty string is an ideal choice. Wouldn't they be better/more explicitly/obviously `undefined`/checkable as such by leaving them as `undefined`.

Then at the point where we are actually inserting them into firestore, we can either replace those keys with the 'delete this key' firestore command, or enable the 'just ignore `undefined` and don't treat it as an error' option on this query itself (not globally).

@mike-lvov has used some of these patterns before, so if you need more specifics, may be able to ask him about it.

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          template: VenueTemplate.partymap,
+        };
+        await createVenue_v2(venue, user);
+        hide();
+      } catch (e) {
+        console.error(e);
+      } finally {
+        setLoading(false);
+      }
+    },
```
0xdevalias, line 56: Instead of `try`/`catch`/`await` this would probably be simpler just using the chained Promise methods.

```typescript
createVenue_v2(venue, user)
  .then(() => {
    hide()
  })
  .catch((e) => {
    // TODO: don't just console.error, Bugsnag.notify and/or display this to the user in some way
  })
  .finally(() => {
    setLoading(false)
  })
```

This aligns with how we tend to do things elsewhere in components/etc. `async`/`await` becomes far more useful/boilerplate reducing in bigger/more complex situations, but it becomes overkill/more verbose in simple situations like this.

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+    reValidateMode: "onSubmit",
+    validationSchema: newVenueSchema,
+    validationContext: {
+      editing: !!venueId,
+    },
+  });
+
+  const values = watch();
+
+  const urlSafeName = values.name
```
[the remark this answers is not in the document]
0xdevalias: Presumably we should extract this into a helper function?

Also, we can construct the URL in a better/safer way than string concatenation like this. I believe we already do similar in `utils/url` using the `new Url` and similar.

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -16,6 +16,16 @@ export interface AdminVenuesProps {
...
 export const AdminVenues: React.FC<AdminVenuesProps> = ({ venues }) => {
+  const [showCreateModal, setShowCreateModal] = useState<boolean>(false);
+
+  const showModal = useCallback(() => {
+    setShowCreateModal(true);
+  }, []);
+
+  const hideModal = useCallback(() => {
+    setShowCreateModal(false);
+  }, []);
```
0xdevalias, line 27: This is a pattern we tend to use all over the codebase. We could extract it into a `useModalControls` or similar custom hook, and then just use that instead of having to set up this boilerplate all the time.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -131,6 +135,39 @@ export const validationSchema_v2 = Yup.object()
...
+    name: Yup.string()
+      .required("Name is required!")
+      .min(
+        VENUE_NAME_MIN_CHAR_COUNT,
+        ({ min }) => `Name must be at least ${min} characters`
+      )
+      .max(
+        VENUE_NAME_MAX_CHAR_COUNT,
+        ({ max }) => `Name must be less than ${max} characters`
+      )
```
0xdevalias, line 149: I believe we have extracted this out of other validation schemas into a reusable 'schema chunk', which should also be reused here.

0xdevalias: And in general, we should be reusing validations as much as possible (whether in their entirety, or smaller chunks of them), rather than duplicating them

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
```diff
@@ -0,0 +1,81 @@
+@import "scss/constants.scss";
+
+.modal-backdrop {
+  opacity: 0.9 !important;
+  background-color: black !important;
+}
+
+.modal-content {
+  background-color: $saturated-black !important;
+}
```
0xdevalias, line 10: Why are we using `!important` here? That's not something we want to be doing.

0xdevalias: (Also the other places within this file where it's being used currently)

denisdimitrov: The bootstrap modal is terrible when it comes to custom styles, and this seems to be the only way to override them. There are 2 props `dialogClassName` and `backdropClassName` which are okey, but they both have parent `classNames` which are the ones that I override with `!important`

I will take a look again because I'm not a fan of `!important` hopefully there is something else as alternative.

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: Noted.

functions/venue.js
denisdimitrov: I agree with you, I didn't know about the `ignoreundefinedproperies` setting. I'll use it, instead of flooding the DB with empty strings.

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
denisdimitrov: Noted.

functions/venue.js
denisdimitrov: @mike-lvov Do you have any experience with this? I tried adding `ignoreUndefinedProperies: true` in the config but I'm still getting rejected by the firebase functions, saying that undefined can not be used as a value.

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
0xdevalias: Theres many other ways to get our styles to override. Checkout:

- https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

functions/venue.js
0xdevalias: > I didn't know about the `ignoreundefinedproperies` setting

@denisdimitrov Yeah, I think we only found it relatively recently, possibly while you were on leave.

---

> but I'm still getting rejected by the firebase functions, saying that undefined can not be used as a value.

@denisdimitrov 😮😢 I'm not sure if we have actually used it yet. I think we ended up going with the delete pattern where I last called this out for a different use case.

Can you show code/errors/etc to help debug more?

denisdimitrov: Nevermind, I'm an idiot. It's working now.

0xdevalias: Haha, what was the issue?

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
denisdimitrov: @0xdevalias I got rid of the second `className` and used the prop `backdropClassName` as an alternative for the first one. But it still needs `!important`. For some strange reason the default styles override the ones provided by the prop class name.

denisdimitrov:
@0xdevalias Everything has been updated and refactored as suggested by your review comments. Feel free to take another look when you are available.

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
denisdimitrov: Found another solution which seems to be working fine. It's `modal-content.in` and then the styles you want. No need for `!important` 🥳

0xdevalias: Nice!

functions/venue.js
denisdimitrov: I was changing the config of `admin.initializeApp` instead of `admin.firestore().settings` That's why it wasn't working.

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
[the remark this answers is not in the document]
denisdimitrov: Looks like nothing of the already existing code works for this case. I just created a new function, you can take a look in the last commit.

functions/venue.js
0xdevalias: Cool, makes sense. A pitfall I will hopefully avoid myself in future now that I know! :p

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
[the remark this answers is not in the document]
0xdevalias: https://github.com/sparkletown/sparkle/pull/1351/commits/5b605b866c48d16a4be9f814684e8b2b6992493f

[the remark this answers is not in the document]
0xdevalias: I didn't expect one of the functions there would have worked, but I was talking about the method we use to construct a URL, not using string concatenation, but using `new URL`:

- https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

Looking at your function, it will probably end up looking something like this:

```typescript
export const generateVenueLandingUrl = (value?: string): string | undefined => {
  if (!value) return undefined

  const path = venueLandingUrl(createUrlSafeName(value))

  return new URL(path, window.location.host).toString()
```

It would be good to have a more contextually relevant name than `value` here too if possible. Is it intended to be the `venueId` or `venueName` or similar? If so, can we call it that?

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -40,6 +44,37 @@ export interface RoomSchemaShape {
...
+const venueNameLengthValidation = Yup.string()
+  .required("Venue name is required!")
+  .min(
+    VENUE_NAME_MIN_CHAR_COUNT,
+    ({ min }) => `Name must be at least ${min} characters`
+  )
+  .max(
+    VENUE_NAME_MAX_CHAR_COUNT,
+    ({ max }) => `Name must be less than ${max} characters`
+  );
```
0xdevalias, line 56: I'm almost certain we already have something like this.. from memory it was from one of @sunny-viktoryia / @malaikaaiyar 's PRs? Possibly still outstanding? Would be good to align with that.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -40,6 +44,37 @@ export interface RoomSchemaShape {
...
+    async (val: string) =>
+      !val ||
+      !(
+        await firebase
+          .firestore()
+          .collection("venues")
+          .doc(createUrlSafeName(val))
+          .get()
+      ).exists
+  );
```
0xdevalias, line 76: How often is this validation going to get executed? I'm guessing not just when the form is submitted? So potentially on every keystroke/similar? This might be a bit heavyweight to include here, and would likely be better done as a backend-triggered error anyway, since we're going to have to check it there regardless.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -75,41 +110,31 @@ const mustBeMinimum = (fieldName: string, min: number) =>
...
+                async (val: string) =>
+                  !val ||
+                  !(
+                    await firebase
+                      .firestore()
+                      .collection("venues")
+                      .doc(createUrlSafeName(val))
+                      .get()
+                  ).exists
+              )
```
0xdevalias, line 135: This seems to be a duplicate with what you extracted in `venueNameValidation` above?

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -131,6 +156,12 @@ export const validationSchema_v2 = Yup.object()
   })
   .required();
 
+export const newVenueSchema = Yup.object()
+  .shape<NewVenueSchema>({
+    name: venueNameValidation,
+  })
+  .required();
```
0xdevalias, line 163: Probably worth also including the 'only this data and nothing else is allowed' property on this schema as well.

0xdevalias: `noUnknown` (like what is being done in `src/types/sounds.ts`):

- https://github.com/jquense/yup#objectnounknownonlyknownkeys-boolean--true-message-string--function-schema

src/pages/Admin/Venue/Details/VenueDetails.tsx
```diff
@@ -50,13 +50,13 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venue }) => {
...
   } = venue;
-  const {
-    subtitle,
-    description,
-    coverImageUrl,
-  } = venue.config.landingPageConfig;
-  const { icon } = venue.host;
+  const subtitle = config?.landingPageConfig?.subtitle;
+  const description = config?.landingPageConfig?.description;
+  const coverImageUrl = config?.landingPageConfig?.coverImageUrl;
```
0xdevalias, line 57: Why was this split into 3 seperate lines? Seemingly no need to extract `config` either?

```typescript
const {
    subtitle,
    description,
    coverImageUrl,
  } = venue.config?.landingPageConfig ?? {};
```

0xdevalias (changes requested):

src/pages/Admin/Details/ValidationSchema.ts
sunny-viktoryia: hmm can't recall anything I've been working on related to that. :thinking: 

But there were @mike-lvov's hot fixes recently where he increased the length of the name as far as I remember. 

Did you mean that?
https://github.com/sparkletown/sparkle/pull/1336

0xdevalias: I think I meant: https://github.com/sparkletown/sparkle/pull/1341

src/pages/Admin/Venue/Details/VenueDetails.tsx
denisdimitrov: Typescript doesn't allow object destructure from optional fields

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: It depends on the UI. `useForm` accepts `reValidateMode` parameter which can be `onChange`, `onSubmit` and `onBlur`. I'm currently using `onSubmit` but if someone changes it to `onChange` it will be heavy indeed.

denisdimitrov: I agree that this should be removed from here, but every single of the admin creation tools is using this and removing the validation means that I need to go and handle every single 403 status code. Maybe I can create an issue for this and we can fix it in another PR?

src/components/organisms/AdminVenues/AdminVenues.tsx
```diff
@@ -1,13 +1,15 @@
 import React, { useMemo } from "react";
 import { Button } from "react-bootstrap";
-import { Link } from "react-router-dom";
 import classNames from "classnames";
 
-import { AnyVenue, isPartyMapVenue } from "types/venues";
```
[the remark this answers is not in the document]
denisdimitrov: I must have moved it bellow `hooks` while rearranging the imports. I don't really know the correct order and I think we should have a linter for this job.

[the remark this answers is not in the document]
denisdimitrov: I'm going to move it back to the top.

src/pages/Admin/Venue/Details/VenueDetails.tsx
0xdevalias: The code I posted above will absolutely work. The potentially undefined will get to be undefined, then the ?? {} defaults it to returning an object, which is then destructured. We already use this pattern across the codebase.

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Ya, backlog it for future cleanup

0xdevalias: Also maybe add an ‘@debt’ comment here calling out the onBlur/etc thing you mentioned above

src/components/organisms/AdminVenues/AdminVenues.tsx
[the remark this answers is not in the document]
0xdevalias: > I don't really know the correct order and I think we should have a linter for this job.

It’s in the works. The PR that needs review for upgrading react-scripts is a blocker on upgrading our linting. Once that’s landed I can add the rule for the imports.

src/api/admin.ts
```diff
@@ -153,6 +153,11 @@ const randomPrefix = () => Math.random().toString();
 export const createUrlSafeName = (name: string) =>
   name.replace(/\W/g, "").toLowerCase();
 
+export const generateVenueLandingUrl = (value?: string) =>
+  value
+    ? `${window.location.host}${venueLandingUrl(createUrlSafeName(value))}`
+    : undefined;
```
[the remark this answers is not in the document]
denisdimitrov: My bad, I accidentally missed it

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
```diff
@@ -0,0 +1,80 @@
+@import "scss/constants.scss";
+
+.modal-backdrop.in {
+  background-color: $black;
+}
+
+.modal-content.in {
+  background-color: $saturated-black;
+}
```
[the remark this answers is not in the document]
denisdimitrov: Sadly, it wasn't described how this works, only that "it works". Your guess is as good as mine.

src/pages/Admin/Venue/Details/VenueDetails.tsx
denisdimitrov: Yeah, I didn't notice `?? {}` at the end. I tried without it and it doesn't work. But with this syntax seems to be accepted.

src/pages/Admin/Details/ValidationSchema.ts
denisdimitrov: Nice catch.

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+        .then(() => {
+          hide();
+        })
+        .catch((e) => {
+          Bugsnag.notify(e, (event) => {
+            event.addMetadata("CreateVenueModal::createVenue_v2", {
+              venueName: vals.name,
+            });
+          });
+        })
```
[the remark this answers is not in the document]
denisdimitrov: State + generic error message should be enough for now. This is very bad UX and we should think of something better.

src/pages/Admin/Venue/Details/VenueDetails.tsx
0xdevalias: Without it it’s trying to destructure undefined; with it it will try and destructure an empty object (which makes all the keys undefined)

src/components/molecules/CreateVenueModal/CreateVenueModal.scss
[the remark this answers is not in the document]
0xdevalias: Checkout the link I shared in #tech recently about CSS specificity (called something like specifishitty). It’s a useful reference sheet that explains how it works/how to easily increase the specificity of things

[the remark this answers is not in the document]
0xdevalias: Can we add a comment to these explaining why they are as they are (eg. To increase the specificity to override the other styles). Otherwise I expect we’ll accidentally refactor the code in future and break this, as it deviates from our standard patterns

src/components/molecules/CreateVenueModal/CreateVenueModal.tsx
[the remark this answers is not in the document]
0xdevalias: I'm guessing that means you implemented state to show an error message, etc?

Not sure if you're aware of these `react-use` (https://github.com/streamich/react-use) async helpers, we don't use them yet (we have this dependency already though), but they will likely reduce the boilerplate around this sort of pattern a lot I expect:

- https://github.com/streamich/react-use/blob/master/docs/useAsync.md
- https://github.com/streamich/react-use/blob/master/docs/useAsyncRetry.md

PR #1352, opened 2021-04-29
add templates and recipes page

denisdimitrov:
Add new page with templates and recipes to choose from.

Partial implementation of https://github.com/sparkletown/internal-sparkle-issues/issues/680


src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -34,6 +34,7 @@ import { venueLandingUrl } from "utils/url";
 import { RoomsForm } from "pages/Admin/Venue/Rooms/RoomsForm";
 import { SchedulePage } from "pages/Schedule/SchedulePage";
 import { VenueAdminPage } from "pages/Admin/Venue/VenueAdminPage";
+import { AdminTemplates } from "pages/AdminTemplates";
```
0xdevalias, line 37: Should this be in `Pages/Admin/*`?

src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -60,6 +61,11 @@ const AppRouter = () => {
         <Route path="/admin/venue/creation" component={VenueWizard} />
         <Route path="/admin/venue/edit/:venueId" component={VenueWizard} />
 
+        <Route
+          exact
+          path="/admin_v2/templates/:venueId"
+          component={AdminTemplates}
+        />
```
0xdevalias, line 68: Why are we setting `exact` on this route? The only case where that should matter for this route is if we're defining additional subroutes in this same `Switch` (not nested within another component), that need to be matched. And in that case, the better solution is to change the order so that the more specific routes are listed before the more generic routes, so that they match first

- https://reactrouter.com/web/api/Switch
- https://reactrouter.com/web/api/Route/exact-bool

src/pages/Account/Venue/VenueMapEdition/PartyMapContainer.tsx
```diff
@@ -1,24 +1,24 @@
 import React, { useMemo } from "react";
+import { Room } from "types/rooms";
 
 import { ExtractProps } from "types/utility";
```
0xdevalias, line 4:
```suggestion

import { Room } from "types/rooms";
import { ExtractProps } from "types/utility";
```

src/pages/Account/Venue/VenueMapEdition/PartyMapContainer.tsx
```diff
@@ -1,24 +1,24 @@
...
-import { PartyMapVenue } from "types/venues";
 
 import { Container, SubVenueIconMap } from "./Container";
 
 type PartyMapContainerProps = Omit<
```
0xdevalias, line 8:
```suggestion
export type PartyMapContainerProps = Omit<
```

src/pages/Account/Venue/VenueMapEdition/PartyMapContainer.tsx
```diff
@@ -1,24 +1,24 @@
...
   "otherIcons"
 > & {
-  venue: PartyMapVenue;
+  rooms: Room[];
   currentRoomIndex?: number;
 };
 
 export const PartyMapContainer: React.FC<PartyMapContainerProps> = (props) => {
-  const { currentRoomIndex, venue, ...rest } = props;
+  const { currentRoomIndex, rooms, ...rest } = props;
```
0xdevalias, line 17: Is `props` ever passed on to anything else (doesn't look like it)? If not, can we use our standard pattern here?

```typescript
export const PartyMapContainer: React.FC<PartyMapContainerProps> = ({
  currentRoomIndex,
  rooms,
  ...rest
}) => {
```

src/pages/Account/Venue/VenueMapEdition/PartyMapContainer.tsx
```diff
@@ -1,24 +1,24 @@
...
 type PartyMapContainerProps = Omit<
   ExtractProps<typeof Container>,
   "otherIcons"
 > & {
-  venue: PartyMapVenue;
+  rooms: Room[];
   currentRoomIndex?: number;
 };
 
 export const PartyMapContainer: React.FC<PartyMapContainerProps> = (props) => {
```
0xdevalias, line 16: Since you've removed `PartyMapVenue`, there is seemingly nothing that makes this a `PartyMapContainer` anymore? Can we rename this to something that better represents it's purpose? I haven't looked too deeply what it does/etc, but off the top of my head, potentially `RoomsContainer` or similar?

```suggestion
export const RoomsContainer: React.FC<RoomsContainerProps> = (props) => {
```

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
+@import "scss/constants.scss";
+
+.templates-page {
```
0xdevalias, line 3:
```suggestion
.AdminTemplate {
```

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
...
+
+  &__nav-buttons {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    width: 80%;
+
+    &--cancel {
+      padding: 0.5rem 1.5rem;
+      border-radius: 22px;
```
0xdevalias, line 19: Can we extract/use this as a variable from `scss/constants` in a similar way to how it's defined in https://doesitsparkle.slack.com/archives/C01JHLMSFND/p1619732604371600?thread_ts=1619732501.371300&cid=C01JHLMSFND

image

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
...
+      background-color: lighten($color: $saturated-black, $amount: 10);
+      color: $white;
+      border: none;
+
+      &:hover {
+        background-color: lighten($color: $saturated-black, $amount: 20);
+      }
```
0xdevalias, line 26: Instead of inlining these `lighten` calls all across the app, it might be a better idea to define them in `scss/constants.scss`, perhaps named something like:

- `$saturated-black--lighten-10`
- `$saturated-black--lighten-20`
- etc

(this applies to all cases where this is used, though I'm only calling it out once here)

0xdevalias: Also, what are the extra `$color` and `$amount` in this call for? In the docs, that is just describing the positional parameters, it doesn't use named params:

- https://sass-lang.com/documentation/modules/color#lighten

So this should just be `lighten($saturated-black, 10)`, etc

src/pages/AdminTemplates/AdminTemplates.scss
0xdevalias: (and all of the other places this is used)

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
...
+
+      &:focus {
+        outline: none;
+      }
+    }
+  }
+
+  &__template-preview {
+    width: 60%;
+    height: 60vh;
```
0xdevalias, line 52: Why are we using `60vh` here, rather than a `%`/etc? In most cases if not all cases we want to stick to `%`.

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  &__recipes {
+    border-radius: 22px;
+    background-color: lighten($color: $saturated-black, $amount: 10);
+    color: $white;
+    border: none;
+  }
+
+  &__recipe {
```
0xdevalias, line 94: I would have to look closer at the structure of the code, but there seem to be a lot of `__elements` here that i'm guessing may make sense to be extracted into sub-components, rather than all combined into this one file.

src/pages/AdminTemplates/AdminTemplates.scss
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    &--disabled {
+      opacity: 0.5;
+    }
+  }
+
+  &__template-background {
+    width: 100%;
+    height: 400px;
```
0xdevalias, line 154: Things like this should be extracted into variables.

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
+import classNames from "classnames";
+import React, { useState } from "react";
```
0xdevalias, line 2:
```suggestion
import React, { useState } from "react";
import classNames from "classnames";
```

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import { DEFAULT_MAP_BACKGROUND } from "settings";
+
+import { Room } from "types/rooms";
+
+import { PartyMapContainer } from "pages/Account/Venue/VenueMapEdition";
+
+import "./AdminTemplates.scss";
+
+interface TemplateType {
```
0xdevalias, line 12:
```suggestion
export interface TemplateType {
```

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+  title: string;
+  description: string;
+  background: string;
+}
+
+interface RecipeType {
```
0xdevalias, line 18:
```suggestion
export interface RecipeType {
```

src/pages/AdminTemplates/AdminTemplates.tsx
0xdevalias: Also, we generally don't suffix our types with `Type`. Why are we doing that here?

It looks like this should just be `Template` or similar.

src/pages/AdminTemplates/AdminTemplates.tsx
0xdevalias: Also, we generally don't suffix our types with `Type`. Why are we doing that here?

It looks like this should just be `Recipe` or similar.

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+    background: DEFAULT_MAP_BACKGROUND,
+  },
+];
+
+const recipes: RecipeType[] = [
+  { title: "Recipe 1", rooms: [] },
+  { title: "Recipe 2", rooms: [] },
+  { title: "Recipe 3", rooms: [] },
+  { title: "Blank Canvas", rooms: [] },
+];
```
0xdevalias, line 51: What's the purpose of this?

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export const AdminTemplates: React.FC = () => {
```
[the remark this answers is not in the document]
0xdevalias: This is probably valid, and I expect we'll want to extract chunks of this component into sub-components rather than having it all defined in this one place.

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const selectTemplate = (template: TemplateType) => {
+    if (isLoading) return;
+
+    setLoading(true);
+
+    setTimeout(() => {
+      setSelectedTemplate(template);
+      setLoading(false);
+    }, 2000);
+  };
```
0xdevalias, line 67: `useCallback`

Extract the timeout into a `const` (probably in `src/settings`) or reuse a relevant one that already exists there rather than using a 'magic number' here.

We probably want to be 'triggering' the `setTimeout` within a `useEffect` and also ensuring that we return a cleanup function that cancels the timeout.

0xdevalias: Also.. what's the intent/purpose of this code in general? It looks like we're introducing an arbitrary delay for... no reason?

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+              onClick={() => console.log("asd")}
+            >
+              Choose this template
+            </button>
+          </div>
+          <div className="templates-page__title">{selectedTemplate.title}</div>
+          <div className="templates-page__subtitle">
+            {selectedTemplate.description}
+          </div>
+          <div className="templates-page__recipes">
```
0xdevalias, line 91: Splitting up 'sibling' code chunks by purpose makes it much easier to read/reason about.

```suggestion
          </div>

          <div className="templates-page__title">{selectedTemplate.title}</div>
          <div className="templates-page__subtitle">
            {selectedTemplate.description}
          </div>

          <div className="templates-page__recipes">
```

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+              <button
+                key={`${recipe.title}-${index}`}
+                onClick={() => setSelectedRecipe(recipe)}
+                className={classNames("templates-page__recipe", {
+                  "templates-page__recipe--selected": recipe === selectedRecipe,
+                })}
+              >
+                {recipe.title}
+              </button>
+            ))}
```
0xdevalias, line 102: `useMemo`

We also ideally don't want to use the `index` as part of the `key` here.

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+          </div>
+          <div className="templates-page__template-preview">
```
0xdevalias, line 104:
```suggestion
          </div>

          <div className="templates-page__template-preview">
```

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+            <PartyMapContainer
+              rooms={[]}
+              iconsMap={{}}
+              coordinatesBoundary={{ width: 60, height: 60 }}
+              interactive={false}
+              resizable={false}
+              backgroundImage={selectedTemplate.background}
+            />
```
0xdevalias, line 112: Why are `rooms`/`iconMap`/etc empty here?

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  {template.title}
+                </div>
+                <div className="templates-page__template-description">
+                  {template.description}
+                </div>
+              </div>
+            ))}
+          </div>
+        </>
+      )}
```
0xdevalias, line 151: This should be extracted out into a sub-component and/or refactored. There's a lot of duplication in structure within this block and the block above it. We shouldn't be doing that. Common structure should use the same component, and then for the 'slots' that need to differ, we can 'inject' that content as needed.

src/pages/AdminTemplates/AdminTemplates.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  className="templates-page__template-background"
+                />
+                <div className="templates-page__template-title">
+                  {template.title}
+                </div>
+                <div className="templates-page__template-description">
+                  {template.description}
+                </div>
+              </div>
+            ))}
```
0xdevalias, line 148: `useMemo` + probably also worth extracting chunks of this into a sub-component. From a quick skim it kind of sounds like it might be a `TemplateCard` or similar.

src/pages/AdminTemplates/AdminTemplates.tsx
0xdevalias: They both seem to have `title`, `subtitle`, a button, then their 'main content' section.

0xdevalias (changes requested):
Questions/comments/changes.

Also, please either link this to the backlog issue with closing keywords, or add it directly to the sprint project board if there is no backlog issue for it.

PR #1358, opened 2021-05-05
Initial approach on posterhall performance optimisation

mike-lvov:
Given that the amount of posters in `posterhall` is expected to be > than 1600, we need a way to display the thing in an efficient & performant way

### DIFF
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -1,4 +1,4 @@
-import { useState, useMemo } from "react";
+import { useState, useMemo, useCallback } from "react";
 import { VenueTemplate } from "types/venues";
 import Fuse from "fuse.js";
 
@@ -7,6 +7,8 @@ import { posterVenuesSelector } from "utils/selectors";
 import { isLoaded, useFirestoreConnect } from "./useFirestoreConnect";
 import { useSelector } from "./useSelector";
 import { useDebounceSearch } from "./useDebounceSearch";
+
+const POSTERS_PER_PAGE = 12;
 
 export const useConnectPosterVenues = (posterHallId: string) => {
   useFirestoreConnect(() => {
@@ -38,7 +40,7 @@ export const usePosterVenues = (posterHallId: string) => {
 };
 
 export const usePosters = (posterHallId: string) => {
-  const { posterVenues } = usePosterVenues(posterHallId);
+  const { posterVenues, isPostersLoaded } = usePosterVenues(posterHallId);
 
   const {
     searchInputValue,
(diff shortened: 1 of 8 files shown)


src/hooks/useDebounceSearch.ts
```diff
@@ -15,9 +15,13 @@ export const useDebounceSearch = () => {
...
     () =>
-      debounce((value: string) => {
-        setSearchQuery(value.toLowerCase());
-      }, SEARCH_DEBOUNCE_TIME),
+      debounce(
+        (value: string) => {
+          setSearchQuery(value.toLowerCase());
+        },
+        SEARCH_DEBOUNCE_TIME,
+        { maxWait: SEARCH_DEBOUNCE_TIME * 3 }
```
mike-lvov, line 23: For some reason, the execution of the denounced func was delayed for unreasonable amount of time (>5 seconds)

mike-lvov:
@0xdevalias merging this just to test the changes with real-world data. Please, make a retrospective review of this

0xdevalias:
@mike-lvov Please make sure to add these 'standalone PRs' (eg. where they aren't linked to an issue) directly to the Sprint project board so that we have an accurate reference of the work done within the sprint.

src/hooks/useDebounceSearch.ts
0xdevalias: Was this while experiencing other performance issues/where the browser was already struggling? If so, might have been related to that more than our use of `debounce` here?

0xdevalias: If we keep this, might be worth extracting `SEARCH_DEBOUNCE_TIME * 3` as a `const` in settings as well.

src/hooks/posters.ts
```diff
@@ -47,37 +49,51 @@ export const usePosters = (posterHallId: string) => {
...
+  const [displayedPostersCount, setDisplayedPostersAmount] = useState(
+    DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT
+  );
+
+  const increaseDisplayedPosterCount = useCallback(() => {
+    setDisplayedPostersAmount(
+      (prevPostersNumber) =>
+        prevPostersNumber + DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT
+    );
+  }, []);
```
0xdevalias, line 61: Would this be better named something like `showMorePosters` or similar?

```typescript
  const showMorePosters = useCallback(() => {
    setDisplayedPostersAmount(
      (prevCount) =>
        prevCount + DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT
    );
  }, []);
```

src/hooks/posters.ts
```diff
@@ -47,37 +49,51 @@ export const usePosters = (posterHallId: string) => {
...
 
   const fuseVenues = useMemo(
     () =>
-      new Fuse(posterVenues, {
+      new Fuse(filteredPosterVenues, {
         keys: ["poster.title", "poster.authorName", "poster.categories"],
       }),
-    [posterVenues]
+    [filteredPosterVenues]
   );
```
0xdevalias, line 77: Why did we move `filteredPosterVenues` up here + make `fuseVenues` consume it?

src/hooks/posters.ts
```diff
@@ -47,37 +49,51 @@ export const usePosters = (posterHallId: string) => {
...
 
   const searchedPosterVenues = useMemo(() => {
-    if (!searchQuery) return posterVenues;
+    if (!searchQuery)
+      return filteredPosterVenues.slice(0, displayedPostersCount);
 
     return fuseVenues
       .search(searchQuery)
+      .slice(0, displayedPostersCount)
```
0xdevalias, line 85: I can see what you're doing, using `.slice` to reduce the number of displayed venues, but I expect the crux of the performance issue here is either in the initial loading of the data, or in processing the parts above. I would be very surprised if it was actually at the 'rendering' part after here that everything slowed down.

0xdevalias (commented):
[Retroactive Review] Retroactive review comments.

PR #1364, opened 2021-05-06
[env/ohbm] modified add-event.js for uploading ohbm schedule

margulies:
> modified `scripts/add-event.js` for uploading ohbm schedule


scripts/add-event.js
```diff
@@ -1,23 +1,22 @@
...
 ${process.argv[1]}: Add an event to a venue
 
 Specify the name, time, duration in minutes, description, collective price, and price.
 
-Usage: node ${process.argv[1]} API_KEY PROJECT_ID VENUE UTC_TIME DURATION_MINUTES PRICE COLLECTIVE_PRICE NAME DESCRIPTION
+Usage: node ${process.argv[1]} API_KEY PROJECT_ID VENUE UTC_TIME DURATION_MINUTES PRICE COLLECTIVE_PRICE NAME DESCRIPTION USERNAME PASSWORD
 
-Example: node ${process.argv[1]} aaazzz111222333 co-reality-map examplevenue 2020-08-01T19:00:00Z 180 500 2500 "The Virtual Jazz Bar" "Join us in Virtual jazz bar on Saturday for three sets of fabulous jazz from the House band."
+Example: node ${process.argv[1]} aaazzz111222333 co-reality-map examplevenue 2020-08-01T19:00:00Z 180 500 2500 "The Virtual Jazz Bar" "Join us in Virtual jazz bar on Saturday for three sets of fabulous jazz from the House band." username password
 `);
```
0xdevalias, line 14: Instead of adding username/password fields to the params like this (which could be a security concern, as they are then going to be logged in the command history in your shell), it might be a better idea to refactor this script to use the firebase admin SDK + service account key pattern that most of our current/modern scripts use. This would also allow us to clean up/simplify a bunch of the patterns used here at the same time.

At the same time, we could also change this to a `.ts` file and rewrite it to benefit from TypeScript types rather than being plain JavaScript.

scripts/add-event.js
```diff
@@ -29,45 +28,40 @@ const price = argv[5];
...
-read({ prompt: "Username:" }, function (err, username) {
-  if (err) {
-    console.error("Error obtaining username:", err);
-    process.exit(1);
-  }
-  read({ prompt: "Password:", silent: true }, function (err, password) {
-    if (err) {
-      console.error("Error obtaining password:", err);
-      process.exit(1);
-    }
```
0xdevalias, line 42: The benefit of reading these credentials in an interactive way like this is that they then won't be logged in the shell's command history, which could pose a security risk.

scripts/add-event.js
```diff
@@ -29,45 +28,40 @@ const price = argv[5];
...
       .catch(function (err) {
-        console.error("Login error:", err);
+        console.error("Add event error:", err);
+        process.exit(0);
       });
+  })
+  .catch(function (err) {
+    console.error("Login error:", err);
+    process.exit(0);
   });
```
0xdevalias, line 67: We could simplify the nesting here by using `async`/`await` patterns.

0xdevalias (changes requested):
Quick review pass + some thoughts/comments.

scripts/add-event.js
margulies: I did notice some issues there that async/await might resolve. Regarding the username/password, no worries — I was simply sharing this script at the request of @mike-lvov to show how I was reading in the data, but no need to pull it into the main repo, as there aren't really any core changes to the content from the current version.

margulies:
Closing this as the minor modifications have been reverted

PR #1366, opened 2021-05-06
remove unused legacy getVenueEvents backend function

mike-lvov:
While I was getting into the way events are saved/fetched, found this unused fn

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -4,7 +4,6 @@ const { checkAuth } = require("./auth");
 const { HttpsError } = require("firebase-functions/lib/providers/https");
 
 const PLAYA_VENUE_ID = "jamonline";
-const MAX_TRANSIENT_EVENT_DURATION_HOURS = 6;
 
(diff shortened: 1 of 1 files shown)


0xdevalias (approved):
I believe this was the legacy way that we implemented the schedule feature before moving to using `useConnectRelatedVenues`. We ditched it because it was chewing up a lot of serverless function time, and relatively costing a lot to run.

LGTM! 💥

0xdevalias:
@mike-lvov FYI the CI build for this is failing when trying to deploy the functions. Looks like our current CI script doesn’t pass correct flags to say ‘yes it’s ok to delete functions that are no longer present in our config’, which causes it to hang waiting for user input until it times out.

- https://app.circleci.com/pipelines/github/sparkletown/sparkle/6062/workflows/25660752-e536-433c-9677-2cd72c454c79/jobs/9821/parallel-runs/0/steps/0-106

I expect the fix for this is likely going to be a line or two change in our Circle CI config.

PR #1370, opened 2021-05-08
add RoomTypes.video + new RoomModal mode that displays an embedded video iframe

margulies:
User request to have a clickable room icon that displays an embedded video within ~the PartyMap template~ the `RoomModal` (on `PartyMap`). ~Specific fix for env/ohbm~

**Edits by @0xdevalias:**

Not directly related, but similar to https://github.com/sparkletown/sparkle/pull/1427


margulies:
Apologies — just noticed this branch includes unnecessary commits. The relevant one here is 7d0c27dc286fab2a3280b886a7118d50b7fe0787 . If I should close and redo the PR, just let me know

0xdevalias:
> Apologies — just noticed this branch includes unnecessary commits. The relevant one here is [7d0c27d](https://github.com/sparkletown/sparkle/commit/7d0c27dc286fab2a3280b886a7118d50b7fe0787) . If I should close and redo the PR, just let me know

@margulies is it just extra commits, or are this extra commits still showing in the diffs on the files tab as unrelated changes? If the latter, that will need to be fixed up before this PR would be mergeable.

If you’re comfortable with rebasing branches/similar (and the source branch you made this PR from doesn’t need to have those extra commits) then you could leave this PR open and just do that + force push over that branch to align it to just what you intended.

If you’re not comfortable doing that, then it might be easier to make a new branch that only contains the minimal commits you intended (potentially by using git’s cherry pick or similar) and then submit that as a new PR that replaces this.

Either way works for me, just depends what you’re more comfortable with.

margulies:
Thanks for explaining — the only additional file that shouldn't be there is `scripts/add-event.js`. All other changes are ready for review. I'll revert `scripts/add-event.js` to the current version tomorrow.

0xdevalias:
> Thanks for explaining — the only additional file that shouldn't be there is `scripts/add-event.js`. All other changes are ready for review. I'll revert `scripts/add-event.js` to the current version tomorrow.

Cool, sounds good.

margulies:
Thanks again for the tip — PR is now cleaned up to the one intended commit. Happy for any comments.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -52,13 +52,20 @@ export const PartyMap: React.FC = () => {
...
+          onHide={unselectRoom}
+        />
+      ) : (
+        <RoomModal
+          room={selectedRoom}
+          venue={currentVenue}
+          show={hasSelectedRoom}
+          onHide={unselectRoom}
+        />
+      )}
```
0xdevalias, line 68: [note] Here you render either `RoomModalVideo` or `RoomModal` based on the `RoomTypes.video`

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -43,6 +46,29 @@ export const RoomModal: React.FC<RoomModalProps> = ({
...
+  >
+    <iframe
+      className="IntroVideoPreviewModal__iframe"
+      title="IntroVideoPreviewModal-iframe"
+      src={room?.url + `?autoplay=1`}
+      allow={IFRAME_ALLOW}
+      allowFullScreen
+    />
+  </Modal>
+);
```
0xdevalias, line 70: [note] Yet here you only show the intro video, with none of the 'normal' `RoomModal` features (including the ability to enter the venue)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -16,6 +16,9 @@ import UserList from "components/molecules/UserList";
 
 import { RoomModalOngoingEvent, ScheduleItem } from "..";
 
+import { IFRAME_ALLOW } from "settings";
+import "components/templates/PosterPage/components/IntroVideoPreviewModal/IntroVideoPreviewModal.scss";
```
0xdevalias, line 20: What are these styles imported here for? We don't tend to 'cross-import' component specific styles from a different/unrelated component like this.

0xdevalias: Though looking at the `IntroVideoPreviewModal` component, it seems practically identical to the  implementation here, except that it sets the `url` to the `introVideoUrl` prop (which is passed into it anyway).

So why wouldn't we just reused that component in its entirety here?

In doing so, it would pretty much make `RoomModalVideo` an irrelevant component.

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias: Instead of handling this different room type selection at the `PartyMap` layer, I would expect it's more of a distinction to be handled within the `RoomModal` component itself. That's where we already handle `RoomTypes.unclickable` currently.

src/components/templates/PartyMap/components/index.ts
```diff
@@ -2,5 +2,6 @@ export { Map } from "./Map/Map";
 export { PartyTitle } from "./PartyTitle/PartyTitle";
 export { RoomAttendance } from "./RoomAttendance";
 export { RoomModal } from "./RoomModal/RoomModal";
+export { RoomModalVideo } from "./RoomModal/RoomModal";
```
0xdevalias, line 5:
```suggestion
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -43,6 +46,29 @@ export const RoomModal: React.FC<RoomModalProps> = ({
...
+    <iframe
+      className="IntroVideoPreviewModal__iframe"
+      title="IntroVideoPreviewModal-iframe"
+      src={room?.url + `?autoplay=1`}
+      allow={IFRAME_ALLOW}
+      allowFullScreen
+    />
+  </Modal>
+);
+
```
0xdevalias, line 71:
```suggestion
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -16,6 +16,9 @@ import UserList from "components/molecules/UserList";
 
 import { RoomModalOngoingEvent, ScheduleItem } from "..";
 
+import { IFRAME_ALLOW } from "settings";
+import "components/templates/PosterPage/components/IntroVideoPreviewModal/IntroVideoPreviewModal.scss";
+
```
0xdevalias, line 21:
```suggestion
```

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -10,7 +10,7 @@ import { useSelector } from "hooks/useSelector";
 import { useUser } from "hooks/useUser";
 import useConnectCurrentVenue from "hooks/useConnectCurrentVenue";
 
-import { Map, RoomModal } from "./components";
+import { Map, RoomModal, RoomModalVideo } from "./components";
```
0xdevalias, line 13:
```suggestion
import { Map, RoomModal } from "./components";
```

src/components/templates/PartyMap/PartyMap.tsx
0xdevalias:
```suggestion
      <RoomModal
        room={selectedRoom}
        venue={currentVenue}
        show={hasSelectedRoom}
        onHide={unselectRoom}
      />
```

0xdevalias: So that we can reuse `IntroVideoPreviewModal` in its entirety, within `RoomModal` I would add a `switch`/`case` (similar to what is done in `src/pages/VenuePage/TemplateWrapper.tsx`) that switches on `room.type`, and chooses to render either `IntroVideoPreviewModal` (if it's `RoomTypes.video`), or `RoomModalContent` (as the `default` case)

This would end up looking roughly something like this:

```typescript
switch (room.type) {
  case RoomTypes.video:
    return <IntroVideoPreviewModal introVideoUrl={room?.url + `?autoplay=1`} isVisible={show} onHide={onHide}>

  default:
    return <RoomModalContent room={room} venueName={venue.name}>
}
```

0xdevalias: Even better than using

```
room?.url + `?autoplay=1`
```

might be to use our `ConvertToEmbeddableUrl` util function that already handles this sort of thing?

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/ConvertToEmbeddableUrl.tsx

0xdevalias: Searching for usages of `ConvertToEmbeddableUrl`:

- https://github.com/sparkletown/sparkle/search?q=ConvertToEmbeddableUrl

I came across `VideoModal`, which might be an even better choice than `IntroVideoPreviewModal` here:

- https://github.com/sparkletown/sparkle/blob/staging/src/components/organisms/VideoModal/VideoModal.tsx

---

@mike-lvov What do you think? Any reason you would/wouldn't use `VideoModal` instead of `IntroVideoPreviewModal` here?

0xdevalias (changes requested):
> User request to have a clickable room icon that displays an embedded video within the partymap template. Specific fix for env/ohbm

@margulies Is this intended to be an entirely separate 'room' that ONLY shows the video in a popup modal, but doesn't actually 'let the user otherwise 'go anywhere'?

---

I can see value in including this in our core product, so might be worth re-targeting this branch from `env/ohbm` to `staging` (+ rebasing/etc to fix any unrelated commits that causes)

sofisparkle:
@0xdevalias is this https://github.com/sparkletown/internal-sparkle-issues/issues/472 this?!

0xdevalias:
> @0xdevalias is this [sparkletown/internal-sparkle-issues#472](https://github.com/sparkletown/internal-sparkle-issues/issues/472) this?!

@sofisparkle Unfortunately not. This PR is about creating a normal room image, that when clicked on will open a modal that contains an embedded video instead of the usual 'room modal' that lets you join a venue.

Off the top of my head, I expect that implementing https://github.com/sparkletown/internal-sparkle-issues/issues/472 would follow a similar pattern of adding a new `RoomsType`, but instead of changing the `RoomModal` component, we would need to change `src/components/templates/PartyMap/components/Map/MapRoom.tsx` (which controls how it is rendered on the map) to return an `iframe`/similar.

0xdevalias:
> > is this sparkletown/internal-sparkle-issues#472 this?!
>
> Unfortunately not. This PR is about creating a normal room image, that when clicked on will open a modal that contains an embedded video instead of the usual 'room modal' that lets you join a venue.
> 
> Off the top of my head, I expect that implementing [sparkletown/internal-sparkle-issues#472](https://github.com/sparkletown/internal-sparkle-issues/issues/472) (in its most basic form) would follow a similar pattern of adding a new `RoomsType`, but instead of changing the `RoomModal` component, we would need to change `src/components/templates/PartyMap/components/Map/MapRoom.tsx` (which controls how it is rendered on the map) to return an `iframe`/similar.

@sofisparkle However https://github.com/sparkletown/sparkle/pull/1427 that @margulies just opened is exactly related to resolving https://github.com/sparkletown/internal-sparkle-issues/issues/472

0xdevalias:
@margulies I have changed the base of this PR from `env/ohbm` to `staging`, that is why all of the extra commits are showing now. To fix this, you'll need to rebase your changes onto `staging` + force push them to this branch. There's no need to close this PR/open a new one.

margulies:
I've completely reworked this PR based on the recent merging of the iframe room type. This also now includes a video and a pdf modal room type. The specific use case for the pdf modal is offering a clickable summary schedule or other detailed information that may be useful to participants during the event.

0xdevalias:
> This also now includes a video and a pdf modal room type. The specific use case for the pdf modal is offering a clickable summary schedule or other detailed information that may be useful to participants during the event.

How does the PDF modal use case differ from just providing a PDF to the existing iframe mode, given it's just using an iframe anyway?

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -37,13 +39,38 @@ export const RoomModal: React.FC<RoomModalProps> = ({
...
+      template = venue && room && (
+        <Modal show={show} onHide={onHide}>
+          <div className="room-modal">
+            <RoomModalContent room={room} venueName={venue.name} />
+          </div>
+        </Modal>
+      );
+  }
+  return <>{template}</>;
+};
```
0xdevalias, line 73: Probably no need to use `let`, the fragment around `template`, etc here. Looks like you may as well just `return` straight from the `switch`/`case` options?

margulies:
> > This also now includes a video and a pdf modal room type. The specific use case for the pdf modal is offering a clickable summary schedule or other detailed information that may be useful to participants during the event.
> 
> @margulies How does the PDF modal use case differ from just providing a PDF to the existing iframe mode, given it's just using an iframe anyway?

The specific use case when I found this helpful is for providing a detailed PDF that contains reference information — for example, a detailed schedule or information sheet. It's too much and too busy to embed in the background, but very helpful to have it quickly and easily accessible in modal form from the partymap page (rather than needing to enter into another venue to view). This also allows the PDF (and user) to stay within sparkle rather than switching between PDF and sparkle for referencing. 

(I find this feature useful, but if the video room type seems ready to go, and the pdf room type will require substantially more work, happy to remove from this PR so as not to delay the other feature.)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: This is now fixed. Thanks

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: now implemented in refactored version

src/components/templates/PartyMap/PartyMap.tsx
margulies: Implemented in new completely refactored version

mike-lvov (approved):
@margulies removed the PDF type in https://github.com/sparkletown/sparkle/pull/1370/commits/aa3cabbdb72b27a9e6cbbe3ed2430107b58262ee

If we decide to implement it, we could just copy from there

PR #1302, opened 2021-04-02
refactor Event Schedule UI for new design + add personal schedule

igachmovmelon:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/485

* Removed `LiveSchedule` component as it's no longer used https://github.com/sparkletown/sparkle/pull/1302/commits/a6cd04b7351b3b004d58205991ae7fd7e65d4ead
* Removed `SchedulePage` component as its' no longer used https://github.com/sparkletown/sparkle/pull/1302/commits/a6cd04b7351b3b004d58205991ae7fd7e65d4ead

Preview:
image

Feature structure:
```
SchedulePageModal                           <---  I suggest renaming it to ScheduleModal
├── ScheduleVenueDescription
├── Schedule
       ├── rooms 
       ├── ScheduleRoomEvents
               ├── ScheduleEvent
```

`SchedulePageModal` is the main component which appears and hides on `Schedule` button click.
`ScheduleVenueDescription` contains general information about the parent (if exists) or current venue.
`Schedule` the component which contains all the information about events of the selected day.
`ScheduleRoomEvents` is just a map of events in the room line.
`ScheduleEvent` is the component of the particular event.

There is a component `EventDisplay` which was modified by Ivan (`ScheduleEvent` in my case). But it's used in several places and I'm not sure if it was safe to modify it. If we don't have anything schedule-related in the project, we might want to remove that piece as well. Open to your thoughts.

PR #1373, opened 2021-05-09
[env/ohbm] add menu to select start date for schedule

margulies:
This is still a bit messy, but allows the user to select the starting date for the schedule. The selection doesn't persist across reloadings, but at least allows for viewing the educational course and conference schedule at this stage. 

---

**Edit by @0xdevalias:** Semi/potentially related: https://github.com/sparkletown/sparkle/pull/1419

### DIFF
--- a/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
+++ b/src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
@@ -14,6 +14,8 @@ import { useVenueId } from "hooks/useVenueId";
 
 import { EventDisplay } from "components/molecules/EventDisplay/EventDisplay";
 
+import { START_DATE } from "settings";
+
 type DatedEvents = Array<{
   dateDay: Date;
   events: Array<WithVenueId<VenueEvent>>;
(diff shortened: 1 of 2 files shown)


0xdevalias:
**Note:** There are heavy refactors coming to this `Schedule` feature in https://github.com/sparkletown/sparkle/pull/1302

src/settings.ts
```diff
@@ -566,3 +566,5 @@ export const POSTERPAGE_MAX_VIDEO_PARTICIPANTS = 10;
 export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 12;
+
+export const START_DATE = "conference"; // options: "conference", "course", "today"
```
0xdevalias, line 570: A little hacky (I would usually extract this into a type on it's own), but you can use TypeScript to restrict this to the values you want.

```suggestion
export const START_DATE: "conference" | "course" | "today" = "conference";
```

src/components/organisms/SchedulePageModal/SchedulePageModal.tsx
```diff
@@ -43,14 +45,56 @@ export const SchedulePageModal: FC<SchedulePageModalProps> = ({
...
+            >
+              <option value="today">Today</option>
+              <option value="course">Educational Course</option>
+              <option value="conference">Conference</option>
+            </select>
+          </div>
+        </label>
+      </form>
+    </>
+  );
```
0xdevalias, line 74: This should either be extracted into a sub-component, or just rendered inline in the JSX below. We try not to use this 'render function' pattern (and where it exists currently is legacy code).

You also don't need to wrap this in the fragment tags (`<>` `</>`) as you already have a single root element (`form`)

0xdevalias: The `onChange` value would need to be extracted into a semantically named `const` and `useCallback`'d, otherwise it will cause re-renders due to the inline function definition.

0xdevalias (changes requested):
@margulies With the big refactor/changes to the Schedule that just landed yesterday, I expect this PR will need to be updated for those.

margulies:
With the other schedule PR now merged, I think this is redundant and can be closed.

margulies:
Closing this as it was addressed in #1419, which has already been merged.

PR #1368, opened 2021-05-07
refactor related venues + event fetching for performance/efficiency

0xdevalias:
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/699

**Note:** This is an alternative / additional implementation plan to https://github.com/sparkletown/sparkle/pull/1367

## TODO

- [x] Implement our context/provider pattern for `useRelatedVenues` ([ref](https://github.com/sparkletown/sparkle/pull/1368#discussion_r629135576))
- [x] Add custom code timing/performance traces to the new code (using `utils/performance`)
  - [x] `fetchSovereignVenue` in `useSovereignVenueId` (fixed in https://github.com/sparkletown/sparkle/pull/1368/commits/a3b4399ef2d47672bd1e36c00ecc3bee561967eb)
  - [x] `fetchRelatedVenues` in `RelatedVenuesProvider` (fixed in https://github.com/sparkletown/sparkle/pull/1368/commits/733cd534d1d4c8c39bdaa52135455fef42067bc3)
  - [x] `fetchAllVenueEvents` in `useVenueEvents` (fixed in https://github.com/sparkletown/sparkle/pull/1368/commits/778bcc9bacafed1d5d74f8a8615e0c01a7395538)
  - [x] Anything else?
- [x] Add code to fetch events for a given array of `venueIds` ~using our 'chunked array in query' pattern~ (can't use that pattern as they are sub-collections) (fixed in https://github.com/sparkletown/sparkle/pull/1368/commits/6ecb7a3126838a5dfed4390598d18c22c7d7b0ed)
- [x] Refactor the codebase to make use of `useRelatedVenues` rather than legacy `useLegacyConnectRelatedVenues` ([ref](https://github.com/sparkletown/sparkle/pull/1368#discussion_r629133383)) (fixed in https://github.com/sparkletown/sparkle/commit/07a9e0e7a64dc56936c88853fd02eb79077e38a0 + https://github.com/sparkletown/sparkle/commit/e905c452a719589a6779047deccdf8c30aa93519 + https://github.com/sparkletown/sparkle/commit/b061f26feec4fe48b393ac38d67d814f528d6990)
- [x] Remove `useLegacyConnectRelatedVenues` once nothing needs it anymore (removed in https://github.com/sparkletown/sparkle/pull/1368/commits/4fecc7e204ff30200009a0ee52f23f74843806bf)
- [x] Anything else?

## Potentially-OOS?

[...]

PR #1375, opened 2021-05-11
poster bookmarking

margulies:
Aim is to enable bookmarking of individual posters, which also adds an event to personalized schedule for poster standby time.

~~Clickable userAvatar on poster card in posterhall template also included in initial form.~~ removed from this PR as per https://github.com/sparkletown/sparkle/pull/1375#issuecomment-845611297

~~**Note:** This branch is building on the PR for new Schedule UI (https://github.com/sparkletown/sparkle/pull/1302) and will require that code to be merged first.~~ this is merged now


0xdevalias:
> One unresolved issue is that events from posters are hierarchically two steps away from /in/home, and don't seem to get picked up for the schedule in that venue.

@margulies My changes in https://github.com/sparkletown/sparkle/pull/1368 will resolve that bug as part of my cleanup/performance refactor

0xdevalias:
> There are still some issues to resolve, but this has all the basic features of filtering out poster events, but pulling them into the personalized schedule if bookmarked.

@margulies This PR is currently in draft. Is it still a WIP, or do you consider it to be ready for review?

0xdevalias:
@margulies Now that the Schedule changes have been merged, it might be worth rebasing this branch + force pushing to clean it up to only the related commits + fix the conflicting files.

margulies:
> > There are still some issues to resolve, but this has all the basic features of filtering out poster events, but pulling them into the personalized schedule if bookmarked.
> 
> @margulies This PR is currently in draft. Is it still a WIP, or do you consider it to be ready for review?
> 
> By the looks of it, it seems to contain a bunch of unrelated commits, so I assume it's based off an invalid branch/similar to be merged into `env/ohbm`. Is this feature relevant to the wider platform in general? If so, we should probably target `staging` with it (which I assume is what it was branched from originally)

I've now rebased on staging, and will continue to develop from there.

margulies:
I've also removed the clickable user avatar feature from this PR to focus here on the bookmarking.

0xdevalias:
@margulies Looks like there are conflicts with `NavBarSchedule`

image

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faBookmark as solidBookmark } from "@fortawesome/free-solid-svg-icons";
+import { faBookmark as regularBookmark } from "@fortawesome/free-regular-svg-icons";
+
+import { updatePersonalizedSchedule, savePosterToProfile } from "api/profile";
+import { useUser } from "hooks/useUser";
+
+// import { UserAvatar } from "components/atoms/UserAvatar";
+// import { useProfileModalControls } from "hooks/useProfileModalControls";
+// import { useWorldUsersById } from "hooks/users";
```
0xdevalias, line 25: Why is this commented out here?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
     [categories]
   );
 
+  const { relatedVenueIds } = useRelatedVenues({
+    //isLoading, relatedVenues,
```
0xdevalias, line 78: Why is this commented out here?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+    currentVenueId: venueId,
+  });
+
+  const {
+    // isEventsLoading,
```
0xdevalias, line 83: Why is this commented out here?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+  );
+
+  // wrap author in memo:
+  // const { worldUsersById } = useWorldUsersById();
+  // const authors = personalizedPoster.owners;
+  // const author = { ...worldUsersById[authors[0]], id: authors[0] };
+  // const { openUserProfileModal } = useProfileModalControls();
+  // const openAuthorProfile = useCallback(() => {
+  //   openUserProfileModal(author);
+  // }, [openUserProfileModal, author]);
```
0xdevalias, line 129: Why is this commented out here?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
-
       <div className="PosterPreview__author">{authorName}</div>
+      {/* don't show author profile if not available... */}
+      {/* {author && (
+        <UserAvatar
+          user={author}
+          // isOnline={}
+          onClick={openAuthorProfile}
+        />
+      )} */}
```
0xdevalias, line 148: Why is this commented out here?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+      <div className="PosterPreview__bookmark" onClick={onBookmarkPoster}>
+        <FontAwesomeIcon
+          icon={isBookmarkedPoster ? solidBookmark : regularBookmark}
+        />
+      </div>
       <p className="PosterPreview__title">{title}</p>
-
       <div className="PosterPreview__categories">{renderedCategories}</div>
-
       <div className="PosterPreview__author">{authorName}</div>
```
0xdevalias, line 140: We probably don't want to remove these whitespace lines here.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
```diff
@@ -56,4 +56,15 @@ $active-poster-shadows: 0 10px 20px 0 rgba($primary, 0.56);
...
+  &__bookmark {
+    cursor: pointer;
+    height: 100%;
+    display: flex;
+    justify-content: right;
+    align-items: right;
+    top: 0%;
+    padding: 0.5rem;
+    color: rgb(255, 255, 255);
+  }
```
0xdevalias, line 69: Where these styles written from scratch, or copied from somewhere? They seem to differ from `&__bookmark` in `ScheduleEvent.scss`. Why is that?

- https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/ScheduleEvent/ScheduleEvent.scss#L88-L98

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
 export interface PosterPreviewProps {
-  posterVenue: WithId<PosterPageVenue>;
   enterVenue: (venueId: string) => void;
+  personalizedPoster: WithId<PersonalizedPoster>;
 }
 
+const emptyRelatedEvents: WithVenueId<VenueEvent>[] = [];
+export const emptyPersonalizedSchedule = {};
+
 export const PosterPreview: React.FC<PosterPreviewProps> = ({
```
0xdevalias, line 42: Keep `FooComponentProps` next to `FooComponent`.

---

Also, what's the need for the `emptyRelatedEvents` and `emptyPersonalizedSchedule` here? Particularly given this file is `PosterPreview` and seems to not relate to the schedule?

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -31,15 +53,19 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+    return posterVenuesPersonalized.map((posterVenue) => (
       <PosterPreview
         key={posterVenue.id}
-        posterVenue={posterVenue}
         enterVenue={enterVenue}
+        personalizedPoster={posterVenue}
       />
     ));
-  }, [posterVenues]);
+  }, [posterVenuesPersonalized]);
```
0xdevalias, line 68: Since all of the changes here related to `userPosterIds`, which is just extracted from `userWithId` which comes from the `useUser` hook, there doesn't seem to be any reason to be doing all of this processing at this 'out leve', or changing things here.

I would just be reverting all of these changes back to how they were.

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -12,13 +12,35 @@ import { Button } from "components/atoms/Button";
...
+const prepareForMyPosters = (usersPosters: MyPersonalizedPosters) => (
+  poster: WithId<PosterPageVenue>
+): WithId<PersonalizedPoster> => {
+  return {
+    ...poster,
+    isSaved: isTruthy(
+      poster.id && usersPosters[poster.id]?.includes(poster.id)
+    ),
+  };
+};
```
0xdevalias, line 31: This should be in a `utils` file of some sort probably (if it's even still needed?)

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -12,13 +12,35 @@ import { Button } from "components/atoms/Button";
...
   venue: WithId<GenericVenue>;
 }
 
+export const emptyPersonalizedPosters = {};
+
 export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
+  const { userWithId } = useUser();
+  const userPosterIds =
+    userWithId?.myPersonalizedPosters ?? emptyPersonalizedPosters;
+
```
0xdevalias, line 43: Instead of including this at this 'outer layer' and then drilling the data through to where it's needed, we can just use the hook directly at the 'deep layer' where we want it.

I would be removing these changes from here.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
+
+import { WithVenueId } from "utils/id";
+
 import "./PosterPreview.scss";
 
 export interface PosterPreviewProps {
-  posterVenue: WithId<PosterPageVenue>;
   enterVenue: (venueId: string) => void;
+  personalizedPoster: WithId<PersonalizedPoster>;
 }
```
0xdevalias, line 37: As mentioned in other comments, this should be reverted to how it was previously.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
 
+const emptyRelatedEvents: WithVenueId<VenueEvent>[] = [];
+export const emptyPersonalizedSchedule = {};
+
 export const PosterPreview: React.FC<PosterPreviewProps> = ({
-  posterVenue,
   enterVenue,
+  personalizedPoster,
 }) => {
```
0xdevalias, line 45: As mentioned in other comments, this should be reverted to how it was previously.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
-  const { title, authorName, categories } = posterVenue.poster ?? {};
```
0xdevalias, line 21: Why did we move this from here? This should be restored.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
 
   const posterClassnames = classNames("PosterPreview", {
-    "PosterPreview--live": posterVenue.isLive,
+    "PosterPreview--live": personalizedPoster.isLive,
   });
 
+  const { title, authorName, categories } = personalizedPoster.poster ?? {};
+
+  const venueId = personalizedPoster.id;
+
```
0xdevalias, line 63: This should all be put back where it was originally.

```suggestion
```

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
-  const { title, authorName, categories } = posterVenue.poster ?? {};
+  const [isBookmarkedPoster, setBookmarkPoster] = useState(
+    personalizedPoster.isSaved
+  );
+
+  useEffect(() => {
+    setBookmarkPoster(personalizedPoster.isSaved);
+  }, [personalizedPoster.isSaved]);
 
-  const venueId = posterVenue.id;
```
0xdevalias, line 23: Why did we move this from here? This should be restored.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
+  const { userId } = useUser();
 
   const posterClassnames = classNames("PosterPreview", {
-    "PosterPreview--live": posterVenue.isLive,
+    "PosterPreview--live": personalizedPoster.isLive,
   });
```
0xdevalias, line 58: As mentioned in other comments, this should be reverted to how it was previously.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -1,31 +1,66 @@
...
+  const [isBookmarkedPoster, setBookmarkPoster] = useState(
+    personalizedPoster.isSaved
+  );
+
+  useEffect(() => {
+    setBookmarkPoster(personalizedPoster.isSaved);
+  }, [personalizedPoster.isSaved]);
 
-  const venueId = posterVenue.id;
+  const { userId } = useUser();
```
0xdevalias, line 54: This is where the logic from `useUser` in `src/components/templates/PosterHall/PosterHall.tsx` can be brought back

```typescript
  const { userWithId } = useUser();
  const userPosterIds =
    userWithId?.myPersonalizedPosters ?? emptyPersonalizedPosters;
  // ..etc..
```

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+          event.id &&
+          updatePersonalizedSchedule({
+            event: event,
+            userId: userId,
+            removeMode: !personalizedPoster.isSaved,
+          });
+        return {};
+      });
+    return relatedVenueEvents;
+  }, [userId, isBookmarkedPoster, personalizedPoster, relatedVenueEvents]);
```
0xdevalias, line 112: Whitespace makes things like this much easier to read:

```suggestion
  const bookmarkPoster = useCallback(() => {
    setBookmarkPoster(!isBookmarkedPoster);
    
    personalizedPoster.isSaved = !personalizedPoster.isSaved;
    
    if (userId && personalizedPoster.id) {
      savePosterToProfile({
        venueId: personalizedPoster.id,
        userId: userId,
        removeMode: !personalizedPoster.isSaved,
      });
    }
    
    relatedVenueEvents
      .filter((event) => event.venueId === personalizedPoster.id)
      .map((event) => {
        userId &&
          event.id &&
          updatePersonalizedSchedule({
            event: event,
            userId: userId,
            removeMode: !personalizedPoster.isSaved,
          });
          
        return {};
      });
      
    return relatedVenueEvents;
  }, [userId, isBookmarkedPoster, personalizedPoster, relatedVenueEvents]);
```

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+
+  const onBookmarkPoster: MouseEventHandler<HTMLDivElement> = useCallback(
+    (e) => {
+      e.stopPropagation();
+      bookmarkPoster();
+    },
+    [bookmarkPoster]
+  );
```
0xdevalias, line 120: We don't want to use this 'passive' `onX` naming style for functions that 'do things'. They should be named for the action they do.

In this case `bookmarkPoster` would be the most obvious name, but it's already been used.

I can't see any reason why we need 2 separate `useCallback`'d functions here, so they can likely be merged into a single definition, called `bookmarkPoster`.

0xdevalias: Also, why do we need `e.stopPropagation();` here? This is something we've been trying to eliminate from our code recently.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -39,13 +74,78 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+  // const openAuthorProfile = useCallback(() => {
+  //   openUserProfileModal(author);
+  // }, [openUserProfileModal, author]);
+
   return (
     <div className={posterClassnames} onClick={handleEnterVenue}>
+      <div className="PosterPreview__bookmark" onClick={onBookmarkPoster}>
+        <FontAwesomeIcon
+          icon={isBookmarkedPoster ? solidBookmark : regularBookmark}
+        />
```
0xdevalias, line 136: Since we've used this 2 times now, and I can imagine it making sense to be used elsewhere in future as well, we should extract this into a tiny atom component, that takes a single `boolean` prop. Then use it both here and in `src/components/molecules/ScheduleEvent/ScheduleEvent.tsx` (and anywhere else that uses this sort of icon, if anywhere else exists) currently.

src/types/User.ts
```diff
@@ -16,6 +16,7 @@ export type VideoState = {
 
 // the structure is { [key: venueId] : eventId[] }
 export type MyPersonalizedSchedule = Partial<Record<string, string[]>>;
+export type MyPersonalizedPosters = Partial<Record<string, string[]>>;
```
0xdevalias, line 19:
```suggestion
export type SavedPosters = Partial<Record<string, string[]>>;
```

src/types/User.ts
```diff
@@ -37,6 +38,7 @@ export interface User {
   enteredVenueIds?: string[];
   status?: string;
   myPersonalizedSchedule?: MyPersonalizedSchedule;
+  myPersonalizedPosters?: MyPersonalizedPosters;
```
0xdevalias, line 41:
```suggestion
  savedPosters?: SavedPosters;
```

src/types/venues.ts
```diff
@@ -233,6 +233,10 @@ export interface PosterPageVenue extends BaseVenue {
   isLive?: boolean;
 }
 
+export interface PersonalizedPoster extends PosterPageVenue {
+  isSaved: boolean;
+}
+
```
0xdevalias, line 239: As mentioned in other comments, I believe this should also be reverted to how it was previously.

```suggestion
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -138,12 +138,18 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
 
     const locatedEvents: LocatedEvents[] = Object.entries(
-      groupBy(daysEvents, buildLocationString)
+      groupBy(
+        daysEvents.filter(
+          (event) => !event.venueId.match(REMOVE_EVENTS_FROM_VENUE)
+        ),
+        buildLocationString
+      )
     ).map(([group, events]) => ({
```
0xdevalias, line 153: What's the intent/need for this change here?

0xdevalias (changes requested):
[CRO] First pass review. A bunch of rough notes that will allow the diff here to be reduced dramatically, which will make it easier to review the actual intended functionality here.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: The aim here is to remove ~3900 posterpage-linked events from appearing in the venue rows of the schedule (unless they are 'saved/bookmarked' in the posterhall, in which case they will appear in the personalized schedule)

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: This was from back when I was also integrating the user avatar into the poster preview card. Will clean up here

margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: cleaned up — from prior idea to show user avatar on poster preview. now removed

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: cleaned up and removed

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: cleaned up and removed

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: cleaned up and removed

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: reverted. thanks

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
margulies: The bookmark on the poster preview cards needs to be positioned differently than the schedule event rectangle because of the layout differences. For the most part, the css is similar, but the additions are generally to address the positioning. One other addition is the `cursor: pointer`, because it helps to indicate the unique function associated with clicking the bookmark icon to the user

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: This may indeed be an OHBM-specific issue, but the issue this PR is trying to address is that each poster is associated with three events. This is far too much to present visually on the schedule, but we want users to be able to bookmark a poster and to have the three events associated with 'saved' poster appear in their personalized schedule.

Also, fixed the props next to component issue

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: reverted

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: done

src/types/User.ts
margulies: Done

src/types/User.ts
margulies: Done

src/types/venues.ts
margulies: Done

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: @mike-lvov This doesn't seem like a generalisable solution, so we might end up needing to merge this directly to `env/ohbm` instead of our main code branch. Thoughts?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
0xdevalias: That makes sense. Sounds good.

Does the bookmark not have this cursor on the schedule? If not, would be good to add it there as well.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: _nods_ that makes sense. I expect it may end up being very OHBM-specific. But let's leave retargeting the branch till we know that for certain.

src/settings.ts
```diff
@@ -594,3 +594,5 @@ export const SCHEDULE_SHOW_DAYS_AHEAD = 7;
  * @see https://firebase.google.com/docs/firestore/query-data/queries#in_not-in_and_array-contains-any
  */
 export const FIRESTORE_QUERY_IN_ARRAY_MAX_ITEMS = 10;
+
+export const REMOVE_EVENTS_FROM_VENUE = /^poster[0-9]{4}$/; // or "INCLUDE_ALL"
```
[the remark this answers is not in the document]
margulies: I've now removed that — an empty string would be sufficient to not filter any venues if need be.

margulies (commented):
accidentally started a review and can't figure out how to cancel it...

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: I've extracted this now to an atom component

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
```diff
@@ -79,17 +76,18 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
...
-        <div className="ScheduleEvent__info">
-          <div className="ScheduleEvent__title">{event.name}</div>
-          <div className="ScheduleEvent__host">by {event.host}</div>
-        </div>
-
-        <div className="ScheduleEvent__bookmark" onClick={bookmarkEvent}>
-          <FontAwesomeIcon
-            icon={event.isSaved ? solidBookmark : regularBookmark}
-          />
+      <div className={containerClasses}>
```
margulies, line 79: This was changed to prevent the event modal from popping up whenever a bookmark was clicked. This appears to be a bug in the current implementation, which is resolved by splitting the class from the onclick activity

sunny-viktoryia (changes requested):
here is the first pass review from me.
I feel that we should change the approach of filtering `PosterPage` venues in the schedule. Using regexp does not seems right to me.

open to your thoughts, @mike-lvov @0xdevalias

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -151,7 +155,13 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
 
     const locatedEvents: LocationEvents[] = Object.entries(
-      groupBy(daysEvents, buildLocationString)
+      groupBy(
+        daysEvents.filter(
+          (event: { venueId: string }) =>
+            !event.venueId.match(REMOVE_EVENTS_FROM_VENUE)
+        ),
+        buildLocationString
+      )
```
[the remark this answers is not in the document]
margulies: We can definitely rework this, but I think hiding all the events from a specific posterpage template might cause other problems — for example, there are other posterhalls than the main one, which have a limited number of events that should be shown. 
One alternative approach might be to add a parameter to the `event` type to `hideEvent?` so that if set to true, it doesn't show in the main schedule, but only in the personal schedule track. Would that be preferable?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -9,14 +9,18 @@ import { enterVenue } from "utils/url";
 
 import { PosterCategory } from "components/atoms/PosterCategory";
 
+import { PosterBookmark } from "../PosterBookmark";
+
 import "./PosterPreview.scss";
 
 export interface PosterPreviewProps {
   posterVenue: WithId<PosterPageVenue>;
+  showBookmarks?: boolean;
```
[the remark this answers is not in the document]
margulies: The idea of this property is to give the admin the option to turn the bookmark feature on or off for a given venue. I could change to `canBeBookmarked` if that would be preferable.

src/components/templates/PosterHall/components/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { events: relatedVenueEvents } = useVenueEvents({
+    venueIds: relatedVenueIds,
+  });
+
+  const { userWithId } = useUser();
+  const userPosterIds = userWithId?.savedPosters ?? emptySavedPosters;
+
+  const [isBookmarkedPoster, setBookmarkPoster] = useState(
+    //@ts-ignore
+    userPosterIds[venueId]?.[0] === venueId
```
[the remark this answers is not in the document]
margulies: I wasn't happy with that either, so any tips are welcome. In the user profile, an array is created where the first item is the name of the poster venue. The `[0]` is simply grabbing the first item, which is always the poster venue name. 

I tried reworking the data structure in several different ways, but continually ran into trouble, so this fix seemed to address it. Any other suggestions for how to set up the structure would be welcome.

src/components/templates/PosterHall/components/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,71 @@
...
+  );
```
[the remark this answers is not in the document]
margulies: Thankt! I'll implement that

margulies:
I managed to get the poster bookmarking (`savedPosters`) aspect fixed, but saving the personalizedSchedule events is still causing some problems

src/components/templates/PosterHall/components/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,71 @@
...
+      .filter((event) => event.venueId === venueId)
+      .map((event) => {
+        userWithId?.id &&
+          event.id &&
+          updatePersonalizedSchedule({
+            event: event,
+            userId: userWithId?.id,
+            removeMode: isBookmarkedPoster,
+          });
+        return {};
```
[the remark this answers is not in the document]
margulies: implemented

margulies:
> I managed to get the poster bookmarking (`savedPosters`) aspect fixed, but saving the personalizedSchedule events is still causing some problems

This now appears to be fixed. The issue seems to have arisen from calling `useVenueEvents`, which also appears to have been noted as an unresolved issue in the code: https://github.com/sparkletown/sparkle/blob/b14a03de0a02764a50e493216ce893ea812bf279/src/hooks/events.ts#L44

I've created an addition hook which only takes a single venue as input (https://github.com/sparkletown/sparkle/pull/1375/commits/37cbbee58bc643dbba4ebfeda08dae23b3df4e78), and that appears to have resolved the issue. I'm sure there's a better way to address the original problem, but didn't want to mess too much with the schedule-related code.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
margulies: Regarding the change in `src/components/molecules/ScheduleEvent/ScheduleEvent.tsx`, I can revert that, but worth noting that if an event is very short, currently the title isn't displayed at all. The result of the change implemented here is that the title exceeds the length of the event box, but is visible, which I think has an advantage because it's possible to see the event title even for short events. I'll revert these changes, but wanted to mention that issue.

src/api/profile.ts
```diff
@@ -137,3 +137,27 @@ export const updatePersonalizedSchedule = async ({
...
+export interface SavePosterToProfileProps {
+  venueId: string;
+  userId: string;
+  removeMode?: boolean;
+}
+
+export function savePosterToProfile({
+  venueId,
+  userId,
+  removeMode = false,
```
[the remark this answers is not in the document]
margulies: I've refactored as requested here. Thanks!

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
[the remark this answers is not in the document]
margulies: Now changed to `canBeBookmarked`

src/components/templates/PosterHall/components/PosterBookmark/PosterBookmark.tsx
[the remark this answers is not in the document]
margulies: This is now resolved with 2bb20a06bf3bf6f56b96c40553c99864a6ee7af8
`const savedPosters = useMemo(() => userWithId?.savedPosters ?? {}, [
    userWithId,
  ]);

  const isSaved = useMemo(
    () => savedPosters[venueId]?.includes(venueId) ?? false,
    [savedPosters, venueId]
  );`

src/api/profile.ts
```diff
@@ -131,3 +131,36 @@ export const updatePersonalizedSchedule = async ({
...
+    Bugsnag.notify(err, (poster) => {
+      poster.addMetadata("context", {
+        location: "api/profile::savePosterToProfile",
+        venueId,
+        userId,
+        removeMode,
+      });
+    });
+  });
+}
```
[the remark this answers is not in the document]
margulies: implement in newest version

src/components/templates/PosterHall/PosterHall.tsx
margulies: no longer relevent

src/components/templates/PosterHall/PosterHall.tsx
margulies: removed in current version

src/components/templates/PosterHall/PosterHall.tsx
margulies: thanks! reverted

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
margulies: done

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: removed during refactoring

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: currently pulled into `PosterBookmark.tsx`

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: All this is now pulled into it's own `PosterBookmark.tsx` file

src/settings.ts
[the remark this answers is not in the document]
margulies: This is the currently unresolved, and any suggestion regarding how to properly address this would be helpful. Filtering based on template type won't work for the use case here, but I think this string search is sufficiently specific 'poster[0-9][0-9][0-9][0-9]' to not impact other venues that are not posters. This is a very OHBM specific feature though, so happy to merge directly to env/ohbm if that would be preferable.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: Any tips on how to resolve this would be helpful. As noted on the navbar schedule issue, I can't seem to find a workaround that doesn't call both the bookmark and underlying poster link when clicked without the `e.stopPropagation()`

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -20,47 +20,42 @@ import { faBookmark as regularBookmark } from "@fortawesome/free-regular-svg-ico
...
-  useEffect(() => {
-    setBookmarkPoster(personalizedPoster.isSaved);
-  }, [personalizedPoster.isSaved]);
+  const { userWithId } = useUser();
+  const userPosterIds = userWithId?.savedPosters ?? emptyPersonalizedPosters;
 
-  const { userId } = useUser();
+  const [isBookmarkedPoster, setBookmarkPoster] = useState(
+    //@ts-ignore
+    userPosterIds[posterVenue.id]?.[0] === posterVenue.id
```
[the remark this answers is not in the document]
margulies: resolved

0xdevalias:
> The issue seems to have arisen from calling `useVenueEvents`, which also appears to have been noted as an unresolved issue in the code:

@margulies that will only cause it to be re-run when that variable changes. It will always change if you pass in an in-line array, that’s what the comment is about. If you only have a single venueId, then memo’ing it so that the array reference isn’t recreated on each render is the simplest way to solve/work around it. Duplicating the entire hook like you did isn’t a good solution. 

There is also a utils helper (maybe in utils/types?) that makes it easy to use a single string or an array of strings as the prop of a function, which could be used to improve the original hook in this case of only needing a single venueId to be passed through (but wouldn’t change the ‘inline array reference changes on every render aspect)

margulies:
> @margulies that will only cause it to be re-run when that variable changes. It will always change if you pass in an in-line array, that’s what the comment is about. If you only have a single venueId, then memo’ing it so that the array reference isn’t recreated on each render is the simplest way to solve/work around it. Duplicating the entire hook like you did isn’t a good solution.

Thanks for pointing this out — I've set the `venueIds` array in a `useMemo` in the following commit: 2a5a6cdf01ae67e47840deba5db4526247d51ccd

margulies:
> This should presumably just be:
> 
> ```ts
>  const venueIds = useMemo(() => [venueId], [venueId]);
> ```

Thanks! Fixed in b1fb3794b0e4bb05e42da7d29a3a28f2918f70ac

src/api/profile.ts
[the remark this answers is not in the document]
sunny-viktoryia: @margulies thank you for your update.
But next time, could you attach the commit URL where the suggestions from the comment were fixed? It will make the reviewer's life a bit easier during the next round of the review.

refactored in https://github.com/sparkletown/sparkle/pull/1375/commits/2ff03724dd3b34e1b988a449566fed0dfc929c05

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -28,7 +31,11 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
   const { push: openUrlUsingRouter } = useHistory();
   const handleEnterVenue = useCallback(
-    () => enterVenue(venueId, { customOpenRelativeUrl: openUrlUsingRouter }),
+    (e) => {
+      if (e.target.closest("PosterPreview__bookmark")) return;
```
sunny-viktoryia, line 35: 1. `closest` takes selector as a parameter, so we should indicate that this is a class name adding `.` in the beginning (`".PosterPreview__bookmark"`)
2. would be better to make this string `"PosterPreview__bookmark"` a const as it's used 2 times in the file: to be able to update it when needed at once.

src/components/molecules/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,69 @@
...
+    () => savedPosters[venueId]?.includes(venueId) ?? false,
+    [savedPosters, venueId]
+  );
+
+  const removeBookmarks = useCallback(() => {
+    if (!userId) return;
+    removePosterFromProfile({ venueId, userId });
+    venueEvents.forEach((event) => {
+      removeEventFromPersonalizedSchedule({ event, userId });
+    });
```
sunny-viktoryia, line 49: I think we can refactor `removeEventFromPersonalizedSchedule` and `addEventToPersonalizedSchedule` to make them take not `event`, but `events array` or `event IDs` as a parameter. As both these methods use `updateUserCollection` which accepts the array.
Thus, we'll reduce the number of database update requests.

src/components/molecules/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }, [venueId, userId, venueEvents]);
+
+  const addBookmarks = useCallback(() => {
+    if (!userId) return;
+    addPosterToProfile({ venueId, userId });
+    venueEvents.forEach((event) => {
+      addEventToPersonalizedSchedule({ event, userId });
+    });
```
sunny-viktoryia, line 57: same here:
>  think we can refactor `removeEventFromPersonalizedSchedule` and `addEventToPersonalizedSchedule` to make them take not `event`, but `events array` or `event IDs` as a parameter. As both these methods use `updateUserCollection` which accepts the array.
Thus, we'll reduce the number of database update requests.

sunny-viktoryia (changes requested):
I guess we still need to understand if this is an ohbm specific feature or it should go to our main codebase. Based on the answer we can understand what is the better approach in poster bookmarking / event hiding.

Personally, I don't like the idea of creating a regexp for that. I thought, that it would be better to hide all the poster venues, but as @margulies noted in his comment https://github.com/sparkletown/sparkle/pull/1375#discussion_r649946286 there are posters that should not be hidden. So how can we distinguish them? Looks like we need to expand the PosterPage interface to include a property which will indicate if this venue should be hidden from the schedule. Or this is a more generic case and it could be useful for any venue? Let's say for a secret venue that can appear at some moment. 

Another thing: event hiding. @margulies suggested to add a property to the `VenueEvent` type: `hideEvent` (or `isHidden`, for example). I believe it can be useful as well. In future, we could turn on/off events and to show them only when needed (if it's a secret event, for example). Not sure if it's something we need though.

@0xdevalias @mike-lvov would be happy to discuss the details of the feature with you. And what way we want to develop this PR?

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
```diff
@@ -84,12 +81,11 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
           <div className="ScheduleEvent__title">{event.name}</div>
```
margulies, line 81: @sunny-viktoryia this is the section I was referring to (where the prior reordering is now reverted). Part of the issue is that if events are very short, the title doesn't display, and is only visible from the event modal. Not a huge problem, but something that might be worth finding a fix for in the future

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
margulies: For sure -- added here: https://github.com/sparkletown/sparkle/pull/1375/files#r651908164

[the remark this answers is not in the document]
margulies: > > One alternative approach might be to add a parameter to the `event` type to `hideEvent`? so that if set to true, it doesn't show in the main schedule, but only in the personal schedule track.
> 
> The approach of marking events is more preferable for me. But why do we need to hide the events from the main schedule if they can appear in the personal schedule? I believe it'd confusing for the user if [s]he sees the events in the personal schedule but does not see the line of the venue/room where it's located.
> 
> one more note about this part of the code.
> 
> We filter all the posters to hide them from the schedule. But what if the poster is bookmarked? In your implementation, we see only its events in the "My Personal schedule line".
> But I believe we need to show the venue line in the schedule as well.
> 
> image
> 

That's a good point — just a note of the issue to consider, which is that there around ~1400 posterpage venues, each of which has 3 events. I've imported those events into the ohbm platform so you can see the issue. The way the schedule is sorting, there are venues at the bottom of the list, after the 1400 posterpage venues, which might easily get lost. I know this is a very specific problem, but this is the one that the posterbookmarking was designed to address.

src/components/molecules/PosterBookmark/PosterBookmark.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { events: venueEvents } = useVenueEvents({ venueIds: venueIds });
+
+  const savedPosters = useMemo(() => userWithId?.savedPosters ?? {}, [
+    userWithId,
+  ]);
+
+  const isSaved = useMemo(
+    () => savedPosters[venueId]?.includes(venueId) ?? false,
+    [savedPosters, venueId]
+  );
```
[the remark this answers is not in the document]
margulies: Thanks!

src/components/atoms/Bookmark/Bookmark.scss
```diff
@@ -0,0 +1,6 @@
+@import "scss/constants.scss";
+
+.Bookmark {
```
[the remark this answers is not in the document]
margulies: Good catch — thanks!

src/components/atoms/Bookmark/Bookmark.tsx
```diff
@@ -0,0 +1,25 @@
...
+  isSaved: boolean;
+}
+
+export const Bookmark: React.FC<BookmarkProps> = ({
+  className = "Bookmark",
+  onClick,
+  isSaved,
+}) => {
+  return (
+    <div className={className} onClick={onClick}>
```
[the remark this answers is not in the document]
margulies: Fixed in f8af7dfd9a3c9af06465643ed3f21805293674a5

src/components/atoms/Bookmark/Bookmark.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { MouseEventHandler } from "react";
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faBookmark as solidBookmark } from "@fortawesome/free-solid-svg-icons";
+import { faBookmark as regularBookmark } from "@fortawesome/free-regular-svg-icons";
+
+import "./Bookmark.scss";
+
+export interface BookmarkProps {
+  className?: string;
```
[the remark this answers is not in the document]
margulies: fixed in f8af7dfd9a3c9af06465643ed3f21805293674a5

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
margulies: I've addressed point 1. here f8f823a7af6a74681675e1a122557c4fae0f4277
Regarding making the string a constant, I ran into some problems there. I'll need to play around with it a bit more, but if you have any tips for dealing with the preceed `.` that differs between the two places, please let me know.

sunny-viktoryia: 2. you can find the example in the `hideEventSchedule` function of the `NavBar` component.
We have `navBarScheduleClassName` const there and used it this way in `closest()` :  \`.${navBarScheduleClassName} `

margulies: Thanks! Implemented here: 7ed6c2ad1c81adfeaa09fde2708bb4c779d093f9

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
0xdevalias: I noticed that for all of these posters, they are being displayed with presumably their `venueId` rather than a name/human readable label? Is that by purpose/design? Why are we doing it that way?

image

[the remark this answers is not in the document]
margulies: Thanks — I hadn't noticed that. The `venue.name` (and `venue/event/room`) is actually set to the `venue.id`, but that doesn't look nice at all. Thanks for the fresh pair of eyes — I'll update the `venue.name` and `room` in events accordingly.

mike-lvov (approved):
LGTM

PR #1376, opened 2021-05-11
fix grabbing events from venues instead of rooms

margulies:
There is still an issue with the isEventLive in ScheduleEvents throwing an error (commented out in order to run here), but this is my quick and dirty fix to switch the new schedule from only using rooms to using venues. Thought there might be something useful in it for fixing up the new schedule UI...


margulies:
This PR was intended merely to share some potential fixes to the new schedule PR. Can be closed now

PR #1379, opened 2021-05-11
increase profile name max length to 40

margulies:
Not sure if this is already taken care of (in which case, please close this PR), but this changes the max alloted char count for user names from 16 to ~30~ 40. If a different number would be preferred, just let me know and I'll make the change.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/670

### DIFF
--- a/src/components/organisms/EditProfileModal/EditProfileModal.tsx
+++ b/src/components/organisms/EditProfileModal/EditProfileModal.tsx
@@ -74,7 +74,7 @@ const EditProfileModal: React.FunctionComponent<PropsType> = ({
               placeholder="Your display name"
               ref={register({
                 required: true,
-                maxLength: 16,
+                maxLength: 30,
               })}
             />
             {errors.partyName && errors.partyName.type === "required" && (
@@ -82,7 +82,7 @@ const EditProfileModal: React.FunctionComponent<PropsType> = ({
             )}
             {errors.partyName && errors.partyName.type === "maxLength" && (
               <span className="input-error">
-                Display name must be 16 characters or less
+                Display name must be 30 characters or less
               </span>
             )}
             {user && (
--- a/src/components/organisms/EditProfileModal/EditProfileModal.tsx
+++ b/src/components/organisms/EditProfileModal/EditProfileModal.tsx
@@ -7,6 +7,7 @@ import { QuestionsFormData } from "pages/Account/Questions";
 import { updateUserProfile } from "pages/Account/helpers";
 import { QuestionType } from "types/Question";
 import ProfilePictureInput from "components/molecules/ProfilePictureInput";
+import { DISPLAY_NAME_MAX_CHAR_COUNT } from "settings";
 import { useUser } from "hooks/useUser";
 import { useSelector } from "hooks/useSelector";
 import { currentVenueSelectorData } from "utils/selectors";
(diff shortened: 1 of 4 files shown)


0xdevalias:
@sparkletown/product @mike-lvov Do you remember where this was discussed/whether there is an existing backlog issues/slack thread about it? In particular, do you know what length we want here?

src/components/organisms/EditProfileModal/EditProfileModal.tsx
```diff
@@ -74,15 +74,15 @@ const EditProfileModal: React.FunctionComponent<PropsType> = ({
               placeholder="Your display name"
               ref={register({
                 required: true,
-                maxLength: 16,
+                maxLength: 30,
```
0xdevalias, line 77: Instead of duplicating this number across the various locations in the code, it would be good to extract it into a `const` or 2 in `src/settings.ts`, then reuse it here.

src/pages/Account/Profile.tsx
```diff
@@ -71,19 +71,19 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
             </span>
             {errors.partyName && errors.partyName.type === "required" && (
               <span className="input-error">Display name is required</span>
             )}
             {errors.partyName && errors.partyName.type === "maxLength" && (
               <span className="input-error">
-                Display name must be 16 characters or less
+                Display name must be 30 characters or less
               </span>
             )}
```
0xdevalias, line 88: @sparkletown/product @sparkletown/design @mike-lvov Should we align the text here so that they both say "Display name" rather than one of them being "public name"?

0xdevalias (changes requested):
Some minor tweaks.

---

@sparkletown/product @mike-lvov Do you remember where this was discussed/whether there is an existing backlog issues/slack thread about it? In particular, do you know what length we want here?

0xdevalias:
> **Edit:** Found it: [sparkletown/internal-sparkle-issues#670](https://github.com/sparkletown/internal-sparkle-issues/issues/670)
> 
> > Extend "User Display Name" length from 16 characters to 40 characters

@margulies Can you please change this to 40 characters long.

src/pages/Account/Profile.tsx
mike-lvov: Yes, `Display name` works. + this should be imported from `settings`

mike-lvov:
@margulies It looks like the branch is created in the fork version of sparkletown and is compared against our staging. Could you please create a branch off of our staging?

margulies:
That makes sense — hadn't noticed that before, but will do my best to get that sorted.

src/pages/Account/Profile.tsx
0xdevalias: > + this should be imported from settings

https://github.com/sparkletown/sparkle/pull/1379#discussion_r630702721 ;)

0xdevalias:
> That makes sense — hadn't noticed that before, but will do my best to get that sorted.

@mike-lvov @margulies See the explanation for this in https://github.com/sparkletown/sparkle/pull/1379#pullrequestreview-657435264

To fix it, rebase the changes on your PR branch onto staging, then force push.

src/pages/Account/Profile.tsx
margulies: > Yes, `Display name` works for me. + this should be imported from `settings`

Fixed

0xdevalias (approved):
Looks good, LGTM!

PR #1381, opened 2021-05-11
remove mapIconImageUrl (aka: 'jam icon') from admin v1 + the codebase

oleksii-front:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/710


src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -145,14 +145,16 @@ export const AdminVenuePreview: React.FC<AdminVenuePreviewProps> = ({
...
+              </div>
+              <div className="content">
+                <img className="icon" src={venue.mapIconImageUrl} alt="icon" />
+              </div>
             </div>
-            <div className="content">
-              <img className="icon" src={venue.mapIconImageUrl} alt="icon" />
-            </div>
-          </div>
+          )}
```
0xdevalias, line 157: We don't want to just render this icon when it's present, we want to remove it from the code entirely. So this block of code should probably be deleted.

0xdevalias: > As part of removing this, we should also be removing the concept of it from our core types/codebase.
> 
> If I remember correctly, I believe @denisdimitrov may have already done some work on this/potentially had a WIP branch for it?
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/710#issuecomment-839195504_

---

It also appears to be used in the following locations:

- https://github.com/sparkletown/sparkle/search?q=mapIconImageUrl

For each of these locations, we will need to assess how it is being used in the code there, and remove any redundant/no longer required code. It might be worth syncing with @mike-lvov about this to double check/confirm what you intend to remove once you look through the locations and formulate your plan.

---

Of particular note, removing it from our `BaseVenue` type is likely going to highlight all of the other places in the app where it exists but shouldn't:

- https://github.com/sparkletown/sparkle/blob/staging/src/types/venues.ts#L124

0xdevalias (changes requested):
Changes/comments.

src/pages/Admin/AdminVenuePreview.tsx
mike-lvov: > It also appears to be used in the following locations:
> 
> https://github.com/sparkletown/sparkle/search?q=mapIconImageUrl
> For each of these locations, we will need to assess how it is being used in the code there, and remove any redundant/no longer required code. It might be worth syncing with @mike-lvov about this to double check/confirm what you intend to remove once you look through the locations and formulate your plan.
> 
> Of particular note, removing it from our BaseVenue type is likely going to highlight all of the other places in the app where it exists but shouldn't:
> 
> https://github.com/sparkletown/sparkle/blob/staging/src/types/venues.ts#L124

@oleksii-front I don't think you addressed all of the above?

0xdevalias (changes requested):
See outstanding changes from previous review in https://github.com/sparkletown/sparkle/pull/1381#discussion_r630700349 + https://github.com/sparkletown/sparkle/pull/1381#discussion_r633486130

src/pages/Admin/AdminVenuePreview.tsx
0xdevalias: @oleksii-front Once again, please follow our standards and post the commit hash + a message in review comment threads communicating when/where they are fixed, and mark it as resolved. You have been asked to do this by both @mike-lvov and myself multiple times now, and we shouldn't have to keep reminding you/asking you to do this. It make things much harder and slower for the reviewer when you don't do this.

Please don't make us need to ask you to do this again in future.

src/components/templates/Playa/Playa.tsx
```diff
@@ -644,7 +644,7 @@ const Playa = () => {
                 <span className="img-vcenter-helper" />
                 <img
                   className="venue-icon"
-                  src={v.mapIconImageUrl || DEFAULT_MAP_ICON_URL}
+                  src={DEFAULT_MAP_ICON_URL}
                   alt={`${v.name} Icon`}
                 />
```
0xdevalias, line 649: Instead of just removing `v.mapIconImageUrl` here, I think we should either be changing it to one of the remaining non-legacy venue images, or removing the entire `img` tag here.

Since Playa is legacy code, perhaps we can just comment out this code instead. In doing that, it's probably worth adding back the `v.mapIconImageUrl ||` before commenting out the whole `img` tag, and then adding a comment explaining why it's commented out.

---

Is `DEFAULT_MAP_ICON_URL` also related to this same concept? Based on the usage here, it seems so. If so, we should probably be removing it from the codebase in this PR as well.

src/components/templates/Playa/PlayaVenue.tsx
```diff
@@ -39,7 +39,7 @@ const PlayaVenue = ({
         <span className="img-vcenter-helper" />
         <img
           className="venue-icon"
-          src={venue.mapIconImageUrl || DEFAULT_MAP_ICON_URL}
+          src={DEFAULT_MAP_ICON_URL}
           alt={`${venue.name} Icon`}
         />
```
0xdevalias, line 44: Instead of just removing `v.mapIconImageUrl` here, I think we should either be changing it to one of the remaining non-legacy venue images, or removing the entire `img` tag here.

Since Playa is legacy code, perhaps we can just comment out this code instead. In doing that, it's probably worth adding back the `v.mapIconImageUrl ||` before commenting out the whole `img` tag, and then adding a comment explaining why it's commented out.

---

Is `DEFAULT_MAP_ICON_URL` also related to this same concept? Based on the usage here, it seems so. If so, we should probably be removing it from the codebase in this PR as well.

src/pages/Admin/Admin.tsx
```diff
@@ -297,14 +298,14 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
...
                               icon: {
                                 width: PLAYA_VENUE_SIZE,
                                 height: PLAYA_VENUE_SIZE,
                                 top: venue.placement.y,
                                 left: venue.placement.x,
-                                url: venue.mapIconImageUrl,
+                                url: DEFAULT_MAP_ICON_URL,
                               },
                             }
                           : {}
```
0xdevalias, line 311: Is `DEFAULT_MAP_ICON_URL` also related to this same concept? Based on the usage here, it seems so. If so, we should probably be removing it from the codebase in this PR as well.

Which means we'll have to change the code here differently.

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -219,25 +210,18 @@ export const editVenueCastSchema = Yup.object()
...
-
-  // possible locations for the map icon
-  .from("config.mapIconImageUrl", "mapIconImageUrl")
-  .from("mapIconImageUrl", "mapIconImageUrl");
+  .from("config.landingPageConfig.bannerImageUrl", "bannerImageUrl");
 
 export const editPlacementCastSchema = Yup.object()
   .shape<Partial<PlacementInput>>({})
 
   // possible locations for the map icon
```
0xdevalias, line 218: This comment seems irrelevant now the associated lines of code are removed?
```suggestion
```

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -219,25 +210,18 @@ export const editVenueCastSchema = Yup.object()
...
-  .from("config.mapIconImageUrl", "mapIconImageUrl")
-  .from("mapIconImageUrl", "mapIconImageUrl")
   .from("placement.addressText", "addressText")
   .from("placement.notes", "notes")
   .required();
 
 export const editPlacementSchema = Yup.object().shape<PlacementInput>({
   mapIconImageFile: createFileSchema("mapIconImageFile", false).notRequired(),
```
0xdevalias, line 224: Given we've removed `mapIconImageUrl`, would we not also be wanting to remove `mapIconImageFile`? Since presumably they are going to relate to the same sort of thing?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -155,8 +155,7 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
   const mapIconUrl = useMemo(() => {
     const file = values.mapIconImageFile;
     if (file && file.length > 0) return URL.createObjectURL(file[0]);
-    return values.mapIconImageUrl;
-  }, [values.mapIconImageFile, values.mapIconImageUrl]);
+  }, [values.mapIconImageFile]);
```
0xdevalias, line 158: I expect we are probably going to want to be assessing this whole piece of code, and potentially removing it entirely; not just removing small bits of it.

0xdevalias: And all of the related bits of code/usages that will then flow on from that.

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -214,26 +214,21 @@ export const editVenueCastSchema = Yup.object()
   .from("config.landingPageConfig.bannerImageUrl", "bannerImageUrl")
 
   // possible locations for the map icon
```
0xdevalias, line 216: This comment seems irrelevant since you've removed the related code.
```suggestion
```

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -214,26 +214,21 @@ export const editVenueCastSchema = Yup.object()
...
   .from("auditoriumColumns", "auditoriumColumns")
   .from("auditoriumRows", "auditoriumRows")
-  .from("mapIconImageUrl", "mapIconImageUrl")
   .from("code_of_conduct_questions", "code_of_conduct_questions")
   .from("profile_questions", "profile_questions");
 
 export const editPlacementCastSchema = Yup.object()
   .shape<Partial<PlacementInput>>({})
 
   // possible locations for the map icon
```
0xdevalias, line 225: This comment seems irrelevant since you've removed the related code.
```suggestion
```

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -214,26 +214,21 @@ export const editVenueCastSchema = Yup.object()
...
-  .from("config.mapIconImageUrl", "mapIconImageUrl")
-  .from("mapIconImageUrl", "mapIconImageUrl")
   .from("placement.addressText", "addressText")
   .from("placement.notes", "notes")
   .required();
 
 export const editPlacementSchema = Yup.object().shape<PlacementInput>({
   mapIconImageFile: createFileSchema("mapIconImageFile", false).notRequired(),
```
0xdevalias, line 231: Given we've removed `mapIconImageUrl`, would we not also be wanting to remove `mapIconImageFile`? Since presumably they are going to relate to the same sort of thing?

src/utils/venue.ts
```diff
@@ -62,10 +62,6 @@ export const createJazzbar = (values: FormValues): JazzbarVenue => {
   return {
     template: VenueTemplate.jazzbar,
     name: values.name || "Your Jazz Bar",
-    mapIconImageUrl: urlFromImage(
-      "/default-profile-pic.png",
-      values.mapIconImageFile
-    ),
```
0xdevalias, line 68: We've removed this code from here, but it looks like `mapIconImageFile` is still going to exist in `FormValues` (and likely related code), which we presumably should also be cleaning up here given it seems to be so tightly related?

src/api/admin.ts
```diff
@@ -139,7 +138,6 @@ type FirestoreRoomInput_v2 = Omit<RoomInput_v2, RoomImageFileKeys> &
 
 export type PlacementInput = {
   mapIconImageFile?: FileList;
```
0xdevalias, line 140: Given we've removed `mapIconImageUrl`, would we not also be wanting to remove `mapIconImageFile`? Since presumably they are going to relate to the same sort of thing?

0xdevalias (changes requested):
Looks like there's still related things to be cleaned up here.

We need to look at the code we are removing, and ensure we remove not just the specific key we've deleted, but any code that then becomes redundant/useless based on that; repeating until we have cleaned up everything related to the feature we are removing.

(cc // @mike-lvov FYI)

src/components/templates/Playa/Playa.tsx
oleksii-front: Should we remove it here as well? https://github.com/sparkletown/sparkle/blob/staging/src/pages/Account/Venue/VenueMapEdition/Container.tsx#L238
https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Playa/PlayaVenue.tsx#L42

src/components/templates/Playa/PlayaVenue.tsx
oleksii-front: @0xdevalias @mike-lvov 
Should we remove it here as well? https://github.com/sparkletown/sparkle/blob/staging/src/pages/Account/Venue/VenueMapEdition/Container.tsx#L238
https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Playa/PlayaVenue.tsx#L42

alexey-petrov:
Closing in favour of https://github.com/sparkletown/sparkle/pull/1873

PR #1382, opened 2021-05-11
remove "Add Events" button from right panel in admin v1

oleksii-front:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/712


src/pages/Admin/Admin.tsx
```diff
@@ -386,16 +384,6 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
...
-            <button
-              className="btn btn-primary"
-              onClick={() => {
-                history.push(`${match.url}/events`);
-                setShowCreateEventModal(true);
-              }}
-              style={{ marginBottom: 10, width: "100%" }}
-            >
-              Create an Event
-            </button>
```
0xdevalias, line 398: > As part of removing this right hand side button, we should also assess what elements it relies on, and whether they have any purpose anymore once it’s removed. If not, they should all be removed as well.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/712#issuecomment-839193139_

0xdevalias: Is the `showCreateEventModal` state still used/needed once we remove this button? Will `AdminEventModal` still be opened here by other areas of the code?

It looks like `setShowCreateEventModal`/`showCreateEventModal` are passed into `VenueInfoComponent` from `VenueDetails`.

After removing this button in `VenueInfoComponent`, it seems the only place `showCreateEventModal`/`setShowCreateEventModal` is used is as props to the `AdminEventModal` (`src/pages/Admin/AdminEventModal.tsx`) component. Double check this, but most if not all of the modals in our app are somewhat 'independent' of the structure of where they are rendered (because they 'float on top' of everything else).

Based on that, we can likely move the `AdminEventModal` 'up' out of `VenueInfoComponent` and render it in `VenueDetails` instead. That will also let us remove the now redundant `setShowCreateEventModal`/`showCreateEventModal` props from `VenueInfoComponent`.

It seems like `EventsComponent` (rendered by `VenueDetails`) also appears to include/render `AdminEventModal` within it's JSX. Since `EventsComponent` is only rendered by `VenueDetails` (which after the refactor above will render `AdminEventModal` for us), and there is no reason to have the code for a modal in the JSX more than once (so long as it is rendered once, it will always show when it's controls are used), then we can likely also remove it from `EventsComponent` (along with the props it will make redundant when we do so, if any)

0xdevalias: > As a creator/admin, I have 3 places to add events, and one is infinitely more useful than the other two.
> 
> _Originally posted by @sofisparkle in https://github.com/sparkletown/internal-sparkle-issues/issues/712_

Looking at the code for the 'Add an Event' button that is described as being the more 'infinitely useful' one, it seems to exist in `src/pages/Admin/AdminVenueRoomDetails.tsx`

- https://github.com/sparkletown/sparkle/blob/staging/src/pages/Admin/AdminVenueRoomDetails.tsx#L167-L176

It uses `setShowCreateEventModal` to show this modal, which is passed in as a prop.

`AdminVenueRoomDetails` is rendered in `AdminVenueRoomsList`, which defines the `useState` (`showCreateEventModal`/`setShowCreateEventModal`)

`showCreateEventModal` is then used to show `AdminEventModal`.

---

So at least the 'bad' and the 'good' versions are rendering the same modal (`AdminEventModal`) and we don't have excess code to remove there.

0xdevalias (dismissed):
Comments/changes.

src/pages/Admin/Admin.tsx
dmitrymarchuk: Removed the button from "Events" tab. Also cleaned up some unused code in the AdminV1-related components. Decided not to do heavy refactoring as Admin V3 is on the way.
In https://github.com/sparkletown/sparkle/pull/1382/commits/9c09b3ba5b3f0dbbc30602d12580a97ecab0af8e

src/pages/Admin/Admin.tsx
```diff
@@ -364,16 +348,17 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
...
+
+            {canBeDeleted(venue) && (
+              <button
+                role="link"
+                className="btn btn-block btn-danger"
+                onClick={showDeleteModal}
+              >
+                Delete space
+              </button>
+            )}
```
mike-lvov, line 360: Where is coming from? I though we already had the delete button in place

mike-lvov (approved):
LGTM, but don't merge

src/pages/Admin/Admin.tsx
dmitrymarchuk: This change comes from merge-commit. I don't know why is it highlighted here.

mike-lvov: Can you confirm that there are no `Delete space` button?

dmitrymarchuk: I don't know if it's a merge error or I accidentally did something wrong but after merging recent changes from staging there really were two "Delete space" buttons. And unused code I remove became used again:) So I think somebody was working on that "Delete space" button and I interfered. Anyway, fixed in https://github.com/sparkletown/sparkle/pull/1382/commits/f8ad957794093233f6547c039f4063c27482d813

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -1,19 +1,14 @@
...
+import "./UserAvatar.scss";
 import classNames from "classnames";
+import { useVenueId } from "hooks/useVenueId";
+import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
+import { useWorldUserLocation } from "hooks/users";
 import { isEqual, isEmpty } from "lodash";
-
+import React, { useMemo } from "react";
 import { DEFAULT_PARTY_NAME, DEFAULT_PROFILE_IMAGE } from "settings";
-
```
mike-lvov, line 6: The imports are broken

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -1,11 +1,14 @@
...
+import { useRelatedVenues } from "hooks/useRelatedVenues";
+import { useRoom } from "hooks/useRoom";
 import React, { useMemo, useState } from "react";
 import Modal from "react-bootstrap/Modal";
-
 import { EVENT_STATUS_REFRESH_MS } from "settings";
-
 import { Room } from "types/rooms";
 import { VenueEvent } from "types/venues";
-
```
mike-lvov, line 8: same everywhere

mike-lvov (changes requested):

mike-lvov (approved):

PR #1384, opened 2021-05-12
User Status: As a user at an event, I can set my status to busy

oleksii-front:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/720


0xdevalias:
@oleksii-front Instead of adding `[WIP]` in the PR title/etc, you can just use GitHub's 'draft pull request' feature, which tells us that it is in-progress and not ready for review yet:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/
  - https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests
  
(cc // @mike-lvov FYI: To improve your onboarding process, it would be good to teach new engineers about this from the beginning)

src/api/profile.ts
```diff
@@ -48,3 +48,40 @@ export const makeUpdateUserGridLocation = ({
       firestore.doc(doc).set(newData);
     });
 };
+
+export const makeUpdateUserOnlineStatus = ({
+  venueId,
+  userUid,
+}: MakeUpdateUserGridLocationProps) => (status: string) => {
+  const doc = `users/${userUid}`;
```
0xdevalias, line 56: Instead of using this 'path' format, we should use `firestore.collection('users').doc(userUid)` below.

src/api/profile.ts
```diff
@@ -48,3 +48,17 @@ export const makeUpdateUserGridLocation = ({
...
+}
+
+export const makeUpdateUserOnlineStatus = ({
+  userUid,
+}: makeUpdateUserOnlineStatusProps) => (status: string) => {
+  const newData = {
+    "data.status": status,
+  };
+
+  firebase.firestore().collection("users").doc(userUid).update(newData);
```
mike-lvov, line 63: Add bugsnag to an error catcher + return firease call.

As per https://github.com/sparkletown/sparkle/blob/feature/new-auditorium/src/api/profile.ts

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -38,7 +38,7 @@ export const ChatMessage: React.FC<ChatProps> = ({
     <div className={containerStyles}>
       <div className="chat-message__text">{text}</div>
       <div className="chat-message__info" onClick={openAuthorProfile}>
-        <UserAvatar user={author} />
+        <UserAvatar user={author} isShowStatus />
```
mike-lvov, line 41:
```suggestion
        <UserAvatar user={author} showStatus />
```

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -20,6 +21,22 @@
...
+      background-color: $ripple-color;
+    }
+
+    &_offline {
+      background-color: $red;
+    }
+
+    &_custom {
+      background-color: $concrete;
+    }
```
mike-lvov, line 39:
```suggestion
    &--online {
      background-color: $green;
    }

    &--busy {
      background-color: $ripple-color;
    }

    &--offline {
      background-color: $red;
    }

    &--custom {
      background-color: $concrete;
    }
```

src/api/profile.ts
```diff
@@ -48,3 +48,17 @@ export const makeUpdateUserGridLocation = ({
...
+
+export interface makeUpdateUserOnlineStatusProps {
+  userUid: string;
+}
+
+export const makeUpdateUserOnlineStatus = ({
+  userUid,
+}: makeUpdateUserOnlineStatusProps) => (status: string) => {
+  const newData = {
+    "data.status": status,
```
mike-lvov, line 60: I think we should store it on the top level.
```suggestion
    "status": status,
```

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -23,7 +23,7 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
   containerClassName,
   imageClassName,
   onClick,
-  isOnline,
```
mike-lvov, line 26: `isOnline` is defined by us. We check if the user has been online lately. We should keep it

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +37,12 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
     "user-avatar--clickable": onClick !== undefined,
   });
 
+  const status = user?.data?.status || "";
+
   const imageClasses = classNames("user-avatar__image", imageClassName);
+  const statusClasses = classNames("user-avatar__status-dot", {
+    [`user-avatar__status-dot_${status}`]: status,
```
mike-lvov, line 44:
```suggestion
    [`user-avatar__status-dot_${status}`]: isOnline && status,
```

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +37,12 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
     "user-avatar--clickable": onClick !== undefined,
   });
 
+  const status = user?.data?.status || "";
```
mike-lvov, line 40: Any reason we want to have "" as a default? Why not just having it undefined?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -293,14 +300,15 @@ const NavBar: React.FC<NavBarPropsType> = ({
...
-                  <div className="navbar-link-profile">
+                  <div className="navbar-link-profile navbar-profile-container">
                     <img
                       src={profileImage}
                       className="profile-icon"
                       alt="avatar"
                       width="40"
                       height="40"
                     />
+                    {profileStatus && <span className={statusClasses} />}
```
mike-lvov, line 311: We should reuse our UserAvatar here to avoid code duplication

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -68,23 +71,49 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
...
   if (!user || !userWithId) return null;
 
+  const handlerChangeStatus = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = e.target;
+    makeUpdateUserOnlineStatus({
+      userUid: userWithId?.id,
+    })(value);
+  };
+
+  const statuses = ["online", "busy", "offline", "custom"];
```
mike-lvov, line 81: This should be implemented as an enum

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -68,23 +71,49 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
 
   if (!user || !userWithId) return null;
 
+  const handlerChangeStatus = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = e.target;
+    makeUpdateUserOnlineStatus({
+      userUid: userWithId?.id,
+    })(value);
+  };
```
mike-lvov, line 79: useCallback. Any reason to destructure e.target here?

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -68,23 +71,49 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
...
-          height="50"
-        />
+        <div className="profile-avatar profile-avatar--clickable">
+          <img
+            className="profile-icon profile-avatar__image"
+            src={profile?.pictureUrl || DEFAULT_PROFILE_IMAGE}
+            alt="profile avatar"
+          />
+          {profileStatus && <span className={statusDotClasses} />}
+        </div>
```
mike-lvov, line 100: Reuse UserAvatar

mike-lvov (changes requested):
Ideas/suggestions/comments

src/api/profile.ts
```diff
@@ -48,3 +48,17 @@ export const makeUpdateUserGridLocation = ({
...
     });
 };
+
+export interface makeUpdateUserOnlineStatusProps {
+  userUid: string;
+}
+
+export const makeUpdateUserOnlineStatus = ({
+  userUid,
+}: makeUpdateUserOnlineStatusProps) => (status: string) => {
```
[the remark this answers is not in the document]
0xdevalias: If it’s a ‘make’ function it’s curried so that the first parameter can be bound when it’s passed to something like useCallback/similar. Is that what this is being used for? If not, the ‘make’ prefix and currying may not make sense

src/components/atoms/UserAvatar/UserAvatar.tsx
0xdevalias: (Also we default with ?? not ||)

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: And not within this component here, but in a types/* location or similar.

src/components/atoms/UserAvatar/UserAvatar.tsx
oleksii-front: I think that status shouldn't be `undefined` in any case. Just my opinion.

0xdevalias: @oleksii-front Why do you say it shouldn't be `undefined` if we have no other value for it? TypeScript makes use of `undefined` all over the place, it's a better value than an arbitrary 'empty string' or other placeholder value as we can easily explicitly check against it and react accordingly.

All of the `?.` 'safe accessor' stuff is neatly wrapping potentially `undefined` cases. We use it a lot in our codebase.

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

src/api/profile.ts
```diff
@@ -48,3 +48,28 @@ export const makeUpdateUserGridLocation = ({
       firestore.doc(doc).set(newData);
     });
 };
+
+export interface UpdateUserOnlineStatusProps {
+  status: string;
```
mike-lvov, line 53: It should have Status enum as a type

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -20,6 +20,19 @@
...
+
+    &--online {
+      background-color: $green;
+    }
+    &--busy {
+      background-color: $ripple-color;
+    }
+    &--offline {
+      background-color: $red;
+    }
```
mike-lvov, line 32: The logic of a user being `online` status is defined in our `hooks/users` `useRecentWorldUsers` hook. If we try to set it to `status`, we will have 2 sources of truth, which is incorrect. That's why, this state should not be stored there. 
Offline status is the default status of the user and that's why we should't expect it to explicitly come from `status` profile field

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -10,10 +11,11 @@ import { WithId } from "utils/id";
 import "./UserAvatar.scss";
 
 export interface UserAvatarProps {
-  user?: WithId<User>;
+  user?: WithId<User> | FirebaseReducer.Profile<User>;
```
mike-lvov, line 14: Why do we need to add `FirebaseReducer.Profile<User>`?

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -10,10 +11,11 @@ import { WithId } from "utils/id";
...
   containerClassName?: string;
   imageClassName?: string;
-  isOnline?: boolean;
+  statusClassName?: string;
```
mike-lvov, line 17: Any usecases we would want to modify statusClassName?

src/components/atoms/UserAvatar/UserAvatar.tsx
mike-lvov: I explained this above. Please, don't silently unresolve the conversations

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +40,12 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
     "user-avatar--clickable": onClick !== undefined,
   });
 
+  const status = user?.status ?? "";
+
   const imageClasses = classNames("user-avatar__image", imageClassName);
+  const statusClasses = classNames("user-avatar__status-dot", statusClassName, {
+    [`user-avatar__status-dot--${status}`]: status,
+  });
```
mike-lvov, line 48: This logic should be refactored to use `isOnline` and `status`. Eg. show status only if isOnline(which makes sense)

src/components/atoms/UserStatusDropdown/UserStatusDropdown.scss
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  &__item {
+    list-style: none;
+    margin-bottom: $spacing--xs;
+    padding: 0 $spacing--sm;
+    cursor: pointer;
+    &:hover {
+      background-color: rgba($white, 0.1);
```
mike-lvov, line 41: use opaque-white css fn

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import { Experience, Statuses } from "types/User";
+
+import "./UserStatusDropdown.scss";
+
+export interface UserStatusDropdownProps {
+  options: Array<Statuses>;
+  onChange: (option: string) => void;
+  label: string | Experience;
```
mike-lvov, line 9: Why does it accept `Experience`?

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
+  options,
+  label,
+  onChange,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
```
mike-lvov, line 17: Can we reuse `useShowHide` hook? Based on the logic, we can.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+        className="dropdown-container__item"
+        onClick={onOptionClicked(option)}
+        key={option}
+      >
+        {option}
+      </li>
+    ));
+
+  return (
+    <div className="dropdown-container">
```
mike-lvov, line 41:
```suggestion
    <div className="Dropdown">
```

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -28,6 +28,16 @@ $border-radius: 28px;
...
+  .navbar-profile-image-container {
+    width: 40px;
+    height: 40px;
+  }
+
+  .navbar-profile-image-status {
+    width: 12px;
+    height: 12px;
+  }
+
```
mike-lvov, line 40: any reason we have these styles? since we reuse UserAvatar everywhere

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -293,15 +293,12 @@ const NavBar: React.FC<NavBarPropsType> = ({
                   overlay={ProfilePopover}
                   rootClose={true}
                 >
-                  <div className="navbar-link-profile">
-                    <img
-                      src={profileImage}
-                      className="profile-icon"
```
mike-lvov, line 299: Have you removed `profile-icon` styles?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -293,15 +293,12 @@ const NavBar: React.FC<NavBarPropsType> = ({
...
-                      alt="avatar"
-                      width="40"
-                      height="40"
-                    />
-                  </div>
+                  <UserAvatar
+                    user={profile}
+                    containerClassName="navbar-profile-image-container"
+                    statusClassName="navbar-profile-image-status"
```
mike-lvov, line 299: I guess it just has different size? In this case we should define the `size` property in UserAvatar and handle it inside

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -66,25 +71,43 @@ const UserInformationContent: React.FunctionComponent<PropsType> = ({
...
+        });
+      }
+    },
+    [userWithId]
+  );
+
   if (!user || !userWithId) return null;
 
+  const profileStatus = profile?.status ?? "";
+
```
mike-lvov, line 89: move this into `useProfileStatus` hook. It will return all the handlers + current status

src/types/User.ts
```diff
@@ -32,4 +32,12 @@ export interface User {
...
   enteredVenueIds?: string[];
+  status?: string;
+}
+
+export enum Statuses {
+  online = "online",
+  busy = "busy",
+  offline = "offline",
+  custom = "custom",
 }
```
mike-lvov, line 43:
```suggestion
export enum UserStatus {
  online = "online",
  busy = "busy",
  offline = "offline",
  custom = "custom",
}
```

mike-lvov (changes requested):
Please, don't silently resolve the comment and answer all of them. You don't necessarily have to make all the fixes, but if you disagree, explain your position.

src/components/atoms/UserAvatar/UserAvatar.scss
0xdevalias: > The logic of a user being online status is defined in our hooks/users useRecentWorldUsers hook. If we try to set it to status, we will have 2 sources of truth, which is incorrect. That's why, this state should not be stored there.

@mike-lvov This seems unrelated to the styles changes here?

src/components/atoms/UserAvatar/UserAvatar.tsx
0xdevalias: It's the 'raw' type of `profile` that comes from `useUser`. We don't want to be adding it here like this. `WithId<User>` is the same thing, but better.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
0xdevalias: Plus make sure to 'rename' the exported data from `useShowHide` so it has 'contextual/semantically correct names' for its purpose.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
0xdevalias: By the name of this file, should it not be `UserStatusDropdown`?

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: What's the different size? I almost added a `size` prop to `UserAvatar` in a previous PR, but then didn't bother. I'd be open to us adding it.

The way I implemented it was with a `AvatarSize` `enum`, with values like `default`, `small`, etc as needed. I then made 'size --variants' in the scss, and applies these with `classNames` based on the value of the `size` prop enum values.

That way we can consistently configure the couple of different avatar sizes we should need across the platform, which IMO will make it easier to keep consistent rather than allowing any arbitrary size params to be passed in.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: @mike-lvov instead of `useCallback`, would this be better using `useAsyncFn`?

oleksii-front: @0xdevalias no, I think that hook works better for that case

src/components/molecules/NavBar/NavBar.tsx
oleksii-front: @0xdevalias prop 'large' added. It is enough for this component

0xdevalias: @oleksii-front Can you please add the commit hash to these comment threads of where you make the changes to make it easy for us to review.

src/components/atoms/UserAvatar/UserAvatar.scss
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/files#diff-edfb3f24596e6fe740f3de347baec9272f305ca2717d22bca5ac4fc5e2eb3b04R37

src/components/molecules/NavBar/NavBar.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/76498f6cf6998b09e94d345c54d36a4a9d199f96

0xdevalias (changes requested):
@oleksii-front When a PR is ready for review, please mark it as 'Ready for review' so that it communicates that to everyone else:

image

As well as re-requesting review from anyone who has already reviewed it after changes are made:

image

When fixing comments, please ensure you follow our process of commenting in each review thread with the commit hash it was fixed in so that reviewers know it was done, where it was fixed, and can quickly assess that fix as part of their re-review: https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1620395110209200

There also seem to be a bunch of conflicts that need fixing here:

image

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +42,26 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
     "user-avatar--clickable": onClick !== undefined,
   });
 
+  const isOnline = recentWorldUsers.find(
+    (worldUser) => worldUser.id === user?.id
+  );
```
mike-lvov, line 47: I don't think it's the most efficient way to decide if a person is `online`, but I don't have any other suggestions. At least this should be wrapper into useMemo. @0xdevalias wdyt?

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +42,26 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
...
+
+  const status = user?.status ?? "";
```
mike-lvov, line 49:
```suggestion
  const status = user?.status;
```

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -37,7 +42,26 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
...
+  });
+  const statusOfflineClasses = classNames(
+    "user-avatar__status-dot",
+    "user-avatar__status-dot--offline",
+    { "user-avatar__status-dot--large": large }
+  );
+
+  const statusComponent = isOnline
+    ? showStatus && status && <span className={statusClasses} />
+    : showStatus && <span className={statusOfflineClasses} />;
```
mike-lvov, line 64: This must be rewritten into one classNames object.

smth like:

```
classNames("user-avatar__status-dot", {
"user-avatar__status-dot--offline": !isOnline,
"user-avatar__status-dot--online": isOnline,
[`user-avatar__status-dot--${status}`]: isOnline && status,
"user-avatar__status-dot--large": large,  });
```

And not display dot if it's offline. Using css

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { UserStatus } from "types/User";
+import { useShowHide } from "hooks/useShowHide";
```
mike-lvov, line 3:
```suggestion
import React from "react";

import { UserStatus } from "types/User";

import { useShowHide } from "hooks/useShowHide";
```

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import "./UserStatusDropdown.scss";
+
+export interface UserStatusDropdownProps {
+  options: Array<UserStatus>;
```
mike-lvov, line 8:
```suggestion
  options: UserStatus[];
```

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onChange: (option: UserStatus) => void;
+  label: UserStatus | string;
```
mike-lvov, line 10: If it's just the label of Dropdown, should it be just string? Or am I missing something?

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  options,
+  label,
+  onChange,
+}) => {
+  const { isShown, hide, toggle } = useShowHide();
+
+  const onOptionClicked = (value: UserStatus) => () => {
+    onChange(value);
+    hide();
+  };
```
mike-lvov, line 23: useCallback + when you useMemo `optionComponents`, you can remove one layer of function and make it not factory () => () => void, but a simple function () => void

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+    .filter((opt) => opt !== label)
+    .map((option) => (
+      <li
+        className="Dropdown__item"
+        onClick={onOptionClicked(option)}
+        key={option}
+      >
+        {option}
+      </li>
+    ));
```
mike-lvov, line 35: useMemo

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onChange: (option: UserStatus) => void;
+  label: UserStatus | string;
+}
+
+export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
+  options,
+  label,
+  onChange,
+}) => {
+  const { isShown, hide, toggle } = useShowHide();
```
mike-lvov, line 18: assign understandable var names.

```suggestion
  const { isShown: isDropdownOptionsShown, hideDropdownOptions, toggle: toggleDropdownOptions } = useShowHide();
```

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.scss
```diff
@@ -0,0 +1,7 @@
+@import "scss/constants";
+
+.profile-avatar-container {
+  width: 50px;
+  height: 50px;
+  margin-right: 10px;
+}
```
mike-lvov, line 7: since we have `--large`, can we get rid of this?

src/hooks/useProfileStatus.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback } from "react";
+
+import { useUser } from "hooks/useUser";
+import { UserStatus } from "types/User";
+import { updateUserOnlineStatus } from "api/profile";
+
+export const useProfileStatus = () => {
+  const { userWithId, profile } = useUser();
```
mike-lvov, line 8: you can just import UserId from `useUser`

src/settings.ts
```diff
@@ -580,6 +581,13 @@ export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 12;
 
+export const USER_STATUSES = [
+  UserStatus.online,
+  UserStatus.busy,
+  UserStatus.offline,
+  UserStatus.custom,
```
mike-lvov, line 588: remove online/offline from this list, please as a person can't set it himself

src/types/User.ts
```diff
@@ -35,5 +35,13 @@ export interface User {
...
   enteredVenueIds?: string[];
+  status?: string;
   myPersonalizedSchedule?: MyPersonalizedSchedule;
 }
+
+export enum UserStatus {
+  online = "online",
+  busy = "busy",
+  offline = "offline",
+  custom = "custom",
```
mike-lvov, line 46: As per the comment above, remove offline and online

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -28,6 +28,11 @@ $border-radius: 28px;
     cursor: pointer;
   }
 
+  .navbar-profile-image-container {
+    width: 40px;
+    height: 40px;
+  }
```
mike-lvov, line 34: since we have `--large`, can we remove these values?

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {label || "change status"}
+      </div>
+      {isShown && (
+        <div className="Dropdown__list-container">
+          <ul className="Dropdown__list">{optionsComponents}</ul>
+        </div>
+      )}
+    </div>
+  );
+};
```
mike-lvov, line 49: I think we should also implement `default` status. A person should be able to `unset` the `busy/custom` status

mike-lvov (changes requested):
Comments/ideas

src/components/molecules/NavBar/NavBar.tsx
oleksii-front: used in `Participant.tsx` and `JazzBar.tsx`

src/components/atoms/UserAvatar/UserAvatar.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/dfe92ea3574c1de767c22b418ec2a106109a6390

src/components/atoms/UserAvatar/UserAvatar.tsx
oleksii-front: Thanks!

oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/119d07836ecd2582c9bd23a81d188b3e93688825

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/2534af4e00cb14c31c12a45ed214f0d543c84981

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/e5c91ff3474edb5d81fe16691b876c9c77238321

src/components/molecules/NavBar/NavBar.scss
oleksii-front: I changed 50x50 to 40x40 to implement it as you suggested

oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/5bcc570daf621979becff60da99f34b765b09c03

src/hooks/useProfileStatus.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/eb9038d3849d6a10733530fa9fa7cc100e6e6f95

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.scss
oleksii-front: shrieked to 40px

oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/5bcc570daf621979becff60da99f34b765b09c03

src/settings.ts
oleksii-front: set status to online is the way to change status from `busy` to normal

oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/9da797f41a65b4b551961457ff493326a6918da5

src/types/User.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/8c87d4d1376bd2174c883f9f92d1bfcf9a2ac85e

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
oleksii-front: Online is default

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
oleksii-front: There is only limited values could be here

mike-lvov (commented):
A few additions

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -13,8 +14,10 @@ export interface UserAvatarProps {
   user?: WithId<User>;
   containerClassName?: string;
   imageClassName?: string;
-  isOnline?: boolean;
+  showStatus?: boolean;
   onClick?: () => void;
+  large?: boolean;
+  statusLarge?: boolean;
```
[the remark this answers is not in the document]
oleksii-front: merged into one 
https://github.com/sparkletown/sparkle/pull/1384/commits/899e746c105f01f4061f2b1bd9828e2147a5e6eb

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -35,10 +42,25 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
...
+  const isOnline = useMemo(
+    () => recentWorldUsers.find((worldUser) => worldUser.id === user?.id),
+    [user, recentWorldUsers]
+  );
+
+  const status = user?.status;
+
   const imageClasses = classNames("user-avatar__image", imageClassName);
 
+  const statusIndicatorClassName = classNames("user-avatar__status-indicator", {
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/22c1d304b32d327e18fa40fcdb85e3507b79ec14

src/components/molecules/NavBar/NavBar.scss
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/3284feef1b32a7a8b24cf3529fa8da4cf81878f5

src/components/atoms/UserAvatar/UserAvatar.tsx
oleksii-front: boolean can't represent all available statuses

src/settings.ts
```diff
@@ -582,6 +583,12 @@ export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 12;
 
+export const USER_STATUSES = [
+  UserStatus.online,
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1384/commits/479b9c42fd77b2590cc81d122fd609984ac17634

src/components/atoms/UserAvatar/UserAvatar.tsx
0xdevalias: @mike-lvov `recentWorldUsers` would seem to align with other similar patterns that we have used elsewhere I believe?

Looking at where `UserAvatar` is used in existing code:

- `src/components/organisms/ChatSidebar/components/PrivateChatPreview/PrivateChatPreview.tsx`
  - passes in `isOnline` from props
- `src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx`
  - uses `onlineUsers.some((user) => user.id === chatMessage.counterPartyUser.id)`
  - which is set from `const onlineUsers = useOnlineUsersToDisplay();`
- `src/hooks/privateChats.ts`
  - which uses `useRecentWorldUsers`

mike-lvov (approved):
Made some logic + styles changes and merging it in

PR #1386, opened 2021-05-13
[env/ohbm] set venuePage HTML document title basing off PLATFORM_BRAND_NAME

yarikoptic:
This greatly improves UX (User eXperience) by not only making page/tab
show the location, but also making browser history useful.  Without this
change you get the exciting "Sparkle" as the title (hardcoded in index.html)
for every page.

Joint work with @margulies where I was "git grep"ing and he was editing and trying it out.

1st commit is generic and could be applied (cherrypicked) to the main code base.  2nd commit is OHBM vendoring. Let me know if there is another/better way to vendor it.

PR #1389, opened 2021-05-14
add PLATFORM_BRAND_NAME + set page title in VenuePage based on current venue name

yarikoptic:
This greatly improves UX (User eXperience) by not only making page/tab
show the location, but also making browser history useful.  Without this
change you get the exciting "Sparkle" as the title (hardcoded in index.html)
for every page.

Joint work with @margulies where I was "git grep"ing and he was editing and trying it out.

Originally was sent (with vendoring) as #1386 against `env/ohbm`. This PR contains only the first commit from that PR (cherrypicked), without ohbm vendoring.

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -1,5 +1,6 @@
-import React, { useState, useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { Redirect, useHistory } from "react-router-dom";
+import { useTitle } from "react-use";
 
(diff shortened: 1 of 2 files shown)


0xdevalias (approved):
Code changes look good here. LGTM!

---

@mike-lvov @denisdimitrov FYI Seems our preview environment doesn't deploy currently for PR's from external contributors.

0xdevalias:
Thanks for the contribution @yarikoptic! You wont see this end up in `env/ohbm` straight away, but the next time we deploy our main code branch through to there, you'll see this functionality show up.

PR #1383, opened 2021-05-12
add RenderMarkdown component + render markdown in areas across the platform

margulies:
This is an initial commit with the react-markdown package used to parse the banner text. If this works, my plan would be to integrate for parsing of event and venue descriptions as well.

- https://github.com/remarkjs/react-markdown

---

- partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/473
- fixes https://github.com/sparkletown/sparkle/issues/1393
- fixes https://github.com/sparkletown/sparkle/issues/1395

PR #1394, opened 2021-05-14
add support for emojis in chat using react-emoji-render

### LINKED ISSUE #1393: chat: render emojis
emojis already used for reactions etc.  But they are absent where IMHO most needed -- in the chat.  I wonder if it might be easy to implement using e.g. https://www.npmjs.com/package/react-emoji-render to render chat messages?  I think it would be ok to not have some fancy completions etc to start with and just rely on people to know to type `:smile:` or `:)`

---

**Edit by @0xdevalias:** See also https://github.com/sparkletown/sparkle/issues/1395 as a tightly related issue/feature.

yarikoptic:
Closes #1393 

This PR introduces base support, no completions/shortcuts etc for emoji's - just rendering.

https://www.npmjs.com/package/react-emoji-render says license MIT but didn't check licenses of the libraries it uses.

<details>
<summary>
example on ohbm instance:
</summary> 

image

</details>

not sure if this would be an optimal solution if decided to proceed with emoji-picker (#1395) but may be they could be made to play together nicely.

### DIFF
--- a/src/components/atoms/ChatMessage/ChatMessage.tsx
+++ b/src/components/atoms/ChatMessage/ChatMessage.tsx
@@ -12,6 +12,8 @@ import { useShowHide } from "hooks/useShowHide";
 import { ChatMessageInfo } from "components/atoms/ChatMessageInfo";
 import { TextButton } from "components/atoms/TextButton";
 
+import Emoji from "react-emoji-render";
+
 import "./ChatMessage.scss";
 
 export interface ChatProps {
@@ -42,7 +44,7 @@ export const ChatMessage: React.FC<ChatProps> = ({
     () =>
       replies?.map((reply) => (
         <div key={reply.id} className="ChatMessage__reply">
-          {reply.text}
+          <Emoji text={reply.text} />
           <ChatMessageInfo
             message={reply}
             deleteMessage={() => deleteMessage(reply.id)}
@@ -68,8 +70,9 @@ export const ChatMessage: React.FC<ChatProps> = ({
     <div className={containerStyles}>
       <div className="ChatMessage__bulb">
         <div className="ChatMessage__text-content">
-          <div className="ChatMessage__text">{text}</div>
-
+          <div className="ChatMessage__text">
+            <Emoji text={text} />
+          </div>
           <div className="ChatMessage__reply-icon">
             <FontAwesomeIcon
               icon={faReply}
--- a/package-lock.json
+++ b/package-lock.json
@@ -18963,8 +18963,7 @@
     "lodash.isstring": {
       "version": "4.0.1",
       "resolved": "https://registry.npmjs.org/lodash.isstring/-/lodash.isstring-4.0.1.tgz",
-      "integrity": "sha1-1SfftUVuynzJu5XV2ur4i6VKVFE=",
-      "dev": true
+      "integrity": "sha1-1SfftUVuynzJu5XV2ur4i6VKVFE="
     },
     "lodash.keys": {
       "version": "2.4.1",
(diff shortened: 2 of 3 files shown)


0xdevalias:
> not sure if this would be an optimal solution if decided to proceed with emoji-picker (#1395) but may be they could be made to play together nicely.

@yarikoptic See my questions/discussion about this in https://github.com/sparkletown/sparkle/issues/1393#issuecomment-841595445 (let's figure out the path forward on that issue, then decide what to do for this PR)

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -36,7 +38,9 @@ export const ChatMessage: React.FC<ChatProps> = ({
 
   return (
     <div className={containerStyles}>
-      <div className="chat-message__text">{text}</div>
+      <div className="chat-message__text">
+        <Emoji text={text} />
+      </div>
```
0xdevalias, line 43: As per https://github.com/sparkletown/sparkle/pull/1394#issuecomment-841597607, pending discussion in https://github.com/sparkletown/sparkle/issues/1393#issuecomment-841595445 to see if we want to just use `emoji-mart`, which has its own `Emoji` component for rendering emoji:

- https://github.com/missive/emoji-mart#emoji

0xdevalias (changes requested):
Officially marking this as 'changes requested' for now, till we decide the path we want to take forward on this.

All in all though, very cool addition, and keen to get it merged once we decide how we want to go about it! I'm pleasantly surprised at just how simple/little code needs to be changed for this!

0xdevalias:
[OOS] While it's OOS for this PR, in future we could likely use this same lib/pattern to enhance our existing 'reactions' feature as well. Currently it just uses hardcoded emoji:

- https://github.com/sparkletown/sparkle/blob/staging/src/types/reactions.ts#L46-L103

Some of the related files:

- https://github.com/sparkletown/sparkle/search?q=reactions
  - https://github.com/sparkletown/sparkle/blob/staging/src/types/reactions.ts
  - https://github.com/sparkletown/sparkle/blob/staging/src/utils/reactions.ts
  - https://github.com/sparkletown/sparkle/blob/staging/src/hooks/reactions.tsx
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/UserReactions/UserReactions.tsx
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/ReactionPage/ReactionList.tsx#L26
  - etc

We also have a backlog issue (scheduled to be worked on soon, after a couple of higher priority features) to clean up all of the 'send reaction' buttons/etc across the codebase into a single canonical component + logic in a hook, etc (currently each venue template has it's own duplicated code for rendering the 'reaction buttons' in places like the Jazz Bar, Auditorium (`Audience`), etc:

- 🔒 https://github.com/sparkletown/internal-sparkle-issues/issues/637

yarikoptic:
updated PR for current state of `staging` with support for multi-threaded chat.  And without URL's rendering support  ran into the case for which I believe I did test (manually) explicitly before but somehow `javascript://` now falls into the same trap as jitsi chat did:

image

but it happens only for `javascript://` and e.g. `short://` but not `http://` - I guess there is some ad-hoc exclusion of common protocols within -render.

yarikoptic:
overall I think this particular PR is superseded by #1383, so if that one is merged, this could be closed

yarikoptic:
I will just close this one -- it should be superseded by #1383 (markdown). We know where to find it if #1383 doesn't land

0xdevalias:
> I will just close this one -- it should be superseded by #1383 (markdown).

@yarikoptic Yup, agreed that #1383 seems like it will be our best overall path forward here.

Thanks for the initial exploration/version here though!

0xdevalias:
Duplicate of #1383

PR #1396, opened 2021-05-14
add ability to share a PosterPage on social media

oleksii-front:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/725
closes https://github.com/sparkletown/internal-sparkle-issues/issues/726
closes https://github.com/sparkletown/internal-sparkle-issues/issues/727


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -116,8 +113,6 @@ const VenuePage: React.FC = () => {
     setLocationData({ userId, locationName: venueName });
   }, [userId, venueName]);
 
-  useTitle(`${PLATFORM_BRAND_NAME} - ${venueName}`);
```
yarikoptic, line 119: No no no!

yarikoptic (changes requested):
This PR reverts #1389 , probably some rebase went wrong

src/pages/VenuePage/VenuePage.tsx
0xdevalias, line 119: Specifically, https://github.com/sparkletown/sparkle/pull/1396#pullrequestreview-660288190 is referring to the removal of this code and similar that was added in https://github.com/sparkletown/sparkle/pull/1389

src/components/atoms/LinkButton/LinkButton.scss
```diff
@@ -0,0 +1,14 @@
+@import "scss/constants.scss";
+
+.button-link {
+  text-align: center;
+  text-decoration: none;
+  margin-top: 20px;
+  background-color: $lighter-intermediate-grey;
+  padding: 12px 16px;
+  border-radius: 28px;
```
mike-lvov, line 9: margins, paddings, border-radiuses should be imported from `constants`

src/components/atoms/LinkButton/LinkButton.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { FC } from "react";
+import "./LinkButton.scss";
+
+interface LinkButtonProps {
+  href: string;
+}
```
mike-lvov, line 6: Do you want to extend from the default HTML tag? An example can be found in `InputField`/`Checkbox`

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -97,7 +103,11 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
           />
 
           {/* TODO: Implement poster sharing on social media */}
```
mike-lvov, line 105:
```suggestion

```

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
```diff
@@ -0,0 +1,82 @@
+@import "scss/constants.scss";
```
mike-lvov, line 1: As before, most of the px values can be and should be imported from `constants`. If you can't find the exact value, either use the closest one, if the difference is small or create a new constant

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
```diff
@@ -0,0 +1,82 @@
...
+  &__title {
+    position: absolute;
+    font-size: 20px;
+    bottom: 20px;
+    left: 0;
+  }
+
+  &__url-text {
+    position: absolute;
+    color: $concrete;
```
mike-lvov, line 78: you can use `secondary-lighest` for this, when you merge the latest staging

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  onHide: () => void;
+  venue: WithId<PosterPageVenue>;
+}
+
+export const ShareModal: FC<ConfirmationModalProps> = ({
+  show,
+  onHide,
+  venue,
+}) => {
+  const url = `${process.env.REACT_APP_CODE_CHECK_URL}in/${venue.id}`;
```
mike-lvov, line 25: should be moved to utils? @0xdevalias wdyt, where and how to name it. Also, I'm not sure this is the best way to generate the link to the current place, but I don't have the capacity to dive into the options

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const handlerCopied = useCallback(() => {
+    setIsCopied(true);
+  }, []);
+
+  const linkText = isCopied ? "Link Copied" : "Copy Link";
+
+  const linkClasses = classNames("share-modal-container__link", {
+    "share-modal-container__link--copied": isCopied,
+  });
```
mike-lvov, line 36: `useShowHide`? and better naming eg. showCopiedText? Not sure about this tbh @0xdevalias

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const textTitle = `Check out this OHBM Poster, ${venue.poster?.title} by ${venue.poster?.authorName} at ${url}`;
+  const facebookHref = `https://www.facebook.com/sharer/sharer.php?app_id=${process.env.REACT_APP_FACEBOOK_APP_ID}&u=${url}&quote=${textTitle}&thumbnail=${venue.host?.icon}&picture=${venue.host?.icon}&hashtag=#OHBM2021`;
+  const twitterHref = `https://twitter.com/intent/tweet?text=${textTitle}&hashtags=OHBM2021`;
+
+  return (
+    <Modal show={show} onHide={onHide}>
+      <Modal.Body>
+        <div className="share-modal-header">
```
mike-lvov, line 45: ShareModal__header and other classnames

mike-lvov (changes requested):

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: Why is `REACT_APP_CODE_CHECK_URL` getting used here? Pretty sure that relates to a legacy/no longer used external microservice.

What's the high level aim/intent here? To get the full URL of the current page/venue/similar? If so, look at `utils/url`. There may not be an exact function there currently, but there should be the basic pattern to use, eg. `new URL(`/in/${venue.id}`, window.location.origin)`

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: Following our BEM naming standards:

- http://getbem.com/naming/

src/components/atoms/LinkButton/LinkButton.scss
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/042a93353ab9ea1d6ea606ef5b7090d8499ff5f7

src/components/atoms/LinkButton/LinkButton.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/ccd0d1ee39f47f9b38680343ec34f721df49e4e5

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/8c232a5d1ae872885bf6d3d3e41c8104f633279b

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: I probably would personally. Removes the boilerplate of the `useState` + `useCallback`.

[OOS] `useShowHide` probably just needs a better name in general to make it more obvious in these sorts of cases.

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
AlexeyPogorelov: https://github.com/sparkletown/sparkle/pull/1396/commits/99a584cdd3729c9db9813f7e7b941d9c6bcba1e4

src/pages/VenuePage/VenuePage.tsx
AlexeyPogorelov: https://github.com/sparkletown/sparkle/pull/1396/commits/026b25922801214655c53f5ddac1ebdbf802d514

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
AlexeyPogorelov: https://github.com/sparkletown/sparkle/pull/1396/commits/8c232a5d1ae872885bf6d3d3e41c8104f633279b

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <div className="ShareModal__container">
+          <h3 className="ShareModal__container-title">Share this room</h3>
+          {linkButtons.map((linkButton) => (
+            <LinkButton href={linkButton.href} key={linkButton.title}>
+              <FontAwesomeIcon icon={linkButton.icon} />
+              <span className="ShareModal__button-text">
+                {linkButton.title}
+              </span>
+            </LinkButton>
+          ))}
```
mike-lvov, line 76: useMemo

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+      href: facebookHref,
+      icon: faFacebook,
+      title: "Facebook",
+    },
+    {
+      href: twitterHref,
+      icon: faTwitter,
+      title: "Twitter",
+    },
+  ];
```
mike-lvov, line 51: useMemo

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { faFacebook, faTwitter } from "@fortawesome/free-brands-svg-icons";
+import { DEFAULT_MAP_BACKGROUND } from "settings";
+import { WithId } from "utils/id";
+import { getVenueFullLocation } from "utils/url";
+import { PosterPageVenue } from "types/venues";
+import classNames from "classnames";
+import { LinkButton } from "components/atoms/LinkButton";
+import { useShowHide } from "hooks/useShowHide";
+
+import "./ShareModal.scss";
```
mike-lvov, line 14:
```suggestion
import React, { FC } from "react";
import { Modal } from "react-bootstrap";
import { CopyToClipboard } from "react-copy-to-clipboard";
import classNames from "classnames";
import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
import { faFacebook, faTwitter } from "@fortawesome/free-brands-svg-icons";

import { DEFAULT_MAP_BACKGROUND } from "settings";

import { PosterPageVenue } from "types/venues";

import { WithId } from "utils/id";
import { getVenueFullLocation } from "utils/url";

import { useShowHide } from "hooks/useShowHide";

import { LinkButton } from "components/atoms/LinkButton";

import "./ShareModal.scss";
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
```diff
@@ -0,0 +1,82 @@
...
+      width: 195px;
+      height: 70px;
+      background-image: url(/assets/icons/sparkle-nav-logo-text.png);
+      background-size: 195px 70px;
+    }
+  }
+
+  &__title {
+    position: absolute;
+    font-size: 20px;
```
mike-lvov, line 41: from consts

.env.example
```diff
@@ -9,6 +9,7 @@ REACT_APP_LOGROCKET_APP_ID=
 REACT_APP_MIXPANEL_PROJECT_TOKEN=
 REACT_APP_WS_RELAY_URL=
 REACT_APP_CODE_CHECK_URL=
+REACT_APP_FACEBOOK_APP_ID=
```
mike-lvov, line 12: NOTE: We should set this in our env

src/components/atoms/LinkButton/LinkButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const LinkButton: FC<LinkButtonProps> = ({ children, href }) => {
+  return (
+    <a href={href} target="_blank" rel="noreferrer" className="button-link">
+      {children}
+    </a>
+  );
+};
```
mike-lvov, line 18:
```suggestion
export const LinkButton: FC<LinkButtonProps> = ({ children, href }) => (
    <a href={href} target="_blank" rel="noreferrer" className="button-link">
      {children}
    </a>
);
```

mike-lvov (commented):
A few minor tweaks and good to go

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/6ab7d04659ddfbdc2b675e3927a546b513eb0981

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/7cb535973ceeee20c404c91e3d1af6a931267a7b

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/7cb535973ceeee20c404c91e3d1af6a931267a7b

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
AlexeyPogorelov: https://github.com/sparkletown/sparkle/pull/1396/commits/f7ed77b46190076dbf459beadc940a3b9cefb6fb

src/components/atoms/LinkButton/LinkButton.tsx
AlexeyPogorelov: https://github.com/sparkletown/sparkle/pull/1396/commits/f8c9af03f719101e95a7e0e563b7fe6e6d2dda89

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+        title: "Twitter",
+      },
+    ];
+    return linkButtons.map((linkButton) => (
+      <LinkButton href={linkButton.href} key={linkButton.title}>
+        <FontAwesomeIcon icon={linkButton.icon} />
+        <span className="ShareModal__button-text">{linkButton.title}</span>
+      </LinkButton>
+    ));
+  }, [facebookHref, twitterHref]);
```
mike-lvov, line 64:
```suggestion
  const linkButtonsComponent = useMemo(() => {
    const linkButtons = [
      {
        href: facebookHref,
        icon: faFacebook,
        title: "Facebook",
      },
      {
        href: twitterHref,
        icon: faTwitter,
        title: "Twitter",
      },
    ];
    
    return linkButtons.map((linkButton) => (
      <LinkButton href={linkButton.href} key={linkButton.title}>
        <FontAwesomeIcon icon={linkButton.icon} />
        <span className="ShareModal__button-text">{linkButton.title}</span>
      </LinkButton>
    ));
  }, [facebookHref, twitterHref]);
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+            className="ShareModal__background"
+            src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
+            alt="map"
+          />
+          <div className="ShareModal__logo"></div>
+          <h3 className="ShareModal__title">Room Title</h3>
+          <span className="ShareModal__url-text">{url}</span>
+        </div>
+        <div className="ShareModal__container">
+          <h3 className="ShareModal__container-title">Share this room</h3>
```
mike-lvov, line 81:
```suggestion
        <div className="ShareModal__content">
          <h3 className="ShareModal__content-title">Share this room</h3>
```

mike-lvov (approved):
LGTM after 2 minor fixes.

Bear in mind that we have to set `REACT_APP_FACEBOOK_APP_ID` in all of our ENVs

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/fcd83046a8e08e34eb41e0710d78f1d58a6c622d

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/b2176d5b85f556f5afebb1b4163545a154129eab

package.json
```diff
@@ -48,6 +49,7 @@
     "re-resizable": "^6.5.4",
     "react": "^16.13.1",
     "react-bootstrap": "^1.0.1",
+    "react-copy-to-clipboard": "^5.0.3",
```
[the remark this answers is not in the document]
0xdevalias: [NB] We also could have used `useCopyToClipboard` from `react-use` (which is already in our dependencies)

- https://github.com/streamich/react-use/blob/master/docs/useCopyToClipboard.md

Though it seems as though it's a more 'low level' hook/functionality.

package-lock.json
```diff
@@ -3025,6 +3025,21 @@
...
+        "@fortawesome/fontawesome-common-types": "^0.2.35"
+      },
+      "dependencies": {
+        "@fortawesome/fontawesome-common-types": {
+          "version": "0.2.35",
+          "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-common-types/-/fontawesome-common-types-0.2.35.tgz",
+          "integrity": "sha512-IHUfxSEDS9dDGqYwIW7wTN6tn/O8E0n5PcAHz9cAaBoZw6UpG20IG/YM3NNLaGPwPqgjBAFjIURzqoQs3rrtuw=="
+        }
+      }
+    },
```
0xdevalias, line 3042: Looks like we want to use a different version of this dependency so that we aren't changing the  `@fortawesome/fontawesome-common-types` version.

This test should have failed based on this I believe..

- https://github.com/sparkletown/sparkle/blob/staging/package-lock.test.ts#L16-L40

---

Looking at the versions, I believe we want `5.15.1` here:

- https://github.com/FortAwesome/Font-Awesome/blob/5.15.1/js-packages/@fortawesome/free-brands-svg-icons/package.json#L49-L51

src/components/atoms/LinkButton/LinkButton.tsx
```diff
@@ -0,0 +1,16 @@
...
+    HTMLAnchorElement
+  > {
+  href: string;
+}
+
+export const LinkButton: FC<LinkButtonProps> = ({ children, href }) => (
+  <a href={href} target="_blank" rel="noreferrer" className="button-link">
+    {children}
+  </a>
+);
```
0xdevalias, line 16: This should use the `externalUrlAdditionalProps` const from `utils/url`.

Additionally, you say that it accepts all `DetailedHTMLProps` in `LinkButtonProps` yet you don't actually apply them to the `a` tag. In addition to this, we need to make sure that if `className` is passed, it doesn't overwrite our default styles here.

Our style is to use `React.FC` rather than just `FC`.

```suggestion
export const LinkButton: React.FC<LinkButtonProps> = ({ children, href, className, ...extraProps }) => {
  const linkClasses = classNames("button-link", className)
  
  return (
    <a href={href} className={linkClasses} {...externalUrlAdditionalProps, ...extraProps}>
      {children}
    </a>
  );
}
```

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -18,6 +18,7 @@ import { PosterPageControl } from "./components/PosterPageControl";
 import { PosterPageSettingsControl } from "./components/PosterPageSettingsControl";
 
 import { usePosterVideo } from "./usePosterVideo";
+import { ShareModal } from "./components/ShareModal";
```
0xdevalias, line 21: Keep imports from the same paths together.

```suggestion
import { PosterPageSettingsControl } from "./components/PosterPageSettingsControl";
import { ShareModal } from "./components/ShareModal";

import { usePosterVideo } from "./usePosterVideo";
```

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -143,6 +152,7 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
           introVideoUrl={introVideoUrl}
         />
       )}
+      <ShareModal show={isShareModal} onHide={hideShareModal} venue={venue} />
```
0xdevalias, line 155:
```suggestion
      <ShareModal show={isShareModalShown} onHide={hideShareModal} venue={venue} />
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
```diff
@@ -0,0 +1,82 @@
...
+.ShareModal {
+  &__header {
+    position: relative;
+    height: 200px;
+    margin: $spacing--sm;
+    border-radius: $border-radius--lg;
+  }
+
+  &__background {
+    z-index: z(map-background);
```
0xdevalias, line 12: Why is the `map-background` `z-index` layer being used here? That seems wrong..? If this needs a `z-index` layer at all, then it should probably be using it's own?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { WithId } from "utils/id";
+import { getVenueFullLocation } from "utils/url";
+
+import { useShowHide } from "hooks/useShowHide";
+
+import { LinkButton } from "components/atoms/LinkButton";
+
+import "./ShareModal.scss";
+
+interface ConfirmationModalProps {
```
0xdevalias, line 21: Export the props interfaces/types.

Also, this is named incorrectly. This is `ShareModal` not `ConfirmationModal`

```suggestion
export interface ShareModalProps {
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  show: boolean;
+  onHide: () => void;
+  venue: WithId<PosterPageVenue>;
+}
+
+export const ShareModal: FC<ConfirmationModalProps> = ({
```
0xdevalias, line 27:
```suggestion
export const ShareModal: React.FC<ShareModalProps> = ({
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const linkText = isShowCopiedText ? "Link Copied" : "Copy Link";
+
+  const linkClasses = classNames("ShareModal__link", {
+    "ShareModal__link--copied": isShowCopiedText,
+  });
+
+  const textTitle = `Check out this OHBM Poster, ${venue.poster?.title} by ${venue.poster?.authorName} at ${url}`;
+  const facebookHref = `https://www.facebook.com/sharer/sharer.php?app_id=${process.env.REACT_APP_FACEBOOK_APP_ID}&u=${url}&quote=${textTitle}&thumbnail=${venue.host?.icon}&picture=${venue.host?.icon}&hashtag=#OHBM2021`;
+  const twitterHref = `https://twitter.com/intent/tweet?text=${textTitle}&hashtags=OHBM2021`;
```
0xdevalias, line 43: There are hardcoded references to OHBM here that need to be changed.

`process.env.REACT_APP_FACEBOOK_APP_ID` shouldn't be used directly, but should follow the pattern of our existing ENV vars in `src/secrets`.

You're also using potentially undefined variables within strings, yet not handling the `undefined` by defaulting or otherwise, which will lead to the literal string "undefined" being in those strings.

Additionally, we should probably extract these URLs out into helper functions or similar rather than having such large `const`s in the body of the component like this.

0xdevalias: These could also likely be included in the `linkButtonsComponent` `useMemo` below, since that appears to be the only place they are needed.

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      },
+      {
+        href: twitterHref,
+        icon: faTwitter,
+        title: "Twitter",
+      },
+    ];
+
+    return linkButtons.map((linkButton) => (
+      <LinkButton href={linkButton.href} key={linkButton.title}>
```
0xdevalias, line 60: [nit] Prefer key to be the first prop within `.map`'s as it keeps it close to the `.map`:

```suggestion
    return linkButtons.map((linkButton) => (
      <LinkButton key={linkButton.title} href={linkButton.href}>
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <Modal show={show} onHide={onHide}>
+      <Modal.Body className="ShareModal">
+        <div className="ShareModal__header">
+          <img
+            width="100%"
+            className="ShareModal__background"
+            src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
+            alt="map"
+          />
```
0xdevalias, line 76: Do we really want to be loading the entire map background here?

Also, why are we using the `width="100%"` property here? We should be setting this through CSS only.

```suggestion
          <img
            className="ShareModal__background"
            src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
            alt="map"
          />
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <div className="ShareModal__logo"></div>
+          <h3 className="ShareModal__title">Room Title</h3>
+          <span className="ShareModal__url-text">{url}</span>
+        </div>
+        <div className="ShareModal__content">
```
0xdevalias, line 81: [nit] Whitespace between sibling blocks of JSX to increase code readability.
```suggestion
        </div>

        <div className="ShareModal__content">
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <h3 className="ShareModal__content-title">Share this room</h3>
+          {linkButtonsComponent}
+          <CopyToClipboard text={url}>
+            <div className={linkClasses} onClick={showCopiedText}>
+              {linkText}
+            </div>
+          </CopyToClipboard>
```
0xdevalias, line 88: [nit] Whitespace between sibling blocks of JSX to increase code readability.

```suggestion
          <h3 className="ShareModal__content-title">Share this room</h3>

          {linkButtonsComponent}

          <CopyToClipboard text={url}>
            <div className={linkClasses} onClick={showCopiedText}>
              {linkText}
            </div>
          </CopyToClipboard>
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const linkClasses = classNames("ShareModal__link", {
+    "ShareModal__link--copied": isShowCopiedText,
+  });
+
+  const textTitle = `Check out this OHBM Poster, ${venue.poster?.title} by ${venue.poster?.authorName} at ${url}`;
+  const facebookHref = `https://www.facebook.com/sharer/sharer.php?app_id=${process.env.REACT_APP_FACEBOOK_APP_ID}&u=${url}&quote=${textTitle}&thumbnail=${venue.host?.icon}&picture=${venue.host?.icon}&hashtag=#OHBM2021`;
+  const twitterHref = `https://twitter.com/intent/tweet?text=${textTitle}&hashtags=OHBM2021`;
+
+  const linkButtonsComponent = useMemo(() => {
```
0xdevalias, line 45: [nit] There appear to be multiple link buttons here so `linkButtonComponents`, or better yet `renderedLinkButtons` to match our existing patterns:

```suggestion
  const renderedLinkButtons = useMemo(() => {
```

src/utils/url.ts
```diff
@@ -82,3 +82,7 @@ export const externalUrlAdditionalProps = {
 
 export const getExtraLinkProps = (isExternal: boolean) =>
   isExternal ? externalUrlAdditionalProps : {};
+
+export const getVenueFullLocation = (venueId: string) => {
+  return new URL(venueInsideUrl(venueId), window.location.origin).href;
+};
```
0xdevalias, line 88: [NB]

```suggestion
export const getFullVenueInsideUrl = (venueId: string) => {
  return new URL(venueInsideUrl(venueId), window.location.origin).href;
};
```

0xdevalias (changes requested):
There are a bunch of blocking things here, plus a few NB/nit (that would be quick to fix up as well, so preferable to do so)

---

@oleksii-front Also, when using 'closing keywords', you need to use the keyword before each issue, not just once in front of all of them. Otherwise it won't link them correctly. I've fixed this for you on this PR, but please keep it in mind for future:

image

package-lock.json
0xdevalias: > This test should have failed based on this I believe..
> 
> - https://github.com/sparkletown/sparkle/blob/staging/package-lock.test.ts#L16-L40

Bonus points if you figure out why that test wasn't failing here and fix up the bug in it.

.env.example
0xdevalias: @oleksii-front I have deleted your comment here, please don't share this value on our public code repo.

package.json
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/c4833d0122ba61d34bda0001934521f0ef9d7c05

src/components/atoms/LinkButton/LinkButton.scss
```diff
@@ -0,0 +1,14 @@
+@import "scss/constants.scss";
+
+.button-link {
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/d3bbfeb1a8934897731782feef9d80c5d415c695

src/components/atoms/LinkButton/LinkButton.scss
```diff
@@ -0,0 +1,14 @@
...
+  text-align: center;
+  text-decoration: none;
+  margin-top: $spacing--lg;
+  background-color: $lighter-intermediate-grey;
+  padding: $spacing--md $spacing--lg;
+  border-radius: $border-radius--xl;
+  cursor: pointer;
+  &:hover {
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/d3bbfeb1a8934897731782feef9d80c5d415c695

src/components/atoms/LinkButton/LinkButton.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/3632c4de26b887043c8c61f1daad943747517382

0xdevalias: Looks like you changed the ordering of the props added to the `<a` from what I suggested. Why is that?

src/components/templates/PosterPage/PosterPage.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/a6a886bae3d491ba531c9f7bf30a4b617c799724

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -36,6 +37,11 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
     show: showIntroVideoModal,
     hide: hideIntroVideoModal,
   } = useShowHide();
+  const {
+    isShown: isShareModal,
+    show: showShareModal,
+    hide: hideShareModal,
+  } = useShowHide();
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/1db768253f0bf3126450f535d88516b2089316b3

src/components/templates/PosterPage/PosterPage.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/1db768253f0bf3126450f535d88516b2089316b3

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/00d18055b7e7b589a6736734c3d1802eeeae2fe8

package-lock.json
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/atoms/LinkButton/LinkButton.tsx
0xdevalias: @oleksii-front This still seems to be outstanding/requires answering?

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
0xdevalias: @oleksii-front This still seems to be outstanding/requires answering?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front This still seems to be outstanding/requires answering?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front You marked this resolved with out any comment or commit hash. I know that both @mike-lvov and I have asked you not to do this in the past. Please follow our standards.

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front You marked this resolved with out any comment or commit hash. I know that both @mike-lvov and I have asked you not to do this in the past. Please follow our standards.

src/utils/url.ts
0xdevalias: @oleksii-front You marked this resolved with out any comment or commit hash. I know that both @mike-lvov and I have asked you not to do this in the past. Please follow our standards.

package-lock.json
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
0xdevalias: Looks like there are still a lot of `px` values and similar being used directly here (eg. `height`, `top`, `left`, etc)

@mike-lvov Are you ok with these as they are, or should they be extracted further?

src/components/templates/PosterPage/components/ShareModal/helpers.ts
```diff
@@ -0,0 +1,18 @@
+import { FACEBOOK_APP_ID } from "secrets";
```
0xdevalias, line 1: [NB] We tend to call these files `utils` rather than `helpers`.

src/components/templates/PosterPage/components/ShareModal/helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (!venue.poster) return "Check out this OHBM Poster";
+
+  return `Check out this OHBM Poster, ${venue.poster.title} by ${venue.poster.authorName} at ${url}`;
+};
+
+export const getFacebookHref = (url: string, text: string) =>
+  `https://www.facebook.com/sharer/sharer.php?app_id=${FACEBOOK_APP_ID}&u=${url}&quote=${text}&hashtag=#OHBM2021`;
+
+export const getTwitterHref = (text: string) =>
+  `https://twitter.com/intent/tweet?text=${text}&hashtags=OHBM2021`;
```
0xdevalias, line 18: As I called out in a previous review, this contains hardcoded text related to OHBM, which needs to be removed/changed before this PR can be merged. This feature needs to be generic.

(@mike-lvov FYI)

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { WithId } from "utils/id";
+import { getFullVenueInsideUrl } from "utils/url";
+
+import { LinkButton } from "components/atoms/LinkButton";
+
+import {
+  getTitleTextForSharing,
+  getFacebookHref,
+  getTwitterHref,
+} from "./helpers";
```
0xdevalias, line 21: [nit, NB] We tend to call these `utils` rather than `helpers`

```suggestion
} from "./utils";
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  venue: WithId<PosterPageVenue>;
+}
+
+export const ShareModal: React.FC<ShareModalProps> = ({
+  show,
+  onHide,
+  venue,
+}) => {
+  const url = getFullVenueInsideUrl(venue.id);
+  const [{ value: isShowCopiedText }, copyToClipboard] = useCopyToClipboard();
```
0xdevalias, line 37: [NB] This reads better:

```suggestion
  const [{ value: hasCopiedText }, copyToClipboard] = useCopyToClipboard();
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @oleksii-front This still seems to be outstanding?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <Modal show={show} onHide={onHide}>
+      <Modal.Body className="ShareModal">
+        <div className="ShareModal__header">
+          <img
+            className="ShareModal__background"
+            src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
+            alt="map"
+          />
```
0xdevalias, line 83: As I believe I called out in a preview review thread, do we want to be using the entire map background image here? That's a huge image.. I expect we don't want to be doing that.

We might want to use `venue.coverImageUrl` here potentially?

@mike-lvov Thoughts?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { useCallback, useMemo } from "react";
+import { Modal } from "react-bootstrap";
+import { useCopyToClipboard } from "react-use";
+import classNames from "classnames";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faFacebook, faTwitter } from "@fortawesome/free-brands-svg-icons";
+
+import { DEFAULT_MAP_BACKGROUND } from "settings";
+
+import { PosterPageVenue } from "types/venues";
```
0xdevalias, line 10: I expect we don't want to be importing `DEFAULT_MAP_BACKGROUND` here at all.

```suggestion
import { PosterPageVenue } from "types/venues";
```

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <Modal show={show} onHide={onHide}>
+      <Modal.Body className="ShareModal">
+        <div className="ShareModal__header">
+          <img
+            className="ShareModal__background"
+            src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
+            alt="map"
+          />
+          <div className="ShareModal__logo"></div>
```
0xdevalias, line 84: What's the intent/purpose of this `div`? Should it have content within it? If not then it should use the `<div />` single tag shorthand.

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <h3 className="ShareModal__title">Room Title</h3>
```
0xdevalias, line 85: Should this be set to something other than the static text 'Room Title'? What's the intent of this section?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <span className="ShareModal__url-text">{url}</span>
+        </div>
+
+        <div className="ShareModal__content">
+          <h3 className="ShareModal__content-title">Share this room</h3>
```
0xdevalias, line 90: I don't think we should be using the word 'room' here. At a minimum we probably want to be saying 'venue' instead. We could probably just use "Share this page" to be super generic.

```suggestion
          <h3 className="ShareModal__content-title">Share this page</h3>
```

@mike-lvov Thoughts?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+          {renderedLinkButtons}
+
+          <div className={linkClasses} onClick={toCopy}>
+            {linkText}
+          </div>
```
0xdevalias, line 96: We should probably be showing some kind of icon/similar here to show that it's text that can be copied. As well as potentially being within an `input` field or similar rather than just being a `div`?

@mike-lvov Thoughts?

0xdevalias (changes requested):
[CRO] Still seem to be a number of outstanding issues in this PR, including a whole bunch that were called out in previous reviews and yet not fixed before re-requesting review.

Please ensure you have properly fixed ALL of the previous review threads/comments before re-submitting this for review; it just wastes the reviewer's time otherwise.

---

Combing through the previous review threads (which isn't something I as reviewer should have to do, as this is the responsibility of the PR creator), here is a non-exhaustive list of outstanding review threads that need to be addressed one way or another (note: closing them without comment doesn't count as 'addressing' them):

- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637799955
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637797654
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637794424
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637794625
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637794786
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637794962
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637795032
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637803701
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637795791
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637796622
- https://github.com/sparkletown/sparkle/pull/1396#discussion_r637796815

---

@mike-lvov I think we should definitely do a clickthrough of this PR once all of this has been fixed up to ensure it functions as expected.

src/components/atoms/LinkButton/LinkButton.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { DetailedHTMLProps, AnchorHTMLAttributes } from "react";
```
0xdevalias, line 1: Why are we creating a new `LinkButton` component from scratch here, when seemingly we already have `src/components/atoms/Button/Button.tsx` that has an `isLink` prop on it?

(cc // @mike-lvov FYI/thoughts?)

mike-lvov: The Button/Button.tsx uses react-router <Link />, which seemingly can't handle an navigating to an external URL. It probably should be renamed somehow, but I can see value in a component like it.

src/components/templates/PosterPage/components/ShareModal/helpers.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/ce99953a2c8fa07fc9ad3110344481163911e470

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: There was nothing about that in description. Please create another issue if you want me to improve this.

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/5c68b1f651cd65bdc58a1ac373270b0e2bf7cd80

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: Could you please create other issue for that improvement? We need to fix it in the design

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: this is for logotype. No content within

oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/2f104f9be574e578c0dfc260d68a1c1eb2f95337

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/ce99953a2c8fa07fc9ad3110344481163911e470

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/08a1109cf3e296a0fc84656b5461df78f32fa60f

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/08a1109cf3e296a0fc84656b5461df78f32fa60f

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/2deaa63420083f3bd24823d446c4c10e3da4bc2a

src/components/atoms/LinkButton/LinkButton.tsx
oleksii-front: I think that we could improve that later.

src/components/templates/PosterPage/components/ShareModal/helpers.ts
oleksii-front: Currently I leave it here just for test.

mike-lvov (approved):
Last minor tweaks/changes. +

Please, comment out `room title` and make the link white 
image

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: @mike-lvov

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: This is a one word change… the overhead of creating a new issue is infinitely more than fixing it here and now if it needs to be fixed.

@mike-lvov ?

0xdevalias: @oleksii-front also, when considering something OOS, our process is for the PR owner to create the new issue capturing the requested aspect.

src/components/atoms/LinkButton/LinkButton.tsx
0xdevalias: Then we should create a backlog issue to track that.

src/components/templates/PosterPage/components/ShareModal/helpers.ts
0xdevalias: What do you mean leave it here for test?

This needs to be loaded from settings/sovereign venue config/similar, not hard coded.

mike-lvov:
Looks like the Facebook doesn't work
image (9).


NOTE: Set the FACEBOOK_APP_ID across all of the envs

0xdevalias:
> Please, comment out `room title` and make the link white

@mike-lvov @oleksii-front Or we could just actually use the room title there rather than a static string.. which I expect is the original intent in the designs..

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: > Please create another issue if you want me to improve this.

When a PR creator calls something OOS (out of scope), our process is for the PR creator to create the backlog issue to capture that.

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
0xdevalias: @mike-lvov That confuses me, since one of the main things you seem to call out in PR reviews is extracting these sorts of things..?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
0xdevalias: Probably want to handle the potentially `undefined` case here better? @mike-lvov Thoughts?

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
mike-lvov: Using the constants everywhere where possible, yes. heights/width, not always(but I don't know if we should do it, if we are not reusing them, honestly). Distances - I don't think I ever called out

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (!venue.poster) return "Check out this OHBM Poster";
+
+  return `Check out this OHBM Poster, ${venue.poster.title} by ${venue.poster.authorName} at ${url}`;
+};
+
+export const getFacebookHref = (url: string, text: string) =>
+  `https://www.facebook.com/sharer/sharer.php?app_id=${FACEBOOK_APP_ID}&u=${url}&quote=${text}&hashtag=#OHBM2021`;
+
+export const getTwitterHref = (text: string) =>
+  `https://twitter.com/intent/tweet?text=${text}&hashtags=OHBM2021`;
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/6d86720d1506504a9126377db11db39e11afa951

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export const ShareModal: React.FC<ShareModalProps> = ({
+  show,
+  onHide,
+  venue,
+}) => {
+  const url = getFullVenueInsideUrl(venue.id);
+  const [{ value: hasCopiedText }, copyToClipboard] = useCopyToClipboard();
+
+  const toCopy = useCallback(() => {
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/3adbd79cc5c04aa0b50f788bfb986013c722aa1d

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
```diff
@@ -0,0 +1,86 @@
...
+    width: 100%;
+    border-radius: $border-radius--lg;
+    height: 100%;
+    opacity: 0.5;
+  }
+
+  &__logo {
+    $logo-height: 70px;
+    $logo-width: 50px;
+    $logo-background-width: 195px;
```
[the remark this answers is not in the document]
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/27b84ba0b32e5a5edb7cb66ce560214d90bf727d

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
+import { FACEBOOK_APP_ID } from "secrets";
+import { PosterPageVenue } from "types/venues";
+import { WithId } from "utils/id";
+
+const getCategoriesFromVenue = (venue: WithId<PosterPageVenue>) => {
+  return venue?.poster?.categories || [];
+};
```
0xdevalias, line 7: Why do we need this function? This seems like something we would just tend to do directly?

Assuming we need/want it, this aligns with our standards + fixes unneeded safety checks/etc:
```suggestion
const getCategoriesFromVenue = (venue: WithId<PosterPageVenue>) => venue.poster?.categories ?? [];
```

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const getTitleTextForSharing = (
+  venue: WithId<PosterPageVenue>,
+  url: string
+) => {
```
0xdevalias, line 12: We should used named params here, not positional args.
```suggestion
export interface GetTitleTextForSharingProps {
  venue: WithId<PosterPageVenue>;
  url: string;
}
export const getTitleTextForSharing = ({ venue, url }: GetTitleTextForSharingProps) => {
```

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!venue.poster) return "Check out this OHBM Poster";
+
+  return `Check out this OHBM Poster, ${venue.poster.title} by ${venue.poster.authorName} at ${url}`;
```
0xdevalias, line 15: We don't want OHBM references hardcoded into this PR (as I have called out in previous reviews)

There is a `const` in `src/settings` that is used to set the window title, and is the 'main event name' or similar. We could potentially use that (though it's hardcoded per environment)

Or we could just remove it entirely and have the text as "Check out this poster"

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  venue: WithId<PosterPageVenue>,
+  url: string,
+  text: string
+) => {
+  const categories = getCategoriesFromVenue(venue);
+  const hashtagsString = categories.reduce((result, hash, index, array) => {
+    const isLast = index === array.length - 1;
+    const separator = isLast ? "" : ", ";
+    return result + `#${hash}${separator}`;
+  }, "");
```
0xdevalias, line 28: [NB] This seems.. convoluted. Is there a simpler way for us to handle this? What's the intended outcome of this code? It looks like we could just do
```
categories.map(category => `#${category}).join(', ')
```
or similar to solve the same need here?

0xdevalias: Does the sharer accept multiple hashtags in this format? Do we need to url encode the `#`?

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  return `https://www.facebook.com/sharer/sharer.php?app_id=${FACEBOOK_APP_ID}&u=${url}&quote=${text}&hashtag=${hashtagsString}`;
```
0xdevalias, line 30: [NB] We would probably be better to use the JS URL helpers for constructing this URL, rather than joining strings. I believe that will handle encoding/escaping for us automatically then.

- https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams
- https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

src/components/templates/PosterPage/components/ShareModal/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const getTwitterHref = (
+  venue: WithId<PosterPageVenue>,
+  text: string
+) => {
+  const categories = getCategoriesFromVenue(venue);
+  const hashtagsString = categories.join(",");
+  return `https://twitter.com/intent/tweet?text=${text}&hashtags=${hashtagsString}`;
```
0xdevalias, line 39: [NB] We would probably be better to use the JS URL helpers for constructing this URL, rather than joining strings. I believe that will handle encoding/escaping for us automatically then.

- https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams
- https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <Modal show={show} onHide={onHide}>
+      <Modal.Body className="ShareModal">
+        <div className="ShareModal__header">
+          <img
+            className="ShareModal__background"
+            src={venue.config?.landingPageConfig?.coverImageUrl}
+            alt="map"
+          />
```
0xdevalias, line 81: This can be potentially `undefined` yet we're not handling that case. I believe I called this out in my last review as well (or something similar to it)

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <div className="ShareModal__logo" />
+          <h3 className="ShareModal__title">{venue.parentId}</h3>
```
0xdevalias, line 83: Why is this using `venue.parentId` here? From memory this was set to room title or similar previously wasn't it? The `parentId` isn't equivalent to that.

0xdevalias (changes requested):
[CRO] Questions/comments/changes.

@mike-lvov Happy for you to re-review after these changes are made and get it approved/merged if you're then happy with it.

See the things I called out in this review as the concerns I have, and judge based on that.

src/components/templates/PosterPage/components/ShareModal/utils.ts
0xdevalias: Or we could just remove it entirely and have the text as "Check out this poster", which is probably my preference/the simplest/most robust solution here.

0xdevalias: What do you think of this as a solution @mike-lvov? Uses `PLATFORM_BRAND_NAME` (which defaults to `Sparkle` on most environments, and is `OHBM2021` in `env/ohbm`), and seems it would read well in both situations/generally.

```suggestion
  if (!venue.poster) return `Check out this poster on ${PLATFORM_BRAND_NAME}`;

  return `Check out this poster on ${PLATFORM_BRAND_NAME}! ${venue.poster.title} by ${venue.poster.authorName} at ${url}`;
```

0xdevalias: Though another question, what is the point of this line?
```
if (!venue.poster) return `Check out this poster on ${PLATFORM_BRAND_NAME}`;
```

I'm guessing it's guarding against the case when `venue.poster` is `undefined`.. but it will end up making a pointless social media share since it's not able to include any details about the poster. We might be better with just a straight `if (!vnue.poster) return` here, and then handling that case of `undefined` in a way that we communicate the error to the user/don't allow it to post/etc.

src/components/templates/PosterPage/components/ShareModal/utils.ts
0xdevalias: This seems to be the relevant docs for this feature: https://developers.facebook.com/docs/sharing/reference/share-dialog#redirect

Looking at https://developers.facebook.com/docs/sharing/reference/share-dialog#params_share it seems to imply that only a single hashtag can be included?

src/components/templates/PosterPage/components/ShareModal/utils.ts
0xdevalias: And these should be done using functions from `utils/url` (or create new appropriate functions to do so there)

src/components/templates/PosterPage/components/ShareModal/utils.ts
0xdevalias: We could also potentially use `PLATFORM_BRAND_NAME` from `src/settings` as the hashtag? This would work for `env/ohbm` (which is set to `OHBM2021`), in most other environments it will just be set to `Sparkle`.

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1396#issuecomment-847330559

> NOTE: Set the FACEBOOK_APP_ID across all of the envs

@mike-lvov We'll also need to make sure we properly set up the Facebook developer application for Sparkle to create this app id (or ensure that @oleksii-front gives us the owner/etc permissions for the existing one) too.

src/components/templates/PosterPage/components/ShareModal/utils.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/a0cb5302188a3401cc34dbca148d6870a0e629ea

src/components/templates/PosterPage/components/ShareModal/utils.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/9a27d5acb19bde1058ae95272a4ffae0a3b76283

oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/cb33a3f5ddfb6801e8af4f6bebc3d8bdedb72b69

src/components/templates/PosterPage/components/ShareModal/utils.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/7fd77d97e2c82c75f9dfb0f9295fe01606cae835

src/components/templates/PosterPage/components/ShareModal/utils.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/444fbee5a64755e19d21cc3020703b59dbd60b69

src/components/templates/PosterPage/components/ShareModal/utils.ts
oleksii-front: https://github.com/sparkletown/sparkle/pull/1396/commits/ec6ea08642a234ebc11f42649bead453454c5e88

package-lock.json
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1396/commits/a611e3516ee9464e51516aa5c868075acd54f997

src/components/atoms/LinkButton/LinkButton.tsx
mike-lvov: NB

src/components/templates/PosterPage/components/ShareModal/ShareModal.scss
mike-lvov: Was removed

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in prev commits

src/utils/url.ts
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/helpers.ts
mike-lvov: Done in prev commits

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Was removed in https://github.com/sparkletown/sparkle/pull/1396/commits/a458f78ea7f7ac7372a68bfd8011a8f50f477a9c

src/components/atoms/LinkButton/LinkButton.tsx
mike-lvov: What kind of backlog issue should be created?

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: > We should probably be showing some kind of icon/similar here to show that it's text that can be copied. As well as potentially being within an input field or similar rather than just being a div?
> 
> @mike-lvov Thoughts?

I didn't think deeply about it. It looks like in the designs. Unresolve, if you think otherwise

image

src/components/templates/PosterPage/components/ShareModal/ShareModal.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1396/commits/dd17fa0af56572ea25c856c30766c5a91b005134

src/settings.ts
```diff
@@ -22,6 +22,9 @@ export const SPARKLE_TERMS_AND_CONDITIONS_URL =
 export const SPARKLE_PRIVACY_POLICY =
   "https://sparklespaces.com/privacy-policy/";
 
+// Sparkle facebook app id. More settings can be found at https://developers.facebook.com/apps/2633721400264126/dashboard/
+export const FACEBOOK_SPARKLE_APP_ID = "2633721400264126";
```
0xdevalias, line 26: @mike-lvov This absolutely shouldn't have been hardcoded and exposed in the source here like this.. That's the whole point/reason of the ENV vars mapped in through from `src/secrets.js`

0xdevalias: Clicking through to that link I have no access to the project. Can you please share that with me and the rest of the team who needs it.

package.json
```diff
@@ -6,7 +6,7 @@
     "@bugsnag/js": "^7.5.1",
     "@bugsnag/plugin-react": "^7.3.5",
     "@fortawesome/fontawesome-svg-core": "^1.2.32",
-    "@fortawesome/free-brands-svg-icons": "^5.15.3",
+    "@fortawesome/free-brands-svg-icons": "^5.15.1",
```
0xdevalias: You changed the version here, yet didn't run `npm install` to update the `package-lock.json`, which means for all intents and purposes, it's not changed/fixed.

PR #1397, opened 2021-05-14
My inaugural PR for sparkle

soichih:
This is my very first PR for sparkle to see what happens. :) I am not sure if env/ohbm is where I need to be PR-ing to.

Anyhow, this change itself does the following 2 things.

1) Some of the css rules specific to ProfilePictureInput were stored in pages/Account/Accounts.scss. I don't know why there were there, but I've moved them to ProfilePictureInput.scss which I believe is a better place.

2) The CSS rule for "Upload your profile pic" button looked odd, so I've updated it.

Before
before

After
after

That is all.

### DIFF
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
@@ -1,7 +1,5 @@
 @import "scss/constants";
 
-$white: #fff;
-
 .default-avatars-container {
   display: flex;
   flex-direction: row;
@@ -38,7 +36,6 @@ $white: #fff;
   border: solid $primary;
   padding: 10px 30px;
   margin: 10px 0;
-  color: $white;
   font-weight: 700;
 
   &:hover {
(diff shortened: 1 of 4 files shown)


0xdevalias:
> This is my very first PR for sparkle to see what happens. :)

Hi @soichih and welcome! Thanks for your contribution :)

> I am not sure if env/ohbm is where I need to be PR-ing to.

@soichih This will sort of depend on the features/fixes you’re raising the PRs for. As a bit of a ‘rule of thumb’, if it’s likely to be a benefit to the platform as a whole, then it’s better to raise them against ‘staging’ (from your description + screenshots, this PR would seem to fit this category). If it’s a ‘quick custom hack’ for OHBM, then raising it to ‘env/ohbm’ is generally the better way.

It’s the weekend here in Australia at the moment, but I’ll jump in and review this properly on Monday :)

soichih:
@0xdevalias Thanks. I will resubmit this PR on /staging instead.

0xdevalias:
> Thanks. I will resubmit this PR on /staging instead. 

@soichih Sounds good :) For future reference, it's also possible to change the target branch of a PR without needing to close and reopen it. Then you just rebase your changes onto the staging branch + force push (assuming you're familiar with how to do that with git)

0xdevalias:
Duplicate of #1398

PR #1398, opened 2021-05-15
cleanup styling for ProfilePictureInput + 'upload your profile pic'

soichih:
This is my very first PR for sparkle to see what happens. :)

Anyhow, this change includes the following changes.

1) Some of the css rules specific to ProfilePictureInput were stored in pages/Account/Accounts.scss. I don't know why there were there, but I've moved them to ProfilePictureInput.scss which I believe is a better place.

2) The CSS rule for "Upload your profile pic" button looked odd, so I've updated it.

Before
before

After
after

3) PartyMap/Map was using room.title as the key and raising console error. It can not be used the key as there is no guarantee that it is unique. Switching to an array index.

---

**Edit by @0xdevalias:** Re-creation of #1397 to re-target to `staging` rather than `env/ohbm`.


0xdevalias:
Hey @soichih, looks like your branch for this PR is still based off the `env/ohbm` branch. To fix it up, you'll need to git rebase your changes onto the `staging` branch, then force push it here again. That should clean up all of this:

image

Let me know if you're not sure how to do that and I can dig up some resources/similar for you.

0xdevalias (changes requested):
Hey @soichih, looks like your branch for this PR is still based off the `env/ohbm` branch.

---

**Edit:** Actually, it looks like you may have made the changes directly on your `env/ohbm` branch on your fork. That's not a huge issue, but to make your life easier in future, i'll share some tips here. 

I'm not sure how familiar you are with git/GitHub/etc (so forgive me if i'm repeating the obvious), but the general workflow would be to make a 'feature branch' off of the 'source branch' you intend to commit the changes to (eg. `staging` or `env/ohbm`), and then make the PR from that 'feature branch'. That ensures you keep your local copy of the 'source branch' clean/in sync with the rest of the code, and isolates your changes to your 'feature branch')

We have some (very basic) notes/instructions about this in the main README under the 'Git Flow' heading:

- https://github.com/sparkletown/sparkle#git-flow

But you'll probably find better general documentation about on the net, eg.

- https://guides.github.com/introduction/flow/


---

To fix it up, you'll need to git rebase your changes onto the `staging` branch, then force push it here again. That should clean up all of this:

image

Let me know if you're not sure how to do that and I can dig up some resources/similar for you.

Issue #1390, opened 2021-05-14
Completely disable eventbrite if not used


yarikoptic:
Somewhat of a complement in motivation to #1388 (stripe):  while loading a page noted browser saying waiting on eventbrite.
```shell
$> git grep eventbr
package-lock.json:    "react-eventbrite-popup-checkout": {
package-lock.json:      "resolved": "https://registry.npmjs.org/react-eventbrite-popup-checkout/-/react-eventbrite-popup-checkout-1.0.5.tgz",
package.json:    "react-eventbrite-popup-checkout": "^1.0.5",
public/index.html:    <script src="https://www.eventbrite.co.uk/static/widgets/eb_widgets.js"></script>
src/components/organisms/GiftTicketModal/GiftTicketModal.tsx:          "https://www.eventbrite.com/e/sparkleverse-2020-online-burn-tickets-117154948605"
src/react-app-env.d.ts:declare module "react-eventbrite-popup-checkout" {
```
Would be nice to disable that for `env/ohbm` since afaik also not used ("burn-tickets" in above suggest burning man specific venue setup)

0xdevalias:
I'm not deeply familiar with how the eventbrite code works within the app currently.

The dependency from `package.json` appears to be:

- https://github.com/wizardrylabs/react-eventbrite-popup-checkout

That shouldn't have any real impact on things if it's not being used, so should be fine to leave.

Similarly the string in the `GiftTicketModal` component shouldn't matter.

The script in `public.index.html` (`https://www.eventbrite.co.uk/static/widgets/eb_widgets.js`) is likely what you're seeing the 'waiting to load' for. I'm not immediately aware of what (if anything) that is needed/used for within the app.

Googling for the string lead me to this page:

- https://www.eventbrite.com/platform/docs/embedded-checkout
  - > Using the Eventbrite API, you can embed a ticket checkout widget in your app or website. This is done by including HTML code with a Javascript call and the appropriate attributes. The Javascript call is `https://www.eventbrite.com/static/widgets/eb_widgets.js`

Loading this in the `.html` file as is currently done isn't a very 'Reacty' way to do things, and makes it harder to selectively choose to load/not load.

I would expect that `react-eventbrite-popup-checkout` is likely the 'solution' to this, and since it doesn't mention needing to manually include the `eb_widgets.js` in it's setup instructions, it wouldn't surprise me if it already handles that within its own code.

Looking at the code, that seems to be exactly what it does:

[...]

PR #1400, opened 2021-05-15
remove unused legacy eventbrite code + add 2021 online burn ticket url

0xdevalias:
fixes https://github.com/sparkletown/sparkle/issues/1390 (raised by @yarikoptic) as per my exploration in https://github.com/sparkletown/sparkle/issues/1390#issuecomment-841591461


package.json
```diff
@@ -51,7 +51,6 @@
     "react-dnd": "^11.1.3",
     "react-dnd-html5-backend": "^11.1.3",
     "react-dom": "^16.13.1",
-    "react-eventbrite-popup-checkout": "^1.0.5",
```
0xdevalias, line 54: [note] This didn't appear to be used anymore.

public/index.html
```diff
@@ -71,7 +71,6 @@
     />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <title>Sparkle</title>
-    <script src="https://www.eventbrite.co.uk/static/widgets/eb_widgets.js"></script>
```
0xdevalias, line 74: [note] This didn't appear to be used anymore.

src/react-app-env.d.ts
```diff
@@ -1,12 +1 @@
...
-declare module "react-eventbrite-popup-checkout" {
-  interface PropType {
-    ebEventId: string;
-    className: string;
-    isModal?: boolean;
-    onOrderComplete?: () => void;
-  }
-  const EventbriteButton: React.FunctionComponent<PropType>;
-  export default EventbriteButton;
-}
```
0xdevalias, line 12: [note] This relates to the removed dependency, which didn't appear to be used anymore.

src/settings.ts
```diff
@@ -59,6 +59,8 @@ export const VENUE_NAME_MIN_CHAR_COUNT = 3;
 export const VENUE_NAME_MAX_CHAR_COUNT = 50;
 export const PLAYA_VENUE_NAME = "Jam";
 export const PLAYA_VENUE_ID = "jamonline";
+export const GIFT_TICKET_MODAL_URL =
+  "https://here.burningman.org/event/virtualburn";
```
0xdevalias, line 63: [note] This is the URL for the 2021 online burn, taken from the link on http://sparklever.se/

0xdevalias (commented):
[notes] Some notes on this PR to make review easier.

src/settings.ts
yarikoptic: I wonder, why to keep any specific event vendoring in the main/staging version and not just replace with "" so too be customized for a given event?

0xdevalias: Minimising the scope of changes for this PR. The better long term fix would be to load this from configure the database, but that’s OOS here.

yarikoptic: +1 on centralizing event specific configuration (later) and minimizing the score of the pr. 
What will need to be done for ohbm (in env/ohbm ?) to not trigger any eventbrite related functionality/calls?

0xdevalias: Nothing, this PR removes everything eventbrite/related except for this static string.

sunny-viktoryia (approved):
Looks good. :upside_down_face:

PR #1401, opened 2021-05-15
cleanup styling for ProfilePictureInput + 'upload your profile pic'

soichih:
I've moved profilepictureinput css to profilepictureinput component with minor style adjustment. This replaces the initial PR I made (#1398).

image


src/pages/Account/Account.scss
```diff
@@ -163,52 +163,6 @@ $page-max-width: 1240px;
...
-  }
-}
-
-.profile-picture-upload-form {
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  flex-direction: column;
-  margin-top: 10px;
-}
```
0xdevalias, line 210: [note] Moved to `src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss`, appears otherwise unchanged.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
```diff
@@ -8,3 +10,49 @@
     border: 0px solid;
   }
 }
+
+.profile-picture-preview {
```
0xdevalias, line 14: [note] Only appears to be used by this component

src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
```diff
@@ -8,3 +10,49 @@
...
+  width: 80px;
+  height: 80px;
+  border: 3px solid $white;
+  margin-bottom: -10px;
+}
+
+.profile-picture-preview-container {
```
0xdevalias, line 21: [note] Only appears to be used by this component

src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
```diff
@@ -8,3 +10,49 @@
...
+  display: flex;
+  justify-content: center;
+  z-index: z(account-profile-picture-preview);
+}
+
+.profile-picture-input {
```
0xdevalias, line 27: [note] Only appears to be used by this component

src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
```diff
@@ -8,3 +10,49 @@
...
+  display: none;
+}
+
+.profile-picture-button {
```
0xdevalias, line 31: [note] Only appears to be used by this component

src/components/molecules/ProfilePictureInput/ProfilePictureInput.scss
```diff
@@ -8,3 +10,49 @@
...
+    background-image: linear-gradient(
+      0deg,
+      rgba(0, 0, 0, 0.6) 0%,
+      rgba(0, 0, 0, 0.6) 100%
+    );
+    background-color: darken($primary, 20%);
+  }
+}
+
+.profile-picture-upload-form {
```
0xdevalias, line 52: [note] Only appears to be used by this component

0xdevalias (changes requested):
Hey @soichih,

Thanks for moving these styles into the appropriate scss file. I expect they were probably leftover in `src/pages/Account/Account.scss` as a legacy from a former refactor to split these components.

---

If you're open to it, it would be awesome if you could also rename these classes to follow our new BEM naming styles. You can read more about BEM here:

- http://getbem.com/naming/

This allows us to keep the styles 'namespaced' in a way that we can be sure they aren't inadvertently used/duplicated across the platform in unexpected places/ways.

At a very high level, the way we tend to use this is by naming the 'block' the same as the component, and then using the `__element` style for each of 'inner elements'.

We make use of the SCSS `&` helper to shorten what we need to type here:

- https://sass-lang.com/documentation/style-rules/parent-selector
- https://css-tricks.com/the-sass-ampersand/

So for this file, the structure would end up looking something like this:

```scss
.ProfilePictureInput {
  // main container styles

  &__preview {
    // styles
  }

  &__preview-container {
    // styles
  }

 // etc
}
```

This would then produce classes such as:

- `.ProfilePictureInput`
- `.ProfilePictureInput__preview`
- `.ProfilePictureInput__preview-container`
- etc

We would then need to update the appropriate locations in the component `.tsx` file to reference these renamed css classes.

Let me know if you have any questions or need any further guidance on this.

mike-lvov (approved):
LGTM

Issue #1402, opened 2021-05-15
Venue Map flickers when it opens 


soichih:
On chrome, when I first open a Venue, all the rooms are collapsed at the top for a second or so before showing the map correctly.

![image](https://user-images.githubusercontent.com/923896/118371889-9c6ea080-b57c-11eb-8820-3411c6e60ffb.png)

yarikoptic:
confirming that I was observing the same

0xdevalias:
This is definitely a common occurrence. I haven’t looked deeply into the specifics of the ‘why’ of this before, but my guess/gut feel is that it may relate to the height that is being set on the PartyMap/Map by default, which is then only changed once it is loaded. If you wanted to dig into it, that would be where I would start looking.

yarikoptic:
I did observe this issue often in the past but have not spotted it recently . I wonder if it is still relevant or was somehow resolved without #1403 which then could be closed along with this one?

mobilevinay:
migrated to zenhub, check!

PR #1403, opened 2021-05-15
Reducing flickering on the PartyMap loading 

soichih:
This fixes https://github.com/sparkletown/sparkle/issues/1402

I've also corrected a bug where `room.title` (not always unique) is used as MapRoom key - switching to array index instead.


0xdevalias:
> I've also corrected a bug where `room.title` (not always unique) is used as MapRoom key - switching to array index instead.

@soichih Array index is even worse, as the index isn’t guaranteed to always uniquely reference the same record; which causes rendering issues.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -203,9 +203,9 @@ export const Map: React.FC<MapProps> = ({
     () =>
       venue?.rooms
         ?.filter(filterEnabledRooms)
-        .map((room) => (
+        .map((room, index) => (
           <MapRoom
-            key={room.title}
+            key={index}
```
0xdevalias, line 208: See comments in https://github.com/sparkletown/sparkle/pull/1403#issuecomment-841886093 about why we never want to use the array index here.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -235,12 +235,13 @@ export const Map: React.FC<MapProps> = ({
           src={venue.mapBackgroundImageUrl ?? DEFAULT_MAP_BACKGROUND}
           alt=""
         />
-
-        <div className="party-map-grid-container" style={gridContainerStyles}>
-          {mapGrid}
-          {partygoersOverlay}
-          {roomOverlay}
-        </div>
+        {totalRows && (
```
0xdevalias, line 238: @soichih Can you please explain the intent behind your logic here please. What was the underlying issue as you see it (at a technical level), and why does this change resolve it?

0xdevalias (dismissed):
Questions/changes.

src/components/templates/PartyMap/components/Map/Map.tsx
soichih: totalRows is initialized to be 0, so this prevents rooms to be drawn until it is reset to non-0 (truthy) value thus eliminating the flicker.

src/components/templates/PartyMap/components/Map/Map.tsx
soichih: Yes, index is not guaranteed to point to the same room if rooms is "removed" from the list and if sparkle is somehow subscribing to that remove event. How often does that really happen?

If you want "correct" way, then you need to guarantee that `room.title` is always unique (not the case right now) or add another id field (overkill, in my opinion) to the collection. I believe using `index` seems to be an acceptable solution given the context. I don't really think performance is an issue here (have you tried benchmarked it?) unless you are expecting hundreds of rooms to be displayed.

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: That makes sense.

I wonder if there is a better way we can fix/improve this 'higher up'?

eg. (musing, these may not necessarily be good ideas)

- initialising with a better default value
- improving the `useEffect` that loads the image just to calculate it's width/height ratio
  - https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/PartyMap/components/Map/Map.tsx#L71-L85
  - Potentially could use `useLayoutEffect` here instead? Though might block rendering too long while loading image?
    - https://reactjs.org/docs/hooks-reference.html#uselayouteffect
- Maybe just treating the whole map as 'loading' until we have the correct `totalRows` value calculated
- etc

---

@mike-lvov What do you think about the solution in this PR + the potential ideas I listed above + any other solutions you can think of for this?

src/components/templates/PartyMap/components/Map/Map.tsx
0xdevalias: An alternative angle to look at this from, is this change required for the main part of your PR fix here?

As this is the existing pattern used across the codebase currently, and since you're only changing one instance of it here (that doesn't seem like it is required to prevent the flickering), then it seems OOS of this PR and would be better raised in a new PR focussed on this (and consistently changed across the entire codebase)

- https://github.com/sparkletown/sparkle/search?q=key%3D%7Broom.title%7D

There are also other variations on this, that combine the `title` with the `index`, which while not perfect, reduces the likelihood of clashes even further (where there would need to be a different room object with the exact same title in the exact same index to cause an issue):

- https://github.com/sparkletown/sparkle/search?q=key%3D%7B%60room-%24%7Broom.title%7D-%24%7Bindex%7D%60%7D
  - ```key={`room-${room.title}-${index}`}```

The `room-` string is superfluous here, so it could just be written as something like:

```
key={`${room.title}-${index}`}
```

This is the general pattern that we use across the codebase in these instances where we don't otherwise have a consistent unique identifier.

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: Anyway, {totalRows &} will result in displaying `0` on the Map, which we want to avoid. moving it to `hasRows` const and using that value to display loading is a better approach IMO

soichih: How about `{(totalRows > 0) & ... }`?  You should avoid creating unnecessary variables / states, etc.

0xdevalias: > You should avoid creating unnecessary variables / states, etc.

Extracting a const with a semantic name has negligible impact on performance/etc, while increasing the semantic readability of the code, so it’s a pattern we often follow.

0xdevalias: > Anyway, {totalRows &} will result in displaying 0 on the Map, which we want to avoid

@mike-lvov Will it? 0 is a falsy value, so I believe it wouldn’t render and be fine (though it’s not as explicit/obvious code at first ‘read’ as we could make it)

mike-lvov: > Will it? 0 is a falsy value, so I believe it wouldn’t render and be fine

It will

0xdevalias: - https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator
  - > Note that returning a falsy expression will still cause the element after && to be skipped but will return the falsy expression. In the example below, `<div>0</div>` will be returned by the render method.

@mike-lvov Seems you're correct.. I didn't realise it worked that way here.. weird implementation choice on React's side of things.

mike-lvov: > How about {(totalRows > 0) & ... }? You should avoid creating unnecessary variables / states, etc.

I like this idea, if you move that to a const as per 
> Extracting a const with a semantic name has negligible impact on performance/etc, while increasing the semantic readability of the code, so it’s a pattern we often follow

dmitrymarchuk: - Reverted back to `room.title`. Gonna make a separate PR where will try to change all occurrences of `index` as a key to `key={`${room.title}-${index}`}` as @0xdevalias suggested.
- extracted `const hasRows = totalRows > 0`
in https://github.com/sparkletown/sparkle/pull/1403/commits/b5dc99e44573b1e11404f11e41121f3105e299c7

mike-lvov (approved):
LGTM

PR #1404, opened 2021-05-16
add venue details information card + showInfo setting

margulies:
This allows the option for each venue to show a left column information card. This is relevent to the env/ohbm educational courses, but I made the PR to staging as it is a general feature. Just let me know if the PR branch should be changed

### DIFF
--- a/src/components/molecules/InformationCard/InformationCard.tsx
+++ b/src/components/molecules/InformationCard/InformationCard.tsx
@@ -4,7 +4,7 @@ import classNames from "classnames";
 import "./InformationCard.scss";
 
 export interface InformationCardProps {
-  title: string;
+  title?: string;
   children: React.ReactNode;
   className?: string;
 }
@@ -15,7 +15,7 @@ const InformationCard: React.FC<InformationCardProps> = ({
   className,
 }) => (
   <div className={classNames("information-card-container", className)}>
-    <h4 className="title">{title}</h4>
+    {title && <h4 className="title">{title}</h4>}
     <div className="information-card-text">{children}</div>
   </div>
 );
--- a/src/components/organisms/WithInformationCard/WithInformationCard.scss
+++ b/src/components/organisms/WithInformationCard/WithInformationCard.scss
@@ -1,0 +1,10 @@
+.information-title {
+  font-size: 22px;
+  font-weight: bold;
+}
+.information-subtitle {
+  font-size: 18px;
+}
+.information-description {
+  font-size: 18px;
+}
--- a/src/components/organisms/WithInformationCard/WithInformationCard.tsx
+++ b/src/components/organisms/WithInformationCard/WithInformationCard.tsx
@@ -1,0 +1,38 @@
+import React from "react";
+
+import { AnyVenue } from "types/venues";
+
+import { InformationLeftColumn } from "components/organisms/InformationLeftColumn";
+import InformationCard from "components/molecules/InformationCard";
+
+import "./WithInformationCard.scss";
+
+interface PropsType {
+  venue: AnyVenue;
+  showInfo?: boolean;
+}
+
+export const WithInformationCard: React.FunctionComponent<PropsType> = ({
+  venue,
+  showInfo = false,
+}) => (
+  <>
+    {showInfo && (
+      <InformationLeftColumn iconNameOrPath={venue?.host?.icon || "info"}>
+        <InformationCard>
+          <p className="information-title">{venue.name}</p>
+          {venue.config?.landingPageConfig.subtitle && (
+            <p className="information-subtitle">
+              {venue.config?.landingPageConfig.subtitle}
+            </p>
+          )}
+          {venue.config?.landingPageConfig.description && (
+            <p className="information-description">
+              {venue.config?.landingPageConfig.description}
+            </p>
+          )}
+        </InformationCard>
+      </InformationLeftColumn>
+    )}
+  </>
+);
--- a/src/pages/VenuePage/TemplateWrapper.tsx
+++ b/src/pages/VenuePage/TemplateWrapper.tsx
@@ -24,6 +24,7 @@ import { ReactionPage } from "components/templates/ReactionPage";
 import { ChatSidebar } from "components/organisms/ChatSidebar";
 import { UserProfileModal } from "components/organisms/UserProfileModal";
 import { WithNavigationBar } from "components/organisms/WithNavigationBar";
+import { WithInformationCard } from "components/organisms/WithInformationCard";
 
 import { AnnouncementMessage } from "components/molecules/AnnouncementMessage";
 
@@ -142,6 +143,7 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
     <ReactionsProvider venueId={venue.id}>
       <WithNavigationBar fullscreen={fullscreen} hasBackButton={hasBackButton}>
         <AnnouncementMessage message={venue.bannerMessage} />
+        <WithInformationCard venue={venue} showInfo={venue?.showInfo} />
         {template}
         <ChatSidebar />
         <UserProfileModal />
--- a/src/types/venues.ts
+++ b/src/types/venues.ts
@@ -145,6 +145,7 @@ export interface BaseVenue {
   hideVideo?: boolean;
   showLiveSchedule?: boolean;
   showGrid?: boolean;
+  showInfo?: boolean;
   roomVisibility?: RoomVisibility;
   rooms?: Room[];
   width: number;
--- a/src/components/organisms/WithInformationCard/index.ts
+++ b/src/components/organisms/WithInformationCard/index.ts
@@ -1,0 +1,1 @@
+export { WithInformationCard } from "./WithInformationCard";


src/components/organisms/WithInformationCard/WithInformationCard.scss
```diff
@@ -0,0 +1,10 @@
+.information-title {
+  font-size: 22px;
+  font-weight: bold;
+}
+.information-subtitle {
+  font-size: 18px;
+}
+.information-description {
+  font-size: 18px;
+}
```
0xdevalias, line 10: Ideally this would follow our latest BEM styling patterns as described in https://github.com/sparkletown/sparkle/pull/1401#pullrequestreview-660507743

- http://getbem.com/naming/

```suggestion
.WithInformationCard {
  &__title {
    font-size: 22px;
    font-weight: bold;
  }
  
  &__subtitle {
    font-size: 18px;
  }

  &__description {
    font-size: 18px;
  }
}
```

0xdevalias: Actually, why are these styles being set here in `WithInformationCard`, yet seemingly designed to override the styles in `src/components/molecules/InformationCard/InformationCard.scss`? Would they not be better set in `InformationCard.scss`?

Ideally the BEM pattern described above would be used directly in `InformationCard.scss`, though there would be couple of places that would need updating across the codebase to use those new names when you do:

- https://github.com/sparkletown/sparkle/search?q=information-card-container
- https://github.com/sparkletown/sparkle/search?q=information-card-text

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+import InformationCard from "components/molecules/InformationCard";
+
+import "./WithInformationCard.scss";
+
+interface PropsType {
+  venue: AnyVenue;
+  showInfo?: boolean;
+}
+
+export const WithInformationCard: React.FunctionComponent<PropsType> = ({
```
0xdevalias, line 15: Export component props + name them as `[COMPONENTNAME]Props`:

```suggestion
export interface WithInformationCardProps {
  venue: AnyVenue;
  showInfo?: boolean;
}

export const WithInformationCard: React.FC<WithInformationCardProps> = ({
```

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+  venue,
+  showInfo = false,
+}) => (
+  <>
+    {showInfo && (
```
0xdevalias, line 20: Why do we pass this prop in to conditionally render here, rather than just conditionally rendering `WithInformationCard` itself in the calling code/parent component?

```suggestion
}) => (
```

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <InformationLeftColumn iconNameOrPath={venue?.host?.icon || "info"}>
```
0xdevalias, line 21: `"info"` appears to already be the default value for this prop, so just passing in a value that evaluates to `undefined` will use the existing default.

(also, as a general note, we use `??` rather than `||` when defining our defaults for nullish variables)

```suggestion
      <InformationLeftColumn iconNameOrPath={venue?.host?.icon}>
```

0xdevalias: Also, in the types for this component (`WithInformationCardProps`) you've said that `venue` must always be defined, therefore there should be no need to be using the `?` syntax here (at least from the root `venue` variable). It's only needed/used when a variable could be `undefined`/etc.

src/components/molecules/InformationCard/InformationCard.tsx
```diff
@@ -15,7 +15,7 @@ const InformationCard: React.FC<InformationCardProps> = ({
   className,
 }) => (
   <div className={classNames("information-card-container", className)}>
-    <h4 className="title">{title}</h4>
+    {title && <h4 className="title">{title}</h4>}
```
0xdevalias, line 18: What's the thoughts/motivation for making the title here optional out of curiosity?

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <InformationCard>
+          <p className="information-title">{venue.name}</p>
```
0xdevalias, line 23: Why are we adding the title here separately, after making it optional in `InformationCard`?

Why not just use the `title` prop on `InformationCard` as intended?

```suggestion
        <InformationCard title={venue.name}>
```

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+          {venue.config?.landingPageConfig.subtitle && (
```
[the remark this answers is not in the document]
0xdevalias: Instead of repeating these long variable paths here directly in the JSX, it would be better to extract them into `const`s in the main body of this component above.

Eg. something like this:

```typescript
const { subtitle, description } = venue.config?.landingPageConfig;
```

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <p className="information-subtitle">
+              {venue.config?.landingPageConfig.subtitle}
+            </p>
```
0xdevalias, line 27: Would it be better to just add a new optional `subtitle` prop to `InformationCard` and pass this value into it?

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+          )}
+          {venue.config?.landingPageConfig.description && (
```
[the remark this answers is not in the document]
0xdevalias: As described above in https://github.com/sparkletown/sparkle/pull/1404/files#r633069252

src/components/organisms/WithInformationCard/WithInformationCard.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
```
0xdevalias, line 1: This file is called `WithInformationCard`, but it seems to be implementing a very specific type of 'venue description' information card. Perhaps it should be named more explicitly for that purpose?

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -142,6 +143,7 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
     <ReactionsProvider venueId={venue.id}>
       <WithNavigationBar fullscreen={fullscreen} hasBackButton={hasBackButton}>
         <AnnouncementMessage message={venue.bannerMessage} />
+        <WithInformationCard venue={venue} showInfo={venue?.showInfo} />
```
0xdevalias, line 146: As mentioned in an earlier comment, it might be better to conditionally render this here directly rather than passing the `showInfo` prop into the component. Also, `venue` is always defined at this point, so there is no reason to use the `?` 'safe accessor' helper:

```suggestion
        {venue.showInfo && <WithInformationCard venue={venue} />}
```

0xdevalias: Also, how will this 'interact' with other templates throughout the app that already explicitly use `InformationLeftColumn`/related wrappers/etc themselves directly? I expect by rendering it here _and_ within those components, the floating elements are going to end up 'over the top of one another':

- https://github.com/sparkletown/sparkle/search?q=InformationLeftColumn
- https://github.com/sparkletown/sparkle/search?q=InformationCard

eg.

- `DonatePopUp`: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/DonatePopUp/DonatePopUp.tsx#L24-L48
  - seemingly only used on `src/components/templates/Playa/Playa.tsx` (legacy template)
- `SparkleFairiesPopUp`: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/SparkleFairiesPopUp/SparkleFairiesPopUp.tsx#L17-L46
  - used on
    - `src/components/templates/PartyMap/PartyMap.tsx`
    - `src/components/templates/ArtPiece/ArtPiece.tsx`
    - `src/components/templates/Playa/Playa.tsx` (legacy template)
- `CreateEditPopUp`: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/CreateEditPopUp/CreateEditPopUp.tsx
  - seemingly only used on `src/components/templates/Playa/Playa.tsx` (legacy template)
- `JazzBarSkeletonPage`: https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Jazzbar/JazzBarSkeletonPage/JazzBarSkeletonPage.tsx#L23-L77
  - used on `src/components/templates/Jazzbar/Jazzbar.tsx`
  - the usage in `JazzBarSkeletonPage` seems to be the same venue name/subtitle/description usage as implemented in this PR + some default fallbacks to legacy text (Kansas Smitty's). I expect we could probably clean up all of this and just use your implementation here.
    - In doing so, it may be possible to remove the `JazzBarSkeletonPage` component entirely, and just move any relevant styles/etc (eg. `jazz-bar-container`) into `Jazzbar` directly.
    - And then once that is done, it makes `src/components/templates/Jazzbar/Jazzbar.tsx` sort of a basically empty/irrelevant wrapper too, so it could probably be removed, and rename `src/components/templates/Jazzbar/JazzTab/JazzTab.tsx` to `Jazzbar` (+ move it up a level so all of it's files end up in `src/components/templates/Jazzbar/` rather than being in `src/components/templates/Jazzbar/JazzTab`)
- `ArtPiece`: https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/ArtPiece/ArtPiece.tsx#L32-L42
  - used in `src/pages/VenuePage/TemplateWrapper.tsx`
  - it also looks like the `default` exports aren't actually used anywhere, so can be removed from `ArtPiece.tsx` + the associated `index.tsx`
- `ConversationSpace`: https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/ConversationSpace/ConversationSpace.tsx#L36-L46
  - used in `src/pages/VenuePage/TemplateWrapper.tsx`
  - it also looks like the `default` exports aren't actually used anywhere, so can be removed from `ConversationSpace.tsx` + the associated `index.tsx`
- etc

src/components/organisms/WithInformationCard/WithInformationCard.tsx
0xdevalias: Maybe `VenueDetailsPopUp` or similar, to match the naming of other similar components?

0xdevalias (changes requested):
Questions/comments/suggestions.

src/types/venues.ts
```diff
@@ -145,6 +145,7 @@ export interface BaseVenue {
   hideVideo?: boolean;
   showLiveSchedule?: boolean;
   showGrid?: boolean;
+  showInfo?: boolean;
```
0xdevalias, line 148: Should we also be implementing the toggle in the admin tooling to be able to enable/disable this setting?

mike-lvov:
Unrelated PR

PR #1408, opened 2021-05-16
autolink URLs in chat messages using getLinkFromText

### LINKED ISSUE #1406: Autodetect and make URLs in the chat hyperlinks
I do expect audience to provide pointers and links to other chat rooms and external websites. ATM URLs in the chat are not presented as hyperlinks, requiring dances to follow them.  I think it would be great UX feature and trivial to implement to convert them (upon rendering) into `<a>` elements.  Smart code should probably discriminate between external URLs and URLs to the event website and open external in a new window with needed security provisions. Or just open always in external since we should encourage people to stay in the chat.

Somewhat relates to discussion in https://github.com/sparkletown/sparkle/pull/1387 for the internal urls.  But otherwise I think it should be safe and very useful UX feature very easy to implement.

yarikoptic:
Thanks @margulies for the pointer to the existing function!

I have first tried to mess around on top of #1394 with ad-hoc function (pushed that mess to https://github.com/yarikoptic/sparkle/commit/0503b269ecfc6278ab776758ef75ab47e1b40991) but then @margulies pointed me to this one and the fix was trivial to implement.  Closes #1406 

Would conflict with #1394 and I am yet to see what is the best way to "merge" the two, and we are yet to decide on how to proceed with #1394.  In my personal preference I think emoji support is more "valuable" but I hope we will not need to choose one over another

edit: ha -- for emojis, adding Emoji instead of fragment right in that getLinkFromText works nicely:

<details>
<summary>the patch</summary> 

```diff
--- a/src/utils/getLinkFromText.tsx
+++ b/src/utils/getLinkFromText.tsx
@@ -1,5 +1,7 @@
 import React from "react";
 
+import Emoji from "react-emoji-render";
+
 import { isExternalUrl, getExtraLinkProps } from "./url";
 
 export const getLinkFromText = (text: string) => {
@@ -24,7 +26,7 @@ export const getLinkFromText = (text: string) => {
         </a>
       );
     } else {
-      return <React.Fragment key={index}>{word} </React.Fragment>;
+      return <Emoji text={word} />;
     }
   });
 };

```
</details>

<details>
<summary>the looks</summary> 

image

</details>
which show the defect -- absent space between words then, so slight tune up will be needed

### DIFF
--- a/src/components/atoms/ChatMessage/ChatMessage.tsx
+++ b/src/components/atoms/ChatMessage/ChatMessage.tsx
@@ -5,6 +5,8 @@ import { faReply } from "@fortawesome/free-solid-svg-icons";
 
 import { MessageToDisplay } from "types/chat";
 
+import { getLinkFromText } from "utils/getLinkFromText";
+
 import { WithId } from "utils/id";
 
 import { useShowHide } from "hooks/useShowHide";
@@ -42,7 +44,7 @@ export const ChatMessage: React.FC<ChatProps> = ({
     () =>
       replies?.map((reply) => (
         <div key={reply.id} className="ChatMessage__reply">
-          {reply.text}
+          {getLinkFromText(reply.text)}
           <ChatMessageInfo
             message={reply}
             deleteMessage={() => deleteMessage(reply.id)}
@@ -68,7 +70,7 @@ export const ChatMessage: React.FC<ChatProps> = ({
     <div className={containerStyles}>
       <div className="ChatMessage__bulb">
         <div className="ChatMessage__text-content">
-          <div className="ChatMessage__text">{text}</div>
+          <div className="ChatMessage__text">{getLinkFromText(text)}</div>
 
           <div className="ChatMessage__reply-icon">
             <FontAwesomeIcon


yarikoptic:
Re PR description enforcer - seems just failed to swallow: enforce
GitHub Actions has encountered an internal error when running your job.

0xdevalias:
> but I hope we will not need to choose one over another

Ultimately we won't have to pick one vs the other. We'll just need to figure the best way to merge the two PRs in a compatible way.

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -9,6 +9,7 @@ import { MessageToDisplay } from "types/chat";
 import { useProfileModalControls } from "hooks/useProfileModalControls";
 
 import { UserAvatar } from "components/atoms/UserAvatar";
+import { getLinkFromText } from "utils/getLinkFromText";
```
0xdevalias, line 12: Re-arranging import order to match our preferred style (types, utils, hooks, components, local `./` imports, etc)

```suggestion
import { getLinkFromText } from "utils/getLinkFromText";

import { useProfileModalControls } from "hooks/useProfileModalControls";

import { UserAvatar } from "components/atoms/UserAvatar";
```

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -36,7 +37,7 @@ export const ChatMessage: React.FC<ChatProps> = ({
 
   return (
     <div className={containerStyles}>
-      <div className="chat-message__text">{text}</div>
+      <div className="chat-message__text">{getLinkFromText(text)}</div>
```
0xdevalias, line 40: See my thoughts/concerns on the potential security rammifications of exposing this functionality to users (as opposed to just admins/venue owners) without further vetting/checking our current implementation in: 

> >  But otherwise I think it should be safe and very useful UX feature very easy to implement.
> 
> We need to be careful about how this gets implemented, as there are many ways that links/URLs can be turned into a security vulnerability (eg. linking to `javascript:alert('do something malicious here')`)
> 
> ---
> 
> > possible solutions are in stackoverflow.com/questions/33235890/react-replace-links-in-a-text . Haven't spotted any react- library mentioned there already being used but didn't look too hard
> 
> We actually already have this functionality within the `AnnouncementMessage` component (used for the banners displayed at the top screen), using `getLinkFromText` from `src/utils/getLinkFromText.tsx`.
> 
> The 'risk' of that implementation is limited to venue owners/admins at the moment, so there is less scope for a malicious user to be able to take advantage of it. I would want to assess the implementation with more scrutiny before exposing that functionality to any guest that can use the chat feature.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1406#issuecomment-841969107_

0xdevalias: @mike-lvov FYI: Need to check through the implementation of `getLinkFromText` from a security-mindset and ensure it's not going to allow any security vulnerabilities.

0xdevalias (commented):
See thoughts/potential concerns in review comments below. Aside from those, the changes here look good. Once we make a decision/assess the potential risk ,this PR is probably good to merge as is (assuming we don't identify any other vulnerabilities in the current implementation of `getLinkFromText` to fix)

0xdevalias (changes requested):
Marking as 'request changes' to block merging until we finalise the aspects noted in https://github.com/sparkletown/sparkle/pull/1408#pullrequestreview-660539167

yarikoptic:
FTR: rebased against fresh staging with threaded chat support.

src/components/atoms/ChatMessage/ChatMessage.tsx
mike-lvov: > @mike-lvov FYI: Need to check through the implementation of getLinkFromText from a security-mindset and ensure it's not going to allow any security vulnerabilities.

As a solution, we can allow to only render links in messages from the `admins`, but I haven't thought about the implementation for that + if it will be that useful, if it's only admins who can post links

0xdevalias: @mike-lvov I think it'd probably be better UX if everyone was able to have their links/etc. I don't think it would take all that long to assess it from a security standpoint. Just need to dedicate the time for it. Would probably take less time to do that than to add the 'only do it for admins' feature I would expect.

yarikoptic:
overall I think this particular PR is superseded by #1383, so if that one is merged, this could be closed

yarikoptic:
I will just close this one -- it should be superseded by #1383 (markdown). We know where to find it if #1383 doesn't land

0xdevalias:
> I will just close this one -- it should be superseded by #1383 (markdown). We know where to find it if #1383 doesn't land

@yarikoptic Makes sense, sounds good.

PR #1409, opened 2021-05-16
add support for custom default avatars

denisdimitrov:
Custom avatars are now fetched and used if they exist in the database. Otherwise the component fallbacks to the default ones.
This is a partial implementation of the whole feature. As already discussed with Sofi, the admin part of this is (optional) for now, but will be required in the near future.

Partially fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/673

### DIFF
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -42,6 +42,7 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
 
+  // @debt Replace fetchSovereignVenueId with useSovereignVenueId, when the hook is refactored to accept venueId as a param.
   const {
     value: sovereignVenueId,
     loading: isLoadingSovereignVenueId,
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -14,13 +14,12 @@ import {
 
 import { resizeFile } from "utils/image";
 
-import { useQuery } from "hooks/useQuery";
-
 import "./ProfilePictureInput.scss";
 
 type Reference = ReturnType<FirebaseStorage["ref"]>;
 
 interface ProfilePictureInputProps {
+  venueId: string;
   setValue: (inputName: string, value: string, rerender: boolean) => void;
   user: UserInfo;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -31,6 +30,7 @@ interface ProfilePictureInputProps {
 }
 
 const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = ({
+  venueId,
   setValue,
   user,
   errors,
@@ -41,8 +41,6 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
   const [error, setError] = useState("");
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
-  const queryParams = useQuery();
-  const venueId = queryParams.get("venueId") ?? "";
 
   const {
     value: sovereignVenueId,
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -1,5 +1,6 @@
 import React, { useRef, useState } from "react";
 import { useFirebase } from "react-redux-firebase";
+import { useAsync } from "react-use";
 import { UserInfo } from "firebase/app";
 import { FirebaseStorage } from "@firebase/storage-types";
 
@@ -11,10 +12,11 @@ import {
 import { resizeFile } from "utils/image";
 
 import "./ProfilePictureInput.scss";
+import { useQuery } from "hooks/useQuery";
 
 type Reference = ReturnType<FirebaseStorage["ref"]>;
 
-interface PropsType {
+interface ProfilePictureInputProps {
   setValue: (inputName: string, value: string, rerender: boolean) => void;
   user: UserInfo;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -25,13 +27,13 @@ interface PropsType {
 }
 
 const sparkleAvatars = [
-  "default-profile-pic-1.png",
-  "default-profile-pic-2.png",
-  "default-profile-pic-3.png",
-  "default-profile-pic-4.png",
+  "/avatars/default-profile-pic-1.png",
+  "/avatars/default-profile-pic-2.png",
+  "/avatars/default-profile-pic-3.png",
+  "/avatars/default-profile-pic-4.png",
 ];
 
-const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
+const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = ({
   setValue,
   user,
   errors,
@@ -42,6 +44,16 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
   const [error, setError] = useState("");
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
+  const queryParams = useQuery();
+  const venueId = queryParams.get("venueId");
+
+  const customAvatars = useAsync(async () => {
+    const storageRef = firebase.storage().ref();
+    const list = await storageRef.child(`/assets/avatars/${venueId}`).listAll();
+    const promises = list.items.map((item) => item.getDownloadURL());
+    const avatars: string[] = await Promise.all(promises);
+    return avatars;
+  }, []);
 
   const uploadPicture = async (profilePictureRef: Reference, file: File) => {
     setIsPictureUploading(true);
@@ -76,8 +88,14 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
   };
 
   const uploadDefaultAvatar = async (avatar: string) => {
-    setValue("pictureUrl", process.env.PUBLIC_URL + `/avatars/${avatar}`, true);
+    setValue("pictureUrl", avatar, true);
   };
+
+  const isLoadingAvatars = customAvatars.loading;
+
+  const defaultAvatars = customAvatars?.value?.length
+    ? customAvatars.value
+    : sparkleAvatars;
 
   return (
     <div className="profile-picture-upload-form">
@@ -110,21 +128,23 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
       {error && <small>Error uploading: {error}</small>}
       <small>Or pick one from our Sparkle profile pics</small>
       <div className="default-avatars-container">
-        {sparkleAvatars.map((avatar, index) => {
-          return (
-            <div
-              key={`${avatar}-${index}`}
-              className="profile-picture-preview-container"
-              onClick={() => uploadDefaultAvatar(avatar)}
-            >
-              <img
-                src={`/avatars/${avatar}`}
-                className="profile-icon profile-picture-preview"
-                alt="your profile"
-              />
-            </div>
-          );
-        })}
+        {isLoadingAvatars && <div>Loading...</div>}
+        {!isLoadingAvatars &&
+          defaultAvatars.map((avatar, index) => {
+            return (
+              <div
+                key={`${avatar}-${index}`}
+                className="profile-picture-preview-container"
+                onClick={() => uploadDefaultAvatar(avatar)}
+              >
+                <img
+                  src={`${avatar}`}
+                  className="profile-icon profile-picture-preview"
+                  alt="your profile"
+                />
+              </div>
+            );
+          })}
       </div>
       <input
         name="pictureUrl"
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -102,7 +102,9 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
     [setValue]
   );
 
-  const isLoading = isLoadingCustomAvatars || isLoadingSovereignVenueId;
+  const isLoading =
+    (isLoadingSovereignVenueId || isLoadingCustomAvatars) &&
+    (customAvatars !== undefined || error !== undefined);
 
   const defaultAvatars = customAvatars?.length
     ? customAvatars
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -1,18 +1,22 @@
-import React, { useRef, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 import { useFirebase } from "react-redux-firebase";
 import { useAsync } from "react-use";
 import { UserInfo } from "firebase/app";
 import { FirebaseStorage } from "@firebase/storage-types";
 
+import { fetchSovereignVenueId } from "api/sovereignVenue";
+
 import {
   ACCEPTED_IMAGE_TYPES,
+  DEFAULT_AVATARS,
   MAX_AVATAR_IMAGE_FILE_SIZE_BYTES,
 } from "settings";
 
 import { resizeFile } from "utils/image";
 
-import "./ProfilePictureInput.scss";
 import { useQuery } from "hooks/useQuery";
+
+import "./ProfilePictureInput.scss";
 
 type Reference = ReturnType<FirebaseStorage["ref"]>;
 
@@ -26,13 +30,6 @@ interface ProfilePictureInputProps {
   register: any;
 }
 
-const sparkleAvatars = [
-  "/avatars/default-profile-pic-1.png",
-  "/avatars/default-profile-pic-2.png",
-  "/avatars/default-profile-pic-3.png",
-  "/avatars/default-profile-pic-4.png",
-];
-
 const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = ({
   setValue,
   user,
@@ -45,15 +42,26 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
   const queryParams = useQuery();
-  const venueId = queryParams.get("venueId");
+  const venueId = queryParams.get("venueId") ?? "";
 
-  const customAvatars = useAsync(async () => {
+  const {
+    value: sovereignVenueId,
+    loading: isLoadingSovereignVenueId,
+  } = useAsync(async () => await fetchSovereignVenueId(venueId));
+
+  const {
+    value: customAvatars,
+    loading: isLoadingCustomAvatars,
+  } = useAsync(async () => {
     const storageRef = firebase.storage().ref();
-    const list = await storageRef.child(`/assets/avatars/${venueId}`).listAll();
-    const promises = list.items.map((item) => item.getDownloadURL());
-    const avatars: string[] = await Promise.all(promises);
+    const list = await storageRef
+      .child(`/assets/avatars/${sovereignVenueId}`)
+      .listAll();
+    const avatars: string[] = await Promise.all(
+      list.items.map((item) => item.getDownloadURL())
+    );
     return avatars;
-  }, []);
+  }, [sovereignVenueId]);
 
   const uploadPicture = async (profilePictureRef: Reference, file: File) => {
     setIsPictureUploading(true);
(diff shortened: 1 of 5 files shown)


src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -42,6 +44,16 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
   const [error, setError] = useState("");
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
+  const queryParams = useQuery();
+  const venueId = queryParams.get("venueId");
```
0xdevalias, line 48: Instead of reading from URL params this deep in the code, how much effort would be required to pass the `venueId` as a prop?

Does this actually need to be the `venueId` here? It would make more sense to use the `sovereignVenueId` I would think? As that correlates closer to the concepts in the new world architecture.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -42,6 +44,16 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
...
+
+  const customAvatars = useAsync(async () => {
+    const storageRef = firebase.storage().ref();
+    const list = await storageRef.child(`/assets/avatars/${venueId}`).listAll();
+    const promises = list.items.map((item) => item.getDownloadURL());
+    const avatars: string[] = await Promise.all(promises);
+    return avatars;
+  }, []);
```
0xdevalias, line 56: Would this be better wrapped up in a custom hook of sorts? And/or in the `api/*` layer?

What storage security permissions are required to list all of the urls at that `storageRef` location? For this to work like this, it would basically need to be queryable by any user and/or potentially public?

Is `getDownloadURL` going to create/add a new key every time it's called, or will this always return the same URL?

Why wouldn't we just have an array of strings in the venue config itself or similar, rather than connecting directly to the storage like this?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -76,9 +88,15 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
   };
 
   const uploadDefaultAvatar = async (avatar: string) => {
-    setValue("pictureUrl", process.env.PUBLIC_URL + `/avatars/${avatar}`, true);
+    setValue("pictureUrl", avatar, true);
   };
```
0xdevalias, line 92: Why getting rid of `process.env.PUBLIC_URL` here? Presumably that would still be relevant to include in the `sparkleAvatars` array above?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -76,9 +88,15 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
...
 
+  const isLoadingAvatars = customAvatars.loading;
```
0xdevalias, line 94: You can extract this directly inline from the `useAsync` call:

```typescript
const { loading: isLoadingAvatars, value, etc } = useAsync(
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -110,21 +128,23 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
...
-            >
-              <img
-                src={`/avatars/${avatar}`}
-                className="profile-icon profile-picture-preview"
-                alt="your profile"
-              />
-            </div>
-          );
-        })}
+        {isLoadingAvatars && <div>Loading...</div>}
```
0xdevalias, line 131: Do we want to mark just this part as loading, or the whole component? Usually we have done the whole component from what I have seen?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -110,21 +128,23 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
...
+        {!isLoadingAvatars &&
+          defaultAvatars.map((avatar, index) => {
```
0xdevalias, line 133: Extract as a const + `useMemo`

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -11,10 +12,11 @@ import {
 import { resizeFile } from "utils/image";
 
 import "./ProfilePictureInput.scss";
+import { useQuery } from "hooks/useQuery";
```
0xdevalias, line 15: Import ordering

```suggestion
import { resizeFile } from "utils/image";

import { useQuery } from "hooks/useQuery";

import "./ProfilePictureInput.scss";
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -25,13 +27,13 @@ interface PropsType {
...
 const sparkleAvatars = [
-  "default-profile-pic-1.png",
-  "default-profile-pic-2.png",
-  "default-profile-pic-3.png",
-  "default-profile-pic-4.png",
+  "/avatars/default-profile-pic-1.png",
+  "/avatars/default-profile-pic-2.png",
+  "/avatars/default-profile-pic-3.png",
+  "/avatars/default-profile-pic-4.png",
 ];
```
0xdevalias, line 34: This seems like it should probably be extracted into `src/settings` and renamed something like `DEFAULT_AVATARS`/`DEFAULT_AVATAR_URLS` or similar.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -42,6 +44,16 @@ const ProfilePictureInput: React.FunctionComponent<PropsType> = ({
...
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
+  const queryParams = useQuery();
+  const venueId = queryParams.get("venueId");
+
+  const customAvatars = useAsync(async () => {
+    const storageRef = firebase.storage().ref();
+    const list = await storageRef.child(`/assets/avatars/${venueId}`).listAll();
+    const promises = list.items.map((item) => item.getDownloadURL());
+    const avatars: string[] = await Promise.all(promises);
```
0xdevalias, line 54: May as well just inline these:

```suggestion
    const avatars: string[] = await Promise.all(list.items.map((item) => item.getDownloadURL()));
```

0xdevalias (changes requested):
Questions/comments/changes.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: > Does this actually need to be the venueId here? It would make more sense to use the sovereignVenueId I would think? As that correlates closer to the concepts in the new world architecture.

I'm not 100% sure on this one. `sovereignVenueId` is probably a better way, because the avatars can be set only for the parent venue and all of it's children will get the avatars from there.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: > Would this be better wrapped up in a custom hook of sorts? And/or in the api/* layer?

This is a very specific implementation and chances are it will never be reused because of this concept for default avatars.

> What storage security permissions are required to list all of the urls at that storageRef location? For this to work like this, it would basically need to be queryable by any user and/or potentially public?

This needs to be public because unregistered users will be fetching the default avatars.

> Is getDownloadURL going to create/add a new key every time it's called, or will this always return the same URL?

It seems to be returning the same URL every time.

> Why wouldn't we just have an array of strings in the venue config itself or similar, rather than connecting directly to the storage like this?

Even if we add this array of strings in the config, they'd still need to be uploaded in the storage or somewhere else. The only difference is that we wouldn't be fetching them directly. However, this means that in the admin part we'd have to upload them and then update the venue config. Where now, we only need to upload them.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: The default avatars are in the public/assets location and there is no need for this `process.env.PUBLIC_URL` The avatars are still included even without this, I double checked.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: On on side, I can load the whole component and keep it consistent. On the other, whilst the avatars are loading, the user can still type his display name and see the rest of the content on the screen. My idea was not to block his whole page just because the avatars are loading.

0xdevalias: Mm, makes sense. Let's not block the whole page.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: > and there is no need for this `process.env.PUBLIC_URL`

That's only because we currently build the app to be hosted at `/`, so it works as a fluke, not because it's correct. Using `process.env.PUBLIC_URL` is the correct way for this to be done:

- https://create-react-app.dev/docs/using-the-public-folder/#adding-assets-outside-of-the-module-system

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: > This is a very specific implementation and chances are it will never be reused because of this concept for default avatars.

This was less about reuse and more about keeping heavier logic/etc outside of component files to aid in testing/etc.

> This needs to be public because unregistered users will be fetching the default avatars.

The storage collection itself doesn't need to be public/publicly listble/etc. Download URLs contain a 'baked in' key that makes their contents accessible. That's the only part of this that needs to be public I believe.

> It seems to be returning the same URL every time.

Cool

> Even if we add this array of strings in the config, they'd still need to be uploaded in the storage or somewhere else. The only difference is that we wouldn't be fetching them directly. However, this means that in the admin part we'd have to upload them and then update the venue config. Where now, we only need to upload them.

Makes sense. I don't see it being a huge issue that we would need to upload them then update venue config or similar. Though my stance on this is mostly going to come back to the answers on the 'publicly listable/accessible' stuff I mentioned above.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: Cool, let's go with that then. Which means we can use `useSovereignVenueId` (which seemingly currently just reads from `useVenueId` anyway.. _sigh_ :p)

denisdimitrov: `sovereignVenueId` is now fetched and used in https://github.com/sparkletown/sparkle/pull/1409/commits/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe however, the defualt avatars are visible for a split second between the 2 async requests. Any idea on how to fix this? Maybe use just one state for loading and change it manually on `.finally()`

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Changed in https://github.com/sparkletown/sparkle/commit/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Changed in: https://github.com/sparkletown/sparkle/commit/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Changed in: https://github.com/sparkletown/sparkle/commit/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Changed in: https://github.com/sparkletown/sparkle/commit/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Changed in: https://github.com/sparkletown/sparkle/commit/6cdbdb2d81f61c67a0665f91bf358d09fc89bfbe

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: Looks like you're using `fetchSovereignVenueId` (the api layer) in that commit, whereas I said to use `useSovereignVenueId` (the hook that wraps up this functionality and is what we should be using from the component layer) from `src/hooks/useSovereignVenueId.ts`

It stores state in redux, so you may never even need to wait on it, but on the off chance that you do, it returns `isSovereignVenueIdLoading`

That should be combined into a single loading state for both (which it seems you're already basically doing with `isLoading`)

The `avatarImages` `useMemo` should probably have a guard that just returns an empty array while `isLoading` is true, since there's no point to processing/rendering it if we're still loading.

Then in the JSX since it's such a small amount of code now, it probably makes sense to use a ternary.

```jsx
{isLoading ? <div>Loading...</div> : avatarImages}
```

---

> however, the defualt avatars are visible for a split second between the 2 async requests

Is this because `useAsync` doesn't actually 'fire' until the first load/display? If so, we could potentially look at it's return value (assuming it is `undefined` or similar before the hook has run?), and include that in the `isLoading` variable?

Something like:

```typescript
const isLoading = (isLoadingSovereignVenueId || isLoadingCustomAvatars) && customAvatars !== undefined;
```

denisdimitrov: > Looks like you're using fetchSovereignVenueId (the api layer) in that commit, whereas I said to use useSovereignVenueId (the hook that wraps up this functionality and is what we should be using from the component layer) from src/hooks/useSovereignVenueId.ts

`useSovereignVenueId` is only working with params, where in my case the url has search query. It returns `undefined` and that's why I had to use `fetchSovereignVenueId`

denisdimitrov: I fixed the loading issue in: https://github.com/sparkletown/sparkle/pull/1409/commits/d9d670ae3e2af69dca89fb490b8b02209ae9baa4

0xdevalias: > `useSovereignVenueId` is only working with params, where in my case the url has search query. It returns `undefined` and that's why I had to use `fetchSovereignVenueId`

Makes sense. That is changed in https://github.com/sparkletown/sparkle/pull/1368 (but not merged yet)

- https://github.com/sparkletown/sparkle/pull/1368/files?file-filters%5B%5D=.ts&hide-deleted-files=true#diff-972c4ca9a9c40bfadf97da41744c77a76d472da0c96d80ed230fd2b216f8057e

0xdevalias: > I fixed the loading issue in: [d9d670a](https://github.com/sparkletown/sparkle/commit/d9d670ae3e2af69dca89fb490b8b02209ae9baa4)

Sweet, if it works, go with that. But also probably add an `@debt` comment saying to use `useSovereignVenueId` once it's refactored to accept `venueId` as a param.

denisdimitrov: @0xdevalias I will add the @debt in a moment. Meanwhile, can you please review this one final time because Sofi needs this in like an hour.

denisdimitrov: `@debt` has been added in: https://github.com/sparkletown/sparkle/pull/1409/commits/82c66b6da26694ba0cf53c43496e1d22742033b9

I had to commit some additional changes, `venueId` is now passed as a prop because this component is used in multiple places and `useVenueId()` and `useQuery()` are both not suitable. Sometimes the url has search query, sometimes it's param. https://github.com/sparkletown/sparkle/pull/1409/commits/843431ffa592ffb19a91978deb4fec4425982c17

0xdevalias: Makes sense.

Yeah, I can review it in a sec.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -44,16 +41,26 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
...
-    const avatars: string[] = await Promise.all(promises);
+    const list = await storageRef
+      .child(`/assets/avatars/${sovereignVenueId}`)
+      .listAll();
+    const avatars: string[] = await Promise.all(
+      list.items.map((item) => item.getDownloadURL())
+    );
     return avatars;
-  }, []);
+  }, [sovereignVenueId]);
```
0xdevalias: [NB] Still not sure if this pattern of listing/reading the storage like this is a good idea.. I'm surprised it even works without changing our security rules.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -87,15 +94,38 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
...
+  const defaultAvatars = customAvatars?.length
+    ? customAvatars
+    : DEFAULT_AVATARS;
 
-  const defaultAvatars = customAvatars?.value?.length
-    ? customAvatars.value
-    : sparkleAvatars;
+  const avatarImages = useMemo(() => {
+    return defaultAvatars.map((avatar, index) => {
+      return (
```
0xdevalias, line 114: [nit] 
```suggestion
    return defaultAvatars.map((avatar, index) => (
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -87,15 +94,38 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
...
+        <div
+          key={`${avatar}-${index}`}
+          className="profile-picture-preview-container"
+          onClick={() => uploadDefaultAvatar(avatar)}
+        >
+          <img
+            src={`${avatar}`}
```
0xdevalias, line 121:
```suggestion
            src={avatar}
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -87,15 +94,38 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
...
+            className="profile-icon profile-picture-preview"
+            alt="your profile"
```
0xdevalias, line 123:
```suggestion
            alt={`default avatar ${index}`}
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -128,23 +158,8 @@ const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputProps> = (
...
-                <img
-                  src={`${avatar}`}
-                  className="profile-icon profile-picture-preview"
-                  alt="your profile"
-                />
-              </div>
-            );
-          })}
+        {isLoading && <div>Loading...</div>}
+        {!isLoading && avatarImages}
```
0xdevalias, line 162:
```suggestion
        {isLoading ? <div>Loading...</div> : avatarImages}
```

src/pages/Account/Profile.tsx
```diff
@@ -1,16 +1,24 @@
...
 import { useVenueId } from "hooks/useVenueId";
+import { useUser } from "hooks/useUser";
+
+import getQueryParameters from "utils/getQueryParameters";
+
+import { RouterLocation } from "types/RouterLocation";
+
+import { updateUserProfile } from "./helpers";
+
+import ProfilePictureInput from "components/molecules/ProfilePictureInput";
```
0xdevalias, line 19: [nit] secrets, settings, types, utils, hooks, components, ./

0xdevalias (approved):
[CRO] Based on the diff since my last review, couple of tiny things. I haven't gone through and checked that all of my previous review comments were addressed properly, so best to make sure that is done before merging. If it is, and it works with clickthrough and such, then LGTM!

(pre-emptively approving based on the above caveats)

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: There is a read rule which allows us to read `assets/`

denisdimitrov: You can take a look at `storage.rules`

denisdimitrov: For the admin we'd have to add write and delete with some conditions ofc.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: [OOS] Do we have some kind of spinner component in the project?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1409/commits/041df52e62699a3a610407e6d105970072a53ce2

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1409/commits/041df52e62699a3a610407e6d105970072a53ce2

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1409/commits/041df52e62699a3a610407e6d105970072a53ce2

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1409/commits/041df52e62699a3a610407e6d105970072a53ce2

src/pages/Account/Profile.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1409/commits/041df52e62699a3a610407e6d105970072a53ce2

0xdevalias (approved):
[CRO] Changes since last review look good.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: Mm, fair. I didn't realise that would also let us list assets/etc too. Makes sense though.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: > [OOS] Do we have some kind of spinner component in the project?

Not that I remember seeing.. closest is possibly the sparkle animation.. but that's not ideal for this use case.

PR #1411, opened 2021-05-17
Added pdf uploader for posterpage template

soichih:
This change adds the capability to upload a pdf file by the poster admin.

image

On firebase / storage, I've added the following rule

```
    match /posters/{allPaths=**} {
        allow read
        allow write: if request.auth != null;
        allow update: if resource.metadata.owners.matches(".*"+request.auth.uid+".*");
    }
``` 

If poster is already uploaded, it does the auth check using the `customMetadata.owners` as set by the UI code, but if the file doesn't exist, I couldn't figure out a good way to check if the user is *supposed to be* able to upload the poster or not. I think this can be mitigated by pre-uploading all posters before hand (I have a backend script for that). And update can be done if the user is listed in the owners list.

### DIFF
--- a/src/components/organisms/PosterPageAdmin/PosterPageAdmin.scss
+++ b/src/components/organisms/PosterPageAdmin/PosterPageAdmin.scss
@@ -1,0 +1,25 @@
+.ppa {
+  padding: 10px 20px;
+  background-color: #333;
+  border-radius: 10px;
+  margin-top: 20px;
+
+  .ppa-header {
+    font-weight: bold;
+    line-height: 200%;
+  }
+  .ppa-submit {
+    margin: 10px 0;
+    float: right;
+  }
+  iframe {
+    margin-top: 10px;
+    width: 100%;
+    height: 350px;
+    border: 0;
+  }
+
+  br.clear {
+    clear: both;
+  }
+}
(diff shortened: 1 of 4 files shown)


0xdevalias:
Hey @soichih

It's my understanding (confirmed with @sparkletown/product just now) that we aren't hosting the poster PDFs directly on the sparkle environment's storage, but that we would just be storing the links to them (provided by OHBM) on another hosting service handled by OHBM. So it seems the feature implemented within this PR (specifically around uploading/etc) is out of scope/not relevant for that use case?

0xdevalias (changes requested):
Hey @soichih

It's my understanding (confirmed with @sparkletown/product just now) that we aren't hosting the poster PDFs directly on the sparkle environment's storage, but that we would just be storing the links to them (provided by OHBM) on another hosting service handled by OHBM. So it seems the feature implemented within this PR (specifically around uploading/etc) is out of scope/not relevant for that use case?

(note: haven't deeply reviewed the code here yet, marking as changes mostly to block merging/etc)

mike-lvov:
as per https://github.com/sparkletown/sparkle/pull/1411#pullrequestreview-661536353, seems unrelevant

PR #1416, opened 2021-05-18
add keyboard shortcuts to common UI elements (chat, schedule, etc)

### LINKED ISSUE #1413: keyboard shortcuts: are they there or todo?
primary use case to show/hide chat (while it is a floating one). 

I see that some provisions were done already to https://github.com/sparkletown/sparkle/commit/45108f229a94d560233442ec39c7a129e8676cc1 provide support for shortcuts, but I wondered -- are any already there and it is just a matter of documentation (and having some "Help" pane) or it ended up non-trivial to support?

### LINKED ISSUE #1418: InputField: Pressing Escape should unfocus (blur) the input
I have tried my ignorant-react best to implement it but failed.  Especially in the light of shortcuts support (#1416) it would make sense to make InputField (search bar, chat entry) to no longer be focused on if Escape is pressed.  

I have tried to follow https://stackoverflow.com/questions/65740016/react-remove-focus-from-input-element-on-escape-key and 

<details>
<summary>blindly brew this beast</summary> 

```patch
diff --git a/src/components/atoms/InputField/InputField.tsx b/src/components/atoms/InputField/InputField.tsx
index 2333a19b..a4a27a53 100644
--- a/src/components/atoms/InputField/InputField.tsx
+++ b/src/components/atoms/InputField/InputField.tsx
@@ -37,6 +37,14 @@ const renderIcon = (
   );
 };
 
+const onEscape = function (action) {
+    window && window.addEventListener('keydown', (e) => {
+        if (e.key === "Escape") {
+            action();
+        };
+    });
+};
+
 const InputField: React.ForwardRefRenderFunction<
   HTMLInputElement,
   InputFieldProps
@@ -61,6 +69,10 @@ const InputField: React.ForwardRefRenderFunction<
 
   const inputClassNames = classNames("input-field__input", inputClassName);
 
+  const onEscape = () => {
+    ref.current.blur();
+  };
+
   return (
     <div className={containerClassNames}>
       <input ref={ref} className={inputClassNames} {...extraInputProps} />

```
</details>
but it doesn't work.

yarikoptic:
If ever complete, merged Closes #1413 

ATM does just:
- [x] use useMousetrap to make "c" to toggle chat -- tested and it works nicely!
  - [x] odd issue detected.  If go to "Direct messages" and then back to "Party chat" - "c" no longer toggles chat. I wonder if relates to that issue we had for "s" requiring rewrite...
- [x] unfocus input upon Escape. Closes #1418 
  -  ~~Ideally Escape also could be used to collapse Schedule pane~~ The same `S` could be used
- [x] centralize definition of shortcuts.  Should I just rename `MovementKeys` in https://github.com/sparkletown/sparkle/blob/staging/src/hooks/useKeyboardControls.ts#L8-L13 to `ShortcutKeys` and extend? or could be a dedicated `ShortcutKeys` defined in the same file 
- [x] decide what to do (if anything) about copy/pasted TODO comments
- add for other elements:
  - [x] toggle schedule (`S` for "Schedule"?) 
  - [x] focus into search bar (`/` as typically used in many *nixy tools)


yarikoptic:
@0xdevalias any quick idea on how to resolve the "f" key being placed into `input` issue?

yarikoptic:
I have merged @soichih fix for unfocus #1418 and find/search. The only outstanding is schedule

src/components/atoms/InputField/InputField.tsx
```diff
@@ -61,9 +61,20 @@ const InputField: React.ForwardRefRenderFunction<
 
   const inputClassNames = classNames("input-field__input", inputClassName);
 
+  function keyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    if (e.key === "Escape" && e.target) {
+      e.currentTarget.blur();
+    }
+  }
```
0xdevalias, line 68: I wonder if we would also want to use `mousetrap` here for consistency?

Also, I wonder if we should be adding this functionality to our generic `InputField`, as opposed to just within the navbar search/similar 'focussed place' that the feature is desired for.

@mike-lvov Any thoughts?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -140,15 +147,27 @@ const NavBar: React.FC<NavBarPropsType> = ({
   }, [radioFirstPlayStateLoaded]);
 
   const [isEventScheduleVisible, setEventScheduleVisible] = useState(false);
-  const toggleEventSchedule = useCallback(() => {
+  function toggleEventSchedule() {
+    console.log("toggling");
     setEventScheduleVisible(!isEventScheduleVisible);
-  }, [isEventScheduleVisible]);
+  }
```
0xdevalias, line 153: `console.log` will need to be removed from here.

We also use `const foo = () => {` style for our functions, rather than `function foo() {`

Also, the `useCallback` wrapper seems to have been lost here, which is required to be kept to ensure the function maintains a stable reference, rather than being re-created on every render.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -140,15 +147,27 @@ const NavBar: React.FC<NavBarPropsType> = ({
...
   const hideEventSchedule = useCallback((e) => {
     if (e.target.closest(`.${navBarScheduleClassName}`)) return;
-
     setEventScheduleVisible(false);
   }, []);
```
0xdevalias, line 157: Why was the newline removed from here? We keep a blank line after our 'guard statements' like this

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -140,15 +147,27 @@ const NavBar: React.FC<NavBarPropsType> = ({
...
+      toggleEventSchedule();
+    }
+  }
+  useEffect(() => {
+    window.addEventListener("keydown", keyDownHandler);
+    return () => {
+      window.removeEventListener("keydown", keyDownHandler);
+    };
+  });
+
```
0xdevalias, line 170: Why aren't we using `mousetrap` for this as we do with our other keyboard shortcuts?

We also use `const foo = () => {` style for our functions, rather than `function foo() {`

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -141,6 +143,20 @@ const NavSearchBar = () => {
     [searchQuery]
   );
 
+  const focusSearchBar = useCallback((e) => {
+    e.preventDefault();
+    const searchBar = document.getElementById("NavSearchBar__search-input");
+    if (searchBar) searchBar.focus();
+    // TODO: we need to remove "f" from being entered into the input which would happen now
+  }, []);
```
0xdevalias, line 151: Using `document.getElementById` isn't a very Reacty way of doing things. We would likely want to get a `ref` for that element if we were wanting to call `.focus()` on it directly:

- https://reactjs.org/docs/refs-and-the-dom.html
- https://reactjs.org/docs/forwarding-refs.html

0xdevalias: > TODO: we need to remove "f" from being entered into the input which would happen now

This part is interesting, and something I was going to call out in my review to ask how you were handling that case. Not sure of the solution off the top of my head just now.

---

Actually, looking at the docs, it looks like it should handle this case by default:

- https://craig.is/killing/mice#api.bind.text-fields
  - > By default all keyboard events will not fire if you are inside of a `textarea`, `input`, or `select` to prevent undesirable things from happening.
  - > If you want them to fire you can add the class `mousetrap` to the element.
- https://craig.is/killing/mice#api.stopCallback
  - > `Mousetrap.stopCallback` is a method that is called to see if the keyboard event should fire based on if you are inside a form `input` field.

Is that not the behaviour you're observing here?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -161,6 +177,7 @@ const NavSearchBar = () => {
       <InputField
         value={searchInputValue}
         inputClassName="NavSearchBar__search-input"
+        id="NavSearchBar__search-input"
```
0xdevalias, line 180: See my comments above about using a React ref, rather than using `id` + `document.getElementById`.

```suggestion
```

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -141,6 +143,20 @@ const NavSearchBar = () => {
...
+    if (searchBar) searchBar.focus();
+    // TODO: we need to remove "f" from being entered into the input which would happen now
+  }, []);
+
+  useMousetrap({
+    keys: ShortcutKeys.find,
+    callback: focusSearchBar,
+    // TODO: bindRef: (null as never) as MutableRefObject<HTMLElement>,
+    withGlobalBind: true, // TODO: remove this once we have a ref to bind to
+  });
```
0xdevalias, line 158: Given these are intended to be global key bindings, I wonder if we should be setting them up somewhere more globally?

Not sure how we would achieve that in an easy/efficient way though at this stage..

@mike-lvov Any thoughts? I know I pushed back on you including key bindings in a 'deeper' component in the past (though that was related to map keyboard controls that already had a location they were being set/called/etc, so maybe this is a different case)

src/hooks/useKeyboardControls.ts
```diff
@@ -12,6 +12,12 @@ export const MovementKeys = {
   right: ["right", "d"],
 };
 
+export const ShortcutKeys = {
+  chat: ["c"],
+  schedule: ["s"],
+  find: ["f"],
+};
```
0xdevalias, line 19: Instead of adding these to `src/hooks/useKeyboardControls.ts`, it might make sense to move both  `MovementKeys` and `ShortcutKeys` to somewhere like `src/settings`.

They could probably also be combined into the same `const` so that we have a single 'source of truth' for all our shortcuts, something like:

```typescript
export const KeyboardShortcutKeys = {
  chat: "c",
  find: "f",
  movement: {
    up: ["up", "w"],
    down: ["down", "s"],
    left: ["left", "a"],
    right: ["right", "d"],
  },
  schedule: "s",
}
```

It would probably also be good to rename `src/hooks/useKeyboardControls.ts` to `src/hooks/useMapKeyboardControls.ts` or similar to differentiate it (including updating the props/hook to match the filename), since we're adding more shortcuts to the platform now.

0xdevalias (changes requested):
Thanks for the contribution @yarikoptic (+ additions from @soichih)

Have left some thoughts/comments/suggestions in my review below.

---

> Probably NavBar code doesn't run until user clicks on Schedule, so shortcut becomes available only e.g. after schedule is expanded once

@yarikoptic Can you please expand on this a bit more. What is the issue/situation you're observing here? If the `NavBar` is rendered and visible, then any code within it should also have run.

src/components/atoms/InputField/InputField.tsx
yarikoptic: That is a solution @soichih came up with since for some reason mousetrap (as described in TODOs on top, and you asked above) for `s` was not causing any "action", while did just fine for other cases. 
I wonder if it also anyhow relates to fresh issue I observed with current state as of a05339aef51ae3551571ba21eb801118bb79e945: if go to "Direct messages" and then back to "Party chat" - `c` no longer toggles chat until I toggle by mouse or go back to "Direct messages"

src/components/molecules/NavBar/NavBar.tsx
yarikoptic: I have  removed `log`, and converted functions to desired style.  @soichih - I would need your help on the `useCallback` part since I believe it was part of a solution to get schedule toggable by shortcut

src/components/molecules/NavBar/NavBar.tsx
yarikoptic: readded, sorry ;-)

src/components/molecules/NavBar/NavBar.tsx
yarikoptic: function invocation was fixed. re `mousetrap` - didn't work (as e.g. of ea7c9ae3f4e978c5551ab7ec49c9a0d802b45451)

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: I don't think so -- the issue wasn't  about being in the input field already to handle an event but to prevent input field from getting that initial `f`.  @soichih solved it in b75e4d6fb97b0695c57eee8a7107004807aba9f5 by 
```
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -144,6 +144,7 @@ const NavSearchBar = () => {
   );
 
   const focusSearchBar = useCallback((e) => {
+    e.preventDefault();
```

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: well, I guess I could define ```useMousetrapGlobal``` as an adapter over to useMousetrap setting `withGlobalBind: true`. is that you are hinting on @0xdevalias ?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: > Using `document.getElementById` isn't a very Reacty way of doing things. We would likely want to get a `ref` for that element if we were wanting to call `.focus()` on it directly:

ignorant in React me tried and failed... I might give it another shot later on

src/components/molecules/NavBar/NavBar.tsx
yarikoptic: but now `s` while in Search bar focused still causes schedule to toggle @soichih (may be references [below](https://github.com/sparkletown/sparkle/pull/1416#discussion_r635777973) would be of help). It still enters it so no "harm is done" but while focused anywhere  `s` should not toggle schedule

soichih: duh! that's emberassing. I will fix it tonight.

src/components/atoms/InputField/InputField.tsx
0xdevalias: Does it relate to the fact that ‘S’ is already bound to movement in usekeyboardControls()?

0xdevalias: I would have to look at the code to be sure, but guessing the ‘no longer toggles’ is possibly going to relate to having the wrong function reference bound to mousetrap or similar, that only sets the state a certain way or similar (not on a computer to check at the moment)

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: Nah, was more thinking in the same way that the movement controls are all wrapped up in a single hook (useKeyboardControls) currently, can we centralise everything else in a similar way.

I feel like the answer is likely no/doesn’t make sense to; so I expect that centralising the shortcut definitions themselves in src/settings (as described elsewhere on my review comments) is probably the best case scenario

src/hooks/useKeyboardControls.ts
0xdevalias: For these minimal changes, yeah, more than happy for you to tweak the existing code for it.

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: Mentioned this in another comment somewhere, but ‘s’ is already bound is mousetrap for movement (see useKeyboardKeys), so that’s probably why you didn’t see it work?

yarikoptic: Ohh oh, my 4 eyes weren't enough and I was blind! I bet it is the reason. Thanks! I will check/RF back if so, to use / ;-)

src/hooks/useKeyboardControls.ts
yarikoptic: done, but made it KEYBOARD_SHORTCUTS (to stay inline with upper casing in `settings.ts`) and placed the last in the file since could not detect any order to follow. also renamed that adding `Map` into `useMapKeyboardControl*`

src/hooks/useMapKeyboardControls.ts
```diff
@@ -23,13 +18,10 @@ interface UseKeyboardControlsProps {
...
-  venueId,
-  totalRows,
-  totalColumns,
-  isSeatTaken,
-  takeSeat,
-}) => {
+export const useMapKeyboardControls: ReactHook<
+  UseMapKeyboardControlsProps,
+  void
+> = ({ venueId, totalRows, totalColumns, isSeatTaken, takeSeat }) => {
```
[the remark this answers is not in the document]
yarikoptic: not my fault

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: <details>
<summary>here is what I have tried</summary> 

```patch
diff --git a/src/components/molecules/NavSearchBar/NavSearchBar.tsx b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
index 674f3dc4..68ff5f25 100644
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -58,6 +58,7 @@ const NavSearchBar = () => {
   const { worldUsers } = useWorldUsers();
 
   const { openUserProfileModal } = useProfileModalControls();
+  const searchBar = React.useRef(null);
 
   const foundRooms = useMemo<JSX.Element[]>(() => {
     if (!searchQuery) return [];
@@ -148,8 +149,7 @@ const NavSearchBar = () => {
 
   const focusSearchBar = useCallback((e) => {
     e.preventDefault();
-    const searchBar = document.getElementById("NavSearchBar__search-input");
-    if (searchBar) searchBar.focus();
+    this.searchBar.current.focus();
   }, []);
 
   useMousetrap({
@@ -179,7 +179,7 @@ const NavSearchBar = () => {
       <InputField
         value={searchInputValue}
         inputClassName="NavSearchBar__search-input"
-        id="NavSearchBar__search-input"
+        ref={searchBar}
         onChange={onSearchInputChange}
         placeholder="Search for people, rooms, events..."
         autoComplete="off"
```
</details>
I get
```
/home/yoh/proj/misc/sparkle/sparkle/src/components/molecules/NavSearchBar/NavSearchBar.tsx
TypeScript error in /home/yoh/proj/misc/sparkle/sparkle/src/components/molecules/NavSearchBar/NavSearchBar.tsx(152,5):
Object is possibly 'undefined'.  TS2532

    150 |   const focusSearchBar = useCallback((e) => {
    151 |     e.preventDefault();
  > 152 |     this.searchBar.current.focus();
        |     ^
    153 |   }, []);
    154 | 
    155 |   useMousetrap({
```
how could I improve?  @0xdevalias @soichih?

src/hooks/chatSidebar.ts
```diff
@@ -38,6 +42,13 @@ export const useChatSidebarControls = () => {
     }
   }, [expandSidebar, collapseSidebar, isExpanded]);
 
+  useMousetrap({
+    keys: KEYBOARD_SHORTCUTS.chat,
+    callback: toggleSidebar,
+    // TODO: bindRef: (null as never) as MutableRefObject<HTMLElement>,
+    withGlobalBind: true, // TODO: remove this once we have a ref to bind to
+  });
```
yarikoptic, line 50: @0xdevalias note that all uses of `useMousetrap`  are with `withGlobalBind: true`.  Should it be just be made default and be done until the use case comes up requiring RFin for `: false`?

0xdevalias: They are all global right now because non-global is broken. 

Changing the default based on that alone isn't a good reason to change the default.

Do you have other reason to believe that global would be a better default?

yarikoptic: It was just my 1c: unused default isn't a default really, just a hypothetical opportunity for an alternative behavior, which is now exposed in api, and as you said - broken. up to you to keep it, I am ok either way.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: Merged @soichih fix. Should be all set

src/components/molecules/NavBar/NavBar.tsx
soichih: @yarikoptic I forgot I was going to fix this (disabling keyboard shortcut when input is focused).. but I am out of time for this evening. Did you happen to fix this already?

yarikoptic: Yes, no fix needed. I changed shortcut to non overlapping and switched back to the mousetrap - all works nicely now

yarikoptic:
@0xdevalias WDYT about this PR -- anything else to be done to land it?

src/components/atoms/InputField/InputField.tsx
0xdevalias: Based on another comment thread on this PR, I believe the issue/solution was that the key was already bound in mousetrap.

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: Sounds good. Is the `useCallback` aspect here still an issue? Or is this outdated now based on newer patterns?

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: All good. Thanks :)

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: Yay! Glad it works.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: > how could I improve

@yarikoptic The issue in that code snippet you provided is that you're trying to use `this.`, which references object state within a class. Yet we use functional components.

Since your `const searchBar = React.useRef(null);` is within the same 'variable scope' as where you want to use it, you can just use `searchBar` directly (no `this.`)

The only other issue is that it could be `null`/`undefined`, so you would use TypeScript's 'safe accessor'/'optional chaining' syntax. Something like: `searchBar.current?.focus();`

You would also need to ensure you add `searchBar` to the 'deps' array of the `useCallback` so that it knows to break the memoisation when that variable changes.

---

> Merged @soichih fix. Should be all set

Sounds good :)

src/hooks/useKeyboardControls.ts
0xdevalias: While most things follow THIS_PATTERN, we have a few examples in `src/settings` already that don't. I feel like this is a case that make more sense to be `KeyboardShortcutKeys` or similar rather than `KEYBOARD_SHORTCUTS`, as we are using it sort of like an `enum`.

src/components/molecules/NavBar/NavBar.tsx
yarikoptic: I believe it is no longer pertinent as we reverted to use mousetrap after you pointed to the real issue of overlapping shortcuts

src/hooks/useKeyboardControls.ts
yarikoptic: You are the boss, I will change (I have seen only few of such in there).

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: Cool, makes sense :) Mostly just wanted to ensure it was 'covered' now that i'm catching up on the comments/changes here.

src/hooks/useKeyboardControls.ts
0xdevalias: > I have seen only few of such in there

Yeah, there's not many in there that break THIS_PATTERN currently. That file (and it's code style) is a fair bit of legacy code/new stuff hacked into it rather than being overly 'planned'/thought through. So That pattern may change in future (eg. in `utils/*` and `types/*` we export `const`s/`enum`s/etc all the time that don't follow THAT_PATTERN; so it's established in the codebase in general, just not so much represented in `src/settings` currently)

0xdevalias: > You are the boss

Ha! I'm just a guy with strong opinions on our codebase 😛 

> I will change 

Sounds good, thanks :)

src/hooks/chatSidebar.ts
0xdevalias: Cool cool, let's leave it as is for now. Might help me feel the pain enough to find the time to fix it properly.

src/hooks/useKeyboardControls.ts
yarikoptic: done on that rename in 27dae04ab854b2d5429f261bc0fca332b2bd1a54 ATM.  I will leave one more question in a dedicated thread.

src/hooks/chatSidebar.ts
```diff
@@ -38,6 +42,13 @@ export const useChatSidebarControls = () => {
     }
   }, [expandSidebar, collapseSidebar, isExpanded]);
 
+  useMousetrap({
+    keys: KeyboardShortcutKeys.chat,
+    callback: toggleSidebar,
```
yarikoptic, line 47: somehow this is not working for a case I have described in PR description: If go to "Direct messages" and then back to "Party chat" - "c" no longer toggles chat.  Going back to Direct or toggling chat with mouse makes everything work again.  @0xdevalias any ideas?

src/hooks/chatSidebar.ts
yarikoptic: ok, let's then consider resolved (will mark as such also in PR description)

0xdevalias: Sounds good.

src/hooks/chatSidebar.ts
0xdevalias: I believe this issue was what I originally shared https://github.com/sparkletown/sparkle/pull/1416#discussion_r636484294 about.

0xdevalias: I haven't yet looked specifically into this and any deeper potential reasons beyond those original musing thoughts.

src/components/atoms/InputField/InputField.tsx
```diff
@@ -61,9 +61,20 @@ const InputField: React.ForwardRefRenderFunction<
 
   const inputClassNames = classNames("input-field__input", inputClassName);
 
+  const keyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
```
0xdevalias, line 64:
```suggestion
  // @debt Replace this with useMousetrap when we fix it to bind refs (not just globally)
  const keyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
```

src/components/atoms/InputField/InputField.tsx
```diff
@@ -61,9 +61,20 @@ const InputField: React.ForwardRefRenderFunction<
...
+    if (e.key === "Escape" && e.target) {
```
0xdevalias, line 65: Can you please also add this `"Escape"` to `KeyboardShortcutKeys` in `src/settings`. 

Probably under a 'root key/namespace' of `inputField` or similar, then something like `blur` or `unfocus` or `deselect` or some other word that describes the intent of this functionality in a succinct yet descriptive way, then use that here rather than the string literal.

src/components/atoms/InputField/InputField.tsx
```diff
@@ -61,9 +61,20 @@ const InputField: React.ForwardRefRenderFunction<
...
+  };
+
   return (
     <div className={containerClassNames}>
-      <input ref={ref} className={inputClassNames} {...extraInputProps} />
+      <input
+        ref={ref}
+        className={inputClassNames}
+        onKeyDown={keyDown}
+        {...extraInputProps}
```
0xdevalias, line 76: If external code passes the `onKeyDown` prop, then this will end up being overwritten. To fix it we could either create a new `useCallback`'d 'wrapping function' that will call the passed in `onKeyDown` _and_ our handler defined in this code, or we could fix `useMousetrap` to properly use a local `ref`, which would then mean we shouldn't need to explicitly use the `onKeyDown` prop here.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -141,6 +152,19 @@ const NavSearchBar = () => {
     [searchQuery]
   );
 
+  const inputRef = useRef<HTMLInputElement>(null);
+  const focusSearchBar = useCallback((e) => {
+    e.preventDefault();
+    if (inputRef && inputRef.current) inputRef.current.focus();
+  }, []);
```
0xdevalias, line 159: This will have the same effect, though in shorter/clearer code (using TypeScript's 'optional chaining' feature):

```suggestion
  const inputRef = useRef<HTMLInputElement>(null);
  const focusSearchBar = useCallback((e) => {
    e.preventDefault();
    inputRef.current?.focus();
  }, []);
```

0xdevalias: @mike-lvov @sunny-viktoryia I know we've talked a bit about avoiding using `e.preventDefault()` of late, though still don't fully have my head around the 'what/when' of it yet.

Do you have any thoughts/suggestions on whether we need to use it here for this, and if so, what our 'preferred' way of solving this is?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -161,6 +185,7 @@ const NavSearchBar = () => {
       <InputField
         value={searchInputValue}
         inputClassName="NavSearchBar__search-input"
+        ref={inputRef}
```
0xdevalias, line 188: [nit] Prefer `ref` to be the first prop:

```suggestion
      <InputField
        ref={inputRef}
        value={searchInputValue}
        inputClassName="NavSearchBar__search-input"
```

src/hooks/useMapKeyboardControls.ts
```diff
@@ -1,18 +1,13 @@
...
 
-export const MovementKeys = {
-  up: ["up", "w"],
-  down: ["down", "s"],
-  left: ["left", "a"],
-  right: ["right", "d"],
-};
-
-interface UseKeyboardControlsProps {
+interface UseMapKeyboardControlsProps {
```
0xdevalias, line 10: Seems this code predates our current style/pattern of exporting all of the props types:

```suggestion
export interface UseMapKeyboardControlsProps {
```

src/settings.ts
```diff
@@ -591,3 +591,15 @@ export const SCHEDULE_SHOW_DAYS_AHEAD = 7;
...
+  chat: "c",
+  movement: {
+    down: ["down", "s"],
+    left: ["left", "a"],
+    right: ["right", "d"],
+    up: ["up", "w"],
+  },
+  schedule: "S",
+  search: "/",
+};
```
0xdevalias, line 605: Since we already use lowercase `s` for movement, it might be strange to use uppercase `S` for an entirely different feature.

I don't have a good idea off the top of my head, but can we change this to a different key binding please. Maybe `e` for events? (not sure if that is intuitive enough for people)

0xdevalias: As mentioned in another comment in this review, it would also be good to add the `esc` keybinding here, rather than being in a string.

src/hooks/chatSidebar.ts
0xdevalias: Ok, so looking at the definition of `toggleSidebar` I expect that a 'stale function reference' is the most likely case here.

Whereas the other 'keybound togglers' use a function that never changes, `toggleSidebar` gets recreated whenever any of `expandSidebar, collapseSidebar, isExpanded` changes. So when you run it the first time, `isExpanded` will change, and so the function gets recreated. Yet `useMousetrap` likely has the old version of the function bound/cached still, so it just constantly tries to run the 'same' toggle (eg. open sidebar) every time.

Need to look a bit deeper to figure the easiest/best solution here.

0xdevalias (changes requested):
A few more little bits and pieces of polish then this should be good to go.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: (this might actually be the pattern we use in all of our keybinding handlers, based on a comment in `useMapKeyboardControls`

0xdevalias: applied the above suggested changes in https://github.com/sparkletown/sparkle/pull/1416/commits/2dc066902d7fe56f50eb54583cc33294a7f246db

The `e.preventDefault()` part of this is still oustanding.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1416/commits/2dc066902d7fe56f50eb54583cc33294a7f246db

src/hooks/useMapKeyboardControls.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1416/commits/2dc066902d7fe56f50eb54583cc33294a7f246db

src/hooks/chatSidebar.ts
0xdevalias: Looking at the definition of `useMousetrap`, it should unregister+re-register itself whenever the `useEffect`'s deps array changes, which includes the `callback` prop:

- https://github.com/sparkletown/sparkle/blob/staging/src/hooks/useMousetrap.ts#L70-L87

So I would have expected the change of `toggleSidebar` to cause this to be re-registered and work as expected..

---

To confirm/start debugging further from here, I would add a `console.log` within `useMousetrap`'s `useEffect` to see if it is being rebound when it should be; and I would also add a `console.log` to `toggleSidebar` to see if it's being called (and what the value of `isExapanded` is) each time you press the shortcut.

That should at least get more information to figure out whether it's following the 'static' expectations/assumptions i've made above from looking at the code.

src/hooks/useKeyboardControls.ts
0xdevalias: Sounds good

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: I believe it was needed to avoid the key propagate into search input element. So the question remaining is either it should be moved to mousetrap handler itself?

src/settings.ts
yarikoptic: done in f79f6bc94dd969fc81493962b77da308cda7c557.  I will comment further in that original [thread](https://github.com/sparkletown/sparkle/pull/1416#discussion_r638465340)

yarikoptic: Re
> Since we already use lowercase s for movement, it might be strange to use uppercase S for an entirely different feature.

IMHO it is ok since matches `[S]chedule`. I don't mind changing it to whatever (e.g. `e`) if you insist though.

src/components/atoms/InputField/InputField.tsx
yarikoptic: any help on this would be appreciated since sounds beyond my react-free comprehension ;) (attn @soichih)

src/components/atoms/InputField/InputField.tsx
yarikoptic: Defined `.escape` in `KeyboardShortcutKeys` in f79f6bc94dd969fc81493962b77da308cda7c557 . Not sure yet what you meant by the rest of the comment.

src/hooks/chatSidebar.ts
yarikoptic: WONDERFUL explanation and pointers elsewhere -- even I managed to troubleshoot it to death... Added console.log first, saw that no mousetraps when switching back... looked at other *Chat*.ts* performed -- they all seems to have some kind of onClick which triggers specific to them select*Chat.  Added similar one to VenueChat - All is working now . Pushed 08c63ff6272e792a0ccb2474f95a600918a37611 .

yarikoptic: resolved now with 08c63ff6272e792a0ccb2474f95a600918a37611 + 8f02d07eb87046f6f2644eb1cc1a6c583bb42afa

Works good as to me.  Anything else or could be blessed? ;)

0xdevalias: > looked at other Chat.ts* performed -- they all seems to have some kind of onClick which triggers specific to them select*Chat. 

@yarikoptic When you say 'other Chat.ts', which components/etc are you referring to?

Looking at usages of `useChatSidebarControls` i'm guessing you mean these?:

- `src/components/organisms/ChatSidebar/ChatSidebar.tsx`
  - `const { isExpanded, toggleSidebar, chatSettings, selectVenueChat, selectPrivateChat } = useChatSidebarControls();`
- `src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx`
  - `const { selectRecipientChat } = useChatSidebarControls();`
    - note that this is `selectRecipientChat`, yet we are in `PrivateChats`
- `src/components/organisms/ChatSidebar/components/RecipientChat/RecipientChat.tsx`
  - `const { selectPrivateChat } = useChatSidebarControls();`
    - note that this is `selectPrivateChat`, yet we are in `RecipientChat`

Looking at the changes in 08c63ff + 8f02d07, you've added:

- `src/components/organisms/ChatSidebar/components/VenueChat/VenueChat.tsx`
  - `const { selectVenueChat } = useChatSidebarControls();`
    - note that this is `selectVenueChat`, and we are in `VenueChat`
    
While this might have made it work, if it did, it's likely as an unrelated side effect rather than this being the proper fix.

Without looking too deeply into the specifics of that code, I believe:

- `PrivateChats` uses `selectRecipientChat` to switch to the 'show me my private messages with this particular user' view
- `RecipientChat` uses `selectPrivateChat` to switch back to the 'show me a list of all users I have private chats with' view

I don't believe there is any valid reason to add `selectVenueChat` as an `onClick` in `VenueChat`.

src/components/atoms/InputField/InputField.tsx
0xdevalias: > Not sure yet what you meant by the rest of the comment.

@yarikoptic Instead of:
```typescript
export const KeyboardShortcutKeys = {
  chat: "c",
  escape: "escape",
```

Something like (or one of the other words mentioned in place of `inputField` and `deselect` if you think a different one makes more sense):

```typescript
export const KeyboardShortcutKeys = {
  chat: "c",
  inputField: {
    deselect: "escape",
  }
```

src/components/atoms/InputField/InputField.tsx
0xdevalias: Fwiw, my preference would be for us to fix `useMousetrap` so it can bind to single elements using `ref`.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: > I believe it was needed to avoid the key propagate into search input element.

According the the `mousetrap` docs, that shouldn't be required when we're in an `input` field and similar:

- https://craig.is/killing/mice#api.bind.text-fields
- https://craig.is/killing/mice#api.stopCallback

src/components/atoms/InputField/InputField.tsx
0xdevalias: Actually, seems I backlogged a thing about it on our internal issues: 🔒 https://github.com/sparkletown/internal-sparkle-issues/issues/376

Since you won't be able to see that:

> > We'll also want to ensure that the 'scope' of the keyboard shortcuts is limited to the right place.  For that, we'll need to fix the long outstanding bug in `useMousetrap`, that will cause some kind of error/issue/just not work when used with `withGlobalBind: false`:
> > 
> > - https://github.com/sparkletown/sparkle/blob/staging/src/hooks/useMousetrap.ts
> > 
> > _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/371#issuecomment-778909908_
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/376_

Apparently I didn't include a whole lot of detail..

src/components/molecules/NavSearchBar/NavSearchBar.tsx
yarikoptic: I have triple checked that removing e.preventDefault pastes `/` into input.  If mousetrap is supposed, it doesn't (may be due to all the layering to get to that `input`).

Also there is a comment
```
// TODO: use e.preventDefault() or return false or similar in the keyboard handlers (check mousetrap docs) so we don't scroll on arrow key presses
```
for `useMapKeyboardControls` so observed needed elsewhere.  Not certain if digging deeper is productive on this one.

src/components/atoms/InputField/InputField.tsx
yarikoptic: done. force-pushed with rebase on top of current staging (conflicts start to appear, pain, suffering...)

src/components/atoms/InputField/InputField.tsx
yarikoptic: > Fwiw, my preference would be for us to fix useMousetrap so it can bind to single elements using ref.

I would have loved that too but overall feels like OOS for the purpose of this PR since it is not the one introduced useMousetrap  and might take awhile, and there are bigger/nastier fishes to fry ATM.  As far as I see it - it all works ATM and there is that saying "ain't broken, don't fix it" ;)

src/hooks/chatSidebar.ts
yarikoptic: nice analysis! and yes -- it is those which I saw but missed that "pairing" pattern you point to.  It somewhat partially explains it though I guess -- since PrivateChats does have a call to `useChatSidebarControls` - it overloads them. When going to `venueChat`, no call to `useChatSidebarControls` and some things are stale, shortcuts do not work.  So it might be just a matter of having it bound just to trigger that code and ensure proper rebinds. I have looked around but see no better place to stick that onClick.    I have tested quite a bit and did not spot any oddity (besides it working very well ;-)).

src/components/atoms/InputField/InputField.tsx
0xdevalias: > and there is that saying "ain't broken, don't fix it"

Well, it is broken currently, as described in https://github.com/sparkletown/sparkle/pull/1416#discussion_r638466103

Fixing `useMousetrap` + using that is the 'optimal best' solution. The `useCallback` workaround is the 'passable for now' hack solution.

yarikoptic: Re

> If external code passes the onKeyDown prop,

Could it be checked at compile time and crash if so to provide safeguard for now?

0xdevalias: Even if we could, there's no point/it would be more convoluted/slower to do than to just implement the solution I described originally in https://github.com/sparkletown/sparkle/pull/1416#discussion_r638466103

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: > not even sure if that would be a good strategy from mouseTrap to just do it by default - I might need a callback to just act on the character but do pass it through to input -- who knows?!

Which is why they provide the documentation that explains how it works and how to change it if you need to, that I provided above in https://github.com/sparkletown/sparkle/pull/1416#discussion_r639442598

---

> I have triple checked that removing e.preventDefault pastes / into input. If mousetrap is supposed, it doesn't

Fair. Potentially because we are using global keybindings rather than directly on the element.

---

> Also there is a comment for useMapKeyboardControls so observed needed elsewhere

Yup, because movement on the map doesn't relate to `input` and similar things that `mousetrap` disables it for by default. That comment may also not even be strictly true, it was written quite 
a while ago (by me) when I had less understanding of how it all interacts.

src/components/atoms/InputField/InputField.tsx
yarikoptic: I have resolved merge conflicts and pushed -- it still works nicely.  If you see how it could be improved to address  shortcomings of used here existing sparkle codebase -- please do not hesitate to push them directly.  Even with your nice descriptions of desired changes they are "above my react-grade" (which is very very low ;-))

mike-lvov:
@sofisparkle says it's a good feature, but it's not the highest priority & has a lot of conflicts, so it'll stay stale for a while

PR #1419, opened 2021-05-18
add optional scheduleStartDate to DB + force NavBarSchedule to start from that date when present

margulies:
Added an initial fix to enable setting the start date for the schedule using `start_utc_seconds` in `venue`. Planning to develop this further, but having this feature available for testing purposes would be helpful at this stage.

### DIFF
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -5,13 +5,7 @@ import React, {
   MouseEventHandler,
   useCallback,
 } from "react";
-import {
-  addDays,
-  startOfToday,
-  format,
-  getUnixTime,
-  fromUnixTime,
-} from "date-fns";
+import { addDays, format, getUnixTime, fromUnixTime } from "date-fns";
 import { groupBy, range } from "lodash";
 import classNames from "classnames";
 
@@ -63,9 +57,21 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const userEventIds =
     userWithId?.myPersonalizedSchedule ?? emptyPersonalizedSchedule;
 
-  const { isLoading, relatedVenues, relatedVenueIds } = useRelatedVenues({
+  const {
+    isLoading,
+    relatedVenues,
+    relatedVenueIds,
+    sovereignVenue,
+  } = useRelatedVenues({
     currentVenueId: venueId,
   });
+
+  const isScheduleStartDate: boolean = sovereignVenue?.scheduleStartDate
+    ? true
+    : false;
+  const today = fromUnixTime(
+    sovereignVenue?.scheduleStartDate || getUnixTime(new Date())
+  );
 
   const {
     isEventsLoading,
@@ -79,8 +85,6 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const today = startOfToday();
-
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
       const day = addDays(today, dayIndex);
       const classes = classNames("NavBarSchedule__weekday", {
@@ -97,11 +101,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
           className={classes}
           onClick={onWeekdayClick}
         >
-          {dayIndex === 0 ? "Today" : format(day, "E")}
+          {isScheduleStartDate
+            ? format(day, "E, LLL d")
+            : dayIndex === 0
+            ? "Today"
+            : format(day, "E")}
         </li>
       );
     });
-  }, [selectedDayIndex]);
+  }, [selectedDayIndex, today, isScheduleStartDate]);
 
   const getEventLocation = useCallback(
     (locString: string): VenueLocation => {
@@ -114,9 +122,13 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   );
 
   const schedule: ScheduleDay = useMemo(() => {
-    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const dayStart = addDays(today, selectedDayIndex);
     const daysEvents = relatedVenueEvents
-      .filter(isEventWithinDate(selectedDayIndex === 0 ? Date.now() : dayStart))
+      .filter(
+        isEventWithinDate(
+          selectedDayIndex === 0 && !isScheduleStartDate ? Date.now() : dayStart
+        )
+      )
       .map(prepareForSchedule(dayStart, userEventIds));
 
     const locatedEvents: LocatedEvents[] = Object.entries(
@@ -132,7 +144,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
       dayStartUtcSeconds: getUnixTime(dayStart),
       personalEvents: daysEvents.filter((event) => event.isSaved),
     };
-  }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
+  }, [
+    relatedVenueEvents,
+    userEventIds,
+    selectedDayIndex,
+    getEventLocation,
+    today,
+    isScheduleStartDate,
+  ]);
 
   const containerClasses = classNames("NavBarSchedule", {
     "NavBarSchedule--show": isVisible,
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -5,7 +5,14 @@ import React, {
   MouseEventHandler,
   useCallback,
 } from "react";
-import { addDays, format, getUnixTime, fromUnixTime } from "date-fns";
+import {
+  addDays,
+  format,
+  getUnixTime,
+  fromUnixTime,
+  startOfToday,
+  startOfDay,
+} from "date-fns";
 import { groupBy, range } from "lodash";
 import classNames from "classnames";
 
@@ -33,6 +40,7 @@ import {
 } from "./utils";
 import { isEventWithinDate } from "utils/event";
 import { WithVenueId } from "utils/id";
+import { isTruthy } from "utils/types";
 
 import "./NavBarSchedule.scss";
 
@@ -66,12 +74,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     currentVenueId: venueId,
   });
 
-  const isScheduleStartDate: boolean = sovereignVenue?.scheduleStartDate
-    ? true
-    : false;
-  const today = fromUnixTime(
-    sovereignVenue?.scheduleStartDate || getUnixTime(new Date())
+  const hasScheduleStartDate: boolean = isTruthy(
+    sovereignVenue?.scheduleStartDate
   );
+  const scheduleStartDate: Date = sovereignVenue?.scheduleStartDate?.toDate()!;
+  const firstDay = useMemo(() => {
+    return hasScheduleStartDate
+      ? startOfDay(scheduleStartDate)
+      : startOfToday();
+  }, [hasScheduleStartDate, scheduleStartDate]);
 
   const {
     isEventsLoading,
@@ -86,7 +97,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
 
   const weekdays = useMemo(() => {
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(today, dayIndex);
+      const day = addDays(firstDay, dayIndex);
       const classes = classNames("NavBarSchedule__weekday", {
         "NavBarSchedule__weekday--active": dayIndex === selectedDayIndex,
       });
@@ -101,7 +112,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
           className={classes}
           onClick={onWeekdayClick}
         >
-          {isScheduleStartDate
+          {hasScheduleStartDate
             ? format(day, "E, LLL d")
             : dayIndex === 0
             ? "Today"
@@ -109,7 +120,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
         </li>
       );
     });
-  }, [selectedDayIndex, today, isScheduleStartDate]);
+  }, [selectedDayIndex, firstDay, hasScheduleStartDate]);
 
   const getEventLocation = useCallback(
     (locString: string): VenueLocation => {
@@ -122,11 +133,13 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   );
 
   const schedule: ScheduleDay = useMemo(() => {
-    const dayStart = addDays(today, selectedDayIndex);
+    const dayStart = addDays(firstDay, selectedDayIndex);
     const daysEvents = relatedVenueEvents
       .filter(
         isEventWithinDate(
-          selectedDayIndex === 0 && !isScheduleStartDate ? Date.now() : dayStart
+          selectedDayIndex === 0 && !hasScheduleStartDate
+            ? Date.now()
+            : dayStart
         )
       )
       .map(prepareForSchedule(dayStart, userEventIds));
@@ -149,8 +162,8 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     userEventIds,
     selectedDayIndex,
     getEventLocation,
-    today,
-    isScheduleStartDate,
+    firstDay,
+    hasScheduleStartDate,
   ]);
 
   const containerClasses = classNames("NavBarSchedule", {
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -12,6 +12,7 @@ import {
   fromUnixTime,
   startOfToday,
   startOfDay,
+  isToday,
 } from "date-fns";
 import { groupBy, range } from "lodash";
 import classNames from "classnames";
@@ -40,7 +41,6 @@ import {
 } from "./utils";
 import { isEventWithinDate } from "utils/event";
 import { WithVenueId } from "utils/id";
-import { isTruthy } from "utils/types";
 
 import "./NavBarSchedule.scss";
 
@@ -74,15 +74,13 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     currentVenueId: venueId,
   });
 
-  const hasScheduleStartDate: boolean = isTruthy(
-    sovereignVenue?.scheduleStartDate
-  );
-  const scheduleStartDate: Date = sovereignVenue?.scheduleStartDate?.toDate()!;
-  const firstDay = useMemo(() => {
-    return hasScheduleStartDate
-      ? startOfDay(scheduleStartDate)
+  const scheduledStartDate = sovereignVenue?.start_utc_seconds;
+
+  const firstDayOfSchedule = useMemo(() => {
+    return scheduledStartDate
+      ? startOfDay(fromUnixTime(scheduledStartDate))
       : startOfToday();
-  }, [hasScheduleStartDate, scheduleStartDate]);
+  }, [scheduledStartDate]);
 
   const {
     isEventsLoading,
@@ -96,8 +94,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
+    const formatDayLabel = (day: Date) => {
+      if (isToday(day)) return "Today";
+      if (isToday(firstDayOfSchedule)) return format(day, "E");
+      return format(day, "E, LLL d");
+    };
+
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(firstDay, dayIndex);
+      const day = addDays(firstDayOfSchedule, dayIndex);
       const classes = classNames("NavBarSchedule__weekday", {
         "NavBarSchedule__weekday--active": dayIndex === selectedDayIndex,
       });
@@ -112,15 +116,11 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
           className={classes}
           onClick={onWeekdayClick}
         >
-          {hasScheduleStartDate
-            ? format(day, "E, LLL d")
-            : dayIndex === 0
-            ? "Today"
-            : format(day, "E")}
+          {formatDayLabel(day)}
         </li>
       );
     });
-  }, [selectedDayIndex, firstDay, hasScheduleStartDate]);
+  }, [selectedDayIndex, firstDayOfSchedule]);
 
   const getEventLocation = useCallback(
     (locString: string): VenueLocation => {
@@ -133,16 +133,10 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   );
 
   const schedule: ScheduleDay = useMemo(() => {
-    const dayStart = addDays(firstDay, selectedDayIndex);
+    const startOfSelectedDay = addDays(firstDayOfSchedule, selectedDayIndex);
     const daysEvents = relatedVenueEvents
-      .filter(
-        isEventWithinDate(
-          selectedDayIndex === 0 && !hasScheduleStartDate
-            ? Date.now()
-            : dayStart
-        )
-      )
-      .map(prepareForSchedule(dayStart, userEventIds));
+      .filter(isEventWithinDate(startOfSelectedDay))
+      .map(prepareForSchedule(startOfSelectedDay, userEventIds));
 
     const locatedEvents: LocatedEvents[] = Object.entries(
       groupBy(daysEvents, buildLocationString)
@@ -154,7 +148,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     return {
       locatedEvents,
       isToday: selectedDayIndex === 0,
-      dayStartUtcSeconds: getUnixTime(dayStart),
+      dayStartUtcSeconds: getUnixTime(startOfSelectedDay),
       personalEvents: daysEvents.filter((event) => event.isSaved),
     };
   }, [
@@ -162,8 +156,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     userEventIds,
     selectedDayIndex,
     getEventLocation,
-    firstDay,
-    hasScheduleStartDate,
+    firstDayOfSchedule,
   ]);
 
   const containerClasses = classNames("NavBarSchedule", {
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -82,6 +82,8 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
       : startOfToday();
   }, [scheduledStartDate]);
 
+  const isScheduleTimeshifted = !isToday(firstDayOfSchedule);
+
   const {
     isEventsLoading,
     events: relatedVenueEvents = emptyRelatedEvents,
@@ -94,10 +96,12 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const formatDayLabel = (day: Date) => {
-      if (isToday(day)) return "Today";
-      if (isToday(firstDayOfSchedule)) return format(day, "E");
-      return format(day, "E, LLL d");
+    const formatDayLabel = (day: Date | number) => {
+      if (isScheduleTimeshifted) {
+        return format(day, "E, LLL d");
+      } else {
+        return isToday(day) ? "Today" : format(day, "E");
+      }
     };
 
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
@@ -120,7 +124,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
         </li>
       );
     });
-  }, [selectedDayIndex, firstDayOfSchedule]);
+  }, [selectedDayIndex, firstDayOfSchedule, isScheduleTimeshifted]);
 
   const getEventLocation = useCallback(
     (locString: string): VenueLocation => {
--- a/src/types/venues.ts
+++ b/src/types/venues.ts
@@ -171,7 +171,6 @@ export interface BaseVenue {
   showRadio?: boolean;
   showBadges?: boolean;
   showZendesk?: boolean;
-  scheduleStartDate?: firebase.firestore.Timestamp;
 }
 
 export interface GenericVenue extends BaseVenue {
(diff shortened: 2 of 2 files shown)


src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -63,10 +57,22 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+  } = useRelatedVenues({
     currentVenueId: venueId,
   });
 
+  const isScheduleStartDate: boolean = sovereignVenue?.scheduleStartDate
+    ? true
+    : false;
+  const today = fromUnixTime(
+    sovereignVenue?.scheduleStartDate || getUnixTime(new Date())
+  );
```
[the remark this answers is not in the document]
0xdevalias: This code isn't perfect, but cleans this up a little bit.

We can use `isTruth` to simplify the first line.
We use `??` to default undefined things rather than `||`.

```suggestion
  const hasScheduleStartDate: boolean = isTruthy(sovereignVenue?.scheduleStartDate);
  
  const scheduleStartDate = sovereignVenue?.scheduleStartDate ? fromUnixTime(sovereignVenue?.scheduleStartDate) ?? Date.now()
  );
```

We can't use `Date.now()` (or `new Date`) in the main component body like this, as it will change on every render. This would need to be done within the `useMemo` below most likely.

So I would probably just extract the `scheduleStartDate` similar to this here (though probably it should be named better in the database to imply that it is in utc seconds):

```suggestion
  const scheduleStartDate = sovereignVenue ?? {}
```

Then would use this within the `useMemo` below (see code suggestions below)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -79,8 +85,6 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const today = startOfToday();
-
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
```
0xdevalias, line 88:
```suggestion
    const firstDayOfSchedule = scheduleStartDateUtcSeconds ? startOfDay(fromUnixTime(scheduleStartDateUtcSeconds)) : startOfToday();
    
    return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
```

Further Reading:

- https://date-fns.org/v2.21.3/docs/startOfDay
- https://date-fns.org/v2.21.3/docs/startOfToday
- https://date-fns.org/v2.21.3/docs/fromUnixTime

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,11 +101,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
           className={classes}
           onClick={onWeekdayClick}
         >
-          {dayIndex === 0 ? "Today" : format(day, "E")}
+          {isScheduleStartDate
+            ? format(day, "E, LLL d")
+            : dayIndex === 0
+            ? "Today"
+            : format(day, "E")}
```
0xdevalias, line 108: In the main body of the code before this JSX (probably directly after `firstDayOfSchedule` is defined) I would do something like:

```typescript
const firstDayOfScheduleLabel = isToday(firstDayOfSchedule) ? "Today" : format(day, "E, LLL d");
```

Then change the code here to:

```suggestion
          {dayIndex === 0 ? firstDayOfScheduleLabel : format(day, "E")}
```

Further Reading:

- https://date-fns.org/v2.21.3/docs/isToday
- https://date-fns.org/v2.21.3/docs/format
  - > Day of week (formatting) | E..EEE | Mon, Tue, Wed, ..., Sun
  - > Month (stand-alone) | LLL | Jan, Feb, ..., Dec
  - > Day of month | d | 1, 2, ..., 31

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,11 +101,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
         </li>
       );
     });
-  }, [selectedDayIndex]);
+  }, [selectedDayIndex, today, isScheduleStartDate]);
```
0xdevalias, line 112: As the code is currently defined above in this PR, `today` will change on every render, and thus break memo on this every render (which we don't want).

The solution would likely be to move the relevant aspects of it within this `useMemo` (as described elsewhere in this review)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,9 +122,13 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   );
 
   const schedule: ScheduleDay = useMemo(() => {
-    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const dayStart = addDays(today, selectedDayIndex);
```
0xdevalias, line 125: It seems like we probably also want to use `firstDayOfSchedule` here as well. So it would make sense to pull it out of the `useMemo` above into its own `useMemo` block so it can easily be used in both places without code duplication/causing re-renders.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: As noted in another review comment, since we actually need to reference `firstDayOfSchedule` from multiple blocks, we should define it within it's own `useMemo`, rather than this one here.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,9 +122,13 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
     const daysEvents = relatedVenueEvents
-      .filter(isEventWithinDate(selectedDayIndex === 0 ? Date.now() : dayStart))
+      .filter(
+        isEventWithinDate(
+          selectedDayIndex === 0 && !isScheduleStartDate ? Date.now() : dayStart
+        )
+      )
```
0xdevalias, line 131: What's the intent/reason for this code change here?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -132,7 +144,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
     };
-  }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
+  }, [
+    relatedVenueEvents,
+    userEventIds,
+    selectedDayIndex,
+    getEventLocation,
+    today,
+    isScheduleStartDate,
+  ]);
```
0xdevalias, line 154: As the code is currently defined above in this PR, `today` will change on every render, and thus break memo on this every render (which we don't want).

The solution would likely be to move the relevant aspects of it within this `useMemo` (as described elsewhere in this review)

src/types/venues.ts
```diff
@@ -171,6 +171,7 @@ export interface BaseVenue {
   showRadio?: boolean;
   showBadges?: boolean;
   showZendesk?: boolean;
+  scheduleStartDate?: number;
```
0xdevalias, line 174: Why is there a new `scheduleStartDate` config item here, rather than re-using `venue.start_utc_seconds` or similar as the PR description suggests?

0xdevalias (changes requested):

---

A few notes for PRs in general:

It would be good if you could also include a screenshot to show how the changes made in this PR look/are intended (to help 'show off' the feature)

As a minor sidenote, instead of naming PRs "added X"/"fixed X"/etc (past tense), we use "add X"/"fix X"/etc (present tense) or similar.

margulies:
I've revised the code to use the Timestamp in the firestore database, as it makes it much easier to select a date. This is an initial version of this feature to enable viewing of a schedule that does not begin today. In future iterations that plan would be integrate this into the admin user interface

src/types/venues.ts
margulies: now changed to Timestamp for ease of entry in firestore database

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: vscode seems to do that by default — I've since turned off that feature, so additional linebreaks shouldn't be a problem in the future

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: I had a bit of trouble with this one, and left (in a slightly modified form) as is. As I understand it, the proposed change would render every day other than the first as a day of the week. The aim here is to render all days as 'Mon, Sep 1' if the start date is selected, and 'Today', 'Mon', ... if the start date of the schedule is driven by the current day

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
0xdevalias: @margulies You didn't answer my question in https://github.com/sparkletown/sparkle/pull/1419#discussion_r634894934 ?

> Why is there a new `scheduleStartDate` config item here, rather than re-using `venue.start_utc_seconds` as the PR description suggests?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: > The aim here is to render all days as 'Mon, Sep 1' if the start date is selected, and 'Today', 'Mon', ... if the start date of the schedule is driven by the current day

Mm, fair. Nesting ternaries like you did makes it super hard to read the code/infer the intended logic, which is why I made the logic error that lead to my suggestion.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: Cool. We use `prettier` for our formatting, so if it has a mode that tells it to use that (which I imagine it should) then that would work even better (though if you don't, it should just happen as part of our git hooks on commit)

src/types/venues.ts
0xdevalias: Fair, though as asked in https://github.com/sparkletown/sparkle/pull/1419#discussion_r635709031, that doesn't really answer my question about _why_ we are introducing this new field in the first place.

In your original description for this PR you talk about using `venue.start_utc_seconds`, why not use that here?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
margulies: The answer is below — I intended this to be a way to easily modify the start date directly in the firestore database. The interfact for putting in a Timestamp using the built-in calendar is much easier that searching utc_seconds. I've left it in as is, because with the type change in this revision, it seemed better not to modify `venue.start_utc_seconds`

src/types/venues.ts
margulies: I think the basic reason is that @mike-lvov suggested I use `scheduleStartDate` as a field name. But since changing it to a different type than start_utc_seconds, which facilitates input by the user, it seems to make sense to keep it separate

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -63,10 +65,25 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
 
+  const hasScheduleStartDate: boolean = isTruthy(
+    sovereignVenue?.scheduleStartDate
+  );
+  const scheduleStartDate: Date = sovereignVenue?.scheduleStartDate?.toDate()!;
+  const firstDay = useMemo(() => {
+    return hasScheduleStartDate
+      ? startOfDay(scheduleStartDate)
+      : startOfToday();
+  }, [hasScheduleStartDate, scheduleStartDate]);
```
0xdevalias, line 85: Why are you using `!` here? We don't use that feature of TypeScript as it basically breaks types (eg. the whole point of using TypeScript in the first place). In this instance, you're saying the value will always exist, which is explicitly untrue based on your use of `?.` before it. This will stop errors showing in TypeScript at coding time, but will just crash the platform at runtime.

For the sake of a few extra characters in the variable name, `firstDayOfSchedule` is more semantically explicit about its purpose than `firstDay`, which leads do a better developer experience (which is why I suggested that naming in my previous review)

This will correctly keep the type of `scheduleStartDate` as `Date | undefined`, and then use the 'falsiness`' of `undefined` in the ternary to choose which value to use.

```suggestion
  const scheduleStartDate = sovereignVenue?.scheduleStartData?.toDate();
 
  const firstDayOfSchedule = useMemo(() => {
    return scheduleStartDate
      ? startOfDay(scheduleStartDate)
      : startOfToday();
  }, [scheduleStartDate]);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in code suggested in my latest review

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -79,10 +96,8 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const today = startOfToday();
-
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(today, dayIndex);
+      const day = addDays(firstDay, dayIndex);
```
0xdevalias, line 100:
```suggestion
      const day = addDays(firstDayOfSchedule, dayIndex);
      const formatDayLabel = (day: Date | number) => {
        if (isScheduleTimeshifted) {
          return format(day, "E, LLL d");
        } else {
          return dayIndex === 0 ? "Today" : format(day, "E");
        }
      }
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,11 +112,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
           className={classes}
           onClick={onWeekdayClick}
         >
-          {dayIndex === 0 ? "Today" : format(day, "E")}
+          {hasScheduleStartDate
+            ? format(day, "E, LLL d")
+            : dayIndex === 0
+            ? "Today"
+            : format(day, "E")}
```
0xdevalias, line 119:
```suggestion
          {formatDayLabel(day)}
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -79,10 +96,8 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const today = startOfToday();
-
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
```
0xdevalias, line 99:
```suggestion
    const isScheduleTimeshifted = !isToday(firstDayOfSchedule);

    return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,11 +112,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+          {hasScheduleStartDate
+            ? format(day, "E, LLL d")
+            : dayIndex === 0
+            ? "Today"
+            : format(day, "E")}
         </li>
       );
     });
-  }, [selectedDayIndex]);
+  }, [selectedDayIndex, firstDay, hasScheduleStartDate]);
```
0xdevalias, line 123:
```suggestion
  }, [selectedDayIndex, firstDayOfSchedule]);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,9 +133,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   );
 
   const schedule: ScheduleDay = useMemo(() => {
-    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const dayStart = addDays(firstDay, selectedDayIndex);
```
0xdevalias, line 136:
```suggestion
    const dayStart = addDays(firstDayOfSchedule, selectedDayIndex);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: @margulies Actually.. there is also a logic change here, which I believe is what I was actually asking about here. You adde `!isScheduleStartDate` into the ternary. Why?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -132,7 +157,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
       personalEvents: daysEvents.filter((event) => event.isSaved),
     };
-  }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
+  }, [
+    relatedVenueEvents,
+    userEventIds,
+    selectedDayIndex,
+    getEventLocation,
+    firstDay,
+    hasScheduleStartDate,
```
0xdevalias, line 166:
```suggestion
    firstDayOfSchedule,
```

src/types/venues.ts
```diff
@@ -171,6 +171,7 @@ export interface BaseVenue {
   showRadio?: boolean;
   showBadges?: boolean;
   showZendesk?: boolean;
+  scheduleStartDate?: firebase.firestore.Timestamp;
```
0xdevalias, line 174: Pending answer as to why we are adding a new field here at all (rather than using existing), as per https://github.com/sparkletown/sparkle/pull/1419#discussion_r635710738 + https://github.com/sparkletown/sparkle/pull/1419#discussion_r635709031

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,9 +133,15 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+    const dayStart = addDays(firstDay, selectedDayIndex);
     const daysEvents = relatedVenueEvents
-      .filter(isEventWithinDate(selectedDayIndex === 0 ? Date.now() : dayStart))
+      .filter(
+        isEventWithinDate(
+          selectedDayIndex === 0 && !hasScheduleStartDate
+            ? Date.now()
+            : dayStart
+        )
+      )
```
0xdevalias, line 144: Pending answer to https://github.com/sparkletown/sparkle/pull/1419#discussion_r635719319 RE: the purpose of adding `!hasScheduleStartDate` here.

0xdevalias (changes requested):
Questions/comments/changes.

src/types/venues.ts
0xdevalias: Copying https://github.com/sparkletown/sparkle/pull/1419#discussion_r635712102 here as well to just reply in one place:

> The answer is below (see #1419 (comment)) — I intended this to be a way to easily modify the start date directly in the firestore database. The interfact for putting in a Timestamp using the built-in calendar is much easier that searching utc_seconds. I've left it in as is, because with the type change in this revision, it seemed better not to modify venue.start_utc_seconds

0xdevalias: > But since changing it to a different type than `start_utc_seconds`, which facilitates input by the user, it seems to make sense to keep it separate

The 'extra overhead' of having to calculate the timestamp when hacking it into the database seems like a 'less important' tradeoff than reusing an existing field that already contains the correct data/etc.

Does the existing `start_utc_seconds` field represent the same/required date here? If so, we should be using that.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
0xdevalias: Replied in https://github.com/sparkletown/sparkle/pull/1419#discussion_r635721776

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: Looking at the code above this.

```typescript
// Original Code
const dayStart = addDays(startOfToday(), selectedDayIndex);

// Current PR Code
const dayStart = addDays(firstDay, selectedDayIndex);
```

I believe the usage of `Date.now` in this filter is irrelevant.

In the original code we used `startOfToday()` + `Date.now()`, which are both going to represent the same day.

Since the first item on the schedule has index `0`, when it is selected, `dayStart` is going to be set to essentially:

```typescript
// Original Code
const dayStart = addDays(startOfToday(), 0);

// Current PR Code
const dayStart = addDays(firstDay, 0);
```

And since adding `0` days to the date is a noop, we essentially get:

```typescript
// Original Code
const dayStart = startOfToday();

// Current PR Code
const dayStart = firstDay;
```

Looking at `isEventWithinDate`, it already uses `startOfDay`/`endOfDay` on the provided `checkDate` value, meaning so long as `checkDate` represents the same date, it doesn't matter what time of day it is when passed in:

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/event.ts#L65-L76

Based on this, it shouldn't matter if we're checking todays date, or a future date, and so should be able to just do the following:

```suggestion
      .filter(isEventWithinDate(startOfSelectedDay))
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: See my exploration of the logic + suggested solution in https://github.com/sparkletown/sparkle/pull/1419/files#r635726723

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: Which it seems I already wrote in my review on the original feature.. I knew i'd written it somewhere already.. https://github.com/sparkletown/sparkle/pull/1302/files#r634036719

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: Thanks for pointing that out.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: Done

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: Done

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: Good catch — fixed

src/types/venues.ts
margulies: I've now modified this PR to use the `start_utc_seconds` value

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: Cool :)

Note that I think there might be slight tweaks to this again as part of @sunny-viktoryia's https://github.com/sparkletown/sparkle/pull/1426 PR that is currently in review (which is the cleanup/follow-on PR from #1302)

So depending on which gets merged first between it and this PR, we'll need to make sure we don't overwrite/lose any of the tweaks/changes.

src/types/venues.ts
0xdevalias: Cool, sounds good.

src/types/venues.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: No worries. I find that explaining the context/'why' of things allows people to make better proactive judgement calls on their own in future. Rather than me needing to provide the feedback each time.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

0xdevalias: Actually, looking at the code suggestion above (+ in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef#r635716283) vs what is implemented in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef, it looks like you didn't end up going with the `isScheduleTimeshifted` implementation.

Why was that?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/types/venues.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -79,10 +94,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
   const weekdays = useMemo(() => {
-    const today = startOfToday();
+    const formatDayLabel = (day: Date) => {
+      if (isToday(day)) return "Today";
+      if (isToday(firstDayOfSchedule)) return format(day, "E");
+      return format(day, "E, LLL d");
+    };
```
0xdevalias, line 101: I'm not sure the logic here is correct, is it?

---

Related to outstanding question in https://github.com/sparkletown/sparkle/pull/1419#discussion_r638459361:

> image
>
> image
>
> Actually, looking at the code suggestion above (+ in https://github.com/sparkletown/sparkle/pull/1419/files#r635716283) vs what is implemented in https://github.com/sparkletown/sparkle/pull/1419/commits/b78373170c83bbc19324c8694b04ea35b9ab06ef, it looks like you didn't end up going with the `isScheduleTimeshifted` implementation.
> 
> Why was that?
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1419#discussion_r638459361_
>

0xdevalias:
```suggestion
    const isScheduleTimeshifted = !isToday(firstDayOfSchedule);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -79,10 +94,14 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
 
     return range(0, SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(today, dayIndex);
+      const day = addDays(firstDayOfSchedule, dayIndex);
```
0xdevalias, line 104: See comment in this review related to this for more context.

```suggestion
      const day = addDays(firstDayOfSchedule, dayIndex);
      const formatDayLabel = (day: Date | number) => {
        if (isScheduleTimeshifted) {
          return format(day, "E, LLL d");
        } else {
          return dayIndex === 0 ? "Today" : format(day, "E");
        }
      }
```

0xdevalias: Potentially slightly more robust, instead of following the assumption that `dayIndex === 0` will imply today when not time shifted, we could be explicit about it, which might make the code easier to read.

```suggestion
      const day = addDays(firstDayOfSchedule, dayIndex);
      const formatDayLabel = (day: Date | number) => {
        if (isScheduleTimeshifted) {
          return format(day, "E, LLL d");
        } else {
          return isToday(day) ? "Today" : format(day, "E");
        }
      }
```

0xdevalias (changes requested):
[CRO] I haven't clicked through this to check that it works, but I believe these 2 small changes (which can likely be applied as is) should be all that's needed before this can be merged?

@mike-lvov If you have the capacity to check this and apply the changes, happy for you to handle approve+merge from there.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: done

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
margulies: brought back in `isScheduledTimeshrifted`. done

mike-lvov (approved):
LGTM. Didn't click-through though.

sunny-viktoryia (approved):
checked the logic and run the code locally. Looks good to me.
Great enhancement! Thank you @margulies ! :heart:

0xdevalias:
> LGTM. Didn't click-through though.

> checked the logic and run the code locally. Looks good to me.
> Great enhancement! Thank you @margulies ! ❤️

@sunny-viktoryia @mike-lvov Awesome! Thanks for your reviews/checks!

@margulies As @sunny-viktoryia already said, thanks for the awesome contribution here! 🎉

PR #1420, opened 2021-05-19
add OHBM Screening room

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/572
closes https://github.com/sparkletown/internal-sparkle-issues/issues/735
closes https://github.com/sparkletown/internal-sparkle-issues/issues/737
closes https://github.com/sparkletown/internal-sparkle-issues/issues/738

TODOs:
- [x] Replace mock data with Firestore
- [x] Playing now component


src/components/templates/ScreeningRoom/useScreeningRoom.ts
```diff
@@ -0,0 +1,148 @@
+import { useState, useMemo, useCallback } from "react";
```
denisdimitrov, line 1: Is this file/hook intentionally in `/ScreeningRoom/`

src/components/templates/ScreeningRoom/useScreeningRoom.ts
```diff
@@ -0,0 +1,148 @@
...
+    },
+  ]);
+};
+
+export const useScreeningRoomVideos = (screeningRoomVenueId: string) => {
+  useConnectScreeningRoomVideos(screeningRoomVenueId);
+
+  const screeningRoomVideos = useSelector(screeningRoomVideosSelector);
+
+  console.log({ screeningRoomVenueId, screeningRoomVideos });
```
denisdimitrov, line 31: log

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
```diff
@@ -0,0 +1,153 @@
...
+        iconStart={faSearch}
+        placeholder="Search for a talk..."
+        value={searchInputValue}
+        onChange={onInputFieldChange}
+      />
+      {renderedCategoryOptions}
+      {renderedSubCategoryOptions}
+
+      <div className="ScreeningRoom__video-previews">
+        {isVideosLoaded ? renderedVideoPreviews : "Loading videos"}
```
denisdimitrov, line 141: Can be extracted as a const

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import { IFRAME_ALLOW } from "settings";
+
+import { GenericVenue } from "types/venues";
+import { ScreeningRoomVideo } from "types/screeningRoom";
+
+import { WithId } from "utils/id";
+
+import { useScreeningRoom } from "./useScreeningRoom";
```
denisdimitrov, line 12: Should this go to the bottom of the imports since its in the same dir?

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
```diff
@@ -0,0 +1,153 @@
...
+    [setSearchInputValue]
+  );
+
+  const renderedVideoPreviews = useMemo(
+    () =>
+      videos.map((video) => (
+        <ScreeningVideoPreview
+          key={video.id}
+          video={video}
+          selectThisVideo={() => setSelectedVideo(video)}
```
denisdimitrov, line 66: Maybe useCallback + func?

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }, [categoryList, categoryFilter, setCategoryFilter]);
+
+  const renderedSubCategoryOptions = useMemo(
+    () => (
+      <div className="ScreeningRoom__subcategories">
+        {subCategoryList.map((subCategory) => (
+          <PosterCategory
+            key={subCategory}
+            category={subCategory}
+            onClick={() => setSubCategoryFilter(subCategory)}
```
denisdimitrov, line 96: Same here.

denisdimitrov (changes requested):

src/components/templates/ScreeningRoom/useScreeningRoom.ts
mike-lvov: Yeah, not sure if this is the best approach, but it's only used here, so I decided there was no sense to move it into hooks/

src/components/templates/ScreeningRoom/useScreeningRoom.ts
mike-lvov: Sorry, it was an outdated code. It's gone now

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
mike-lvov: It's a simple ternary and we reuse this kind of codestyle across the codebase. Not sure, if this will add to readibility

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
mike-lvov: It's useMemoed 
image

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
mike-lvov: It's useMemoed
image

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1420/commits/a548b05be944d45a37282dc7de1064038aa08f7f

denisdimitrov (approved):
LGTM!

src/components/templates/ScreeningRoom/useScreeningRoom.ts
0xdevalias: I'm fine with this pattern for 'local dependencies', we follow it in a few places already, and it aligns closer with (what I feel is) a superior pattern of 'code layout', which is 'feature focussed' rather than 'type of thing focussed'.

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
0xdevalias: @denisdimitrov @mike-lvov FYI: For short '1 element' ternaries like this, I don't mind them being in the JSX directly since it's simple enough to 'take it all in' in a single glance. Where I have issue with using ternaries in JSX is when it's a whole giant block of code on one side or the other of it, as that becomes hard to read.

If I was to try and convert this to a non-ternary form, off the top of my head, I would probably end up moving the 'Loading videos' inside the same (presumably memo'd) code as `renderedVideoPreviews`, and return it as the result of the 'guard statement'.

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
0xdevalias: > Should this go to the bottom of the imports since its in the same dir

@denisdimitrov @mike-lvov Yup. Though still in the same vague 'hierarchy' within that 'local imports' (`./`) section (so hooks, then components, then scss, etc)

(the change @mike-lvov made is correct, I just wanted to be clear about the answer so you know for sure in future)

PR #1423, opened 2021-05-19
add full/personal schedule export as static .ics (calendar) files

htwangtw:
I received an invitation from @margulies to take on this challenge...
We would like to make a calender exporter for users' personal schedule.
Currently I have successfully made a button that downloads ics file on click. For now I am using some random number as the time. I have marked all the relevant fields needed to be changed.

From my understanding, there's already some features to generate personal schedule in sparkle. 
The current bottle neck for me is to find out how to access the personal schedule, how's the information structured, and how to parse it.
I will also need to look into the ics file format a bit more to understand how multiple events are stored.
Any pointers are welcome!

**To-do**
- [x] find out how to access event info
- [x] better output file name
- [x] ics file for mutiple event - what does it look like so I can figure out how to parse the events


src/components/organisms/NavBarSchedule/ics.js
```diff
@@ -0,0 +1,290 @@
+/* global saveAs, BlobBuilder */
+/* exported ics */
```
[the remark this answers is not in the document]
0xdevalias: Based on the above, it looks like we probably want to use

- https://bundlephobia.com/result?p=ics@2.27.0
  - https://www.npmjs.com/package/ics
  - https://github.com/adamgibbons/ics

`npm install ics`

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const createCalendar = (
+  allEvents: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[],
+  filename: string
+) => {
+  const cal = ics();
+  if (!cal) return;
+  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) => {
+    return cal.addEvent(
+      event.id!,
```
0xdevalias, line 16: We don't use TypeScript's `!` operator in our codebase as it breaks type safety.

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
+import { eventEndTime, eventStartTime } from "utils/event";
```
0xdevalias, line 1: This file should probably be in `utils/calendar` or similar rather than here?

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+import { PersonalizedVenueEvent } from "types/venues";
+import { VenueEvent } from "types/venues";
+import { WithVenueId } from "utils/id";
+
+import { ics } from "./ics";
```
0xdevalias, line 6: This should be included from a proper `npm install`'d dependency.

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+  filename: string
+) => {
+  const cal = ics();
+  if (!cal) return;
+  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) => {
+    return cal.addEvent(
+      event.id!,
+      event.name,
+      event.host || "",
+      window.location.origin.toString() + "/in/" + event.venueId,
```
0xdevalias, line 19: Look in `utils/url` for this sort of thing. Our preference would be to use a pattern with `new URL()`. While it's not in `staging` yet, there is a new function for exactly this purpose in https://github.com/sparkletown/sparkle/pull/1396

- https://github.com/sparkletown/sparkle/pull/1396/files#diff-aab4ceb6d3254a7b948f95d8057cd8d91cd43c09f95569885cbce6d8b079d926
- (see also my review suggestion for renaming it https://github.com/sparkletown/sparkle/pull/1396/files#r635905264)

```typescript
export const getFullVenueInsideUrl = (venueId: string) =>
  new URL(venueInsideUrl(venueId), window.location.origin).href;
```

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ics } from "./ics";
+
+export const createCalendar = (
+  allEvents: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[],
+  filename: string
+) => {
+  const cal = ics();
+  if (!cal) return;
+  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) => {
+    return cal.addEvent(
```
0xdevalias, line 15: By getting rid of the `{` you can also get rid of `return`.

```suggestion
  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) =>
    cal.addEvent(
```

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+      event.id!,
+      event.name,
+      event.host || "",
```
0xdevalias, line 18: We default with `??` not `||`.

```suggestion
      event.host ?? "",
```

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+      window.location.origin.toString() + "/in/" + event.venueId,
+      formatDate(eventStartTime(event)),
+      formatDate(eventEndTime(event)),
+      ""
```
0xdevalias, line 22: What is this extra empty string here for?

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+    );
+  });
+  cal.download(filename + "_calendar", ".ics");
```
0xdevalias, line 25: Does this trigger the actual file download?

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+    pad(dateTime.getUTCDate()),
+    "T",
+    pad(dateTime.getUTCHours()),
+    pad(dateTime.getUTCMinutes()) + "00Z",
+  ].join("");
+}
+
+function pad(num: string | number) {
+  return num < 10 ? "0" + num : num;
+}
```
0xdevalias, line 42: We use `date-fns` for our date formatting, and have a number of helper functions already defined in `utils/time`.

I expect there's a better way to solve this need using those.

```suggestion
```

src/components/organisms/NavBarSchedule/calendar.ts
0xdevalias, line 1: Sort `utils/*` imports with other `utils/*` imports (after `types/*` imports)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +140,30 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     };
   }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
 
+  const downloadCalenderPersonal = useCallback(() => {
+    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
+      .map(prepareForCalendar(userEventIds))
+      .filter((event) => event.isSaved);
+    createCalendar(allPersonalEvents, "Personal");
+  }, [relatedVenueEvents, userEventIds]);
```
0xdevalias, line 148: Whitespace makes code easier to read.

```suggestion
  const downloadCalenderPersonal = useCallback(() => {
    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
      .map(prepareForCalendar(userEventIds))
      .filter((event) => event.isSaved);

    createCalendar(allPersonalEvents, "Personal");
  }, [relatedVenueEvents, userEventIds]);
```

src/components/organisms/NavBarSchedule/calendar.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { ics } from "./ics";
+
+export const createCalendar = (
+  allEvents: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[],
+  filename: string
+) => {
+  const cal = ics();
+  if (!cal) return;
+  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) => {
```
0xdevalias, line 14: Whitespace makes code easier to read.

```suggestion
) => {
  const cal = ics();
  
  if (!cal) return;
  
  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) => {
```

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -31,6 +31,22 @@ export const prepareForSchedule = (
...
+
+  return {
+    ...event,
+    start_utc_seconds: getUnixTime(startOfEventToShow),
+    duration_minutes: differenceInMinutes(endOfEventToShow, startOfEventToShow),
+    isSaved: isTruthy(
+      event.id && usersEvents[event.venueId]?.includes(event.id)
+    ),
+  };
+};
```
0xdevalias, line 48: What is this function adding/doing differently (and why) compared to `prepareForSchedule` above? Is it just the differences in `startOfEventToShow` / `endOfEventToShow`?

It would be good to avoid duplicating code here.

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -31,6 +31,22 @@ export const prepareForSchedule = (
   };
 };
 
+export const prepareForCalendar = (usersEvents: MyPersonalizedSchedule) => (
+  event: WithVenueId<VenueEvent>
+): PersonalizedVenueEvent => {
+  const startOfEventToShow = eventStartTime(event);
+  const endOfEventToShow = eventEndTime(event);
+
+  return {
```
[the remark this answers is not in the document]
0xdevalias: See my comment below.

0xdevalias (changes requested):
Hey @htwangtw, thanks for this PR!

I've gone through and left a bunch of comments/questions/suggestions for how we can align this closer to our code standards and get it merged.

Let me know if you have any questions/concerns.

src/components/organisms/NavBarSchedule/ics.js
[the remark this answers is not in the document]
0xdevalias: Actually.. somehow I missed this one in my exploration above, which also seems like it might be a decent choice:

- https://bundlephobia.com/result?p=ical-generator@2.1.0 `8.9kB`
  - https://github.com/sebbo2002/ical-generator
    - last updated 7 days ago

src/components/organisms/NavBarSchedule/calendar.ts
htwangtw: For now I just added this new function in the same file, I hope that's okay.

htwangtw:
Thanks for the detailed review from @0xdevalias. The comments helped a lot.
I have implemented the suggested changes but introduced some issues that will really appreciate some inputs on.

- File saving: I fall back to use `window.open` to export the URL file. The filename is not as pretty. Would be helpful to have a better way to save the file.
- Adding more information to calendar event: adding `event.host` introduces errors. I am still trying to work on this and include more fields, but if anyone has an idea that would be helpful too.

yarikoptic:
I don't want to derail but wondered: is there a way to account for unbookmarked (thus kinda deleted) events? I emailed or posted my earlier investigation somewhere (but failed to quickly find it): if we have information which event was at some point added but then "removed"  -- iirc there is a way to annotate that in .ics so upon re-import calendar should account for that.  I could dig out more if there is a faint possibility for that to get supported (i.e. the way events or exported calendars stored).

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +139,32 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
+      .map(prepareForSchedule(dayStart, userEventIds, "calender"))
+      .filter((event) => event.isSaved);
+
+    createCalendar(allPersonalEvents, "Personal");
+  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
+
+  const downloadCalenderAll = useCallback(() => {
+    createCalendar(relatedVenueEvents, "Full");
```
yarikoptic, line 152: recommendation: prefix with `PLATFORM_BRAND_NAME` from `src/settings.ts` (which will be `OHBM2021` in `env/ohbm` branch) and `_`.  This way filename would be more descriptive.  Similarly for personal schedule

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+      start: eventStartTime(event),
+      end: eventEndTime(event),
+      description: event.host,
+      summary: event.name ? event.host : "",
+      url: getFullVenueInsideUrl(event.venueId),
+    })
+  );
+
+  const outputCal = cal.toURL();
+  const outputFile = `${calendername}_calendar.ics`;
```
yarikoptic, line 29: IMHO `_calendar` suffix is not needed since implied by `.ics` extension.
```suggestion
  const outputFile = `${calendername}.ics`;
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +139,32 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     };
   }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
 
+  const downloadCalenderPersonal = useCallback(() => {
```
yarikoptic, line 142: here and elsewhere a typo - should be  Calendar not Calender.
```
$> git grep -i calender
src/components/organisms/NavBarSchedule/NavBarSchedule.tsx:  const downloadCalenderPersonal = useCallback(() => {
src/components/organisms/NavBarSchedule/NavBarSchedule.tsx:      .map(prepareForSchedule(dayStart, userEventIds, "calender"))
src/components/organisms/NavBarSchedule/NavBarSchedule.tsx:  const downloadCalenderAll = useCallback(() => {
src/components/organisms/NavBarSchedule/NavBarSchedule.tsx:        <Button onClick={downloadCalenderPersonal}>
src/components/organisms/NavBarSchedule/NavBarSchedule.tsx:        <Button onClick={downloadCalenderAll}>Download full schedule</Button>
src/components/organisms/NavBarSchedule/utils.ts:  if (target === "calender") {
src/utils/calendar.ts:  calendername: string
src/utils/calendar.ts:  const cal = ical({ name: calendername });
src/utils/calendar.ts:  const outputFile = `${calendername}_calendar.ics`;
```

smth like
```
git grep -l alender | xargs sed -i -e 's,alender,alendar,g'
```
should fix them all

yarikoptic (changes requested):
at least the typo needs to be fixed ;-)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
htwangtw: I was wondering if that's possible! Thanks for pointing out where this info sits!

yarikoptic:
another wishlist (to be ignored or not) from ignorant in all things web me:  since it is all clicky buttons and not just a URL - can't check if it provides proper annotation somehow that it is a `webcal`.  In principle, with a proper annotation of the type browsers should be able to handle it as not to download but possibly react and import to the configured calendar app/site?

may be [this](https://stackoverflow.com/questions/5329529/i-want-html-link-to-ics-file-to-open-in-calendar-app-when-clicked-currently-op/7532838) could be of help which states: "Ok here's the thing: Chrome recognizes it as well if you put it into a link (as opposed to putting it directly into address bar and hitting enter): `<a href="webcal://path/to/.ics">sub link</a>` – "

so if there was "Download" (the one I guess now, with all the content-disposition filename) and "Export" (which does the webcal:// trick somehow) -- that would have been awesome. May be that library you use has some docs/API for that?

htwangtw:
> another wishlist (to be ignored or not) from ignorant in all things web me: since it is all clicky buttons and not just a URL - can't check if it provides proper annotation somehow that it is a `webcal`. In principle, with a proper annotation of the type browsers should be able to handle it as not to download but possibly react and import to the configured calendar app/site?
> 
> may be [this](https://stackoverflow.com/questions/5329529/i-want-html-link-to-ics-file-to-open-in-calendar-app-when-clicked-currently-op/7532838) could be of help which states: "Ok here's the thing: Chrome recognizes it as well if you put it into a link (as opposed to putting it directly into address bar and hitting enter): `<a href="webcal://path/to/.ics">sub link</a>` – "
> 
> so if there was "Download" (the one I guess now, with all the content-disposition filename) and "Export" (which does the webcal:// trick somehow) -- that would have been awesome. May be that library you use has some docs/API for that?

I see the difference. From what I understand, [ics-generator](https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalcalendar.html#toblob) has really straightforward API for generating a URL or sting, but nothing beyond that. I can see if I can modify the URL with the `webcal` annotation? This is the relevant chunch of code in my current PR:
```typescript
  const outputCal = cal.toURL();
  const outputFile = `${calendarname}.ics`;
  const link = document.createElement("a");
  link.download = outputFile;
  link.href = outputCal;
  link.click();
```

0xdevalias:
@yarikoptic @htwangtw Off the top of my head, to make it a webcam subscribe me URL we would need to generate it from a backend function rather than the frontend code like this. It would likely be practically the same code as used on the frontend to do so (but given the somewhat inefficient way we have to query related venues and events of them right now due to our current data structures/architecture, it’s probably not worth implementing in a backend function now, as it would drive costs up.

src/components/organisms/NavBarSchedule/ics.js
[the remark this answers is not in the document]
0xdevalias: Awesome, sounds good :)

src/components/organisms/NavBarSchedule/calendar.ts
0xdevalias: Yup, that works well :) That way when it comes time to merge things together, git should just see it as already existing code/similar, and 'just work' without us having to do anything more :)

htwangtw:
@0xdevalias Please let me know if there's more to change, if we are not doing the `webcal` export.
Just making sure I didn't drop the ball on my side.

0xdevalias:
> Please let me know if there's more to change, if we are not doing the `webcal` export.
> Just making sure I didn't drop the ball on my side.

@htwangtw I'll let you know if there's anything else needed. Just been super busy and not had a chance to get back to doing a full review of the latest changes in this PR yet.

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1423#issuecomment-846019228

> may be [this](https://stackoverflow.com/questions/5329529/i-want-html-link-to-ics-file-to-open-in-calendar-app-when-clicked-currently-op/7532838) could be of help which states: "Ok here's the thing: Chrome recognizes it as well if you put it into a link (as opposed to putting it directly into address bar and hitting enter): `<a href="webcal://path/to/.ics">sub link</a>` – "

@yarikoptic @htwangtw Re-reading this now, I see that I read it wrong originally. My reply in https://github.com/sparkletown/sparkle/pull/1423#issuecomment-846369903 was making the assumption we wanted to have 'live subscribable webcals', rather than just using that as the prefix to trigger calendar apps for import/etc.

---

RE: https://github.com/sparkletown/sparkle/pull/1423#issuecomment-846248593

> I see the difference. From what I understand, [ics-generator](https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalcalendar.html#toblob) has really straightforward API for generating a URL or sting, but nothing beyond that (as far as I can tell). I can see if I can modify the URL with the `webcal` annotation? This is the relevant chunch of code in my current PR:
> 
> ```ts
>   const outputCal = cal.toURL();
>   const outputFile = `${calendarname}.ics`;
>   const link = document.createElement("a");
>   link.download = outputFile;
>   link.href = outputCal;
>   link.click();
> ```

@yarikoptic @htwangtw What does `cal.toURL()` generate at the moment? That would determine whether it's likely that it could be changed to use a `webcal://` prefix.

---

@yarikoptic @htwangtw Will do another full pass of review now as well.

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -16,10 +16,19 @@ import { isTruthy } from "utils/types";
 
 export const prepareForSchedule = (
   day: Date,
-  usersEvents: MyPersonalizedSchedule
+  usersEvents: MyPersonalizedSchedule,
+  target: string
 ) => (event: WithVenueId<VenueEvent>): PersonalizedVenueEvent => {
```
0xdevalias, line 21: Since we have so many parameters here now, it would be good to use an object with named params rather than these 3 positional params.

The change will end up something like:

```typescript
export interface PrepareForScheduleProps {
  day: Date;
  usersEvents: MyPersonalizedSchedule;
  isForCalendarFile: boolean;
}

export const prepareForSchedule = ({
  day,
  usersEvents,
  isForCalendarFile,
}: PrepareForScheduleProps) => (event: WithVenueId<VenueEvent>): PersonalizedVenueEvent => {
```

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -16,10 +16,19 @@ import { isTruthy } from "utils/types";
...
+  let startOfEventToShow;
+  let endOfEventToShow;
+
+  if (target === "calendar") {
+    startOfEventToShow = eventStartTime(event);
+    endOfEventToShow = eventEndTime(event);
+  } else {
+    startOfEventToShow = max([eventStartTime(event), startOfDay(day)]);
+    endOfEventToShow = min([eventEndTime(event), endOfDay(day)]);
+  }
```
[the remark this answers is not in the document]
0xdevalias: note that with my other suggested change in this review, the ternary's would end up using `isForCalendarFile` instead of `target === "calendar"`

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
+import { PersonalizedVenueEvent } from "types/venues";
+import { VenueEvent } from "types/venues";
+import { eventEndTime, eventStartTime } from "utils/event";
+import { WithVenueId } from "utils/id";
+import { getFullVenueInsideUrl } from "utils/url";
+
+import ical from "ical-generator";
```
0xdevalias, line 7: External lib imports at the top, whitespace between `types/*` and `utils/*`, and combined imports from the same file:

```suggestion
import ical from "ical-generator";

import { PersonalizedVenueEvent, VenueEvent } from "types/venues";

import { eventEndTime, eventStartTime } from "utils/event";
import { WithVenueId } from "utils/id";
import { getFullVenueInsideUrl } from "utils/url";
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const createCalendar = (
+  allEvents: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[],
+  calendarname: string
+) => {
+  const cal = ical({ name: calendarname });
```
0xdevalias, line 13: Would be good to use named props here as well:

```suggestion
export interface CreateCalendarProps {
  calendarName: string,
  events: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[]
}

export const createCalendar = ({
  calendarName,
  events,
}: CreateCalendarProps) => {
  const cal = ical({ name: calendarName });
```

0xdevalias: It would also be good to set an explicit return type on this function.

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!cal) return;
+
+  allEvents.map((event: PersonalizedVenueEvent | WithVenueId<VenueEvent>) =>
```
0xdevalias, line 17: Since we're using the 'side effect' of processing these, rather than the return value, we should use `forEach` here:

I think we should also be able to leave off the explicit typing and just let TypeScript infer it:

```suggestion
  allEvents.forEach((event) =>
```

Further Reading:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+    cal.createEvent({
+      id: event.venueId,
```
0xdevalias, line 19: I think we'll need to use something different for `id` here, since `event.venueId` isn't going to be unique.

I don't believe we have a unique `id` representing our events currently. Not sure what to use here, as i'm not sure what the calendar 'does with it'.

Further Reading:

- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalevent.html#id

---

@mike-lvov Any thoughts?

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+      start: eventStartTime(event),
+      end: eventEndTime(event),
+      description: event.host,
```
0xdevalias, line 22: Might be able to put `event.host` in `organizer`?

- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalevent.html#organizer

---

You can find the `VenueEvent` type definition in `src/types/venues.ts`

It has a `description` field, so that would probably be good to include here.

- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalevent.html#description

---

```suggestion
      organizer: event.host,
      description: event.description,
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+      summary: event.name ? event.host : "",
```
0xdevalias, line 23: What's the intent of this ternary here? Currently it will check if `event.name` is truthy, and if it is, use `event.host`. If `event.name` is falsy it will use `""`.

I expect what we want here is to use event.name if present, if not fall back to event.host, and if not that then fall back to the empty string (or similar)?

If so, that would be implemented using TypeScript's "default the value if empty" `??` (nullish coalescing) operator as follows:

```suggestion
      summary: event.name ?? event.host ?? "",
```

Though looking at the definition of `VenueEvent` in `src/types/venues.ts`, `name`/`host` can never be `undefined` anyway.. so not sure what the intent of this is, we can probably just be as simple as this here:

```suggestion
      summary: event.name,
```

Further Reading:

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalevent.html#summary

src/utils/calendar.ts
0xdevalias: Since `PersonalizedVenueEvent` (defined in `src/types/venues.ts`) already extends from `WithVenueId<VenueEvent>`, unless we need the `isSaved` property that it adds (which it doesn't look like we do), we can just simplify the type this function accepts to:

```typescript
export interface CreateCalendarProps {
  calendarName: string,
  events: WithVenueId<VenueEvent>[]
}
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+      url: getFullVenueInsideUrl(event.venueId),
+    })
+  );
+
+  const outputCal = cal.toURL();
+  const outputFile = `${calendarname}.ics`;
+  const link = document.createElement("a");
+  link.download = outputFile;
+  link.href = outputCal;
+  link.click();
```
0xdevalias, line 33: This part probably belongs in a different function (maybe called `downloadCalendar`). `createCalendar` should just create and `return` the calendar object (with the added events).

We could have a new function that accepts a calendar object, and then triggers this download (`downloadCalendar`)

0xdevalias: See also the discussion in https://github.com/sparkletown/sparkle/pull/1423#issuecomment-848630538 for what the `cal.toURL()` produces here, and whether we can make that work with a `webcal://` prefix or similar.

Further Reading:

- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalcalendar.html#tourl
- https://sebbo2002.github.io/ical-generator/develop/reference/classes/icalcalendar.html#toblob
- https://developer.mozilla.org/en-US/docs/Web/API/Blob
- https://developer.mozilla.org/en-US/docs/Web/API/Blob#creating_a_url_representing_the_contents_of_a_typed_array
- https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL
- https://developer.mozilla.org/en-US/docs/Web/API/File/Using_files_from_web_applications#using_object_urls

src/utils/calendar.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import ical from "ical-generator";
+
+export const createCalendar = (
+  allEvents: PersonalizedVenueEvent[] | WithVenueId<VenueEvent>[],
+  calendarname: string
+) => {
+  const cal = ical({ name: calendarname });
+
+  if (!cal) return;
```
0xdevalias, line 15: Is this needed here? Can `cal` be `undefined`/etc? If so, under what circumstances?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +139,32 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
   }, [relatedVenueEvents, userEventIds, selectedDayIndex, getEventLocation]);
 
+  const downloadCalendarPersonal = useCallback(() => {
+    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
+      .map(prepareForSchedule(dayStart, userEventIds, "calendar"))
+      .filter((event) => event.isSaved);
+
+    createCalendar(allPersonalEvents, `${PLATFORM_BRAND_NAME}_Personal`);
+  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
```
0xdevalias, line 149: There will be some minimal changes required here based on other suggestions in this review (mostly around the `prepareForSchedule` props/args)

As well as to call the separated `createCalendar` and the proposed `downloadCalendar` functions.

0xdevalias: Note also, when https://github.com/sparkletown/sparkle/pull/1419 is merged there will likely be some tweaks required here to use the correct date.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -117,7 +122,7 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
     const dayStart = addDays(startOfToday(), selectedDayIndex);
     const daysEvents = relatedVenueEvents
       .filter(isEventWithinDate(selectedDayIndex === 0 ? Date.now() : dayStart))
-      .map(prepareForSchedule(dayStart, userEventIds));
+      .map(prepareForSchedule(dayStart, userEventIds, "web"));
```
0xdevalias, line 125: There will be some minimal changes required here based on other suggestions in this review (mostly around the `prepareForSchedule` props/args)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -37,6 +39,9 @@ import {
   extractLocation,
   prepareForSchedule,
 } from "./utils";
+
+import { createCalendar } from "utils/calendar";
+
 import { isEventWithinDate } from "utils/event";
 import { WithVenueId } from "utils/id";
```
0xdevalias, line 46: All of these `utils/*` imports (not the `./utils` import) should be moved up after `settings` import, before the `hook/*` imports, with a single line of whitespace before/after the `utils/*` block of imports (and no spaces in between them)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +139,32 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
+      .map(prepareForSchedule(dayStart, userEventIds, "calendar"))
+      .filter((event) => event.isSaved);
+
+    createCalendar(allPersonalEvents, `${PLATFORM_BRAND_NAME}_Personal`);
+  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
+
+  const downloadCalendarAll = useCallback(() => {
+    createCalendar(relatedVenueEvents, `${PLATFORM_BRAND_NAME}_Full`);
+  }, [relatedVenueEvents]);
```
0xdevalias, line 153: There will be some minimal changes required here based on other suggestions in this review to call the separated `createCalendar` and the proposed `downloadCalendar` functions.

0xdevalias: Renamed: `downloadAllEventsCalendar`
```suggestion
  const downloadAllEventsCalendar = useCallback(() => {
    createCalendar(relatedVenueEvents, `${PLATFORM_BRAND_NAME}_Full`);
  }, [relatedVenueEvents]);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: Renamed `downloadPersonalEventsCalendar`:
```suggestion
  const downloadPersonalEventsCalendar = useCallback(() => {
    const dayStart = addDays(startOfToday(), selectedDayIndex);
    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
      .map(prepareForSchedule(dayStart, userEventIds, "calendar"))
      .filter((event) => event.isSaved);

    createCalendar(allPersonalEvents, `${PLATFORM_BRAND_NAME}_Personal`);
  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
```

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -39,6 +39,19 @@ $weekday--margin-right: 10px;
...
+    cursor: pointer;
+    padding-left: $NavBarSchedule--padding;
+    opacity: 0.8;
+    text-shadow: 0 2px 4px rgba($black, 0.3);
+    box-shadow: 0 3px 6px rgba($black, 0.2);
+
+    &:hover {
+      opacity: 1;
+    }
+  }
```
0xdevalias, line 53: [note] I haven't deeply reviewed these styles visually, they look frome a code point of view though.

0xdevalias (changes requested):
[CRO] A bunch of changes, most of them should be pretty minor/quick to get fixed up, then I think this PR will be good to go.

@mike-lvov I haven't clicked through this yet, once the changes are made, if you get a chance to look at it before I do, can you click through and check that it works as expected, and if so, feel free to approve and merge.

src/utils/calendar.ts
htwangtw: Good point - this was some left over from I was testing with raw ics string

src/utils/calendar.ts
htwangtw: I will remove it for now since you mention it's not a unique number

htwangtw:
I decided to drop the `webcal` protocol improvement. See commit message ec68e83 for details. But the rest of the changes are addressed. (Sorry @yarikoptic)
One remaining issue is the calendar library seems to require supplying along a name for the organiser info. This is an issue of the upstream so I just set email to undefined as default.

package.json
```diff
@@ -31,6 +31,7 @@
     "firebase": "^7.14.2",
     "fuse.js": "^6.4.1",
     "husky": "^4.2.5",
+    "ical-generator": "^2.1.0",
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +139,32 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
 
   return (
     <div className={containerClasses}>
       {venueId && <ScheduleVenueDescription venueId={venueId} />}
+      <ul className="NavBarSchedule__downloads">
+        <Button onClick={downloadCalendarPersonal}>
+          Download your schedule
+        </Button>
+        <Button onClick={downloadCalendarAll}>Download full schedule</Button>
+      </ul>
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/utils.ts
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/utils.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: Sounds good. Does the `ical` lib need us to set it? Or will it just generate its own internally if required? Given it depends on a `uuid` lib, i'm guessing it will probably generate it on it's own if needed.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/utils/calendar.ts
0xdevalias: Makes sense :)

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/ec68e834fab42e9549a285d7a454e4c29e1f24e4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +144,46 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
+      .map(
+        prepareForSchedule({
+          day: dayStart,
+          usersEvents: userEventIds,
+          isForCalendarFile: true,
+        })
+      )
+      .filter((event) => event.isSaved);
```
0xdevalias, line 157: I'm guessing `.isSaved` only becomes available here after the `.map`? If so, all good. If not, we should move the `.filter` before the `.map`

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -14,12 +14,25 @@ import { WithVenueId } from "utils/id";
...
+  day: Date;
+  usersEvents: MyPersonalizedSchedule;
+  isForCalendarFile: boolean;
+}
+
+export const prepareForSchedule = ({
+  day,
+  usersEvents,
+  isForCalendarFile,
+}: PrepareForScheduleProps) => (
```
0xdevalias, line 27: We can default `isForCalendarFile` here:

```suggestion
export const prepareForSchedule = ({
  day,
  usersEvents,
  isForCalendarFile = false,
}: PrepareForScheduleProps) => (
```

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -14,12 +14,25 @@ import { WithVenueId } from "utils/id";
...
-  day: Date,
-  usersEvents: MyPersonalizedSchedule
-) => (event: WithVenueId<VenueEvent>): PersonalizedVenueEvent => {
-  const startOfEventToShow = max([eventStartTime(event), startOfDay(day)]);
-  const endOfEventToShow = min([eventEndTime(event), endOfDay(day)]);
+export interface PrepareForScheduleProps {
+  day: Date;
+  usersEvents: MyPersonalizedSchedule;
+  isForCalendarFile: boolean;
+}
```
0xdevalias, line 21: We can make `isForCalendarFile` an optional prop here:
```suggestion
export interface PrepareForScheduleProps {
  day: Date;
  usersEvents: MyPersonalizedSchedule;
  isForCalendarFile?: boolean;
}
```

src/components/organisms/NavBarSchedule/utils.ts
```diff
@@ -14,12 +14,25 @@ import { WithVenueId } from "utils/id";
...
+  day,
+  usersEvents,
+  isForCalendarFile,
+}: PrepareForScheduleProps) => (
+  event: WithVenueId<VenueEvent>
+): PersonalizedVenueEvent => {
+  const startOfEventToShow = isForCalendarFile
+    ? eventStartTime(event)
+    : max([eventStartTime(event), startOfDay(day)]);
+  const endOfEventToShow = isForCalendarFile
```
0xdevalias, line 33: Let's add a newline here to make it a bit easier to read:
```suggestion
    : max([eventStartTime(event), startOfDay(day)]);

  const endOfEventToShow = isForCalendarFile
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
+import ical from "ical-generator";
+
+import { VenueEvent } from "types/venues";
+
+import { eventEndTime, eventStartTime } from "utils/event";
+
+import { WithVenueId } from "utils/id";
+
+import { getFullVenueInsideUrl } from "utils/url";
```
0xdevalias, line 9: I thought my description was probably going to end up too verbose/complex to be understandable 😛 My intent was to say that we don't need the newlines between these individual imports (just above/below the block of `utils/url` imports as a whole):

```suggestion
import { eventEndTime, eventStartTime } from "utils/event";
import { WithVenueId } from "utils/id";
import { getFullVenueInsideUrl } from "utils/url";
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface CreateCalendarProps {
+  calendarName: string;
+  events: WithVenueId<VenueEvent>[];
+}
+
+const createCalendar = ({ calendarName, events }: CreateCalendarProps) => {
```
0xdevalias, line 16: We'll want to replace the `???` with an explicit type here (whatever type `cal` is)

```suggestion
const createCalendar = ({ calendarName, events }: CreateCalendarProps): ??? => {
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+    cal.createEvent({
+      start: eventStartTime(event),
+      end: eventEndTime(event),
+      organizer: `${event.host} <undefined>`, // string format: "name <email>". email cannot be blank
+      description: event.description,
+      summary: event.name,
+      url: getFullVenueInsideUrl(event.venueId),
+    })
+  );
+  return cal;
```
0xdevalias, line 29: [nit]
```suggestion
  );

  return cal;
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const downloadCalendar = ({
+  calendarName,
+  events,
+}: CreateCalendarProps) => {
+  const cal = createCalendar({ calendarName: calendarName, events: events });
```
0xdevalias, line 36: My intent here was that we would pass the already created calendar object as the only param here (replace `???` with the correct type for the generated calendar, this will be the same value used for the `???` for `createCalendar`'s return type:

```suggestion
export const downloadCalendar = (calendar: ???): void => {
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+  const outputCal = cal.toURL();
+  const outputFile = `${calendarName}.ics`;
```
0xdevalias, line 38: I think this should be equivalent? Also renamed to `outputFileName`

```suggestion
  const outputFileName = `${calendar.name}.ics`;
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+  const link = document.createElement("a");
```
0xdevalias, line 39: [nit] Let's add a newline here to keep the 'related chunks' separated:
```suggestion

  const link = document.createElement("a");
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+  link.download = outputFile;
+  link.href = outputCal;
+  console.log(link.href);
```
0xdevalias, line 42:
```suggestion
```

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+  return cal;
+};
+
+export const downloadCalendar = ({
+  calendarName,
+  events,
+}: CreateCalendarProps) => {
+  const cal = createCalendar({ calendarName: calendarName, events: events });
+  const outputCal = cal.toURL();
```
0xdevalias, line 37: Not sure if we will need to explicitly release this from memory once we are done with it, or if the lib will handle that for us:

- https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL#memory_management
- https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL

src/utils/calendar.ts
```diff
@@ -0,0 +1,44 @@
...
+  const outputFile = `${calendarName}.ics`;
+  const link = document.createElement("a");
+  link.download = outputFile;
```
0xdevalias, line 40:
```suggestion
  link.download = outputFileName;
```

src/utils/calendar.ts
0xdevalias: Then, where we call this in `src/components/organisms/NavBarSchedule/NavBarSchedule.tsx`, we would call both `createCalendar` and then `downloadCalendar`

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +144,46 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+
+    downloadCalendar({
+      calendarName: `${PLATFORM_BRAND_NAME}_Personal`,
+      events: allPersonalEvents,
+    });
```
0xdevalias, line 162:
```suggestion
    downloadCalendar(createCalendar({
      calendarName: `${PLATFORM_BRAND_NAME}_Personal`,
      events: allPersonalEvents,
    }));
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -134,13 +144,46 @@ export const NavBarSchedule: FC<NavBarScheduleProps> = ({ isVisible }) => {
...
+  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
+
+  const downloadAllEventsCalendar = useCallback(() => {
+    downloadCalendar({
+      calendarName: `${PLATFORM_BRAND_NAME}_Full`,
+      events: relatedVenueEvents,
+    });
```
0xdevalias, line 169:
```suggestion
    downloadCalendar(createCalendar({
      calendarName: `${PLATFORM_BRAND_NAME}_Full`,
      events: relatedVenueEvents,
    }));
```

0xdevalias (changes requested):
Tiniest bits of polish then this is good to go!

(Note: from the 'files' tab on this PR, you can use the 'add suggestion to batch' button on each suggestion, and then apply them all as a single commit from the web UI; which is usually quicker than copy/pasting the changes.)

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/components/organisms/NavBarSchedule/utils.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/components/organisms/NavBarSchedule/utils.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/components/organisms/NavBarSchedule/utils.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/utils/calendar.ts
0xdevalias: We'll still need to replace this with the actual correct type

src/utils/calendar.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4

src/utils/calendar.ts
0xdevalias: We'll still need to replace this with the actual correct type

src/utils/calendar.ts
0xdevalias: It doesn't look like this was applied in https://github.com/sparkletown/sparkle/pull/1423/commits/1346ad92bc3006ba3bbbe1f5695c2806dcc4f4c4 ?

src/utils/calendar.ts
0xdevalias: This was closed, but with no reason/context given one way or the other as to why?

htwangtw:
If there's more code style related comments, please feel free to just make the change directly, if PR editing permission is set for you.

src/utils/calendar.ts
htwangtw: `calendar.name` doesn't work as intended. fall back to using `calendarName`

src/utils/calendar.ts
htwangtw: Not adapting this solution see ec68e83

src/utils/calendar.ts
htwangtw: See L47 in src/utils/calendar.ts d339d9a

yarikoptic:
@htwangtw conflicts came up, could you please merge/resolve?

mike-lvov (approved):
Added polish bits in https://github.com/sparkletown/sparkle/pull/1423/commits/a0f03a4c887f384f5ce79e2b2003af867fa063b6

Ready to be merged LGTM ! 🎉

mike-lvov:
@0xdevalias turned off the `code-owner` review as you have seen the addition to the `package.json`

mike-lvov:
Thanks @htwangtw.
You did all the heavy lifting, code works amazing, I tested it locally myself. It's a pleasure to work with such outstanding opensource contributors as you are!

0xdevalias:
> @0xdevalias turned off the `code-owner` review as you have seen the addition to the `package.json`

@mike-lvov Sounds good 👌🏻

PR #1427, opened 2021-05-20
added room type for embedding iframes on partymap

margulies:
The idea here is to allow folks to embed an iframe on the partymap. This request has come up a couple times for OHBM, so thought to integrate the feature. Happy for any suggestions.


src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -50,6 +50,22 @@ $label-visible: flex;
...
+      &:hover {
+        z-index: z(map-room--iframe);
+
+        transform: none;
+
+        .maproom__image {
+          filter: none;
+        }
+      }
+    }
```
0xdevalias, line 68: Since these are all the same styles as `&--unclickable`, I think we should be able to combine their definitions by changing line 39 to something like this:

```scss
 &--unclickable, &--iframe {
```

src/scss/constants.scss
```diff
@@ -148,6 +148,7 @@ $z-layers: (
   map-room: 4,
   map-grid: 3,
   map-room--unclickable: 2,
+  map-room--iframe: 4,
   map-background: 1,
```
0xdevalias, line 152: We aim to keep the order here (within each 'section') based on the `z-index` 'level'. So if we leave `map-room--iframe` as `4` then we should move it up next to `map-room`.

Though I wonder, should we have it set to `4`? Or should we have it set to `2`, the same as 'unclickable'? Or something entirely new (potentially bumping other layers 'up a number' to make room for it. When I originally envisaged this layering for the map, my intention was that we would have each item on it's own layer, so that we can explicitly control the positioning (rather than 'natural layering' occuring when multiple items of different 'types' on the same layer interact)

@margulies @sofisparkle Do you have any thoughts on what the layering here should be?

src/types/rooms.ts
```diff
@@ -2,6 +2,7 @@ import { SoundConfigReference } from "./sounds";
 
 export enum RoomTypes {
   unclickable = "UNCLICKABLE",
+  iframe = "IFRAME",
 }
```
0xdevalias, line 6: Between this PR and https://github.com/sparkletown/sparkle/pull/1370, I feel like the naming of the `RoomTypes` might get a little confusing.

In particular, this `iframe` relates to changing how the room is displayed on the map itself (`MapRoom`), whereas in https://github.com/sparkletown/sparkle/pull/1370 we are using `video` to change how the `RoomModal` popup is shown.

I wonder if we can build a pattern into our naming choices here that distinguish between 'different thing on the map (`MapRoom`)' versus 'different thing in the modal (`RoomModal`)' in a clean/simple way?

Maybe it's as simple as using `videoModal` or similar for https://github.com/sparkletown/sparkle/pull/1370's type? What do you think?

(cc // @mike-lvov @sofisparkle in case you have any thoughts/ideas)

src/components/templates/PartyMap/components/Map/MapRoom.scss
0xdevalias: Oh.. actually.. except for the `z-index`s..

See my questions/comments/thoughts in `src/scss/constants.scss` about the layering in general. Depending on what we decide there will influence what changes (if any) to make here.

Even if we do need to keep the z-index layers seperate, we could (if we wanted) use this combined style type for the main 'shared styles', and then add the individual entries below that to just set the different `z-index`'s.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -44,15 +45,19 @@ export const MapRoom: React.FC<MapRoomProps> = ({
 
   const containerClasses = classNames("maproom", {
     "maproom--unclickable": isUnclickable,
+    "maproom--iframe": isIframe,
     "maproom--always-show-label":
       !isUnclickable &&
+      !isIframe &&
```
0xdevalias, line 51: As per other comments in this review, refactor to use `NON_INTERACTABLE_ROOM_TYPES` (or similar) extracted into `src/settings`:

```suggestion
      !NON_INTERACTABLE_ROOM_TYPES.includes(room.type) &&
```

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -44,15 +45,19 @@ export const MapRoom: React.FC<MapRoomProps> = ({
...
       (venue.roomVisibility === RoomVisibility.nameCount ||
         (venue.roomVisibility === RoomVisibility.count && hasRecentRoomUsers)),
   });
 
   const titleClasses = classNames("maproom__title", {
     "maproom__title--count":
-      !isUnclickable && venue.roomVisibility === RoomVisibility.count,
+      !isUnclickable &&
+      !isIframe &&
```
0xdevalias, line 59: Using the `isNonInteractableRoom` pattern mentioned in another comment in this review:

```suggestion
      !isNonInteractableRoom &&
```

0xdevalias: Though.. another question here actually.. do we want to be able to have a 'room title' shown for the `iframe` embedded on the map?

This is probably a @sofisparkle / @sparkletown/product question.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: In fact.. since we're going to use this in a few places in this file, it probably makes sense to make it a standalone `const` up near `isUnclickable`/`isIframe`, then reuse that here/elsewhere:

```typescript
const isNonInteractableRoom = NON_INTERACTABLE_ROOM_TYPES.includes(room.type);
```

Then here we would just use:
```suggestion
      !isNonInteractableRoom &&
```

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -82,13 +87,23 @@ export const MapRoom: React.FC<MapRoomProps> = ({
     <div
       className={containerClasses}
       style={roomInlineStyles}
-      onClick={isUnclickable ? noop : selectRoomWithSound}
-      onMouseEnter={isUnclickable ? noop : handleRoomHovered}
-      onMouseLeave={isUnclickable ? noop : handleRoomUnhovered}
+      onClick={isUnclickable || isIframe ? noop : selectRoomWithSound}
+      onMouseEnter={isUnclickable || isIframe ? noop : handleRoomHovered}
+      onMouseLeave={isUnclickable || isIframe ? noop : handleRoomUnhovered}
```
0xdevalias, line 92: Using the `isNonInteractableRoom` pattern mentioned in another comment in this review:

```suggestion
      onClick={isNonInteractableRoom ? noop : selectRoomWithSound}
      onMouseEnter={isNonInteractableRoom ? noop : handleRoomHovered}
      onMouseLeave={isNonInteractableRoom ? noop : handleRoomUnhovered}
```

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: It might even be simpler/more obvious to just make it:

```typescript
const isRoomInteractable = !NON_INTERACTABLE_ROOM_TYPES.includes(room.type);
```

(to avoid the 'double negative' everywhere it's used)

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -82,13 +87,23 @@ export const MapRoom: React.FC<MapRoomProps> = ({
...
+        <img className="maproom__image" src={room.image_url} alt={room.title} />
+      ) : (
+        <iframe
+          title={room.title}
+          id={`${room.type}-${room.title}`}
+          scrolling="no"
+          allow="autoplay"
+          src={room.url}
+        />
+      )}
```
0xdevalias, line 104: This feels a little messy.. I wonder if there is a better way we can handle this than having the ternary directly in the JSX like this? Though my brain isn't coming up with any great solutions right now.. @mike-lvov any thoughts?

0xdevalias: Why did we add the `id` prop to the `iframe`? We basically never use/need to use `id`'s in React.

---

I expect we're probably going to want to add a `className` to the iframe (probably `&__iframe`) or similar in the `.scss`) to apply some styling directly to this. At a minimum, setting the width/height to `100%` (so it fills it's parent container) seems like a decent idea?

---

It seems like `scrolling="no"` is deprecated. I expect this should also be controlled from CSS styles these days:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#deprecated_attributes

Further reading:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement

---

Instead of `allow="autoplay"`, we may/probably want to use our standard `IFRAME_ALLOW` from `src/settings`?

---

```suggestion
      {!isIframe ? (
        <img className="maproom__image" src={room.image_url} alt={room.title} />
      ) : (
        <iframe
          className="maproom__iframe"
          src={room.url}
          title={room.title}
          allow={IFRAME_ALLOW}
        />
      )}
```

0xdevalias: [OOS?] Eventually this (and all iframes in our codebase) should use a standard/generic 'Iframe wrapper component' that encapsulates our patterns/etc, but I don't think we have one in `staging` yet..

I believe @mike-lvov may have mentioned that he has one in a recent PR he has been working on potentially?

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -82,13 +87,23 @@ export const MapRoom: React.FC<MapRoomProps> = ({
...
 
-      {!isUnclickable && (
+      {!isUnclickable && !isIframe && (
```
0xdevalias, line 106:
```suggestion
      {!isNonInteractableRoom && (
```

0xdevalias (changes requested):
Thanks for the contribution @margulies!

I expect this one is going to make @sofisparkle extremely happy, as it should resolve (the non-admin controls side of) https://github.com/sparkletown/internal-sparkle-issues/issues/472

Have left my thoughts/comments/questions/changes below.

margulies:
Thanks for the feedback, @0xdevalias. It looks as though it might make sense to integrate #1370 into this as well. Any suggestions about whether to merge the two or how to best proceed?

0xdevalias:
> It looks as though it might make sense to integrate #1370 into this as well. Any suggestions about whether to merge the two or how to best proceed?

@margulies Generally speaking I would say no, keep them separate. While they are similar, they are distinct features. Smaller more focussed PR's are easier to review/land/etc than trying to combine 'all the things' into a single PR.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -62,7 +62,7 @@ const NavSearchBar = () => {
       venue?.rooms
         ?.filter(
           (room) =>
-            room.type !== RoomTypes.unclickable &&
+            room.type !== (RoomTypes.unclickable || RoomTypes.iframe) &&
```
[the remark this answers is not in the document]
margulies: Thanks — for the time being I've implemented with `COVERT_ROOM_TYPES`, which I think sounds cool and captures the slightly beneath the surface aspect.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -29,7 +29,7 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
   const hasSelectedRoom = !!selectedRoom;
 
   const selectRoom = useCallback((room: Room) => {
-    if (room.type === RoomTypes.unclickable) return;
+    if (room.type === (RoomTypes.unclickable || RoomTypes.iframe)) return;
```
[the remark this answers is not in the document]
margulies: Done. I had an issue with `room.type` potentially being `undefined`, so implemented as follows:  
`if (room.type && COVERT_ROOM_TYPES.includes(room.type)) return;`  
Same general approach for above point.

src/components/templates/PartyMap/components/Map/MapRoom.scss
margulies: I've renamed them `covert` and separated out `iframe` and `unclickable` for the `z-index` values

src/scss/constants.scss
margulies: I think it would be better to show this above the `map-grid`, but could shift `map-room` and everything higher up by 1

src/components/molecules/NavSearchBar/NavSearchBar.tsx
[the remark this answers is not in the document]
margulies: I've implemented the logic as: `(!room.type || !COVERT_ROOM_TYPES.includes(room.type)) &&` because otherwise I encountered the error that `room.type` may be `undefined`

src/types/rooms.ts
margulies: Not sure how to best address this point, but happy for any guidance

src/components/templates/PartyMap/components/Map/MapRoom.tsx
margulies: For what it's worth, I have a few requests where title wouldn't be necessary

src/components/templates/PartyMap/components/Map/MapRoom.tsx
margulies: done

src/components/templates/PartyMap/components/Map/MapRoom.tsx
margulies: done

src/components/templates/PartyMap/components/Map/MapRoom.tsx
margulies: replaced with `!isCovertRoom`

src/components/templates/PartyMap/PartyMap.tsx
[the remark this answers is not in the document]
0xdevalias: Cool, sounds good.

src/scss/constants.scss
0xdevalias: Yeah, above the `map-grid` definitely makes sense to me as well, since it's more 'content you want to see' rather than 'part of the background' (like `unclickable` was intended as).

Pushing the other `z-index`s 'up by 1' to 'make room' seems like a good choice here.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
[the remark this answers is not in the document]
0xdevalias: Sounds good.

src/types/rooms.ts
0xdevalias: I've just started a thread on this in our internal tech channel ([ref](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1621918122122300) 🔒) so will hopefully be able to give you some guidance one way or another soon.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: Makes sense. Originally my thinking was that this can already be controlled by the venue settings, but since this applies to all 'rooms' in the map, it's not really the same concept.

For now let's leave it as not having a name, and we can address that in future if we find we have a need for it to be otherwise.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: Sounds good.

src/types/rooms.ts
margulies: Any suggestions for how to address this one? Otherwise seems we're more or less ready to go with this feature

src/scss/constants.scss
mike-lvov: Changed the order in https://github.com/sparkletown/sparkle/pull/1427/commits/998a0612be928ccb5278cc3cb058ef0399f292e7

src/types/rooms.ts
mike-lvov: > In particular, this iframe relates to changing how the room is displayed on the map itself (MapRoom)

Same with `unclickable`? Seems like it's #1370 that needs to be changed, not `RoomTypes`.

Since, 
> Maybe it's as simple as using videoModal or similar for #1370 type? What do you think?

is related to #1370, resolving this comment

src/components/templates/PartyMap/components/Map/MapRoom.tsx
mike-lvov: > This feels a little messy.. I wonder if there is a better way we can handle this than having the ternary directly in the JSX like this? Though my brain isn't coming up with any great solutions right now.. @mike-lvov any thoughts?

Not particularly, but I exchanged them in https://github.com/sparkletown/sparkle/pull/1427/commits/f5c38c702a23cf5d322715fa3c24eae1fb42e052

src/components/templates/PartyMap/components/Map/MapRoom.tsx
mike-lvov: Since the naming has changed, I don't think it's relevant?

All of the other comments seem to be addressed

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -31,6 +33,8 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   const hasRecentRoomUsers = recentRoomUsers.length > 0;
 
   const isUnclickable = room.type === RoomTypes.unclickable;
+  const isIframe = room.type === RoomTypes.iframe;
+  const isCovertRoom = room.type && COVERT_ROOM_TYPES.includes(room.type);
```
mike-lvov, line 37: The only thing that concerns me is the naming here, but it's definitely NB

mike-lvov (approved):
LGTM. A note about the naming choice @0xdevalias wdyt?

src/types/rooms.ts
0xdevalias: @mike-lvov cant remember where but there was discussion around this on slack, I know @sofisparkle chimed in, and if I remember correctly we may have even come to a conclusion on what to call each of these.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: As I mentioned in the other comment thread (can’t get link as on phone atm), I believe there was a slack discussion + decisions made around naming somewhere.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: I think my original pondering was around potentially using `switch`/`case` or similar, but seems fine enough as is for now given there are only 2 choices.

---

@mike-lvov Did you see the rest of my comments in this thread around deprecated features/etc and how they should be done through styling? It looks like you may have applied my suggested changes, or similar, but there aren't any styling changes made in the commit you linked?

src/types/rooms.ts
0xdevalias: In fact, it's the thread linked from https://github.com/sparkletown/sparkle/pull/1427#discussion_r638451622

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: In fact, it's the thread linked from https://github.com/sparkletown/sparkle/pull/1427#discussion_r638451622

src/types/rooms.ts
mike-lvov: Is renamed in https://github.com/sparkletown/sparkle/pull/1427/commits/2ce258fde4bbcf2151b14086f53390bb874f2910

src/components/templates/PartyMap/components/Map/MapRoom.tsx
mike-lvov: > @mike-lvov Did you see the rest of my comments in this thread around deprecated features/etc and how they should be done through styling? It looks like you may have applied my suggested changes, or similar, but there aren't any styling changes made in the commit you linked?

They were fixed before me. The latest version looks like this:
image

mike-lvov:
After testing the implementation, I found out that it lacks two things:
- correct sizing for the iframe

Fixed in https://github.com/sparkletown/sparkle/pull/1427/commits/b9101ff867b5c632e45f72fb622e3a31c06092a7
NOTE: The changes regarding the room.url were reverted in https://github.com/sparkletown/sparkle/pull/1427/commits/a4dac1fb0f26bf0b1fce29310f42629bd5fc47c8

PR #1432, opened 2021-05-21
[env/ohbm] Display subvenue cards in PosterHall template based on live events

margulies:
Adds feature to show venue cards at the top of the posterhall template for specified live events (as determined by events in subvenues).  This is a specific feature needed for sessions called 'poster highlights' at OHBM, which only last for the first half hour of poster sessions. If this is too specific, happy to move the PR to env/ohbm. 

Also, currently having difficulty getting the event to update live (though refreshing does the trick). Any react tips to address that would be appreciated. 

Thanks!

### DIFF
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -69,6 +69,16 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     venueIds: subVenueIds,
   });
 
+  useMemo(
+    () =>
+      setLiveVenueIds(
+        subVenueEvents
+          .filter((event) => isEventLive(event))
+          .map((event) => event.venueId)
+      ),
+    [subVenueEvents]
+  );
+
   useInterval(() => {
     setLiveVenueIds(
       subVenueEvents
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -1,23 +1,16 @@
-import React, { useMemo, useState } from "react";
+import React, { useMemo } from "react";
 
-import { GenericVenue, VenueTemplate } from "types/venues";
-
-import { PosterPageVenue } from "types/venues";
+import { GenericVenue } from "types/venues";
 
 import { WithId } from "utils/id";
-import { isEventLive } from "utils/event";
 
-import { usePosters } from "hooks/posters";
-import { useVenueEvents } from "hooks/events";
-import { useRelatedVenues } from "hooks/useRelatedVenues";
-import { useInterval } from "hooks/useInterval";
+import { usePosters, usePosterHallSubVenues } from "hooks/posters";
 
 import { Button } from "components/atoms/Button";
 
 import { PosterPreview } from "./components/PosterPreview";
+import { SubVenuePreview } from "./components/SubVenuePreview";
 import { PosterHallSearch } from "./components/PosterHallSearch";
-
-import { POSTERHALL_SUBVENUE_STATUS_MS } from "settings";
 
 import "./PosterHall.scss";
 
@@ -47,57 +40,18 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     ));
   }, [posterVenues]);
 
-  const { relatedVenues } = useRelatedVenues({
-    currentVenueId: venue.id,
-  });
-
-  const subVenues = useMemo(() => {
-    return relatedVenues.filter(
-      (relatedVenue) =>
-        relatedVenue.parentId === venue.id &&
-        relatedVenue.template !== VenueTemplate.posterpage
-    );
-  }, [relatedVenues, venue]);
-
-  const subVenueIds = useMemo(() => subVenues.map((venue) => venue.id), [
-    subVenues,
-  ]);
-
-  const { events: subVenueEvents } = useVenueEvents({
-    venueIds: subVenueIds,
-  });
-
-  const initLiveVenueIds = useMemo(
-    () =>
-      subVenueEvents
-        .filter((event) => isEventLive(event))
-        .map((event) => event.venueId),
-    [subVenueEvents]
-  );
-
-  const [liveVenueIds, setLiveVenueIds] = useState(initLiveVenueIds);
-
-  useInterval(() => {
-    const updateLiveVenueIds = subVenueEvents
-      .filter((event) => isEventLive(event))
-      .map((event) => event.venueId);
-
-    if (updateLiveVenueIds === liveVenueIds) return;
-    setLiveVenueIds(updateLiveVenueIds);
-  }, POSTERHALL_SUBVENUE_STATUS_MS);
+  const { liveVenueEvents } = usePosterHallSubVenues(venue.id);
 
   const renderedSubvenues = useMemo(() => {
-    if (liveVenueIds)
-      return subVenues
-        .filter((subVenue) => liveVenueIds.includes(subVenue.id))
-        .map((subVenue) => (
-          <PosterPreview
-            key={subVenue.id}
-            posterVenue={subVenue as WithId<PosterPageVenue>}
-            enterVenue={enterVenue}
-          />
-        ));
-  }, [liveVenueIds, subVenues]);
+    return liveVenueEvents.map((subVenueEvent) => (
+      <SubVenuePreview
+        key={subVenueEvent.venueId}
+        venueId={subVenueEvent.venueId}
+        title={subVenueEvent.name}
+        host={subVenueEvent.host}
+      />
+    ));
+  }, [liveVenueEvents]);
 
   return (
     <div className="PosterHall">
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -49,7 +49,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
 
   const [liveVenueIds, setLiveVenueIds] = useState<string[]>();
 
-  const { isRelatedVenuesLoading, relatedVenues } = useRelatedVenues({
+  const { relatedVenues } = useRelatedVenues({
     currentVenueId: venue.id,
   });
 
@@ -61,8 +61,12 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     );
   }, [relatedVenues, venue]);
 
+  const subVenueIds = useMemo(() => subVenues.map((venue) => venue.id), [
+    subVenues,
+  ]);
+
   const { events: subVenueEvents } = useVenueEvents({
-    venueIds: subVenues.map((venue) => venue.id),
+    venueIds: subVenueIds,
   });
 
   useInterval(() => {
@@ -73,27 +77,22 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     );
   }, LOCATION_INCREMENT_MS);
 
-  const liveSubVenues = useMemo(() => {
-    if (!liveVenueIds) return;
-    return subVenues.filter((subVenue) => liveVenueIds.includes(subVenue.id));
-  }, [subVenues, liveVenueIds]);
-
   const renderedSubvenues = useMemo(() => {
-    if (liveSubVenues)
-      return liveSubVenues.map((subVenue) => (
-        <PosterPreview
-          key={subVenue.id}
-          posterVenue={subVenue as WithId<PosterPageVenue>}
-          enterVenue={enterVenue}
-        />
-      ));
-  }, [liveSubVenues]);
+    if (liveVenueIds)
+      return subVenues
+        .filter((subVenue) => liveVenueIds.includes(subVenue.id))
+        .map((subVenue) => (
+          <PosterPreview
+            key={subVenue.id}
+            posterVenue={subVenue as WithId<PosterPageVenue>}
+            enterVenue={enterVenue}
+          />
+        ));
+  }, [liveVenueIds, subVenues]);
 
   return (
     <div className="PosterHall">
-      <div className="PosterHall__related">
-        {isRelatedVenuesLoading ? renderedSubvenues : "Loading"}
-      </div>
+      <div className="PosterHall__related">{renderedSubvenues}</div>
 
       <PosterHallSearch
         setSearchInputValue={setSearchInputValue}
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -3,8 +3,12 @@ import React, { useMemo } from "react";
 import { GenericVenue } from "types/venues";
 
 import { WithId } from "utils/id";
+import { isEventLive } from "utils/event";
 
 import { usePosters } from "hooks/posters";
+import { useVenueId } from "hooks/useVenueId";
+import { useVenueEvents } from "hooks/events";
+import { useRelatedVenues } from "hooks/useRelatedVenues";
 
 import { Button } from "components/atoms/Button";
 
@@ -18,6 +22,14 @@ export interface PosterHallProps {
 }
 
 export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
+  const venueId = useVenueId();
+  const { isRelatedVenuesLoading, relatedVenueIds } = useRelatedVenues({
+    currentVenueId: venueId,
+  });
+  const { events: relatedVenueEvents } = useVenueEvents({
+    venueIds: relatedVenueIds,
+  });
+
   const {
     posterVenues,
     isPostersLoaded,
@@ -42,20 +54,30 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     ));
   }, [posterVenues]);
 
+  const liveRelatedVenues = useMemo(() => {
+    const liveVenueIds = relatedVenueEvents
+      .filter((event) => isEventLive(event))
+      .map((event) => event.venueId);
+    return posterRelatedVenues.filter((posterRelatedVenue) =>
+      liveVenueIds.includes(posterRelatedVenue.id)
+    );
+  }, [relatedVenueEvents, posterRelatedVenues]);
+
   const renderedPosterRelatedPreviews = useMemo(() => {
-    return posterRelatedVenues.map((posterRelatedVenue) => (
+    return liveRelatedVenues.map((posterRelatedVenue) => (
       <PosterPreview
         key={posterRelatedVenue.id}
         posterVenue={posterRelatedVenue}
-        enterVenue={enterVenue}
       />
     ));
-  }, [posterRelatedVenues]);
+  }, [liveRelatedVenues]);
 
   return (
     <div className="PosterHall">
       <div className="PosterHall__related">
-        {isPosterRelatedLoaded ? renderedPosterRelatedPreviews : "Loading..."}
+        {isPosterRelatedLoaded || isRelatedVenuesLoading
+          ? renderedPosterRelatedPreviews
+          : "Loading..."}
       </div>
       <PosterHallSearch
         setSearchInputValue={setSearchInputValue}
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -17,7 +17,7 @@ import { Button } from "components/atoms/Button";
 import { PosterPreview } from "./components/PosterPreview";
 import { PosterHallSearch } from "./components/PosterHallSearch";
 
-import { LOCATION_INCREMENT_MS } from "settings";
+import { POSTERHALL_SUBVENUE_STATUS_MS } from "settings";
 
 import "./PosterHall.scss";
 
@@ -75,7 +75,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
         .filter((event) => isEventLive(event))
         .map((event) => event.venueId)
     );
-  }, LOCATION_INCREMENT_MS);
+  }, POSTERHALL_SUBVENUE_STATUS_MS);
 
   const renderedSubvenues = useMemo(() => {
     if (liveVenueIds)
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -23,6 +23,9 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     isPostersLoaded,
     hasHiddenPosters,
 
+    posterRelatedVenues,
+    isPosterRelatedLoaded,
+
     increaseDisplayedPosterCount,
 
     searchInputValue,
@@ -39,8 +42,21 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     ));
   }, [posterVenues]);
 
+  const renderedPosterRelatedPreviews = useMemo(() => {
+    return posterRelatedVenues.map((posterRelatedVenue) => (
+      <PosterPreview
+        key={posterRelatedVenue.id}
+        posterVenue={posterRelatedVenue}
+        enterVenue={enterVenue}
+      />
+    ));
+  }, [posterRelatedVenues]);
+
   return (
     <div className="PosterHall">
+      <div className="PosterHall__related">
+        {isPosterRelatedLoaded ? renderedPosterRelatedPreviews : "Loading..."}
+      </div>
       <PosterHallSearch
         setSearchInputValue={setSearchInputValue}
         searchInputValue={searchInputValue}
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -47,8 +47,6 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     ));
   }, [posterVenues]);
 
-  const [liveVenueIds, setLiveVenueIds] = useState<string[]>();
-
   const { relatedVenues } = useRelatedVenues({
     currentVenueId: venue.id,
   });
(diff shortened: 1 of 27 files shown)


src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -41,8 +60,36 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+      const liveVenueIds = relatedVenueEvents
+        .filter((event) => isEventLive(event))
+        .map((event) => event.venueId);
+      const livePosterRelatedVenues = posterRelatedVenues.filter(
+        (posterRelatedVenue) => liveVenueIds.includes(posterRelatedVenue.id)
+      );
+      setPosterRelatedPreviews(livePosterRelatedVenues);
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [posterRelatedVenues, relatedVenueEvents]);
```
0xdevalias, line 74: > Also, currently having difficulty getting the event to update live (though refreshing does the trick). Any react tips to address that would be appreciated.

@margulies Instead of using a `useEffect` + `setInterval` here, try just using `useInterval`.

The 'interval time' here will need to be extracted to `src/settings` or one of the existing ones defined there should be used.

0xdevalias: @margulies Though taking a step further back here.. what is the actual intent/desire for this code? There may be a simpler/better way to go about it.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -24,6 +24,7 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
   const posterClassnames = classNames("PosterPreview", {
     "PosterPreview--live": posterVenue.isLive,
+    "PosterPreview--related": !posterVenue.template.includes("posterpage"),
```
0xdevalias, line 27: Why `.includes` here? I believe `venue.template` is just a string, not an array?

```suggestion
    "PosterPreview--related": posterVenue.template !== "posterpage",
```

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -19,10 +23,21 @@ export interface PosterHallProps {
 }
 
 export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
+  const venueId = useVenueId();
+  const { isRelatedVenuesLoading, relatedVenueIds } = useRelatedVenues({
+    currentVenueId: venueId,
+  });
+  const { events: relatedVenueEvents } = useVenueEvents({
+    venueIds: relatedVenueIds,
+  });
```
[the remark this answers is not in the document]
0xdevalias: Looking at the query config for `useConnectPosterRelatedVenues` in `src/hooks/posters.ts`:

```typescript
{
  collection: "venues",
  where: [
    ["parentId", "==", posterHallId],
    ["template", "in", [VenueTemplate.audience, VenueTemplate.jazzbar]],
  ],
  storeAs: "posterRelatedVenues",
}
```

It's getting the `audience`/`jazzbar` venues that are direct children of the `posterHallId`, which is passed in as props from `usePosters`, and so in this file is set to `venue.id`

We're already passing `venueId` into `useRelatedVenues` here, so we can probably just `.filter` the `relatedVenues` array in a `useMemo` or similar to find venues where:

```typescript
const fooRelatedVenues = relatedVenues.filter((relatedVenue) => relatedVenue.parentId === venue.id && [VenueTemplate.audience, VenueTemplate.jazzbar].includes(relatedVenue.template))
```

This wouldn't live update, but I expect that isn't needed here as the venues are unlikely to be changing.

We would then get the `id`s from just these `fooRelatedVenues` (needs better name; this would likely be done with `.map`, and would also need to be done in a `useMemo`), and pass that to `useVenueEvents`, which would then only fetch the events for these venues, rather than fetching all of them; which you could then filter to get just the live ones in the `useInterval`, and then set that result to the `useState`, etc.

This would then allow us to get rid of all of the `useFirestoreConnect`/selector/firestore types/etc that were added in this PR and simplify things right down here.

src/hooks/posters.ts
```diff
@@ -4,7 +4,10 @@ import Fuse from "fuse.js";
 
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
-import { posterVenuesSelector } from "utils/selectors";
+import {
+  posterVenuesSelector,
+  posterRelatedVenuesSelector,
+} from "utils/selectors";
```
0xdevalias, line 10:
```suggestion
import { posterVenuesSelector } from "utils/selectors";
```

src/hooks/posters.ts
```diff
@@ -25,6 +28,21 @@ export const useConnectPosterVenues = (posterHallId: string) => {
...
+        where: [
+          ["parentId", "==", posterHallId],
+          ["template", "in", [VenueTemplate.audience, VenueTemplate.jazzbar]],
+        ],
+        storeAs: "posterRelatedVenues",
+      },
+    ];
+  });
+};
+
```
0xdevalias, line 45:
```suggestion
```

src/hooks/posters.ts
```diff
@@ -39,8 +57,25 @@ export const usePosterVenues = (posterHallId: string) => {
...
+  return useMemo(
+    () => ({
+      posterRelatedVenues: posterRelatedVenues ?? [],
+      isPosterRelatedLoaded: isLoaded(posterRelatedVenues),
+    }),
+    [posterRelatedVenues]
+  );
+};
+
```
0xdevalias, line 73:
```suggestion
```

src/hooks/posters.ts
```diff
@@ -39,8 +57,25 @@ export const usePosterVenues = (posterHallId: string) => {
...
 export const usePosters = (posterHallId: string) => {
   const { posterVenues, isPostersLoaded } = usePosterVenues(posterHallId);
+  const { posterRelatedVenues, isPosterRelatedLoaded } = usePosterRelatedVenues(
+    posterHallId
+  );
```
0xdevalias, line 78:
```suggestion
```

src/hooks/posters.ts
```diff
@@ -90,6 +125,9 @@ export const usePosters = (posterHallId: string) => {
     posterVenues: searchedPosterVenues,
     isPostersLoaded,
 
+    posterRelatedVenues,
+    isPosterRelatedLoaded,
+
```
0xdevalias, line 130:
```suggestion
```

src/types/Firestore.ts
```diff
@@ -100,6 +100,7 @@ export interface FirestoreOrdered {
   userModalVisits?: WithId<UserVisit>[];
   userPurchaseHistory?: WithId<Purchase>[];
   privateChatMessages?: WithId<PrivateChatMessage>[];
+  posterRelatedVenues?: WithId<PosterPageVenue>[];
```
0xdevalias, line 103:
```suggestion
```

src/utils/selectors.ts
```diff
@@ -199,6 +199,10 @@ export const posterVenuesSelector: SparkleSelector<
   WithId<PosterPageVenue>[] | undefined
 > = (state) => state.firestore.ordered.posterVenues;
 
+export const posterRelatedVenuesSelector: SparkleSelector<
+  WithId<PosterPageVenue>[] | undefined
+> = (state) => state.firestore.ordered.posterRelatedVenues;
+
```
0xdevalias, line 205:
```suggestion
```

0xdevalias (changes requested):
Comments/suggestions including patterns to clean up/simplify the implementation a bunch.

src/components/templates/PosterHall/PosterHall.tsx
margulies: Thanks for the feedback — to explain the aim of this PR: We have events related to the posterhall that will only be occuring for the first half of each session. So as not to clutter the main landing page (partymap), and to emphasize the relationship to the posterhall, I would ideally like the venues related to these event to appear at the top of the posterhall for only the half hour in which they are occuring. Happy for any other thoughts related to implementation. Otherwise, I'll continue addressing your comments here. Thanks!

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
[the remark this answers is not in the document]
margulies: implemented

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -41,8 +49,54 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+
+  const subVenues = useMemo(() => {
+    return relatedVenues.filter(
+      (relatedVenue) =>
+        relatedVenue.parentId === venue.id &&
+        relatedVenue.template !== VenueTemplate.posterpage
+    );
+  }, [relatedVenues, venue]);
+
+  const { events: subVenueEvents } = useVenueEvents({
```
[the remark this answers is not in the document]
margulies: refactored a bit to get the `subVenueIds` from line 66, and it seems to work smoothly now

src/components/templates/PosterHall/PosterHall.tsx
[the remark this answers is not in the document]
margulies: Thanks for all the help with this — I managed to implement what you described, and seems to finally be working smoothly again.

src/components/templates/PosterHall/PosterHall.tsx
margulies: I've now implemented with `useInterval` and the additional refactoring you suggested. Thanks!

margulies:
I added back in a `useInterval` check every `1 sec` because the 10 sec wait every time the page loaded for any live `subVenues` to show up seemed too long for the user

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -41,8 +60,36 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+    ));
+  }, [posterRelatedPreviews]);
+
   return (
     <div className="PosterHall">
+      <div className="PosterHall__related">
+        {isPosterRelatedLoaded || isRelatedVenuesLoading
+          ? renderPosterRelatedPreviews
+          : "Loading..."}
+      </div>
```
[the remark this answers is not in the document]
margulies: As the default behavior is that no `subVenues` are shown, I've removed this

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -39,8 +47,52 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+    venueIds: subVenueIds,
+  });
+
+  useInterval(() => {
+    setLiveVenueIds(
+      subVenueEvents
+        .filter((event) => isEventLive(event))
+        .map((event) => event.venueId)
+    );
+  }, POSTERHALL_SUBVENUE_STATUS_MS);
```
mike-lvov, line 78: This will cause fetching of the events every second

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -39,8 +47,52 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
...
+
+  const renderedSubvenues = useMemo(() => {
+    if (liveVenueIds)
+      return subVenues
+        .filter((subVenue) => liveVenueIds.includes(subVenue.id))
+        .map((subVenue) => (
+          <PosterPreview
+            key={subVenue.id}
+            posterVenue={subVenue as WithId<PosterPageVenue>}
```
mike-lvov, line 87: Since subVenue is not same as PosterPage, it's very error-prone approach

mike-lvov (changes requested):
I'm afraid this PR can cause some performace-related issues. @0xdevalias wdyt?

margulies:
Good question — might be worth mentioning that the events are only checked
for the filtered set of ’subVenues’ (in the current case: 5), so it’s a
very limited fetch. If the frequency is the key factor here, we could
certainly decrease it.


Regarding the `subVenue as WithId<PosterPageVenue>` line, I’d be happy for
better ways to implement this. I struggled for a while to make this work,
and I know this solution isn’t ideal. I could, for instance, fill the
poster props for the venue using existing AnyVenue info such as venue.name,
etc… Any suggestion would be helpful here.

src/components/templates/PosterHall/PosterHall.tsx
0xdevalias: `subVenueEvents` is loaded from `useVenueEvents`, which unless it has changed since I last looked, will only load the events when the `venueIds` passed to it changes. In this case, that comes from the `useMemo`'d `subVenueIds`, which comes from the `useMemo`'d `subVenues`, which filters `relatedVenues`, which comes from `useRelatedVenues` which will only break memo/change when `sovereignVenueId` changes (so basically never).

So every second (or whatever `POSTERHALL_SUBVENUE_STATUS_MS` equates to), this `useInterval` will run and call `setLiveVenueIds`, which I believe (not 100% sure) will cause a re-render, even if the values end up being the same. Since `liveVenueIds` will have changed, it will break the memo on `renderedSubvenues`.

If you use the 'functional' form of setState (eg. `setLiveVenueIds(prevState => newState`), and then check if the values actually changed, you can return the `prevState` when they haven't, which will skip re-renders as per:

- https://reactjs.org/docs/hooks-reference.html#usestate
  - > If the new state is computed using the previous state, you can pass a function to `setState`. The function will receive the previous value, and return an updated value.
  - > **If your update function returns the exact same value as the current state, the subsequent rerender will be skipped completely.**

0xdevalias: In short, I don't think the situation is as dire as @mike-lvov originally suspected it to be, but there are definitely a number of optimisations that could be made, as spelled out above.

I haven't looked at the changes @margulies made in his most recent changed mentioned in https://github.com/sparkletown/sparkle/pull/1432/files/189959cad6f72e004d8186d16b63876952d17c8c#r649509502 since the hash wasn't linked for it. Generally speaking reducing the frequency of the `useInterval` to 1min will reduce the impact of any re-renders/etc.

src/components/templates/PosterHall/PosterHall.tsx
0xdevalias: I haven't looked too deeply into this, but I assume the issue is because `subVenues` (which comes from `relatedVenues`) isn't guaranteed to be a `PosterPage`.

Yet we could guarantee that by adding a check into the `.filter` above that makes the `subVenues`, to ensure it only includes `template === VenueTemplate.posterpage`

Then you should be able to just change the type on `subVenues` without using `as` to the appropriate type, and have it 'just work' down here.

margulies: The only issue with the filtering approach is that the actual venue types for the current subVenues are 'audience', to avoid the space limits of the posterpage template. I have 5 of them set up for this, and have added in the `poster` fields to ensure they'll display properly as posterpreview cards in the posterhall, but if `subVenues` are filtered for template `posterpage` these will be filtered out.

src/components/templates/PosterHall/PosterHall.tsx
margulies: I've implemented a change in this commit: c14d9678647777019fa4f36de1482747a532ab75
that only calls `setLiveVenueIds` inside `useInterval` if there is a change in the `liveVenueIds` list

If there's another way you'd prefer that I implement that, just let me know

src/components/templates/PosterHall/PosterHall.tsx
margulies: I've addressed this issue by adding a component `SubVenuePreview` 783cd299733235039740168864e5ebb3ed5a6bc0
which renders any venues that are not `=== VenueTemplate.posterpage` and that contain live events

src/components/templates/PosterHall/PosterHall.tsx
margulies: This has now been moved to `hook/poster.ts`, implemented in a similar way to the calculation of `currentTimePosition` in `components/molecules/Schedule/Schedule.tsx`. Commit: 2f5bb2560106081d10ef3eb1f733c32207b59406

0xdevalias: Looking at your original commit there, the separate useMemo for the unit code init  needed. If you make the initial value of the useState a function, it will only be run the first time it’s needed to set the initial state.

Then the === check to see if it should update won’t work because it will be checking that the 2 arrays have the literal same reference, not that they contain the same items.

0xdevalias: In the 2nd commit.

checkLiveEvents should be named better for what it’s doing, maybe filterLiveEvents or similar.

The variable subVenues is somewhat misleading, as it’s really representing subVenuesWithoutPosterPages.

setLiveEvents would probably be better named updateLiveEvents. Also, within it you are needlessly running the filter multiple times. Run it once and store it in a const, then use that in your checks. You still have the same === bug as mentioned above that will check that they have the same literal array reference, not that the items in the array are the same. I’ll often use JavaScript’s Set type when I need to efficiently perform this sort of check, but unsure of the overhead of converting both arrays to sets every time just for the comparison here, probably an easier way to check they have all values. Probably a lodash function that does it. You also need a newline after your guard ‘if’.

Whats the point of the useEffect that calls setLiveEvents? You already run the equivalent code in the initialiser of the useState.

Then once that is removed, since you only use setLiveEvents in the useInterval, there doesn’t seem to be a need for it to be extracted as a useCallback anyway, so can just be inclined into the useInterval

src/components/templates/PosterHall/PosterHall.tsx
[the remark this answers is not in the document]
0xdevalias: Glad to hear it :)

0xdevalias:
> I added back in a `useInterval` check every `1 sec` because the 10 sec wait every time the page loaded for any live `subVenues` to show up seemed too long for the user

@margulies Given your latest code seems to run this check as part of the initialiser, there wouldn’t need to be a 10sec wait for the first update, so I expect it would be fine to set this back to a longer rips for subsequent updates.

src/components/templates/PosterHall/PosterHall.tsx
0xdevalias: @mike-lvov FYI. The use of audience templates/etc here sounds.. odd to say the least..

margulies:
> > I added back in a `useInterval` check every `1 sec` because the 10 sec wait every time the page loaded for any live `subVenues` to show up seemed too long for the user
> 
> @margulies Given your latest code seems to run this check as part of the initialiser, there wouldn’t need to be a 10sec wait for the first update, so I expect it would be fine to set this back to a longer rips for subsequent updates.

It's now set to a check interval of one minute.

src/components/templates/PosterHall/PosterHall.tsx
margulies: We have 4-5 parallel, live 'Poster Highlights' sessions that run during the first half hour of each of the poster sessions. These are 30-min moderator-guided presentations of 10-12 posters with summary videos. The aim of these sessions to to show folks some of the exciting posters to go and visit. These are not posters, but are rather talks related to the poster session. I anticipate we may have upwards of 100 attendees at some of these sessions. 

The initial idea was to put them in the symposium halls (audience venues), but we were told by the AV company that a 15 min break would be necessary between sessions for us to use those. 

The solution I came up with was to have venue cards appear at the top of the poster hall in a unique card style when these Poster Highlights events are live.

In brief I know this is a very OHBM-specific feature, but it fits quite well with what we're trying to achieve with these sessions. If there's a better way to implement, I'm open to that.

src/components/templates/PosterHall/PosterHall.tsx
margulies: Thanks for explaining these points. Several of them are addressed in the current version. I've also removed the `if` line from `useEffect` now. 

I'll rename `checkLiveEvents` and `subVenue` now as well.

margulies: I've renamed `subVenue` -> `nonPosterSubVenue` (including variants such as `liveNonPosterSubVenueEvents`)

0xdevalias: > I've also removed the if line from useEffect now.

@margulies Why did you remove it?

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -43,6 +43,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
   const { liveVenueEvents } = usePosterHallSubVenues(venue.id);
 
   const renderedSubvenues = useMemo(() => {
+    if (!liveVenueEvents) return;
     return liveVenueEvents.map((subVenueEvent) => (
```
0xdevalias: Remember to always include a blank line of space after a 'guard `if`' like this.

margulies: Thanks! Will do in the future — this `if` statement has been removed in the current version

margulies: Thanks! implemented in 3ac268ed04961eb5ee453b9a5cc33cf3893a0da3

src/components/templates/PosterHall/PosterHall.tsx
margulies: Sorry, that was actually a type -- I'd removed the `if` statement from `useInterval`. As you'd mentioned initially, I assumed that `setLiveNonPosterSubVenueEvents(() => filterLiveEvents(nonPosterSubVenueEvents));` would only update if there was a change. Let me know if I should bring back the `if` statement though to prevent this from running if there's no change.

0xdevalias: > Let me know if I should bring back the if statement though to prevent this from running if there's no change.

Definitely will need to do this.

---

From https://github.com/sparkletown/sparkle/pull/1432#discussion_r649631983:

Particularly note the bolded part:

> If you use the 'functional' form of setState (eg. setLiveVenueIds(prevState => newState), **and then check if the values actually changed, you can return the prevState when they haven't, which will skip re-renders**

From https://github.com/sparkletown/sparkle/pull/1432#discussion_r650445196:

> You still have the same === bug as mentioned above that will check that they have the same literal array reference, not that the items in the array are the same. I’ll often use JavaScript’s Set type when I need to efficiently perform this sort of check, but unsure of the overhead of converting both arrays to sets every time just for the comparison here, probably an easier way to check they have all values. Probably a lodash function that does it. You also need a newline after your guard ‘if’.

margulies: Thanks! I've now implemented the following `if` guard in 4c6a772dfac90f498a767199c71964886ab7740e
I'm unclear if this properly addresses the first point you raised regarding the functional form of `setState`, so if that needs to change as well, please let me know.

```
useInterval(() => {
    if (
      isEqual(
        sortBy(filterLiveEvents(nonPosterSubVenueEvents)),
        sortBy(liveNonPosterSubVenueEvents)
      )
    )
      return;

    setLiveNonPosterSubVenueEvents(() =>
      filterLiveEvents(nonPosterSubVenueEvents)
    );
  }, POSTERHALL_SUBVENUE_STATUS_MS);
```

0xdevalias: This has the same problem of running the filter code multiple times that I said to extract it as a const and reuse that.

Do you need to sort them? Is there no helper that will directly check that one array contains all the elements of the other and vice versa?

Having the guard where it is is going to close over the value of liveNonPosterVenues at the time the useInterval is created, so it’s not going to be updated. That’s why you need to put the guard within the functional form of the  set* function. When they are the same, return the prevState. When they aren’t, return the new state as I originally described.

margulies: Sorry for being a bit dense on this one. I've refactored as follows in 1a8356b3d62b2e0d19b285bb64737b00ecfae908 

Here's my understanding:
```
  // const that will be called to pull an object containing only live events:

  const filteredLiveEvents = filterLiveEvents(nonPosterSubVenueEvents);

  // set up state type -- I may have already misunderstood -- is this where 
  // the functional form below should already be created?:

  const [
    liveNonPosterSubVenueEvents,
    setLiveNonPosterSubVenueEvents,
  ] = useState<WithVenueId<VenueEvent>[]>();

  // run set... here to ensure it runs after the consts have all loaded 
  // and doesn't wait for the first useInterval timer

  useEffect(
    () =>
      setLiveNonPosterSubVenueEvents(filteredLiveEvents),
    [filteredLiveEvents]
  );

  // check every 1 min if there's a change from prevState using lodash `isEqual`. 
  // If they're equal, return preState (so no reloading). If they aren't equal, return the 
  // new state, initiating a reloading.

  useInterval(() => {
    setLiveNonPosterSubVenueEvents(function(prevState){
      if (isEqual(prevState,filteredLiveEvents)) return prevState;

      return filteredLiveEvents;
    });
  }, POSTERHALL_SUBVENUE_STATUS_MS);
```

margulies: Or is this the correct syntax?: 
```
useInterval(() => {
    setLiveNonPosterSubVenueEvents(prevState => {
      if (isEqual(prevState, filteredLiveEvents)) return prevState;

      return filteredLiveEvents;
    });
  }, POSTERHALL_SUBVENUE_STATUS_MS);
```

margulies: One last attempt, and then I promise to be patient. Is this any closer to what we're going for: 
```
const [
    liveNonPosterSubVenueEvents,
    setLiveNonPosterSubVenueEvents,
  ] = useState<WithVenueId<VenueEvent>[]>();

  const updateLiveEvents = useCallback(() => {
    const filteredLiveEvents = filterLiveEvents(nonPosterSubVenueEvents);

    setLiveNonPosterSubVenueEvents((prevState) => {
      if (isEqual(prevState, filteredLiveEvents)) return prevState;

      return filteredLiveEvents;
    })
  },[nonPosterSubVenueEvents]);

  useEffect(() => updateLiveEvents());

  useInterval(() => {
    updateLiveEvents();
  }, POSTERHALL_SUBVENUE_STATUS_MS);
```

margulies: I've checked the refresh rate on a version similar to the one above, and it seems to be working now: fd09688df03497142bbd53fd069876cc53995f3e Of course, let me know if any changes should be made, but hope this is on the right track

mike-lvov (commented):
Looks good. Please, change the base name of this PR to env/ohbm, rebase out all of the staging merge commits and LGTM to merge in

mike-lvov (approved):
LGTM

PR #1435, opened 2021-05-24
remove unused legacy dependencies: three, three-obj-mtl-loader

0xdevalias:
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/617

replaces https://github.com/sparkletown/sparkle/pull/1226

---

> - install `three` `0.127.0` (major=`false`, depth=`null`)
>   - I'm not sure that we are even using this for anything anymore, might be able to remove it?
>   - `npm ls three` shows it as a top level dependency in our `package.json`
>   - `npx depcheck` lists it as under "unused dependencies"
>   - Searching for imports in the code doesn't seem to show anything
>   - Looking at the 'blame' on GitHub, looks like it was added a while ago, seemingly related to one of our legacy templates (Playa)
>     - https://github.com/sparkletown/sparkle/blame/staging/package.json#L72-L73
>   - We can potentially remove both `three` and `three-obj-mtl-loader` 
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/617#issuecomment-814592460_


mike-lvov (approved):
LGTM

PR #1441, opened 2021-05-25
Adds user import from the file

mike-lvov:
Had to modify our default script. Maybe we could reuse/add this code @0xdevalias ?

This modification gets password + username + avatar from the file & sets it


mike-lvov:
Unrelevant

PR #1445, opened 2021-05-26
increase number of poster previews shown from 12 to 48

yarikoptic:
I have tried with 1200 and it still works ok.  The next UX fiasco which is not
fixed in this PR anyhow is "Show more posters" for a search query which has no
more to show.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -581,7 +581,7 @@ export const POSTERPAGE_MAX_VIDEO_PARTICIPANTS = 10;
 
 export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
-export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 12;
+export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 48;
 
 export const USER_STATUSES = [UserStatus.available, UserStatus.busy];
 


src/settings.ts
```diff
@@ -581,7 +581,7 @@ export const POSTERPAGE_MAX_VIDEO_PARTICIPANTS = 10;
 
 export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
-export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 12;
+export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 48;
```
0xdevalias, line 584: @mike-lvov This PR is tiny so will be a quick review. Since you made the `PosterHall` you're probably best at assessing whether we should make this change or not.

mike-lvov (approved):
LGTM

0xdevalias:
> The next UX fiasco which is not
> fixed in this PR anyhow is "Show more posters" for a search query which has no
> more to show.

@yarikoptic Could you please make a backlog issue about that. Sounds like a bug.

0xdevalias:
@mike-lvov Thanks for reviewing + getting this merged!

@yarikoptic Thanks for the contribution :)

PR #1446, opened 2021-05-26
show poster number alongside the title in PosterPreview / PosterPage

yarikoptic:
Poster number is important information since people refer to their posters by
numbers for easy finding/association/check if in the right place.  But it was
nowhere shown.

I guess ideally it should not be melded into the same `<element>` as title, but
this is the fastest and easiest way since title is in `<p>` and we do not want to
break it all across multiple lines etc.  Specific recommendations or better
tune up commits pushed directly would be most welcome to expedite this
PR.

Search for number yet to be supported (will see if could do in #1443) and is
orthogonal IMHO to having them displayed.

### DIFF
--- a/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
+++ b/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
@@ -19,7 +19,6 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   posterVenue,
 }) => {
   const { title, authorName, categories } = posterVenue.poster ?? {};
-  const number = posterVenue.name.replace("poster", "");
 
   const venueId = posterVenue.id;
 
@@ -44,7 +43,7 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   return (
     <div className={posterClassnames} onClick={handleEnterVenue}>
       <p className="PosterPreview__title">
-        {number}: {title}
+        {posterVenue.name}: {title}
       </p>
 
       <div className="PosterPreview__categories">{renderedCategories}</div>
--- a/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
+++ b/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
@@ -19,6 +19,7 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   posterVenue,
 }) => {
   const { title, authorName, categories } = posterVenue.poster ?? {};
+  const number = posterVenue.name.replace("poster", "");
 
   const venueId = posterVenue.id;
 
@@ -42,7 +43,9 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
   return (
     <div className={posterClassnames} onClick={handleEnterVenue}>
-      <p className="PosterPreview__title">{title}</p>
+      <p className="PosterPreview__title">
+        {number}: {title}
+      </p>
 
       <div className="PosterPreview__categories">{renderedCategories}</div>
 
--- a/src/components/templates/PosterPage/PosterPage.tsx
+++ b/src/components/templates/PosterPage/PosterPage.tsx
@@ -30,6 +30,7 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
   const { id: venueId, isLive: isPosterLive, poster, iframeUrl } = venue;
 
   const { title, introVideoUrl, categories } = poster ?? {};
+  const number = venue.name.replace("poster", "");
 
   const {
     isShown: isIntroVideoShown,
@@ -84,7 +85,9 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
         <div />
 
         <div className="PosterPage__header--middle-cell">
-          <p className="PosterPage__title">{title}</p>
+          <p className="PosterPage__title">
+            {number}: {title}
+          </p>
           <div className="PosterPage__categories">{renderedCategories}</div>
         </div>
 
--- a/src/components/templates/PosterPage/PosterPage.tsx
+++ b/src/components/templates/PosterPage/PosterPage.tsx
@@ -30,7 +30,6 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
   const { id: venueId, isLive: isPosterLive, poster, iframeUrl } = venue;
 
   const { title, introVideoUrl, categories } = poster ?? {};
-  const number = venue.name.replace("poster", "");
 
   const {
     isShown: isIntroVideoShown,
@@ -86,7 +85,7 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
 
         <div className="PosterPage__header--middle-cell">
           <p className="PosterPage__title">
-            {number}: {title}
+            {venue.name}: {title}
           </p>
           <div className="PosterPage__categories">{renderedCategories}</div>
         </div>


src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -19,6 +19,7 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   enterVenue,
 }) => {
   const { title, authorName, categories } = posterVenue.poster ?? {};
+  const number = posterVenue.name.replace("poster", "");
```
0xdevalias, line 22: Probably want to `useMemo` this so we don't have to do the string replace on every render.

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -29,6 +29,7 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
   const { id: venueId, isLive: isPosterLive, poster, iframeUrl } = venue;
 
   const { title, introVideoUrl, categories } = poster ?? {};
+  const number = venue.name.replace("poster", "");
```
0xdevalias, line 32: Probably want to `useMemo` this so we don't have to do the string replace on every render.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: @sparkletown/product @mike-lvov Would be good to get your thoughts/input here for the more general codebase as well.

If we don't want this (or a variation of it) in our main codebase, then we can make this an `env/ohbm` specific 'hack'.

0xdevalias (dismissed):
[CRO] A couple of tiny changes + some questions.

@sparkletown/product @mike-lvov Would be good to get your thoughts/insight on this and whether it's the kind of change we want in our main codebase.

---

> I guess ideally it should not be melded into the same <element> as title, but
this is the fastest and easiest way since title is in <p> and we do not want to
break it all across multiple lines etc. Specific recommendations or better
tune up commits pushed directly would be most welcome to expedite this
PR.

@yarikoptic I'm not too concerned about this being within the same element here. We don't have a need to target the element and extract the text to get the 'raw title', which is probably the only reason this would matter.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
yarikoptic: > @yarikoptic I'm curious, would you also be happy with showing the full `posterVenue.name`, without this `replace`?

if needed - I am ok with such compromise. Indeed since DB-based value might not be generic enough, albeit it is a "poster hall". Can't fathom how this would take any notable amount of time though in comparison to the rest

src/components/templates/PosterPage/PosterPage.tsx
yarikoptic: wouldn't `useMemo` require more cycles (dunno what it does but assume some kind of hashing + caching) than a simple replace in under 10 characters string?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: > Can't fathom how this would take any notable amount of time though in comparison to the rest

@yarikoptic If you mean when rendering, it's less about that, and more that poster pages aren't guaranteed to have the word poster in their `name`/`id`/etc. So the assumption that we can do a `name.replace('poster', '')` to wipe it out is invalid/not generalisable.

src/components/templates/PosterPage/PosterPage.tsx
0xdevalias: Potentially.. the thinking here is that it's better to do the calculation once rather than every single time the component function is called to check if it should re-render (even if nothing changes when it does). React function components are called a LOT (anytime any of its parent's get called to check if they need to re-render at a minimum). So my theory/perception here is that even if the memo takes slightly longer once, it's probably going to be net-'cheaper' than doing the replace every single time the component is called.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
yarikoptic: stripped away my stripping of the poster.  It still looks good and indeed more factual -- who knows, might be a "demo" since says "Search posters/demos". So now the diff is really tiny ;-)

src/components/templates/PosterPage/PosterPage.tsx
yarikoptic: gone now. IMHO can't be simpler now - only aesthetically inspired folks needing separate xml/html markup should complain ;-)

0xdevalias:
> image
> the design for the screening room could be a good way to go, (number in the top left corner of the preview) I would not want for counting to be implemented in the title as ultra long titles would be made longer and it's just not friendly.

@sofisparkle I believe one of the features OHBM called out as missing from the `PosterPreview` already is the count of users there (one of the solutions I detailed on email), so if that 'belongs' in the same place in this design, that top left corner will already be 'used up' once fixed.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -41,7 +41,9 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
   return (
     <div className={posterClassnames} onClick={handleEnterVenue}>
-      <p className="PosterPreview__title">{title}</p>
+      <p className="PosterPreview__title">
+        {posterVenue.name}: {title}
+      </p>
```
0xdevalias, line 46: @sparkletown/product @mike-lvov Would you be happy with this change being made (to show the `posterVenue` name in full + the poster title) instead of https://github.com/sparkletown/internal-sparkle-issues/issues/755 (which isn't implementable in a generalisable way as we don't (to my knowledge) have a concept of 'poster number' outside of the implied form of it through `posterVenue.name`)

The changes in this file represent the `PosterPreview` 'cards' in the `PosterHall`.

And if so, would you be happy for that to be a change made/merged into our core codebase? Or would we do this as an `env/ohbm` specific hack?

src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -77,7 +77,9 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
         <div />
 
         <div className="PosterPage__header--middle-cell">
-          <p className="PosterPage__title">{title}</p>
+          <p className="PosterPage__title">
+            {venue.name}: {title}
+          </p>
```
0xdevalias, line 82: @sparkletown/product @mike-lvov Would you be happy with this change being made (to show the `posterVenue` name in full + the poster title) instead of https://github.com/sparkletown/internal-sparkle-issues/issues/755 (which isn't implementable in a generalisable way as we don't (to my knowledge) have a concept of 'poster number' outside of the implied form of it through `posterVenue.name`)

The changes in this file represent how it's displayed in the `PosterPage` venue itself.

And if so, would you be happy for that to be a change made/merged into our core codebase? Or would we do this as an `env/ohbm` specific hack?

0xdevalias (commented):
@sparkletown/product @mike-lvov Would be good to get your thoughts/insights/etc on this.

src/components/templates/PosterPage/PosterPage.tsx
yarikoptic: @sparkletown/product @mike-lvov pretty please?

0xdevalias: @sofisparkle ?

mike-lvov: Looks good

mike-lvov: >  Or would we do this as an env/ohbm specific hack?

OHBM specific hack as venueName is not restricted to be Poster{NNN}

mike-lvov (changes requested):
Blocking this just not to merge it yet as in contains recent `staging` updates, which are not in `env/ohbm` yet

yarikoptic:
Should I rebase it against env/ohbm ?

mike-lvov:
> Should I rebase it against env/ohbm ?

I don't think it will help since the merge commit from staging is there.

yarikoptic:
> > Should I rebase it against env/ohbm ?
> 
> I don't think it will help since the merge commit from staging is there.

That's ok, I can rebase only relevant commits and abandon all the merges

yarikoptic:
done - now it is cleanly against env/ohbm

mike-lvov:
@yarikoptic looks good, merging in

PR #1452, opened 2021-05-27
refactor usePosters to display a random selection of posters when no searchQuery

### LINKED ISSUE #1449: posterhall: randomize posters if no search query given
to not overpromote the lucky first X and leave the others behind.

randomization sounds like the easiest/best way to give fair exposure

yarikoptic:
Closes #1449

works nicely
- upon cleaning up search query I get a new random selection ;-)

### DIFF
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -1,7 +1,7 @@
 import { useState, useMemo, useCallback } from "react";
 import { VenueTemplate } from "types/venues";
 import Fuse from "fuse.js";
-import { sampleSize } from "lodash";
+import { shuffle } from "lodash";
 
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
@@ -90,13 +90,11 @@ export const usePosters = (posterHallId: string) => {
   const searchedPosterVenues = useMemo(() => {
     const normalizedSearchQuery = searchQuery.trim();
 
-    if (!normalizedSearchQuery) return filteredPosterVenues;
-
     const tokenisedSearchQuery = tokeniseStringWithQuotesBySpaces(
       normalizedSearchQuery
     );
 
-    if (tokenisedSearchQuery.length === 0) return filteredPosterVenues;
+    if (tokenisedSearchQuery.length === 0) return shuffle(filteredPosterVenues);
 
     return fuseVenues
       .search({
@@ -117,7 +115,7 @@ export const usePosters = (posterHallId: string) => {
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
 
   const displayedPosterVenues = useMemo(
-    () => sampleSize(searchedPosterVenues, displayedPostersCount),
+    () => searchedPosterVenues.slice(0, displayedPostersCount),
     [searchedPosterVenues, displayedPostersCount]
   );
 
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -1,6 +1,7 @@
 import { useState, useMemo, useCallback } from "react";
 import { VenueTemplate } from "types/venues";
 import Fuse from "fuse.js";
+import { sampleSize } from "lodash";
 
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
@@ -116,7 +117,7 @@ export const usePosters = (posterHallId: string) => {
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
 
   const displayedPosterVenues = useMemo(
-    () => searchedPosterVenues.slice(0, displayedPostersCount),
+    () => sampleSize(searchedPosterVenues, displayedPostersCount),
     [searchedPosterVenues, displayedPostersCount]
   );
 
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -1,9 +1,10 @@
 import { useState, useMemo, useCallback } from "react";
-import { VenueTemplate } from "types/venues";
 import Fuse from "fuse.js";
 import { shuffle } from "lodash";
 
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
+
+import { VenueTemplate } from "types/venues";
 
 import { posterVenuesSelector } from "utils/selectors";
 import { tokeniseStringWithQuotesBySpaces } from "utils/text";
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -91,6 +91,8 @@ export const usePosters = (posterHallId: string) => {
   const searchedPosterVenues = useMemo(() => {
     const normalizedSearchQuery = searchQuery.trim();
 
+    if (!normalizedSearchQuery) return shuffle(filteredPosterVenues);
+
     const tokenisedSearchQuery = tokeniseStringWithQuotesBySpaces(
       normalizedSearchQuery
     );


src/hooks/posters.ts
```diff
@@ -78,7 +79,7 @@ export const usePosters = (posterHallId: string) => {
 
   const searchedPosterVenues = useMemo(() => {
     if (!searchQuery)
-      return filteredPosterVenues.slice(0, displayedPostersCount);
+      return sampleSize(filteredPosterVenues, displayedPostersCount);
```
0xdevalias, line 82: > Depending on where it's put, we'll need to pay attention to ensure that it doesn't needlessly 're-randomize' things when the user wouldn't expect it to. Eg. if we just put it in the existing `searchedPosterVenues` `useMemo` then it will re-randomize when any of `[searchQuery, fuseVenues, filteredPosterVenues, displayedPostersCount]` change, which is likely undesirable.
> 
> I would probably do it in it's own `useMemo` that relies on just `filteredPosterVenues`, and possibly  `searchQuery`. If we include `searchQuery`, then it would have a guard that `if (searchQuery) return` to avoid needlessly processing things. Including `searchQuery` means that whenever the search is used (and then cleared again) we would end up with a different set of random posters.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1449#issuecomment-849198650_

- `searchedPosterVenues` memo's on `[searchQuery, fuseVenues, filteredPosterVenues, displayedPostersCount]`
  - ✅ we definitely want to react when `searchQuery` changes
  - ✅ `fuseVenues` only memo's on `filteredPosterVenues`
  - ✅ we would definitely want to react if `filteredPosterVenues` ever changed
  - ⚠️ i'm not sure we should be re-randomising when `displayedPostersCount` changes though.. as that means clicking 'show more' will completely change everything you see, which is undesirable

See my proposed solution that would avoid this in the quoted text above: creating a new `useMemo` that depends on only `filteredPosterVenues` and possibly `searchQuery`. Probably call this `randomisedPosterVenues` or similar.

This can then probably be passed into `searchedPosterVenues` and used as it's 'guard `return`'.

That way it should only randomise when:

- `filteredPosterVenues` changes (which will mostly only happen when `isLive` changes, which is effectively akin to changing the `searchQuery`)
- `searchQuery` changes, and it is cleared (which we want/expect)
- whenever the component is first rendered and there is no `searchQuery` set (also what we want/expect)

---

@yarikoptic Clean solution choice with `lodash`'s `sampleSize` by the way. I wouldn't have even considered that 👌🏻✨ 

---

**Further Reading:**

- https://lodash.com/docs/4.17.15#sampleSize
  - > Gets `n` random elements at unique keys from collection up to the size of collection.

src/hooks/posters.ts
```diff
@@ -1,6 +1,7 @@
 import { useState, useMemo, useCallback } from "react";
 import { VenueTemplate } from "types/venues";
 import Fuse from "fuse.js";
+import { sampleSize } from "lodash";
 
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
```
0xdevalias, line 6: This issue isn't from your PR, but may as well fix it while i'm here:

```suggestion
import { useState, useMemo, useCallback } from "react";
import Fuse from "fuse.js";
import { sampleSize } from "lodash";

import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";

import { VenueTemplate } from "types/venues";
```

0xdevalias (changes requested):
Looks like we need to tweak the logic here slightly and extract a new `useMemo`, then this should be good to go!

src/hooks/posters.ts
yarikoptic: Hm... Good point about displayedPostersCount and Show more action. For that we would need to reshuffle all once (more expensive) and then do the slicing... Let me sleep on it

0xdevalias: > Let me sleep on it

While you may come up with a different solution, I pretty much spelled out the solution as I see it above. If I get a moment to, I will put that in code.

yarikoptic: FTR: #1460 might be bringing us close(r) to that implementation (or not ;-) ). In any case probably worth waiting for that one to land or build on top of it instead of staging.

0xdevalias: FYI: #1460 is merged now.

yarikoptic: Thanks for that merge and the buzz here. I have merged staging and resolved merge by switching from sampleSize to full `shuffle` as part of the resolution.  Now we need some memo to ensure that for a user it would be a singular consistent shuffle, or otherwise they return to the posterhall with always reshuffled selection.  And I don't get why since the entire `searchedPosterVenues` where `shuffle` is called is already in `useMemo` on `[searchQuery, fuseVenues, filteredPosterVenues]` which should be the same: `searchQuery` should be empty, `fuseVenues` is also useMemo on `filteredPosterVenues`.  `filteredPosterVenues` is a memo on `[posterVenues, liveFilter]`... sorry -- don't know exact mechanics of `useMemo` memoization in React.

0xdevalias: The mechanics are basically when any of the things listed in that department array change, it will break the memo.

I can’t say for certain, but I believe there may have been another var in that department array previously that may have broken it more often (explained the full situation in one of my earlier comments here).

So it’s possible that way things were changed in the recently merged PR might altered my original assumptions/etc slightly.

Couldn’t say for certain without looking closer at the code changes + what I said above originally. Won’t be doing that till next week (late Friday night here currently; or.. technically early Saturday I suppose)

src/hooks/posters.ts
```diff
@@ -89,7 +91,7 @@ export const usePosters = (posterHallId: string) => {
   const searchedPosterVenues = useMemo(() => {
     const normalizedSearchQuery = searchQuery.trim();
 
-    if (!normalizedSearchQuery) return filteredPosterVenues;
```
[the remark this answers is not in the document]
0xdevalias: (Delete line below this: if (!normalosedSearchQuery) etc

src/hooks/posters.ts
0xdevalias: I think the thing that would previously break memo was the display poster count variable. Which has been moved into its own memo on that other PR. So off the top of my head I think this one will just work now (once the duplicated line I called out in other comments is fixed up)

src/hooks/posters.ts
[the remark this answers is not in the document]
yarikoptic: removed, moved shuffle. merged staging. 
going back from a poster to posterhall still reshuffles - and it looks like it might first show original one for a moment but then reshuffle happens - so something might indeed be "flipping" (even if just being changed from null to undefined or smth like that?)
pushed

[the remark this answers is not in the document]
yarikoptic: well, this one I think is resolved -- I will move my comment to the above one

src/hooks/posters.ts
yarikoptic: removed, moved shuffle. merged staging.
going back from a poster to posterhall still reshuffles - and it looks like it might first show original one for a moment but then reshuffle happens - so something might indeed be "flipping" (even if just being changed from null to undefined or smth like that?)
pushed

0xdevalias: Ah.. so going to a poster and back again is an entirely different batch of problems, as that completely unrenders this component (and thus the memos) to load the PosterPage, so when you come back to the PosterHall, it’s basically ‘starting from scratch.

To solve that we would basically have to move the data out of the hook, memo won’t help us here.

Global redux state is the best solution that comes to mind, but there is more boilerplate/react specifics involved to get into that side of things.

0xdevalias: As for the ‘seems to show first then changes again’, would need to look deeper at the code/debug to know why that might be.

yarikoptic: May be it was some perceptual trick... I will try to see if I could catch it. Use the an easy way to instrument useMemo to see when it returns memorized value? (for new I could just log)

yarikoptic: I thought of another approach: we could just use seeded (based on user or session id) RNG for shuffle - will make it always consistent.  Unfortunately with lodash it isn't possible natively but there is a workaround which could be tried: https://github.com/lodash/lodash/issues/3289#issuecomment-434854622 . But I wonder if such global seeding would not be carrying some undersired side effect on another point in sparkle which uses some "random" .  So an alternative could be to find another `shuffle` which would use a dedicated (seeded) RNG.

yarikoptic: ha -- for all means and purposes of this PR -- we do not need to shuffle AT ALL!  We can just choose random offset within the array of posters and roll circularly to that index.  Then each user would get a different selection.  It would be fast since would just require slicing. We better do that!.  Now the question is only on the best way to get some session or user id into that poster.ts:usePosters I guess.  Or there could be may be some global `sessionSpecificRoll(a: array) -> rolled_a` somewhere?

0xdevalias: While the slice would solve the ‘need’ for not always showing the same top posters, it feels like more of a hack than a solution, as the properly random seems a better choice.

Seeded random for it seems like a neat way to solve the problem. Definitely needs to be only at a local (not global) scale if we did that. Not sure if it’s the best solution though. If we needed to get details about the user we would likely just use the useUser hook.

See my comments about using redux and/or react context provider patterns on the new issue you opened recently for what I feel is potentially a more ‘proper’/‘better’ solution.

(fwiw, the seeded random + full shuffle still intrigues me, but even that would need to be mixed with some global state to store the seed, at which point I figure we may as well just keep current patterns that work well and use the global state/context provider for it all)

yarikoptic:
> mixed with some global state to store the seed,
Not really, if userId is used as a seed - it is already stored

0xdevalias:
> Not really, if userId is used as a seed - it is already stored

@yarikoptic Yet that somewhat defeats the purpose of randomising the posters in the first place doesn't it? As then for each user one set of posters is always going to unfairly biased to be shown at the top.

src/hooks/posters.ts
0xdevalias: > > mixed with some global state to store the seed,
> 
> Not really, if userId is used as a seed - it is already stored
> 
> _Originally posted by @yarikoptic in https://github.com/sparkletown/sparkle/issues/1452#issuecomment-850921738_

0xdevalias: > > Not really, if userId is used as a seed - it is already stored
> 
> @yarikoptic Yet that somewhat defeats the purpose of randomising the posters in the first place doesn't it? As then for each user one set of posters is always going to unfairly biased to be shown at the top.
> 
> EDIT: Also, this conversation would be better continued in the comment thread rather than spamming this main feed.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1452#issuecomment-850925408_

yarikoptic: > While the slice would solve the ‘need’ for not always showing the same top posters, it feels like more of a hack than a solution, as the properly random seems a better choice.

thought about it a bit more:
- doing user specific "roll" to position user within a predefined order of posters might actually be more "biologically plausible": as if a visitor is dumped somewhere in a poster hall -- he/she is not getting a "random" selection but merely a different position within a set order
- we can even avoid using "random" altogether (to not bring another library in etc): we just need some function which spreads uses evenly.  checksum (over userId), such as md5 or any other, is as good. So would be just needed to take checksum(userId) remainder from division over `posterVenues.length` and have our offset.

So ranom or not it should be done within `posters.ts:usePosters` but ignorant in react me still can't grasp how to bring `useUser` in there + we might need to parametrize it by a venue -- not all venues should have them randomized I guess.  But then where such parametrization should live -- sounds like in a DB as a property of a poster room. I.e. main poster hall -- random.  The others, not sure.

0xdevalias: Even with your latest thoughts/suggestion above, I'm still heavily leaning towards the React context based solution I described previously in https://github.com/sparkletown/sparkle/pull/1452#discussion_r641999204 / related.

mike-lvov (approved):
LGTM

mike-lvov (changes requested):
Withdrawing my approval, once I read the thread https://github.com/sparkletown/sparkle/pull/1452#pullrequestreview-669697532 

@yarikoptic any updates on this?

src/hooks/posters.ts
mike-lvov: > Yet that somewhat defeats the purpose of randomising the posters in the first place doesn't it? As then for each user one set of posters is always going to unfairly biased to be shown at the top.

I guess the desired effect is to spread the load on the posters. + Even with the context & storing value in redux, I guess the problem is still there? User will get a random set of posters, when he enters the platform

yarikoptic: The problem I any trying to solve is not technical (load on the posters) but social (unfair better visibility of the posters on top of the list if all users get the same listing). What I dislike about current PR implantation is that every visit by the same user gets a new random order. What I think is desirable is either random (but reproducible for the same user) order, or just a random roll of the listing (again reproducible for the user) which would be less demanding on the user since just needs 1 single random number for each user to roll by.

yarikoptic: Re solution, unfortunately I still have little too no clue in what "redux and/or react context provider patterns" would entail since I am not a react developer, just trying to contribute with my very little of JS knowledge to provide needed functionality/features. So any help would be appreciated.

mike-lvov: IMO we should use shuffle-seed. And use `userId` as seed for every user as per @yarikoptic's 

> I thought of another approach: we could just use seeded (based on user or session id) RNG for shuffle - will make it always consistent

https://www.npmjs.com/package/shuffle-seed

@0xdevalias wdyt? Looks like the only option, which is fast & decently functional

0xdevalias: @mike-lvov I already explained my thoughts/solution as I see it above.

mike-lvov: @0xdevalias 

> the only option, **which is fast & decently functional**

0xdevalias: I stand by my statement above? How is it not fast and decently functional?

goran-peoski-work:
Switched base branch to env/ohbm because of the specifics of the current PR state

goran-peoski-work (approved):

src/hooks/posters.ts
yarikoptic: @mike-lvov re shuffle-seed -- looking at the code: https://github.com/webcaetano/shuffle-seed/blob/master/shuffle-seed.js#L4  (and below) it seems to use `seedrandom` library I have considered before (https://github.com/sparkletown/sparkle/pull/1452#discussion_r641941290 and below), and uses it as just `Math.seedrandom` and not `new Math.seedrandom` thus (I guess, since not JS programmer) making that seeding global (according to https://github.com/davidbau/seedrandom docs if I got it right), which I do not think would be desired.  I (or someone) would need first to check if there would be no side-effect on global RNG

@0xdevalias the solution you explained (in https://github.com/sparkletown/sparkle/pull/1452#discussion_r642846045 and the chain there) boils down to "we better use (seeded) shuffle" (as opposed to rolling the list) + "use Redux as mentioned somewhere else". Insofar I found no way to get seeded shuffle without affecting global RNG (see e.g. paragraph above). As for shuffle vs roll, although I agree that shuffle would not be worse, for all means and purposes (as described two comments back in https://github.com/sparkletown/sparkle/pull/1452#discussion_r648794892)  roll would be sufficient and would be much easier/faster to just get a new list of posters (as `filteredPosterVenues[random_user_specific_index:] + filteredPosterVenues[:random_user_specific_index]` if expressed in Python I speak better) and be done to the same overall desired effect of distributing visitors across the entire list of posters.

<details>
<summary>here is a quick&dirty patch for trying out slicing, although due to logic code duplication needs to be added into the 2nd invocation as well</summary> 

```shell
diff --git a/package.json b/package.json
index 30de342dd..67e6ba76c 100644
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "remark-gfm": "^1.0.0",
     "sass": "^1.32.5",
     "scheduler": "^0.19.1",
+    "seedrandom": "^3.0.5",
     "styled-components": "^5.2.1",
     "three": "^0.119.1",
     "three-obj-mtl-loader": "^1.0.3",
index 82649c773..9ed970ae7 100644
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -2,6 +2,8 @@ import { useState, useMemo, useCallback } from "react";
 import Fuse from "fuse.js";
 import { shuffle } from "lodash";
 
+// import seedrandom from "seedrandom";
+
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
 import { VenueTemplate } from "types/venues";
@@ -91,7 +93,14 @@ export const usePosters = (posterHallId: string) => {
     const normalizedSearchQuery = searchQuery.trim();
 
-    if (!normalizedSearchQuery) return shuffle(filteredPosterVenues);
+    var seedrandom = require('seedrandom');
+    const rng = new seedrandom('hello3.');  // TODO: userId not fixed string
+    const userSliceIndex = Math.floor(rng() * filteredPosterVenues.length);
+
+    if (!normalizedSearchQuery) {
+        return filteredPosterVenues.slice(userSliceIndex, filteredPosterVenues.length).concat(
+                 filteredPosterVenues.slice (0, userSliceIndex));
+    }
 
     const tokenisedSearchQuery = tokeniseStringWithQuotesBySpaces(
       normalizedSearchQuery

```
</details>

As for "Redux implementation" - if you or @mike-lvov could give me a hand on that with either clear pointers or just e.g. making this PR into a "Redux'ed version", would be appreciated.

Meanwhile I will rebase it on top of `env/ohbm` due to change of the base in https://github.com/sparkletown/sparkle/pull/1452#issuecomment-860625398

goran-peoski-work: **TL;DR:**  You can be relatively safe if you get the `sampleSize` code and probably replace that `Math.random()` call with
```
// Local PRNG: does not affect Math.random.
const seedrandom = require('seedrandom');
const rng = seedrandom(seed); // string taken from user and/or usePosters params maybe
```

I got an idea from
> according to https://github.com/davidbau/seedrandom docs if I got it right

something like: _when in doubt, don't read the docs, but the code_. 

- By looking at `sampleSize`, the code is at https://github.com/lodash/lodash/blob/master/sampleSize.js and released under The MIT License (Copyright JS Foundation and other contributors), meaning you can modify it and re-use it.

- By looking at `seedrandom`, the code at https://github.com/davidbau/seedrandom/blob/released/seedrandom.js, there is one comment that sticks out:
  ```
  // When seedrandom.js is loaded, we immediately mix a few bits
  // from the built-in RNG into the entropy pool.  Because we do
  // not want to interfere with deterministic PRNG state later,
  // seedrandom will not call math.random on its own again after
  // initialization.
  ```
  The `new` ambiguity stems from how it has been used in browsers since v1, but that's not npm/nodejs convention
  ```
  // If called as a method of Math (Math.seedrandom()), mutate
  // Math.random because that is how seedrandom.js has worked since v1.0.
  if (is_math_call) { math[rngname] = prng; return seed; }

  // Otherwise, it is a newer calling convention, so return the
  // prng directly.
  ```

0xdevalias: FYI since it was apparently never linked directly in this thread (only referred to), this is the method I keep suggesting/referring back to: https://github.com/sparkletown/sparkle/issues/1477#issuecomment-850917978

goran-peoski-work: @yarikoptic here https://github.com/sparkletown/sparkle/tree/pr/random-posters-redux-way--1452 is "The Redux Way" version I could come up following @0xdevalias prescription in the short time I had to do it. It is a POC and work in progress, so does not produce the desired result at this moment

Also, at this branch https://github.com/sparkletown/sparkle/tree/pr/random-posters--1452 there is a version that uses modified `sampleSize` and `seedrandom` with `userId` and 
```javascript
const POSTERS_TIME_GRANULARITY = 1000 * 60 * 60; // one hour in ms
```
for more or less persistent posters (but still randomized)

goran-peoski-work: if it makes it easier (certainly complicates PR management) I've created a PR for the `seedrandom` version at https://github.com/sparkletown/sparkle/pull/1549

yarikoptic: THANK YOU @goran-peoski-work for the #1549 . I will close this PR .
One question: in aforementioned https://github.com/sparkletown/sparkle/tree/pr/random-posters-redux-way--1452 you used Redux and it gave an awesome feature -- being able to return to the "searched up" listing of posters. IT IS JUST AWESOME!  Could it be brought into #1549 as well, pretty please????

yarikoptic:
superseded by https://github.com/sparkletown/sparkle/pull/1549 but some discussion might still be ongoing above in https://github.com/sparkletown/sparkle/pull/1452#discussion_r651868410 to may be gain some more super-powers there.

src/hooks/posters.ts
yarikoptic: could redux also save the state of the scroll bar in the long list of posters? if it could  -- it would have been really perfect! ;-)

goran-peoski-work: Unfortunately, saving the state in Redux, even though more structured and idiomatic solution, would've required more time to do reliably and without introducing regressions with the short time at hand.

In general, though, the provider pattern is a useful tool for managing application state. You can read further at
- https://flexiple.com/react/provider-pattern-with-react-context-api <- general explanation
- https://reactjs.org/docs/hooks-reference.html#usecontext <- hooks based implementation

0xdevalias: RE: https://github.com/sparkletown/sparkle/pull/1452#discussion_r651735570

@goran-peoski-work Looking at the compare view: https://github.com/sparkletown/sparkle/compare/staging...pr/random-posters-redux-way--1452

In `src/pages/VenuePage/TemplateWrapper.tsx` you're using `PostersProvider` twice. Whereas it should be used only once wrapping both PosterHall and PosterPage.

Something like the following done on line 167 where `template` is rendered in the JSX (though formatted properly:

```diff
- {template}
+ {[VenueTemplate.posterhall, VenueTemplate.posterpage].includes(
+     venue.template
+ ) ? (
+     <PostersProvider value={postersState}>{template}</PostersProvider>
+ ) : (
+     template
+ )}
```

The way you're using `let postersState = usePosters(venue.id);` outside of the provider doesn't look like it's going to be right either, but not 100%

0xdevalias: FYI: None of this is using redux. It's just using React context provider patterns. Redux was a 3rd alternative suggested, but one we don't have as good patterns established around.

0xdevalias: > and it gave an awesome feature -- being able to return to the "searched up" listing of posters. IT IS JUST AWESOME! 

Almost like it was the optimal solution to be used from the very beginning.. and had people followed the instructions I laid out originally we would have had it solved well, ages ago, and people wouldn't be stressing to rush in hacky suboptimal implementations now.

> could redux also save the state of the scroll bar in the long list of posters? if it could -- it would have been really perfect! ;-)

As above, it's not using redux at all, just context/provider patterns, which basically allow us to store any sort of JavaScript data we want.

Since there are APIs to both check and set the scroll position, I would say it would absolutely be possible. Though i'm not sure of the best way to implement it off the top of my head:

- https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset
- https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll
- https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

My gut feel is that using `scrollIntoView` to somehow re-focus the poster that we're returning back from would be the cleanest implementation, though for that you need a reference to the element (`useRef` or similar maybe?) or could use React's 'callback ref' pattern potentially.

Not sure of the best way to know what the ID of the venue we're returning back from is off hand. React Router lets state get passed along when navigating, so potentially refactoring the back button logic to pass the `venueId` we're navigating from as part of the state would work.

- https://reactrouter.com/web/api/Hooks/usehistory
  - https://reactrouter.com/web/api/history
    - > `push(path, [state])`
- https://reactrouter.com/web/api/Hooks/uselocation
  - https://reactrouter.com/web/api/location
    - `state`

0xdevalias:
@yarikoptic @goran-peoski-work @mike-lvov Feedback on the proper React context/provider solution implementation + further ways to improve it to also keep the scroll state in https://github.com/sparkletown/sparkle/pull/1452#discussion_r652529136 + https://github.com/sparkletown/sparkle/pull/1452#discussion_r652537451

goran-peoski-work:
As https://github.com/sparkletown/sparkle/compare/staging...pr/random-posters-redux-way--1452 is still work in progress, comments like https://github.com/sparkletown/sparkle/pull/1452#discussion_r652529136 will help me push it forward to working and idiomatic solution. It may just take a bit more time.

Two things to note:
- can't judge of what "far superior choice" is because at different occasions, different yardsticks can be used to measure it, but I do think it's more in line with how React was intended to be used, thus more idiomatic, robust and maintainable way going forward
- the branch name is a bit of a misnomer, was related to a comment, so was useful to name it for the purpose. going forward, before another PR is created, I should resolve that as well.

goran-peoski-work:
Here is a PR https://github.com/sparkletown/sparkle/pull/1557 for the Context/Provider solution with implemented TemplateWrapper suggestions from @0xdevalias , but not sure if it works as intended.

src/hooks/posters.ts
0xdevalias: Instead of passing on the last venueId through state, we could probably just read it from the router history directly:

- https://reactrouter.com/web/api/Hooks/usehistory
- https://reactrouter.com/web/api/history

PR #1460, opened 2021-05-27
improve PosterHall search by adjusting fuse.js options + related

### LINKED ISSUE #1391: Search results in poster hall are not based on actual data
Searching for "ayzasdasdaaaaaa" gave me 3 posters.  May be expected ATM (i.e. search is not really there)

### LINKED ISSUE #1453: PosterHall 'show more posters'  is shown even when there are no more posters to be displayed
> The next UX fiasco which is not
> fixed in this PR anyhow is "Show more posters" for a search query which has no
> more to show.
> 
> _Originally posted by @yarikoptic in https://github.com/sparkletown/sparkle/pull/1445_

mike-lvov:
**Note:** This is a re-created PR (to fix CI issues/etc) for the work tested & developed by @yarikoptic in #1443 

**Edit:** While the original PR accidentally overwrote the commit history, that has since been fixed and restored. Details in https://github.com/sparkletown/sparkle/pull/1460#issuecomment-850127899

---

This is basically a copy of #1443 + some additions/tweaks.

The reason for copying it was to test it & slightly clean up the code

closes #1391
closes #1453 

## Original PR Description

> At large closes #1391 (**Edit by @0xdevalias:** also closes https://github.com/sparkletown/internal-sparkle-issues/issues/745) although remains notably suboptimal.  With settings @margulies came up with seems to go a few steps further than my exploration (initial commit) 
> - additional explicit trimming of the search query (we aren't into [Whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language)) here aren't we? but adding a space trailing space has an immediate detrimental effect)
> - split by space into words and search for each within `$and` across all keys.  See below for more info on how plain fuse.js seems not capable of that
> - allow for `" "` to enclose phrase to not be split
> 
> With the above changes search becomes **pragmatically useful**! That splitting **resolved** following gotchas: 
> - `LastName FirstName` would return no result whenever `FirstName LastName` is ok.
> - results often make no real sense from UX PoV: e.g. there is a poster by Tsai with "connectivity" in the title.  Entering Tsai - shows the poster. Entering "Tsai connectivity" shows bunch of posters none of which is of Tsai
> 
> edits:
> -  I think this is the most related issue https://github.com/krisk/Fuse/issues/235 (closed without resolution pointing to built-in deficiency of fuse.js for our use case)
>   - to workaround people breed monsters like https://github.com/krisk/Fuse/issues/235#issuecomment-821220246
> - the approach of the bot closing issues without resolution there is really annoying, keep running into them (tokenize was removed but no changelog https://github.com/krisk/Fuse/issues/511, so not clear why/how to workaround... started to be removed in https://github.com/krisk/Fuse/commit/4343f71e4159e5ee046a821b2be700eee0708148 in favor of "extended" mode I guess; but in extended mode you seems need to specify fields for logical operation)
> - with that in mind it might need some JS foo to first split query string, and prep it to be fed to "advanced" mode
>   - may be dirty sufficient workaround would be to meld all 3 fields into one to give to fuse
>   - or just perform separate queries per each split term and take intersection

### DIFF
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -22,6 +22,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
   const {
     posterVenues,
     isPostersLoaded,
+    hasHiddenPosters,
 
     increaseDisplayedPosterCount,
 
@@ -55,7 +56,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
       </div>
 
       <div className="PosterHall__more-button">
-        {isPostersLoaded && (
+        {hasHiddenPosters && isPostersLoaded && (
           <Button onClick={increaseDisplayedPosterCount}>
             Show more posters
           </Button>
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -32,6 +32,8 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
     setLiveFilter,
   } = usePosters(venue.id);
 
+  const shouldShowMorePosters = isPostersLoaded && hasHiddenPosters;
+
   const renderedPosterPreviews = useMemo(() => {
     return posterVenues.map((posterVenue) => (
       <PosterPreview
@@ -56,7 +58,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
       </div>
 
       <div className="PosterHall__more-button">
-        {hasHiddenPosters && isPostersLoaded && (
+        {shouldShowMorePosters && (
           <Button onClick={increaseDisplayedPosterCount}>
             Show more posters
           </Button>
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -5,7 +5,7 @@ import Fuse from "fuse.js";
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
 import { posterVenuesSelector } from "utils/selectors";
-import { breakStringWithQuotesBySpaces } from "utils/text";
+import { tokeniseStringWithQuotesBySpaces } from "utils/text";
 
 import { isLoaded, useFirestoreConnect } from "./useFirestoreConnect";
 import { useSelector } from "./useSelector";
@@ -90,19 +90,20 @@ export const usePosters = (posterHallId: string) => {
 
     if (!normalizedSearchQuery) return filteredPosterVenues;
 
-    const matchEntries: string[] =
-      breakStringWithQuotesBySpaces(normalizedSearchQuery) ?? [];
+    const tokenisedSearchQuery = tokeniseStringWithQuotesBySpaces(
+      normalizedSearchQuery
+    );
 
-    if (!matchEntries) return filteredPosterVenues;
+    if (tokenisedSearchQuery.length === 0) return filteredPosterVenues;
 
     return fuseVenues
       .search({
-        $and: matchEntries.map((x: string) => {
+        $and: tokenisedSearchQuery.map((searchToken: string) => {
           const orFields: Fuse.Expression[] = [
-            { name: x },
-            { "poster.title": x },
-            { "poster.authorName": x },
-            { "poster.categories": x },
+            { name: searchToken },
+            { "poster.title": searchToken },
+            { "poster.authorName": searchToken },
+            { "poster.categories": searchToken },
           ];
 
           return {
@@ -110,7 +111,7 @@ export const usePosters = (posterHallId: string) => {
           };
         }),
       })
-      .map((fuseSearchItem) => fuseSearchItem.item);
+      .map((fuseResult) => fuseResult.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
 
   const displayedPosterVenues = useMemo(
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -102,7 +102,7 @@ export const usePosters = (posterHallId: string) => {
         //@ts-ignore
         $and: searchQuery
           .trim()
-          .match(/(".*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472
+          .match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472 + fix
           .map((x) => ({
             $or: [
               { "poster.title": x },
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -71,13 +71,25 @@ export const usePosters = (posterHallId: string) => {
   const fuseVenues = useMemo(
     () =>
       new Fuse(filteredPosterVenues, {
-        keys: ["poster.title", "poster.authorName", "poster.categories"],
-        threshold: 0.4, // default 0.6: brings too distant if anyhow related hits
+        keys: [
+          {
+            name: "poster.title",
+            weight: 2,
+          },
+          {
+            name: "poster.authorName",
+            weight: 3,
+          },
+          {
+            name: "poster.categories",
+            weight: 1,
+          },
+        ],
+        threshold: 0.2, // 0.1 seems to be exact, default 0.6: brings too distant if anyhow related hits
         ignoreLocation: true, // default False: True - to search ignoring location of the words.
-        // Otherwise can't find a word in a sentence
-        // includeScore: true   // so it could be displayed!?
-        // distance: 40        //  default 100 - ignored if ignoreLocation: true
-        // useExtendedSearch: true // might be neat!
+        findAllMatches: true,
+        minMatchCharLength: 3,
+        // useExtendedSearch: true,  // might be neat but confusing. might be worthwhile a UI switch
       }),
     [filteredPosterVenues]
   );
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -97,7 +97,7 @@ export const usePosters = (posterHallId: string) => {
           .match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472 + fix
           .map((x) => ({
             $or: [
-              { "name": x },
+              { name: x },
               { "poster.title": x },
               { "poster.authorName": x },
               { "poster.categories": x },
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -99,7 +99,19 @@ export const usePosters = (posterHallId: string) => {
       return filteredPosterVenues.slice(0, displayedPostersCount);
 
     return fuseVenues
-      .search(searchQuery.trim()) // trim so adding space does not kill all results at once
+      .search({
+        //@ts-ignore
+        $and: searchQuery
+          .trim()
+          .split(" ")
+          .map((x) => ({
+            $or: [
+              { "poster.title": x },
+              { "poster.authorName": x },
+              { "poster.categories": x },
+            ],
+          })),
+      })
       .slice(0, displayedPostersCount)
       .map((fuseSearchItem) => fuseSearchItem.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues, displayedPostersCount]);
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -88,7 +88,6 @@ export const usePosters = (posterHallId: string) => {
         threshold: 0.2, // 0.1 seems to be exact, default 0.6: brings too distant if anyhow related hits
         ignoreLocation: true, // default False: True - to search ignoring location of the words.
         findAllMatches: true,
-        minMatchCharLength: 3,
         // useExtendedSearch: true,  // might be neat but confusing. might be worthwhile a UI switch
       }),
     [filteredPosterVenues]
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -99,7 +99,7 @@ export const usePosters = (posterHallId: string) => {
       return filteredPosterVenues.slice(0, displayedPostersCount);
 
     return fuseVenues
-      .search(searchQuery)
+      .search(searchQuery.trim()) // trim so adding space does not kill all results at once
       .slice(0, displayedPostersCount)
       .map((fuseSearchItem) => fuseSearchItem.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues, displayedPostersCount]);
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -103,7 +103,7 @@ export const usePosters = (posterHallId: string) => {
         //@ts-ignore
         $and: searchQuery
           .trim()
-          .split(" ")
+          .match(/(".*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472
           .map((x) => ({
             $or: [
               { "poster.title": x },
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -5,6 +5,7 @@ import Fuse from "fuse.js";
 import { DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT } from "settings";
 
 import { posterVenuesSelector } from "utils/selectors";
+import { breakStringWithQuotesBySpaces } from "utils/text";
 
 import { isLoaded, useFirestoreConnect } from "./useFirestoreConnect";
 import { useSelector } from "./useSelector";
@@ -89,19 +90,21 @@ export const usePosters = (posterHallId: string) => {
 
     if (!normalizedSearchQuery) return filteredPosterVenues;
 
+    const matchEntries = breakStringWithQuotesBySpaces(normalizedSearchQuery);
+
+    if (!matchEntries) return filteredPosterVenues;
+
     return fuseVenues
       .search({
-        //@ts-ignore
-        $and: normalizedSearchQuery
-          .match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472
-          .map((x) => ({
-            $or: [
-              { name: x },
-              { "poster.title": x },
-              { "poster.authorName": x },
-              { "poster.categories": x },
-            ],
-          })),
+        $and: matchEntries.map((x) => ({
+          $or: [
+            { name: x },
+            { "poster.title": x },
+            { "poster.authorName": x },
+            { "poster.categories": x },
+            // @debt Rewrite the search query to match the $and type
+          ] as Record<string, string>[],
+        })),
       })
       .map((fuseSearchItem) => fuseSearchItem.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -69,6 +69,7 @@ export const usePosters = (posterHallId: string) => {
     [posterVenues, liveFilter]
   );
 
+  // See https://fusejs.io/api/options.html
   const fuseVenues = useMemo(
     () =>
       new Fuse(filteredPosterVenues, {
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -106,9 +106,9 @@ export const usePosters = (posterHallId: string) => {
       .map((fuseSearchItem) => fuseSearchItem.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
 
-  const displayedPosterVenues = searchedPosterVenues.slice(
-    0,
-    displayedPostersCount
+  const displayedPosterVenues = useMemo(
+    () => searchedPosterVenues.slice(0, displayedPostersCount),
+    [searchedPosterVenues, displayedPostersCount]
   );
 
   const hasHiddenPosters =
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -72,6 +72,12 @@ export const usePosters = (posterHallId: string) => {
     () =>
       new Fuse(filteredPosterVenues, {
         keys: ["poster.title", "poster.authorName", "poster.categories"],
+        threshold: 0.4, // default 0.6: brings too distant if anyhow related hits
+        ignoreLocation: true, // default False: True - to search ignoring location of the words.
+        // Otherwise can't find a word in a sentence
+        // includeScore: true   // so it could be displayed!?
+        // distance: 40        //  default 100 - ignored if ignoreLocation: true
+        // useExtendedSearch: true // might be neat!
       }),
     [filteredPosterVenues]
   );
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -86,7 +86,7 @@ export const usePosters = (posterHallId: string) => {
   );
 
   const searchedPosterVenues = useMemo(() => {
-    if (!searchQuery)
+    if (!searchQuery.trim())
       return filteredPosterVenues.slice(0, displayedPostersCount);
 
     return fuseVenues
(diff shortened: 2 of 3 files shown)


yurii-lubynets (approved):
LGTM

yarikoptic:
FWIW: :+1: from me!   ;-)
Note: Would close a yet not filed issue about "More posters" button persistently visible.

src/hooks/posters.ts
```diff
@@ -71,24 +71,53 @@ export const usePosters = (posterHallId: string) => {
...
+
+    if (!normalizedSearchQuery) return filteredPosterVenues;
 
     return fuseVenues
-      .search(searchQuery)
-      .slice(0, displayedPostersCount)
+      .search({
+        //@ts-ignore
+        $and: normalizedSearchQuery
+          .match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g) // source: https://stackoverflow.com/a/16261693/1265472
```
denisdimitrov, line 96: Should we extract this regex somewhere else, is there any chance it might be used in another place other than this one?

src/hooks/posters.ts
```diff
@@ -71,24 +71,53 @@ export const usePosters = (posterHallId: string) => {
...
-      return filteredPosterVenues.slice(0, displayedPostersCount);
+    const normalizedSearchQuery = searchQuery.trim();
+
+    if (!normalizedSearchQuery) return filteredPosterVenues;
 
     return fuseVenues
-      .search(searchQuery)
-      .slice(0, displayedPostersCount)
+      .search({
+        //@ts-ignore
```
denisdimitrov, line 94: What seems to be the issue here? Can we remove this and fix it easily?

src/hooks/posters.ts
```diff
@@ -71,24 +71,53 @@ export const usePosters = (posterHallId: string) => {
...
-  }, [searchQuery, fuseVenues, filteredPosterVenues, displayedPostersCount]);
+  }, [searchQuery, fuseVenues, filteredPosterVenues]);
+
+  const displayedPosterVenues = useMemo(
+    () => searchedPosterVenues.slice(0, displayedPostersCount),
+    [searchedPosterVenues, displayedPostersCount]
+  );
+
+  const hasHiddenPosters =
+    searchedPosterVenues.length > displayedPosterVenues.length;
```
denisdimitrov, line 115: `searchedPosterVenues` is returning `filteredPosterVenues`  When `(!normalizedSearchQuery)`. 
After that `filteredPosterVenues` seems return `posterVenues` which is using this selector `posterVenuesSelector`. Looks like it can be undefined and this `.length` can fail. 

Maybe you should have `const posterVenues = useSelector(posterVenuesSelector) ?? []` in `usePosterVenues`

denisdimitrov (dismissed):

src/hooks/posters.ts
denisdimitrov: Nevermind, I just saw this: https://github.com/sparkletown/sparkle/pull/1460/files#diff-06ebbfcf8e627f41bd6b15a2648431dc8a5dfd21f7d64f9bc798ed55051c0c58R35

You can ignore this comment and resolve the conversation.

src/hooks/posters.ts
yarikoptic: IIRC it was about untyped 
```
 .map((x) =&gt; ({            $or: [...
```
below

src/hooks/posters.ts
yarikoptic: do you have unit tests for the codebase? extra :+1: for moving away would then be -- would allow to get a nice unittest to ensure its correct operation.

yarikoptic:
> Note: Would close #1453 ~a yet not filed issue~ about "More posters" button persistently visible.

<details>
<summary>FWIW -- confirming! (after full reload to get rid of Memos)</summary> 

image

</details>

yarikoptic (approved):

src/hooks/posters.ts
mike-lvov: > do you have unit tests for the codebase?
Unfortunately, no

0xdevalias: Technically we have the infrastructure for it, but it needs a little love, and is only used in a couples of places at the moment (possible only one, testing the package-lock file). I believe the filename ends in .test.ts or similar if you wanted to find it.

src/hooks/posters.ts
```diff
@@ -71,24 +71,53 @@ export const usePosters = (posterHallId: string) => {
...
+        ],
+        threshold: 0.2, // 0.1 seems to be exact, default 0.6: brings too distant if anyhow related hits
+        ignoreLocation: true, // default False: True - to search ignoring location of the words.
+        findAllMatches: true,
       }),
     [filteredPosterVenues]
   );
 
   const searchedPosterVenues = useMemo(() => {
-    if (!searchQuery)
```
0xdevalias, line 80: Can searchQuery ever be undefined/similar? The old code guarded against that, the new code doesn’t seem to.

src/hooks/posters.ts
```diff
@@ -71,24 +71,53 @@ export const usePosters = (posterHallId: string) => {
...
-      return filteredPosterVenues.slice(0, displayedPostersCount);
+    const normalizedSearchQuery = searchQuery.trim();
+
+    if (!normalizedSearchQuery) return filteredPosterVenues;
```
0xdevalias, line 91: The old version of this returned a slice of filteredVenues, the new code doesn't. Why is that?

src/hooks/posters.ts
0xdevalias, line 80: Can searchQuery ever be undefined/similar? The old code guarded against that, the new code doesn’t seem to.

src/hooks/posters.ts
0xdevalias: This was raised/shown on the previous PR but has been lost here since recreating as a new PR (which I have questions about in general @mike-lvov , as it complicates things in an I believe unnessecary way, let’s discuss on slack)

 D0BC2A4C-6C2F-4909-B626-2C6137846633
977D09BF-B995-4200-8126-09CF9036E797

mike-lvov:
@0xdevalias you did the review on a bit outdated code

src/hooks/posters.ts
mike-lvov: @0xdevalias no, it's always a string

src/hooks/posters.ts
mike-lvov: It slices it below. It was needed for correct `Show more posters` button implementation

0xdevalias:
> @0xdevalias you did the review on a bit outdated code

@mike-lvov It wasn’t at the time I was doing it. It wasn’t intended as a deep thorough review anyway, more to point out some things I noticed on my phone while in bed.

src/hooks/posters.ts
0xdevalias: Sounds good

src/hooks/posters.ts
yarikoptic: This `!` issue remained a mystery since couldn't trigger and regex doesn't have any `!`. Concern is also IMHO minor since would only affect the malicious (or the one with stuck keyboard) user on the client side.

yarikoptic: <details>
<summary>looking at such a code analysis fiasco</summary> 

image

where we can't get to that line if that variable is `None`
</details>

and the complexity of the regex analysis [codeql query](https://raw.githubusercontent.com/github/codeql/main/javascript/ql/src/Performance/ReDoS.ql) used for this one, I would definitely not get a sleepless night because of this "issue" and just consider it a "false positive" bug in the analysis query

yarikoptic: asked the "source": https://github.com/github/codeql/issues/5964

yarikoptic:
tip of trade: @0xdevalias a nice strategy is to use html trick

```html
<details>
<summary>SUMMARY HERE</summary> 

CONTENT HERE

</details>
```

so looks likes

<details>
<summary>SUMMARY HERE</summary> 

CONTENT HERE

</details>

to make content (especially images which span multiple pages) collapsable.  could be done post-fact via `Edit` ;)

0xdevalias:
> tip of trade: @0xdevalias a nice strategy is to use html trick

@yarikoptic Yup, am aware and use it regularly. That 'tip' isn't particularly easily usable from a mobile device though. Thus the above.

src/hooks/posters.ts
0xdevalias: Thanks for opening the issue about it @yarikoptic :)

src/utils/text.ts
```diff
@@ -0,0 +1,2 @@
+export const breakStringWithQuotesBySpaces = (string: string) =>
+  string.match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g); // source: https://stackoverflow.com/a/16261693/1265472
```
0xdevalias, line 2: See discussion about this regex issue (and the `Code scanning: Inefficient regular expression` error that is raisued for it) in https://github.com/sparkletown/sparkle/pull/1460#discussion_r640977315

src/hooks/posters.ts
0xdevalias: Looking at the 'explain' for the regex on https://regexr.com/ :

```
("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)
```

<details><summary>Expand for screenshot..</summary>

Screen Shot 2021-05-28 at 1 40 15 pm

</details>

Breaking this down:

- `(` `"[^"]*?"` `|` `[^"\s]+` `)+`
  - `"[^"]*?"`: Match `"`, not `"`, and then another `"`; 0 or more times, lazily
  - `|`: or
  - `[^"\s]+`: match not `"` or a space; 1 or more times
- `(?=` `\s*` `|` `\s*$` `)`
  - `(?=`: positive lookahead (matches a group after the main expression without including it in the result)
  - `\s*`: match 0 or more whitespace characters
  - `|`: or
  - `\s*$`: match 0 or more whitespace characters, then the end of the string
 
---
 
Won't look into this any further for now, and wait to see what (if anything) https://github.com/github/codeql/issues/5964 turns up.

0xdevalias: Fwiw, from what I am seeing in that explain, I don't think it's likely to be a false positive directly, though the `!` it mentions is likely incorrect (and possibly should refer to a `"` or a space based on the regex i'm guessing).

Can't quite wrap my 🧠 around it right now though to be able to definitively say that one way or another right now though.

0xdevalias: For 888 bytes, this lib seems to support it:

- https://github.com/jonschlinkert/split-string
  - seemingly not using regex: https://github.com/jonschlinkert/split-string/blob/master/index.js
- https://bundlephobia.com/result?p=split-string@6.1.0

0xdevalias: This is another SO that seems to provide some regex based solutions:

- https://stackoverflow.com/questions/2817646/javascript-split-string-on-space-or-on-quotes-to-array

yarikoptic: I like that split-words - has tests! Nice find.

yarikoptic: Re regex - ideally shouldn't split on - but if required, I guess could be improved. Didn't analyze/try yet.

yarikoptic:
Somewhat unrelated to the excited topic of splitting. If @mike-lvov had a moment: for this solution I did `$or` across all keys, but if split words started with "{key}:" , restricting only to that key (with making to more obscure DB names), would've been awesome. Then I could say eg

"title: blah blah" author:Пушкин anywhere

and restrict searches - that would've been awesome. Unknown keys ignored and passed as is without splitting or just to all. Can express a brief solution in python, but JS is pain for me.

yarikoptic:
<details>
<summary> BTW I am on the phone too, but I cheat: </summary> 

Screenshot_20210528-002309


</details>

0xdevalias:
> RE: https://github.com/sparkletown/sparkle/pull/1443#issuecomment-849644406
> > Closing it in favor of tested #1460 (copy of this branch) + some additions
> 
> RE: https://github.com/sparkletown/sparkle/pull/1443#issuecomment-849675175
> > A better workflow though would have been to join this PR (maintainers are allowed to push directly into this branch) while preserving the history (useful even if later squashed to see what problems were encountered and resolved) and authorship for the work done (not that I care much but kinda nice to have due credit wherever it is due).
> 
> @mike-lvov @yarikoptic I have now cloned both branches, properly rebased @mike-lvov's changes on top of @yarikoptic's original work, and force pushed that branch back to #1460
> 
> This should resolve any of the 'lost contribution history' issues that were raised and discussed above, and given @mike-lvov's new PR references back to this one (and gives credit to @yarikoptic for the bulk of the work), there shouldn't be an issue as far as attribution goes.
> 
> In future though, let's try to avoid the need to do this rework by ensuring we get it right the first time please :)
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1443#issuecomment-850127763_

src/hooks/posters.ts
0xdevalias: > ideally shouldn't split on `-` 

@yarikoptic Do you mean `_` (underscore, not dash)?

If so, I think the solution there is to just use `[a-zA-z0-9]` instead of `\w`

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1460#issuecomment-850119547

> If @mike-lvov had a moment: for this solution I did `$or` across all keys unconditionally.
> But if split words started with "{key}:" , restricting only to that key (with making to more obscure DB names), would've been awesome. Then I could say eg

@yarikoptic I'm going to call that OOS for this PR in the interest of landing it sooner and avoiding scope creep. If you wanted to open a new backlog issue describing the request, we can handle that as a potential future improvement.

src/hooks/posters.ts
```diff
@@ -71,24 +72,55 @@ export const usePosters = (posterHallId: string) => {
   const fuseVenues = useMemo(
     () =>
       new Fuse(filteredPosterVenues, {
```
0xdevalias, line 74:
```suggestion
      // Ref: https://fusejs.io/api/options.html
      new Fuse(filteredPosterVenues, {
```

src/utils/text.ts
0xdevalias, line 2: Adds explicit return type + reformats comment
```suggestion
/**
 * @see https://stackoverflow.com/a/16261693/1265472
 */
export const breakStringWithQuotesBySpaces = (
  string: string
): RegExpMatchArray | null => string.match(/("[^"]*?"|[^"\s]+)+(?=\s*|\s*$)/g);
```

src/hooks/posters.ts
```diff
@@ -71,24 +72,55 @@ export const usePosters = (posterHallId: string) => {
...
+      .search({
+        $and: matchEntries.map((x) => ({
+          $or: [
+            { name: x },
+            { "poster.title": x },
+            { "poster.authorName": x },
+            { "poster.categories": x },
+            // @debt Rewrite the search query to match the $and type
+          ] as Record<string, string>[],
+        })),
```
0xdevalias, line 107: What is the relevance of this `@debt` comment here? It's ambiguous currently, since there already seems to be an `$and` used here?

src/hooks/posters.ts
```diff
@@ -71,24 +72,55 @@ export const usePosters = (posterHallId: string) => {
...
+
+    const matchEntries = breakStringWithQuotesBySpaces(normalizedSearchQuery);
+
+    if (!matchEntries) return filteredPosterVenues;
 
     return fuseVenues
-      .search(searchQuery)
-      .slice(0, displayedPostersCount)
+      .search({
+        $and: matchEntries.map((x) => ({
```
0xdevalias, line 99: @yarikoptic @mike-lvov Instead of hacking our own workaround to match quoted strings together, why not enable/use the built in syntax/mode supported by Fuse.js?

- https://fusejs.io/examples.html#extended-search
  - > White space acts as an AND operator, while a single pipe (`|`) character acts as an OR operator. To escape white space, use double quote ex. `="scheme language"` for exact match.

src/hooks/posters.ts
```diff
@@ -71,24 +72,55 @@ export const usePosters = (posterHallId: string) => {
...
-        keys: ["poster.title", "poster.authorName", "poster.categories"],
+        keys: [
+          "name",
+          "poster.title",
+          "poster.authorName",
+          "poster.categories",
+        ],
+        threshold: 0.2, // 0.1 seems to be exact, default 0.6: brings too distant if anyhow related hits
+        ignoreLocation: true, // default False: True - to search ignoring location of the words.
+        findAllMatches: true,
```
0xdevalias, line 83: [note] See the following for more details on these options + how the scoring works:

- https://fusejs.io/api/options.html
- https://fusejs.io/concepts/scoring-theory.html#scoring-theory

0xdevalias: See also my suggestion elsewhere in this review of using extended search instead of trying to hack our own tokenisation/support of exact matching quoted terms.

- https://fusejs.io/api/options.html#useextendedsearch

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -55,7 +56,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
       </div>
 
       <div className="PosterHall__more-button">
-        {isPostersLoaded && (
+        {hasHiddenPosters && isPostersLoaded && (
```
0xdevalias, line 59: [NB] We should extract this into something like, then use that here:

```typescript
const shouldShowMorePosters = isPostersLoaded && hasHiddenPosters
```

src/hooks/posters.ts
```diff
@@ -71,24 +72,55 @@ export const usePosters = (posterHallId: string) => {
...
-      .slice(0, displayedPostersCount)
+      .search({
+        $and: matchEntries.map((x) => ({
+          $or: [
+            { name: x },
+            { "poster.title": x },
+            { "poster.authorName": x },
+            { "poster.categories": x },
+            // @debt Rewrite the search query to match the $and type
+          ] as Record<string, string>[],
```
0xdevalias, line 106: Why are we explicitly forcing the type here with `as`? Surely there's a better solution?

0xdevalias (approved):
[CRO] @mike-lvov I'm going to preemptively approve this to unblock you, as I don't see any huge issues, but note that I haven't tested it yet, and there are a few notes in here that are probably worth checking/fixing before it gets merged.

src/hooks/posters.ts
0xdevalias: This seems to be hiding an error: 
image

The last line seems to imply that `undefined` is not an allowable value here, yet I believe that's the only reason why the `as` type override is needed here.

0xdevalias: Actually.. that doesn't seem to fully solve it.. 🔍 👀

0xdevalias: This seems to solve it:

```patch
index 9714f48a..853270cf 100644
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -90,21 +90,26 @@ export const usePosters = (posterHallId: string) => {
 
     if (!normalizedSearchQuery) return filteredPosterVenues;
 
-    const matchEntries = breakStringWithQuotesBySpaces(normalizedSearchQuery);
+    const matchEntries: string[] =
+      breakStringWithQuotesBySpaces(normalizedSearchQuery) ?? [];
 
     if (!matchEntries) return filteredPosterVenues;
 
+    // @debt Rewrite the search query to match the $and type
     return fuseVenues
       .search({
-        $and: matchEntries.map((x) => ({
-          $or: [
+        $and: matchEntries.map((x: string) => {
+          const orFields: Fuse.Expression[] = [
             { name: x },
             { "poster.title": x },
             { "poster.authorName": x },
             { "poster.categories": x },
-            // @debt Rewrite the search query to match the $and type
-          ] as Record<string, string>[],
-        })),
+          ];
+
+          return {
+            $or: orFields,
+          };
+        }),
       })
       .map((fuseSearchItem) => fuseSearchItem.item);
   }, [searchQuery, fuseVenues, filteredPosterVenues]);
```

src/hooks/posters.ts
0xdevalias: Tried playing around with this a little (with extended search enabled) using the search term `eeg`.

---

Just using:

```typescript
.search(normalizedSearchQuery)
```

---

Just using a top level `$or` for the fields, with no 'custom tokenisation' fails when trying to use a search term and the authors name to narrow it down (eg. `eeg rux` fails to find the poster I want)

Presumably because of it's 'short circuit':

- https://fusejs.io/api/query.html#or
  - > The `$or` operator performs a logical OR operation on an array expressions and selects the entries that satisfy at least one of the expressions. **The `$or` operator uses short-circuit evaluation (i.e, if the first expression evaluates to true, Fuse.js will not evaluate the remaining expressions).**

```typescript
.search({
  $or: [
    { name: normalizedSearchQuery },
    { "poster.title": normalizedSearchQuery },
    { "poster.authorName": normalizedSearchQuery },
    { "poster.categories": normalizedSearchQuery },
  ],
})
```

---

I think I played around with some other variations of things, basically attempting to find a method of using the extended search features without needing to tokenise the search string ourselves.. but.. basically didn't manage to figure anything useful/it all seemed to work less good than the 'custom tokenisation' version (even though it feels weirdly hacky that we would need to do that for some reason.. maybe just because the lib isn't so much designed for passing user-facing search input directly through)

src/hooks/posters.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1460/commits/56686771a3a97eb5d0a537bb191dfad4754d362f

src/components/templates/PosterHall/PosterHall.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1460/commits/8051f616d516d57fde53964d6b45ad92e447b09f

src/hooks/posters.ts
0xdevalias: The conclusion on that other thread is likely that tokenisation is actually the best solution we're going to get, and that it probably doesn't make sense to add the extended search here.

Based on that, going to mark this as resolved.

src/hooks/posters.ts
0xdevalias: I believe this related to what I called out in https://github.com/sparkletown/sparkle/pull/1460#discussion_r641295362

Which was fixed in 5668677

src/utils/text.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1460/commits/dd079984f7ece5b6a3ab9c8a974b1ab0d92d8c5a

src/hooks/posters.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1460/commits/9d293acfb44e50f362e410eec6042786a34aa6fd

src/hooks/posters.ts
0xdevalias: This seems to have been fixed. Not sure what commit it was in. Probably one of @mike-lvov 's

Resolving.

src/hooks/posters.ts
0xdevalias: In the interest of getting this merged, I'm going to call it OOS to muse on/change the regex/splitting method anymore here.

I've added a `@debt` comment to it in https://github.com/sparkletown/sparkle/pull/1460/commits/11cca352507b27f8d17583c7d0453347c590b920 referencing https://github.com/github/codeql/issues/5964 so that we have the relevant details captured there in future if needed.

0xdevalias:
Copying this out of the comment thread for visibility purposes.

> In the interest of getting this merged, I'm going to call it OOS to muse on/change the regex/splitting method anymore here.
> 
> I've added a `@debt` comment to it in https://github.com/sparkletown/sparkle/pull/1460/commits/11cca352507b27f8d17583c7d0453347c590b920 referencing https://github.com/github/codeql/issues/5964 so that we have the relevant details captured there in future if needed.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1460#discussion_r641352179_

0xdevalias:
Thanks for all of your work on this @yarikoptic! (and @mike-lvov for your additional cleanup/polish) 🎉

yarikoptic:
Woohoo -- thank you @mike-lvov and @0xdevalias !!! de-@dept (it was a false positive) PR #1470

PR #1464, opened 2021-05-27
add sorting + proper 'show more' button in ScreeningRoom

mike-lvov:
Adds sorting + proper show more button logic in screening room


yurii-lubynets (approved):

src/components/templates/ScreeningRoom/useScreeningRoom.ts
```diff
@@ -119,24 +120,21 @@ export const useScreeningRoom = (screeningRoomVenueId: string) => {
...
-  }, [
-    searchQuery,
-    fuseVideos,
-    filteredVideosBySubCategory,
-    displayedVideosAmount,
-  ]);
+  }, [searchQuery, fuseVideos, filteredVideosBySubCategory]);
+
+  const displayedVideos = searchedVideos
+    .sort((a, b) => alphabeticalOrderSort(a.title, b.title))
```
sunny-viktoryia, line 131: can we use `a.title.localeCompare(b.title)` here?

see: https://stackoverflow.com/a/2167619

mike-lvov: @sunny-viktoryia good catch!

Fixed in https://github.com/sparkletown/sparkle/pull/1464/commits/b406c60efbda905b3a7f0fca95cb9633c4496fab

src/components/templates/ScreeningRoom/useScreeningRoom.ts
```diff
@@ -70,7 +70,7 @@ export const useScreeningRoom = (screeningRoomVenueId: string) => {
   const increaseDisplayedVideosAmount = useCallback(() => {
     setDisplayedVideosAmount(
       (prevPostersNumber) =>
-        prevPostersNumber + DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT
+        prevPostersNumber + DEFAULT_DISPLAYED_VIDEO_PREVIEW_COUNT
     );
```
0xdevalias, line 74: [nit] change `prevPostersNumber` to `prevVideosNumber` or similar.

0xdevalias (approved):
One tiny nit change, then LGTM!

It might also be good to add some `// @debt` comments pointing out that this code seems to be almost a direct duplication from the `PosterHall`, and to handle cleaning that up/de-duplicating it in future.

src/settings.ts
```diff
@@ -585,6 +585,7 @@ export const POSTERPAGE_MAX_VIDEO_PARTICIPANTS = 10;
 export const SEARCH_DEBOUNCE_TIME = 200; // ms
 
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 48;
```
denisdimitrov, line 587: Please remove the old `const` if it's not being used anymore.

src/components/templates/ScreeningRoom/useScreeningRoom.ts
```diff
@@ -119,24 +119,21 @@ export const useScreeningRoom = (screeningRoomVenueId: string) => {
...
-    displayedVideosAmount,
-  ]);
+  }, [searchQuery, fuseVideos, filteredVideosBySubCategory]);
+
+  const displayedVideos = searchedVideos
+    .sort((a, b) => a.title.localeCompare(b.title))
+    .slice(0, displayedVideosAmount);
 
-  const hasHiddenVideos = searchedVideos.length > displayedVideosAmount;
+  const hasHiddenVideos = searchedVideos.length > displayedVideos.length;
```
sunny-viktoryia, line 133: and can you explain why was that changed?
I guess `displayedVideosAmount` should have the same value. 🤔

src/settings.ts
0xdevalias: I expect this is still being used in the `PosterHall`

src/components/templates/ScreeningRoom/useScreeningRoom.ts
0xdevalias: Since `displayedVideos` created from a `.slice(0, displayedVideosAmount)` it should at most have an array length of `displayedVideosAmount`; but I expect it could possibly have less elements than that if `displayedVideosAmount` is set to a value higher than the total length of `searchedVideos`

mike-lvov: @sunny-viktoryia that logic was broken. Now it's fixed

src/settings.ts
mike-lvov: @denisdimitrov that const is used in PosterHall

PR #1473, opened 2021-05-28
add loading when editing venue table

denisdimitrov:
- Fixes a bug where only the first table can be edited.
- Adds loading flag that disables the UI elements when the request is pending.


functions/venue.js
```diff
@@ -730,7 +730,7 @@ exports.updateTables = functions.https.onCall((data, context) => {
       (table) => table.reference === data.updatedTable.reference
     );
 
-    if (currentTableIndex > 0) {
+    if (currentTableIndex < 0) {
```
mike-lvov, line 733: What's the intent of this change?

src/components/molecules/TableHeader/components/EditTableTitleModal/EditTableTitleModal.tsx
```diff
@@ -41,6 +42,7 @@ export const EditTableTitleModal: React.FC<EditTableTitleModalProps> = ({
   onHide,
 }) => {
   const venueId = useVenueId();
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
mike-lvov, line 45: [NB] Can this be rewritten using `showHide` hook? Looks like it's a good case for using it

mike-lvov (changes requested):

functions/venue.js
denisdimitrov: It was reversed by mistake. Fixes this:
> Fixes a bug where only the first table can be edited.

src/components/molecules/TableHeader/components/EditTableTitleModal/EditTableTitleModal.tsx
denisdimitrov: It can. However, is it okey to do so. When the name implies that it's show/hide but in my case is loading or not

denisdimitrov: I suppose, I can create a hook that's only used for loading.

mike-lvov: It's handled like this

image

denisdimitrov: But the benefit is almost non existent. 
Instead of:
```tsx
const [isLoading, setIsLoading] = useState<boolean>(false);
```
It will be:
```tsx
const { isLoading, setIsLoading } = useLoading();
```

It's still going to be one line code.

denisdimitrov: Actually useAsyncFn returns `loading` I will use that instead.

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1473/commits/5ca74272522dfcd2eae36fa854d48158194d67e0

denisdimitrov: I think the `useAsyncFn`'s out of the box loading is better approach. If you disagree I will use `useShowHide`

mike-lvov (approved):
LGTM

src/components/molecules/TableHeader/components/EditTableTitleModal/EditTableTitleModal.tsx
0xdevalias: Agreed RE: useAsyncFn

0xdevalias: With useShowHide, we need a better name for it that allows it to be used more generically tbh.

PR #1474, opened 2021-05-29
Fixes object freezing err

mike-lvov:
Fixes an error, I still don't fully understand why happens, but it was described in the stack overflow thread, linked in the comment


0xdevalias:
> I still don't fully understand why happens, but it was described in the stack overflow thread, linked in the comment

@mike-lvov skimming the fix here, off the top of my head, the .slice() creates a new shallow clone of the array, which is needed because if I remember correctly .sort() mutates the array it’s called on.

PR #1481, opened 2021-05-31
Fix profile questions page

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/722

Before:  
image

After:
Screen Shot 2021-05-31 at 12 00 06 PM


src/pages/Account/Account.scss
```diff
@@ -55,11 +55,21 @@ $page-max-width: 1240px;
   background-color: $dark;
   flex-direction: column;
   flex: 1;
+  padding-bottom: 30px;
```
yurii-lubynets, line 58: please use vars from `src/scss/constants.scss`

yurii-lubynets (approved):
LGTM, please cover the comment

src/pages/Account/Account.scss
```diff
@@ -107,7 +117,9 @@ $page-max-width: 1240px;
 
   &.sparkle {
     margin: 0 auto;
-    background-image: url(/assets/images/sparkle-logo.png);
```
sunny-viktoryia, line 110: do we still need that image? if it's not used anywhere else can we remove it as well?
and if it's removed we could rename `Sparkle_logo_white_yellow.png` to `sparkle-logo.png`.

malaikaaiyar: `scripts/bootstrap-new-environments.ts` seems to use it, so I won't delete it

src/pages/Account/Account.scss
malaikaaiyar: fixed it!

src/pages/Account/Account.scss
0xdevalias: FYI: That script is only using it as a default image. If we have a different/better image we can use than can update that script to use it.

0xdevalias: Also, why did we remove this image from here in general out of curiosity?

0xdevalias: Since the intent of this change seems to be to replace the old logo with the new, it makes sense to delete the old outdated logo and name the new correct logo as @sunny-viktoryia suggested above.

malaikaaiyar: Ok- will make a new pr for this

PR #1468, opened 2021-05-28
refactor openUrl/openRoomUrl/enterVenue to allow custom openers + use React Router for navigation in PosterPreview / NavBar's back button

0xdevalias:
The changes in this PR basically pave the way to us being able to use React Router for relative link navigation, rather than having to entirely load the application every time someone changes spaces (or as @yarikoptic describes it: the "blue screen of sparkle")

- https://reactrouter.com/web/api/history
  - > `push(path, [state])` - (`function`) Pushes a new entry onto the history stack

It adds this support in `openUrl`, `openRoomUrl`, and `enterVenue` within `utils/url`.

It uses this in only 2 places currently (we can add more in future followup PRs):

- the `backToParentVenue` button in the `NavBar` component
- within the `PosterPreview` component when accessing a `PosterPage`

I want to keep the scope of this PR fairly small for now, as while it does increase the perceived performance of the app quite a bit when navigating between spaces; there are cases where we actually benefit from the "forced to fully reload the page" navigation, as it can force aspects of our code with potential memory leaks/etc to get reset/cleaned up. I'm not certain if/where/how much of an impact that will be, so I want to avoid updating this across the entire app until we have a better idea around that.

---

partially fixes https://github.com/sparkletown/sparkle/issues/1461

Will likely be useful for https://github.com/sparkletown/internal-sparkle-issues/issues/752#issuecomment-850333759

PR #1482, opened 2021-05-31
fix back button navigation in partymap template

denisdimitrov:
When `useConnectCurrentVenue()` is called inside the venue template, the venue doesn't update in the store. Maybe because it's already called in the top layer, inside ```VenuePage.tsx```?

Removing it fixes the issue, and adding it to another venue template reproduces the bug.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/774


yurii-lubynets (approved):

0xdevalias:
> When `useConnectCurrentVenue()` is called inside the venue template, the venue doesn't update in the store. Maybe because it's already called in the top layer, inside `VenuePage.tsx`?
> 
> Removing it fixes the issue, and adding it to another venue template reproduces the bug.

`useConnectCurrentVenue` calls `useVenueId` which tries to read the `venueId` from the URL params using `useParams`. If `venueId` is defined, it will the connect the queries using `useFirestoreConnect`.

@denisdimitrov It would be good if you could `console.log` the `venueId` value in `useConnectCurrentVenue` in one of the 'failing case' situations so we know what it's set to. I expect it's likely to be `undefined` or something else similarly 'falsy', or at least starts out that way.

One of the limitations of our `useConnectX` pattern of hooks is that since Redux is a global datastore, when we use the same name for `storeAs`, it will write to the same area. While `useFirestoreConnect` and/or the underlying firestore libraries are meant to theoretically de-duplicate queries (assuming they are exactly the same), I expect the reality is that that isn't always the case.

Another thing that confuses me about this is that it seemingly only became an issue very recently, which implies it probably was introduced by my https://github.com/sparkletown/sparkle/pull/1468 PR; though based on what's been said/shown so far, I'm not sure why that would be the case. Any better thoughts/theories on that @denisdimitrov?

0xdevalias:
> @denisdimitrov It would be good if you could `console.log` the `venueId` value in `useConnectCurrentVenue` in one of the 'failing case' situations so we know what it's set to. I expect it's likely to be `undefined` or something else similarly 'falsy', or at least starts out that way.

Just tried this locally with my env mapped through to `env/ohbm`, looks like the `venueId` is always set throughout, so even more confused as to why the issue would be expressing like this.

0xdevalias:
> From what I observed just now, it feels like it might be a race condition or similar.

I played around with this idea for a little bit, but still don't really have any better theories/explanations as to why the changes here fix it. Even when I dig into the data stored in Redux, it doesn't look like changing what is changed in this PR should have any impact.

denisdimitrov:
> I played around with this idea for a little bit, but still don't really have any better theories/explanations as to why the changes here fix it. Even when I dig into the data stored in Redux, it doesn't look like changing what is changed in this PR should have any impact.

Yeah, me too. I can't really figure out, why this is not updating. The url and param changes but the venue itself doesn't in the redux. It's almost like hook doesn't detect the change. 

I tried adding the same hook to another template (Artpiece) and the bug was reproduced. It has something to do with that deprecated hook but I still can't figure it out.

@0xdevalias Yes. it was introduced in: https://github.com/sparkletown/sparkle/pull/1468/files because before that it was using href and it was reloading the whole page and that's why it was re-rendering the whole page. Now using `.push()` it changes the url but it doesn't detect the param change, at least the hook doesn't. I'm still trying to figure this out. I don't feel comfortable merging this yet.

denisdimitrov:
Another potential workaround is, removing this line: https://github.com/sparkletown/sparkle/pull/1468/files#diff-d2dbbd9743869736d76cf73512c615856022b7a3ee80ec9b63bed997c40f7e37R104 but that just avoids the problem by not using router and forcing the browser to reload the whole page by changing it's `href`

0xdevalias:
> Another potential workaround is, removing this line: https://github.com/sparkletown/sparkle/pull/1468/files#diff-d2dbbd9743869736d76cf73512c615856022b7a3ee80ec9b63bed997c40f7e37R104 but that just avoids the problem by not using router and forcing the browser to reload the whole page by changing it's `href`

@denisdimitrov Yeah, we definitely want to keep that. I think even if we can’t understand the ‘why’ of this fix in the PR here, so long as we’re sure it’s not going to break anything, it’s probably fine to remove. Particularly if as you said this hook is already connected ‘higher up’ anyway.

I wonder how many places this deprecated connect/selector is used currently, maybe we can end up getting rid of it entirely?

denisdimitrov:
@0xdevalias Just checked and the `useConnectCurrentVenue` hook is used in a couple of places and all of them are different routes and don't have the hook called in a higher place. Except for `PlayaRouter.tsx` which is deprecated. All of the other venue templates don't have this hook, so we should be fine.

denisdimitrov:
I also tried a couple of things but the bug still exists.

1. Update `react-router-dom`
2. Use https://github.com/supasate/connected-react-router. Some people say that this is a potential fix but didn't work.
3. Use `.replace()` instead of `.push()`
4. Pass the whole URL to `.push()`
5. Get the venue directly from `useConnectCurrentVenue` instead of `useSelector(currentVenueSelector)`
6. I tried adding `createBrowserHistory()` to our `Router`
7. Replacing the `BrowserRouter` with another one
8. Making the router, `exact`, `sensitive` and `strict`

0xdevalias:
> @0xdevalias Just checked and the `useConnectCurrentVenue` hook is used in a couple of places and all of them are different routes and don't have the hook called in a higher place. Except for `PlayaRouter.tsx` which is deprecated. All of the other venue templates don't have this hook, so we should be fine.

@denisdimitrov did you also look for usages of the selector? One of the problems with this hook/pattern/how we use it is that searching for usages of the hook itself are insufficient, as we need to find anywhere that uses the selector, and then ensure that it has been ‘properly connected’ somewhere higher up.

0xdevalias:
> I also tried a couple of things but the bug still exists.

@denisdimitrov thanks for trying/listing what you tried. I guess if we can’t figure it out + we’re sure that the initial change in this PR won’t break anything then we can just merge that and leave the ‘why’ unknown.

denisdimitrov:
@0xdevalias The selector is being used in a couple of places but it doesn't seem to be problematic. Looks like the hooks is causing the problem, this code specifically, when called twice: 
```tsx
  useFirestoreConnect(
    venueId
      ? {
          collection: "venues",
          doc: venueId,
          storeAs: "currentVenue",
        }
      : undefined
  );
```

The problem is when the hook is called 2 or more times in the parent/children structure.

denisdimitrov:
> did you also look for usages of the selector?

Yes, everything seems to be connected properly. At the top level. `VenuePage.tsx`

0xdevalias:
> The selector is being used in a couple of places but it doesn't seem to be problematic.

@denisdimitrov Yup. The reason I said to check it was more to ensure it was still connected properly, as you called out seems to be fine in https://github.com/sparkletown/sparkle/pull/1482#issuecomment-852991671

0xdevalias:
@denisdimitrov When merging in future, please ensure the name of the PR and the name on the merge commit are aligned.
image

image

PR #1056, opened 2020-12-23
Add new admin panel v2.0

IvanBuljovcic:
Adds features to the admin panel to update partymap template

PR #1485, opened 2021-06-01
refactor updateVenue / updateVenue_v2 in functions/venue to DRY duplicate code

0xdevalias:
- Admin (v1) Venue entrance 'Add entrance step' was broken, as the backend `updateVenue` function didn't handle it (though `updateVenue_v2` did)
  - There's a lot of duplication/room for error between `updateVenue` / `updateVenue_v2` that should be DRY'd up to reduce cognitive burden, engineer effort, and the possibility of issues like this arising in future.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/777
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/714
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/842


functions/venue.js
```diff
@@ -462,20 +462,26 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
...
 
+  // @debt this is exactly the same as in updateVenue_v2
   if (!doc || !doc.exists) {
     throw new HttpsError("not-found", `Venue ${venueId} not found`);
   }
 
+  // @debt this is exactly the same as in updateVenue_v2
   const updated = doc.data();
 
+  // @debt this is missing from updateVenue_v2, why is that? Do we need it there/here?
```
denisdimitrov, line 484: [NB] After the new admin v3 design the venue can be created without these fields and that's why they are missing from that function.

functions/venue.js
```diff
@@ -462,20 +462,26 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
   }
 });
 
+// @debt this is almost a line for line duplicate of exports.updateVenue_v2, we should de-duplicate/DRY these up
 exports.updateVenue = functions.https.onCall(async (data, context) => {
   const venueId = data.id || getVenueId(data.name);
   checkAuth(context);
 
+  // @debt updateVenue_v2 uses checkUserIsAdminOrOwner rather than checkUserIsOwner. Should these be the same? Which is correct?
   await checkUserIsOwner(venueId, context.auth.token.user_id);
```
denisdimitrov, line 471: `checkUserIsAdminOrOwner` is the correct one, because the user should be an owner of the admin venue + he should be in the `admins` collection

src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -198,20 +198,22 @@ export const roomEditSchema = Yup.object().shape<RoomSchemaShape>({
...
+// @debt I'm pretty sure every one of these .from that have the same fromKey / toKey are redundant noops and should be removed
 export const editVenueCastSchema = Yup.object()
   .shape<Partial<VenueInput>>({})
   // possible locations for the subtitle
   .from("subtitle", "subtitle")
   .from("config.landingPageConfig.subtitle", "subtitle")
 
   .from("config.landingPageConfig.description", "description")
-  .from("profile_questions", "profileQuestions")
+  .from("profile_questions", "profile_questions")
```
denisdimitrov, line 216: I'm okey with this for now, but we should get rid of the underscore syntax because the rest of our fields are using camelCase and it's confusing.

denisdimitrov (approved):
LGTM!

functions/venue.js
0xdevalias: That function is super misleading, as it doesn't do anything extra:

```javascript
const checkUserIsAdminOrOwner = async (venueId, uid) => {
  try {
    return await checkUserIsOwner(venueId, uid);
  } catch (e) {
    return e.toString();
  }
};
```

src/pages/Admin/Details/ValidationSchema.ts
0xdevalias: Eventually sure, but the entire app uses this in the `_` form, and nowhere in the app references it in the `profileQuestions` form. So it will basically need to be a refactor of the entire app to do so.

0xdevalias:
RE: https://github.com/sparkletown/sparkle/pull/1485#pullrequestreview-691630654

@denisdimitrov apparently I can't dismiss reviews in draft mode, but this PR is WIP and absolutely not ready to be merged.

0xdevalias:
@denisdimitrov This should be good to go now I think. I gave it a tiny quick clickthrough locally and it seems to fix the 2 things intended. Calling further refinement/de-duplication of the 'not exactly the same' aspects OOS of this PR to get the main fixes in.

Please feel free to merge it once approved.

functions/venue.js
denisdimitrov: In this case we can remove it later on.

denisdimitrov:
> @denisdimitrov apparently I can't dismiss reviews in draft mode, but this PR is WIP and absolutely not ready to be merged.

I saw the state of the PR but I was asked for a review from the product team because the profile questions weren't working properly and it seemed like this fixes the issue.

functions/venue.js
```diff
@@ -462,20 +548,29 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
...
+  // @debt this is exactly the same as in updateVenue_v2
   if (!doc || !doc.exists) {
     throw new HttpsError("not-found", `Venue ${venueId} not found`);
   }
 
+  // @debt this is exactly the same as in updateVenue_v2
   const updated = doc.data();
 
+  // note: This function will mutate the passed in updated object
+  createBaseUpdateVenueData(data, updated);
```
denisdimitrov, line 571: IMO this should be returning the updated object instead of mutating it. Also, the second object is called `updated` but it's not actually updated yet.

0xdevalias: Yeah, I agree, but I couldn’t wrap my brain around it last night, and this is basically the equivalent pattern of how it was already implemented, just extracting the common parts into a shared function.

I didn’t want to have to make a deep clone of the existing object, and I didn’t want to have to build up the deeply nested structures and deal with potentially undefined fields and such.

I absolutely expect we’ll iterate and improve on this more in future, but I kind of want to wait till we have typescript in place on the backend to reduce the cognitive burden of trying to do so.

Given this is more or less equivalent to the existing pattern, and the importance of some of the fixes here, I’d like to get it out if there’s no ‘major issues/regressions’ with it.

0xdevalias: If you have some quick/efficient solutions to avoiding the mutation I’m happy to hear them. But otherwise I think we should call that a future iterations improvement.

functions/venue.js
```diff
@@ -269,6 +270,91 @@ const createVenueData_v2 = (data, context) => ({
...
+      updated.host = {};
+    }
+    updated.host.icon = data.logoImageUrl;
+  }
+
+  if (data.profile_questions) {
+    updated.profile_questions = data.profile_questions;
+  }
+
+  if (data.entrance) {
```
sunny-viktoryia, line 300: the `updateVenue` works now. But `createVenue` still misses `.entrance` data. We need to edit that function too to be able to add `entrance` data while creating a venue.

\+ to be able to update `entrance` data in Admin UI while editing venue, you should add `editing` property to the `EntranceInput` component

```tsx
        <EntranceInput
          register={register}
          fieldName="entrance"
          editing={state.detailsPage?.venue.entrance}
        />

```

sunny-viktoryia (changes requested):
looks like it doesn't fully work in the Admin UI (tried to run it locally). Suggested changes should help fix that.

P.S. thanks for the `@debt` comments about the difference between adminV1 and adminV2/3

src/pages/Admin/Venue/DetailsValidationSchema.ts
```diff
@@ -197,14 +197,15 @@ export const validationSchema = Yup.object()
...
+// @debt I'm pretty sure every one of these .from that have the same fromKey / toKey are redundant noops and should be removed
 export const editVenueCastSchema = Yup.object()
   .shape<Partial<VenueInput>>({})
   // possible locations for the subtitle
   .from("subtitle", "subtitle")
   .from("config.landingPageConfig.subtitle", "subtitle")
 
   .from("config.landingPageConfig.description", "description")
-  .from("profile_questions", "profileQuestions")
+  .from("profile_questions", "profile_questions")
```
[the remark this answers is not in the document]
0xdevalias: Ah, nice catch! I noticed it wasn't popping up the view of previous values, but I thought that was just our terrible UI/UX flow in admin v1. Thanks for the solution!

functions/venue.js
0xdevalias: > couldn't check because of cors issues locally

Were you running the function emulator? That's usually the problem.

0xdevalias: > But `createVenue` still misses .entrance data. We need to edit that function too to be able to add entrance data while creating a venue.

@sunny-viktoryia Nice catch. I didn't try creating a new venue.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1485/commits/f395d82ff53b0de5ce1c2ce5d1a151bcd16c6729 + https://github.com/sparkletown/sparkle/pull/1485/commits/f2b495fb6dada938f332df513adaaaa39dcc3830

and a bonus one for `code_of_conduct_questions` https://github.com/sparkletown/sparkle/pull/1485/commits/fc267a813da2bc1c912d640d6062d74798059100

src/pages/Admin/Venue/DetailsValidationSchema.ts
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1485/commits/f395d82ff53b0de5ce1c2ce5d1a151bcd16c6729

functions/venue.js
goran-peoski-work: Maybe just rename them for now, anything in the form of source/destination, old/new  and maybe mark the intent for the future with @debt

functions/venue.js
goran-peoski-work: This looks more like UI utility function that is supposed to print a message. Probably solved inline in the TSX that calls it. 
Otherwise, suggested type replacement: `boolean|Error` or just an enum having unknown/admin/owner/regular...

functions/venue.js
sunny-viktoryia: super awesome! 🎉

sunny-viktoryia: @0xdevalias  here is my CORS issue
image

not sure why it does not work for me.
I have running:
```
npm run start
npm run firebase:emulate-functions
```

sunny-viktoryia (approved):
Wonderful! 🙃 

Looks like we're going to have a less buggy Admin V1!!! 🎉

functions/venue.js
0xdevalias: Hrmm.. are both your .env and firebase use set to the same environment?

functions/venue.js
0xdevalias: This is in the backend functions, which are only raw JavaScript at the moment, no types. I'll dig back into the git history sometime in the future to figure where/why it was implemented, but till then, `@debt` suffices.

0xdevalias: Ended up looking into it now instead..

Looking at the git blame it was introduced in https://github.com/sparkletown/sparkle/pull/1056, at the start of our admin v2 tech debt:

- https://github.com/sparkletown/sparkle/blame/staging/functions/venue.js#L152-L158

So i'm just going to clean it up/remove it in this PR I think.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1485/commits/5df8975f5ddadec52b9f43a1089de3230a39246c

functions/venue.js
0xdevalias: `@debt` was added in https://github.com/sparkletown/sparkle/pull/1485/commits/682627b2ba302bdf6cf5c58d0c4a8fc71fca84a4

Nuance of naming and 'not yet updated' is pre-existing from the code that was here before, so calling OOS.

functions/venue.js
sunny-viktoryia: ohh yes, you were right. I used `sparkle-github` for functions for some reason. 🙈 
Thank you. 🙃 It works now.

PR #1486, opened 2021-06-01
fix to return chat to original mode after polling post

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/779


src/components/molecules/PollBox/PollBox.tsx
```diff
@@ -36,7 +38,7 @@ export const PollBox: React.FC<PollBoxProps> = ({ createPoll }) => {
       topic,
       questions: questions.map(({ name }, id) => ({ name, id })),
     });
-    reset();
+    unselectOption();
   });
```
0xdevalias, line 42: Instead of passing in both `createPoll` and `unselectOption` to this component to then create `onPollSubmit` here, we should just be passing in a single prop that already combines them as a single handler.

This is an 'implementation detail' of the calling component (`src/components/molecules/Chatbox/Chatbox.tsx`) that is 'leaking' down into this component currently. This component shouldn't need to 'know/think about/care about' it. It should just know that it needs to call a single handler function when the form is submitted.

0xdevalias (changes requested):
Comments/changes.

yurii-lubynets:
@sparkletown/engineering please check this tiny request, hope to merge it asap

src/components/molecules/Chatbox/Chatbox.tsx
```diff
@@ -55,6 +55,14 @@ export const Chatbox: React.FC<ChatboxProps> = ({
     setActiveOption(undefined);
   }, []);
 
+  const onPollSubmit = useCallback(
+    (data) => {
+      createPoll(data);
+      unselectOption();
+    },
+    [createPoll, unselectOption]
+  );
```
0xdevalias, line 64: [OOS] Calling this OOS as it's a bit more involved than the specific fix intended here, but `createPoll` should really be returning it's `Promise`, and then we would be able to use `createPoll.then(() => unselectOption())` here or similar. Then it might make sense to use `useAsync` here with `onPollSubmit` so that we can easily capture/surface the loading/error/etc states from it.

0xdevalias: Actually, it would be nice to add `@debt` comments as part of this PR calling this out though, as a future reminder/etc.

src/components/molecules/PollBox/PollBox.tsx
```diff
@@ -21,23 +21,20 @@ const DEFAULT_VALUES = {
...
   const topic = watch("topic");
 
-  const onPollSubmit = handleSubmit(({ topic, questions }) => {
-    createPoll({
+  const handlePollSubmit = handleSubmit(({ topic, questions }) =>
+    onPollSubmit({
       topic,
       questions: questions.map(({ name }, id) => ({ name, id })),
-    });
-    reset();
```
0xdevalias, line 39: [NB] Is this `reset()` definitely no longer needed here?

0xdevalias: It looks like this whole `PollBox` component will be unrendered when `activeOption` is no longer `ChatOptionType.poll`, so I expect that will 'clear the old data' appropriately. 👌🏻

0xdevalias (approved):
A couple of minor things that would be nice to note as part of this PR, but then LGTM!

src/components/molecules/Chatbox/Chatbox.tsx
yurii-lubynets: fixed in 7126439c4407614d55434c09a72581140e6c94b8

0xdevalias:
@yurii-lubynets this is approved so feel free to merge it

PR #1489, opened 2021-06-02
[scripts] add new script for firestore auth user deletion

denisdimitrov:
New script which checks if the firebase `auth` user exists in the `users` collection. if he doesn't the user is then deleted. Anyone who's account was not created from the sparkle registration page will be deleted!

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/776

### DIFF
--- a/scripts/remove-unregistered-users.ts
+++ b/scripts/remove-unregistered-users.ts
@@ -1,0 +1,125 @@
+#!/usr/bin/env node -r esm -r ts-node/register
+
+import { resolve } from "path";
+
+import admin from "firebase-admin";
+
+import {
+  checkFileExists,
+  initFirebaseAdminApp,
+  makeScriptUsage,
+  parseCredentialFile,
+} from "./lib/helpers";
+
+const usage = makeScriptUsage({
+  description: `Bulk remove of user accounts that exist in the auth but are not in the firestore users collection`,
+  usageParams: `CREDENTIAL_PATH`,
+  exampleParams: `fooAccountKey.json`,
+});
+
+const [credentialPath] = process.argv.slice(2);
+if (!credentialPath) {
+  usage();
+}
+
+if (!checkFileExists(credentialPath)) {
+  console.error("Credential file path does not exists:", credentialPath);
+  process.exit(1);
+}
+
+const { project_id: projectId } = parseCredentialFile(credentialPath);
+
+if (!projectId) {
+  console.error("Credential file has no project_id:", credentialPath);
+  process.exit(1);
+}
+
+initFirebaseAdminApp(projectId, {
+  credentialPath: credentialPath
+    ? resolve(__dirname, credentialPath)
+    : undefined,
+});
+
+// 1000 is the maximum number of users firestore can fetch. check .listUsers()'s documentation for more information.
+// It is also 1000 by default but I decided to set it explicitly to avoid confusion.
+const MAX_USERS = 1000;
+
+const getAuthUsers = async (maxUsers: number, pageToken?: string) =>
+  await admin.auth().listUsers(maxUsers, pageToken);
+
+(async () => {
+  const totalAuthUsers: admin.auth.UserRecord[] = [];
+
+  console.log("Fetching firestore auth users...");
+  const { users: authUsers, pageToken } = await getAuthUsers(MAX_USERS);
+  totalAuthUsers.push(...authUsers);
+
+  // @debt This is going to make another request with offset if there are the maximum users returned by the previous request. This should be using recursion.
+  if (authUsers.length >= MAX_USERS) {
+    console.log(
+      `Auth users are more than ${MAX_USERS}. Fetching additional batch of firestore auth users...`
+    );
+    const { users: additionalAuthUsers } = await getAuthUsers(
+      MAX_USERS,
+      pageToken
+    );
+    totalAuthUsers.push(...additionalAuthUsers);
+  }
+
+  const userDocIds: string[] = await admin
+    .firestore()
+    .collection("users")
+    .get()
+    .then((result) => {
+      console.log("Firestore users successfully fetched.");
+      return result.docs.map((doc) => doc.id);
+    })
+    .catch((err) => {
+      console.error(
+        "Failed when trying to fetch the users collection, skipping...",
+        err
+      );
+      return [];
+    });
+
+  const unregisteredUsers: string[] = [];
+
+  totalAuthUsers.forEach((user) => {
+    if (!userDocIds.includes(user.uid)) {
+      unregisteredUsers.push(user.uid);
+    }
+  });
+
+  const numberOfUnregisteredUsers = unregisteredUsers.length;
+
+  if (!numberOfUnregisteredUsers) {
+    console.error(
+      "There are no unregistered users in the database, end of script."
+    );
+    return;
+  }
+
+  console.error(`Deleting ${numberOfUnregisteredUsers} users...`);
+
+  await admin
+    .auth()
+    .deleteUsers(unregisteredUsers)
+    .then(() => {
+      console.error(
+        `${numberOfUnregisteredUsers} users have been successfuly deleted.`
+      );
+    })
+    .catch((err) => {
+      console.error(
+        "Failed when trying to delete the unregistered users, end of script.",
+        err
+      );
+    });
+})()
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
+  })
+  .finally(() => {
+    process.exit(0);
+  });
--- a/scripts/remove-unfinished-profiles.ts
+++ b/scripts/remove-unfinished-profiles.ts
@@ -12,9 +12,10 @@ import {
 } from "./lib/helpers";
 
 const usage = makeScriptUsage({
-  description: `Bulk remove of user accounts that exist in the auth but are not in the firestore users collection`,
-  usageParams: `CREDENTIAL_PATH`,
-  exampleParams: `fooAccountKey.json`,
+  description:
+    "Bulk remove of user accounts that exist in the auth but are not in the firestore users collection",
+  usageParams: "CREDENTIAL_PATH",
+  exampleParams: "fooAccountKey.json",
 });
 
 const [credentialPath] = process.argv.slice(2);
@@ -40,6 +41,8 @@ initFirebaseAdminApp(projectId, {
     : undefined,
 });
 
+const usersCollectionRef = admin.firestore().collection("users");
+
 // 1000 is the maximum number of users firestore can fetch. check .listUsers()'s documentation for more information.
 // It is also 1000 by default but I decided to set it explicitly to avoid confusion.
 const MAX_USERS = 1000;
@@ -54,7 +57,7 @@ const getAuthUsers = async (maxUsers: number, pageToken?: string) =>
   const { users: authUsers, pageToken } = await getAuthUsers(MAX_USERS);
   totalAuthUsers.push(...authUsers);
 
-  // @debt This is going to make another request with offset if there are the maximum users returned by the previous request. This should be using recursion.
+  // @debt This is going to make another request with offset if the maximum number of users have been returned by the previous request. This should be using recursion, because there might be more than 2k users.
   if (authUsers.length >= MAX_USERS) {
     console.log(
       `Auth users are more than ${MAX_USERS}. Fetching additional batch of firestore auth users...`
@@ -66,9 +69,7 @@ const getAuthUsers = async (maxUsers: number, pageToken?: string) =>
     totalAuthUsers.push(...additionalAuthUsers);
   }
 
-  const userDocIds: string[] = await admin
-    .firestore()
-    .collection("users")
+  const userDocIds: string[] = await usersCollectionRef
     .get()
     .then((result) => {
       console.log("Firestore users successfully fetched.");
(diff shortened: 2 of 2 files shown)


scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+  initFirebaseAdminApp,
+  makeScriptUsage,
+  parseCredentialFile,
+} from "./lib/helpers";
+
+const usage = makeScriptUsage({
+  description: `Bulk remove of user accounts that exist in the auth but are not in the firestore users collection`,
+  usageParams: `CREDENTIAL_PATH`,
+  exampleParams: `fooAccountKey.json`,
+});
```
0xdevalias, line 18: [nit] No need for the backtick quotes:

```suggestion
const usage = makeScriptUsage({
  description: 'Bulk remove of user accounts that exist in the auth but are not in the firestore users collection',
  usageParams: 'CREDENTIAL_PATH',
  exampleParams: 'fooAccountKey.json',
});
```

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+if (!credentialPath) {
+  usage();
+}
+
+if (!checkFileExists(credentialPath)) {
+  console.error("Credential file path does not exists:", credentialPath);
+  process.exit(1);
+}
+
+const { project_id: projectId } = parseCredentialFile(credentialPath);
```
0xdevalias, line 30: While we can extract this automatically, I wonder if we should. Often I use the fact that you have to enter it up front + point it at the credential key, and then it ensures they match, as a bit of a 'safety rail' to ensure scripts are run against the thing we intend. Particularly when they are super destructive scripts.

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+  await admin.auth().listUsers(maxUsers, pageToken);
+
+(async () => {
+  const totalAuthUsers: admin.auth.UserRecord[] = [];
+
+  console.log("Fetching firestore auth users...");
+  const { users: authUsers, pageToken } = await getAuthUsers(MAX_USERS);
+  totalAuthUsers.push(...authUsers);
+
+  // @debt This is going to make another request with offset if there are the maximum users returned by the previous request. This should be using recursion.
```
0xdevalias, line 57: [NB] Why should it be using recursion? To avoid mutation/etc?

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (authUsers.length >= MAX_USERS) {
+    console.log(
+      `Auth users are more than ${MAX_USERS}. Fetching additional batch of firestore auth users...`
+    );
```
0xdevalias, line 61: Is this ever going to trigger? Presumably `authUsers.length` will be at most 1000, since that is the size that will be returned, and there will be some other method of checking if there are more users to be fetched?

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+      console.log("Firestore users successfully fetched.");
+      return result.docs.map((doc) => doc.id);
+    })
+    .catch((err) => {
+      console.error(
+        "Failed when trying to fetch the users collection, skipping...",
+        err
+      );
+      return [];
+    });
```
0xdevalias, line 83: I would break this down, it makes the code simpler to read:

I usually add this right after `initFirebaseAdminApp`
```typescript
const usersCollectionRef = admin.firestore().collection("users");
```

Name `userSnapshot` whatever is correct for the actual type of that var, I forget off the top of my head:
```typescript
console.log("Fetching firestore users...");
const userIds = await usersCollectionRef.get().map(userSnapshot => userSnapshot.id)
```

I probably wouldn't bother `.catch`'ing the error here. That seems like a catastrophic failure condition, so should probably just exit then and there, by throwing the error for the final `.catch` at the bottom of the script.

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const unregisteredUsers: string[] = [];
```
0xdevalias, line 85: The name of this script/this var/etc seems confusing. The users are technically registered, they are just in a weird half registered state. I'm not actually sure exactly what that state is currently. But the name should reflect/describe that better. They're more like users without profiles created or similar presumably?

0xdevalias: [NB] And in figuring that out, I wonder if a better solution to this problem in general might be to fix whatever the bug is that causes these half created accounts to be unusable?

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  totalAuthUsers.forEach((user) => {
+    if (!userDocIds.includes(user.uid)) {
+      unregisteredUsers.push(user.uid);
+    }
+  });
```
0xdevalias, line 91: `.forEach` and `.push` should just be a `.map` and no mutation.

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+      "There are no unregistered users in the database, end of script."
+    );
+    return;
+  }
+
+  console.error(`Deleting ${numberOfUnregisteredUsers} users...`);
+
+  await admin
+    .auth()
+    .deleteUsers(unregisteredUsers)
```
0xdevalias, line 106: Does `.deleteUsers()` have a max size limit it can process? If so, we need to ensure we stay within that, and likely use the 'chunk' pattern to handle it.

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+      console.error(
+        `${numberOfUnregisteredUsers} users have been successfuly deleted.`
+      );
+    })
+    .catch((err) => {
+      console.error(
+        "Failed when trying to delete the unregistered users, end of script.",
+        err
+      );
+    });
```
0xdevalias, line 117: We're already using `await`, instead of doing these in `.then` we should just do it on the next line after the `await`.

The `.catch` would only really make sense to re-throw the error with a better message so that the final 'catastrophic failure' `.catch` handles it and logs it.

scripts/remove-unregistered-users.ts
```diff
@@ -0,0 +1,125 @@
...
+  totalAuthUsers.forEach((user) => {
+    if (!userDocIds.includes(user.uid)) {
+      unregisteredUsers.push(user.uid);
+    }
+  });
+
+  const numberOfUnregisteredUsers = unregisteredUsers.length;
+
+  if (!numberOfUnregisteredUsers) {
+    console.error(
```
0xdevalias, line 96: Why are you using `console.error` for all of these rather than `console.log`?

0xdevalias (commented):
Comments/questions/suggestions/etc.

This seems like it should do the 'delete users that have no corresponding user record in the firestore DB' thing, and if we feel that is the best solution to the original problem, then that should be fine.

I'm explicitly holding off on marking this as changes requested/approved for someone else to make that call, as I have just rushed through this review to get some comments to unblock moving forward as much as possible.

scripts/remove-unregistered-users.ts
denisdimitrov: Because it will fetch only 2k users and nothing more. We need to add recursion and every time it fetches 1000 users it should trigger again. Where right now, it happens only once.

scripts/remove-unregistered-users.ts
denisdimitrov: Good catch. Looks like it has a limit of 1000 as well.

scripts/remove-unregistered-users.ts
denisdimitrov: Fair enough. I'll replace them.

scripts/remove-unregistered-users.ts
denisdimitrov: > The name of this script/this var/etc seems confusing. The users are technically registered, they are just in a weird half registered state. I'm not actually sure exactly what that state is currently. But the name should reflect/describe that better. They're more like users without profiles created or similar presumably?

Maybe rename the script to `remove-unfinished-profiles.ts` and this line to
```tsx
const unfinishedProfiles: string[] = [];
```


> [NB] And in figuring that out, I wonder if a better solution to this problem in general might be to fix whatever the bug is that causes these half created accounts to be unusable?

Definitely, I can easily reproduce this by removing my account from the `users` collection and then fix the problem.

scripts/remove-unregistered-users.ts
denisdimitrov: IMO it doesn't really matter because this is a script. And the rest of the scripts are following this same pattern. If you really want this change I will do it but I don't think it's something huge.

scripts/remove-unregistered-users.ts
denisdimitrov: They have to be filtered first otherwise `.map` will return `null` when it skips the `if` statement. This means one more iteration but since this is a script I don't see a problem.

scripts/remove-unregistered-users.ts
denisdimitrov: > Is this ever going to trigger?

Yes. I suppose I can leave it to `==` instead of `>=`. 

> and there will be some other method of checking if there are more users to be fetched?

There is no built-in method for it. Some people keep track of their users by incrementing a value in the database, which feels hacky. I suppose checking if the number of users fetched `==` the maximum that means there are more. 

Of course there is the edge case where the database can have exactly 1000 users and this logic will trigger again. I'm not sure if this will return an empty array or `undefined`.

scripts/remove-unregistered-users.ts
denisdimitrov: > I probably wouldn't bother .catch'ing the error here. That seems like a catastrophic failure condition, so should probably just exit then and there, by throwing the error for the final .catch at the bottom of the script.

It will exit but it's it better to add error message and 'print' it in the console just for a user feedback that the script has failed at that particular point.

scripts/remove-unregistered-users.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1489/commits/9e681f33610b17a47d185592323c0fb8443f6785

scripts/remove-unregistered-users.ts
denisdimitrov: Naming fixed in: https://github.com/sparkletown/sparkle/pull/1489/commits/9e681f33610b17a47d185592323c0fb8443f6785

scripts/remove-unregistered-users.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1489/commits/9e681f33610b17a47d185592323c0fb8443f6785

scripts/remove-unregistered-users.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1489/commits/9e681f33610b17a47d185592323c0fb8443f6785

scripts/remove-unregistered-users.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1489/commits/9e681f33610b17a47d185592323c0fb8443f6785

denisdimitrov:
@mike-lvov Take a look at this PR when you are available. It might help us with the OHBM user issue. Glenn is burned out and told me to ask for your advice.

scripts/remove-unregistered-users.ts
denisdimitrov: I can't really say anything about this subject. It's your call.

scripts/remove-unregistered-users.ts
mike-lvov: As I can see from the file diff, the recursion was removed and now we only fetch 2k users at most? Doesn't seem right to me.

scripts/remove-unregistered-users.ts
mike-lvov: I think changing to `===` should be okay?

scripts/remove-unregistered-users.ts
0xdevalias: Ah, makes sense. Instead of recursion this seems to just be handled with loops and modifying result variables.

While I’m not so much a fan of that pattern in general, and recursion would probably allow us to avoid it, I also don’t think the solution here would need to inherently involve recursion.

Could just be something like: while (stillMoreToGet) results.push(…getMore())

0xdevalias: @mike-lvov the recursion was never there and it only ever fetches 2k users. Which I missed, and what @denisdimitrov explained above.

scripts/remove-unregistered-users.ts
0xdevalias: Ah, yeah ok, makes sense. == is probably more obvious when reading the code, but maybe also a comment explaining this aspect of how we’re finding more + checking the docs to see what is returned when there are no more to get to ensure we handle that ‘final case’ correctly.

There’s other scripts that use this same SDK endpoint that seem to handle it in a loop, maybe check what they’re doing?

scripts/remove-unregistered-users.ts
0xdevalias: So that’s a what I meant by rethrowing the error with a better message.

.catch(err => throw new Error(‘my new message’, err)

scripts/remove-unregistered-users.ts
0xdevalias: Sounds good. Fixing the bug in general is OOS of this PR, but we should backlog it if we haven’t already got one for it (I feel like there is a really old bug about this already)

scripts/remove-unregistered-users.ts
0xdevalias: I’m not going to block this PR over it, but if existing scripts do it it’s only for legacy reasons, none of the newer scripts I’ve written/seen do it this way, it makes the code simpler overall, and it makes it easier to write in the first place; so using our best practice patterns means less bad examples for people to copy from and thibk are the way to do things in future, so I’d prefer it.

mike-lvov:
@denisdimitrov do we still need this PR?

denisdimitrov:
Outdated PR, the script was never needed and the problem wasn't in the accounts at all.

PR #1490, opened 2021-06-02
Add thumbnails to posterhall

matteomancini:
A few things to improve the poster hall:

- [x] add the thumbnails from each poster;
- [ ] add full authors list;
- [ ] show profile pic for first author.

### DIFF
--- a/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
+++ b/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
@@ -18,7 +18,8 @@ export interface PosterPreviewProps {
 export const PosterPreview: React.FC<PosterPreviewProps> = ({
   posterVenue,
 }) => {
-  const { title, authorName, categories } = posterVenue.poster ?? {};
+  const { title, authorName, categories, thumbnailUrl } =
+    posterVenue.poster ?? {};
 
   const venueId = posterVenue.id;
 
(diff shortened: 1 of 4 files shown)


0xdevalias:
@mike-lvov Do you have capacity to take a look at this? Seems like a fairly small PR/diff, so shouldn't be too onerous to review.

src/types/posters.ts
```diff
@@ -3,4 +3,5 @@ export type Poster = {
   categories: string[];
   introVideoUrl: string;
   authorName: string;
+  thumbnailUrl: string;
```
0xdevalias, line 6: Is this guaranteed to always be defined on every poster? My guess is probably not, so we should probably mark it as 'optional', and then handle that potentially missing state appropriately in the code where needed:
```suggestion
  thumbnailUrl?: string;
```

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -46,6 +47,8 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
       <div className="PosterPreview__categories">{renderedCategories}</div>
 
+      <img src={thumbnailUrl} alt="Thumbnail" />
```
yarikoptic, line 50: could this be made conditional on availability of the thumbnail and replace thumbnailUrl with some placeholder (e.g. just stating "No thumbnail") better than default "no image" icon etc I see currently:

image

margulies: Just an idea, but perhaps if the thumbnail is missing, we could simply display a colored rectangle, where is the color is randomly selected from a few options?

mike-lvov:
Any progress on this one?

matteomancini:
I've been quite swamped with organising the OHBM Brainhack, but I'd like to wrap this up quite shortly - do you have by any chance any code section where it is checked if an attribute is defined (in my case, the `thumbnailUrl`) and something is done about it? That would really help me to quickly finish this. Thank you.

mike-lvov:
Thumbnails were added in another PR. Seems irrevant now

PR #1496, opened 2021-06-03
Add 4 new environments: demo, unity, clever, burn

sunny-viktoryia:
Add the config settings for 4 new firebase environments.

partially closes: https://github.com/sparkletown/internal-sparkle-issues/issues/731
partially closes: https://github.com/sparkletown/internal-sparkle-issues/issues/780
partially closes: https://github.com/sparkletown/internal-sparkle-issues/issues/782
partially closes: https://github.com/sparkletown/internal-sparkle-issues/issues/784

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -20,7 +20,11 @@
     "memrise": "sparkle-memrise",
     "unesco": "sparkle-unesco",
     "ohbm": "sparkle-ohbm",
-    "sparkle-pa": "sparkle-pa"
+    "sparkle-pa": "sparkle-pa",
+    "sparkle-demo": "sparkle-demo-e65b1",
+    "sparkle-clever": "sparkle-clever",
+    "sparkle-unity": "sparkle-unity",
+    "sparkle-burn": "sparkle-burn"
   },
   "targets": {
     "co-reality-map": {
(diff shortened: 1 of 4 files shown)


mike-lvov (approved):
LGTM

.firebaserc
```diff
@@ -20,7 +20,11 @@
     "memrise": "sparkle-memrise",
     "unesco": "sparkle-unesco",
     "ohbm": "sparkle-ohbm",
-    "sparkle-pa": "sparkle-pa"
+    "sparkle-pa": "sparkle-pa",
+    "sparkle-demo": "sparkle-demo-e65b1",
```
0xdevalias, line 24: Curious, why did the demo environment end up with this suffix on the projectId?

.firebaserc
```diff
@@ -175,6 +179,34 @@
           "sparkle-pa"
         ]
       }
+    },
+    "sparkle-demo": {
```
0xdevalias, line 183: Not 100% but I think this chunk may be wrong based on the demo projectId

firebase.json
```diff
@@ -237,6 +237,50 @@
         }
       ],
       "ignore": ["firebase.json", "**/.*", "**/node_modules/**"]
+    },
+    {
+      "target": "sparkle-demo",
```
0xdevalias, line 242: Does this work correctly given it differs from the demo projectId?

PR #1499, opened 2021-06-03
Profile popup UX/UI changes

oleksii-front:
> AAU I can add button links into my profile

closes: https://github.com/sparkletown/internal-sparkle-issues/issues/773


sofisparkle:
Just a little push on this as we need it for 17th now on GitHub

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
   const toggleAnonMode = useCallback(async () => {
-    if (user && profile) {
-      profile.anonMode = !profile?.anonMode;
-      await updateUserProfile(user.uid, { anonMode: profile.anonMode });
+    if (user) {
+      user.anonMode = !user?.anonMode;
+      await updateUserProfile(user.id, { anonMode: user.anonMode });
     }
-  }, [profile, user]);
+  }, [user]);
```
mike-lvov, line 74: What is the intent of this change?

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
 
   const toggleMirrorVideo = useCallback(async () => {
-    if (user && profile) {
-      profile.mirrorVideo = !profile?.mirrorVideo;
-      await updateUserProfile(user.uid, { mirrorVideo: profile.mirrorVideo });
+    if (user) {
+      user.mirrorVideo = !user?.mirrorVideo;
+      await updateUserProfile(user.id, { mirrorVideo: user.mirrorVideo });
     }
-  }, [profile, user]);
```
mike-lvov, line 68: What is the intent of this change?

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+      <div className="UserInformationContent__information">
+        <div>
+          <UserAvatar user={user} showStatus large />
         </div>
+        <div className="UserInformationContent__text-container">
+          <h3 className="UserInformationContent__user-name">
+            {user?.partyName ?? DEFAULT_PARTY_NAME}
+          </h3>
+          <div
+            title={email ?? ""}
```
mike-lvov, line 105: Any reason to have `?? ""`?

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+      {profileQuestions &&
+        profileQuestions.map((question: QuestionType) => (
+          <div key={question.name} className="question-section">
+            <div className="question">{question.text}</div>
+            <div>
+              {
+                // @ts-ignore question.name is a correct index for type User
+                (profile && profile[question.name]) ||
+                  DEFAULT_PROFILE_QUESTION_ANSWER
+              }
```
mike-lvov, line 133:
```suggestion
                (profile?[question.name]) ||
                  DEFAULT_PROFILE_QUESTION_ANSWER
              }
```

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+            </div>
+          </div>
+        ))}
+
+      <ul className="UserInformationContent__profile-links">
+        {user?.profileLinks?.map((profileLink) => (
+          <li key={profileLink.title}>
+            {profileLink.title}{" "}
```
mike-lvov, line 141:
```suggestion
            {profileLink.title}
```

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+                // @ts-ignore question.name is a correct index for type User
+                (profile && profile[question.name]) ||
+                  DEFAULT_PROFILE_QUESTION_ANSWER
+              }
+            </div>
+          </div>
+        ))}
+
+      <ul className="UserInformationContent__profile-links">
+        {user?.profileLinks?.map((profileLink) => (
```
mike-lvov, line 139: useMemo

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+          <li key={profileLink.title}>
+            {profileLink.title}{" "}
+            <button
+              className="button--a"
+              onClick={editProfileLink(profileLink)}
```
mike-lvov, line 144: useCallback

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
+            >
+              Edit
+            </button>
+          </li>
+        ))}
+      </ul>
+
+      <button
+        className="UserInformationContent__add-profile-link button--a"
+        onClick={editProfileLink(undefined)}
```
mike-lvov, line 154: useCallback

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
-        onClick={() => setIsPasswordEditMode(true)}
-      />
-
-      <input
-        className="btn button"
-        value="Log out"
-        onClick={logout}
-        type="button"
-      />
-    </>
```
mike-lvov, line 175: Where did we move these controls?

src/settings.ts
```diff
@@ -59,6 +59,7 @@ export const SPARKLEVERSE_LOGO_URL = sparkleverseLogo;
 
 export const DEFAULT_PARTY_NAME = "Anon";
 export const DISPLAY_NAME_MAX_CHAR_COUNT = 40;
+export const DEFAULT_PROFILE_QUESTION_ANSWER = "unknown";
```
mike-lvov, line 62: @sofisparkle is this the correct default answer?

src/components/organisms/Badges/Badges.tsx
```diff
@@ -146,28 +146,28 @@ export const Badges: React.FC<{
...
 
   return (
-    <div className="badges-component">
-      <div className="visits">
-        <div className="visit-item">
-          <span className="visit-item__value">
+    <div className="Badges">
+      <div className="Badges__visits visits">
+        <div className="visits__item">
+          <span className="visits__item-value">
```
mike-lvov, line 152: Why is it not aligned with our patterns?

I'd expect it to be smth like `Badges__visits`, `Badges__visit`, `Badges__visit-value`, etc

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
```diff
@@ -68,64 +75,69 @@ const EditProfileForm: React.FunctionComponent<PropsType> = ({
...
-            name="partyName"
-            className="input-block input-centered"
-            placeholder="Your display name"
-            ref={register({
-              required: true,
-              maxLength: DISPLAY_NAME_MAX_CHAR_COUNT,
-            })}
+    <div className="EditProfileForm">
+      <h1 className="EditProfileForm__title">Edit profile</h1>
+      <div className="form">
```
mike-lvov, line 80: should it have onSubmit?

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
```diff
@@ -68,64 +75,69 @@ const EditProfileForm: React.FunctionComponent<PropsType> = ({
...
+          containerClassName="EditProfileForm__input-container-name"
+          inputClassName="EditProfileForm__input-name"
+          name="partyName"
+          placeholder="Your display name"
+          ref={register({
+            required: true,
+            maxLength: DISPLAY_NAME_MAX_CHAR_COUNT,
+          })}
+        />
+        {errors.partyName && errors.partyName.type === "required" && (
```
mike-lvov, line 91:
```suggestion
        {errors.partyName?.type === "required" && (
```

mike-lvov: Is it possible that the error types are exported from the library?

mike-lvov: If not, we should create an enum for those

mike-lvov: Oh, since it was copyied, NB

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+        user?.profileLinks?.filter(
+          (userProfileLink) =>
+            userProfileLink.title === profileLink?.title ||
+            userProfileLink.title === title
+        ).length ?? 0;
+
+      return sameTitleCount === allowedCountOfSameTitles;
+    },
+    [profileLink, user?.profileLinks, edit]
+  );
```
mike-lvov, line 48: It looks too complicated, but it solves the edgecase, if a person renames the Profile link title into an existing one. I can't suggest a better approach now, but I'm sure there is one @0xdevalias @sunny-viktoryia

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+      const newProfileLinks = user.profileLinks ?? [];
+      const profileLinkToEdit = newProfileLinks.find(
+        (userProfileLink) => userProfileLink.title === profileLink?.title
+      );
+
+      if (profileLinkToEdit) {
+        Object.assign(profileLinkToEdit, profileLinkFormData);
+      } else {
+        newProfileLinks.push(profileLinkFormData);
+      }
```
mike-lvov, line 82: It looks too complicated. And I don't understand what is the intent here.

mike-lvov (approved):
We can merge it, most of the comments are NB/stylish. There are a few moments that I'd like to clarify. Eg. the implementation of onSubmit/etc

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
mike-lvov: Also, Object.assign must re rewritten not to mutate object, but to create a new one

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -1,179 +1,214 @@
...
-  }, [firebase, hideModal, history, venueId]);
+  }, [firebase, history, venueId]);
 
   const toggleKidsMode = useCallback(async () => {
-    if (user && profile) {
-      profile.kidsMode = !profile?.kidsMode;
-      await updateUserProfile(user.uid, { kidsMode: profile.kidsMode });
+    if (user) {
+      user.kidsMode = !user?.kidsMode;
+      await updateUserProfile(user.id, { kidsMode: user.kidsMode });
```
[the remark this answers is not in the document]
0xdevalias: This is mutating `user.kidsMode`, which is not ok. This is a blocker. This cannot be merged in it's current state.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: ⚠️ This is mutating `user.anonMode`, which is not ok. This is a blocker. This cannot be merged in its current state.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: ⚠️ This is mutating `user.mirrorVideo`, which is not ok. This is a blocker. This cannot be merged in its current state.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: No point, if `email` is `undefined` then the prop will not be set, which is more correct/valid than setting it to an empty string. This is a blocker, please change.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: FYI: I think it needs to be `profile?.[question.name]`

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: @mike-lvov Why are you saying to remove this here? This is what React does to ensure a space is included. Do we not want a space here?

0xdevalias: Looking at the wider code, this looks like it will include a space between the title and the button, which, without looking closer at the PR/designs, sounds pretty logical to me.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: Also, unless `editProfileLink` is a function returning a function, the syntax here is incorrect and won't even work. This will run that function on every render.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: Also, unless `editProfileLink` is a function returning a function, the syntax here is incorrect and won't even work. This will run that function on every render.

src/settings.ts
0xdevalias: Do we need a default answer? I thought we explicitly removed the display of unanswered questions recently in one of OHBM's open source contributions.

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
0xdevalias: Should it be within a proper `form` tag?

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
0xdevalias: Agreed that this looks way too convoluted/complicated. You're also shadowing variables like `title`/etc here which makes the code way harder to reason about.

It's hard to even follow the logic of what is intended here to then be able to re-write it cleaner.

0xdevalias: This is used in the `Yup` `validationSchema` defined inline below, as the handler to `Yup`'s `.test`:

- https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema

Basically the handler is passed the `value` to be tested + an optional `context` that contains `path`, `schema`, `options`, `parent`, `originalValue`, `createError`

---

`user` is of type `WithId<User>`, optionally passed in from props, so may be `undefined`

---

`validationSchema` is passed into `useForm`.. which i'm not even sure is correct based on what I am seeing in the usage docs?

- https://react-hook-form.com/api/useform

0xdevalias: Looking at `node_modules/react-hook-form/dist/useForm.d.ts:4` it definitely mentions `validationSchema` there + we use it across the codebase in as similar way.. maybe the API changed between the version we are using + the latest docs on the website?

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
0xdevalias: ES6+ object spread syntax, never `Object.assign`

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
0xdevalias: Ok, so I think the intent of this is that:
- if we're in `edit` mode, then there can be 1 matching duplicate title (presumably intended to be the current title being edited)
- if we're not in edit mode, then there can't be any duplicates

We get passed the `user` and the `profileLink`.

`ProfileLink` has a `title` and `url`.

0xdevalias: I'm not really sure of the mechanics/usage of `edit` and didn't want to dig further trying to get the context in my head, but something like this should work and be far simpler:

```typescript
const validateUniqueTitle = useCallback(
  (checkTitle: string) => {
    if (title === profileLink.title) return true

    const isDuplicateTitle = user?.profileLinks?.find(link => link.title === title) ?? false

    return !isDuplicateTitle
  [profileLink, user?.profileLinks]
);
```

0xdevalias (changes requested):
I haven't reviewed this PR in full (even at a high 'quick pass' level), but from what i've seen in @mike-lvov's review in https://github.com/sparkletown/sparkle/pull/1499#pullrequestreview-688402780, this shouldn't be approved/mergeable right now, and I have concerns that this may need a deeper review + further fixes before it will be.

Marking as changes required so that this isn't merged before needed fixes are made.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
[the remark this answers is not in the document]
mike-lvov: I didn't call it a blocker, since it was mutated before the change, so it's probably part of the logic that used to work

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
mike-lvov: > @mike-lvov Why are you saying to remove this here? This is what React does to ensure a space is included. Do we not want a space here?

I don't mind having a space there. I would expect it to be handled with CSS

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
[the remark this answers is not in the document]
0xdevalias: Fair. Makes sense. Still, we should be eliminating mutations like this everywhere we see them when we see them. They're a hella antipattern + can cause some super weird and hard to figure out bugs, so better to squash them as they come up. They're also generally pretty trivial to fix properly.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
0xdevalias: > I would expect it to be handled with CSS

Unsure in this particular instance, but generally speaking this pattern of `{" "}` is pretty standard react for when we need a space between elements in JSX. It's basically (if not literally) the equivalent of using `&nbsp;` in HTML.

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: agree, there is no point  for `string | undefined` type. But we have `null` allowed. Should I remove it and leave just `string`?
```ts
export interface UserInfo {
  displayName: string | null;
  email: string | null;
  phoneNumber: string | null;
  photoURL: string | null;
  providerId: string;
  uid: string;
}
```

> Type 'null' is not assignable to type 'string | undefined' .ts(2322)

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: The view is changed a bit. 

It looked this way:
image


This is the way it looks now:
image

* edit input -> `editIcon` https://github.com/sparkletown/sparkle/blob/b40d96493471fd353df1a05041e18c3b7e9cbd00/src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx#L117:L122
* change password, logout -> https://github.com/sparkletown/sparkle/blob/b40d96493471fd353df1a05041e18c3b7e9cbd00/src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx#L203:L211

src/components/organisms/ProfileModal/EditPasswordForm/EditPasswordForm.tsx
```diff
@@ -104,14 +113,12 @@ const EditPasswordForm: React.FunctionComponent<PropsType> = ({
           value="Save changes"
         />
       </form>
-      <div
-        className="cancel-button"
-        onClick={() => setIsPasswordEditMode(false)}
+      <button
+        className="button--a"
+        onClick={() => setUserProfileMode(UserProfileMode.DEFAULT)}
```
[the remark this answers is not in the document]
sunny-viktoryia: fixed here: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-20f5d23e6f3fb3af8d06ad430f2e5ce192619c8303852251bffb6e5cc64e5b5aR49-R53

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.scss
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  &__input {
+    text-align: left;
+    margin-bottom: 0;
+  }
+
+  &__input-error {
+    color: $input--invalid;
+    font-size: 0.8rem;
```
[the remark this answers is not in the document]
sunny-viktoryia: added here: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-23e4e421aa6f45fdd17a7cbdd6aec453653be326cbd3d6c5fe5fb23da5204a72R4-R5

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.scss
```diff
@@ -0,0 +1,77 @@
...
+    margin-top: -$spacing--sm;
```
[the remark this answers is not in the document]
sunny-viktoryia: great catch! this `&__input-error` was used before my enhancement of `InputField`. So this selector is not needed anymore. Removed here: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-23e4e421aa6f45fdd17a7cbdd6aec453653be326cbd3d6c5fe5fb23da5204a72L25-L31

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: fixed: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9f9d7328f3ef637d17d442f038b2da17f90a06672d7fb0845d12e066305ee141R148-R156

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: yes, that space is needed to separate the `title` and the `button` which looks like an `a`-tag (does not have any borders / margins).
image

sunny-viktoryia: as the owner of the PR I resolve this conversation leaving `{" "}` in place! 😎

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: here is the `useMemo` https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9f9d7328f3ef637d17d442f038b2da17f90a06672d7fb0845d12e066305ee141R83-R84

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: moved that to the `useMemo` of the profile links and renamed `editProfileLink` to `makeEditProfileLink` as it's a currying function.
https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9f9d7328f3ef637d17d442f038b2da17f90a06672d7fb0845d12e066305ee141R83-R84

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: moved that to the `useMemo` of the profile links and renamed `editProfileLink` to `makeEditProfileLink` as it's a currying function.
https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9f9d7328f3ef637d17d442f038b2da17f90a06672d7fb0845d12e066305ee141R83-R84

src/settings.ts
sunny-viktoryia: there was a default variable when I came to that PR, I just reworked that.
I've just checked the logic that was initially here, and realized that it had been changed a bit, in the newer version we did show all the questions even if they hadn't been answered by the user.

So I've fixed that. Now we see only the answered questions (as it was before the PR): https://github.com/sparkletown/sparkle/pull/1499/commits/a5c297f5834a0cb31d6f35d0afe32ddc1f003ac1

(`DEFAULT_PROFILE_QUESTION_ANSWER` is removed)

src/components/organisms/Badges/Badges.tsx
sunny-viktoryia: renamed the class names: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
sunny-viktoryia: added `onSubmit`, `form`-tag: https://github.com/sparkletown/sparkle/commit/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
sunny-viktoryia: added `Yup` validation here: 
https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9456a36066fb33e91efde7cd544a40ca21b20f06174f7bf2a401eb0da7f1e7e1R28-R37

removed these blocks: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-9456a36066fb33e91efde7cd544a40ca21b20f06174f7bf2a401eb0da7f1e7e1L91-L99
(because I have improved `InputField` to accept the input-error and show the error-message.)

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
sunny-viktoryia: @0xdevalias OMG!!!! SO EASY!!! THANK YOU! ❤️ 
I was also not happy with the way I implemented that... 

Amazing fix here: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-7e60858c69455cc177571d9c3680b6df13c67916f9827da273b5e0bf46ee9ee7R36-R42

src/components/organisms/ProfileModal/EditProfileLinkForm/EditProfileLinkForm.tsx
sunny-viktoryia: no-mutation update is here: https://github.com/sparkletown/sparkle/pull/1499/commits/4b9c4fee3c3c8ac3d75c4a18814995447bfa29da#diff-7e60858c69455cc177571d9c3680b6df13c67916f9827da273b5e0bf46ee9ee7R70-R72

```ts
      const newProfileLinks = [...(user.profileLinks ?? [])];
      const profileLinkToEdit = newProfileLinks.findIndex(
        (link) => link.title === profileLink?.title
      );

      if (profileLinkToEdit > -1) {
        newProfileLinks[profileLinkToEdit] = profileLinkFormData;
      } else {
        newProfileLinks.push(profileLinkFormData);
      }
```

`profileLinkToEdit` if for the check if there is an profile link we need to replace / update (and if so, which exactly)

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
mike-lvov: I guess it's legacy then. Just leave a comment

mike-lvov (approved):
Looks good

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
sunny-viktoryia: added a `@debt` comment https://github.com/sparkletown/sparkle/pull/1499/commits/354d8df4363f9df93752628de4e030466872f219

PR #1510, opened 2021-06-07
add howto on running emulators in README.md

goran-peoski-work:
New section in README.md for running the Firebase emulators suite


0xdevalias:
Had a quick skim, will give a more in depth review when I have capacity. Looks like a good start! I think instead of documenting the code/version changes that might be required here, we might be better to actually make them (probably in a separate PR). I have an existing draft PR for upgrading our firebase libs.

yurii-lubynets (approved):

goran-peoski-work:
> we might be better to actually make them (probably in a separate PR). I have an existing draft PR for upgrading our firebase libs.

@0xdevalias 

I've already anticipated this, only, didn't think of noting it here as well, in a rush I just put a comment on the PR:
https://github.com/sparkletown/internal-sparkle-issues/issues/781#issuecomment-855750046

Basically, I pushed 2 branches, one for the docs, other for the code (minimal changes) that I used to run the stuff.
This only because I know more works needs to be done when you upgrade major version of your database adapter.

goran-peoski-work:
Oh, branch is https://github.com/sparkletown/sparkle/tree/feature/emulators-code

0xdevalias:
@yurii-lubynets @goran-peoski-work As per my comments in https://github.com/sparkletown/sparkle/pull/1510#issuecomment-855763213, this PR wasn't really in a state that it should have been merged as-is. The README presents our 'front marketing' and 'first line of support' for the open source project, and how to use it. As my comments suggested, there were further changes that we needed to land + adjustments needed to be made to the text here before it was ready to be merged.

(cc // @jonboutelle FYI)

PR #1511, opened 2021-06-07
add initial skeleton for admin v3 final design

denisdimitrov:
This is the initial skeleton of the admin v3's final design. It just splits the admin logic into two separate screens. One with the old implementation and one for the new one with 3 new tabs. Me, @sunny-viktoryia and @goran-peoski-work are going to work on this 3 new tabs in parallel and we are currently blocked. 

Unblocks:
https://github.com/sparkletown/internal-sparkle-issues/issues/681
https://github.com/sparkletown/internal-sparkle-issues/issues/680
https://github.com/sparkletown/internal-sparkle-issues/issues/683


src/components/molecules/AdminVenueCard/AdminVenueCard.scss
```diff
@@ -1,5 +1,7 @@
 @import "scss/constants.scss";
 
+$admin-venue-card-button-margin: 5px 0;
```
goran-peoski-work, line 3: Is there app wide constant for more uniformed margins and paddings? Something to be used instead of `5px` ?

denisdimitrov: There is not. There are some padding/margin values in `/scss/constants.scss` but we've decided to keep global/common things in there and the component specific ones at the top of their own `.scss` file. @0xdevalias @mike-lvov Correct me, If I'm wrong.

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const renderAdminAdvancedTabs = useMemo(() => {
+    return Object.entries(adminVenueTabs).map(([id, text]) => (
+      <Nav.Link
+        key={id}
+        className={classNames("AdminVenueView__tab", {
+          "AdminVenueView__tab--selected": selectedTab === id,
+        })}
+        eventKey={id}
+      >
```
goran-peoski-work, line 54: `key={id}` vs `key={key}`
or
`selectedTab === id` vs `selectedTab === tab`
or
`eventKey={id}` vs `eventKey={key}`

Any objective preference for `id` over the rest?

denisdimitrov: > Any objective preference for id over the rest?

Not really. `key` makes more sense than `id`. I will change it.

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/b2285b9a5ca9ba5346f299db9386d0571fb1e9d3

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
goran-peoski-work: Could there be something like a base value, a unit of whitespace in constants, then each component use multiples of it for itself? Right now just having 5px or 16px vs 1 * $space, 3 * $space makes it look inconsistent (why 16, not 15, or 5 instead of 4) and not allowing to just use IDE shortcut to list all places one might need to modify in some future update.

src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -60,6 +62,12 @@ const AppRouter = () => {
         <Route path="/admin/venue/creation" component={VenueWizard} />
         <Route path="/admin/venue/edit/:venueId" component={VenueWizard} />
 
+        <Route path="/admin_v2/venue/:venueId?" component={AdminVenueView} />
+        <Route
+          path="/admin_v2/advanced-settings/:venueId?"
```
goran-peoski-work, line 67: Why is path using both underscore and dash? Is there an obstacle of sticking to one?

denisdimitrov: This was introduced this way by an ex-dev and he decided to use the underscore syntax. We just went with it. It's just a temp url to test the new functionality. Our final goal is to make this `/admin`

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
denisdimitrov: What's the benefit? IMO, It will add unnecessary complexity and decrease the readability. You should bring up questions like this one in slack. That way everyone can give input on the subject.

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+  basicInfo = "basic_info",
+  entranceExperience = "entrance_experience",
+  advancedMapSettings = "advanced_map_settings",
+}
+
+const adminAdvancedTabs: Record<AdminAdvancedTab, String> = {
+  [AdminAdvancedTab.basicInfo]: "Start",
+  [AdminAdvancedTab.entranceExperience]: "Entrance",
+  [AdminAdvancedTab.advancedMapSettings]: "Advanced",
+};
```
goran-peoski-work, line 29: For a "real" constant object:
```suggestion
const adminAdvancedTabs: Record<AdminAdvancedTab, String> = Object.freeze({
  [AdminAdvancedTab.basicInfo]: "Start",
  [AdminAdvancedTab.entranceExperience]: "Entrance",
  [AdminAdvancedTab.advancedMapSettings]: "Advanced",
});
```

goran-peoski-work (approved):

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: The object is not being mutated or modified anywhere, I understand that this is a 'secure' way to ensure that this is not going to be ever changed, but maybe we should be using `readonly` instead of ES5 functions. We also use immutability whenever we can.

goran-peoski-work: `Record<AdminAdvancedTab, String>` can be then changed to `Readonly<Record<AdminAdvancedTab, String>>` as a signal for "immutability".

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
goran-peoski-work: My bad, commented before checking. There already **is** something like it in `/scss/constants.scss` :

```scss
$spacing--xs: 4px;
$spacing--sm: 8px;
$spacing--md: 12px;
$spacing--lg: 16px;
$spacing--xl: 24px;
$spacing--xxl: 30px;
```

yurii-lubynets (approved):
LGTM

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,60 @@
...
 
-const DEFAULT_TAB = AdminVenueTab.dashboard;
+const DEFAULT_TAB = AdminVenueTab.spaces;
 
-export interface AdminVenueViewProps {
-  venue: Venue_v2;
-}
+export const AdminVenueView: React.FC = () => {
+  const { user } = useUser();
+  useAdminVenues(user?.uid);
```
sunny-viktoryia, line 30: (⚠️ warning, probably a silly question) can I ask why do we need to use this hook here? seems like it fetches all the venues which are current user admin for. But we don't use the list of those venues here in the component, do we?

would like to understand the logic and the reason why it's here.

src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -60,6 +62,12 @@ const AppRouter = () => {
         <Route path="/admin/venue/creation" component={VenueWizard} />
         <Route path="/admin/venue/edit/:venueId" component={VenueWizard} />
 
+        <Route path="/admin_v2/venue/:venueId?" component={AdminVenueView} />
```
sunny-viktoryia, line 65: if we end up creating functions for the routes in `utils/url` (like I've suggested in other places)  we can use it here as well, if we add a default value for the parameter there 
```ts
export const adminV3VenueUrl = (venueId : string = ':venudId?') => `/admin_v2/venue/${venueId}`
```

src/pages/Admin/Admin_v2.tsx
```diff
@@ -63,7 +61,7 @@ const Admin_v2: React.FC = () => {
         <S.ViewWrapper>
           {selectedVenue ? (
             // @debt Venue_v2 has different structure than AnyVenue, 1 of them should be deprecated.
-            <AdminVenueView venue={selectedVenue as Venue_v2} />
+            <AdminVenueView />
```
sunny-viktoryia, line 64: looks like we don't need the `@debt` comment here anymore?

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+          activeKey={selectedTab}
+          onSelect={setSelectedTab}
+        >
+          {renderAdminAdvancedTabs}
+        </Nav>
+      </div>
+      {selectedTab === AdminAdvancedTab.basicInfo && <VenueWizard />}
+      {selectedTab === AdminAdvancedTab.entranceExperience && (
+        <EntranceExperience
+          venue={venue as Venue_v2}
```
sunny-viktoryia, line 78: I guess we should move the `@debt` comment here:
```ts
// @debt Venue_v2 has different structure than AnyVenue, 1 of them should be deprecated.
```

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -30,8 +30,17 @@ export const AdminVenueCard: React.FC<AdminVenueCardProps> = ({ venue }) => (
       />
       <h3>{venue.name}</h3>
     </div>
-    <Link className="admin-venue-card__button" to={`/admin_v2/${venue.id}`}>
-      Manage Party
+    <Link
+      className="admin-venue-card__button"
```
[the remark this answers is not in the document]
goran-peoski-work: Not acquainted with the rest of the code base, so my comment stems only from this diff. It looks the `className` follows BEM and as per CSS tradition it uses kebob case for the block name (while double underscore separates the element section).

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -30,8 +30,17 @@ export const AdminVenueCard: React.FC<AdminVenueCardProps> = ({ venue }) => (
...
+      to={`/admin_v2/venue/${venue.id}`}
```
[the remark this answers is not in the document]
goran-peoski-work: +1 on the idea for easier routes management, but not sure if and how it's best to go about it since parts of those routes are constants and context dependent,

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
0xdevalias: Our more recent variation on BEM uses the component name as the block as @sunny-viktoryia suggested.

I expect this code was originally written as we were starting to foray into using BEM, but before we established this aspect of the pattern.

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: The _ is legacy and will go, the - is likely canon and will stay.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
0xdevalias: @goran-peoski-work what aspects are ‘context dependent’ in a way that you feel it wouldn’t work?

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,60 @@
...
 
-export const AdminVenueView: React.FC<AdminVenueViewProps> = ({ venue }) => {
+  const venueId = useVenueId();
   const [selectedTab, setSelectedTab] = useState<string>(DEFAULT_TAB);
 
-  const selectDefaultTab = useCallback(() => {
-    setSelectedTab(DEFAULT_TAB);
-  }, []);
+  const { currentVenue: venue } = useConnectCurrentVenueNG(venueId);
```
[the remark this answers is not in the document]
0xdevalias: ‘Next generation’. The idea was basically for this function to replace the legacy useCurrentVenue stuff as it uses suboptimal patterns, but without needing to refactor it across the entire codebase.

Eventually we want to be able to remove the legacy version, and then this will be the only version (at which point we can rename it)

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: I think react router may already have some helper functions designed for constructing urls in an efficient way. Could be worth checking those out too.

0xdevalias: Less relevant for this route, but for the others, I’m curious why we need to use absolute paths everywhere, rather than relative paths when navigating around the subsections of the admin panel. Relative would make it simpler/less rigid IMO

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: I will change the routes, Its not a big deal.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,60 @@
...
-import { Button, Nav } from "react-bootstrap";
-import { Link } from "react-router-dom";
+import React, { useMemo, useState } from "react";
+import { Nav } from "react-bootstrap";
 import classNames from "classnames";
 
-import AdvancedSettings from "pages/Admin/AdvancedSettings";
-import BasicInfo from "pages/Admin/BasicInfo";
-import EntranceExperience from "pages/Admin/EntranceExperience";
-import VenueDetails from "pages/Admin/Venue/Details";
```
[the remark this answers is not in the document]
denisdimitrov: You are correct, `BasicInfo` is no longer being used and `VenueWizard` is used instead. It can be removed the the other screen can be adapted/changed if needed.

`VenueDetails` is being used in one more place. `Admin.tsx`

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: This logic fetches the venues and checks if you are an admin in that venue. Previously you could access directly this page if you had the link for it. Our check was only in `/admin_v2`. If you had the direct link to edit a specific venue EX: `/admin_v2/viktoriasvenue` you could have accessed the page. The requests were probably going to return 403. But it's still leaking information, imo.

goran-peoski-work: > This logic fetches the venues and checks if you are an admin in that venue.
Just to be clear, fetching all venues from Firebase and filter after or does the filter happen on the DB level and only get the relevant ones?

src/components/organisms/AdminVenueView/AdminVenueView.tsx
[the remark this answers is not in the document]
goran-peoski-work: Oh well, completely opposite from what I suspected.

"The letters NG, shown enclosed in a square, which are an acronym for the words No Good." - https://emojipedia.org/ng-button/

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
goran-peoski-work: Still have to learn the codebase, so I'm just making a strawman case here.
In context as in Admin or Login/Auth, if we consider them a modules on their own, should they come with their own routing tucked in their own directory structure or should one general utility/constants file for the entire project suffice?

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/0d4d60cd9f66fb4682d6f3cbf71193467e5c5a52

denisdimitrov: The names of the components and imports need to be changed too, but I don't want to make this PR too big, because we should merge this asap. I will make sure to check if there is an issue for this, if there is not will create one.

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/0d4d60cd9f66fb4682d6f3cbf71193467e5c5a52

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/0d4d60cd9f66fb4682d6f3cbf71193467e5c5a52

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: @0xdevalias @sunny-viktoryia 
> if we end up creating functions for the routes in utils/url (like I've suggested in other places) we can use it here as well, if we add a default value for the parameter there
> ```export const adminV3VenueUrl = (venueId : string = ':venudId?') => '/admin_v2/venue/${venueId}'```

This is not going to work because some of our screens rely on `venueId` and if it exists. In this case it will always exist by falling back to `:venueId`

What I've done before is have something like this for the routes:
```tsx
const SCREENS = {
  ADMIN_VENUE: {
    path: '/admin-v3/venue/:venueId',
    url: (venueId: string) => `/admin-v3/venue/${venueId}`,
    component: AdminVenue
  },
  ADMIN_SETTINGS: {
    path: '/admin-v3/advanced-settings/:venueId',
    url: (venueId: string) => `/admin-v3/advanced-settings/${venueId}`,
    component: AdminAdvancedSettings
  }
  ....
}



Usage: 

<Route path={SCREENS.ADMIN_VENUE.path} component={SCREENS.ADMIN_VENUE.component} /> 

and 

<Link to={SCREENS.ADMIN_VENUE.url(venueId)}  />
```

The only problem is that this introduces small duplication because paths without params are going to be exactly like the urls.

Ex: 
```tsx
    path: '/home/entrance',
    url: '/home/entrance',
```

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/767b8397b022e9e276ed46c3f1bd384570fa4751

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -30,8 +30,17 @@ export const AdminVenueCard: React.FC<AdminVenueCardProps> = ({ venue }) => (
...
+    >
+      Manage Space
+    </Link>
+    <Link
+      className="admin-venue-card__button"
+      to={`/admin_v2/advanced-settings/${venue.id}`}
```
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/767b8397b022e9e276ed46c3f1bd384570fa4751

src/components/organisms/AdminVenueView/AdminVenueView.tsx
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/767b8397b022e9e276ed46c3f1bd384570fa4751

src/pages/Admin/Admin_v2.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/767b8397b022e9e276ed46c3f1bd384570fa4751

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/767b8397b022e9e276ed46c3f1bd384570fa4751

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -6,6 +6,7 @@ import { AnyVenue } from "types/venues";
 import { DEFAULT_VENUE_LOGO } from "settings";
 
 import { WithId } from "utils/id";
+import { adminV3SettigsUrl, adminv3VenueUrl } from "utils/url";
```
goran-peoski-work, line 9: Slight naming inconsistency with `*V3*` vs `*v4*`, so maybe just drop the letter V so we don't always think about proper casing of single letter words. Just `admin3` and `Admin3` are clear enough I think.

denisdimitrov: This is a topic I want to talk about and I will bring it up in slack. Currently from product/design perspective this is 'Admin v3' but on our side we are actually replacing Admin v2 with Admin v3 and there will be no trace of Admin v2 in our codebase. It will be very miss-leading to have Admin and Admin v3 logic in our codebase. I suggest we come up with a better temporary name for these new screens until it's all implemented and becomes the only final Admin.
I'd really like everyone's input on this topic. @sparkletown/engineering

src/utils/url.ts
```diff
@@ -9,6 +9,14 @@ export const venueInsideUrl = (venueId: string) => {
...
 };
 
+export const adminv3VenueUrl = (venueId: string) => {
+  return `/admin-v2/venue/${venueId}`;
+};
+
+export const adminV3SettigsUrl = (venueId: string) => {
+  return `/admin-v2/advanced-settings/${venueId}`;
+};
+
```
goran-peoski-work, line 19: Naming inconsistency, same comment as https://github.com/sparkletown/sparkle/pull/1511/files#r647324726

There's probably good reason why the function names have 3 while routes 2, so electing not to go there.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
0xdevalias: I’m guessing the hook throws an error/redirects somewhere/etc when forbidden? The name of the hook doesn’t imply that sort of functionality, which makes it a little confusing. I would expect this to return a list of venues I had admin access to or similar based on the name.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
0xdevalias: @goran-peoski-work For the utilise suggested above that won’t really matter, as they are more localised to just ‘wrapping’ the url string creation as we do for other urls in utilise/url. That way we use a single semantic function call for it rather than ‘magic strings’ all over the place.

As far as routing goes, react router is designed to be used effectively with nested subroutes. We almost certainly don’t want to keep all routing at the root level in a single routes file. The nuance right now is a little arbitrary, but this will become a lot more important in future when it comes to bundle splitting, tree shaking, lazy loading, and similar optimisations.

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: https://reactrouter.com/web/api/generatePath

0xdevalias: You can handle an optional venueId in the function with a ternary?

const foo = (venueId?: string) => venueId ? ‘/foo/${venueId}’ : ‘/foo’

src/components/molecules/AdminVenueCard/AdminVenueCard.scss
0xdevalias: This is something we basically need @sparkletown/design to weigh in on. The expectation is we should be able to choose a minimal set of standards spacing and be able to use that in all cases.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
0xdevalias: @denisdimitrov can you please link the slack thread to this issue for continuity

0xdevalias: Fwiw, I’m 👎🏻 on dropping the v and just having a numeric suffix

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: It's actually `useRoles` and `useIsAdminUser` that's checks for that. I just added them in this commit: https://github.com/sparkletown/sparkle/pull/1511/commits/2afccca403af90d38dca5a81a91041709b2afac3

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: I just used `generatePath` for the functions: https://github.com/sparkletown/sparkle/pull/1511/commits/2afccca403af90d38dca5a81a91041709b2afac3

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
denisdimitrov: NG has been used for naming: https://github.com/sparkletown/sparkle/pull/1511/commits/2afccca403af90d38dca5a81a91041709b2afac3

src/utils/url.ts
denisdimitrov: New naming convention in: https://github.com/sparkletown/sparkle/pull/1511/commits/2afccca403af90d38dca5a81a91041709b2afac3

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: Backlogged here: https://github.com/sparkletown/internal-sparkle-issues/issues/807

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
denisdimitrov: Leaving slack link for continuity: https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1623149690277100

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,66 @@
...
+        eventKey={key}
       >
         {text}
       </Nav.Link>
     ));
   }, [selectedTab]);
 
+  if ((venueId && !venue) || !roles) {
+    return <LoadingPage />;
+  }
```
mike-lvov, line 55: you can use `isCurrentVenueLoaded` here, which is exposed in useCurrentVenueNG

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,66 @@
...
+
+  if (!roles.includes("admin") || !isAdminUser) {
```
mike-lvov, line 57: Why does it doublechecks if a user is an admin?

src/pages/Admin/Admin_v2.tsx
```diff
@@ -61,12 +59,7 @@ const Admin_v2: React.FC = () => {
     <>
       <S.Wrapper className="no-venue-selected">
         <S.ViewWrapper>
-          {selectedVenue ? (
-            // @debt Venue_v2 has different structure than AnyVenue, 1 of them should be deprecated.
-            <AdminVenueView venue={selectedVenue as Venue_v2} />
-          ) : (
-            <AdminVenues venues={venues} />
-          )}
+          {selectedVenue ? <AdminVenueView /> : <AdminVenues venues={venues} />}
```
mike-lvov, line 62: I think we need to pass selectedVenue here so that we don't need to handle `useCurrentVenue` inside. We have this pattern in all of our Templates

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Nav } from "react-bootstrap";
+import classNames from "classnames";
+
+import { useVenueId } from "hooks/useVenueId";
+import { useConnectCurrentVenueNG } from "hooks/useConnectCurrentVenueNG";
+import { useUser } from "hooks/useUser";
+import { useRoles } from "hooks/useRoles";
+import { useIsAdminUser } from "hooks/roles";
+
+import { Venue_v2 } from "types/venues";
```
mike-lvov, line 11:
```suggestion
import { Venue_v2 } from "types/venues";

import { useVenueId } from "hooks/useVenueId";
import { useConnectCurrentVenueNG } from "hooks/useConnectCurrentVenueNG";
import { useUser } from "hooks/useUser";
import { useRoles } from "hooks/useRoles";
import { useIsAdminUser } from "hooks/roles";

```

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+import VenueWizard from "pages/Admin/Venue/VenueWizard/VenueWizard";
+import { LoadingPage } from "components/molecules/LoadingPage";
+
+import "./AdminAdvancedSettings.scss";
+
+export enum AdminAdvancedTab {
+  basicInfo = "basic_info",
+  entranceExperience = "entrance_experience",
+  advancedMapSettings = "advanced_map_settings",
+}
```
mike-lvov, line 24: [NB] I think these enums should move somewhere into the common `type` file

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const adminAdvancedTabs: Readonly<Record<AdminAdvancedTab, String>> = {
+  [AdminAdvancedTab.basicInfo]: "Start",
+  [AdminAdvancedTab.entranceExperience]: "Entrance",
+  [AdminAdvancedTab.advancedMapSettings]: "Advanced",
```
mike-lvov, line 29: I'm not a fan of keeping the names tighted to the `display text representation`, but is there a reason to have them different?

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const DEFAULT_TAB = AdminAdvancedTab.basicInfo;
+
+export const AdminAdvancedSettings: React.FC = () => {
+  const venueId = useVenueId();
+  const [selectedTab, setSelectedTab] = useState<string>(DEFAULT_TAB);
```
mike-lvov, line 36:
```suggestion
  const [selectedTab, setSelectedTab] = useState<AdminAdvancedTab>(DEFAULT_TAB);
```

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const { user } = useUser();
```
mike-lvov, line 38: userId

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useCallback, useMemo, useState } from "react";
```
mike-lvov, line 1: This file looks a lot like a copy-paste of `AdminVenueView.tsx `. Is there a reason for that?

src/utils/url.ts
```diff
@@ -9,6 +11,18 @@ export const venueInsideUrl = (venueId: string) => {
...
+  return generatePath("/admin-ng/venue/:venueId?", {
+    venueId: venueId,
+  });
+};
+
+export const adminNGSettigsUrl = (venueId?: string) => {
+  return generatePath("/admin-ng/advanced-settings/:venueId?", {
+    venueId: venueId,
+  });
+};
```
mike-lvov, line 24:
```suggestion
export const adminNGSettigsUrl = (venueId?: string) => generatePath("/admin-ng/advanced-settings/:venueId?", {
    venueId: venueId,
  });
```

mike-lvov, line 24:
```suggestion
export const adminNGSettigsUrl = (venueId?: string) => generatePath("/admin-ng/advanced-settings/:venueId?", {
    venueId: venueId,
  });
```

mike-lvov (changes requested):
Questions/suggestions

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,66 @@
...
-const adminVenueTabs: Record<AdminVenueTab, String> = {
-  [AdminVenueTab.basicInfo]: "Start",
-  [AdminVenueTab.entranceExperience]: "Entrance",
-  [AdminVenueTab.advancedMapSettings]: "Advanced",
-  [AdminVenueTab.dashboard]: "Dashboard",
+const adminVenueTabs: Readonly<Record<AdminVenueTab, String>> = {
+  [AdminVenueTab.spaces]: "Spaces",
+  [AdminVenueTab.timing]: "Timing",
+  [AdminVenueTab.run]: "Run",
 };
```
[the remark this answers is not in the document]
goran-peoski-work: That object is a mapping from the tab key to the text displayed.

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
0xdevalias: See thread raised on Slack around this question in https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1623302331331700

src/components/organisms/AdminVenueView/AdminVenueView.tsx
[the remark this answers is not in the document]
mike-lvov: See the thread in https://github.com/sparkletown/sparkle/pull/1511#discussion_r648868400

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Both screens look identical. The only difference is the object that's being mapped. I'd like to create a follow up PR which will create a reusable screen with tabs, and you can just pass the object. That way it can be reused.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -6,6 +6,7 @@ import { AnyVenue } from "types/venues";
 import { DEFAULT_VENUE_LOGO } from "settings";
 
 import { WithId } from "utils/id";
+import { adminNGSettigsUrl, adminNGVenueUrl } from "utils/url";
```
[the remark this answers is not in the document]
denisdimitrov: I think `adminSettingsUrlNG` will imply that there already exists an old version of the function `adminSettingsUrl`.

In this case it's a settings url build function for `admin-ng` (admin v3)

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: There are 2 places which contain admin user ids. The `users` collection, and the venue field `owners`. This checks both places.

src/pages/Admin/Admin_v2.tsx
denisdimitrov: I'm glad you saw this, now that there is another new tab for the AdminVenueView, this can be removed.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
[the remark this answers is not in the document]
mike-lvov: Logical

src/components/organisms/AdminVenueView/AdminVenueView.tsx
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

The const was renamed as we discussed in the slack convo mentioned above.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,65 +1,66 @@
...
-
-export const AdminVenueView: React.FC<AdminVenueViewProps> = ({ venue }) => {
+export const AdminVenueView: React.FC = () => {
+  const venueId = useVenueId();
   const [selectedTab, setSelectedTab] = useState<string>(DEFAULT_TAB);
 
-  const selectDefaultTab = useCallback(() => {
-    setSelectedTab(DEFAULT_TAB);
-  }, []);
+  const { user } = useUser();
```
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/pages/Admin/Admin_v2.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/utils/url.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/484c4afb77a3b370f73be7ab7dbd7911b30553c0

src/components/organisms/AdminVenueView/AdminVenueView.tsx
mike-lvov: I don't think it's the case. Looking at the way the hooks are implemented, neither of them checks for venue ownership

image

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -1,85 +1,82 @@
...
         })}
-        eventKey={id}
+        eventKey={key}
       >
         {text}
       </Nav.Link>
     ));
   }, [selectedTab]);
 
+  const selectTab = useCallback((tab: string) => {
```
mike-lvov, line 54:
```suggestion
  const selectTab = useCallback((tab: AdminVenueTab) => {
```

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
mike-lvov: Withdrawing this as per latest discussions

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -0,0 +1,101 @@
...
+        eventKey={key}
+      >
+        {text}
+      </Nav.Link>
+    ));
+  }, [selectedTab]);
+
+  const selectTab = useCallback((tab: string) => {
+    setSelectedTab(tab as AdminAdvancedTab);
+  }, []);
```
mike-lvov, line 67: any reason to wrap this into a function? Looks like you can just use `setSelectedTab`

mike-lvov (approved):
A few NB comments and good to go

src/components/organisms/AdminVenueView/AdminVenueView.tsx
mike-lvov: Or as per https://github.com/sparkletown/sparkle/pull/1511#discussion_r649088680, just remove this and use `setSelectedTab`

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
denisdimitrov: We should avoid inline functions and use `useCallback` to avoid unnecessary re-renders. In this case I need to cast the parameter because it's a string.

Ex: 
```tsx
<Nav
  className="AdminAdvancedSettings__options"
  activeKey={selectedTab}
  onSelect={(value) => setSelectedTab(value as AdminAdvancedTab)}
>
```

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: This value is NavLink's onSelect first parameter, it's already typed as string, I have to cast it otherwise I get compile error from TS.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: You are right, `owners` is being used only to display the list for venues in the admin screen when you open it. 

In this case `useRoles` is unnecessary, I will remove it and use `useIsAdminUser`

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/d74ce15ca534e3215cda469caf147a7a6d26e9d5

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1511/commits/d74ce15ca534e3215cda469caf147a7a6d26e9d5

PR #1515, opened 2021-06-08
More markdown: event modal description

### LINKED ISSUE #1505: Markdown renderer needed for event modal
image

also in the admin interface -- renders the body nicely but not the title.

ping @margulies

yarikoptic:
This is how it would look like with this PR (edit: title is no longer rendered so must not be markdown)

image

in comparison to 

image

Closes #1505

### DIFF
--- a/src/components/organisms/EventModal/EventModal.scss
+++ b/src/components/organisms/EventModal/EventModal.scss
@@ -24,13 +24,6 @@ $EventModal--padding-top: 40px;
     opacity: 0.8;
   }
 
-  &__description {
-    font-size: $font-size--md;
-    opacity: 0.8;
-    padding-top: $spacing--xl;
-    padding-bottom: $spacing--xl;
-  }
-
   // @debt we need to transform the rule to `&__button { ... }` once `<Button />` styles stop appearing later than the styles from this file
   & .EventModal__button {
     background-color: $primary--live;
--- a/src/components/organisms/EventModal/EventModal.scss
+++ b/src/components/organisms/EventModal/EventModal.scss
@@ -24,6 +24,13 @@ $EventModal--padding-top: 40px;
     opacity: 0.8;
   }
 
+  &__description {
+    font-size: $font-size--md;
+    opacity: 0.8;
+    padding-top: $spacing--xl;
+    padding-bottom: $spacing--xl;
+  }
+
   // @debt we need to transform the rule to `&__button { ... }` once `<Button />` styles stop appearing later than the styles from this file
   & .EventModal__button {
     background-color: $primary--live;
--- a/src/components/organisms/EventModal/EventModal.tsx
+++ b/src/components/organisms/EventModal/EventModal.tsx
@@ -78,7 +78,9 @@ export const EventModal: React.FC<EventModalProps> = ({
           </button>
         </span>
 
-        <RenderMarkdown text={event.description} />
+        <div className="EventModal__description">
+          <RenderMarkdown text={event.description} />
+        </div>
 
         <Button
           customClass="EventModal__button"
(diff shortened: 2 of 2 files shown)


src/components/molecules/InformationCard/InformationCard.tsx
```diff
@@ -15,7 +17,9 @@ const InformationCard: React.FC<InformationCardProps> = ({
   className,
 }) => (
   <div className={classNames("information-card-container", className)}>
-    <h4 className="title">{title}</h4>
+    <h4 className="title">
```
0xdevalias, line 20: Pretty sure rendering in a title tag like this will cause issues similar to the p in p issue

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
```diff
@@ -81,7 +82,9 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
     <>
       <div className={containerClasses} onClick={showEventModal}>
         <div className="ScheduleEvent__info">
-          <div className="ScheduleEvent__title">{event.name}</div>
+          <div className="ScheduleEvent__title">
```
0xdevalias, line 85: Why do we want to render markdown in titles?

src/components/organisms/EventModal/EventModal.scss
```diff
@@ -24,13 +24,6 @@ $EventModal--padding-top: 40px;
     opacity: 0.8;
   }
 
-  &__description {
```
0xdevalias, line 27: Why was this removed?

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -68,15 +70,17 @@ export const EventModal: React.FC<EventModalProps> = ({
...
+          <RenderMarkdown text={event.name} />
+        </h4>
         <span className="EventModal__subtitle">
           by {event.host} in{" "}
           <button className="button--a" onClick={goToEventLocation}>
             {eventLocationToDisplay}
           </button>
         </span>
 
-        <p className="EventModal__description">{event.description}</p>
```
0xdevalias, line 79: I assume this was done to avoid the ‘p in p’ error, but I don’t think the solution here is what we want to be doing. Without looking too deeply, I would assume that changing the p to a dig and nesting RenderMarkdown within that would have been the needed solution.

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -68,15 +70,17 @@ export const EventModal: React.FC<EventModalProps> = ({
   return (
     <Modal show={show} onHide={onHide} className="EventModal">
       <div className="EventModal__content">
-        <h4 className="EventModal__title">{event.name}</h4>
+        <h4 className="EventModal__title">
```
0xdevalias, line 73: As above, why do we want to render markdown in titles?

0xdevalias (commented):
High level quick pass of comments/questions (from my phone)

src/components/molecules/InformationCard/InformationCard.tsx
yarikoptic: didn't see the problem, but indeed decided to not bother about titles (left sample event though with markdown for testing).  Removed them all.  Now this PR has only markdown rendering for event description

src/components/organisms/EventModal/EventModal.scss
yarikoptic: because was not used in that version. now brought back (although I do not see any effect of it when used as `div`)

src/components/organisms/EventModal/EventModal.tsx
yarikoptic: no more

mike-lvov (approved):
LGTM

PR #1520, opened 2021-06-08
[scripts] add update-venue-owners script

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/854
partially closes https://github.com/sparkletown/internal-sparkle-issues/issues/15

the script updates the list of owners of the specified venue


0xdevalias:
> * added `update-venue-owners.ts ` to be able to update venue ownership

@sunny-viktoryia This should really be in its own standalone PR, or the title of this PR should be changed. As is, it feels unrelated/OOS here.

scripts/update-venue-owners.ts
```diff
@@ -0,0 +1,130 @@
...
+enum Mode {
+  Add = "ADD",
+  Remove = "REMOVE",
+}
+
+const VALID_MODES: string[] = [Mode.Add, Mode.Remove];
+
+const usage = makeScriptUsage({
+  description: `Bulk add users to the owners list based on the supplied email address(es).\n\n(Note: the accounts must already exist.)`,
+  usageParams: `CREDENTIAL_PATH [${Mode.Add}/${Mode.Remove}] EMAIL1 EMAIL2 EMAIlN...`,
```
0xdevalias, line 33: I would add the venueIds here as well, using the venueId1,venueId2,etc pattern used in other scripts, probably before the ADD

scripts/update-venue-owners.ts
```diff
@@ -0,0 +1,130 @@
...
+
+(async () => {
+  for (const venue of VENUES_TO_CHANGE_OWNERSHIP) {
+    console.log(
+      `\n===== Updating the ownership of the following venue: ${venue}`
+    );
+    const venueDocRef = app.firestore().collection("venues").doc(venue);
+
+    for (const email of allEmailAddresses) {
+      const existingUser = await findUserByEmail(app)(email);
```
0xdevalias, line 76: We should only need to lookup the user details once, regardless of how many venues there are. This doesn’t need to be done in the inner loop.

scripts/update-venue-owners.ts
```diff
@@ -0,0 +1,130 @@
...
+
+if (!projectId) {
+  console.error("Credential file has no project_id:", credentialPath);
+  process.exit(1);
+}
+
+const app = initFirebaseAdminApp(projectId, { credentialPath });
+
+(async () => {
+  for (const venue of VENUES_TO_CHANGE_OWNERSHIP) {
```
0xdevalias, line 69: It might be good to use the batch write pattern here to do all of the updates in one atomic write. There should be an example of this in a script I worked on recently, I think it might have been the venue access one? Might still be in an unnerved PR

0xdevalias: Instead of looping through things with for like this, we tend to use .map, .filter, etc

scripts/update-venue-owners.ts
sunny-viktoryia: a good idea. 🙃 
added: https://github.com/sparkletown/sparkle/pull/1520/commits/c9c2aefa61c3f70c9702ee5722905f14a693a29e

sunny-viktoryia: a good idea. 🙃 
added: https://github.com/sparkletown/sparkle/pull/1520/commits/c9c2aefa61c3f70c9702ee5722905f14a693a29e

0xdevalias: I meant where they are all literally touching with a comma between them, then using split to get the array. That’s how the existing scripts do it.

I’m concerned about searching for position/etc in the current way in case there’s a name clash/etc that leads to weird behaviour

Issue #1462, opened 2021-05-27
PosterPreview <div> to have more actionable choices besides going into the room


yarikoptic:
It is **highly** desired (if not desperately needed) to be able to quickly get a better idea about any particular poster displayed in a poster hall.  It is needed for a number of reasons, e.g. user might want to quickly get more information about the poster/research without going to a room  since that takes time and "cancels" search result (see #1447).

I think a nice approach would be to overlay clickable icons which would perform extra actions:
- :page_facing_up:  - bring up an overlayed pane (scrollable) with rendering of the abstract (ideally in markdown -- #1383 ) . Would have some `x` to close or just allow to close outside to close, or `escape` button press (relevant: #1416) 
  - note: this is the main feature needed.  could be partially mitigated by the next one - :arrow_down:  
- :arrow_down: - download PDF of the poster
- "share" icon -- social media share - #1396
- :bookmark: -- I think this is already in development somewhere to be able to add to "my schedule" -- so might be establishing the framework for this

0xdevalias:
> 📄 - bring up an overlayed pane (scrollable) with rendering of the abstract (ideally in markdown -- #1383 ) . Would have some x to close or just allow to close outside to close, or escape button press (relevant: #1416)

@yarikoptic Is the abstract something we already have stored in firestore, or do you mean pulling it out of the poster PDF itself? If pulling it from the PDF, that's almost certainly going to be OOS.

If the data is already in firestore (I assume it would be part of `PosterPageVenue` if it is?), then it would be possible to implement some form of `Modal` (see our existing patterns for these) or similar to display it from `PosterPreview` without needing to navigate into the venue, as `PosterPreview` already has the full `PosterPageVenue` available to it in it's props.

---

> ⬇️ - download PDF of the poster

@yarikoptic Since `PosterPreview` already has the full `PosterPageVenue` (`posterVenue`) passed in it's props. `PosterPage` uses the venue config's `iframeUrl` to embed the PDF in an `iframe`.

Therefore, you could likely fairly easily add a download button in `PosterPreview` using `posterVenue.iframeUrl`, using `externalUrlAdditionalProps` or `getExtraLinkProps` from `utils/url` to ensure it opens in a new window/tab/etc. Then apply those to an `<a href` element and use `FontAwesomeIcon` with a relevant download icon from one of our fontawesome icon libs.

That would be a fairly quick/localised way to solve this core need.

---

> "share" icon -- social media share - #1396

[...]

PR #1522, opened 2021-06-08
Implements opening poster preview modal

mike-lvov:
Open modal preview, if there is intro video.
If there is no video, redirect directly to the venue


yarikoptic:
> Open modal preview, if there is intro video.
> If there is no video, redirect directly to the venue

IMHO such inconsistent among posters behavior (if I get it right) would be confusing and possibly annoying.  I would have thought that adding an explicit icon/button (e.g. 📺) in the preview for a poster (see #1462 and partial/sketch implementation for preview/download in #1527) would be a much better UX:

- consistent action upon clicking a poster
- ability to see which posters have video preview (no 📺 for those which don't)
- no unexpected audio playing all of a sudden while kids are asleep

question (may be @margulies knows) - any sample poster with such video preview for OHBM 2021?

mike-lvov:
Is irrelevant now

PR #1523, opened 2021-06-09
add I4A SSO via custom OAuth2 'Authorization Code Grant' flow

0xdevalias:
Add SSO via custom OAuth2 'Authorization Code Grant' flow.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/31

~~**Note:** This PR may need to be re-targeted onto `env/ohbm` depending on how it is finished off, as the I4A aspects of it are super specific.. Chunks of this would be useful for implementing a more generic custom OAuth provider though..~~ I think this is fine to go to `staging` based on how it's implemented now.

**Note:** When testing locally, you can't just use `npm start` as that won't map the 'rewrites' to the functions correctly. You need to use `npm run firebase:emulate-functions-hosting`.

## Reviewer Notes

If running locally to test, you need to:
  - Ensure your 'production' ENV values are mapped to staging
  - Comment out the `if (process.env.NODE_ENV === "development") {` check around `firebaseFunctions.useFunctionsEmulator("http://localhost:5001");`
  - Build the assets: `npm run build`
  - Run the function + hosting emulator: `npm run firebase:emulate-functions-hosting`
  - Access the login screen: http://localhost:5006/v/devaliashacksville
  - You should see the quicklogin icon:
    - image
  - Test credentials are available [here](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1624020034185200?thread_ts=1624019098.182500&cid=C01ER4M61TL)

If you don't do this, then you'll likely get a weird CORS error or similar, as per this slack thread: https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1624019260183800

It seems as though the preview URL doesn't work for this either, with a similar CORS error. I don't believe that will happen in production, as it seems as though we get CORS errors from various thing on the platform in general when running on the preview deploys.

I have configured `/in/devaliashacksville/` to use the custom auth, which you can find in firebase in `authConfigs/devaliashacksville`

## TODO

- [x] Refactor `AUTH_ORIGIN` to be an array of whitelisted values on `authConfig` in firestore + read the 'redirect back to origin' from URL params + validate it against the whitelist: https://github.com/sparkletown/sparkle/pull/1523#discussion_r653470100 (https://github.com/sparkletown/sparkle/pull/1523/commits/a5147f88ee70065c69b17cabac44c9e058963088 + https://github.com/sparkletown/sparkle/pull/1523/commits/eb523e2c71d8dd19a4edbf50a2172f5bc5f48e78)
- [x] UI to actually redirect to I4A to initiate the login process
  - [x] Probably need a backend function to retrieve whatever details the frontend needs (eg. to check if it has some kind of auth config)
  - [x] Some `api/*` to call it
  - [x] Then integrate that in `src/pages/Account/Login.tsx` within the 'quick login' section to navigate to `/auth/connect/i4a?venueId=foo` (ideally returned from the aforementioned check)
- [ ] See TODO notes in this review: https://github.com/sparkletown/sparkle/pull/1523#pullrequestreview-685871766
  - [ ] Redirect back to platform and display 'not allowed, please register/buy a ticket' error: https://github.com/sparkletown/sparkle/pull/1523#discussion_r653233947
- [ ] Anything else?

### Potentially OOS?

- [ ] implement finding the sovereign venue efficiently/safely in backend functions, then use it to get the appropriate authConfig: https://github.com/sparkletown/sparkle/pull/1523#discussion_r653235698
  - Note: we could potentially get away without needing this for now so long as `authConfigs` are defined for all expected 'entrypoint' venues (which should already be the sovereign venues in most if not all cases anyway..)
- [ ] implement OAuth logout url correctly: https://github.com/sparkletown/sparkle/pull/1523#discussion_r653236835


firestore.rules
```diff
@@ -1,5 +1,7 @@
 rules_version = '2';
 
+// TODO: Add appropriate data structures/rules based on + how we decide we want to lay things out: https://github.com/firebase/functions-samples/blob/master/instagram-auth/database.rules.json
```
0xdevalias, line 3: See the following thread for seeking ideas on how to lay this out 🔒 https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1623754294143800

functions/.runtimeconfig.json.example
```diff
@@ -2,6 +2,9 @@
   "project": {
     "id": ""
   },
+  "auth": {
+    "origin": "https://foo.example.com"
+  },
```
0xdevalias, line 7: [TODO] Do we want to keep this here, or move it into the `authConfig` in firestore? I think `authConfig`. And maybe name it better. It's essentially used as part of the 'redirect back to the platform' code.

functions/auth.js
```diff
@@ -1,5 +1,16 @@
...
+const { HttpsError } = require("firebase-functions/lib/providers/https");
+
+const { fetchAuthConfig } = require("./src/api/auth");
+
+const { createOAuth2Client } = require("./src/utils/auth");
+const { getJson, postJson } = require("./src/utils/fetch");
+const { checkIfValidVenueId } = require("./src/utils/venue");
+
 const PROJECT_ID = functions.config().project.id;
+const { origin: AUTH_ORIGIN } = functions.config().auth;
```
0xdevalias, line 13: [TODO] Do we want to keep this here, or move it into the authConfig in firestore? I think authConfig. And maybe name it better. It's essentially used as part of the 'redirect back to the platform' code.

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
   submittedPassword.trim() === actualPassword.trim() ||
   lowercaseFirstChar(submittedPassword.trim()) ===
     lowercaseFirstChar(actualPassword.trim());
+
+/**
+ * Redirect the user to the authentication consent screen. The 'state' cookie is set for later state verification.
```
0xdevalias, line 35:
```suggestion
 * Redirect the user to the authentication consent screen.
```

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+ *
+ * @see https://github.com/lelylan/simple-oauth2/blob/3.x/API.md#authorizeurlauthorizeoptions--string
+ */
+// TODO: rename this authorize or similar?
```
0xdevalias, line 39:
```suggestion
```

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+
+  if (!checkIfValidVenueId(venueId)) {
+    throw new HttpsError("invalid-argument", "venueId is not a valid venue id");
+  }
+
+  const authConfig = await fetchAuthConfig(venueId);
+
+  const authClient = createOAuth2Client(authConfig);
+
+  // TODO: configure this in cloud config and/or firestore or similar
```
0xdevalias, line 55:
```suggestion
```

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+  const authCodeReturnUri = `${AUTH_ORIGIN}/auth/connect/i4a/handler?venueId=${venueId}`;
+
+  // TODO: configure this in cloud config and/or firestore or similar
+  const scope = "";
```
0xdevalias, line 59: [TODO] configure `scope` in firestore `authConfig`

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+  });
+
+  functions.logger.log("Redirecting to:", redirectUri);
+  res.redirect(redirectUri);
+});
+
+/**
+ * Exchanges a given auth code passed in the 'code' URL query parameter for a Firebase auth token.
+ * The request also needs to specify a 'state' query parameter which will be checked against the 'state' cookie.
+ */
```
0xdevalias, line 73:
```suggestion
/**
 * Exchanges a given auth code passed in the 'code' URL query parameter for an access token,
 * looks up the associated I4A user's details, creates/fetches a firebase account, and then
 * finally returns a custom Firebase auth token that the frontend can use to login as this user.
 */
```

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+
+  const {
+    i4aApiKey,
+    i4aOAuthUserInfoUrl,
+    i4aGetUserMeetingInfoUrl,
+  } = authConfig;
+
+  const authClient = createOAuth2Client(authConfig);
+
+  const { code: authCode } = req.query;
```
0xdevalias, line 95: [TODO] Move this to the top of the function, and error out if it isn't present.

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+
+  functions.logger.log("Received auth code:", authCode);
+  const results = await authClient.authorizationCode.getToken({
+    code: authCode,
+    redirect_uri: "http://localhost:3000/in/foo", // TODO: what do we want to use here?
```
0xdevalias, line 100: [TODO] What should we be using for the redirect URL here? I'm not even sure if/what this is used for in this I4A flow. I think in a normal OAuth flow it might be used to check the redirect URL against a whitelist or similar?

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+    Authorization: `Bearer ${accessToken}`,
+  });
+
+  functions.logger.log("I4A User ID:", i4aUserId);
+
+  if (!i4aUserId) {
+    // TODO: redirect to some kind of '500 error' page?
+    res.redirect("/in/TODO/error");
+    return;
+  }
```
0xdevalias, line 117: [TODO] redirect to some kind of '500 error' page? Or maybe just throw a 500 error? This shouldn't really happen in reality, so it's probably fine to not show the user a perfectly pretty error page if it does.

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+
+  // TODO: configure this in cloud config and/or firestore or similar
+  const meetingIdsToCheck = [131, 136];
+  const eventIdsToCheck = [504, 506, 554];
```
0xdevalias, line 121: [TODO] configure this in firestore `authConfig`?

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+    isRegistered,
+    registeredMeetings,
+    registeredEvents,
+  });
+
+  if (!isRegistered || !emailRaw) {
+    // TODO: redirect to some kind of 'not allowed' page
+    res.redirect("/in/TODO/not-registered");
+    return;
+  }
```
0xdevalias, line 157: [TODO] redirect back to the platform and show some kind of 'not allowed' page, that also prompts them to go and register/buy a ticket. We could potentially also send them straight to the buy a ticket page, but that might be confusing for them?

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+  functions.logger.log("User:", {
+    i4aUserId,
+    userId: userRecord.uid,
+    email: userRecord.email,
+  });
+
+  // Create a custom token for the frontend to use to sign into firebase auth as this user
+  const customToken = await admin.auth().createCustomToken(userRecord.uid, {
+    registeredMeetings,
+  });
```
0xdevalias, line 189: Do we want to include `registeredMeetings` in the custom claims here, is it relevant? If we do keep it, we should also add `registeredEvents`

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+
+  // TODO: redirect back to the main application in a way that we can provide the custom auth token back to it
```
0xdevalias, line 191:
```suggestion
```

functions/auth.js
```diff
@@ -19,3 +30,164 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
+  res.redirect(`${AUTH_ORIGIN}/in/${venueId}?customToken=${customToken}`);
```
0xdevalias, line 192: [TODO] Should we have a different route here used to handle the custom token then redirect back to the venue? That would seem cleaner to me..

Maybe something like `/login/:venueId/:token` or similar?

functions/src/api/auth.js
```diff
@@ -0,0 +1,42 @@
...
+
+const { checkIfValidVenueId } = require("../utils/venue");
+
+const getAuthConfigsCollectionRef = () =>
+  admin.firestore().collection("authConfigs");
+
+const getAuthConfigRef = (venueId) =>
+  getAuthConfigsCollectionRef().doc(venueId);
+
+// TODO: do we need to implement finding the sovereign venue to implement this properly..? Probably should..
```
0xdevalias, line 16: [TODO] do we need to implement finding the sovereign venue to implement this properly..? Probably should..

functions/src/types/auth.js
```diff
@@ -0,0 +1,21 @@
+const yup = require("yup");
+
+// TODO: should we nest these different subsets of keys?
```
0xdevalias, line 3: [TODO] should we nest these different subsets of keys? Probably..

functions/src/utils/auth.js
```diff
@@ -0,0 +1,35 @@
...
+  const credentials = {
+    client: {
+      id: clientId,
+      secret: clientSecret,
+    },
+    auth: {
+      tokenHost,
+      tokenPath,
+      revokePath,
+      // TODO: how can we ensure we add the correct redirect URL here to revokePath? Maybe use a placeholder in the revokePath string or similar?
```
0xdevalias, line 26: [TODO] how can we ensure we add the correct redirect URL here to revokePath? Maybe use a placeholder in the revokePath string or similar?

This looks like:

```
/oauth/logout?redirect_uri=TODO_full_url_here
```

Probably return to the platform at `/v/${venueId}`?

functions/src/utils/venue.js
```diff
@@ -0,0 +1,12 @@
+const getVenueId = (name) => {
+  return name.replace(/\W/g, "").toLowerCase();
+};
+
+const checkIfValidVenueId = (venueId) => {
+  if (typeof venueId !== "string") return false;
+
+  return /[a-z0-9_]{1,250}/.test(venueId);
+};
```
0xdevalias, line 9: [TODO] Extract the `venueId` regex as a const
[TODO] Make a `yup` schema for `VenueId`: https://github.com/jquense/yup#stringmatchesregex-regex-message-string--function-schema

src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -34,6 +34,9 @@ import { venueLandingUrl } from "utils/url";
 import { RoomsForm } from "pages/Admin/Venue/Rooms/RoomsForm";
 import { VenueAdminPage } from "pages/Admin/Venue/VenueAdminPage";
 
+// TODO: Add a route/some kind of related functionality here or similar for handling receiving the custom auth token back from firebase, then using that to sign in.
+//   Alternatively, should we handle this in src/index.ts?
+
```
0xdevalias, line 39: [TODO] Add a route/some kind of related functionality here or similar for handling receiving the custom auth token back from firebase

See other comments in this review for thoughts on what to use for the route here.

src/pages/Account/Login.tsx
```diff
@@ -26,6 +33,39 @@ export const Login: FC<LoginProps> = ({ formType = "initial" }) => {
...
+    currentUrlWithoutCustomToken.searchParams.delete("customToken");
+    // @debt This should work, but it doesn't seem to be properly updating the URL for some reason. Bug with react-router's replace method?
+    replaceUrlUsingRouter(currentUrlWithoutCustomToken.toString());
+
+    // @debt I feel like it would be better to explicitly redirect somewhere once we have logged in.. but our other signIn
+    //  code in LoginForm doesn't seem to..
+  }, [firebase, replaceUrlUsingRouter, search]);
+
+  if (isCustomTokenLoginLoading) return <LoadingPage />;
+
```
0xdevalias, line 68: [TODO] Refactor this into a new component that will be rendered at a specific route, and then redirect to the venue after successful login.

src/utils/venue.ts
```diff
@@ -28,8 +28,11 @@ export const canBeDeleted = (venue: AnyVenue): boolean =>
 export const canHavePlacement = (venue: AnyVenue): boolean =>
   PLAYA_TEMPLATES.includes(venue.template);
 
-export const checkIfValidVenueId = (venueId: string): boolean =>
-  /[a-z0-9_]{1,250}/.test(venueId);
+export const checkIfValidVenueId = (venueId?: string): boolean => {
+  if (typeof venueId !== "string") return false;
+
+  return /[a-z0-9_]{1,250}/.test(venueId);
+};
```
0xdevalias, line 35: [note] This change is mostly to keep this implementation aligned with the backend version, but essentially, `.test` can only take a `string`, and if anything else is passed to it (eg. `undefined`, `null`, etc), then it will seemingly implicitly call `.toString()` on it, which often then causes the value to pass our test here even though it shouldn't (eg. because the string `"undefined"` matches the regex)

0xdevalias (commented):
TODO notes and similar

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/28638e8502fe6cfd5e7b11b43493cb15271c4326

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/28638e8502fe6cfd5e7b11b43493cb15271c4326

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/28638e8502fe6cfd5e7b11b43493cb15271c4326

functions/auth.js
0xdevalias: fixed in 8579155d8e510c0304d266b62135365d95e668f1

functions/auth.js
0xdevalias: fixed in 8579155d8e510c0304d266b62135365d95e668f1

src/pages/Account/Login.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/bba51569720b6b1215b83b06e481f811d677ed52

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/bba51569720b6b1215b83b06e481f811d677ed52

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/646840a005538c122568ce36779385940539f595

functions/.runtimeconfig.json.example
0xdevalias: `authConfig` won't work well though, as it will need to be different between both hosted `staging` and local `staging`. So I think it might need to stay as it is here.

functions/auth.js
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1523#discussion_r653436631 for why this might need to stay as is.

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/93a18c8a51b1c23ece18542ecbe856fcd65e6c67

functions/auth.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/7a3e6cffa5ddb2ed7bf20ad7ddaa983717dcf12d

functions/src/types/auth.js
0xdevalias: Potentially can be OOS of this PR

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/bba51569720b6b1215b83b06e481f811d677ed52

src/pages/Account/Login.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/bba51569720b6b1215b83b06e481f811d677ed52

functions/.runtimeconfig.json.example
0xdevalias: Though if we make it a list of whitelisted origins in `authConfig`, then we could read it from the headers/url params of the request to the function, check that it's on the whitelist, and if it's valid we can use it.

That way we can configure it to work in multiple environments, and we avoid the security issues of blindly trusting the headers/URL params.

src/pages/Account/Login.tsx
```diff
@@ -17,20 +27,47 @@ import "./Login.scss";
...
 }) => {
+  const venueId = venue.id;
+
   const [formToDisplay, setFormToDisplay] = useState(formType);
 
   const { signInWithSAML, hasSamlAuthProviderId } = useSAMLSignIn(
     venue.samlAuthProviderId
   );
-  // It will be extended with addition of new providers
-  const hasAlternativeLogins = hasSamlAuthProviderId;
```
mike-lvov, line 33: Why was it removed? What if there are no authConfig/saml integrations? We don't want to show the `Quick login block`

functions/index.js
```diff
@@ -7,8 +7,6 @@ const functions = require("firebase-functions");
 
 const functionsConfig = functions.config();
 
-console.log("functions.config()", JSON.stringify(functionsConfig, null, 2));
```
0xdevalias, line 10: [note] This was explicitly removed as it logs secrets, which is a bad practice.

src/api/auth.ts
```diff
@@ -20,3 +21,25 @@ export const checkIsEmailWhitelisted = async (data: CheckIsEmailWhitelisted) =>
...
+  await firebase
+    .functions()
+    .httpsCallable("auth-getCustomAuthConfig")({ venueId })
+    .then<CustomAuthConfig>((result) => result.data)
+    .catch((err) => {
+      Bugsnag.notify(err, (event) => {
+        event.addMetadata("context", {
+          location: "api/auth::fetchCustomAuthConfig",
+        });
+      });
```
0xdevalias, line 42:
```suggestion
      Bugsnag.notify(err, (event) => {
        event.addMetadata("context", {
          location: "api/auth::fetchCustomAuthConfig",
          venueId,
        });
      });
```

src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -56,7 +56,12 @@ const AppRouter = () => {
         <Route path="/account/profile" component={Profile} />
         <Route path="/account/questions" component={Questions} />
         <Route path="/account/code-of-conduct" component={CodeOfConduct} />
-        <Route path="/login" component={Login} />
+        <Route
+          path="/login/:venueId/:customToken"
+          component={LoginWithCustomToken}
+        />
+        {/* @debt The /login route doesn't work since we added non-defaulted props to the Login component */}
+        {/*<Route path="/login" component={Login} />*/}
```
0xdevalias, line 64: [note] Pretty sure this is tech debt anyway, not sure it ever really worked currently?

src/pages/Account/Login.tsx
0xdevalias: Look at line 70 where it still exists, but without the comment which is rendundant now.

src/pages/Account/Login.tsx
```diff
@@ -17,20 +27,47 @@ import "./Login.scss";
...
+  }, [venueId]);
+
+  const { customAuthName, customAuthConnectPath } = customAuthConfig ?? {};
+
+  const hasCustomAuthConnect = isDefined(customAuthConnectPath);
+  const signInWithCustomAuth = useCallback(() => {
+    openUrl(
+      `${customAuthConnectPath}?venueId=${venueId}&returnOrigin=${window.location.origin}`
+    );
+  }, [customAuthConnectPath, venueId]);
```
0xdevalias, line 68: [TODO] Wrap this into a custom hook

src/pages/Account/LoginWithCustomToken.tsx
```diff
@@ -0,0 +1,63 @@
...
+      .signInWithCustomToken(customToken)
+      // .catch((error) => {
+      //   // TODO: Handle this with Bugsnag or similar?
+      //   console.error(error);
+      //
+      //   throw error;
+      // });
+      .then(() => {
+        enterVenue(venueId, { customOpenRelativeUrl: replaceUrlUsingRouter });
+      });
```
0xdevalias, line 49: [TODO] move this into api/auth or similar? Add Bugsnag or cleanup commented part.

functions/src/api/auth.js
0xdevalias: Nope, so long as we have an `authConfig` on the main 'entrypoint' for the event, that is all we should need. Usually (if not always) that's the sovereign venue anyway.

functions/src/utils/auth.js
0xdevalias: Did this.. not sure of the commit hash off the top of my head.

functions/src/utils/venue.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1523/commits/b1a0c754fa2a8deae8481d5e1458786e05b42428 + https://github.com/sparkletown/sparkle/pull/1523/commits/660cbc74e22012af2a7d5c753ff092717664a131

functions/auth.js
0xdevalias: Fixed this, not sure of the commit off the top of my head. Basically moved the claims to the user itself rather than the custom token, and added both of them.

mike-lvov (commented):
i skimmed through, will test & if it works approve

functions/auth.js
```diff
@@ -19,3 +30,209 @@ exports.passwordsMatch = (submittedPassword, actualPassword) =>
...
     lowercaseFirstChar(actualPassword.trim());
+
+exports.getCustomAuthConfig = functions.https.onCall(async (data) => {
+  const { venueId } = data;
+
+  assertValidVenueId(venueId, "venueId");
+
+  const { customAuthName, customAuthConnectPath } = await fetchAuthConfig(
+    venueId
+  ).catch(() => ({}));
```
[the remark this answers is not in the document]
0xdevalias: It already logs in `functions/src/api/auth.js` as part of `AuthConfigSchema.validate`

mike-lvov (approved):
Tested locally, it works

mike-lvov:
@0xdevalias 
Since you are reusing the SSO login(just like I do), we should figure out how to handle this. (Is not a problem for now, but will be, if we have both integrations)

mike-lvov:
Also, as an OOS, probably we should add some kind of loader, when a person clicks a button. It takes about 2-3 secs for me to actually get redirected

0xdevalias:
> Since you are reusing the SSO login icon(just like I do), we should figure out how to handle this. (Is not a problem for now, but will be, if we have both integrations)

@mike-lvov Agreed. Future cleanup for sure. Probably can move that to custom auth config or similar I expect. Would also be good to get better labels on them.

0xdevalias:
> Also, as an OOS, probably we should add some kind of loader, when a person clicks a button. It takes about 2-3 secs for me to actually get redirected

@mike-lvov Oh? Interesting.. I wonder how long that will be in prod. Sounds like a reasonable idea though. Would just be changing `signInWithCustomAuth` to `useAsync` rather than `useCallback`. OOS for this PR, but not bad idea for future improvement.

PR #1527, opened 2021-06-09
posterhall: Add a way to quickly preview abstract or download in a separate tab/window (reused)

yarikoptic:
Ideally should sit on top of    https://github.com/sparkletown/sparkle/pull/1375
which would also add a bookmarking capability.  Both bookmark and this page icon should stay side by side.

With this tiny PR visitors could arrange it like
image

and quickly navigate through the posters until you decide to enter one.  We can trust external urls in this case IMHO so I do not think there are security implications.

edit: added also a "download" icon (following @0xdevalias instructions on how to make it harmoneously into a separate new window for this one).  Now just need to figure out how to not do overall `display: flex;` of the entire `PosterPreview` class, since they stack up vertically:
image

If this would be an acceptable solution, we would need 
- [ ] get proper IDs or full URLs in DB 
- [ ] harmonize with #1375 in positioning of bookmark and this 📄 icon (partially addresses #1462)

### DIFF
--- a/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
+++ b/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
@@ -5,7 +5,7 @@ import { useHistory } from "react-router-dom";
 import { PosterPageVenue } from "types/venues";
 
 import { WithId } from "utils/id";
-import { enterVenue } from "utils/url";
+import { enterVenue, externalUrlAdditionalProps } from "utils/url";
 
 import { PosterCategory } from "components/atoms/PosterCategory";
 
@@ -51,6 +51,14 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
       <a href={posterAbsUrl} target="_ohbm-poster-preview">
         📄
       </a>
+      {posterVenue.iframeUrl?.trim() !== "" ? (
+        <a href={posterVenue.iframeUrl} {...externalUrlAdditionalProps}>
+          {" "}
+          ⬇{" "}
+        </a>
+      ) : (
+        <div />
+      )}
       <div onClick={handleEnterVenue}>
         <p className="PosterPreview__title">
           {posterVenue.name}: {title}
--- a/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
+++ b/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
@@ -21,6 +21,12 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   const { title, authorName, categories } = posterVenue.poster ?? {};
 
   const venueId = posterVenue.id;
+  const posterId = venueId.replace("poster", "");
+  // TODO: the posterId doesn't correspond to abs ID. We will need proper URL in DB.
+  // This version is just to demonstrate possibility
+  const posterAbsUrl =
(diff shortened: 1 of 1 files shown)


src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -21,6 +21,12 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   const { title, authorName, categories } = posterVenue.poster ?? {};
 
   const venueId = posterVenue.id;
+  const posterId = venueId.replace("poster", "");
+  // TODO: the posterId doesn't correspond to abs ID. We will need proper URL in DB.
+  // This version is just to demonstrate possibility
+  const posterAbsUrl =
+    "https://ww4.aievolution.com/hbm2101/index.cfm?do=abs.viewAbs&src=ext&abs=" +
+    posterId;
```
mike-lvov, line 29: Doesn't it make more sense to just store it in `poster`? Looks like an unnecessary complex implementation

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -41,14 +47,27 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+        📄
+      </a>
+      {posterVenue.iframeUrl?.trim() !== "" ? (
+        <a href={posterVenue.iframeUrl} {...externalUrlAdditionalProps}>
+          {" "}
+          ⬇{" "}
+        </a>
+      ) : (
+        <div />
+      )}
```
mike-lvov, line 61: move `posterVenue.iframeUrl?.trim() !== ""` to a const. eg. const hasIframeUrl.

then it'll look like

`hasIframeUrl && <a></a>`

mike-lvov (changes requested):

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
yarikoptic: See TODO comment above the const definition :-)

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
yarikoptic: Thanks, will do.

But in general you don't oppose such a solution?

yarikoptic: A hint on how to tune css (see original description on top) to line them up horizontally would be  appreciated.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
mike-lvov: @yarikoptic oops, my bad. Haha, it means that we are on the same page though 😄

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
mike-lvov: No, I don't, except for, I'd still leave the `onClick` handler on the whole poster. And added the `stopPropagation` for both icons' handlers

yarikoptic: note: for a proper url, needs #1534 , but may be that PR can just supersede this one altogether.

yarikoptic:
replaced with #1548 now that #1534 is merged

PR #1531, opened 2021-06-11
Fix tables generator: rows/columns issue

sunny-viktoryia:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/811

There were default values for the number of rows and columns set. Although, we need to define the number of rows based on the capacity of the table.


src/utils/table.ts
```diff
@@ -1,5 +1,7 @@
 import { Table } from "types/Table";
 
+const DEFAULT_COLUMNS_COUNT = 3;
```
yurii-lubynets, line 3: IMHO it's worth keeping this const in `src/settings.ts`. WDYT?

0xdevalias: Makes sense to do so to me.

0xdevalias:
**Note:** It probably makes sense to land this other PR first, then solve the minor merge conflicts once it does:

- https://github.com/sparkletown/sparkle/pull/1821

src/utils/table.ts
sunny-viktoryia: it's already there -- appeared when I merged changes from staging. `DEFAULT_TABLE_COLUMNS`. 🙃 

(but haven't tested the script after that merge)

yurii-lubynets:
LGTM, looks like @sunny-viktoryia has fixed tables view for `Jazzbar` and `ConversationSpace`

Screenshot 2021-07-16 at 13 18 34

src/components/templates/Jazzbar/JazzTab/JazzTab.scss
```diff
@@ -20,7 +20,7 @@
   .music-bar-content {
     display: grid;
 
-    padding: 10px 70px;
+    padding: 10px 70px 40px;
```
yurii-lubynets, line 23: fixed bottom indent for tables on the `Jazzbar` template. They were cropped.

Screenshot 2021-07-16 at 13 26 55

yurii-lubynets:
@sparkletown/engineering feel free to review it

src/utils/table.ts
```diff
@@ -1,4 +1,4 @@
-import { DEFAULT_TABLE_ROWS, DEFAULT_TABLE_COLUMNS } from "settings";
```
mike-lvov, line 1: I dont think we use `DEFAULT_TABLE_ROWS` anymore. If this is the case, delete it from the settings, please

mike-lvov (approved):

src/utils/table.ts
yurii-lubynets: `DEFAULT_TABLE_ROWS` is used in calculation of `DEFAULT_TABLE_CAPACITY` in `src/settings.ts`
and in `src/components/molecules/TablesUserList/TablesUserList.tsx`

mike-lvov: Okay, sounds good

PR #1534, opened 2021-06-11
Feature/adds poster fields

margulies:
adding ohbm-specific poster fields. If this PR should be made directly to branch [env/ohbm], just let me know.


margulies:
Just a note that the styling isn't worked out at this stage — at this stage simply building in the features for the posterpreview cards

yarikoptic:
I think this PR is not only pre-requisite for https://github.com/sparkletown/sparkle/pull/1527 but also might be a replacement altogether! Thank you @margulies .  When I try it locally though -- doesn't work for me: console reports

```
react-dom.development.js:13413 Uncaught Error: Objects are not valid as a React child (found: object with keys {renderMoreInfoUrl}). If you meant to render a collection of children, use an array instead.
```

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -40,13 +52,77 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+  }, [worldUsers, authorName]);
+
+  const recentPosterUsers = useRecentLocationUsers(posterVenue.name);
+
+  const numUsers = recentPosterUsers.recentLocationUsers.length;
+
+  const renderMoreInfoUrl = useMemo(() => {
+    if (!moreInfoUrl) return;
+    return (
+      <a href={moreInfoUrl} target="_blank" rel="noreferrer">
```
yarikoptic, line 72: instead of explicit "rel" probably 
```suggestion
      <a href={moreInfoUrl} target="_blank"  {...externalUrlAdditionalProps}>
```
should be used (would need an import ` from "utils/url"`

mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1534/commits/f17936a2d37b1c1e621b314cd5867c28016f6073

mike-lvov (approved):
Tested locally, polshed a few things

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -27,6 +39,7 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
   });
 
   const { push: openUrlUsingRouter } = useHistory();
+
   const handleEnterVenue = useCallback(
     () => enterVenue(venueId, { customOpenRelativeUrl: openUrlUsingRouter }),
```
mike-lvov, line 44: NOTE: This has potential risks

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -40,13 +53,89 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
+  }`;
+
+  const renderInfoLink = useCallback(
+    (url: string) => (
+      <a href={url} {...externalUrlAdditionalProps}>
+        {url.replace(/(^\w+:|^)\/\//, "")}
+      </a>
+    ),
+    []
+  );
```
0xdevalias, line 80: This seems not an ideal thing to be doing..

src/types/posters.ts
```diff
@@ -1,6 +1,15 @@
...
+  contactEmail?: string;
+  introVideoUrl?: string;
+  thumbnailUrl?: string;
+  keywords?: string[];
+  introduction?: string;
+  moreInfoUrl?: string;
+  moreInfoUrls?: string[];
+  moreInfoUrlTitle?: string;
+  posterId?: string;
 };
```
0xdevalias, line 15: The fields added here seem kind of suboptimal and redundant.. definitely needs to be cleaned up in future.

Issue #1487, opened 2021-06-02
Refactor 'room event' loading in `RoomModal` so that it correctly displays events


0xdevalias:
When on a `PartyMap`, the `RoomModal should show the upcoming events for that room/venue.

It seems that currently, this will only work correctly if the events have the `room` key set to the name of the room, AND the events are all created on the `PartyMap` venue itself.

Since all venues can have events created on them in our firestore structure, it makes sense that this is also reflected in how the logic works here. For that to happen we will need to ensure that we're reading the events from the venue the room relates to and using them here. To maintain backwards compatibility, we also will need to continue reading the events from the current `PartyMap` venue (and filtering them for any events that are related to this room.

## Context / Exploration

_Originally posted by @0xdevalias in [`#external-open-source` on Slack](https://doesitsparkle.slack.com/archives/C01P53GTXKN/p1622598637055100?thread_ts=1622515189.045800&cid=C01P53GTXKN)._

`RoomModal` gets `roomEvents` from `venueEvents`, which is pulled in from `venueEventsSelector` (`src/utils/selectors.ts`) which reads from `state.firestore.ordered.venueEvents`, which is connected/updated by the 2nd query in `useConnectCurrentVenue` (`src/hooks/useConnectCurrentVenue.ts`), which is currently called from:

- `src/components/templates/PartyMap/PartyMap.tsx`
- `src/pages/VenuePage/VenuePage.tsx`

and a few other places that aren’t really relevant for this discussion.

`RoomModal` also passes some of the calculated data to `RoomModalOngoingEvent` (`src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx`)

## Plan

Based on the usage of `venueEventsSelector` / `useConnectCurrentVenue`, as the code currently is, the events would need to be added to the `PartyMap` venue itself to work in this way, which isn’t something we want to do.

This suggestion seems like the correct way to go about solving this properly:

> @margulies: Ideally, I imagine it would be preferable for venue-level events to be grabbed by the parent venue to be shown on the venue modal, as I do see the logic of organizing events within venues (rather than within partymap rooms)

We would most likely use something like `useVenueEvents` (`src/hooks/events.ts`) instead of `venueEventsSelector` / `useConnectCurrentVenue` ; by refactoring `RoomModalContent` to accept the `venueId`, then connecting it there.

[...]

PR #1538, opened 2021-06-13
fix RoomModal to display events created directly on venues + show day of events

margulies:
This is an attempt to fix the issue of events from rooms on a partymap are not showing in the room modal if they are saved on the venue-level:

- fixes https://github.com/sparkletown/sparkle/issues/1487
- partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/783

The initial commit a96b69e49858ae600f5952b8d9f57a6f7697176a appears to grab the events properly to display them in the room modal, but there is a pause between opening the room modal and rendering the events. Any suggestions for how to address this would be appreciated.


margulies:
Moved the `venueEvents` call to `partymap`, which seems to dramatically speed up loading of the events in the room modal, as suggested towards the end of the issue explanation https://github.com/sparkletown/sparkle/issues/1487

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -72,7 +72,7 @@ export const EventModal: React.FC<EventModalProps> = ({
       <div className="EventModal__content">
         <h4 className="EventModal__title">{event.name}</h4>
         <span className="EventModal__subtitle">
-          by {event.host} in{" "}
+          {event.host && "by"} {event.host} in{" "}
```
0xdevalias, line 75: This doesn't look right? Presumably intended to be a ternary or similar?

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,6 +24,12 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
   const { user, profile } = useUser();
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const { relatedVenueIds } = useRelatedVenues({ currentVenueId: venue.id });
+
+  const venueIds = useMemo(() => relatedVenueIds, [relatedVenueIds]);
```
0xdevalias, line 29: `relatedVenueIds` from `useRelatedVenues` is already `useMemo`'d, so you don't need to do it again here.

Also, this `useMemo` as written is a noop that will waste resources, as it is just re-memoing the value whenever the value changes, which has no use/purpose.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,6 +24,12 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
...
+
+  const { events: venueEvents } = useVenueEvents({ venueIds: venueIds });
```
0xdevalias, line 31: `relatedVenueEvents` would probably be a better name here.

0xdevalias: Though this is going to fetch venues for literally every related venue, parents, children, siblings, etc. Which I don't believe we need here at all.

All we need is the events for the venues of the rooms shown on this map, which I believe I described in my original overview/explanation (https://github.com/sparkletown/sparkle/issues/1487) of how to resolve this issue.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -79,10 +84,12 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
...
     if (!venueEvents) return [];
 
-    return venueEvents.filter(
-      (event) =>
-        event.room === room.title && isBefore(Date.now(), eventEndTime(event))
-    );
+    return venueEvents
+      .filter(
+        (event) =>
+          event.room === room.title && isBefore(Date.now(), eventEndTime(event))
```
0xdevalias, line 90: I think there might be a helper in `utils/events` that can replace `isBefore(Date.now(), eventEndTime(event)`

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -79,10 +84,12 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
...
+      )
+      .sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
```
0xdevalias, line 92: I believe we may already have a sorting function we can use here in one of the `utils/*` files.

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -54,9 +54,11 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
...
           <div className="artist-ongoing-container">
             <div className="event-title">{eventToDisplay.name}</div>
-            <div>
-              by <span className="artist-name">{eventToDisplay.host}</span>
-            </div>
+            {eventToDisplay.host && (
+              <div>
+                by <span className="artist-name">{eventToDisplay.host}</span>
+              </div>
+            )}
```
0xdevalias, line 61: Why is this change needed here? Is the `host` not always set?

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -29,6 +29,9 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
   return (
     <div className="schedule-item-container">
       <div className={`time-section ${isCurrentEvent ? "primary" : ""}`}>
+        <div>
+          <b>{labelDayRoomSchedule(event.start_utc_seconds)}</b>
+        </div>
```
0xdevalias, line 34: What's the intent of this change?

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -44,9 +47,11 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
               <b>{event.name}</b>
             </div>
-            <div>
-              by <b>{event.host}</b>
-            </div>
+            {event.host && (
+              <div>
+                by <b>{event.host}</b>
+              </div>
+            )}
```
0xdevalias, line 54: Why is this change needed here? Is the `host` not always set?

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: We don't want to use `<b>` tags in general. Either use `<strong>`, or even better, following a similar pattern as used in `src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx` with the `<span>` and a semanticly BEM named class.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: We don't want to use `<b>` tags in general. Either use `<strong>`, or even better, following a similar pattern as used in `src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx` with the `<span>` and a semanticly BEM named class.

src/utils/time.ts
```diff
@@ -232,3 +234,9 @@ export const getDayInterval = (date: Date | number) => ({
   start: startOfDay(date),
   end: endOfDay(date),
 });
+
+export const labelDayRoomSchedule = (startUtcSeconds: number) => {
+  if (isToday(fromUnixTime(startUtcSeconds))) return "";
+  else if (isTomorrow(fromUnixTime(startUtcSeconds))) return "Tomorrow";
+  else return formatDate(startUtcSeconds);
+};
```
0xdevalias, line 242: What's the intent of this change?

0xdevalias: @sparkletown/engineering I think this may relate to another item we have on our backlog somewhere, would be good to find which one it us, and then properly link that to this PR with closing keywords/etc.

0xdevalias (changes requested):
Quick high level review pass, not necessarily overly thorough.

src/components/organisms/EventModal/EventModal.tsx
margulies: I'll remove all these — just checked and indeed `event.host` is a required parameter

src/components/templates/PartyMap/PartyMap.tsx
margulies: fixed

src/components/templates/PartyMap/PartyMap.tsx
margulies: I've now implemented a filter for venues that have `parentId` set to the current `venue.id` or the venue that is the current `venue`: 
 
`const relatedVenueRoomIds = relatedVenues.filter((roomVenue) => roomVenue.parentId === venue.id || roomVenue.id === venue.id).map((roomVenue) => roomVenue.id);`

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
margulies: reverted

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
margulies: The intent of the change is that currently all events in the room modal are only showing the time irrespective of the day. If the event is now occuring today, this will display 'tomorrow' in the case of tomorrow, and `[Abbrev Month] Date` if the event is occuring after tomorrow.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
margulies: `host` conditional is removed, and `<b>` tags removed as in`RoomModalOngoingEvent.tsx`

src/utils/time.ts
margulies: The intent of this is to provide the day information in the event that an event shown in the room modal isn't today. It it is tomorrow, it will return `Tomorrow`, and if the event is after tomorrow, it will return the abbreviated form of the date.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: `venuePreview.tsx` line 163 appears to have an identical sort. I could pull this into a utils function, but could use some direction as to where to place it.

margulies: replaced with `isEventLiveOrFuture` from `utils/events` 8fbc3f39505799cf9aba6470dac21321f5f7db56

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: replaced with `isEventLiveOrFuture` from `utils/events` 8fbc3f39505799cf9aba6470dac21321f5f7db56

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: I've now replaced this sorting with the new function `sortEventsByStartUtcSeconds` in ef37ccd771f8e359c04cdb48cf443d54b6ae1ffc

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: @sparkletown/engineering make sure we find and link the backlog issue about this from the internal board to this PR with closing keywords

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Sounds good. Can you please also replace the identical location with the function as well.

margulies: great — replaced it in 76b8c3290427682c4d69e62fe42038740b3fd395

0xdevalias: Thanks :)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -79,9 +87,10 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
   const roomEvents = useMemo(() => {
     if (!venueEvents) return [];
 
-    return venueEvents.filter(
-      (event) =>
-        event.room === room.title && isBefore(Date.now(), eventEndTime(event))
+    return sortEventsByStartUtcSeconds(
+      venueEvents.filter(
+        (event) => event.room === room.title && isEventLiveOrFuture(event)
+      )
```
sunny-viktoryia, line 93: if you apply the changes from my another comment (about RoomModal events), we won't need this filtering here,
so we'll need just

```js
const roomEvents = useMemo(() => {
    if (!venueEvents) return [];

    return sortEventsByStartUtcSeconds(venueEvents);
  }, [venueEvents]);
```

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -29,28 +29,25 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-          <b>{formatUtcSeconds(event.start_utc_seconds)}</b>
-        </div>
-        <div>
+        <span className="event-date">
+          {labelDayRoomSchedule(event.start_utc_seconds)}
+        </span>
+        <span className="event-time">
+          {formatUtcSeconds(event.start_utc_seconds)}
+        </span>
+        <span className="event-time">
```
sunny-viktoryia, line 38: we prefer BEM-styled class names, so could we have `schedule-item-container__event-date`, `schedule-item-container__event-time`.

And also would be just awesome, if you could rename `schedule-item-container` -> `ScheduleItem` (and change other class names accordingly: `ScheduleItem__event-date`, `ScheduleItem__event-time`, etc). Our approach is to give the base class the same name as its component has.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -29,28 +29,25 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-              <b>{event.name}</b>
-            </div>
-            <div>
-              by <b>{event.host}</b>
-            </div>
-          </div>
-          <div className="event-description">
+        <div className={`${isCurrentEvent ? "primary" : ""}`}>
+          <span className="event-name">{event.name}</span>
+          by <span className="event-host">{event.host}</span>
```
sunny-viktoryia, line 47: I guess we should do the same check here as we do in the `ScheduleEvent` component? 
(Note: the class name should be updated too, see my another comment)

```suggestion
          {event.host && (
            <>
              by <span className="event-host">{event.host}</span>
            </>
          )}
```

src/components/templates/Playa/VenuePreview.tsx
```diff
@@ -159,11 +160,12 @@ const VenuePreview: React.FC<VenuePreviewProps> = ({
       .then(function (array) {
         const futureEvents = array.docs
           .map((doc) => doc.data())
-          .filter((event) => event.start_utc_seconds > nowSeconds)
-          .sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
+          .filter((event) => event.start_utc_seconds > nowSeconds);
```
sunny-viktoryia, line 163: I guess we might want that to be in the `utils/event.ts` -> `isFutureEvent()`

sunny-viktoryia (commented):
first pass review from me. 🙃

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,6 +24,23 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
   const { user, profile } = useUser();
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const { relatedVenues } = useRelatedVenues({ currentVenueId: venue.id });
+
+  const relatedVenueRoomIds = useMemo(
```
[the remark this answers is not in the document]
margulies: Renamed in 616edc44d8a69c925c93e94ebae816784d0641db

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,6 +24,23 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
...
+    () =>
+      relatedVenues
+        .filter(
+          (relatedVenue) =>
+            relatedVenue.parentId === venue.id || relatedVenue.id === venue.id
```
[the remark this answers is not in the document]
margulies: If we don't include that, any events save at the room-level in the partymap venue won't get pulled in. Just tested to confirm. For maintaining current functionality, seems important to keep that in.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -52,6 +71,7 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
       <RoomModal
         room={selectedRoom}
         venue={venue}
+        venueEvents={relatedVenueEvents}
```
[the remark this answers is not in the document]
margulies: renamed it `selectedRoomEvents` -- hope that works: 6596e1e30a007868ea48ab048ff311ce7883055b

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
margulies: fixed here: 31c51d873d13046cb48881bf72a6fe0f66072998

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
margulies: I'm fine with putting that in, but @0xdevalias suggested I remove that, as `event.host` is a required field. Just let me know whether to implement, and I'll do it

src/components/templates/Playa/VenuePreview.tsx
margulies: I struggled with this a bit due to the `array.docs` part. It's a good idea to move the function to `utils/event.ts`, but seems a bit out of scope for this PR. I only added the `sortEventsByStartUtcSeconds` function to be consistent with the change elsewhere in this PR

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -26,15 +24,17 @@ export interface RoomModalProps {
...
 
 export const RoomModal: React.FC<RoomModalProps> = ({
   onHide,
   room,
   show,
   venue,
+  venueEvents,
 }) => {
-  if (!venue || !room) return null;
+  if (!venue || !room || !venueEvents) return null;
```
mike-lvov, line 37: I don't think we should have this check. Especially, when we check for that down in the subtree

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -27,39 +27,40 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-      <div className={`time-section ${isCurrentEvent ? "primary" : ""}`}>
-        <div>
-          <b>{formatUtcSeconds(event.start_utc_seconds)}</b>
-        </div>
-        <div>
+    <div className="ScheduleItem">
+      <div
+        className={`ScheduleItem__time-section ${
+          isCurrentEvent ? "ScheduleItem__primary" : ""
+        }`}
```
mike-lvov, line 34: classnames

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -27,39 +27,40 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-            <div>
-              <b>{event.name}</b>
-            </div>
-            <div>
-              by <b>{event.host}</b>
-            </div>
-          </div>
-          <div className="event-description">
+      <div className="ScheduleItem__event-section">
+        <div className={`${isCurrentEvent ? "ScheduleItem__primary" : ""}`}>
```
mike-lvov, line 49: classnames

src/components/templates/Playa/VenuePreview.tsx
```diff
@@ -159,11 +160,12 @@ const VenuePreview: React.FC<VenuePreviewProps> = ({
...
 
         // TODO: is this type cast correct?
-        setEventsFuture(futureEvents as VenueEvent[]);
+        setEventsFuture(
+          sortEventsByStartUtcSeconds(futureEvents as VenueEvent[])
```
mike-lvov, line 167: is it a necessity to have `as`?

mike-lvov (changes requested):
A few items/ideas

src/components/templates/Playa/VenuePreview.tsx
mike-lvov: Oh, it's Playa, nwm

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/05e69b6ea9cffce574fb944d32462f8f82570d45

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
mike-lvov: FIxed in https://github.com/sparkletown/sparkle/pull/1538/commits/05e69b6ea9cffce574fb944d32462f8f82570d45

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
mike-lvov: Removed in https://github.com/sparkletown/sparkle/pull/1538/commits/05e69b6ea9cffce574fb944d32462f8f82570d45

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
```diff
@@ -9,30 +9,48 @@
...
     color: $concrete;
     white-space: pre-wrap;
+
+    p {
+      display: inline;
+    }
   }
 
-  .primary {
+  &__primary {
```
sunny-viktoryia, line 28:
```suggestion
  &--primary {
```

sunny-viktoryia (commented):
Skim review. Looks good to merge. 🙃

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -29,7 +29,7 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
   const dispatch = useDispatch();
 
   const schedulePrimaryClasses = classNames({
-    ScheduleItem--primary: isCurrentEvent,
+    "ScheduleItem--primary": isCurrentEvent,
```
margulies, line 32: Is this change correct? It was throwing an error without the quotes

sunny-viktoryia: oops, yeah, didn't account that. Sorry. 🙈

src/components/templates/PartyMap/PartyMap.tsx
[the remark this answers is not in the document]
0xdevalias: > (Also, reverted the renaming from that commit here: 31c51d8)

Was was the naming reverted/etc?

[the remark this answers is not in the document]
margulies: I can reimplement the renaming if you think it's preferable, and also possible to do the sorting there. Just give me the thumbs up and I'll take care of it

src/components/templates/Playa/VenuePreview.tsx
0xdevalias: Given it's a 1 liner to implement, it would be nice to fix it up in this PR. Specifically this part, using `date-fns`:

```typescript
(event) => event.start_utc_seconds > nowSeconds
```

We already have `isEventLiveOrFuture` there though.. which I assume would solve this need, so we should just be using that.

margulies: Thanks for the tip — done: 20f39aee52b31f9839f5a9773aafa974ecdaec0b

src/components/templates/PartyMap/PartyMap.tsx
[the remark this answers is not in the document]
margulies: Here is the commit with the event sorting moved to the `partymap`: ab8bac8fcda0d580b6a6a9d34e8c54397c960b72

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -61,49 +65,40 @@ export const RoomModal: React.FC<RoomModalProps> = ({
...
+        //   and causes rendering issues, we construct a key that, while not guaranteed to be unique,
+        //   is far less likely to clash
+        key={event.id ?? `${event.room}-${event.name}-${index}`}
+        event={event}
+        isCurrentEvent={currentEvent && event.name === currentEvent.name}
+        onRoomEnter={enterRoomWithSound}
+        roomUrl={room.url}
+      />
+    ));
+  }, [enterRoomWithSound, room.url, venueEvents]);
```
0xdevalias, line 101: Since `getCurrentEvent` already processes the array with `.find`, and then we are processing the array again below with `.map`, this is inefficient.

Looking at the definition of `getCurrentEvent`:

- https://github.com/sparkletown/sparkle/blob/staging/src/utils/event.ts#L18-L26

The main bit of work is:

```
(event) =>
      event.start_utc_seconds < currentTimeInUTCSeconds &&
      event.start_utc_seconds + event.duration_minutes > currentTimeInUTCSeconds
```

In english:
- is the event's start before the current time
- does the event end after the current time

`isEventLive` uses `getEventInterval` to get the start/end time of the event, then uses `isWithinInterval` to check if the current time is within that.

- https://date-fns.org/v2.21.3/docs/isWithinInterval

So basically, `isEventLive` does the equivalent to this code.

---

Based on the above, we should change `getCurrentEvent` to be as follows:

```typescript
export const getCurrentEvent = (roomEvents: VenueEvent[]) => roomEvents.find(isEventLive);
```

And then there is no need to use `getCurrentEvent` here, because we can just use `isEventLive` directly within the `.map`:

```suggestion
    return venueEvents.map((event, index: number) => (
      <ScheduleItem
        // @debt Ideally event.id would always be a unique identifier, but our types suggest it
        //   can be undefined. Because we can't use index as a key by itself (as that is unstable
        //   and causes rendering issues, we construct a key that, while not guaranteed to be unique,
        //   is far less likely to clash
        key={event.id ?? `${event.room}-${event.name}-${index}`}
        event={event}
        isCurrentEvent={isEventLive(event)}
        onRoomEnter={enterRoomWithSound}
        roomUrl={room.url}
      />
    ));
  }, [enterRoomWithSound, room.url, venueEvents]);
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -121,11 +116,12 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
...
-        <RoomModalOngoingEvent
-          roomEvents={roomEvents}
-          onRoomEnter={enterRoomWithSound}
-        />
+        {venueEvents && (
+          <RoomModalOngoingEvent
+            roomEvents={venueEvents}
+            onRoomEnter={enterRoomWithSound}
+          />
+        )}
```
0xdevalias, line 124: This previously wasn't needed because we explicitly ensured that there would always be a defined array here. That changed based on @mike-lvov's comments.

Instead of conditionally rendering here, we should continue to allow `RoomModalOngoingEvent` to make its own decisions. It's written to handle an array, even when it's empty, so we should give it that.

I don't really know why @mike-lvov suggested to make the param optional in both of these components, and IMO it's a regression and not something we should be doing. By switching it back to always being defined, we can leave this as it was, which worked well.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
```diff
@@ -9,30 +9,48 @@
...
 
-  .event-description {
+  &__event-description {
+    display: block;
     color: $concrete;
     white-space: pre-wrap;
+
+    p {
+      display: inline;
+    }
```
0xdevalias, line 25: Why was this added here? We don't tend to write styles that target HTML elements like this.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
```diff
@@ -9,30 +9,48 @@
...
+    font-weight: 700;
+  }
+  &__event-date {
+    display: block;
+    font-weight: 700;
+    color: $concrete;
+  }
+  &__event-host {
+    font-weight: 700;
+  }
```
0xdevalias, line 55: Please include whitespace between blocks like this.. I know i've asked for this to be followed in the past. It creates extra work and delay for us when I have to keep asking for it on each new PR.

```suggestion
  }

  &__event-time,
  &__event-name {
    display: block;
    font-weight: 700;
  }

  &__event-date {
    display: block;
    font-weight: 700;
    color: $concrete;
  }

  &__event-host {
    font-weight: 700;
  }
```

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -26,40 +27,47 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
+        <span className="ScheduleItem__event-time">
+          {formatUtcSeconds(event.start_utc_seconds)}
+        </span>
+        <span className="ScheduleItem__event-time">
           {formatUtcSeconds(
             event.start_utc_seconds + event.duration_minutes * 60
           )}
-        </div>
+        </span>
       </div>
```
0xdevalias, line 54: It would be better if we could use `eventStartTime` / `eventEndTime` from `utils/event` here rather than doing the calculations all over again.

`formatUtcSeconds` is defined as follows:

```typescript
export function formatUtcSeconds(utcSeconds?: number | null) {
  return utcSeconds ? format(fromUnixTime(utcSeconds), "p") : "(unknown)";
}
```

The main part of this is the `format` with pattern `p` (Long localized time eg. `12:00 AM`):

- https://date-fns.org/v2.21.3/docs/format

We already have `formatTimestampToDisplayHoursMinutes` which uses the pattern `h:mm aa`

- Hour [1-12]: `h`, eg. `1, 2, ..., 11, 12`
- Minute: `mm`, eg. `00, 01, ..., 59`
- AM, PM: `a..aa`, eg. `AM, PM`

This currently accepts `timestamp` (`number`) as it's param, but since it's just passing directly to `format`, the type can be changed to accept `Date | number`.

That would then allow the return value from `eventStartTime` / `eventEndTime` to be passed directly to it.

---

`formatTimestampToDisplayHoursMinutes` is currently only used in `src/components/atoms/ChatMessageInfo/ChatMessageInfo.tsx` and would be fine if it used the `p` format.

0xdevalias: Refactoring, it would end up looking like (this is directly copy/pastable):

```typescript
/**
 * Format UTC seconds as a string representing the time in long localized time format (eg. 12:00 AM).
 *
 * @example
 *   formatTimestampToDisplayHoursMinutes(1618509600)
 *   // '9:35 AM'
 *
 * @param date
 *
 * @see https://date-fns.org/docs/format
 */
export const formatTimestampToDisplayHoursMinutes = (date: Date | number) =>
  format(date, "p");
```

And then we can use the following here:

```typescript
 <span className="ScheduleItem__event-time">
  {formatTimestampToDisplayHoursMinutes(eventStartTime(event))}
</span>
<span className="ScheduleItem__event-time">
  {formatTimestampToDisplayHoursMinutes(eventEndTime(event))}
</span>
```

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -26,40 +27,47 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-          </div>
-          <div className="event-description">
+      <div className="ScheduleItem__event-section">
+        <div className={schedulePrimaryClasses}>
+          <span className="ScheduleItem__event-name">{event.name}</span>
+          by <span className="ScheduleItem__event-host">{event.host}</span>
+          <span className="ScheduleItem__event-description">
             <RenderMarkdown text={event.description} />
-          </div>
+          </span>
```
0xdevalias, line 61: Why did we change this to use `span` rather than `div` here?

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -26,40 +27,47 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
-          <div className="entry-room-button">
+          <div className="ScheduleItem__entry-room-button">
             <a
               onMouseOver={() => dispatch(retainAttendance(true))}
               onMouseOut={() => dispatch(retainAttendance(false))}
-              className="btn room-entry-button"
+              className="btn ScheduleItem__room-entry-button"
               onClick={onRoomEnter}
-              id={`enter-room-from-schedule-event-${event}`}
+              id={`ScheduleItem__enter-room-from-schedule-event-${event}`}
```
0xdevalias, line 70: We shouldn't need an `id` on this at all I don't believe. React doesn't really make use of them at all.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -26,40 +27,47 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
...
               href={roomUrl}
               target="_blank"
               rel="noopener noreferrer"
```
0xdevalias, line 73: This should be using `externalUrlAdditionalProps`:
```suggestion
              {...externalUrlAdditionalProps}
```

src/components/templates/Playa/VenuePreview.tsx
```diff
@@ -159,11 +160,14 @@ const VenuePreview: React.FC<VenuePreviewProps> = ({
...
-          .sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
+          .filter((event) => isEventLiveOrFuture(event as VenueEvent));
 
         // TODO: is this type cast correct?
-        setEventsFuture(futureEvents as VenueEvent[]);
+        setEventsFuture(
+          sortEventsByStartUtcSeconds(
+            futureEvents as WithVenueId<WithId<VenueEvent>>[]
+          )
+        );
```
0xdevalias, line 170: [note] This is playa (legacy code to be deleted), so we don't mind that this doesn't fit our patterns/type safety.

src/utils/event.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from "date-fns";
 
 import { VenueEvent } from "types/venues";
+import { WithId, WithVenueId } from "./id";
```
0xdevalias, line 11:
```suggestion

import { WithId, WithVenueId } from "./id";
```

src/utils/event.ts
```diff
@@ -85,3 +86,7 @@ export const getEventStatus = (event: VenueEvent) => {
     return `Starts ${formatUtcSecondsRelativeToNow(event.start_utc_seconds)}`;
   }
 };
+
+export const sortEventsByStartUtcSeconds = (
+  events: WithVenueId<WithId<VenueEvent>>[]
+) => events.sort((a, b) => a.start_utc_seconds - b.start_utc_seconds);
```
0xdevalias, line 92: Similar to `chatSort` in `src/utils/chat.ts`, we should just be making the sorting function here, not the whole `.sort` part of it, which is redundant. eg.

```suggestion
export const eventSort = (a: VenueEvent, b: VenueEvent) => a.start_utc_seconds - b.start_utc_seconds
```

So that it can then be used in any `.sort` as:

```typescript
fooEvents.sort(eventSort)
```

0xdevalias (changes requested):
Comments/questions/suggestions/changes.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: This is the related issue https://github.com/sparkletown/internal-sparkle-issues/issues/783

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: Leave it without IMO. I also called out another location to remove it from in my most recent review as well.

0xdevalias:
@margulies Is this PR configured to allow us to push changes directly to it? It doesn't seem to be working as it normally would/should on my side.

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,10 +26,38 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
   const { user, profile } = useUser();
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const { relatedVenues } = useRelatedVenues({ currentVenueId: venue.id });
+
+  const childVenueIds = useMemo(
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/8103d5be13631708fd932e152c95e10a937f4c10

src/components/templates/PartyMap/PartyMap.tsx
```diff
@@ -22,10 +26,38 @@ export const PartyMap: React.FC<PartyMapProps> = ({ venue }) => {
...
+      relatedVenues
+        .filter(
+          (relatedVenue) =>
+            relatedVenue.parentId === venue.id || relatedVenue.id === venue.id
+        )
+        .map((childVenue) => childVenue.id),
+    [relatedVenues, venue]
+  );
+
+  const { events: relatedVenueEvents } = useVenueEvents({
```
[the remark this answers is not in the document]
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/8103d5be13631708fd932e152c95e10a937f4c10

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/7eba21f8e2fa302d88d03ab53d4809bd15fc2449 + https://github.com/sparkletown/sparkle/pull/1538/commits/ff2350207191e67a797f79326ad10ed16dea66bf + https://github.com/sparkletown/sparkle/pull/1538/commits/f901ee13b588fafddff3985c612f818ae78a34a2 + https://github.com/sparkletown/sparkle/pull/1538/commits/3597fb1cb5fa80d044f763613e9b568055366e8c

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/171b2904631fd39b954ae9727b9b4ef0be5af6b0

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/6fc557eeb19867a64d8bbfa9c463c96e3c54531a

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
```diff
@@ -93,7 +93,9 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
       <div className={containerClasses} onClick={onEventBoxClick}>
         <div className="ScheduleEvent__info">
           <div className="ScheduleEvent__title">{event.name}</div>
-          <div className="ScheduleEvent__host">by {event.host}</div>
+          {event.host && (
+            <div className="ScheduleEvent__host">by {event.host}</div>
+          )}
```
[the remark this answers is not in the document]
margulies: agreed - reverted: b0972b161baf80c738715618494944d1dc6ccde9

src/components/organisms/EventModal/EventModal.scss
```diff
@@ -29,6 +29,7 @@ $EventModal--padding-top: 40px;
     opacity: 0.8;
     padding-top: $spacing--xl;
     padding-bottom: $spacing--xl;
+    white-space: pre-wrap;
```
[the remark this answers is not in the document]
margulies: I added this to allow for proper handling of line breaks in the event modal. I struggled to get line breaks to work correctly without this, but if there's another way, just let me know

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: In `src/components/organisms/NavBarSchedule/NavBarSchedule.tsx` we have `formatDayLabel` defined inline, which uses:

```typescript
const formatDayLabel = (day: Date | number) => {
  if (isScheduleTimeshifted) {
    return format(day, "E, LLL d");
  } else {
    return isToday(day) ? "Today" : format(day, "E");
  }
};
```

`labelDayRoomSchedule` in `utils/time` in this PR is defined as:

```typescript
export const labelDayRoomSchedule = (startUtcSeconds: number) => {
  const date = fromUnixTime(startUtcSeconds);

  if (isToday(date)) return "";
  if (isTomorrow(date)) return "Tomorrow";

  return formatDate(startUtcSeconds);
};
```

`formatDate` in `utils/time` is defined as:

```typescript
export function formatDate(utcSeconds: number) {
  return format(fromUnixTime(utcSeconds), "MMM do");
}
```

---

Both of these are sort of different variations on `formatRelative`, but only for dates rather than times:

- https://date-fns.org/v2.22.1/docs/formatRelative

Looking at `format`:

- https://date-fns.org/v2.22.1/docs/format

For these patterns:

- `E, LLL d`, eg. `Mon, Jun 7`
  - `E`: Day of week (formatting), eg. `Mon, Tue, Wed, ..., Sun`
  - `LLL`: Month (stand-alone), eg. `Jan, Feb, ..., Dec`
  - `d`: Day of month, eg. `1, 2, ..., 31`
- `MMM do`, eg. `Jan 7th`
  - `MMM`: Month (formatting), eg. `Jan, Feb, ..., Dec`
  - `do`: Day of month, eg. `1st, 2nd, ..., 31st`

Based on that:
- `LLL` and `MMM` appear to be equivalent.
- We could potentially replace `d` with `do` or vice versa quite easily

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
margulies: This was the only way I was able to find to present an addition blank line occuring after each event description in the event modal, while also allowing line breaks. If you have any hints for better ways to present line breaks like that, happy to implement differently.

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: - refactor formatTimestampToDisplayHoursMinutes to use format pattern '… …49a8bec
- refactor ScheduleItem to use eventStartTime / eventEndTime with forma… … 626cf09
- refactor formatDate in utils/time to accept dateOrTimestamp 7b7b63a
- replace labelDayRoomSchedule with formatDateRelativeToNow b58a244
- refactor formatDayLabel in NavBarSchedule to use formatDateRelativeToNow a825845

margulies: Thanks! Done: 9bd0adfd98424ff18e366d672c9a30b86838ca37

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/78054d17ebf3e5222fe89f80fc682d83b39b96e3

src/utils/event.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/c87e9c8094ad8ceb7a56aaa668ecf94c895afba2

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
margulies: The `span` after 'by' was necessary to prevent a line break. I've reverted the others. dc3f55a5c5b9c8dc6fd06fbf87da70238b8ac19a

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
margulies: removed: ba91d4acfad3fa35d02bc654f28ced6da605a514

src/utils/event.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1538/commits/6e240c48936c15f5ee1103f2ac1e7fb4f6f1ec67

src/components/organisms/EventModal/EventModal.scss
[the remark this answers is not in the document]
0xdevalias: The way that was proposed in the past was using the markdown rendering as the solution to line breaks as well as formatting within various descriptions across the platform. I believe that was one of the original reasons for bringing it in in the first place. Is there a reason we can't just use that here?

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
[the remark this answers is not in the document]
0xdevalias: Thanks :)

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
0xdevalias: I asked a question about why we changed it, I didn't say to change it. PRs are a conversation, and questions are used to get the required context to make proper decisions. I didn't say reverting it was the best choice. I asked why it was done.

So asking again, why was it changed in the first place?

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.scss
0xdevalias: I thought I already replied to this.. but can't find it now. Was line breaks and similar not one of the original reasons we implemented markdown rendering? Why can't that be used here?

0xdevalias: Though a further question here is why do we need to allow line breaks in the first place?

0xdevalias (commented):
[CRO] Have just gone through this again, and all of the resolved comments from past reviews seem fine to me.

There is the following outstanding from my last review still:

- https://github.com/sparkletown/sparkle/pull/1538#discussion_r653206914
- https://github.com/sparkletown/sparkle/pull/1538#discussion_r653208959

And given I wrote the fixes for a lot of my last review suggestions, it would be good to have someone else look over them as a sanity check.

mike-lvov (approved):
LGTM + tested locally in OHBM env. Added max-height + scroll for long lists

PR #1542, opened 2021-06-14
initial fix for overlapping events

margulies:
This is a bit hacky, but seems to roughly achieve the goal of spreading events vertically if they're overlapping within the room/venue-row of the dropdown schedule navbar.

### DIFF
--- a/src/components/molecules/ScheduleEvent/ScheduleEvent.scss
+++ b/src/components/molecules/ScheduleEvent/ScheduleEvent.scss
@@ -4,7 +4,7 @@ $ScheduleEvent--margin-top: 0.25rem;
 $ScheduleEvent--height: 3.75rem;
 $ScheduleEvent--padding: 0 0.25rem 0 0.75rem;
 $ScheduleEvent--border-radius: 18px;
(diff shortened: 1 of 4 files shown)


0xdevalias:
@margulies Under what circumstances will a single room/venue have 2 things happening in it at the same time to need this change?

margulies:
> @margulies Under what circumstances will a single room/venue have 2 things happening in it at the same time to need this change?

I think there are two situations: in the env/ohbm case, one concrete example is there are 4 category tracks running simultaneously in the posterhall at the same time. This could be addressed through a combined event that described all four together, but in some ways that makes it trickier for an attendee to bookmark their personalized schedules.

Personal schedule: The more critical (and perhaps general) case is where an attendee has bookmarked a bunch of events that are of interest and happen to be overlapping, and they can't see them in their personal schedule. Again, this is primarily a problem related to aspects such as the posterhall, but could also be relevant for planning a personal schedule in advance of deciding what to attend. 

I'm sure there are better ways to address this issue, and it's not urgent, but I thought to give it a whirl to see if this was a manageable fix.

mike-lvov:
Seems to be irrelevant

PR #1543, opened 2021-06-14
Allow camera and microphone for embeddable

whoaitsaimz:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/662


whoaitsaimz:
This really only needs to go to one particular client environment -- not sure how to make that happen?

whoaitsaimz:
I think I am pushing to a feature branch. Would love to have this deployed for a specific Sparkle-hosted client environment. Not totally sure how to do so.

0xdevalias:
> I think I am pushing to a feature branch. Would love to have this deployed for a specific Sparkle-hosted client environment. Not totally sure how to do so.

This is how to change it:

- https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

Though depending on what branch you forked from originally, and whether there are any unrelated changes, you may also need to rebase your branch to clean it up:

- https://docs.github.com/en/github/getting-started-with-github/using-git/about-git-rebase
- https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase
- https://git-scm.com/book/en/v2/Git-Branching-Rebasing

src/components/templates/Embeddable/Embeddable.tsx
```diff
@@ -22,7 +22,7 @@ export const Embeddable: React.FC<EmbeddableProps> = ({ venue }) => {
         src={iframeUrl}
         className="embeddable__iframe"
         style={iframeStyles}
-        allow={IFRAME_ALLOW}
+        allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture; fullscreen; camera; microphone;"
```
0xdevalias, line 25: Instead of hardcoding directly into the component like this, it would be better to extract a new `IFRAME_ALLOW_EMBEDDABLE` or similar in `src/settings` and use that here.

0xdevalias, line 25: Instead of hardcoding directly into the component like this, it would be better to extract a new `IFRAME_ALLOW_EMBEDDABLE` or similar in `src/settings` and use that here.

0xdevalias (changes requested):
Suggestions.

See also the questions/comments raised in https://github.com/sparkletown/internal-sparkle-issues/issues/662#issuecomment-862775122

src/components/templates/Embeddable/Embeddable.tsx
0xdevalias: Do we know that these settings are all required here? How were they chosen?

whoaitsaimz: I took the existing IFRAME_ALLOW settings and added the camera; microphone. I imagine that autoplay and fullscreen would be good to keep but cannot personally speak to why accelerometer, encrypted-media, gyroscope, or picture-in-picture would be needed

whoaitsaimz: @0xdevalias I updated per your requests

whoaitsaimz:
Closing to do this in a cleaner way (correct base branch of env/github) with https://github.com/sparkletown/sparkle/pull/1643

PR #1544, opened 2021-06-14
fix covert room type classes in MapRoom to not scale on hover

margulies:
Minor bug fix to return prior functionality to covert room type. The recent update to this code in https://github.com/sparkletown/sparkle/pull/1427 seems to have allowed unclickable and mapframe room types to scale on hover. This PR reverts that change.

### DIFF
--- a/src/components/templates/PartyMap/components/Map/MapRoom.tsx
+++ b/src/components/templates/PartyMap/components/Map/MapRoom.tsx
@@ -47,8 +47,9 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   }, [dispatch]);
 
   const containerClasses = classNames("maproom", {
-    "maproom--covert maproom--unclickable": isUnclickable,
-    "maproom--covert maproom--iframe": isMapFrame,
+    "maproom--covert": isCovertRoom,
+    "maproom--unclickable": isUnclickable,
+    "maproom--iframe": isMapFrame,
     "maproom--always-show-label":
       !isCovertRoom &&
       (venue.roomVisibility === RoomVisibility.nameCount ||
--- a/src/components/templates/PartyMap/components/Map/MapRoom.tsx
+++ b/src/components/templates/PartyMap/components/Map/MapRoom.tsx
@@ -47,8 +47,8 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   }, [dispatch]);
 
   const containerClasses = classNames("maproom", {
-    "maproom--covert--unclickable": isUnclickable,
-    "maproom--covert--iframe": isMapFrame,
+    "maproom--covert maproom--unclickable": isUnclickable,
+    "maproom--covert maproom--iframe": isMapFrame,
     "maproom--always-show-label":
       !isCovertRoom &&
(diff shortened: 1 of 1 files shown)


yurii-lubynets (approved):

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -47,8 +47,8 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   }, [dispatch]);
 
   const containerClasses = classNames("maproom", {
-    "maproom--covert--unclickable": isUnclickable,
-    "maproom--covert--iframe": isMapFrame,
+    "maproom--covert maproom--unclickable": isUnclickable,
+    "maproom--covert maproom--iframe": isMapFrame,
```
sunny-viktoryia, line 51:
```suggestion
    "maproom--covert": isCovertRoom,
    "maproom--unclickable": isUnclickable,
    "maproom--iframe": isMapFrame,
```

margulies: Thanks!

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -47,8 +47,9 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   }, [dispatch]);
 
   const containerClasses = classNames("maproom", {
-    "maproom--covert--unclickable": isUnclickable,
-    "maproom--covert--iframe": isMapFrame,
+    "maproom--covert": isCovertRoom,
+    "maproom--unclickable": isUnclickable,
+    "maproom--iframe": isMapFrame,
```
0xdevalias, line 52: > The recent update to this code seems to have allowed unclickable and mapframe room types to scale on hover. This PR reverts that change.

@margulies Which 'recent update' changed this, and which is the original PR? It's best practice to link the PRs where changes are made so we have continuity + can click through to them to understand what changed/why it's an issue/etc.

In this case, I don't see any changes to any styling files? So why does this fix the issue? Do those classes that have been removed here still exist and need to be cleaned up? Etc.

Please provide more thorough information/details when making PRs to make it easier for the reviewer to know what is going on/why, rather than needing to spend extra time back and forthing to ask those questions for details that could have been provided up front.

margulies: Good point -- sorry about that. This is the relevant PR: https://github.com/sparkletown/sparkle/pull/1427

mike-lvov: @0xdevalias 
The classes were inside in the initial PR

like that:
`.maproom {
 &--covert {
    &--unclickable
 }
}`

They were moved outside to the same level, but were not changed here

mike-lvov (approved):
LGTM

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: Makes sense. Thanks.

PR #1545, opened 2021-06-14
show the number of participants on the poster preview

denisdimitrov:
Shows the number of participants on the poster preview. If no users are inside the poster, the icon is not displayed.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/810


yarikoptic:
I believe #1534 has done the same, but may be styling could be improved. See #1548 displaying current state (as as of that PR)

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -104,7 +103,13 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
         )}
         {hasUsers && (
-          <div className="PosterPreview__visiting">{userCountText}</div>
+          <div className="PosterPreview__users">
+            <FontAwesomeIcon
+              className="PosterPreview__users-icon"
+              icon={faUserAlt}
+            />
+            {userCount}
+          </div>
```
yarikoptic, line 112: FWIW, here is how it looks like  as of 212773beda38578f3cbdb06acac8339ee8898359 (I with there were some "release tags" in this repo)

image

denisdimitrov: Yeah this is because of the poster Id being rendered. It's not in the design of the issue and I'm not sure how to proceed. When I get to it, I will ask Sofi about it.

denisdimitrov: This is now aligned to be visually correct and not outside the container.

mike-lvov:

PR #1546, opened 2021-06-14
add SAML SSO login flow

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/767

I wonder if it should be merged into staging given its abstract implementation. My only concern is that `Google identity` is turned off for everything else, except `env/github`

More info in https://docs.google.com/document/d/1Ai7eN8U-epxnaPDl5NPJZKOdHrtkVWBLmddlt5xr99Q/edit

NOTE:
To test it, set the APP ENVS to
`REACT_APP_AUTH_DOMAIN=sparkle-github.firebaseapp.com`
`REACT_APP_SSO_PROVIDER_ID=saml.okta`

NOTE: They were set in CircleCI for GitHub


0xdevalias:
> I wonder if it should be merged into staging given its abstract implementation. My only concern is that `Google identity` is turned off for everything else, except `env/github`

@mike-lvov I haven't looked at the code here yet, but in general I would say we should be merging it into `staging`. If it's coded defensively/etc enough then it should deploy fine/'sit there' fine, and just be 'disabled' when the appropriate env things aren't set up for it.

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -94,6 +103,25 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
...
+
+          <div className="LoginForm__alternative-logins">
+            {hasSSOProvider && (
+              <img
+                className="LoginForm__sso-login"
+                src={SSO_LOGIN_ICON}
+                onClick={signInSSO}
+                alt="SSO login"
+              />
+            )}
```
0xdevalias, line 118: Instead of just showing an icon here, it might be better for us to show some text as well?

0xdevalias: Have you looked at https://firebase.google.com/docs/reference/js/firebase.auth.Auth?authuser=2#fetchsigninmethodsforemail at all? Not sure if it would be relevant here.. but figured I would mention it in case.

src/hooks/useSSO.ts
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from "react";
+import firebase from "firebase/app";
+
+import { SSO_PROVIDER_ID } from "secrets";
```
0xdevalias, line 4: [NB] May be considered OOS of this PR, but instead of needing to configure this at the environment secrets level, we should configure it on the sovereign `venue` in firebase (and eventually it would be on the `world`)

`LoginForm` already seems to have `venue` in scope anyway, so would be easy to pass the relevant details through here. That way we would also be able to easily configure different SSO methods for different venues (and would take less env setup/code changes to do so, etc)

src/hooks/useSSO.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const useSSO = () => {
+  const hasSSOProvider = SSO_PROVIDER_ID !== undefined;
+
+  const signInSSO = useCallback(() => {
+    if (!SSO_PROVIDER_ID) return;
+
+    const SSOProvider = new firebase.auth.SAMLAuthProvider(SSO_PROVIDER_ID);
```
0xdevalias, line 12: This hook is written generically as 'SSO', yet this implementation is specifically SAML related. There are other forms of SSO that can be implemented. What're the thoughts/plan/etc here?

src/hooks/useSSO.ts
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from "react";
+import firebase from "firebase/app";
+
+import { SSO_PROVIDER_ID } from "secrets";
+
+export const useSSO = () => {
+  const hasSSOProvider = SSO_PROVIDER_ID !== undefined;
+
+  const signInSSO = useCallback(() => {
```
0xdevalias, line 9: [NB]

Maybe something like this? Naming flows weirdly currently:
```suggestion
  const signInWithSSO = useCallback(() => {
```

src/secrets.js
```diff
@@ -12,6 +13,7 @@ export const MIXPANEL_PROJECT_TOKEN =
 
 export const WS_RELAY_URL = process.env.REACT_APP_WS_RELAY_URL;
 export const IS_BURN = process.env.REACT_APP_IS_BURN;
+export const SSO_PROVIDER_ID = process.env.REACT_APP_SSO_PROVIDER_ID;
```
0xdevalias, line 16: See earlier comments about configuring this via `venue` config rather than ENV.

src/settings.ts
```diff
@@ -49,6 +50,7 @@ export const SPARKLE_ICON = "/sparkle-icon.png";
 export const DEFAULT_MAP_BACKGROUND = "/maps/Sparkle_Field_Background.jpg";
 export const DEFAULT_VENUE_BANNER = "/assets/Sparkle_Banner_Default.jpg";
 export const DEFAULT_VENUE_LOGO = "/assets/Sparkle_SquareLogo_Default.jpg";
+export const SSO_LOGIN_ICON = "/assets/SSO_Login_Logo.png";
```
0xdevalias, line 53: See comments elsewhere about importing this icon from fontawesome or similar.

If we do choose to keep doing it this way, then it shouldn't be loaded as a magic string like this, but should moved to somewhere in `src/assets/*`, and then loaded via `import` so that it is properly bundled/checked/etc.

src/settings.ts
```diff
@@ -553,6 +555,7 @@ export const FIREBASE_CONFIG = {
   measurementId: MEASUREMENT_ID,
   projectId: PROJECT_ID,
   storageBucket: BUCKET_URL,
+  authDomain: AUTH_DOMAIN,
```
0xdevalias, line 558: The rest of the keys here seem to be sorted alphabetically. Would be good to maintain that.

src/secrets.js
```diff
@@ -1,6 +1,7 @@
 export const PROJECT_ID = process.env.REACT_APP_PROJECT_ID;
 export const API_KEY = process.env.REACT_APP_API_KEY;
 export const APP_ID = process.env.REACT_APP_APP_ID;
+export const AUTH_DOMAIN = process.env.REACT_APP_AUTH_DOMAIN;
```
0xdevalias, line 4: [OOS] We should really prioritize some work to use the 'automatic' setup option and remove the need for us to configure/pass through all of this stuff from ENV vars (with an option to fall back to/override using the ENV vars still probably)

See https://firebase.google.com/docs/hosting/reserved-urls?authuser=2#sdk_auto-configuration

src/hooks/useSSO.ts
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from "react";
+import firebase from "firebase/app";
+
+import { SSO_PROVIDER_ID } from "secrets";
+
+export const useSSO = () => {
+  const hasSSOProvider = SSO_PROVIDER_ID !== undefined;
```
0xdevalias, line 7: We should pass in `SSO_PROVIDER_ID` as a prop here, and read it off the `venue` config itself rather than a const like this.

0xdevalias (commented):
[CRO]

---

Why are we using `public/assets/SSO_Login_Logo.png` rather than something from our typical icon import libs?

Eg.

- https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=lock
- https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=key
- etc

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -45,6 +50,10 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
     reValidateMode: "onChange",
   });
 
+  const { signInSSO, hasSSOProvider } = useSSO();
+  // It will be extended with addition of new providers
+  const hasAlternativeLogins = hasSSOProvider;
```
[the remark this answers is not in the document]
mike-lvov: Yep, correct

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -94,6 +103,25 @@ const LoginForm: React.FunctionComponent<PropsType> = ({
 
       <h2>Log in</h2>
 
+      {hasAlternativeLogins && (
+        <>
+          <p>Quick log in with</p>
```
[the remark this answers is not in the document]
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1546/commits/416b11d804e90a1eca8f44969646cc4c0006949b

src/hooks/useSSO.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1546/commits/cd6ed2a394579a1a482db8a751630924672ad528

src/secrets.js
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1546/commits/cd6ed2a394579a1a482db8a751630924672ad528

src/hooks/useSSO.ts
mike-lvov: > LoginForm already seems to have venue in scope anyway, so would be easy to pass the relevant details through here. That way we would also be able to easily configure different SSO methods for different venues (and would take less env setup/code changes to do so, etc)

Done in https://github.com/sparkletown/sparkle/pull/1546/commits/cd6ed2a394579a1a482db8a751630924672ad528

mike-lvov: > [NB] May be considered OOS of this PR, but instead of needing to configure this at the environment secrets level, we should configure it on the sovereign venue in firebase (and eventually it would be on the world)

👍 OOS

src/settings.ts
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1546/commits/c5f8848eb58b5921d9ff5785eba01ee6b0bb2278

src/settings.ts
mike-lvov: > See comments elsewhere about importing this icon from fontawesome or similar.

There were no needed icon

> If we do choose to keep doing it this way, then it shouldn't be loaded as a magic string like this, but should moved to somewhere in src/assets/*, and then loaded via import so that it is properly bundled/checked/etc.


Done in https://github.com/sparkletown/sparkle/pull/1546/commits/c5f8848eb58b5921d9ff5785eba01ee6b0bb2278

src/hooks/useSSO.ts
mike-lvov: Changed the name in https://github.com/sparkletown/sparkle/pull/1546/commits/62de340d0e1eda6f11434e9fa870896e6698c0f7

src/hooks/useSSO.ts
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1546/commits/62de340d0e1eda6f11434e9fa870896e6698c0f7

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
mike-lvov: Not sure what it's about 

> Have you looked at https://firebase.google.com/docs/reference/js/firebase.auth.Auth?authuser=2#fetchsigninmethodsforemail at all? Not sure if it would be relevant here.. but figured I would mention it in case.

whoaitsaimz: @mike-lvov we got some notes on sign-in flow / descriptive text requests based on what we shared with client yesterday, I'll tag you in main issue if we need to add something here

whoaitsaimz: Also adding @malaikaaiyar to support some textual additions / revisions to this flow

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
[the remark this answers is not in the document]
0xdevalias: Cool cool, makes sense.

src/hooks/useSSO.ts
0xdevalias: As per our processes, if you're calling something OOS then you need to make a backlog issue for it + link that from the thread so that it's not lost. Please follow our proccesses.

src/pages/Account/Login.tsx
```diff
@@ -1,17 +1,28 @@
...
+import { useSSO } from "hooks/useSSO";
+
 import LoginForm from "components/organisms/AuthenticationModal/LoginForm";
 import PasswordResetForm from "components/organisms/AuthenticationModal/PasswordResetForm";
 import RegisterForm from "components/organisms/AuthenticationModal/RegisterForm";
 import { InitialForm } from "components/organisms/AuthenticationModal/InitialForm";
-import React, { FC, useState } from "react";
+
 import "./Account.scss";
+import "./Login.scss";
```
0xdevalias, line 15: If that `Account.scss` file isn't being used anywhere else (which ideally it shouldn't be) then we should renaming it to `Login.scss` and combine it with the new styles introduced here.

At an absolute minimum, if we can't do that, we should add an `@debt` comment to do so in future.

src/hooks/useSSO.ts
0xdevalias: Looking at the code changes it looks like you did this anyway? So no clue why you're calling it OOS?

src/settings.ts
0xdevalias: > There were no needed icon

😕 I showed a bunch of icons that would have worked equally as well for this in those other comments. And from the design mockup in https://github.com/sparkletown/internal-sparkle-issues/issues/767#issuecomment-861814262 it looks like using the GitHub logo would be appropriate)

src/pages/Account/Login.tsx
mike-lvov: image

mike-lvov: Comment added in https://github.com/sparkletown/sparkle/pull/1546/commits/4fe8c301fc88ac30c81c63af80e6ab29fcbe538b

src/pages/Account/Login.tsx
```diff
@@ -1,17 +1,37 @@
...
 }
 
-export const Login: FC<LoginProps> = ({ formType = "initial" }) => {
+export const Login: React.FC<LoginProps> = ({
+  formType = "initial",
+  venue,
+}) => {
   const [formToDisplay, setFormToDisplay] = useState(formType);
 
+  const { signInWithSAML, hasSamlAuthProviderId } = useSAMLSignIn(
```
denisdimitrov, line 29: You can remove the const at line 33 and convert it directly here
```suggestion
  const { signInWithSAML, hasSamlAuthProviderId: hasAlternativeLogins } = useSAMLSignIn(
```

src/pages/Account/Login.tsx
```diff
@@ -32,6 +52,22 @@ export const Login: FC<LoginProps> = ({ formType = "initial" }) => {
         <img src="/sparkle-header.png" alt="" width="100%" />
       </div>
       <div className="auth-form-container">
+        {hasAlternativeLogins && (
+          <div className="Login__login-box">
+            <span>Quick log in with</span>
+
+            <div className="Login__alternative-logins">
+              {hasSamlAuthProviderId && (
```
denisdimitrov, line 60: You are already checking for `hasAlternativeLogins` above and they have the same value, looks like this check here is redundant.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -172,15 +172,19 @@ const VenuePage: React.FC = () => {
...
+    return <LoadingPage />;
   }
 
   if (venueRequestStatus && !venue) {
     return <>This venue does not exist</>;
   }
 
-  if (!venue || !venueId || !profile) {
+  if (!user) {
+    return <Login formType="initial" venue={venue} />;
```
denisdimitrov, line 184: `initial` is the default value, you can remove the prop `formType`

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -172,15 +172,19 @@ const VenuePage: React.FC = () => {
...
+  }
+
+  if (!profile) {
```
denisdimitrov, line 187: There are 2 separate checks that render `<LoadingPage/>` why is that?

One at line 175, and one here.

src/hooks/useSAMLSignIn.ts
```diff
@@ -0,0 +1,31 @@
+import { useCallback } from "react";
+import Bugsnag from "@bugsnag/js";
+import firebase from "firebase/app";
+
+export const useSAMLSignIn = (samlAuthProviderId?: string) => {
+  const hasSamlAuthProviderId = samlAuthProviderId !== undefined;
```
denisdimitrov, line 6: If the id is empty string this will be true, is this the desired flow? Should be do something like this:
```suggestion
const hasSamlAuthProviderId = !!samlAuthProviderId
```

denisdimitrov (changes requested):

src/pages/Account/Login.tsx
mike-lvov: It's an explicit choice to have it both `hasSamlAuthProviderId` and `hasAlternativeLogins`.

The reason for that is that hasAlternativeLogins will have more things in it(Google auth, facebook auth, etc...)

src/pages/Account/Login.tsx
mike-lvov: As per https://github.com/sparkletown/sparkle/pull/1546#discussion_r652831141, it is the intended flow

src/pages/VenuePage/VenuePage.tsx
mike-lvov: I don't think it's true

image

mike-lvov: So, I moved the `!venue || !venueId` check before the `Login page`, because I need `venue` to be loaded before we display it. However, if you add `!profile` before `!user`, it will cause an infinite loading

src/hooks/useSAMLSignIn.ts
mike-lvov: I guess it's the standart way we check if something is not defined

image

src/pages/VenuePage/VenuePage.tsx
mike-lvov: Removed in https://github.com/sparkletown/sparkle/pull/1546/commits/3eaee127c14e82bae162ec5a42eeba2fdfcb888a

src/hooks/useSAMLSignIn.ts
denisdimitrov: Fair enough

src/pages/VenuePage/VenuePage.tsx
denisdimitrov: Okey, sounds reasonable.

However, looks like because of the first statement, the second one will never render, yes?

```tsx
  if (!venue || !venueId) {
    return <LoadingPage />;
  }

  if (venueRequestStatus && !venue) {
    return <>This venue does not exist</>;
  }
```

Maybe you should swap them or add some additional condition to the first one.

mike-lvov: 👍 Fixed in https://github.com/sparkletown/sparkle/pull/1546/commits/882adbf34503dedf8195a6d981d431cbd2ca71a9

denisdimitrov (approved):
LGTM!

src/settings.ts
mike-lvov: > I showed a bunch of icons that would have worked equally as well for this in those other comments. And from the design mockup in sparkletown/internal-sparkle-issues#767 (comment) it looks like using the GitHub logo would be appropriate)

This is a placeholder icon, I'll remove it as soon as we have final designs

0xdevalias: If it’s a placeholder then even more reason to use something from the icon line and not add yet another file to forever clutter and bloat our git history.

src/hooks/useSAMLSignIn.ts
0xdevalias: If it’s an empty string it will just get an error during attempting to signin which will be handled in standard ways

PR #1547, opened 2021-06-15
BF: place only the "main body" of poster preview to onClick

yarikoptic:
Otherwise clicking on the URL in the PosterPreview__header would also
trigger onClick for the entire div and thus enter the poster

Followup to #1534


yarikoptic:
current state of env/ohbm already does it (after merging the  bookmark PR) so no need (for ohbm at least)

PR #1548, opened 2021-06-15
[env/ohbm] add download icon (⬇) alongside with the bookmark

yarikoptic:
Replacement for #1527.
Sits on top of #1547 for a fix.

But the effect of it fixes the incorrect (off the div) rendering of the number of current visitors. compare
image
as of staging to
image
as of this PR

edit 1: if someone could help to style it better (icons should be larger but somehow my tune ups to font-size had no effect, didn't inverstigate) correctly would be appreciated


yarikoptic:
PR is redone on top of env/ohbm and only download icon is added now since link to the abstract is already there!

yarikoptic:
@mike-lvov may be you would have a moment to have a look at this now quite tiny (now that it is on top of merged #1375) PR?

mike-lvov:
Isn't relevant anymore

PR #1549, opened 2021-06-15
Time and userId based randomized usePosters when no searchQuery - one of the proposed solutions to PR #1452

goran-peoski-work:
See discussion at https://github.com/sparkletown/sparkle/pull/1452


src/hooks/posters.ts
```diff
@@ -87,15 +96,19 @@ export const usePosters = (posterHallId: string) => {
   );
 
   const searchedPosterVenues = useMemo(() => {
+    // Local PRNG: does not affect Math.random.
+    const random = seedrandom(`${userId}${timeSlice}`);
```
yarikoptic, line 100: I do not quite understand why bother with `timeSlice`.  I would say `userId` should be sufficient and actually the desired effect. IMHO it would only be odd (if not annoying) for users to have their view of posters to change all of a sudden whenever a new hour comes.

yarikoptic (changes requested):
THANK YOU @goran-peoski-work for working on this one!!!  FWIW --  I have tested locally as of e24c6f417c419e4a727333dc334ebb54c5cdd1aa on OHBM venue and it worked awesomely well for its original intent!  I left one comment on the unclear need for time slicing to participate in seeding -- IMHO should not be done. (I am not a sparkle person, hopefully it is ok to leave my "Review")

src/hooks/posters.ts
goran-peoski-work: Simply:
* remove the `timeSlice` number from the string, and you get random posters on each request (maybe even noticeable "weird" behavior within the same request).
* put it at 1 minute, then going back and forward for that one minute, you get the same posters, the next minute you get a fresh batch.

Basically, it's a tweaking tool for as how long a cache of posters gets reused for the same `userId` before a new one gets generated. You can use the `POSTERS_TIME_GRANULARITY` constant for it, set it to  every 5 minutes, every 5 hours, remove it, make it configurable through `.env` file etc.

goran-peoski-work:
I've enabled Auto-merge, so do take care if approving the PR.

src/hooks/posters.ts
yarikoptic: > * remove the `timeSlice` number from the string, and you get random posters on each request (maybe even noticeable "weird" behavior within the same request).

why is that so? they should then be seeded by `userId`... 

<details>
<summary>with this "strip away" patch they seems to be stable and nice and seems to be specific to me </summary> 

```diff
diff --git a/src/hooks/posters.ts b/src/hooks/posters.ts
index 3bdb3abab..951179cb1 100644
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -45,10 +45,7 @@ export const usePosterVenues = (posterHallId: string) => {
   );
 };
 
-const POSTERS_TIME_GRANULARITY = 1000 * 60 * 60; // one hour in ms
-
 export const usePosters = (posterHallId: string) => {
-  const timeSlice = Math.floor(new Date().getTime() / POSTERS_TIME_GRANULARITY);
   const { userId } = useUser();
   const { posterVenues, isPostersLoaded } = usePosterVenues(posterHallId);
 
@@ -97,7 +94,7 @@ export const usePosters = (posterHallId: string) => {
 
   const searchedPosterVenues = useMemo(() => {
     // Local PRNG: does not affect Math.random.
-    const random = seedrandom(`${userId}${timeSlice}`);
+    const random = seedrandom(`${userId}`);
     const normalizedSearchQuery = searchQuery.trim();
 
     if (!normalizedSearchQuery)
@@ -126,7 +123,7 @@ export const usePosters = (posterHallId: string) => {
         }),
       })
       .map((fuseResult) => fuseResult.item);
-  }, [searchQuery, fuseVenues, filteredPosterVenues, userId, timeSlice]);
+  }, [searchQuery, fuseVenues, filteredPosterVenues, userId]);
 
   const displayedPosterVenues = useMemo(
     () => searchedPosterVenues.slice(0, displayedPostersCount),

```
</details>

> * put it at 1 minute, then going back and forward for that one minute, you get the same posters, the next minute you get a fresh batch.

I put it at 1 second, and what happens - it randomizes 2 (or 3?) times and settles.  Then if I click to enter some poster - it again randomizes few times and enters the poster I had clicked.

Anyways -- thinking about it, I guess it could be indeed a nice feature to promote more posters exploration!  We just would need to announce it as a feature so whenever someone sees them getting shuffled (at hour boundary), they don't get surprised.

goran-peoski-work: As I said it can be 5 minutes or 5 hours, it can very well  be 5 days, so there are ways to certainly make it more or less noticeable. But, since I'm not familiar to the original requirement, I can accept any changes, even removal of the time factor from the seed.

yarikoptic (approved):
Works for me! ;)

src/hooks/posters.ts
yarikoptic: FWIW - all good with me on this extra-feature.  Unfortunately I am nobody here so can't mark it resolved AFAIK.

0xdevalias:
> @yarikoptic @goran-peoski-work @mike-lvov Feedback on the proper React context/provider solution implementation + further ways to improve it to also keep the scroll state in https://github.com/sparkletown/sparkle/pull/1452#discussion_r652529136 + https://github.com/sparkletown/sparkle/pull/1452#discussion_r652537451
> 
> My recommendation is still that the react context/provider solution is a far superior choice to the hacky seed/shuffle/etc stuff proposed in https://github.com/sparkletown/sparkle/pull/1549, and given it's practically implemented already now, and better, I don't really see why we wouldn't use it.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1452#issuecomment-862223934_

src/hooks/posters.ts
0xdevalias: >  I guess it could be indeed a nice feature to promote more posters exploration! We just would need to announce it as a feature so whenever someone sees them getting shuffled (at hour boundary), they don't get surprised.

If I am viewing a list of posters, deeply scrolled down, find something interesting, and am about to click on it and the hour ticks, i'm gonna be mighty... frustrated by this 'feature'

0xdevalias (changes requested):
This PR contains unrelated commits, I assume it was created off an incorrect branch and needs to be properly rebased and cleaned up.

---

The addition of the time in the randomisation is a bad idea:

> >  I guess it could be indeed a nice feature to promote more posters exploration! We just would need to announce it as a feature so whenever someone sees them getting shuffled (at hour boundary), they don't get surprised.
> 
> If I am viewing a list of posters, deeply scrolled down, find something interesting, and am about to click on it and the hour ticks, i'm gonna be mighty... frustrated by this 'feature'.
> 
> Seems like a terrible idea to me tbh.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1549#discussion_r652545103_

src/hooks/posters.ts
goran-peoski-work: First
> announce it as a feature so whenever someone sees them
then
> i'm gonna be mighty... frustrated by this 'feature'.

Sure, that might happen. Fun question might be _what's the percent of users that will fall into this category?_. I'm guessing there might be a fun graph with cutoff interval (1h, 1d, 1w) vs count of frustrated users... but that's not really the issue.

The issue is, this PR is just one of the proposed solutions to https://github.com/sparkletown/sparkle/issues/1449, one that I got working first, not one I would generally promote as best way going forward, especially if (as originally) with base to staging.

It's just easier to compare pros and cons on working code.

goran-peoski-work:
> This PR contains unrelated commits, I assume it was created off an incorrect branch and needs to be properly rebased and cleaned up.

@0xdevalias what is the correct branch / are the unrelated commits?

goran-peoski-work:
Created new branch https://github.com/sparkletown/sparkle/tree/pr/random-posters-seedrandom--1452 out of `env/ohbm` and applied the patch with the seed randomization.

0xdevalias:
@goran-peoski-work Can we please close this PR in favour of https://github.com/sparkletown/sparkle/pull/1557

goran-peoski-work:
superseded by https://github.com/sparkletown/sparkle/pull/1557

PR #1555, opened 2021-06-16
fix event modal popping up when clicking bookmark

margulies:
Fixes event modal popup bug when clicking on bookmark.

### DIFF
--- a/src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
+++ b/src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
@@ -63,6 +63,8 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
     containerCssVars
   );
 
+  const ScheduleEventBookmarkClass = "ScheduleEvent__bookmark";
+
   const bookmarkEvent: MouseEventHandler<HTMLDivElement> = useCallback(() => {
     if (!userId || !event.id) return;
 
@@ -77,15 +79,24 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
     hide: hideEventModal,
   } = useShowHide();
 
+  const onEventBoxClick = useCallback(
+    (e) => {
+      if (e.target.closest(`.${ScheduleEventBookmarkClass}`)) return;
+
+      showEventModal();
+    },
+    [showEventModal]
+  );
+
   return (
     <>
-      <div className={containerClasses} onClick={showEventModal}>
+      <div className={containerClasses} onClick={onEventBoxClick}>
         <div className="ScheduleEvent__info">
           <div className="ScheduleEvent__title">{event.name}</div>
           <div className="ScheduleEvent__host">by {event.host}</div>
         </div>
 
-        <div className="ScheduleEvent__bookmark" onClick={bookmarkEvent}>
+        <div className={ScheduleEventBookmarkClass} onClick={bookmarkEvent}>
           <FontAwesomeIcon
             icon={event.isSaved ? solidBookmark : regularBookmark}
           />
--- a/src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
+++ b/src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
@@ -26,6 +26,8 @@ import { calcStartPosition } from "components/molecules/Schedule/utils";
 
 import "./ScheduleEvent.scss";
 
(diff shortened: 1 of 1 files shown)


src/hooks/useShowHide.ts
```diff
@@ -1,11 +1,23 @@
...
-  const show = useCallback(() => {
-    setShown(true);
-  }, []);
+  const show = useCallback(
+    (e?) => {
+      if (targetClosest && e?.target.closest(targetClosest)) return;
+      setShown(true);
+    },
+    [targetClosest]
+  );
```
sunny-viktoryia, line 20: It looks a bit hacky to my eyes. 
I see your implemented fix this way: we have some `targetClosest` exception. When the event was fired within that `targetClosest` selector we ignore the modal opening. Am I correct?

Let's consider another case. Let's say we want to make clickable not only the bookmark icon but the host in the event box in the schedule too.
image
It means we have 2 clickable child elements. And we don't want the event modal to open when we click on them. And in fact, we can have more complex containers with more than 2 clickable child elements. Your implement version of the hook can't handle that case.

You could suggest transforming `targetClosest` to the array of selectors. But I wouldn't support that. I suppose would be better to keep things as isolated as possible. The name of the hook implies a straightforward logic of having a boolean status: "is shown" / "is hidden". Adding feature-specific things would one day overcomplicate the hook and increase the possibility of a bug.

Although I like your idea of building a generic solution, currently I can't quickly come up with a better solution than explicitly add the exceptions to the functions where it's needed. 

@0xdevalias @mike-lvov would love to hear your thoughts. 🙃

sunny-viktoryia (commented):
Posting here a comment to discuss the idea of the fix.

Thanks for your contributions, @margulies!

src/hooks/useShowHide.ts
margulies: All sounds totally reasonable. Just wanted to mention that the current implemtation should be able to handle an array of clickable child elements. For example, here: https://github.com/sparkletown/sparkle/blob/4e9c93cfe4b8d5f6e31a26fac8cb0c0f6fe8b040/src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx#L50

By passing a string array to `targetClosest` similar to the example linked above, it should achieve a similar functionality. 

More broadly though, you're absolutely right that a more elegent general solution would be possible here.

margulies: Oh! Sorry, one other question: the real goal of this is to deal with the bug in the event modal. If there's a simple way to add in the specific exception in `ScheduleEvent.tsx` without needing to copy over all the code from `useShowHide`, that's seems totally sufficient for the time being. Happy to do that if you could point me in the simplest direction to achieve that goal.

sunny-viktoryia: I'd do something like this in the `ScheduleEvent.tsx`: 
```ts
  const onEventBoxClick = useCallback(
    (e) => {
      if (e.target.closest(".ScheduleEvent__bookmark")) return;

      showEventModal();
    },
    [showEventModal]
  );
  
  // some logic 
  
      <div className={containerClasses} onClick={onEventBoxClick}>
  // ...
```

margulies: That was so much easier! Thanks! Cleaned up and much simpler now... really appreciate it the help

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
```diff
@@ -63,6 +63,8 @@ export const ScheduleEvent: React.FC<ScheduleEventProps> = ({
     containerCssVars
   );
 
+  const ScheduleEventBookmarkClass = "ScheduleEvent__bookmark";
```
sunny-viktoryia, line 66: can we move this static const out of the `ScheduleEvent` component not to create it on every render?
(would be better to see it right after the imports before the `ScheduleEventProps` interface)

sunny-viktoryia (approved):
Looks good to me (would be perfect to address my comment). 🙃

Thanks for providing the fix, @margulies.

src/components/molecules/ScheduleEvent/ScheduleEvent.tsx
margulies: Done! 34b0660cab61d257db73fc7da72b9313eaa75589
Thanks! (and good to know how to creating a const on every render. It seems obvious now that you mention it, but hadn't realized it before.)

PR #1556, opened 2021-06-16
[env/ohbm] include thumbnail in posterpreview

margulies:
This is the final styling issue for the posterpreview cards on the posterhall template. It incorporates the poster thumbnails, and in the absence of an active link, grabs a default image whose location is specified in `settings.ts`.

It's not perfect -- as it would be nice to lock the position of the thumbnail irrespective of the surrounding text. But I think it's better than nothing.

Two pages for testing: 
- https://ohbm.sparkle.space/in/posterhall
- https://ohbm.sparkle.space/in/exhibithall


src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
```diff
@@ -60,12 +60,17 @@ $active-poster-shadows: 0 10px 20px 0 rgba($primary, 0.56);
...
   }
 
   &__categories {
     display: flex;
     flex-wrap: wrap;
+    margin: $spacing--sm 0;
+
+    -webkit-line-clamp: 2;
+    line-clamp: 2;
+    overflow: hidden;
```
0xdevalias, line 73: What do we need this for?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -127,6 +131,15 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
...
       <p className="PosterPreview__title">{title}</p>
 
+      <img
+        src={thumbnailUrl}
+        alt={title}
+        onError={(e) => {
+          e.currentTarget.src = DEFAULT_POSTER_THUMBNAIL_IMG;
+        }}
+        style={{ borderRadius: "15px" }}
+      />
```
0xdevalias, line 141: Having this function inline like this will cause re-renders. It should be `useCallback`'d.

Also, we shouldn't be using inline styles like this (and it will also cause re-renders). It should be made as a new class in the SCSS following BEM naming patterns.

src/settings.ts
```diff
@@ -601,6 +601,9 @@ export const POSTERPAGE_MORE_INFO_URL_TITLE = "Full abstract";
 
 export const POSTERHALL_POSTER_IS_LIVE_TEXT = "Presenter is online";
 
+export const DEFAULT_POSTER_THUMBNAIL_IMG =
+  "https://firebasestorage.googleapis.com/v0/b/sparkle-ohbm.appspot.com/o/assets%2Fposter_thumbnail_default.png?alt=media&token=fff064bc-0c62-4a4d-8f5e-0bd25dd06680";
```
0xdevalias, line 605: [note] This is ok since it's targeting `env/ohbm`, but if we wanted to merge this into our main code branch we would obviously need to use something different here.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: Why are we using an `img` here? If we used a `div` or similar then we wouldn't need a hacky workaround with the default image, which is basically just a solid block of colour:

image

We can use the CSS `background-image` property on a `div`, and also set the `background-color`, which is used as a fallback when there is no image/it fails to render/etc.

I believe there should be examples of this already in the codebase. They'll use `useCss` and CSS `--properties` to pass the URL of the image from the JS into SCSS. This will then be combined with the `classNames` helper to apply the variables + specific class for the element to the `className` in the JSX. There are standard naming patterns for these, so we should follow suit. The properties tend to be set on the main container of component.

0xdevalias (changes requested):
Comments/changes for how to improve on this using our existing patterns. 

@mike-lvov These changes I suggested will also remove the `env/ohbm` specific hackiness from this I believe, so if we wanted to merge this into our main `staging` branch, I believe it would be fine to do so.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.scss
margulies: There are cases where some posters in the exhibitors hall have a dozen categories — this prevents them from overflowing the posterpreview card, but still allows them to be searchable.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
```diff
@@ -127,6 +146,14 @@ export const PosterPreview: React.FC<PosterPreviewProps> = ({
 
       <p className="PosterPreview__title">{title}</p>
 
+      <div className="PosterPreview__thumbnail">
+        {renderImg ? (
+          renderThumbnailUrl
+        ) : (
+          <rect className="PosterPreview__rect" />
```
margulies, line 153: I know this is a different approach, but I had difficulty getting rid of the missing image icon. Here I'm simply switching with a rectangle. Are there issues I should be aware of with this approach?

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: `useCss` + `classNames` pattern: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/ScheduleVenueDescription/ScheduleVenueDescription.tsx#L26-L33

Using a `div`: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/ScheduleVenueDescription/ScheduleVenueDescription.tsx#L41

With SCSS styling: https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/ScheduleVenueDescription/ScheduleVenueDescription.scss#L32

Just seems to need a `background-colour` added as well.

src/components/templates/PosterHall/components/PosterPreview/PosterPreview.tsx
0xdevalias: Not sure why you would hav difficulty getting rid of missing image icon if you followed what I laid out in https://github.com/sparkletown/sparkle/pull/1556#discussion_r654096370 as there would be no missing image icon anymore.

I did a shift+control+f and searched for `useCss` as I suggested previously, and instantly located an exact example, which I have now spelled out the aspects of more specifically in https://github.com/sparkletown/sparkle/pull/1556#discussion_r654553111

sunny-viktoryia (approved):
I've fixed the code. Looks good to me now.

Going to merge to OHBM only, if we want to merge to other envs, we can do that later, I guess. Since we've not agreed on that yet. 

FYI: @0xdevalias @mike-lvov

PR #1557, opened 2021-06-16
refactor usePosters using context/provider pattern to display a random selection of posters when no searchQuery

goran-peoski-work:
Fresh branch out of `env/ohbm` with the Context/Provider randomization of `usePosters` based on https://github.com/sparkletown/sparkle/pull/1452#discussion_r652529136

closes https://github.com/sparkletown/sparkle/issues/1449


goran-peoski-work:
I feel like I'm missing something from the original requirement. Can't remember if shuffle is supposed to be used and under which conditions.

Also, still haven't look at the scroll to position described in https://github.com/sparkletown/sparkle/pull/1452#discussion_r652537451

yarikoptic:
I have tested this branch in isolation (no merge of now conflicting env/ohbm, or any randomization attempt) and for its purpose of returning to the filtered by search list of posters -- it worked wonderfully!  Thank you @goran-peoski-work !
I have tested in both http://localhost:3000/in/exhibithall and http://localhost:3000/in/posterhall to the same effect.

yarikoptic:
FWIW, also test-driven in the d9187068485cae5046ded804f3a07463c8fa0226 state -- as before reverting the shuffle.  It is better IMHO than no randomization at all for the purpose of the OHBM meeting.  

But I think the combination of this PR + #1549 (userId + time based seeded randomization) would be nearly a perfect solution for the purpose of OHBM.  In other words: any of these two PRs would significantly improve UX, and in combination - it would be a huge boost.  If neither of the two is merged -- would be really really sad. This PR in particular is nearly a "must" since otherwise, without even ability to open a poster room in a new window or preview a PDF (unless #1548 is merged) -- it would be really hard to go through any searched up collection of the posters.

goran-peoski-work:
The shuffle add/revert was me checking out possible combination of both, which at first glance it looked like they don't mix as well in the state they are ATM. 

But, for the purpose of
>But I think the combination of this PR + #1549 (userId + time based seeded randomization) would be nearly a perfect solution for the purpose of OHBM.

I'm interested in any randomization idea before implementing, since care should probably be taken to extract the randomization out of the deeply nested `usePosters` and used higher up where the provider state is set if we want to achieve some navigation consistency within each randomization instance.

src/hooks/posters.tsx
```diff
@@ -162,5 +182,31 @@ export const usePosters = (posterHallId: string) => {
...
     setBookmarkedFilter,
-  };
+  } as PostersContextState;
+};
+
+export const PostersContext = createContext<PostersContextState | undefined>(
+  undefined
+);
+
+type PostersProviderProps = Readonly<{ venueId: string }>;
```
0xdevalias, line 192:
```suggestion
export interface PostersProviderProps {
  venueId: string
}
```

src/hooks/posters.tsx
```diff
@@ -162,5 +182,31 @@ export const usePosters = (posterHallId: string) => {
     setSearchInputValue,
     setLiveFilter,
     setBookmarkedFilter,
-  };
+  } as PostersContextState;
```
0xdevalias, line 185: Using `as` isn't type safe, and shouldn't be needed.

src/hooks/posters.tsx
```diff
@@ -118,7 +138,7 @@ export const usePosters = (posterHallId: string) => {
       normalizedSearchQuery
     );
 
-    if (tokenisedSearchQuery.length === 0) return filteredPosterVenues;
+    if (!tokenisedSearchQuery.length) return filteredPosterVenues;
```
0xdevalias, line 141: We were explicit in using `===` here on purpose. Please revert.

src/hooks/posters.tsx
```diff
@@ -44,7 +51,20 @@ export const usePosterVenues = (posterHallId: string) => {
   );
 };
 
-export const usePosters = (posterHallId: string) => {
+interface PostersContextState {
```
0xdevalias, line 54:
```suggestion
export interface PostersContextState {
```

src/hooks/posters.tsx
```diff
@@ -10,6 +16,7 @@ import { tokeniseStringWithQuotesBySpaces } from "utils/text";
 import { isLoaded, useFirestoreConnect } from "./useFirestoreConnect";
 import { useSelector } from "./useSelector";
 import { useDebounceSearch } from "./useDebounceSearch";
+import { WithId } from "../utils/id";
```
0xdevalias, line 19: Please ensure imports use their canonical paths, not `../` relative paths.

This should also be sorted with the other `utils/*`
```suggestion
import { WithId } from "utils/id";
```

src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -4,7 +4,7 @@ import { GenericVenue } from "types/venues";
 
 import { WithId } from "utils/id";
 
-import { usePosters } from "hooks/posters";
+import { usePostersContext } from "../../../hooks/posters";
```
0xdevalias, line 7:
```suggestion
import { usePostersContext } from "hooks/posters";
```

src/hooks/posters.tsx
```diff
@@ -1,5 +1,11 @@
-import { useState, useMemo, useCallback } from "react";
-import { VenueTemplate } from "types/venues";
+import React, {
+  useState,
+  useMemo,
+  useCallback,
+  useContext,
+  createContext,
+} from "react";
+import { VenueTemplate, PosterPageVenue } from "types/venues";
```
0xdevalias, line 8: `types/*` imports after `settings`, before `utils/*`

src/hooks/posters.tsx
```diff
@@ -162,5 +182,31 @@ export const usePosters = (posterHallId: string) => {
...
+export const PostersContext = createContext<PostersContextState | undefined>(
+  undefined
+);
+
+type PostersProviderProps = Readonly<{ venueId: string }>;
+export const PostersProvider: React.FC<PostersProviderProps> = ({
+  venueId,
+  children,
+}) => (
+  <PostersContext.Provider value={usePosters(venueId)}>
```
0xdevalias, line 197: I don't know that it's even valid to use hooks like this? At the very least, we shouldn't be. Please call the hook outside of the JSX, set the value to a `const`, then pass that through to the `value` prop.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -28,6 +28,7 @@ import { UserProfileModal } from "components/organisms/UserProfileModal";
 import { WithNavigationBar } from "components/organisms/WithNavigationBar";
 
 import { AnnouncementMessage } from "components/molecules/AnnouncementMessage";
+import { PostersProvider } from "../../hooks/posters";
```
0xdevalias, line 31: This should be sorted with other `hooks/*` imports.

```suggestion
import { PostersProvider } from "hooks/posters";
```

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -152,7 +153,13 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
           hasBackButton={hasBackButton}
         >
           <AnnouncementMessage message={venue.bannerMessage} />
-          {template}
+          {[VenueTemplate.posterhall, VenueTemplate.posterpage].includes(
```
0xdevalias, line 156: It might be worth extracting `[VenueTemplate.posterhall, VenueTemplate.posterpage]` into a `const` outside of the JSX, potentially in `src/settings` where we have similar things defined that are also used in this `.includes()` kind of way.

0xdevalias (changes requested):
Quick pass of review alongside the comments made on slack in https://doesitsparkle.slack.com/archives/C01JHLMSFND/p1623920660202300?thread_ts=1623915612.199400&cid=C01JHLMSFND

src/hooks/posters.tsx
goran-peoski-work: IDE artefact I had already dealt with. Took a while to catch it.

src/hooks/posters.tsx
goran-peoski-work: Should have had a comment for the reasoning  above it I guess. What's the purpose?

src/hooks/posters.tsx
goran-peoski-work: Any reasoning, comment, code style docs that can be referred to as to this kind of rules not seem arbitrary?

yarikoptic:
> I'm interested in any randomization idea before implementing, since care should probably be taken to extract the randomization out of the deeply nested `usePosters` and used higher up where the provider state is set if we want to achieve some navigation consistency within each randomization instance.

+1 on that. Also note that not all venues might want randomization.  BUT in the scope of ohbm, AFAIK it is posterhall and exhibithall which use it and probably even in the case of exhibithall randomization would be somewhat desired to give all exhibitors a "fair chance" to be "the first" (now it is alphabetically sorted, and I was "smart" to have our booth name start with `Ce` ;))

src/pages/VenuePage/TemplateWrapper.tsx
goran-peoski-work: Had the same thought, opted to keep it hardcoded inside that file  since I saw it more connected to the large hardcoded constants `switch`board above.
I could move it or leave it in place, looks the same to me at this point, but should we refactor it into smaller chunks, `if`s, factory functions etc., the settings array would be reusable in at least 2 places, surely it can be moved elsewhere.

goran-peoski-work:
OK, the story so far:
- original solution idea from https://github.com/sparkletown/sparkle/pull/1452#discussion_r640250266
- using context provider (https://reactjs.org/docs/context.html#contextprovider) with a hook https://reactjs.org/docs/hooks-reference.html#usecontext
- said `PostersProvider` replaces `usePosters` and is tucked inside `TemplateWrapper` wrapping up `VenueTemplate.posterhall`, and `VenueTemplate.posterpage` venue templates
- navigating to the parent component for the first time having empty `searchQuery` will shuffle the poster venues and store it in the provider own state
- navigating to and from e.g. from a poster hall to a poster, then clicking back to the hall, should persist the same randomization
- with no automatic scrolling to the poster ATM
- to use different kind of randomization, one can change the snippet 
  ```javascript
  const shuffledPosterVenues = useMemo(() => shuffle(loadedPosterVenues), [
    loadedPosterVenues,
  ]);```
  
- and to determine when to re-randomize the posters, snippet 

  ```javascript
  useEffect(() => {
    // only save the shuffled poster venues state if the previous ones were empty
    // additional logic can be added to (re-/in-)validate based on time, userId or other parameters
    if (isEmptyQuery && isPostersLoaded && !previousPosterVenues.length) {
    }
  }, [
    isEmptyQuery,
    isPostersLoaded,
    shuffledPosterVenues,
    previousPosterVenues.length,
  ]);``` 
 - extra logic/checks/care should be taken on picking what array to show due to the provider being called multiple times from different components and at times of waiting for result to be fetched and/or read from previous state.

goran-peoski-work:
> Also note that not all venues might want randomization. 

My suggestion would be to determine which parameters can be added to `PostersProviderProps` or in `src/settings.ts` to let the provider be tailored per specific uses

src/hooks/posters.tsx
goran-peoski-work: done

src/hooks/posters.tsx
goran-peoski-work: reverted, with a comment above

src/hooks/posters.tsx
goran-peoski-work: done

src/hooks/posters.tsx
goran-peoski-work: done.

although I guess I should collect links to comments like these for code style guide, or at least examples for do's and don'ts

src/components/templates/PosterHall/PosterHall.tsx
goran-peoski-work: done

src/pages/VenuePage/TemplateWrapper.tsx
goran-peoski-work: will leave it as is for now

yarikoptic: FWIW: test driven the f04735d193f4fcc125643d3277c3f6322378c4c7 and it is nearly perfect for my UX.  Even if scrolling position is not to be cached as well -- it is just a minor aspect. Thank you @goran-peoski-work for all this work!

src/hooks/posters.tsx
0xdevalias: It’s just our style to explicitly check the length rather than check if the length is falsy, no need to comment.

0xdevalias: Also our style is to post the commit hash where changes were made in the review comments when marking them resolved so that others can easily see where/what was changed. Please do so.

src/hooks/posters.tsx
0xdevalias: There is a backlog issue about enabling linting for it that also has the full list in order as I described it once.

As far as collecting links to comments like this. There’s a super old PR here that mentioned documentation that’s comments have sort of ended up as a little bit of a de facto dumping ground for references to comments like this about code style/etc. With the hope/intent of getting the tone to turn that into proper doco.

It would be great if you could keep notes/references as well (ideally somewhere accessible, not just locked away on your computer/similar), would make the process way easier when it comes time to make those docs.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: @goran-peoski-work not sure if/what you changed here as you didn’t leave a commit hash, but our existing standard is to extract these arrays to a const in settings as described above, so would make sense to me to follow suit. It’s a different usecase to the ‘switchboard’ defined in this file

0xdevalias: Also I’m not sure what you mean by ‘refactor into smaller chunks’/ifs/etc. What are you referring to here, and for what purpose?

src/hooks/posters.tsx
0xdevalias: Please add the commit hash of changes made to comment threads when resolving as per our standards

src/hooks/posters.tsx
goran-peoski-work: > It’s just our style to explicitly check the length rather than check if the length is falsy, no need to comment, so probably remove that again.

Few notes on this statement:

1. There isn't explicit code style (yet, but comments like this are helpful in compiling one) that can read beforehand
2. With no explicit code style, there also isn't explicit reasoning on why strict equality (in type and value) is preferable over falsy value only
3. With no explicit code style comments could exist explaining why something isn't a code smell, like in e.g.
```javascript
    if (!normalizedSearchQuery) return filteredPosterVenues;

    const tokenisedSearchQuery = tokeniseStringWithQuotesBySpaces(
      normalizedSearchQuery
    );

    if (tokenisedSearchQuery.length === 0) return filteredPosterVenues;
```
  Granted, check of a boolean and of a length of an array are different (green to red apples, not oranges), still few comments in code may save further comments in PRs since not everyone at every time will be fully up to speed with the code style and reasoning behind it
4. Current comment was provided just because of prior art, already done one change and reversion w/o readily visible rationale, something in the line of
  // this is how it's supposed to be for the reason of...
  
Let me see if I get this hash thing right: 

Code style update at ecd77d03

src/pages/VenuePage/TemplateWrapper.tsx
goran-peoski-work: > It’s a different usecase to the ‘switchboard’ defined in this file

Didn't clam it's the same use case, just the same concern: _which templates have randomized posters_, that currently is contained within that one file.

 The original code I wrote didn't change the JSX at the `return` point but baked the Provider within the template. Ultimately that harder to visually track (with fall through cases and if check in), even if it provided for simpler return block.

> What are you referring to here, and for what purpose?

`switch` statements aren't just glorified `goto`s, but can also be considered code smell that something simple like few `if-else`s has been overengineered, or something too complicated hasn't been engineered enough with more robust code using factory functions, single responsibility etc.

@0xdevalias  Changes were made at d9761b2c, please do provide clear yes/no if you still need the array moved to `src/settings`

src/hooks/posters.tsx
0xdevalias: The solution to that is to make the time to properly document our code style, which we intend to do already. Not to litter the codebase with things that don’t belong there.

(Sarcasm) otherwise we should add a comment to every single line everywhere explaining why we follow that particular code style in that place.

It’s not how it’s done.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: Do you feel like this switch is ‘overengineered’? Doesn’t seem that way to me. And so I’m confused why you bring it up/make those comments here in general?

0xdevalias: I don’t see how the commit hash your posted above relates to this at all?

goran-peoski-work: > Do you feel like this switch is ‘overengineered’?

I don't feel like it. Example of "overengineered" would be a simple `if-else` turned into `switch`, not the case here.

>  And so I’m confused why you bring it up/make those comments here in general?

I can't answer if you're confused, can only why I brought it up here - as an attempted answer to

> What are you referring to here, and for what purpose?

I guess I didn't property address the "purpose", so to be short: **in case it got refactored some day**.

goran-peoski-work: Added the constant in d404b6c3244edd64977f6e9cec73f64506bb5fca

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -144,6 +146,10 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
       template = <div>Unknown Template: ${(venue as AnyVenue).template}</div>;
   }
 
+  if (HAS_RANDOMIZED_POSTERS_TEMPLATES.includes(venue.template)) {
+    template = <PostersProvider venueId={venue.id}>{template}</PostersProvider>;
+  }
+
```
0xdevalias, line 152: We shouldn't be mutating/reassigning the `template` var here like this.

Now that we are using `HAS_RANDOMIZED_POSTERS_TEMPLATES`, just put the rest of this back in the JSX where it was before. There was no need to extract it out here like this.

src/settings.ts
```diff
@@ -551,6 +551,12 @@ export const ALL_BURN_TEMPLATES: Array<VenueTemplate> = [
   VenueTemplate.themecamp,
 ];
 
+// @debt Refactor this constant into types/templates along with the rest *_TEMPLATES constants
```
0xdevalias, line 554: This array is different to the situation where the other venue related things talk about being moved to `types/venue`. This is a configurable setting, and belongs here. Most of the others are related to inherent properties of the venue types, not things that can be configured.

Please remove the `@debt` comment from this line.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: Actually.. thinking about this more.. `HAS_RANDOMIZED_POSTERS_TEMPLATES` is definitely not the right name for this.. If you try and render any of the venues that relies on `PostersProvider` when it's not present it will throw an error. So this has nothing to do with whether they should be randomised, and everything to do with poster provider itself.

0xdevalias: Just call it something like `POSTER_PROVIDER_TEMPLATES` or similar.

src/settings.ts
0xdevalias: Actually after the fix in naming as called out in https://github.com/sparkletown/sparkle/pull/1557/commits/d404b6c3244edd64977f6e9cec73f64506bb5fca#r655090559 this `@debt` comment may be relevant again.

src/pages/VenuePage/TemplateWrapper.tsx
goran-peoski-work: renamed b4c9d461ba2e3bb154e34624998df49e1c40bf73

src/settings.ts
goran-peoski-work: kept the comment b4c9d461ba2e3bb154e34624998df49e1c40bf73

sofisparkle:
Hello @goran-peoski-work @yarikoptic @0xdevalias we had a comment from a user that the randomisation is refreshing the order in real time every 10-20 seconds. Witnessed https://ohbm.sparkle.space/in/posterhall can you look into this?

0xdevalias:
@goran-peoski-work Merged a fix for this today. Can’t remember the PR off the top of my head, otherwise I would link it here.

yarikoptic:
Awesome! Thank you @goran-peoski-work and @0xdevalias . https://github.com/sparkletown/sparkle/pull/1656

0xdevalias:
> Awesome! Thank you @goran-peoski-work and @0xdevalias . #1656

That’s the one! :)

PR #1558, opened 2021-06-16
use venue.chatTitle to set search field message

margulies:
This fix addresses the default "Search Posters/Demos" message that appears in the posterhall template. This may not be appropriate for all contexts where posterhall is used. This PR addresses this issue by reusing `chatTitle` to provide a customized search field message. 

Thanks to @yarikoptic for catching this issue!

### DIFF
--- a/src/components/templates/PosterHall/PosterHall.tsx
+++ b/src/components/templates/PosterHall/PosterHall.tsx
@@ -46,6 +46,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
         searchInputValue={searchInputValue}
         liveFilterValue={liveFilter}
         setLiveValue={setLiveFilter}
+        searchTitleBase={venue?.chatTitle}
       />
 
       <div className="PosterHall__posters">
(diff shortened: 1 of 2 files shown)


src/components/templates/PosterHall/PosterHall.tsx
```diff
@@ -46,6 +46,7 @@ export const PosterHall: React.FC<PosterHallProps> = ({ venue }) => {
         searchInputValue={searchInputValue}
         liveFilterValue={liveFilter}
         setLiveValue={setLiveFilter}
+        searchTitleBase={venue?.chatTitle}
```
denisdimitrov, line 49: `venue.chatTitle` is being used for the side chat. Check `/hooks/chatSidebar.ts`. I don't thing the placeholder of posterhall and the sidechat title should use the same value.

margulies:
We were simply trying to avoid adding a new field to the venue type, but if
that's the easier solution, I think it would be worthwhile to be able to
specify that on a venue-level

src/components/templates/PosterHall/PosterHall.tsx
0xdevalias: Agreed

mike-lvov:

PR #1560, opened 2021-06-16
[env/github] minor text changes to create new account form

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/821 
except for the 9th step (making questions required or not, and adding some undesigned format to the last two, but I did put an asterisk next to the required questions so that they at least look required)


malaikaaiyar:
Not sure if I did this PR correctly because it added a bunch of commits that I didn't make (?)

malaikaaiyar:
something seems wrong with this, closing it

PR #1564, opened 2021-06-17
fixes styling with categories in screeningroom

margulies:
Minor css changes to address the following issue in cases where many cateogories exist in the screeningroom template:

Before:
Screen Shot 2021-06-17 at 5 08 58 AM

After:
Screen Shot 2021-06-17 at 5 07 49 AM


mike-lvov (approved):
LGTM

PR #1567, opened 2021-06-17
[env/ohbm] add venue logos to poster hall template

denisdimitrov:
Conditionally renders logos in the poster hall. Left, right, both or none.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/694

posterhall


src/settings.ts
```diff
@@ -149,6 +149,8 @@ export const PLAYA_VENUE_STYLES: Record<string, CSSProperties> = {
   },
 };
 
+export const POSTER_LOGO_SIZE = 100;
```
mike-lvov, line 152: Any reason to have it like this rather than in classes

mike-lvov (changes requested):
And I think it should be based off `env/ohbm` as the changes are very specific

src/settings.ts
denisdimitrov: Not really, I'm just using it with img's width prop like this ``

denisdimitrov:
> And I think it should be based off env/ohbm as the changes are very specific

This PR is going in `staging` instead of `env/ohbm` maybe I should change it?

mike-lvov:
> This PR is going in staging instead of env/ohbm maybe I should change it?

Yes

denisdimitrov:
Merged with `env/ohbm`

src/settings.ts
mike-lvov: Can we do it with the classes just to follow the patterns?

mike-lvov (approved):
LGTM

0xdevalias:
> > This PR is going in staging instead of env/ohbm maybe I should change it?
> 
> Yes

@mike-lvov @denisdimitrov FWIW (and we can assess this in the future when we have more capacity), this doesn't seem like a bad idea to have in the main code branch for the poster hall.

PR #1578, opened 2021-06-17
ENH: poster search - give authors more weightand authorName even twice more

yarikoptic:
Without such a change, search for a well cited (in introduction) scientist
leads to have those hits ahead of the hits where that person is the author
or among authors.  Since names are typically "unique" in sense of not overlapping
with keywords or words used in the title, I think a side-effect on search by
titles etc is (tested on some) non-existing or very minor.

Sample searches to do on env/ohbm would be Fornito or Poldrack to see the effect

### DIFF
--- a/src/hooks/posters.ts
+++ b/src/hooks/posters.ts
@@ -76,9 +76,15 @@ export const usePosters = (posterHallId: string) => {
         keys: [
           "name",
           "poster.title",
-          "poster.authorName",
+          {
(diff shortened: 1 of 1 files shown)


mike-lvov (approved):
LGTM

0xdevalias:
@sparkletown/engineering @jonboutelle @sparkletown/product These changes would be worth pulling out into our main code branch at some point.

PR #1580, opened 2021-06-17
add events by csv

ermiaSparkle:
Now you can add events to venues by CSV file.

The CSV file separator is `","`.
It will add each event to its venue and rooms( rooms are optional as we have some spaces which do not have any rooms )
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/770


0xdevalias (changes requested):
It would be good to make the changes to align with spec detailed in https://github.com/sparkletown/internal-sparkle-issues/issues/770#issuecomment-863661332

Issue #1563, opened 2021-06-17
env/ohbm: Sponsors urls on the main page should be adjusted


yarikoptic:
ATM they all point to non existing  https://ohbm.sparkle.space/in/exhibitionhall which should be https://ohbm.sparkle.space/in/exhibithall or actually specific booths like https://ohbm.sparkle.space/in/boothsiemenshealthineers for Siemens

Attn @margulies et al

yarikoptic:
now points to https://ohbm.sparkle.space/in/%20boothsiemenshealthineers so rogue space added

yarikoptic:
tried to adjust in admin interface from a localhost instance -- there is indeed a space pasted in url -- but got 
"""
One or more errors occurred when saving the form:
"""

without any details upon clicking "Update room".

yarikoptic:
fixed directly in DB (didn't check others)

PR #1585, opened 2021-06-18
fix EventModal location links

margulies:
Now that events may have a room parameter, but still be located on the venue-level, the link in the event modal seems to break. This fix changes the logic of grabbing the link url to update to events having `event.room`, but still being organized at the venue collection level.

attempts to fix https://github.com/sparkletown/sparkle/issues/1622


src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -52,10 +52,10 @@ export const EventModal: React.FC<EventModalProps> = ({
   const goToEventLocation = () => {
     onHide();
 
-    if (event.room) {
-      enterRoom();
-    } else {
+    if (event.venueId) {
```
margulies, line 55: Sorry, the logic here might wrong and could fail if the event is actually stored at the room-level. I haven't checked that use case, so definitely worth testing there.

0xdevalias: @margulies what was wrong with the existing logic? If there is a room, enter the room, otherwise enter the venue? That seems correct and straightforward to me? Not sure what the changes you're making here are for, but it doesn't look right to me?

0xdevalias: Additionally, `enterRoom` is defined in `src/hooks/useRoom.ts` and basically just ends up calling either `enterVenue` or `enterExternalRoom` depending on whether `roomVenue` exists.

margulies:
I agree there's something off about the logic. Perhaps the problem is
rather with `useRoom.ts` that is causing the links to break?

src/components/organisms/EventModal/EventModal.tsx
margulies: I know it looks weird, but if you try the OHBM setup, you'll see that clicking on a room location from the event modal returns: `//`
At least in the case of our current env/ohbm setup, where all the events are stored on the venue-level, the fix I suggested addressed this.

0xdevalias: Have we figured out the ‘why’ of that occurring? It’s better to know why than to make arbitrary changes that seem to fix the issue we haven’t even properly defined yet.

sparkleed:
This is an urgent fix requested today by OHBM. Having personally inspected the feature in action, I concur.

The commit here is more to illustrate the issue, which is what we should focus on. The issue is that links in the schedule to the rooms in which events are taking place are broken -nothing happens when you click on them- and this therefore steals the utility from the schedule.

Request is that this be alongside #1587 on the first-things-for-Monday-morning list. Many thanks!

mobilevinay:
Marked as critical as Ed would like to see this added soonish.

0xdevalias (changes requested):
Officially marking this as 'changes requested' since the current implementation doesn't look like it will be safe to merge/fix things.

0xdevalias:
@margulies It would be useful if explicit examples / steps to reproduce were included alongside fixes like this (ideally in an existing environment) so that they can be looked into. Without those details, it's a bunch of guesswork to even identify a valid 'test case', before we can even start looking into the 'why' of it + fixing things.

0xdevalias:
@margulies @yarikoptic is this just another case of the ‘not a platform/code issue’ (eg. Problem exists in the data added to firebase, not with our code) as per https://github.com/sparkletown/sparkle/issues/1563 ?

mobilevinay:
Issue needs more information before we can accept. @sparkleed and @jonboutelle to communicate.

mobilevinay:
Assigned to Viktoria to lead the investigation.

yarikoptic:
Insofar no feedback from investigators, and problem seems to persist.

0xdevalias:
> Insofar no feedback from investigators, and problem seems to persist.

@yarikoptic That's because @sunny-viktoryia was an integral part of the team working on the event fetching performance work that was put in overnight, which was deemed a higher overall priority.

alexey-petrov:
@sunny-viktoryia @0xdevalias i'm going to work on this one

0xdevalias:
FYI: This will likely be superceded by https://github.com/sparkletown/sparkle/pull/1696 as per https://github.com/sparkletown/sparkle/pull/1585#issuecomment-870400409

PR #1586, opened 2021-06-18
Fix/add category filter to posterhall

margulies:
This PR brings the functionality for filtering posters based on categories to the posterhall template. It is implemented based on the category-level filtering approach currently available in the screeningroom template. 

Given some of the challenges we're having with search specificity, this brings a valuable tool for exploring the contents of the posterhall.

I thought this might be sufficiently general for staging, but can rebase on env/ohbm if more appropriate there.


mike-lvov:

Issue #1576, opened 2021-06-17
[PosterHall]: add the same category  (and keyword?) selection as Screening Room 


yarikoptic:
adopt solution implemented in the `screeningroom` to have them visible/selectable to start with, and  
 then pre-filter for search:
![image](https://user-images.githubusercontent.com/39889/122460857-a2dba800-cf80-11eb-8915-95c8e72f0c4c.png)

PR #1587, opened 2021-06-18
[env/ohbm] add category/subcategory filtering to PosterHall

margulies:
This PR contains filtering on the category and subcategory level. It is highly specific to env/ohbm, where a postervenue has multiple category each with subcategories organized hierarchically. The category and subcategory filters are integrated with the isLive and Bookmark filters as well.


yarikoptic:
it would close https://github.com/sparkletown/sparkle/issues/1576 AFAIK! ;)  Well done!!!

yarikoptic:
most likely would need "integration" with https://github.com/sparkletown/sparkle/pull/1557 by @goran-peoski-work . Might as well be worthwhile to get ready and prepare a merge if (hopefully) that PR gets merged soon ;)

src/settings.ts
```diff
@@ -154,7 +154,7 @@ export const PLAYA_VENUE_STYLES: Record<string, CSSProperties> = {
 };
 
 export const ACCEPTED_IMAGE_TYPES =
-  "image/png,image/x-png,image/gif,image/jpg,image/jpeg,image/tiff,image/bmp,image/gif";
+  "image/png,image/x-png,image/gif,image/jpg,image/jpeg,image/tiff,image/bmp,image/gif,image/webp";
```
yarikoptic, line 157: May be with submitting this commit as a separate PR right away against staging?

margulies: Good call -- thanks! https://github.com/sparkletown/sparkle/pull/1601

sparkleed:
This improvement to the UX has been requested by OHBM, on basis that it dramatically improves UX of finding posters (and makes up for the fact that search results on title key terms currently present in reverse-order, an issue for which there is no obvious current understanding or fix.

The code has been run by different people and different devices, and has been correspondingly QA-ed.

@margulies has said that the code, though carefully structured and limited to this context, may not be of truly Sparkle-generality and quality; I'd nonetheless like to see this merged as a first priority overnight - for among other reasons that it saves us from a different issue (which Yarik will shortly file) which is that search results are reversed.

It will be noted that this PR also includes ability to play video from poster page, which is also a good UX improvement- even if not strictly speaking best practice.

src/hooks/posters.ts
```diff
@@ -75,17 +93,91 @@ export const usePosters = (posterHallId: string) => {
...
+  const liveFilteredPosterVenues = useMemo(
+    () =>
+      liveFilter
+        ? filteredPostersBySubCategory.filter(
+            (filteredPostersBySubCategory) =>
+              filteredPostersBySubCategory.isLive
+          )
+        : filteredPostersBySubCategory,
+    [filteredPostersBySubCategory, liveFilter]
+  );
```
0xdevalias, line 180: There is already filtering going based on 'live' above this in `liveFilterResultsSubCategory` and similar. Why is this also needed here?

0xdevalias: Additionally, use the guard pattern as described above rather than nesting complex ternaries like this.

src/hooks/posters.ts
```diff
@@ -170,14 +262,22 @@ export const usePosters = (posterHallId: string) => {
...
     liveFilter,
     bookmarkedFilter,
 
     increaseDisplayedPosterCount,
     setSearchInputValue,
     setLiveFilter,
     setBookmarkedFilter,
+    setCategoryFilter,
+    setSubCategoryFilter,
+    unsetCategoryFilter,
```
0xdevalias, line 280: Subcategory is a single word, so shouldn't be cased as 'SubCategory':

- https://en.wikipedia.org/wiki/Subcategory

src/settings.ts
0xdevalias, line 157: Revert this unrelated change, as it's handled in https://github.com/sparkletown/sparkle/pull/1601

```suggestion
  "image/png,image/x-png,image/gif,image/jpg,image/jpeg,image/tiff,image/bmp,image/gif";
```

src/types/posters.ts
```diff
@@ -1,6 +1,12 @@
+export interface Category {
+  category: string;
+  subcategory: string;
+}
+
 export type Poster = {
   authorName: string;
   categories: string[];
+  subcategories?: Array<Category>;
```
0xdevalias, line 9: Why does `Category` have both 'category' and 'subcategory' keys on it? And given it's only used in the `Poster`'s `subcategories` key.. should it not be a `Subcategory` at the very least?

Why is `Poster`'s `categories` of type `string[]`, yet `subcategories` is `Array<Category>`?

Why are we using `Array<foo>` instead of our standard `foo[]` syntax?

src/types/venues.ts
```diff
@@ -179,6 +179,7 @@ export interface BaseVenue {
   leftLogo?: string;
   rightLogo?: string;
   canBeBookmarked?: boolean;
+  showCategoriesFilter?: boolean;
```
0xdevalias, line 182: It would make sense to sort this with all of the other `show*` keys here.

src/components/templates/PosterHall/PosterHall.scss
```diff
@@ -29,4 +31,30 @@
...
+    justify-content: center;
+    margin: $spacing--xs;
+    flex-wrap: wrap;
+  }
+
+  &__subcategory {
+    font-size: $font-size--xxs;
+    cursor: pointer;
+    text-transform: capitalize;
+  }
```
0xdevalias, line 59: It would probably make sense to extract the categories/subcategories stuff into their own component rather than batching it all together in `PosterHall`?

mobilevinay:
Glenn reviewed this today...
Someone else should look at it... it looks like it could break stuff.

mobilevinay:
CTO approves not accepting these changes.

mike-lvov:
Doesn't seem relevant anymore

PR #1589, opened 2021-06-18
remove iframe border in PosterHall

yarikoptic:
- IMHO it makes display more "coherent" (there are no borders around other
  elements, e.g. drop down Schedule, Chat, etc)
- make it possible for presented in iframe material to "meld" into the page
  better, again making it more "coherent" with the rest
  - opens possibility to take advantage of that e.g. with black background
    so presented stuff is nicely integrated into the display instead of aleniated
    with the border


src/components/templates/PosterPage/PosterPage.tsx
```diff
@@ -183,6 +183,7 @@ export const PosterPage: React.FC<PosterPageProps> = ({ venue }) => {
             title={title}
             allow={IFRAME_ALLOW}
             allowFullScreen
+            frameBorder="0"
```
0xdevalias, line 186: I’m not 100% off the top of my head but I’m pretty sure this is a legacy prop and this should be done with CSS these days.

0xdevalias (changes requested):

src/components/templates/PosterPage/PosterPage.tsx
yarikoptic: done! commit was rewritten and force pushed

0xdevalias:
@mike-lvov and/or @sparkletown/product You’re more versed in the PosterHall designs. Is removing the border like this fine?

@yarikoptic can you please provide before/after screenshots to make comparison/decision easier.

src/components/templates/PosterPage/PosterPage.tsx
0xdevalias: Thanks :)

yarikoptic:
> @yarikoptic can you please provide before/after screenshots to make comparison/decision easier.

hm... have I spoiled someone? ;)

before | after
--- | ---
image| image

yarikoptic:
Please let me know if I could facilitate the review of this non elaborate PR.

mike-lvov (approved):
LGTM

yarikoptic:
oh, it was for staging, not ohbm, oh well -- may be would still percolate ;)  Thank you @mike-lvov !

0xdevalias:
@yarikoptic 
image

PR #1597, opened 2021-06-18
FIX check if event.host is empty when creating calendar

htwangtw:
The calendar ics download was not working properly and pointed to an error related to `event.host`.
I suspect empty string broke the check on host name and email in the upstream package. Now I add a dirty check to ensure the calendar generator either use the value from event.host or has a string saying `undefined`
cc @margulies


mike-lvov:
Was implemented and is in staging currently

PR #1598, opened 2021-06-18
[env/ohbm] do not disable "See all" for user list in RoomModal

### LINKED ISSUE #1579: venue modal should not list all participants!
With @soichih we are testing env/ohbm and main auditorium has >1000 participants ATM. Clicking on "Main auditorium" opens a modal with all 1000 listed BEFORE the schedule block etc, so you need to scroll down for a while to get to what is actually useful...

I think listing first (random) 20 participants and saying "+ 1000 more" would be sufficient.

edit 1:
<details>
<summary>Here is how it looks like -- pay attention to the scroll bar on the right</summary> 

image

</details>

also all those constantly flicker replacing with others etc, resizing window left me at some point with a stuck view of only the half of the screen, then browser locked up (I have many other tabs though), took minutes to go normal.

limiting number of avatars would help with that I hope.

yarikoptic:
From looking at the code, if disableSeeAll is True, all users
are shown.  At OHBM we will have > 1000 and that leads to
a very long list!  I think it would be useful for the event to
just collapse users by default and allow to expand if someone
really desires it

Closes #1579


src/components/molecules/UserList/UserList.tsx
```diff
@@ -32,7 +32,7 @@ export const UserList: React.FC<UserListProps> = ({
   limit = DEFAULT_USER_LIST_LIMIT,
   imageSize = 40,
   activity = "partying",
-  disableSeeAll = true,
+  disableSeeAll = false,
```
0xdevalias, line 35: Given you’re changing the default here it’s going to have impact in more places across the app than just the RoomModal. Have you found/asses the impact this will have on those locations? Where else is this used currently?

The minimal localised fix to the RoomModal would be to just pass through this prop on the RoomModa, rather than changing the default.

0xdevalias (changes requested):

src/components/molecules/UserList/UserList.tsx
yarikoptic: I did click around and didn't spot oddity ... and couldn't see a use-case where I would want to see all potential 1000  (or even > 10) avatars.  If it is less than 11 (IIRC -- some magical number) it still shows all of them and without "See all", so I thought it would be actually of advantage generally.

I have tried to pass `disableSeeAll = {false}` in EventModal but for some reason it had no effect.

0xdevalias: In EventModal? Isn’t this in the RoomModal?

yarikoptic: yeap -- I thought it must have been that one since none other Modal has it
```
$> git grep UserList | grep Modal
src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx:import { UserList } from "components/molecules/UserList";
src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx:      <UserList
```
but looking there I actually see the `11` which I didn't saw back then -- must have been some wrong place ;)
```
      <UserList
        users={recentRoomUsers}
        limit={11}
        activity="in this room"
        attendanceBoost={room.attendanceBoost}
      />
```
so , reverted default (although still don't quite grasp why by default to display all possible 100s) and changed to have it only in RoomModal - seems to work. pushed

0xdevalias: Changing the default may eventually make sense, but without the capacity to assess the impact across all places, making a smaller localised change is easier.

0xdevalias (approved):
LGTM! :shipit:

0xdevalias:
@sparkletown/engineering Note: this probably makes sense to change in our main branch as well, possibly even changing the default across the platform, but I’ve merged into env/ohbm for now to expedite this.

PR #1599, opened 2021-06-18
Adjust .ScheduleItem itself and reset padding/margin left/right

### LINKED ISSUE #1582: venue modal -- make a tiny bit wider or adjust some margins/padding?
ATM it requires horizontal scroll bar on the main auditorium event modal

image

yarikoptic:
instead toying with .schedule-item-container margin/padding

Closes #1582

With this looks like

before | after 
----|----
image | image

### DIFF
--- a/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
+++ b/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
@@ -68,12 +68,18 @@ $spacing: 20px;
   &__events {
     max-height: 35vh;
     overflow-y: auto;
+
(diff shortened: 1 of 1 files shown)


0xdevalias:
@yarikoptic What is the difference between your before/after screenshots? They look exactly the same to me?

yarikoptic:
Before has horizontal scroll bar add described in the original issue

0xdevalias:
> Before has horizontal scroll bar as described in the original issue

Ah! I see it now! 😛 I think with tired eyes my brain filtered it as being part of the background rather than part of the screenshot.

PR #1601, opened 2021-06-20
adds support for webp image type

margulies:
This PR adds the possibility of uploading webp image types: https://developers.google.com/speed/webp 

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/841


jonboutelle:
Whoa. @0xdevalias what do you think of this?

src/components/molecules/ImageInput/ImageCollectionInput.tsx
```diff
@@ -108,7 +108,7 @@ export const ImageCollectionInput: React.FC<ImageInputProps> = (props) => {
           disabled={disabled}
           type="file"
           onChange={handleFileChange}
-          accept="image/png,image/x-png,image/gif,image/jpeg"
+          accept="image/png,image/x-png,image/gif,image/jpeg,image/webp"
```
0xdevalias, line 111: @margulies This (+ anywhere else across the platform that deals with uploading images) should be using `ACCEPTED_IMAGE_TYPES`

0xdevalias:
> Whoa. @0xdevalias what do you think of this?

@jonboutelle Specific thoughts in more detail left in https://github.com/sparkletown/internal-sparkle-issues/issues/841#issuecomment-864710696 This is a nice iterative improvement, but it's no 10x performance gain as it was originally described in that issue (unless there are metrics to show how that is possible you can refer to)

src/components/molecules/ImageInput/ImageCollectionInput.tsx
malaikaaiyar: fixed in 95a5b0761781b55a76026bd102f9dd080222675a

malaikaaiyar: that is the only place which seems to deal with image types

mike-lvov (approved):
LGTM

PR #1607, opened 2021-06-20
remove description from ongoing event in room modal

margulies:
This PR addresses the issue that a long description for the current (or upcoming) event in the room modal can make the rest of the modal unwieldy -- it becomes difficult to view the future events below. As the current event is actually duplicated in the list of events below, the description seems redundent at the top and coule be removed:
Before: 
Screen Shot 2021-06-20 at 10 01 28 PM
After: 
Screen Shot 2021-06-20 at 10 01 11 PM


0xdevalias:
Agree with this change in general, quick skim of code changes from my phone seem fine. Where is that component used currently though? Is the only place it’s rendered?

PR #1608, opened 2021-06-21
Enh: disable leave seat

katjaq:
Thank you for all your amazing and hard work! Sparkle is awesome, and OHBM will be fantastic!  ✨  I love the final design! 😍 

This PR is just a little enhancement. It removes the ability to "Leave my seat" while attending an event in the main auditorium and symposia halls. This functionality is not needed, (unnatural and even confusing as it does not make the attendee leave the room, but just turns them into a floating ghost). 

I also added a tooltip for the speaker icon in the emoji container for main auditorium and symposia halls to prevent confusion (feedback included that this could be to unmute oneself to talk to audience).

(I did not add the tooltip to the Jazz Lounge but if you'd prefer to have it there for consistency, happy to add it.)

### DIFF
--- a/src/components/templates/Audience/Audience.scss
+++ b/src/components/templates/Audience/Audience.scss
@@ -87,27 +87,6 @@ $spacing: calc(1 * min(var(--seat-size), var(--seat-size-min)));
             font-size: 20px;
             cursor: pointer;
           }
-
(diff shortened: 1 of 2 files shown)


mike-lvov:
@sofisparkle @Sparkle-T wdyt?

PR #1610, opened 2021-06-21
[env/ohbm] refactor Login component to fetchCustomAuthConfig from sovereignVenueId

0xdevalias:
Currently when using the custom auth config (originally added in https://github.com/sparkletown/sparkle/pull/1523) you need to log in from the exact venue that it's configured on, but I noticed a bunch of errors coming through where people seemed to be trying to login on some of the related venues, which didn't have access to the custom auth config.

This PR refactors that feature so that it will always try and load it from the sovereign venue instead, which should allow it to be available from any of the related venues.

An edgecase/limitation of this PR in it's current minimal state is:

- if you try to access a related venue `B`
- the auth config will be loaded from the sovereign venue `A`
- login should work correctly
- but you will be returned to the sovereign venue `A` afterwards (not `B`) that you originally tried to access

**Note:** This PR is probably beneficial to bring back into the main code branch at some point, but for now I didn't want to complicate/slow things down by needing to deploy it through there + assess the wider impact of the changes in general.

### DIFF
--- a/src/pages/Account/Login.tsx
+++ b/src/pages/Account/Login.tsx
@@ -11,6 +11,7 @@ import { isDefined } from "utils/types";
 import { openUrl } from "utils/url";
 
 import { useSAMLSignIn } from "hooks/useSAMLSignIn";
+import { useSovereignVenueId } from "hooks/useSovereignVenueId";
 
 import { InitialForm } from "components/organisms/AuthenticationModal/InitialForm";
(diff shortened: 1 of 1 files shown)


denisdimitrov (approved):
LGTM!

mike-lvov:
Doesn't seem to be relevant now

PR #1614, opened 2021-06-21
Peformance fix: reduce the number of user profile updates

sunny-viktoryia:
using `{ }` notation we force all the related components to re-render creating a new object every time. 
`Object.assign()` uses the same target object. 
Hopefully, that can reduce the number of component re-renders on the users' side.

I've chosen `env/ohbm` as a base branch to reduce the fire, maybe we could find a better approach for our main codebase `staging`.

### DIFF
--- a/src/hooks/useUser.ts
+++ b/src/hooks/useUser.ts
@@ -21,7 +21,8 @@ export const useUser = (): UseUserResult => {
   return {
     user: !auth.isEmpty ? auth : undefined,
     profile: !profile.isEmpty ? profile : undefined,
-    userWithId: auth && profile ? { ...profile, id: auth.uid } : undefined,
+    userWithId:
+      auth && profile ? Object.assign(profile, { id: auth.uid }) : undefined,
     userId: auth && profile ? auth.uid : undefined,
   };
 };
--- a/src/hooks/useUser.ts
+++ b/src/hooks/useUser.ts
@@ -14,6 +14,8 @@ type UseUserResult = {
   userId?: string;
 };
 
+const emptyUserProfile = {};
+
 export const useUser = (): UseUserResult => {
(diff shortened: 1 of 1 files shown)


src/hooks/useUser.ts
```diff
@@ -14,14 +14,19 @@ type UseUserResult = {
...
   return {
     user: !auth.isEmpty ? auth : undefined,
     profile: !profile.isEmpty ? profile : undefined,
-    userWithId: auth && profile ? { ...profile, id: auth.uid } : undefined,
+    userWithId:
+      auth && profile
+        ? Object.assign(emptyUserProfile, profile, { id: auth.uid })
+        : undefined,
     userId: auth && profile ? auth.uid : undefined,
   };
```
0xdevalias, line 31: As written here, this is essentially mutating global state, and will have concurrency issues where literally every usage of the `useUser` hook is going to try and update the same object.

The proper way to fix this here is to `useMemo` the `userWithId` result, which seems like a worthwhile improvement to make in this PR, but it's not worth us using the verbose `Object.assign` syntax for in place of the current object spread syntax.

This is also unlikely to have a major impact on the performance issues across the platform, as most of those aren't coming via the current user, but by users read in through other places, which this will have no impact on.

Further Reading on object spread syntax:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals

0xdevalias (changes requested):
Comments/changes/etc.

Worth re-targeting this to `staging` when it's worked on as well too, as a 'generally good improvement to the codebase' that isn't critical/immediate in resolving the current fires.

mike-lvov:
Doesn't seem to be relevant as it mutates the state with Object.assign

PR #1617, opened 2021-06-21
[env/burn-staging] add new AnimateMap venue template

YarRainbow:
- The template was added following the example of the PartyMap.
- Added dependencies to the package.json
- Added venue reducer and another functions for work with redux-store
- Added basic objects for upcoming works in the canvas.
- Added vieports and quadtree for rendering objects
- Cuted out all demo data


0xdevalias:
@YarRainbow Out of curiosity, was this branch originally opened targeting `staging`/similar? I'm curious why https://github.com/sparkletown/sparkle/pull/1617#issuecomment-865285655 ran on this PR, as I don't think it should have currently based on the trigger rules in our workflow ([ref](https://github.com/sparkletown/sparkle/blob/staging/.github/workflows/firebase-hosting-pull-request.yml#L4-L15)). So trying to understand what happened with our systems here. (cc // @sunny-viktoryia FYI)

YarRainbow:
@0xdevalias sorry, I was late for the train and was in a hurry. All our branches forked from env/burn-staging and target on this branch.
If this broken some github actions, you can reopen this correctly and close broken PR. I will be grateful to you.
I will be available after 17pm (GMT+10)

src/store/actions/AnimateMap.ts
```diff
@@ -0,0 +1,73 @@
...
+  UPDATE_USERS = "AnimateMapActionTypes.UPDATE_USERS",
+  UPDATE_VENUES = "AnimateMapActionTypes.UPDATE_VENUES",
+}
+
+export type setAnimateMapStageOptionsAction = ReduxAction<
+  AnimateMapActionTypes.UPDATE_STAGE_OPTIONS,
+  {
+    options: AnimateMapStageOptions;
+  }
+>;
```
0xdevalias, line 20: This file seems to be based on our legacy/outdated/tech debt redux patterns. For all new work we want to be following Redux best practices and use Redux Toolkit, as per:

- https://redux.js.org/style-guide/style-guide#use-redux-toolkit-for-writing-redux-logic
- https://redux-toolkit.js.org/
  - https://redux-toolkit.js.org/usage/usage-with-typescript
  - https://redux-toolkit.js.org/tutorials/typescript

---

I though there should be an example of existing code I could show you in the codebase currently, but I couldn't find it, so I expect it may be buried away in one of my open WIP PR's (if I can find it, I will link it here so you can see a 'real world code implemented example' of this)

In essence, we basically want to be using `createSlice` for the main bulk of Redux state setup:

- https://redux-toolkit.js.org/usage/usage-with-typescript#createslice

Potentially falling back to using `createAction` or other things in extreme edgecases (though most of the time `createSlice` will handle that need for us anyway:

- https://redux-toolkit.js.org/usage/usage-with-typescript#createaction

Sometimes `createAsyncThunk` will also be useful:

- https://redux-toolkit.js.org/usage/usage-with-typescript#createasyncthunk

This pattern batches all of the actions/reducer/etc into a single place, as is best practice, so instead of including this in `store/actions/*`, etc; we should just make the file directly in `store/` directly.

src/store/actions/AnimateMap.ts
```diff
@@ -0,0 +1,73 @@
...
+    venues: Map<string, ReplicatedVenue>;
+  }
+>;
+
+export const setAnimateMapStageOptions = (
+  options?: AnimateMapStageOptions
+): setAnimateMapStageOptionsAction => ({
+  type: AnimateMapActionTypes.UPDATE_STAGE_OPTIONS,
+  payload: { options },
+});
```
0xdevalias, line 48: The name of these Redux actions should be aligned to their type. So instead of `setAnimateMapStageOptions` this would be `updateAnimateMapStageOptions` or similar.

0xdevalias: In general though, actions shouldn't be defined as 'get' / 'set' / etc, but should rather be modelled as events. See the Redux styleguide/best practices for more information:

- https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+// see: https://dev.to/kjleitz/comment/gb5d
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DropFirstInTuple<T extends any[]> = T extends [arg: any, ...rest: infer U]
+  ? U
+  : T;
+
+// Gets the type of the last element of a tuple.
+// see: https://dev.to/kjleitz/comment/gb5d
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type LastInTuple<T extends any[]> = T[LengthOfTuple<DropFirstInTuple<T>>];
```
0xdevalias, line 25: These are utility types, so belong in `types/utility`

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+export interface AnimateMapPoint {
+  x: number; //integer
+  y: number; //integer
+}
+
+export interface ReplicatedUserData {
+  id: string;
+  videoUrlString: string;
+  avatarUrlString: string;
+  dotColor: number; //hex
```
0xdevalias, line 42: If color is intended as hex (i'm assuming a hex color code here, eg. `#FFAABB`) then shouldn't it be a `string`?

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface ReplicatedVenueData {
+  id: string;
+  videoUrlString: string;
+  imageUrlString: string;
+}
+
+export interface ReplicatedVenue extends AnimateMapPoint {
+  data: ReplicatedVenueData;
+}
```
0xdevalias, line 57: I expect these types belong in a `types/*` location rather than here within this reducer file.

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+interface AnimateMapState {
+  stageOptions: AnimateMapStageOptions;
+  worldWidth: number;
+  worldHeight: number;
+  zoom: number;
+  users: Map<string, ReplicatedUser>;
+  usersQT: QuadTree | null;
+  venues: Map<string, ReplicatedVenue>;
+  venuesQT: QuadTree | null;
+}
```
0xdevalias, line 68: Preferencing `undefined` shorthand over `null` + using proper names rather than abbreviations + ensuring that the `Map` type doesn't 'lie' about being able to produce a value for literally any given string as it previously did:



```suggestion
interface AnimateMapState {
  stageOptions: AnimateMapStageOptions;
  worldWidth: number;
  worldHeight: number;
  zoom: number;
  users: Partial<Map<string, ReplicatedUser>>;
  usersQuadTree?: QuadTree;
  venues: Partial<Map<string, ReplicatedVenue>>;
  venuesQuadTree?: QuadTree;
}
```

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+export type AnimateMapStageOptions = LastInTuple<
+  ConstructorParameters<typeof PIXI.Application>
+>;
+
+export type AnimateMapWorldBounds = { width: number; height: number };
+
+export interface AnimateMapPoint {
+  x: number; //integer
+  y: number; //integer
+}
```
0xdevalias, line 36: We already have a type for this (`Point`) in `src/types/utility.ts`

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+// Gets the type of the last element of a tuple.
+// see: https://dev.to/kjleitz/comment/gb5d
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type LastInTuple<T extends any[]> = T[LengthOfTuple<DropFirstInTuple<T>>];
+
+export type AnimateMapStageOptions = LastInTuple<
+  ConstructorParameters<typeof PIXI.Application>
+>;
+
+export type AnimateMapWorldBounds = { width: number; height: number };
```
0xdevalias, line 31: We already have a type for this (`Dimensions`) in `src/types/utility.ts`

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface AnimateMapPoint {
+  x: number; //integer
+  y: number; //integer
+}
+
+export interface ReplicatedUserData {
+  id: string;
```
0xdevalias, line 39: Is the `id` guaranteed to always be present everywhere this type is used? We have a `WithId` type + related helpers in `src/utils/id.ts` if not. So that we can represent the type easily in cases where it both does and doesn't have the `id` 'baked in' to the data.

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface ReplicatedUser extends AnimateMapPoint {
+  data: ReplicatedUserData;
+}
+
+export interface ReplicatedVenueData {
+  id: string;
+  videoUrlString: string;
+  imageUrlString: string;
+}
```
0xdevalias, line 53: Appending 'string' to the variable name here is redundant/goes against our naming patterns:

```suggestion
export interface ReplicatedVenueData {
  id: string;
  videoUrl: string;
  imageUrl: string;
}
```

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface ReplicatedVenue extends AnimateMapPoint {
+  data: ReplicatedVenueData;
+}
+
+interface AnimateMapState {
+  stageOptions: AnimateMapStageOptions;
+  worldWidth: number;
+  worldHeight: number;
```
0xdevalias, line 62: Curious why not use the `Dimensions` type for this? Is there a reason this was kept flat/etc?

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+  zoom: number;
+  users: Map<string, ReplicatedUser>;
+  usersQT: QuadTree | null;
+  venues: Map<string, ReplicatedVenue>;
+  venuesQT: QuadTree | null;
+}
+
+const initialAnimateMapState: AnimateMapState = {
+  stageOptions: {
+    backgroundColor: 0x10bb99,
```
0xdevalias, line 72: Is there a reason we're using a hex number here instead of a hex string/similar?

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+  venuesQT: null,
+};
+
+export type AnimateMapReducer = Reducer<AnimateMapState, AnimateMapActions>;
+
+function generateNewQuadtree(
+  items: Map<string, ReplicatedVenue | ReplicatedUser>,
+  quadTree: QuadTree | null,
+  state: AnimateMapState
+): QuadTree {
```
0xdevalias, line 92: We use the `const` + `() => {` pattern for functions + the `?` shorthand for `undefined` values instead of explicit `| null` + named parameters rather than multiple positional parameters:

```suggestion
export interface GenerateNewQuadtreeProps {
  items: Map<string, ReplicatedVenue | ReplicatedUser>;
  quadTree?: QuadTree;
  state: AnimateMapState;
}

const generateNewQuadtree = ({ items, quadTree, state): QuadTree => {
```

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (!quadTree) {
+    // create tree
+    quadTree = new QuadTree(
+      new Box(0, 0, state.worldWidth, state.worldHeight),
+      { maximumDepth: 100 },
+      Array.from(items).map(([key, value]) => value)
+    );
```
0xdevalias, line 99: This is mutating the passed on `quadTree` prop, which we don't do. In most if not all cases, functions should be pure (in the sense that they don't mutate their params) + return should return their changes.

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+    return quadTree;
+  } else {
+    // update tree
+    quadTree.clear(); //NOTE: can optimize if we will remove certain elements before update
+    quadTree.insert(Array.from(items).map(([key, value]) => value));
```
0xdevalias, line 104: Is there a need to mutate the provided quadTree param rather than just returning a new one?

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+  action: AnimateMapActions
+): AnimateMapState => {
+  switch (action.type) {
+    case AnimateMapActionTypes.UPDATE_STAGE_OPTIONS:
+      const { options } = action.payload;
+      return { ...state, stageOptions: options };
+
+    case AnimateMapActionTypes.UPDATE_ZOOM:
+      const { zoom } = action.payload;
+      return { ...state, zoom: zoom };
```
0xdevalias, line 120: In cases like this where the key and the value are named the same, you can use the following shorthand:
```suggestion
      return { ...state, zoom };
```

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    case AnimateMapActionTypes.UPDATE_USERS:
+      const { users } = action.payload;
+      const usersQT = generateNewQuadtree(users, state.usersQT, state);
+      return { ...state, users: users, usersQT: usersQT };
```
0xdevalias, line 125: Looking at where/how `generateNewQuadtree` is being used here, it should DEFINITELY not be modifying/mutating the parameters passed in to it. It should return a new copy of the data. Redux state is immutable:

- https://redux.js.org/style-guide/style-guide#do-not-mutate-state

src/store/reducers/AnimateMap.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    case AnimateMapActionTypes.UPDATE_VENUES:
+      const { venues } = action.payload;
+      const venuesQT = generateNewQuadtree(venues, state.venuesQT, state);
```
0xdevalias, line 129: As above RE: not mutating state, etc.

src/store/reducers/index.ts
```diff
@@ -1,14 +1,16 @@
...
 import { sovereignVenueReducer } from "./SovereignVenue";
 import { userProfileReducer } from "./UserProfile";
+import { animateMapReducer } from "./AnimateMap";
+import { VenueTemplate } from "../../types/venues";
 
 // Reducers per VenueTemplate (eg. reducer for playa template)
-export const VenueTemplateReducers: { [key: string]: Reducer } = {};
+export const VenueTemplateReducers = {
+  [VenueTemplate.animatemap]: animateMapReducer,
+};
```
0xdevalias, line 13: Why was the explicit typing of this function removed?

Should be either (matching how it used to be):
```typescript
export const VenueTemplateReducers: { [key: string]: Reducer } = {
  [VenueTemplate.animatemap]: animateMapReducer,
};
```

Or if the typing works, this would seem more explicit:
```typescript
export const VenueTemplateReducers: { [key: VenueTemplate]: Reducer } = {
  [VenueTemplate.animatemap]: animateMapReducer,
};
```

Which could also likely be simplified to just:
```typescript
export const VenueTemplateReducers: Partial<Map<VenueTemplate, Reducer>> = {
  [VenueTemplate.animatemap]: animateMapReducer,
};
```

src/types/venues.ts
```diff
@@ -178,6 +181,11 @@ export interface GenericVenue extends BaseVenue {
   template: GenericVenueTemplates;
 }
 
+export interface AnimateMapVenue extends BaseVenue {
+  id: string;
+  template: VenueTemplate.animatemap;
+}
```
0xdevalias, line 187: `id` shouldn't be baked directly into this type like this, as in a majority of cases, it's not guaranteed to be present based on how we query the data. We have a helper type (`WithId`) for showing this distinction where necessary.

```suggestion
export interface AnimateMapVenue extends BaseVenue {
  template: VenueTemplate.animatemap;
}
```

0xdevalias: Once the `id` key is removed, this is essentially just the same as `BaseVenue`. Is it anticipated that more specific data is going to be added here in future? If so, it's fine to keep this as is for now. But if not, then we should just use `GenericVenue`

src/types/venues.ts
```diff
@@ -56,6 +58,7 @@ export type AnyVenue =
   | EmbeddableVenue
   | JazzbarVenue
   | PartyMapVenue
+  | AnimateMapVenue
   | PosterPageVenue;
```
0xdevalias, line 62: These are in alphabetic order (except for `GenericVenue` which is always at the top)
```suggestion
  | AnimateMapVenue
  | EmbeddableVenue
  | JazzbarVenue
  | PartyMapVenue
  | PosterPageVenue;
```

src/types/venues.ts
```diff
@@ -46,6 +47,7 @@ export type GenericVenueTemplates = Exclude<
   | VenueTemplate.embeddable
   | VenueTemplate.jazzbar
   | VenueTemplate.partymap
+  | VenueTemplate.animatemap
   | VenueTemplate.posterpage
   | VenueTemplate.themecamp
```
0xdevalias, line 52: These are in alphabetic order:
```suggestion
  | VenueTemplate.animatemap
  | VenueTemplate.embeddable
  | VenueTemplate.jazzbar
  | VenueTemplate.partymap
  | VenueTemplate.posterpage
  | VenueTemplate.themecamp
```

src/utils/selectors.ts
```diff
@@ -20,6 +20,13 @@ import {
   makeIsRequestingSelector,
   makeOrderedSelector,
 } from "./firestoreSelectors";
+import {
+  AnimateMapStageOptions,
+  AnimateMapWorldBounds,
+  ReplicatedUser,
+  ReplicatedVenue,
+} from "../store/reducers/AnimateMap";
```
0xdevalias, line 28: These types should all be moved to `types/*` (as called out earlier in this review), not imported from the store/reducer file.

src/utils/selectors.ts
```diff
@@ -243,3 +250,36 @@ export const noopSelector: SparkleSelector<undefined> = () => undefined;
 
 export const emptyArray = [];
 export const emptyArraySelector = <T>(): T[] => emptyArray;
+
+// AnimateMap
+export const animateMapStageOptionsSelector: SparkleSelector<AnimateMapStageOptions> = (
+  state
+) => state.animatemap.stageOptions;
+
+export const animateMapWorldBoundsSelector: SparkleSelector<AnimateMapWorldBounds> = (
```
0xdevalias, line 259: Use our existing types:

```suggestion
export const animateMapWorldBoundsSelector: SparkleSelector<Dimensions> = (
```

src/utils/selectors.ts
```diff
@@ -243,3 +250,36 @@ export const noopSelector: SparkleSelector<undefined> = () => undefined;
...
+  Map<string, ReplicatedVenue>
+> = (state) => state.animatemap.venues;
+
+export const animateMapUsersQTSelector: SparkleSelector<QuadTree | null> = (
+  state
+) => state.animatemap.usersQT;
+
+export const animateMapVenuesQTSelector: SparkleSelector<QuadTree | null> = (
+  state
+) => state.animatemap.venuesQT;
```
0xdevalias, line 285: As called out earlier in review, `undefined` rather than `null`:

```suggestion
export const animateMapUsersQTSelector: SparkleSelector<QuadTree | undefined> = (
  state
) => state.animatemap.usersQT;

export const animateMapVenuesQTSelector: SparkleSelector<QuadTree | undefined> = (
  state
) => state.animatemap.venuesQT;
```

src/utils/selectors.ts
```diff
@@ -243,3 +250,36 @@ export const noopSelector: SparkleSelector<undefined> = () => undefined;
...
+export const animateMapZoomSelector: SparkleSelector<number> = (state) =>
+  state.animatemap.zoom;
+
+export const animateMapUsersSelector: SparkleSelector<
+  Map<string, ReplicatedUser>
+> = (state) => state.animatemap.users;
+
+export const animateMapVenuesSelector: SparkleSelector<
+  Map<string, ReplicatedVenue>
+> = (state) => state.animatemap.venues;
```
0xdevalias, line 277: There may be good reasons for it, but curious what the reason needs to be for having so many fine grained selectors here?

If we're going to do this, there should be a 'base selector' for `state.animatemap` that is then used by all of these other 'sub-selectors'

tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
0xdevalias, line 20: Can you please explain the intent/purpose/need of this change.

src/components/templates/AnimateMap/AnimateMap.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <div ref={containerRef} className="animatemap-venue-container">
+      {container && (
+        <ReactReduxContext.Consumer>
+          {({ store }) => (
+            <Stage
+              width={container.offsetWidth}
+              height={container.offsetHeight}
+              options={Object.assign(options, { resizeTo: container })}
```
0xdevalias, line 40: Instead of `Object.assign`, use the ES6+ spread syntax:

```suggestion
              options={{...options, resizeTo: container }}
```

src/components/templates/AnimateMap/AnimateMap.tsx
```diff
@@ -0,0 +1,85 @@
...
+            >
+              <AppConsumer>
+                {(app) => (
+                  <Provider store={store}>
+                    {app ? <LoadingContainer /> : <Container />}
+                  </Provider>
```
0xdevalias, line 46: Please use the redux hooks-based API: https://react-redux.js.org/api/hooks

src/components/templates/AnimateMap/AnimateMap.tsx
```diff
@@ -0,0 +1,85 @@
...
+      setIsLoading(true);
+
+      app.loader.reset().add(MAP_IMAGE).load();
+      app.loader.onLoad.add(() => setIsLoaded(true));
+      app.loader.onError.add(
+        (loader: PIXI.Loader, resource: PIXI.LoaderResource) =>
+          console.error(resource)
+      );
+    }
+  }, [isLoading, app]);
```
0xdevalias, line 74: Instead of nesting entire blocks within an `if`, we use 'guard statements' at the start of the block, that 'shortcut' by returning when the condition isn't met. Note that there should always be a blank line of space after a 'guard statement' like this,=:

```suggestion
  useEffect(() => {
    if (isLoading || !app) return;

    setIsLoading(true);

    app.loader.reset().add(MAP_IMAGE).load();
    app.loader.onLoad.add(() => setIsLoaded(true));
    app.loader.onError.add(
      (loader: PIXI.Loader, resource: PIXI.LoaderResource) =>
        console.error(resource)
    );
  }, [isLoading, app]);
```

0xdevalias: Does `app.loader` need to be 'cleaned up' in the `useEffect`'s cleanup return?

- https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

src/components/templates/AnimateMap/components/Common/Viewport.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { PixiComponent } from "@inlet/react-pixi";
+import { MovedEventData, Viewport, ZoomedEventData } from "pixi-viewport";
+
+type ViewportProps = {
+  screenWidth?: number;
+  screenHeight?: number;
+  worldWidth: number;
+  worldHeight: number;
+  interaction: PIXI.InteractionManager | undefined;
```
0xdevalias, line 11:
```suggestion
  interaction?: PIXI.InteractionManager;
```

0xdevalias (commented):
Gave this a quick pass over to leave some comments/suggestions/pointers to help you align with how we do things, our codebase style, etc.

Learning about these things early on, and embedding them in your workflow will allow you to write code aligned to our standards from the very beginning, and will reduce the need for extra work/frustration/etc to try and align it later on.

Please share this among the rest of your engineering/dev team to ensure that everyone learns/benefits from the knowledge shared here.

Thanks.

(cc // @jonboutelle @sunny-viktoryia @mike-lvov FYI)

0xdevalias:
> sorry, I was late for the train and was in a hurry. All our branches forked from env/burn-staging and target on this branch.
> If this broken some github actions, you can reopen this correctly and close broken PR. I will be grateful to you.
> I will be available after 17pm (GMT+10)

@YarRainbow All good, didn't 'break anything' at all, more just confused my why that workflow ran. Thanks for letting me know the process you followed.. sounds like it might be something weird going on in our actions workflow then (might not be anything you've done 'wrong' with this PR at all). I'll make a mental note to look into it more later on and see if I can figure out the 'why' of it. Not something you need to look into. Thanks for the insight! :)

src/components/templates/AnimateMap/AnimateMap.tsx
```diff
@@ -0,0 +1,85 @@
+import * as PIXI from "pixi.js";
```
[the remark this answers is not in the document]
YarRainbow: This is a rudiment that I missed by relying on "optimize import" in IDE. We hid everything behind react-pixi abstractions. This string will just be deleted.

src/components/templates/AnimateMap/AnimateMap.scss
```diff
@@ -0,0 +1,29 @@
...
+  @include scrollbar;
+
+  &.no-scroll {
+    overflow: hidden;
+  }
+
+  &.modal-open {
+    padding: 0 0 0 0 !important;
+    overflow-y: scroll !important;
+  }
```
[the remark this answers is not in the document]
YarRainbow: We relied on the PartyMap template from Victoria's advice and a quick inspection showed that it was not known which modal the style was intended for. I left him for fear that he was doing some kind of specific work for the venue templates. If it just trash, confirm this please, and we remove these styles.

src/components/templates/AnimateMap/AnimateMap.scss
```diff
@@ -0,0 +1,29 @@
+@import "scss/constants.scss";
+
+.navbar-margin {
+  height: 100%;
+}
+
+.animatemap-venue-container {
```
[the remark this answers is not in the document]
YarRainbow: I love BEM! This methodology goes well with your atomic approach. I was confused again by PartyMap as an example. I will gladly correct that.

tsconfig.json
YarRainbow: I was pretty sure I removed this. Someone from the packages made this change to tsconfig. I haven't found the culprit and haven't even been able to reproduce the problem on my main machine, but at the moment this change is definitely not necessary. I'll remove this.

src/components/templates/AnimateMap/AnimateMap.tsx
YarRainbow: Usually we need a callback for "cleanup". But now we don't need it. Moreover, in one of the next sprints, we will one way or another take on a full-fledged resource management system and all this code will be rework.

src/components/templates/AnimateMap/AnimateMap.tsx
```diff
@@ -0,0 +1,85 @@
...
+  if (!user || !profile) {
+    return <>Loading..</>;
+  }
+
+  const container = containerRef.current ?? null;
+
+  return (
+    <div ref={containerRef} className="animatemap-venue-container">
+      {container && (
+        <ReactReduxContext.Consumer>
```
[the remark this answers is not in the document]
YarRainbow: @0xdevalias I consulted with Vasily about his decision. We are facing a known issue with react-pixi. We only use this wrapper to patch the provider. And then we use only hooks according to your recommendations.
We can encapsulate the wrapper in a separate component, but we don't see a way to get rid of it entirely.
If you have an idea how to fix this, I will be glad to consider it.
PS https://github.com/inlet/react-pixi/issues/77

src/components/templates/AnimateMap/AnimateMap.tsx
0xdevalias: See this for potentially related reason why this isn't currently using hooks: https://github.com/sparkletown/sparkle/pull/1617/files/6340eaee9d3913df6b50ab525569bfa35da9e5ea#r656036377

src/components/templates/AnimateMap/AnimateMap.tsx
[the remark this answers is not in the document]
0xdevalias: Makes sense, thanks for the context. Following some of those links I ended up at this section of the README:

- https://github.com/michalochman/react-pixi-fiber#react-context-api-limitations

This links to the following React issue:

- https://github.com/facebook/react/issues/13336

Which seems to have been closed by this PR:

- https://github.com/facebook/react/pull/13728

Which seems to have been included in the React `16.6.0` changelog in this PR:

- https://github.com/facebook/react/pull/13927

Since we seem to be using `16.13.1`, my guess/theory is that we should have that fix available to us:

- https://github.com/sparkletown/sparkle/blob/staging/package.json#L56

I'm not really sure of the specifics of how this issue works, how it was fixed, or how that applies to the lib you're using here, but it would be good if you could quickly see if it is resolved in a 'proper' way now, and if so, align the code here to that 'proper' way.

If it's going to take too long/be too hard/etc, then can you add the following comment (replacing the `XXX`'s with the relevant details) just before the `return` JSX so we have a note captured for future:

```
// @debt We need to use Redux's XXX here to work around a limitation in react-pixi-fibre as per:
//   https://github.com/michalochman/react-pixi-fiber#react-context-api-limitations
//
// While this issue was resolved in React in v16.6.0 via facebook/react#13728, we can't use that
//   fix here yet because XXX. Once that is resolved, we should refactor this accordingly.
```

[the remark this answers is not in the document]
0xdevalias: Also, if we still do need to hack around it like this, I wonder if we can still simplify/move the code here to a 'closer to Redux hooks' based implementation by making use of `useStore` instead of `ReactReduxContext.Consumer`, etc:

- https://react-redux.js.org/api/hooks#usestore
- https://react-redux.js.org/using-react-redux/accessing-store

src/components/templates/AnimateMap/AnimateMap.tsx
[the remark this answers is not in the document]
0xdevalias: Makes sense

src/components/templates/AnimateMap/AnimateMap.scss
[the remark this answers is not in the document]
0xdevalias: I'm also unsure off the top of my head. I expect this is likely leftover from old legacy outdated code.

---

`⇒  git log -S'modal-open' --oneline --name-status`
```
342d1348 fix room modal spacing in party map (#856)
M	src/components/templates/PartyMap/PartyMap.scss

7e50fa0e Run happy path on staging after deployment
A	cypress/integration/examples/cypress-register.spec.js
```

`⇒  git log -S'modal-open' -p | grep 'modal-open' -C5`
```diff
@@ -45,4 +45,9 @@ body {
     border-radius: 4px;
     background-color: rgba($white, 0.1);
   }
+
+  &.modal-open {
+    padding: 0 0 0 0 !important;
+    overflow-y: scroll !important;
+  }
 }

--
--
+
+    // cy.get(
+    //   ".private-chat-container > .chatbox-container > form > .chat-form > .chat-input"
+    // ).click();
+
+    // cy.get(".modal-open > .modal").click();
+  });
+});
```

`⇒  git log -S'modal-open' --patch`
```diff
commit 342d1348e6772284415e180e0a05cf3dc1349250
Author: Denis Dimitrov <dedimitrov@melontech.com>
Date:   Wed Dec 2 17:41:15 2020 +0200

    fix room modal spacing in party map (#856)

diff --git a/src/components/templates/PartyMap/PartyMap.scss b/src/components/templates/PartyMap/PartyMap.scss
index a9a816fd..6cf05f05 100644
--- a/src/components/templates/PartyMap/PartyMap.scss
+++ b/src/components/templates/PartyMap/PartyMap.scss
@@ -45,4 +45,9 @@ body {
     border-radius: 4px;
     background-color: rgba($white, 0.1);
   }
+
+  &.modal-open {
+    padding: 0 0 0 0 !important;
+    overflow-y: scroll !important;
+  }
 }
```

---

Looks like it was added to `PartyMap` in #856

And it seems to be part of Bootstrap's Modal styling:

- https://getbootstrap.com/docs/4.0/components/modal/#usage

It's still an issue in `src/components/templates/PartyMap/PartyMap.scss` as well.. so let's just leave it as is here for now. But can you please add a `@debt` comment above it:

```scss
// @debt refactor this to get rid of !important
&.modal-open {
```

src/components/templates/AnimateMap/AnimateMap.scss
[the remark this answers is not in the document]
0xdevalias: Unfortunately there's still a lot of legacy code patterns cluttering things up that we haven't had time to refactor/cleanup.. but slowly we're getting there!

tsconfig.json
0xdevalias: Ha, no worries, makes sense.

src/components/templates/AnimateMap/AnimateMap.tsx
0xdevalias: Fair, makes sense.

PR #1619, opened 2021-06-21
chat - place only author specific elements under onClick={openAuthorProfile}

### LINKED ISSUE #1618: Clicking on delete brings up a modal for the author
didn't check staging but on ongoing OHBM 2021 -- it seems to be impossible to delete a message in the chat.  Clocking on trashcan just brings up a modal with my name/profile.

yarikoptic:
So that delete below could have a chance to be clickable

Closes #1618 

applicable to staging as well most likely, didn't check

PS edit didn't check too hard, but clicking on trashcan seems to do have desired effect but message seems to not be removed from visible chat in the browser... but that would be a separate issue for someone to file if confirmed


mike-lvov:
Is unrelevant now

PR #1620, opened 2021-06-21
remove fetching all venues across the codebase

mike-lvov:
Replace it with `useRelativeVenues` instead


src/components/organisms/AppRouter/AppRouter.jsx
```diff
@@ -84,7 +87,15 @@ const AppRouter = () => {
...
-        <Route path="/in/:venueId" component={VenuePage} />
+
+        <Route
+          path="/in/:venueId"
+          render={(props) => (
+            <RelatedVenuesProvider venueId={props.match.params.venueId}>
+              <VenuePage />
+            </RelatedVenuesProvider>
+          )}
+        />
```
0xdevalias, line 98: Not sure that this functional router render pattern is the right way to go here. Also not sure about including it the in `AppRouter` like this in general.

src/components/templates/Playa/Playa.tsx
```diff
@@ -41,7 +41,7 @@ import { WithId } from "utils/id";
 // import { updateLocationData } from "utils/userLocation";
```
0xdevalias, line 41: [note] Playa is legacy code to be removed so we don't really care if it breaks here.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -145,19 +144,14 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
...
-    </RelatedVenuesProvider>
+    <ReactionsProvider venueId={venue.id}>
+      <WithNavigationBar fullscreen={fullscreen} hasBackButton={hasBackButton}>
+        <AnnouncementMessage message={venue.bannerMessage} />
+        {template}
+        <ChatSidebar venue={venue} />
+        <UserProfileModal venue={venue} />
+      </WithNavigationBar>
+    </ReactionsProvider>
   );
```
0xdevalias, line 155: Why was `RelatedVenuesProvider` removed from here to move it to `AppRouter`? Feels wrong to me.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -49,6 +48,7 @@ import { updateTheme } from "./helpers";
 import "./VenuePage.scss";
 
 import Login from "pages/Account/Login";
+import { useRelatedVenues } from "hooks/useRelatedVenues";
```
0xdevalias, line 51: Sort with other `hooks/*`

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -79,16 +79,17 @@ const VenuePage: React.FC = () => {
...
 
   const venueName = venue?.name ?? "";
   const venueTemplate = venue?.template;
 
   const event = currentEvent?.[0];
 
+  const { isRelatedVenuesLoading } = useRelatedVenues({
+    currentVenueId: venueId,
+  });
+
```
0xdevalias, line 92: It seems the only thing `useRelatedVenues` is used for in this component is `isRelatedVenuesLoading`. I'm not sure that that's something we need to necessarily have here..  and if we can remove it I expect that will mean that:

- we can move `RelatedVenuesProvider` back into `TemplateWrapper` where it used to be + where I prefer it
- it will most likely increase the perceived loading performance of the app, as we no longer will need to block the rest of the app based on this being available (and if there is something 'deeper down' that needs it, then best for that component to handle its own 'loading' rather than up here)

0xdevalias: Hopefully this is possible:
```suggestion
```

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -231,7 +232,7 @@ const VenuePage: React.FC = () => {
     }
   }
 
-  if (!user) {
+  if (!user || isRelatedVenuesLoading) {
```
0xdevalias, line 235: Hopefully this is possible:
```suggestion
  if (!user) {
```

src/utils/selectors.ts
```diff
@@ -71,51 +71,6 @@ export const worldUsersByIdSelector: SparkleSelector<
...
-export const makeVenueSelector = (venueId: string) => (
-  state: RootState
-): WithId<AnyVenue> | undefined => {
-  const venues = venuesSelector(state);
-
-  if (!venues.hasOwnProperty(venueId)) return undefined;
-
-  return { ...venues[venueId], id: venueId };
-};
-
```
0xdevalias, line 118: 🎉 💥 👋🏻

0xdevalias: Make sure to also remove these aspects from the core firestore types + any hooks/other places might have referred to them as well

src/pages/Admin/Admin.tsx
```diff
@@ -78,9 +76,11 @@ const VenueList: React.FC<VenueListProps> = ({
   roomIndex,
 }) => {
   // @debt This selector relies on all venues in firebase being loaded into memory.. not very efficient
-  const venues = useSelector(orderedVenuesSelector);
+  const { isRelatedVenuesLoading, relatedVenues } = useRelatedVenues({
+    currentVenueId: selectedVenueId,
+  });
 
-  if (!venues) return <>Loading...</>;
+  if (!isRelatedVenuesLoading) return <>Loading...</>;
```
0xdevalias, line 83: [nit] Replace this with `Loading` or `LoadingPage` components as appropriate.

src/pages/Admin/Admin.tsx
```diff
@@ -91,7 +91,7 @@ const VenueList: React.FC<VenueListProps> = ({
         </Link>
       </div>
       <ul className="page-container-adminsidebar-venueslist">
-        {venues.map((venue, index) => (
+        {relatedVenues.map((venue, index) => (
           <li
             key={index}
```
0xdevalias, line 96: [nit] Since we're here, it's a trivial change, and I believe `relatedVenues` provides the `venue.id` for us, we should change this `key` to use the `id` rather than the `index`.

src/pages/Admin/Admin.tsx
```diff
@@ -129,11 +129,9 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId, roomIndex }) => {
   const { url: matchUrl } = useRouteMatch();
   const { pathname: urlPath } = useLocation();
 
-  const venueSelector = useCallback(
-    (state) => makeVenueSelector(venueId)(state),
-    [venueId]
-  );
-  const venue = useSelector(venueSelector, shallowEqual);
```
0xdevalias, line 136: [note] I forgot that this 'custom comparator'/'shallowEqual' form of `useSelector` existed... I wonder if we can make use of that as part of a 'quickfix' hack to `worldUsers` / etc?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -18,14 +17,13 @@ import { AnyVenue, isVenueWithRooms } from "types/venues";
...
 import { useProfileModalControls } from "hooks/useProfileModalControls";
-import { useSelector } from "hooks/useSelector";
-import { useFirestoreConnect } from "hooks/useFirestoreConnect";
 import { useChatSidebarControls } from "hooks/chatSidebar";
 
 import { Badges } from "components/organisms/Badges";
 import Button from "components/atoms/Button";
 
 import "./UserProfileModal.scss";
+import { useRelatedVenues } from "hooks/useRelatedVenues";
```
0xdevalias, line 26: Sort with other `hooks/*`

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -141,25 +139,26 @@ const SuspectedLocation: React.FC<{
   currentVenue: WithId<AnyVenue>;
 }> = ({ user, currentVenue }) => {
   // @debt This will currently load all venues in firebase into memory.. not very efficient
```
0xdevalias, line 141: `@debt` comment no longer relevant once this PR lands 🎉 

```suggestion
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -141,25 +139,26 @@ const SuspectedLocation: React.FC<{
...
-  useFirestoreConnect("venues");
-  const allVenues = useSelector(orderedVenuesSelector);
+  const { relatedVenues } = useRelatedVenues({
+    currentVenueId: currentVenue.id,
+  });
 
   const suspectedLocation = useMemo(
     () => ({
-      venue: allVenues?.find(
+      venue: relatedVenues?.find(
```
0xdevalias, line 148: `relatedVenues` is guaranteed to always be an array:
```suggestion
      venue: relatedVenues.find(
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -141,25 +139,26 @@ const SuspectedLocation: React.FC<{
...
         (v) =>
           (user.lastSeenIn && user.lastSeenIn[currentVenue?.name ?? ""]) ||
           v.name === user.room
       ),
-      room: allVenues?.find(
+      room: relatedVenues?.find(
```
0xdevalias, line 153: `relatedVenues` is guaranteed to always be an array:
```suggestion
      venue: relatedVenues.find(
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -141,25 +139,26 @@ const SuspectedLocation: React.FC<{
...
       ),
     }),
-    [user, allVenues, currentVenue]
+    [user, relatedVenues, currentVenue]
   );
 
-  if (!user.room || !allVenues) {
+  if (!user.room) {
     return null;
   }
```
0xdevalias, line 163: [NB] I was going to ask if we still need this `return null`, but I think this entire `SuspectedLocation` component is legacy tech debt from the old `Playa` days that will be removed eventually. It's only rendered in this file when `ENABLE_SUSPECTED_LOCATION` is `true`, and it's hardcoded to `false`.

Can you please add an `@debt` comment just before the component definition, something like:

```typescript
/**
 * @debt I believe this is a legacy component leftover from the Playa template/similar outdated code, and should be removed as part of that cleanup
 * @deprecated legacy component, to be removed
 */
const SuspectedLocation: React.FC<{
```

src/pages/Admin/VenueList/VenueList.tsx
```diff
@@ -3,21 +3,21 @@ import { Link } from "react-router-dom";
...
   roomIndex,
 }) => {
-  // @debt This selector relies on all venues in firebase being loaded into memory.. not very efficient
-  const venues = useSelector(orderedVenuesSelector);
+  const { relatedVenues, isRelatedVenuesLoading } = useRelatedVenues({
+    currentVenueId: selectedVenueId,
+  });
 
-  if (!venues) return <>Loading...</>;
+  if (!isRelatedVenuesLoading) return <>Loading...</>;
```
0xdevalias, line 20: [NB] Would be good to replace this with `Loading` or `LoadingPage` as appropriate.

src/pages/Admin/VenueList/VenueList.tsx
```diff
@@ -28,7 +28,7 @@ const VenueList: React.FC<VenueListProps> = ({
         </Link>
       </div>
       <ul className="page-container-adminsidebar-venueslist">
-        {venues.map((venue, index) => (
+        {relatedVenues.map((venue, index) => (
           <li
             key={index}
```
0xdevalias, line 33: `relatedVenues` will always have the `id` available, so we may as well use it here instead of the antipattern of using the `index` as the `key`:
```suggestion
        {relatedVenues.map((venue) => (
          <li
            key={venue.id}
```

src/hooks/useRoom.ts
```diff
@@ -3,13 +3,12 @@ import { useCallback, useMemo } from "react";
...
 
 import { enterExternalRoom } from "utils/userLocation";
-import { orderedVenuesSelector } from "utils/selectors";
 import { enterVenue } from "utils/url";
 import { getExternalRoomSlug } from "utils/room";
 
-import { useSelector } from "hooks/useSelector";
 import { useRecentLocationUsers } from "hooks/users";
 import { useUser } from "hooks/useUser";
+import { useRelatedVenues } from "./useRelatedVenues";
```
0xdevalias, line 11: Use canonical import path + alphabetical sorting:

```suggestion
import { useRelatedVenues } from "hooks/useRelatedVenues";
import { useRecentLocationUsers } from "hooks/users";
import { useUser } from "hooks/useUser";
```

src/hooks/useRoom.ts
```diff
@@ -22,12 +21,11 @@ export const useRoom = ({ room, venueName }: UseRoomProps) => {
 
   const roomUrl = room?.url ?? "";
 
-  // @debt This selector relies on all venues in firebase being loaded into memory.. not very efficient
-  const venues = useSelector(orderedVenuesSelector);
+  const { relatedVenues } = useRelatedVenues({});
```
0xdevalias, line 24: This seems wrong? I suspect it was done because you have no need for 'current venue', yet the props are required (even though `currentVenueId` is optional)?

---

[potentially-OOS] I was going to suggest that we should just pass in the `venue.id` since it was in-scope anyway, but it seems like we end up finding the venue based on this data, so that wouldn't work.

I was also going to suggest that we should potentially refactor `useRoom` to accept `venueId` rather than `venueName`, but it's used `~5` locations across the app currently, and at least 2 of them don't immediately have access to the `venue` directly (only `venueName`) which could complicate this further.. though looking at those cases:

- `EventModal`: `src/components/organisms/EventModal/EventModal.tsx`
  - This seems to have `event.venueId` though, which is all we would need.. and this is already passed to `useRelatedVenues` to get `eventVenue`, so this is fine.
- `RoomModalContent`: `src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx`
  - This is only rendered by `RoomModal`, which already has `venue` present, so it could just pass on the `venue.id` rather than `venue.name`.

So it seems like it wouldn't be all that hard to do, and would make `useRoom` more canonical to our standards, while also avoiding this weird hacky looking syntax here at the moment.

If you're going to call OOS, then at least add 2 `@debt` comments, one to refactor `useRoom` to take `venueId` instead of `venueName` (at the top of this hook), and one on this line to say to pass `venueId` through to `useRelatedVenues` when we do; and then also please create a backlog issue that references this snippet + details for future.

---

[OOS] It would be a tiny change to make the props optional so we don't need to hack around it like this.. but that would seemingly require changing `ReactHook` + ensuring everywhere that uses that would work properly in that case.. so maybe best to leave that to a future problem.

src/hooks/useRoom.ts
```diff
@@ -22,12 +21,11 @@ export const useRoom = ({ room, venueName }: UseRoomProps) => {
...
 
   const roomVenue = useMemo(
-    () => venues?.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
-    [roomUrl, venues]
+    () => relatedVenues?.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
```
0xdevalias, line 27: `relatedVenues` will always be a defined array, so no need for `?`:

```suggestion
    () => relatedVenues.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
```

src/pages/Admin/Admin_v2.tsx
```diff
@@ -28,13 +26,13 @@ const Admin_v2: React.FC = () => {
   useAdminVenues(user?.uid);
 
   // @debt This selector relies on all venues in firebase being loaded into memory.. not very efficient
```
0xdevalias, line 28: This `@debt` comment is no longer relevant with the changes in this PR.
```suggestion
```

src/pages/Admin/Admin_v2.tsx
```diff
@@ -28,13 +26,13 @@ const Admin_v2: React.FC = () => {
...
-  const venues = useSelector(orderedVenuesSelector);
+  const { relatedVenues, isRelatedVenuesLoading } = useRelatedVenues({});
```
0xdevalias, line 29: I don't think this usage is valid here actually.. I believe the usage here needs to fetch all venues that this user is going to be able to have access to (would have to check code to be certain), not 'related venues'.

0xdevalias (changes requested):
Thoughts/comments/changes/etc.

I'm not sure that any of the usages in the admin components/etc will work currently as this code is, because there is no `RelatedVenuesProvider` in-scope for them (and to do so requires being able to give it a specific `venueId`).

I haven't looked too deeply into things.. but I expect that most if not all usages of it across the admin components may be wrong/not a good usage of it. It wasn't really written with the admin panel in mind, and I suspect in most cases it's just going to lead to needlessly overfetching data.

src/components/organisms/AppRouter/AppRouter.jsx
denisdimitrov: Maybe VenuePage can have a container component and all of this declerative code can be moved there. That way this will still keep the `component` router syntax.

src/components/templates/Playa/Playa.tsx
```diff
@@ -126,10 +126,12 @@ const isPlaced = (venue: AnyVenue) => {
...
 const minZoom = () => (window.innerWidth - 2 * PLAYA_MARGIN_X) / PLAYA_WIDTH;
 
+const venues = [] as WithId<AnyVenue>[];
+
 const Playa = () => {
   // @debt This will currently load all venues in firebase into memory.. not very efficient
-  useFirestoreConnect("venues");
-  const venues = useSelector(orderedVenuesSelector);
+  // useFirestoreConnect("venues");
+  // const venues = useSelector(orderedVenuesSelector);
```
denisdimitrov, line 134: Can we remove these, since we are 100% sure that this is a bad pattern and shouldn't be used, there is no point in having it here commented.

src/components/organisms/AppRouter/AppRouter.jsx
0xdevalias: That’s more or less what TemplateWrapper ends up as. It’s the bulk of stuff from VenuePage minus a bunch of legacy stuff IMO

src/components/templates/Playa/Playa.tsx
0xdevalias: This is the playa template. Not worth spending time on. It’s getting removed anyway. Commenting out is minimal effort while still vaguely explaining things

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: This line was the only reason why the `useRelatedVenues` hook was used in `VenuePage` and if there isn't some hidden specific reason for it being there, it's probaby best to accept the change request.

sunny-viktoryia:
@alexey-petrov, please, fix the `prettier` issues and let me know when the PR is ready for re-review.
image

sunny-viktoryia:
@alexey-petrov, not sure why you do force-pushes instead of regular push, but I guess it'd be better if you haven't changed the history of your commits as there is a sub-pr that is dependent on yours. 
https://github.com/sparkletown/sparkle/pull/1713

and I guess it has to be force-pushed after yours as well or possible can cause some difficulties resolving merge conflicts.

goran-peoski-work:
@sunny-viktoryia  I have resolved merge conflicts, and can push proper https://github.com/sparkletown/sparkle/pull/1713 albeit possibly with some more unrelated commits along the ride. So, to facilitate easier review, I have created separate branch with just PR related changes
https://github.com/sparkletown/sparkle/pull/1713#issuecomment-876451459 and the only missing will be probably changes done in the merges themselves.

@0xdevalias Suggested it to be sub-PR, but at this point, I'm like of a mind to ditch 1620 and merge 1713 into staging, just to cut the red tape.

sunny-viktoryia:
@alexey-petrov @goran-peoski-work @mike-lvov what's the status of this PR? not sure if it's ready for review.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
goran-peoski-work: Added in 1e8fb879bcd345ad7c00ceb417ac1dfa78c5bbff

src/components/organisms/AppRouter/AppRouter.jsx
goran-peoski-work: Superseded by `AdminRouter.tsx`'s `Provided` wrapper in 5f2ba2663620de1631552bd47e3d59a3dcab5ce2

```
          <Route path="/in/:venueId">
            <Provided withWorldUsers withRelatedVenues>
              <VenuePage />
            </Provided>
          </Route>
```

src/pages/VenuePage/TemplateWrapper.tsx
goran-peoski-work: Superseded by `AdminRouter.tsx`'s `Provided` wrapper in 5f2ba2663620de1631552bd47e3d59a3dcab5ce2

```
          <Route path="/in/:venueId">
            <Provided withWorldUsers withRelatedVenues>
              <VenuePage />
            </Provided>
          </Route>
```

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: Done in 4908ac4abd39131d79fdaedea96781e81ebae0cb

src/hooks/useRoom.ts
goran-peoski-work: Added `@debt` comments in 9b83f5d27a1ee3495619618af0a00b9f28337ab7 and opened a stub of an issue https://github.com/sparkletown/internal-sparkle-issues/issues/911

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: import removed in 4908ac4abd39131d79fdaedea96781e81ebae0cb

src/utils/selectors.ts
goran-peoski-work: As of 35d7c118100f97df09a60af2d491344e9e364a93 selectors can only be found in 2 comments inside `Playa.tsx`
```
import {
  currentVenueSelectorData,
  // orderedVenuesSelector,
} from "utils/selectors";
```
```
const Playa = () => {
  // @debt This will currently load all venues in firebase into memory.. not very efficient
  // useFirestoreConnect("venues");
  // const venues = useSelector(orderedVenuesSelector);
```

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: Removed in 4908ac4abd39131d79fdaedea96781e81ebae0cb

src/pages/Admin/Admin.tsx
goran-peoski-work: done 06fc69c809796e1f2f134b311c6fe8f756ee5d4d

src/components/organisms/UserProfileModal/UserProfileModal.tsx
goran-peoski-work: done 1e8fb879bcd345ad7c00ceb417ac1dfa78c5bbff

src/components/organisms/UserProfileModal/UserProfileModal.tsx
goran-peoski-work: removed in 78346abc

src/components/organisms/UserProfileModal/UserProfileModal.tsx
goran-peoski-work: done 1e8fb879bcd345ad7c00ceb417ac1dfa78c5bbff

src/components/organisms/UserProfileModal/UserProfileModal.tsx
goran-peoski-work: done 1e8fb879bcd345ad7c00ceb417ac1dfa78c5bbff

src/pages/Admin/VenueList/VenueList.tsx
goran-peoski-work: replaced with `<Loading />` in 4944f3417cc3a961b2743f9c4db7316d816602dc

src/pages/Admin/Admin.tsx
goran-peoski-work: replaced with `<Loading />` in 4944f3417cc3a961b2743f9c4db7316d816602dc

src/pages/Admin/Admin_v2.tsx
```diff
@@ -50,7 +47,7 @@ const Admin_v2: React.FC = () => {
     <>
       <S.Wrapper className="no-venue-selected">
         <S.ViewWrapper>
-          <AdminVenues venues={venues} />
+          <AdminVenues venues={relatedVenues} />
```
denisdimitrov, line 50: This should be a list of partymap venues that belong to you (you are owner). I don't think there is a hook or something else that returns such list, that's why we had all of the venues passed in and then the component `AdminVenues` filters them.

If you are to adapt `relatedVenues` to pass these owner venues you should also add `withRelatedVenues` to the subrouter.

src/pages/Admin/VenueList/VenueList.tsx
```diff
@@ -3,21 +3,23 @@ import { Link } from "react-router-dom";
 
 import { isVenueWithRooms } from "types/venues";
 
-import { orderedVenuesSelector } from "utils/selectors";
 import { canHaveSubvenues } from "utils/venue";
 
-import { useSelector } from "hooks/useSelector";
+import { useRelatedVenues } from "hooks/useRelatedVenues";
```
denisdimitrov, line 8: This component is not being imported or used anywhere. `Admin.tsx` is using `VenueList` that's already defined inside it's own file.

src/pages/Admin/Admin.tsx
```diff
@@ -91,9 +92,9 @@ const VenueList: React.FC<VenueListProps> = ({
         </Link>
       </div>
       <ul className="page-container-adminsidebar-venueslist">
-        {venues.map((venue, index) => (
+        {relatedVenues.map((venue, index) => (
```
denisdimitrov, line 95: The same comment for Admin NG applies here as well, except that these should be all venues that you are owner of, not only partymap.

denisdimitrov (changes requested):

src/hooks/usePreloadAssets.ts
```diff
@@ -0,0 +1,41 @@
...
+    link.type = type;
+  }
+
+  return link;
+};
+
+export const usePreloadAssets = (assets: PreloadAsset[]) => {
+  const headRef = useRef<HTMLHeadElement>(document.head);
+
+  return useEffect(() => {
```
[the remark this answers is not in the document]
goran-peoski-work: It just makes `usePreloadAssets()` a de facto `useEffect()`.
Nothing special, since there isn't requirement to manipulate the cleanup callback from the inner hook outside the outer one.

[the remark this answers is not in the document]
goran-peoski-work: removed in 6067f321184ada9bc10ccb2158c32d07389ce3b7 as to not confuse people

src/pages/Admin/VenueList/VenueList.tsx
goran-peoski-work: removed in cebe716d91ee6f1f6f15bcf2cadeecbc2c31a060

jonboutelle:
Do this after main github (thurs fri)

src/pages/Admin/Admin_v2.tsx
goran-peoski-work: Replaced with `useAdministeredVenues()` in f97606d53e7ceba1ab4ad0433711a96c16d5b7ea

src/pages/Admin/Admin_v2.tsx
goran-peoski-work: Replaced with `useAdministeredVenues()` that returns `administeredVenues` instead of `relatedVenues`  in f97606d53e7ceba1ab4ad0433711a96c16d5b7ea

src/pages/Admin/Admin.tsx
goran-peoski-work: Replaced with `useAdministeredVenues()` that returns `administeredVenues` instead of `relatedVenues`  in f97606d53e7ceba1ab4ad0433711a96c16d5b7ea

src/hooks/usePreloadAssets.ts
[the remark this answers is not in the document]
0xdevalias: Since useEffect doesn’t ever return anything to my knowledge, not having the return is correct.

src/pages/Admin/Admin_v2.tsx
0xdevalias: Can we Calle it useOwnedVenues or similar please

goran-peoski-work: Makes no difference to me if `useOwnedVenues` or `useAdministeredVenues`, I was following the `useAdminVenues` prior art which  https://github.com/sparkletown/sparkle/pull/1620/commits/103baa1a25f079ee2d0ccf9158c295a66ee594c9 removes as redundant.

Just in case, will RFC @denisdimitrov before doing the rename.

goran-peoski-work: renamed in ef8651a789a447dbfc954eb4e516ff7d298adeea

src/components/templates/Playa/Playa.tsx
```diff
@@ -40,7 +40,7 @@ import { WithId } from "utils/id";
 // import { updateLocationData } from "utils/userLocation";
```
mike-lvov, line 40: remove all of the commented code, since we are not going to uncomment it in the future. Please, do it to all of the occurances

src/components/templates/Playa/Playa.tsx
```diff
@@ -527,26 +529,23 @@ const Playa = () => {
...
-          closestVenue = venue;
-          distanceToClosestVenue = distance;
-        }
-      });
-      return closestVenue;
-    },
-    [venues]
-  );
+  const getNearbyVenue = useCallback((x: number, y: number) => {
+    if (!venues) return;
```
mike-lvov, line 533:
```suggestion
    if (!venues) return;
    
```

src/components/templates/Playa/Playa.tsx
```diff
@@ -527,26 +529,23 @@ const Playa = () => {
...
+    let closestVenue: WithId<AnyVenue> | undefined;
+    let distanceToClosestVenue: number;
```
mike-lvov, line 535:
```suggestion
    let distanceToClosestVenue: number;
    
```

src/components/templates/Playa/Playa.tsx
```diff
@@ -527,26 +529,23 @@ const Playa = () => {
...
+        distance &&
+        distance <= VENUE_NEARBY_DISTANCE &&
+        (!distanceToClosestVenue || distance < distanceToClosestVenue)
+      ) {
+        closestVenue = venue;
+        distanceToClosestVenue = distance;
+      }
+    });
+    return closestVenue;
+  }, []);
```
mike-lvov, line 548:
```suggestion
    venues.forEach((venue) => {
      const distance = distanceToVenue(x, y, venue.placement);
      
      if (
        isTruthy(distance) &&
        distance <= VENUE_NEARBY_DISTANCE &&
        (!distanceToClosestVenue || distance < distanceToClosestVenue)
      ) {
        closestVenue = venue;
        distanceToClosestVenue = distance;
      }
    });
    
    return closestVenue;
  }, []);
```

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { AnyVenue } from "types/venues";
+
+import { WithId, withId } from "utils/id";
+
+import { useFirestoreConnect, isLoaded } from "hooks/useFirestoreConnect";
+import { useSelector } from "hooks/useSelector";
+
+export const ownedVenuesSelector: SparkleSelector<
+  Record<string, AnyVenue> | undefined
+> = (state) => state.firestore.data.ownedVenues;
```
mike-lvov, line 14: move into the selectors

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,125 @@
...
+    () => Object.entries(venues ?? {}).map(([id, venue]) => withId(venue, id)),
+    [venues]
+  );
+
+  const ownedVenuesIds = useMemo(() => Object.keys(venues ?? {}), [venues]);
+
+  const findVenueInOwnedVenues = useCallback(
+    (searchedForVenueId: string): WithId<AnyVenue> | undefined =>
+      ownedVenues.find((venue) => venue.id === searchedForVenueId),
+    [ownedVenues]
```
mike-lvov, line 66: If we are using useFirestoreConnect, there is no need in Context/Provider. This was done for users only because it was the first step to improve performance. After that the useConnect was removed from there

mike-lvov: This should be done similar to `auditorium/useAllAuditoriumSections`

src/hooks/usePreloadAssets.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useRef } from "react";
+
+import { isDefined } from "utils/types";
+
+export interface PreloadAsset {
+  url: string;
+  as?: string;
+  type?: string;
```
mike-lvov, line 8: I guess these can be more strictly defined?

mike-lvov (commented):
A few more additions and good to go

src/hooks/usePreloadAssets.ts
goran-peoski-work: Not really any benefit in it.

It's passed directly to the HTML element (declared `interface HTMLLinkElement extends HTMLElement, LinkStyle`) which has them as strings.

And I would consider it OOS to have one more think to keep track of such a thing as constraining something so basic it wasn't constrained even in the standard library.

mike-lvov: @goran-peoski-work if they are defined as strings in the HTML default types, then It's all good. Thanks for the research

src/components/templates/Playa/Playa.tsx
mike-lvov: @goran-peoski-work if you think something is irrelevant, please say it out loud. I've just noticed that it's the outdated Playa template, so this comment is irrelevant, but you shouldn't resolve the comments silently

src/components/templates/Playa/Playa.tsx
goran-peoski-work: removed d0e862618e8e3afacd238ae21284bf4230067a79

src/components/templates/Playa/Playa.tsx
mike-lvov: @goran-peoski-work if you think something is irrelevant, please say it out loud. I've just noticed that it's the outdated Playa template, so this comment is irrelevant, but you shouldn't resolve the comments silently

src/components/templates/Playa/Playa.tsx
mike-lvov: @goran-peoski-work if you think something is irrelevant, please say it out loud. I've just noticed that it's the outdated Playa template, so this comment is irrelevant, but you shouldn't resolve the comments silently

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: moved 8819037c8d4dc531a31279aa15a3e1979048a188

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: removed the Context/Provider 8819037c8d4dc531a31279aa15a3e1979048a188

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,78 @@
...
+> = ({ currentVenueId }): useOwnedVenuesData => {
+  const { userId } = useUser();
+  useConnectOwnedVenues(userId);
+
+  const venues = useSelector(ownedVenuesSelector);
+
+  const ownedVenues = useMemo(
+    () => Object.entries(venues ?? {}).map(([id, venue]) => withId(venue, id)),
+    [venues]
+  );
```
mike-lvov, line 54: ownedVenuesSelector should use .ordered, not .data.
you should create one more selector. ownedVenuesDataSelector which would return data

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const findVenueInOwnedVenues = useCallback(
+    (searchedForVenueId: string): WithId<AnyVenue> | undefined =>
+      ownedVenues.find((venue) => venue.id === searchedForVenueId),
+    [ownedVenues]
+  );
```
mike-lvov, line 60: use ownedVenuesDataSelector here and get O(1) complexity instead of O(N)

src/utils/selectors.ts
```diff
@@ -254,3 +209,7 @@ export const noopSelector: SparkleSelector<undefined> = () => undefined;
 
 export const emptyArray = [];
 export const emptyArraySelector = <T>(): T[] => emptyArray;
+
+export const ownedVenuesSelector: SparkleSelector<
+  Record<string, AnyVenue> | undefined
+> = (state) => state.firestore.data.ownedVenues;
```
mike-lvov, line 215:
```suggestion
export const ownedVenuesSelector: SparkleSelector<
  Record<string, AnyVenue> | undefined
> = (state) => state.firestore.ordered.ownedVenues;
```

Also, add the data selector

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,78 @@
...
+          where: [["owners", "array-contains", userId]],
+          storeAs: "ownedVenues",
+        }
+      : undefined
+  );
+};
+
+export interface useOwnedVenuesProps {
+  currentVenueId?: string;
+}
```
mike-lvov, line 26:
```suggestion
export interface UseOwnedVenuesProps {
  currentVenueId?: string;
}
```

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,78 @@
...
+  currentVenue?: WithId<AnyVenue>;
+
+  ownedVenues: WithId<AnyVenue>[];
+  ownedVenuesIds: string[];
+
+  findVenueInOwnedVenues: (
+    searchedForVenueId: string
+  ) => WithId<AnyVenue> | undefined;
+}
+
```
mike-lvov, line 41:
```suggestion
export interface UseOwnedVenuesData {
  isLoaded: boolean;
  isLoading: boolean;

  currentVenue?: WithId<AnyVenue>;

  ownedVenues: WithId<AnyVenue>[];
  ownedVenuesIds: string[];

  findVenueInOwnedVenues: (
    searchedForVenueId: string
  ) => WithId<AnyVenue> | undefined;
}

```

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,78 @@
...
+    [ownedVenues]
+  );
+
+  return useMemo(
+    () => ({
+      isLoaded: isLoaded(venues),
+      isLoading: !isLoaded(venues),
+
+      ownedVenues,
+      ownedVenuesIds: Object.keys(venues ?? {}),
```
mike-lvov, line 68: useMemo

mike-lvov (commented):
A few last changes and good to ship

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: done 91bb8cbdd6f157e7c45f103389ebc034a5cdfa35

src/utils/selectors.ts
goran-peoski-work: now with 2 selectors 91bb8cbdd6f157e7c45f103389ebc034a5cdfa35

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: I should've caught this when I copied the code. Changed in 91bb8cbdd6f157e7c45f103389ebc034a5cdfa35

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: using `ownedVenuesDataSelector` now 91bb8cbdd6f157e7c45f103389ebc034a5cdfa35

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  findVenueInOwnedVenues: (
+    searchedForVenueId: string
+  ) => WithId<AnyVenue> | undefined;
+}
+
+export const useOwnedVenues: ReactHook<
+  useOwnedVenuesProps,
+  useOwnedVenuesData
+> = ({ currentVenueId }): useOwnedVenuesData => {
```
mike-lvov:
```suggestion
export const useOwnedVenues: ReactHook<
  UseOwnedVenuesProps,
  UseOwnedVenuesData
> = ({ currentVenueId }) => {
```

src/hooks/useConnectOwnedVenues.tsx
```diff
@@ -0,0 +1,88 @@
...
+    [venues]
+  );
+
+  const findVenueInOwnedVenues = useCallback(
+    (searchedForVenueId: string) => {
+      const found = venues?.[searchedForVenueId];
+      return found ? withId(found, searchedForVenueId) : undefined;
+    },
+    [venues]
+  );
```
mike-lvov, line 64:
```suggestion
  const ownedVenuesData = useSelector(ownedVenuesDataSelector);
  const ownedVenues = useSelector(ownedVenuesSelector);
  
  const ownedVenuesIds = useMemo(() => Object.keys(venues ?? {}), [venues]);

  const getVenueInOwnedVenues = useCallback(
    (searchedForVenueId: string) => {
      const foundVenue = venues?.[searchedForVenueId];
      return foundVenue ? withId(found, searchedForVenueId) : undefined;
    },
    [venues]
  );
```

mike-lvov (commented):
Approved, once reused the selector properly. I can't approve my own PR, but I'll ask @jonboutelle to do it

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: this got fixed, but was too tired to notice I didn't type `git push` 354fa97f162e3211548b0d13dfaf429aa7932460

src/hooks/useConnectOwnedVenues.tsx
goran-peoski-work: ` useSelector(ownedVenuesSelector);` doesn't return `WithId<>` objects I think it's better not to proliferate non-id values

mike-lvov: @goran-peoski-work it should return `WithId`

image

jonboutelle (approved):
gogogo

PR #1621, opened 2021-06-22
load venue events from pre-cached content stored in gs://

soichih:
To load the venue calendar, sparkle currently must download all venues (which takes about 30-45 seconds for OHBM), followed by downloading all events from all venues (takes another 10-15 seconds). In total, user must sit and wait for about 40-60 seconds before they can see the schedule. Most user won't wait that long.

This PR replaces the call to firebase to download venues and venues/events, and instead, it loads from the cached json file stored in gs://cache/events.json (or wherever we set it to). By loading from the gs://, the calendar can be opened almost as soon as the main venue opens. 

The cache is uploaded from a server that queries all venues/events and we can run this as frequently as we need - during our conference. Any server with firebase admin sdk configured can do the caching (I've already set it up on my jetstream VM). 

I tried to make this feature optional by setting `SCHEDULE_LOADFROM_GS`, but I couldn't quite get the useVenueEvent function to use `useAsync` conditionally. When we are loading from `events.json` it does not require venueIds. So we have to make this somewhat dynamic.

### DIFF
--- a/src/hooks/events.ts
+++ b/src/hooks/events.ts
@@ -10,7 +10,7 @@ import { WithId, WithVenueId } from "utils/id";
 //import { tracePromise } from "utils/performance";
 import { useFirebase } from "react-redux-firebase";
 
-import { SCHEDULE_LOADFROM_GS } from "settings";
+import { SCHEDULE_LOAD_FROM_GS } from "settings";
 
 const emptyArray: never[] = [];
 
@@ -40,7 +40,7 @@ export const useVenueEvents: ReactHook<VenueEventsProps, VenueEventsData> = ({
     const storage = firebase.storage();
     const url = await storage
       .ref()
-      .child(SCHEDULE_LOADFROM_GS)
+      .child(SCHEDULE_LOAD_FROM_GS)
       .getDownloadURL();
     return fetch(url).then((res) => res.json());
 
--- a/src/hooks/events.ts
+++ b/src/hooks/events.ts
@@ -1,12 +1,16 @@
 import { useAsync } from "react-use";
 
-import { fetchAllVenueEvents } from "api/events";
+//import { fetchAllVenueEvents } from "api/events";
 
 import { ReactHook } from "types/utility";
 import { VenueEvent } from "types/venues";
 
 import { WithId, WithVenueId } from "utils/id";
-import { tracePromise } from "utils/performance";
+
+//import { tracePromise } from "utils/performance";
+import { useFirebase } from "react-redux-firebase";
+
+import { SCHEDULE_LOADFROM_GS } from "settings";
 
 const emptyArray: never[] = [];
 
@@ -25,13 +29,24 @@ export interface VenueEventsData {
 export const useVenueEvents: ReactHook<VenueEventsProps, VenueEventsData> = ({
   venueIds,
 }) => {
+  const firebase = useFirebase();
+
   const {
     loading: isEventsLoading,
     error: eventsError,
     value: events = emptyArray,
   } = useAsync(async () => {
-    if (!venueIds) return emptyArray;
+    //load from gS
+    const storage = firebase.storage();
+    const url = await storage
+      .ref()
+      .child(SCHEDULE_LOADFROM_GS)
+      .getDownloadURL();
+    return fetch(url).then((res) => res.json());
 
+    //load from firebase - depends on venueIds
+    /*
+    if (!venueIds) return emptyArray;
     return tracePromise(
       "useVenueEvents::fetchAllVenueEvents",
       () => fetchAllVenueEvents(venueIds),
@@ -41,8 +56,8 @@ export const useVenueEvents: ReactHook<VenueEventsProps, VenueEventsData> = ({
         },
       }
     );
-  }, [venueIds]); // TODO: figure out this deps in an efficient way so it doesn't keep re-rendering
-
+    */
+  }, [firebase /*venueIds*/]);
   return {
     isEventsLoading,
     isError: eventsError !== undefined,
--- a/src/hooks/events.ts
+++ b/src/hooks/events.ts
@@ -1,16 +1,14 @@
 import { useAsync } from "react-use";
 
-//import { fetchAllVenueEvents } from "api/events";
-
 import { ReactHook } from "types/utility";
 import { VenueEvent } from "types/venues";
 
 import { WithId, WithVenueId } from "utils/id";
 
 //import { tracePromise } from "utils/performance";
-import { useFirebase } from "react-redux-firebase";
+//import { fetchAllVenueEvents } from "api/events";
 
-import { SCHEDULE_LOAD_FROM_GS } from "settings";
+import { useSelector } from "hooks/useSelector";
 
 const emptyArray: never[] = [];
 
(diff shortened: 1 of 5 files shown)


soichih:
note to myself.. I believe the bookmark button won't work until all venues are loaded?

src/hooks/events.ts
```diff
@@ -1,12 +1,16 @@
...
-import { fetchAllVenueEvents } from "api/events";
+//import { fetchAllVenueEvents } from "api/events";
 
 import { ReactHook } from "types/utility";
 import { VenueEvent } from "types/venues";
 
 import { WithId, WithVenueId } from "utils/id";
-import { tracePromise } from "utils/performance";
+
+//import { tracePromise } from "utils/performance";
```
0xdevalias, line 10: Commenting these out directly isn't making use of `SCHEDULE_LOADFROM_GS`?

src/settings.ts
```diff
@@ -624,6 +624,7 @@ export const SCHEDULE_SHOW_DAYS_AHEAD = 7;
 export const FIRESTORE_QUERY_IN_ARRAY_MAX_ITEMS = 10;
 
 export const REMOVE_EVENTS_FROM_VENUE = /^poster[0-9]{4}$/;
+export const SCHEDULE_LOADFROM_GS = "cache/events.json";
```
0xdevalias, line 627:
```suggestion
export const SCHEDULE_LOAD_FROM_GS = "cache/events.json";
```

src/hooks/events.ts
```diff
@@ -25,13 +29,24 @@ export interface VenueEventsData {
...
     error: eventsError,
     value: events = emptyArray,
   } = useAsync(async () => {
-    if (!venueIds) return emptyArray;
+    //load from gs
+    const storage = firebase.storage();
+    const url = await storage
+      .ref()
+      .child(SCHEDULE_LOADFROM_GS)
+      .getDownloadURL();
```
0xdevalias, line 44: Are we not going to need to grant some kind of 'is allowed to read storage' permissions in the rules/etc for this to work? I don't see any changes to `storage.rules` in this PR?

src/hooks/events.ts
```diff
@@ -25,13 +29,24 @@ export interface VenueEventsData {
 export const useVenueEvents: ReactHook<VenueEventsProps, VenueEventsData> = ({
```
0xdevalias, line 29: `useVenueEvents` is called in many places across the platform, and the events it fetches/returns depend on the `venueId`s provided. As is currently implemented here, that functionality isn't included/handled at all.

0xdevalias (changes requested):
Comments/changes/etc. As is, this isn't sufficient to replace the existing functionality without risking creating further issues/breaking things.

---

> The cache is uploaded from a server that queries all venues/events and we can run this as frequently as we need - during our conference. Any server with firebase admin sdk configured can do the caching (I've already set it up on my jetstream VM).

How is this done? Traditionally that would be through a script/similar. It would be good if that was included as part of the PR given the changes here directly rely on and necessitate having it/being able to run it.

src/hooks/events.ts
yarikoptic: Code which used them is commented out and replaced with use of cache below

src/hooks/events.ts
yarikoptic: So someone should check if filtering the events by venueId would still at the end would provide the speedup overall I guess.

Note that this PR is just an attempt to help to make sparkle usable for the ongoing event. It is provided as a prototype for you, sparkle folks, to try/improve etc, to save the event. If you see something to improve, just join/push.

src/hooks/events.ts
yarikoptic: check if needed

src/hooks/events.ts
yarikoptic: also might be worth adding a condition and use cached version only if many venueIds are requested, and if not -- resort to non-cached version

src/hooks/events.ts
0xdevalias: @yarikoptic I understand that, I can read code :) Yet this PR explicitly states that `SCHEDULE_LOADFROM_GS` was added so it could be used to conditionally use the old method or the new. Which isn't the case here since it's commented out. Thus why I called it out.

0xdevalias: You can't conditionally render a hook. But you can do things within the hook conditionally.

src/hooks/events.ts
0xdevalias: @yarikoptic I don't understand if you're making a note here for yourself, telling me to do something, or something else entirely?

src/hooks/events.ts
0xdevalias: > Note that this PR is just an attempt to help to make sparkle usable for the ongoing event. It is provided as a prototype for you, sparkle folks, to try/improve etc, to save the event. If you see something to improve, just join/push.

We're working on our own more specific/better performance fixes. I spent the time to review this PR out of good will and to see if it would help. In its current state, it's not really usable. It's not worth me diverting my efforts from proper bigger better fixes to try and piece together that half implementation provided here.

src/hooks/events.ts
soichih: I had to update the cors but you don't need any update for storage rules as far as I know. The way it's setup right now works just fine.

src/hooks/events.ts
soichih: I tried to make the dependency array conditional (and it worked) but then the linter didn't let me commit it. I don't know if it breaks something with react or not.. but the way it is written now works and it makes loading of the schedule much quicker.

src/hooks/events.ts
soichih: @0xdevalias Thank you for reviewing this PR. Until better fixes are available, would it be possible to merge this now and later replace it with something better?

src/hooks/events.ts
soichih: Actually, I needed to add this to the rule

```
    match /cache/{allPaths=**} {
        allow read
    }
```

I will move it under /assets so we don't have to add this.

src/hooks/events.ts
yarikoptic: @soichih wouldn't it need at least addition of filtering for the `venueIds` given to the call, and thus also just returning if none was requested?

src/hooks/events.ts
```diff
@@ -25,13 +29,24 @@ export interface VenueEventsData {
...
     value: events = emptyArray,
   } = useAsync(async () => {
-    if (!venueIds) return emptyArray;
+    //load from gs
+    const storage = firebase.storage();
+    const url = await storage
+      .ref()
+      .child(SCHEDULE_LOAD_FROM_GS)
+      .getDownloadURL();
+    return fetch(url).then((res) => res.json());
```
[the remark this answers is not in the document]
margulies: Also, is there a way to filter by `venueIds` array is it is available? An issue has come up in the posterhall where the live events aren't filtered to only those specified by parentId (== posterhall). Something like?: 

`events.filter((event) => venueIds.includes(event.venueId))`

src/hooks/events.ts
soichih: Yes, that's true. I should filter them by requested venueIds. I will update

src/hooks/events.ts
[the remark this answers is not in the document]
soichih: Yes. I should filter them.. I just did that. I've also updated to store the cache in the redux store instead of relying on fetch() to return the cached output.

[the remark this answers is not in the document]
margulies: Thanks! Just tested, and seems to work smoothly in the posterhall now

soichih:
Once #1630 is merged, I believe this PR is not necessary. Loading of calendar still take a few seconds but it's quick enough that ppl could probably live with it.

PR #1627, opened 2021-06-22
[env/ohbm] increase user location update frequency from 5min to 120min

sunny-viktoryia:
Change user location update frequency from `5 mins` to `120 mins` for a (theoretically maybe) slight performance improvement.


0xdevalias (approved):
While I don't expect this to have a big performance improvement, it might provide some slight assistance.

LGTM! :shipit:

sunny-viktoryia:
@0xdevalias FYI, I have increased the update window to 120 mins as per @jonboutelle's request.

PR #1628, opened 2021-06-22
Introduce the script that adds venueId and sovereignId to events

sunny-viktoryia:
Introduce the script that adds venueId and sovereignId to events for a new events structure.

That'll help us fetch events without fetching all the related venues.

### DIFF
--- a/scripts/extend-events-with-venue-ids.ts
+++ b/scripts/extend-events-with-venue-ids.ts
@@ -1,0 +1,95 @@
+#!/usr/bin/env node -r esm -r ts-node/register
+
+import { resolve } from "path";
+
+import admin from "firebase-admin";
+
+import { initFirebaseAdminApp, makeScriptUsage } from "./lib/helpers";
+
+const usage = makeScriptUsage({
+  description: "Update events details: { venueId, sovereignVenueId }",
+  usageParams: "PROJECT_ID [CREDENTIAL_PATH]",
+  exampleParams: "co-reality-map [theMatchingAccountServiceKey.json]",
+});
+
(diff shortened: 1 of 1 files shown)


scripts/extend-events-with-venue-ids.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  if (maxDepth && maxDepth <= 0)
+    throw new Error("Maximum depth reached before finding the sovereignVenue.");
+
+  return fetchSovereignVenue(venue.data().parentId, venues, {
+    ...options,
+    previouslyCheckedVenueIds: [...previouslyCheckedVenueIds, venueId],
+    maxDepth: maxDepth ? maxDepth - 1 : undefined,
+  });
+};
```
0xdevalias, line 68: Why is this duplicated here? Why can't we just use the existing functions from `api/*` in the main app (we should be able to)?

scripts/extend-events-with-venue-ids.ts
```diff
@@ -0,0 +1,95 @@
...
+      const venueData = {
+        venueId,
+        sovereignVenueId,
+      };
+
+      console.log(`updating event "${eventDoc.data().name}"`, venueData);
+      eventDoc.ref.update(venueData);
+    });
+  });
+})();
```
0xdevalias, line 95: Instead of using `.forEach` and having to synchronously block each update/etc, we should make use of `.map`, `Promise.all`, firebase batched writes, potentially firebase transactions, etc.

There are examples of these patterns in our other scripts.

0xdevalias (commented):
Notes

PR #1629, opened 2021-06-22
add events fetch per one request

mike-lvov:
Fixes the issue with thousands of events requests.

NOTE: We should also create an index

image


0xdevalias:
@yarikoptic @soichih FYI

src/types/venues.ts
```diff
@@ -319,6 +319,7 @@ export interface VenueEvent {
   collective_price: number;
   host: string;
   room?: string;
+  venueId?: string;
```
0xdevalias, line 322: Is this going to be on every event always? We have a WithVenueId helper.

firestore.rules
```diff
@@ -48,13 +52,13 @@ service cloud.firestore {
       allow read: if request.auth.uid != null;
       allow write: if true;
     }
+    match /{path=**}/events/{event} {
+      allow read: if true;
+      allow write: if request.auth.uid in venueData().owners;
+    }
```
mike-lvov, line 58: NOTE: This needs to be explicitly set for groupQuery to work as per the docs
https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_collection_groups

firestore.rules
```diff
@@ -8,6 +8,10 @@ function role(name) {
   return get(/databases/$(database)/documents/roles/$(name)).data
 }
 
+function venueData() {
+  return get(/databases/$(database)/documents/venues/$(venue)).data
+}
```
mike-lvov, line 13: I needed to move it to use in a new rule

src/types/venues.ts
mike-lvov: we are going to prepopulate our data with it. So in this environment, yes

firestore.rules
```diff
@@ -48,6 +48,9 @@ service cloud.firestore {
       allow read: if request.auth.uid != null;
       allow write: if true;
     }
+    match /{path=**}/events/{event} {
+      allow read: if true;
+    }
```
mike-lvov, line 53: NOTE: This needs to be explicitly set for groupQuery to work as per the docs
https://firebase.google.com/docs/firestore/security/rules-query#secure_and_query_documents_based_on_collection_groups

0xdevalias: This should probably be `if request.auth.uid != null;` at a minimum. Currently any unauthenticated user on the internet could query this.

src/api/events.ts
```diff
@@ -2,32 +2,24 @@ import firebase from "firebase/app";
...
-): Promise<WithVenueId<WithId<VenueEvent>>[]> =>
-  Promise.all(asArray(venueIdOrIds).map(fetchVenueEvents)).then((result) =>
-    result.flat()
-  );
+  return eventsSnapshot.docs
+    .map((venueEvent) => venueEvent.data())
+    .filter((venueEvent) => venueEvent.venueId) as WithVenueId<
+    WithId<VenueEvent>
+  >[];
+};
```
0xdevalias, line 22: Ignoring the `fetchAllVenueEvents` part which is fine to remove. Why was the `.then` removed + separated like this? It should have just been left inline as it was by the looks of things.

src/hooks/events.tsx
```diff
@@ -0,0 +1,119 @@
...
+  venueId,
+}) => {
+  const {
+    sovereignVenueId,
+    isSovereignVenueIdLoading,
+
+    errorMsg: sovereignVenueIdError,
+  } = useSovereignVenueId({
+    venueId,
+  });
```
0xdevalias, line 70: `EventsProvider` is setup to need `venueId` passed to it. Yet where you instantiate it, you're not passing `venueId` to it at all?

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -146,17 +147,19 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
...
-          fullscreen={fullscreen}
-          hasBackButton={hasBackButton}
-        >
-          <AnnouncementMessage message={venue.bannerMessage} />
-          {template}
-          <ChatSidebar venue={venue} />
-          <UserProfileModal venue={venue} />
-        </WithNavigationBar>
-      </ReactionsProvider>
+      <EventsProvider>
```
0xdevalias, line 150: Where's the `venueId` prop on `EventsProvider` here?

src/types/venues.ts
0xdevalias: Makes sense.

0xdevalias:
@mike-lvov We should think about bringing this change + related out of `env/ohbm` into the main `staging` code branch + cleaning up some related manual aspects of it when we do.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: Seems this was fixed in https://github.com/sparkletown/sparkle/pull/1631

src/hooks/events.tsx
0xdevalias: Seems this was fixed in https://github.com/sparkletown/sparkle/pull/1631

0xdevalias:
> NOTE: We should also create an index

@mike-lvov An index on what/for what purpose? Doesn't this only use single fields? Firestore indexes all single fields by default.

PR #1632, opened 2021-06-23
add toggle to force hide room labels individually per room in admin (v1)

ermiaSparkle:
add label room to admin page
add a new field to rooms 

closes https://github.com/sparkletown/internal-sparkle-issues/issues/845

DRIFT-VIDEO-2610861-2699200-1624451986


src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -122,13 +122,28 @@ export const AdminVenueRoomDetails = ({
...
+                      name={"toggle2-" + index}
+                      checked={room.isLabeled}
+                      onClick={() => {
+                        updateRoom(!room.isLabeled, "isLabeled");
+                      }}
+                    />
+                    <span className="slider round"></span>
+                  </label>
+                  <div>Label: {room.isLabeled ? "On" : "Off"}</div>
+                </div>
```
0xdevalias, line 146: I realise that this is following the pattern of the code above it, but it's quite old/legacy.  Is it possible to use `src/components/atoms/Toggler/Toggler.tsx` here instead? That would be ideal.

I don't think we should need the `id`s here, since we nest the `input` within the `label` as per:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label
  - > Alternatively, you can nest the `<input>` directly inside the `<label>`, in which case the `for` and `id` attributes are not needed because the association is implicit

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -331,6 +331,19 @@ const RoomInnerForm: React.FC<RoomInnerFormProps> = (props) => {
...
+                      <input
+                        disabled={disable}
+                        type="checkbox"
+                        id="isLabeled"
+                        name={"isLabeled"}
+                        ref={register}
+                      />
+                      <span className="slider round"></span>
+                    </label>
+                  </div>
```
0xdevalias, line 346: Same comments as above RE: using `src/components/atoms/Toggler/Toggler.tsx` here.

src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -56,13 +56,13 @@ export const AdminVenueRoomDetails = ({
   const { user } = useUser();
   const history = useHistory();
 
-  const updateRoom = async (newState: boolean) => {
+  const updateRoom = async (newState: boolean, field: string) => {
```
0xdevalias, line 59: We don't really like to use multiple positional parameters like this. We should refactor this to take a 'props' object and use named keys here.

I'm also not sure that we should allow this to be a `string`, as that will make it too easy to accidentally overwrite things, and doesn't make the best use of our type system to guide us towards writing good code. As a minimal implementation here, we should change the type to `"isEnabled" | "isLabelled"`

Since we're changing it anyway, may as well rename this something better at the same time (eg. updateRoomBoolean

```suggestion
  export interface UpdateRoomBooleanProps {
    field: "isEnabled" | "isLabelled";
    newState: boolean;
  }
  
  const updateRoomBoolean = async ({ field: string, newState: boolean }) => {
```

0xdevalias: This function would also be best if it was moved into somewhere like `api/rooms.ts`

src/pages/Admin/AdminVenueRoomDetails.tsx
```diff
@@ -122,13 +122,28 @@ export const AdminVenueRoomDetails = ({
...
+                <div className="toggle-room">
+                  <label id={"toggle2-" + index} className="switch">
+                    <input
+                      type="checkbox"
+                      id={"toggle2-" + index}
+                      name={"toggle2-" + index}
+                      checked={room.isLabeled}
+                      onClick={() => {
+                        updateRoom(!room.isLabeled, "isLabeled");
+                      }}
```
0xdevalias, line 141: Inline function handlers like this will cause re-renders. They should be wrapped with `useCallback` outside of the JSX, then that passed here.

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -35,6 +35,7 @@ export const MapRoom: React.FC<MapRoomProps> = ({
   const isUnclickable = room.type === RoomType.unclickable;
   const isMapFrame = room.type === RoomType.mapFrame;
   const isCovertRoom = room.type && COVERT_ROOM_TYPES.includes(room.type);
+  const isLabeled = room.isLabeled !== undefined ? room.isLabeled : true;
```
0xdevalias, line 38: This will use the value if it exists, or default it to `true` if it's `undefined`:
```suggestion
  const isLabeled = room.isLabeled ?? true;
```

**Further Reading:**

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -80,9 +81,10 @@ export const MapRoom: React.FC<MapRoomProps> = ({
 
   const [play] = useCustomSound(room.enterSound, { interrupt: true });
   const selectRoomWithSound = useCallback(() => {
+    if (!isLabeled) return;
     play();
     selectRoom();
-  }, [play, selectRoom]);
+  }, [play, selectRoom, isLabeled]);
```
0xdevalias, line 87: This will prevent the room / portal from being interacted with, which definitely isn't something we want to do here. We only want to ensure the label isn't shown on mouseover.

Please revert these changes.

src/types/rooms.ts
```diff
@@ -19,6 +19,7 @@ export interface Room {
   width_percent: number;
   height_percent: number;
   isEnabled: boolean;
+  isLabeled?: boolean;
```
0xdevalias, line 22: I'm not a fan of this naming.. Our default is that labels will be shown as normal, so this exception should be named as changing that.

```suggestion
  isLabelHidden?: boolean;
```

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -104,7 +106,7 @@ export const MapRoom: React.FC<MapRoomProps> = ({
         <img className="maproom__image" src={room.image_url} alt={room.title} />
       )}
 
-      {!isCovertRoom && (
+      {!isCovertRoom && isLabeled && (
```
0xdevalias, line 109: It would be good to combine these into a single `const` above, then use that here to keep the logic cleaner within the JSX.

0xdevalias (changes requested):
Comments/changes/etc.

0xdevalias (approved):
Made some changes, now looks good to go.

LGTM!

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1632/commits/ca8fae8cf8aa67b941e899e162433062a81f9564

src/types/rooms.ts
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1632/commits/9e679604114d01a3121e1d3d42ac06b60632cc4a

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1632/commits/274b6319f73d8b047099bb23ee6d33d8bda5ada2

src/components/templates/PartyMap/components/Map/MapRoom.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1632/commits/9e679604114d01a3121e1d3d42ac06b60632cc4a

src/pages/Admin/AdminVenueRoomDetails.tsx
0xdevalias: Decided to call this OOS as this issue is all over the place in admin v1, and it's not worth cleaning up since we plan to remove it as soon as admin v3 reaches feature parity.

src/pages/Admin/AdminVenueRoomDetails.tsx
0xdevalias: Decided to call most of this OOS as this as it's not worth cleaning up since we plan to remove it as soon as admin v3 reaches feature parity.

fixed the typing in https://github.com/sparkletown/sparkle/pull/1632/commits/88062e4868edc765e7f9b6db71d8d76159c296a8

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
0xdevalias: Looked at this, the current implementation of `Toggler` doesn't support all the props we need it to be able to do to use in this kind of situation, so calling this OOS

src/pages/Admin/AdminVenueRoomDetails.tsx
0xdevalias: I ended up removing this toggler entirely, as we don't need this to be exposed on the main view. It's more of an advanced setting, so can be 'hidden away' only in the main 'edit room' dialog.

ermiaSparkle:
@0xdevalias Thank you for all comments, I will try not to repeat any of them.
Should I add a script for toggle this as we don't have the toggle in admin (v1)?

0xdevalias:
> Thank you for all comments, I will try not to repeat any of them.

No worries :) 👌🏻 

> Should I add a script for toggle this as we don't have the toggle in admin (v1)?

What do you mean 'add a script for toggle this'? This PR already adds the toggle in admin (v1). That's what `RoomsForm` is.

Do you mean in admin (v2/3)?

IMO this PR is good to go as is.

sofisparkle:
@whoaitsaimz this thread may give some context for database inputs!

PR #1637, opened 2021-06-23
TEMP: replace env/ohbm with gh-soichih/usersperformance

yarikoptic:
If merged, then it would result in easy way to revert the diff from the prior state

an alternative to hard resetting etc env/ohbm


yarikoptic:
although I think "hard reset" approach might be better since wouldn't mess with the history and allow to actually continue growing from soichi's branch by re-merging interesting PRs etc. With this approach it would be trickier!

yarikoptic:
although I think "hard reset" approach might be better since wouldn't mess with the history and allow to actually continue growing from soichi's branch by re-merging interesting PRs etc. With this approach it would be trickier!

yarikoptic:
no, don't like this option, will close

PR #1638, opened 2021-06-23
Revert "refactor lastSeenAt / lastSeenIn out of main User document + associated refactors"

mike-lvov:
Reverts sparkletown/sparkle#1623

PR #1623, opened 2021-06-22
refactor lastSeenAt / lastSeenIn out of main User document + associated refactors

0xdevalias:
Opening this draft PR super early to track progress.

Aim is to refactor `lastSeenIn` + `lastSeenAt` off the main `User` object/types to prevent it breaking memoisation across the platform when updated.

This is suspect to be one of the core contributors to performance issues, particularly as the number of online users increases.

PR #1639, opened 2021-06-23
Revert "Revert "refactor lastSeenAt / lastSeenIn out of main User document + associated refactors""

mike-lvov:
Reverts sparkletown/sparkle#1638 to restore #1623

Review comments should also be looked at in #1623 + potentially addressed still.


0xdevalias:
Adding further context/exploration here for future reference:

> > I thought may be it would be better to stop them query every minute for a list of users as was needed to be done for OHBM
> > in #1630 ? To say the truth I am not sure if it would make positive or negative impact on the number of requests and thus cost of still running OHBM sparkle instance but just couldn't stay silent seeing $s going wasted ;)
> 
> @yarikoptic Thanks for this PR! To provide some context, I am almost certain that this will have a positive impact on the requests/costs.
> 
> Without going too deeply into the specifics, our firebase costs have been relatively low/stable for across most environments we run. OHBM spiked in cost _significantly_, and from the interim exploration I have done to find out 'why', it seems to point back to @soichih's #1630, though there will almost certainly be some element of it that is just 'increased load' due to OHBM being a larger event in general (and having some 'unique specifics' that differ from previous events we have run, eg. having 1000+ related venues, etc)
> 
> In particular, I believe the way that #1630 is doing a full query every minute or so per user is loading vastly more data than the previous 'live update' method would have. The live update method will load the data the first time, but then only receive the tiny 'snapshot updates' each time. So if a particular user's record hasn't changed then no data will need to be sent down again, and it won't count as a 'read' against the firestore DB.
> 
> While I haven't looked too deeply at the code in #1630, in skimming it just now I see the following:
> 
> - `src/index.tsx` uses `setInterval` to `dispatch` `CacheActionTypes.RELOAD_USER_CACHE` every `60` seconds. Given this is so high up in the code, this will run while a user is connected to the platform in any way, even if they aren't logged in.
> - `src/store/reducers/Cache.ts` handles 'reducing' the various redux actions that were added here. It violates one of the core principles that reducers should be pure functions, and calls async side effect code to load the entire `users` collection from firebase; then violates Redux's "never directly mutate the state, only return an updated state" to update it with the fetched records. This should have used `redux-thunk` or similar patterns for handling async things in redux. This will end up loading a LOT of unneeded data each time it's called, and is almost certainly the reason this drove up costs by several orders of magnitude. A hacky optimisation would have been to only query for records that had changed since the last time the query was run or similar, which should have been a 1-2 line change, and likely would have eliminated most of the excess costs incurred here from ever needing to have been incurred.
> - `useUserInvalidateCache` in `useUser` queries the `users` collection in firestore to retrieve a certain user's document, and then updates the cache with that record
>   - This appears to be used in `src/components/templates/Audience/Audience.tsx` / `src/components/templates/PartyMap/components/Map/Map.tsx` to handle the 'update user's seat position more quickly' issue
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1690#issuecomment-871032253_

PR #1647, opened 2021-06-24
add boilerplate for single-file custom css overrides

whoaitsaimz:
As a way to implement custom CSS, I propose this solution that closes https://github.com/sparkletown/internal-sparkle-issues/issues/597 

Details:

- Addition of a new file, that we can then make edits to on a per-environment basis without affecting other files, thus being able to easily isolate any strangeness and also simplifying code reviews
- This file should be given priority, so I've made it last in hierarchy
- We can follow standard patterns of .scss and use variables
- We maintain performance where everything is served from one, compiled CSS
- Customizations can be made and PRs submitted through our standard process and will be version controlled etc (whereas with database strings, we would not have this benefit)


whoaitsaimz:
I have only requested to merge into env/github for now, but imagine this will be a useful thing for main code base, once we test process

cc: @Sparkle-T

src/scss/ClientCustom/Custom.scss
```diff
@@ -0,0 +1,3 @@
+@import "scss/constants.scss";
```
0xdevalias, line 1: If this is only a single file, why add the ClientCustom folder? Filename should be lowercase. Drop client from this cots irrelevant and doesn’t align to our open source project.

Why not just custom.scss or customOverrides.scss or similar?

src/scss/global.scss
```diff
@@ -6,6 +6,7 @@
 @import "node_modules/bootstrap/scss/bootstrap";
 @import "node_modules/bootswatch/dist/darkly/bootswatch";
 @import "./constants.scss";
+@import "./ClientCustom/Custom.scss";
```
0xdevalias, line 9: I have mixed feels about this. We’re generally trying to eradicate things from the global scss.. It’s probably not the worst thing.. it’s better than hacking in from the DB.

0xdevalias:
> * This file should be given priority, so I've made it last in hierarchy

I’m not sure that will do anything at all tbh. Priority comes from CSS specificity.

src/scss/ClientCustom/Custom.scss
whoaitsaimz: Perfect-o will do

src/scss/global.scss
whoaitsaimz: Precisely, a compromise

whoaitsaimz:
> > * This file should be given priority, so I've made it last in hierarchy
> 
> I’m not sure that will do anything at all tbh. Priority comes from CSS specificity. There’s a cheat sheet called specifishity that’s a good reference.

Yes order in the file (how far down the page) is part of priority. And if we had multiple CSS files loaded individually (rather than imported as we do now) then the ones loaded last would have highest priority. So I set this override file to be imported later as I figure that will also carry over hierarchy.

src/scss/global.scss
0xdevalias, line 9: This needs to be updated based on the renamed file

src/scss/customOverrides.scss
```diff
@@ -0,0 +1,3 @@
+@import "scss/constants.scss";
+
+// Custom CSS file to deal with specific client requests
```
0xdevalias, line 3:
```suggestion
// Custom CSS file to deal with environment specific hacks/overrides
```

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1647/commits/2b61b8cc3f4f16c3310a72d0e2059b7c070d091f

src/scss/global.scss
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1647/commits/2b61b8cc3f4f16c3310a72d0e2059b7c070d091f

0xdevalias (approved):
Let's see how this ends up getting used and evaluate further from there. I still have my concerns about this pattern, and am worried it's going to lead to make more convoluted changes that are harder to notice/detect since they aren't made in the actual `.scss` files they apply to, which will make it harder for engineers to build things without breaking in environments that make use of it.

That said, LGTM!

PR #1651, opened 2021-06-24
Always show Events tab for venues in Admin V1

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/775


src/pages/Admin/Admin.tsx
```diff
@@ -154,7 +138,7 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId, roomIndex }) => {
 
     return [
       { url: matchUrl, label: "Venue Info" },
-      canHaveEvents(venue) && { url: `${matchUrl}/events`, label: "Events" },
```
sunny-viktoryia, line 157: the fix is here

src/pages/Admin/Admin.tsx
```diff
@@ -278,57 +262,6 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
               venue={venue}
               containerStyle={{ marginTop: 20 }}
             />
-            {IS_BURN && PLACEABLE_VENUE_TEMPLATES.includes(venue.template) && (
```
sunny-viktoryia, line 281: I assumed that we don't need `IS_BURN` flag anymore (correct me if I'm wrong). 
If so, we can remove this part and `PLACEABLE_VENUE_TEMPLATES` becomes not used anywhere else, so we can remove it as well.

src/pages/Admin/Admin.tsx
```diff
@@ -153,7 +137,7 @@ const VenueDetails: React.FC<VenueDetailsProps> = ({ venueId, roomIndex }) => {
 
     return [
       { url: matchUrl, label: "Venue Info" },
-      canHaveEvents(currentVenue) && {
```
sunny-viktoryia, line 156: the fix is here

src/pages/Admin/Admin.tsx
dmitrymarchuk: `IS_BURN` is still used in a bunch of places, maybe it's better to remove it entirely?

dmitrymarchuk (approved):
LGTM but needs conflict resolution.

src/pages/Admin/Admin.tsx
goran-peoski-work: +1 if we can weed out the `IS_BURN` uses one by one, especially in this case since there are plenty of venue templates lists in `settings.ts` already

PR #1653, opened 2021-06-24
[env/github] fix TalkShowStudio bugs

dmitrii-zolotukhin:
the list of bugs: https://github.com/sparkletown/internal-sparkle-issues/issues/748#issuecomment-866993048

- fixed repeated sounds in Custom Reactions

(partially?) fixes https://github.com/sparkletown/internal-sparkle-issues/issues/748

## TODO

> - [x] When I share screen, no one can hear or see me
> - [x] When I change seats, I leave the stage
> - [x] When I press STOP screensharing from Chrome, I also have to press stop sharing in Sparkle
> - [x] When someone else is screensharing, I cannot see myself, although others can see me
> - [x] When screenshare stops, my camera appears black but people can still see me 
> - [x] When I am the only admin in the space, I cannot leave the stage
> - [x] When I am the only admin in the space, turning off my camera, leaving the stage restarts my camera
> 
> _Originally posted by @sofisparkle in https://github.com/sparkletown/internal-sparkle-issues/issues/748#issuecomment-866993048_


src/components/molecules/UserReactions/UserReactions.tsx
```diff
@@ -101,7 +101,7 @@ export const DisplayEmojiReaction: React.FC<EmojiReactionProps> = ({
 
       {/* @debt replace this with useSound or calling new Audio in useEffect or similar */}
       {!isMuted && (
-        <audio autoPlay loop>
+        <audio autoPlay>
           <source src={audioPath} />
         </audio>
       )}
```
0xdevalias, line 107: A better fix here would be to use a `useEffect` + the JavaScript audio API (not this `<audio>` tag), or to use `useSound` as per the `@debt` comment. The audio API is the closest to the current code here, and will have less nuances/scope for issues, so that is probably the best quick fix to use here, and then make sure to keep the `@debt` comment saying to replace it with `useSound`.

- https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio

0xdevalias:
@dmitrii-zolotukhin Instead of appending WIP to the PR title, please use draft branches. That's what they exist for:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/

yurii-lubynets:
@dmitrii-zolotukhin, please make sure you are aligned with @0xdevalias PR and changes here https://github.com/sparkletown/sparkle/pull/1673

0xdevalias:
> @dmitrii-zolotukhin, please make sure you are aligned with @0xdevalias PR and changes here #1673

If you wanted to make merging easier, you could add a stubbed version of the api/video getAgoraToken file that just returns the static token you have configured in this branch currently. That way you can align your code here to the new way without needing to be blocked waiting for my merge.

Please don’t merge my branch into this one. I want to ensure it’s deployed properly through staging/etc.

0xdevalias:
> @yurii-lubynets @dmitrii-zolotukhin How does this PR relate to https://github.com/sparkletown/sparkle/pull/1653 ? Is it in addition to; does it supercede/replace it; etc?
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1674#issuecomment-869290345_

.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -38,6 +38,9 @@ jobs:
       REACT_APP_STRIPE_PUBLISHABLE_KEY: '${{ secrets.STRIPE_PUBLISHABLE_KEY }}'
       REACT_APP_BUGSNAG_API_KEY: '${{ secrets.BUGSNAG_API_KEY }}'
       REACT_APP_LOGROCKET_APP_ID: '${{ secrets.LOGROCKET_APP_ID }}'
+      REACT_APP_AGORA_APP_ID: '${{ secrets.AGORA_APP_ID }}'
+      REACT_APP_AGORA_CHANNEL: '${{ secrets.AGORA_CHANNEL }}'
+      REACT_APP_AGORA_TOKEN: '${{ secrets.AGORA_TOKEN }}'
```
0xdevalias, line 43: We won't need any of these here when using the real token fetching from backend method I wrote.

```suggestion
```

scripts/init-env.sh
```diff
@@ -6,7 +6,7 @@ PREFIX=$1
 rm $ENV_FILE 2> /dev/null
 touch $ENV_FILE
 
-for VARIABLE in API_KEY APP_ID AUTH_DOMAIN MEASUREMENT_ID BUCKET_URL PROJECT_ID STRIPE_PUBLISHABLE_KEY BUGSNAG_API_KEY LOGROCKET_APP_ID MIXPANEL_PROJECT_TOKEN WS_RELAY_URL CODE_CHECK_URL IS_BURN
+for VARIABLE in API_KEY APP_ID AUTH_DOMAIN MEASUREMENT_ID BUCKET_URL PROJECT_ID STRIPE_PUBLISHABLE_KEY BUGSNAG_API_KEY LOGROCKET_APP_ID MIXPANEL_PROJECT_TOKEN WS_RELAY_URL CODE_CHECK_URL IS_BURN AGORA_APP_ID AGORA_CHANNEL AGORA_TOKEN
```
0xdevalias, line 9: We won't need any of these here when using the real token fetching from backend method I wrote, so can just revert this back to how it was.

```suggestion
for VARIABLE in API_KEY APP_ID AUTH_DOMAIN MEASUREMENT_ID BUCKET_URL PROJECT_ID STRIPE_PUBLISHABLE_KEY BUGSNAG_API_KEY LOGROCKET_APP_ID MIXPANEL_PROJECT_TOKEN WS_RELAY_URL CODE_CHECK_URL IS_BURN
```

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
-          )
+
+    return remoteUsers.map((agoraRemoteUser) => {
+      const user = setRemoteUserAvatar(agoraRemoteUser.uid);
+      if (!user) return null;
+
+      const isUniqueUser = !uniqueUsersIds.has(user.id);
+      if (isUniqueUser) {
+        uniqueUsersIds.add(user.id);
+      }
```
0xdevalias, line 116: This check shouldn't be needed, as `Set` by default keeps fields unique:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
  - > A value in the Set may only occur once; it is unique in the Set's collection.

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
     const setRemoteUserAvatar = (remoteUserId: number | string) => {
       if (!venue.id) return;
-      const remoteUser = stage.peopleOnStage.find(
+
+      return stage.peopleOnStage.find(
         ({ data }) =>
           data?.[`${venue.id}`]?.cameraClientUid === `${remoteUserId}`
       );
-      return remoteUser;
     };
```
0xdevalias, line 107: What is the purpose of this function? It doesn't seem to 'set' anything? It finds/returns something?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
-        ({ uid }) =>
-          `${uid}` !==
-          userOnStageSharingScreen?.data?.[`${venue.id}`]?.screenClientUid
-      )
-      .map(
-        (user) =>
-          user.uid !== screenClient.uid &&
-          user.uid !== cameraClient.uid && (
-            <div key={user.uid}>
-              {user.hasVideo && (
```
0xdevalias, line 132: Why did we remove the `.filter`'s + check to see if things are defined here? Are they needed?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
+      if (!user) return null;
+
+      const isUniqueUser = !uniqueUsersIds.has(user.id);
+      if (isUniqueUser) {
+        uniqueUsersIds.add(user.id);
+      }
+
+      return (
+        agoraRemoteUser.uid !== screenClient.uid &&
+        agoraRemoteUser.uid !== cameraClient.uid && (
```
0xdevalias, line 120: Instead of inlining this check here, we should use a guard pattern. If those things aren't defined, then just return. Do this separately to the main JSX block here.

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
+          <div key={agoraRemoteUser.uid}>
+            {isUniqueUser && (
```
0xdevalias, line 122: Why do we need/use `isUniqueUser` here? Seems.. weird?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
+                showButtons
+                user={setRemoteUserAvatar(agoraRemoteUser.uid)}
+                videoTrack={agoraRemoteUser.videoTrack}
+                audioTrack={agoraRemoteUser.audioTrack}
+                isCamOn={agoraRemoteUser.hasVideo}
+                isMicOn={agoraRemoteUser.hasAudio}
+                toggleCam={isUserOwner ? stage.toggleUserCamera : undefined}
+                toggleMic={isUserOwner ? stage.toggleUserMicrophone : undefined}
+                containerClass="TalkShowStudio__mode--play"
+              />
```
0xdevalias, line 133: There are a LOT of props needing to be passed through to this component.. I expect there should probably be a better pattern to use here.

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -194,46 +194,25 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
-            containerClass="TalkShowStudio__mode--local-play"
-          />
-        )}
+        <Player
+          videoTrack={screenTrack}
+          containerClass="TalkShowStudio__mode--share"
+        />
       </div>
     );
-  };
```
0xdevalias, line 225: Why is the `cameraTrack` stuff removed from here?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -194,46 +194,25 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
     !stage.isUserRequesting;
 
-  const renderScreenSharing = (
-    screenTrack: VideoPlayerProps["videoTrack"],
-    cameraTrack: VideoPlayerProps["videoTrack"]
-  ) => {
-    if (!screenTrack || !cameraTrack) return;
-    return (
+  const renderScreenSharing = (screenTrack: VideoPlayerProps["videoTrack"]) =>
+    screenTrack && (
```
0xdevalias, line 198: Instead of nesting everything in a conditional render like this, use our standard guard pattern:

```typescript
if (!screenTrack) return

return theJSXHere
```

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -194,46 +194,25 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
       <div className="TalkShowStudio">
         <div className="TalkShowStudio__scene">
           {renderScreenSharing(
-            localScreenTrack || remoteScreenTrack?.videoTrack,
-            localCameraTrack || remoteCameraTrack?.videoTrack
+            localScreenTrack || remoteScreenTrack?.videoTrack
           )}
           <div className="TalkShowStudio__players">
-            {localCameraTrack && !localScreenTrack && (
+            {localCameraTrack && (
```
0xdevalias, line 215: Why is `localScreenTrack` removed from here?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -94,56 +94,56 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
   );
-  const remoteCameraTrack = useMemo(
-    () =>
-      remoteUsers.find(
-        ({ uid }) =>
-          `${uid}` ===
-          userOnStageSharingScreen?.data?.[`${venue.id}`]?.cameraClientUid
-      ),
-    [userOnStageSharingScreen?.data, remoteUsers, venue.id]
-  );
```
0xdevalias, line 105: Why was `remoteCameraTrack` removed from here?

src/components/templates/TalkShowStudio/components/Player/Player.scss
```diff
@@ -51,8 +59,15 @@
       user-select: none;
     }
 
-    &--on {
+    &--off {
+      visibility: visible;
       background: rgb(124, 70, 251);
     }
   }
```
0xdevalias, line 66: Why did we change this from `&--on` to `&--off`?

src/components/templates/TalkShowStudio/components/Player/Player.scss
```diff
@@ -51,8 +59,15 @@
...
+
+  &:hover {
+    .Player__buttons {
+      visibility: visible;
+    }
+  }
```
0xdevalias, line 72: `&:` styles should go before the `&__` styles as they relate to the main class. So this should be before `&__background`

src/components/templates/TalkShowStudio/components/Player/Player.tsx
```diff
@@ -66,18 +66,19 @@ const Player: React.FC<VideoPlayerProps> = ({
...
 
   return (
     <div ref={container} className={classNames("Player", containerClass)}>
+      <div className="Player__background" />
       {showButtons && (
         <div className="Player__overlay">
           <div className="Player__buttons">
             <div
-              className={IconContainerClasses(!isMicOn)}
+              className={iconContainerClasses(!isMicOn)}
```
0xdevalias, line 81: We shouldn't have to add the `!` here. The helper should handle any negations it may need inside itself. The calling code shouldn't have to care.

src/components/templates/TalkShowStudio/components/Player/Player.tsx
```diff
@@ -86,7 +87,7 @@ const Player: React.FC<VideoPlayerProps> = ({
             </div>
 
             <div
-              className={IconContainerClasses(!isCamOn)}
+              className={iconContainerClasses(!isCamOn)}
```
0xdevalias, line 90: We shouldn't have to add the `!` here. The helper should handle any negations it may need inside itself. The calling code shouldn't have to care.

src/components/templates/TalkShowStudio/components/Player/Player.tsx
```diff
@@ -95,7 +96,7 @@ const Player: React.FC<VideoPlayerProps> = ({
             </div>
 
             {isSharing && (
-              <div className={IconContainerClasses(true)}>
+              <div className={iconContainerClasses(true)}>
```
0xdevalias, line 99: We shouldn't be passing through a raw `true` like this. Either add the correct class here directly (preferred IMO), or pass through the variable.

src/secrets.js
```diff
@@ -14,11 +14,9 @@ export const MIXPANEL_PROJECT_TOKEN =
...
 export const IS_BURN = process.env.REACT_APP_IS_BURN;
 
-// @debt doing this for GitHub demo
-export const AGORA_APP_ID = "6f46c1a52afc445c94901fa64877cd05";
-export const AGORA_CHANNEL = "TestSparkle";
-export const AGORA_TOKEN =
-  "0066f46c1a52afc445c94901fa64877cd05IAAzOVWNGWxPaIM6/imvMqJoOMLMl6sn4k9OSmlaL7PZ0bjtc7kAAAAAEAAm+nFWKabUYAEAAQApptRg";
+export const AGORA_APP_ID = process.env.REACT_APP_AGORA_APP_ID;
+export const AGORA_CHANNEL = process.env.REACT_APP_AGORA_CHANNEL;
+export const AGORA_TOKEN = process.env.REACT_APP_AGORA_TOKEN;
```
0xdevalias, line 19: We won't need any of these here when using the real token fetching from backend method I wrote.

src/settings.ts
```diff
@@ -561,6 +564,9 @@ export const FIREBASE_CONFIG = {
   measurementId: MEASUREMENT_ID,
   projectId: PROJECT_ID,
   storageBucket: BUCKET_URL,
+  agoraAppId: AGORA_APP_ID,
+  agoraChannel: AGORA_CHANNEL,
+  agoraToken: AGORA_TOKEN,
```
0xdevalias, line 569: Not sure why these were added here? But they have nothing to do with the firebase config.

```suggestion
```

src/settings.ts
```diff
@@ -8,6 +8,9 @@ import {
   BUCKET_URL,
   PROJECT_ID,
   IS_BURN,
+  AGORA_APP_ID,
+  AGORA_CHANNEL,
+  AGORA_TOKEN,
```
0xdevalias, line 13:
```suggestion
```

0xdevalias (changes requested):
Questions, comments, changes.

Also, would be good to understand if/how this relates to https://github.com/sparkletown/sparkle/pull/1674#issuecomment-869290345 and whether this PR is even needed anymore.

0xdevalias:
> @yurii-lubynets @dmitrii-zolotukhin As per [this slack message](https://doesitsparkle.slack.com/archives/C0229C1EC86/p1624865732437600?thread_ts=1624648679.279900&cid=C0229C1EC86) my understanding is that the plan is to get https://github.com/sparkletown/sparkle/pull/1653 cleaned up, and that will be the canonical bugfix PR.  
> 
> Based on that, I believe we can now consider https://github.com/sparkletown/sparkle/pull/1674 as 'abandoned', and close it off.
> 
> Is that correct?
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1674#issuecomment-869474329_

src/components/molecules/UserReactions/UserReactions.tsx
0xdevalias: @dmitrii-zolotukhin FYI: This is basically the `useEffect` + Audio API patch:

```diff
diff --git a/src/components/molecules/UserReactions/UserReactions.tsx b/src/components/molecules/UserReactions/UserReactions.tsx
index fd89ca5c..801cb007 100644
--- a/src/components/molecules/UserReactions/UserReactions.tsx
+++ b/src/components/molecules/UserReactions/UserReactions.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { useEffect, useMemo } from "react";
 import { useCss } from "react-use";
 import classNames from "classnames";
 import { getSeconds } from "date-fns";
@@ -95,16 +95,24 @@ export const DisplayEmojiReaction: React.FC<EmojiReactionProps> = ({
 }) => {
   const { ariaLabel, text: emojiText, audioPath } = emojiReaction;
 
+  useEffect(() => {
+    if (isMuted) return;
+
+    const audio = new Audio(audioPath);
+
+    audio.play();
+
+    return () => {
+      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio#memory_usage_and_management
+      //   the audio will keep playing and the object will remain in memory until playback ends or is paused (such as by calling pause()).
+      //   At that time, the object becomes subject to garbage collection.
+      audio.pause();
+    };
+  }, [audioPath, isMuted]);
+
   return (
     <div className="UserReactions__reaction" role="img" aria-label={ariaLabel}>
       {emojiText}
-
-      {/* @debt replace this with useSound or calling new Audio in useEffect or similar */}
-      {!isMuted && (
-        <audio autoPlay loop>
-          <source src={audioPath} />
-        </audio>
-      )}
     </div>
   );
 };
```

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
dmitrii-zolotukhin: @0xdevalias I was thinking over about it but I don't have any ideas of how to implement it. In different places that component gets different essences to props and using any common hooks and smth like this won't be possible

.github/workflows/firebase-hosting-pull-request.yml
dmitrii-zolotukhin: @0xdevalias for hooks working (src/hooks/video/agora/index.ts) I need an App Id and Channel Name in addition to token. If I'll delete now these ENV from file, so agora won't work too.

src/components/molecules/UserReactions/UserReactions.tsx
```diff
@@ -95,16 +95,24 @@ export const DisplayEmojiReaction: React.FC<EmojiReactionProps> = ({
...
+
+    const audio = new Audio(audioPath);
+
+    audio.play();
+
+    return () => {
+      // https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio#memory_usage_and_management
+      //   the audio will keep playing and the object will remain in memory until playback ends or is paused (such as by calling pause()).
+      //   At that time, the object becomes subject to garbage collection.
+      audio.pause();
```
yurii-lubynets, line 109: nice catch!

src/hooks/video/agora/useAgoraScreenShare.ts
```diff
@@ -24,10 +24,43 @@ export const useAgoraScreenShare: ReactHook<
...
+
+    await client.unpublish();
+
+    setLocalScreenTrack(undefined);
+    setLocalAudioTrack(undefined);
+  }, [venueId, userId, client, localAudioTrack, localScreenTrack]);
+
+  const leaveChannel = useCallback(async () => {
+    stopShare();
+    await client?.leave();
```
yurii-lubynets, line 52:
```suggestion
  const leaveChannel = useCallback(async () => {
    if(!client) return;
    
    stopShare();
    await client.leave();
```

src/hooks/video/agora/useAgoraScreenShare.ts
```diff
@@ -24,10 +24,43 @@ export const useAgoraScreenShare: ReactHook<
...
+  }, [client, stopShare]);
+
   const shareScreen = useCallback(async () => {
     if (!client) return;
 
     const screenTrack = await AgoraRTC.createScreenVideoTrack({}, "auto");
+    // We add a check since there is no "on" method in screenTrack types,
+    // but in fact there is one and this is the main way to track browser "Stop sharing" button click.
+    if (!screenTrack || !("on" in screenTrack)) return;
```
yurii-lubynets, line 61: nice catch!

yurii-lubynets (approved):
LGTM!
Some comments left, please check

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
0xdevalias: Makes sense.

src/hooks/video/agora/useAgoraScreenShare.ts
0xdevalias: @yurii-lubynets @dmitrii-zolotukhin Do we actually want to not call `stopShare()` if there is no `client`?

Or do we only want to avoid calling `client.leave()` when there is no `client`?

0xdevalias:
@dmitrii-zolotukhin Please ensure you follow our standards with regards to responding to review comments threads with relevant details such as the commit hash where the change was made, etc; as per https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625023021018200?thread_ts=1620395110.209200&cid=C01ER4M61TL

0xdevalias:
@dmitrii-zolotukhin @yurii-lubynets + anyone who comes to review this: Given we have de-prioritised work on this feature for now, we should take the time to properly review/assess the changes here and ensure they won't needlessly add risk to the `env/github` environment when merged. The 'urgency' of getting this merged is somewhat lessened now, but we absolutely don't want to de-stabilise the environment for a feature that isn't getting used.

whoaitsaimz:
There does seem to be a lot of... stuff... left in the console now. And this is just on the register page!

image

dmitrii-zolotukhin:
> > we should take the time to properly review/assess the changes here and ensure they won't needlessly add risk to the env/github environment when merged
> 
> There does seem to be a lot of... Agora stuff... left in the console now in env//github. And this is just on the register page!
> 
> image

@whoaitsaimz did you check that all the permissions for microphone and camera are given?

dmitrii-zolotukhin:
@whoaitsaimz also, can you, please, tell what actions was done before that Agora's stuff?

.github/workflows/firebase-hosting-pull-request.yml
0xdevalias: Look at the implementation on my draft PR where I wire everything up. It also contains some tweaks to the backend functionality to support getting the app id.

0xdevalias:
> > we should take the time to properly review/assess the changes here and ensure they won't needlessly add risk to the env/github environment when merged
> 
> There does seem to be a lot of... Agora stuff... left in the console now in env//github. And this is just on the register page!
> 
> image

This is the issue I called out in #tech yesterday. The agora clients are being made in global scope, thus initialised on app load regardless of whether this template is even being accessed. That shouldn’t be happening. They should be done within the appropriate hook/react component only when needed.

src/hooks/video/agora/useAgoraScreenShare.ts
```diff
@@ -25,7 +25,8 @@ export const useAgoraScreenShare: ReactHook<
   const [localAudioTrack, setLocalAudioTrack] = useState<ILocalAudioTrack>();
 
   const stopShare = useCallback(async () => {
-    if (!venueId || !userId || !client) return;
+    if (!venueId || !userId || !client || !localScreenTrack || !localAudioTrack)
+      return;
 
     updateTalkShowStudioExperience({
```
0xdevalias: Is it correct to not call `updateTalkShowStudioExperience` when `!localScreenTrack || !localAudioTrack`? The previous way would have ensured the backend data stayed in sync with the expected reality I would have thought? Whereas now I expect there is more chance of them being weirdly out of sync.

src/hooks/video/agora/useAgoraScreenShare.ts
```diff
@@ -48,8 +49,10 @@ export const useAgoraScreenShare: ReactHook<
   }, [venueId, userId, client, localAudioTrack, localScreenTrack]);
 
   const leaveChannel = useCallback(async () => {
+    if (!client) return;
+
     stopShare();
-    await client?.leave();
+    await client.leave();
   }, [client, stopShare]);
```
0xdevalias: Is it correct to not call `stopShare` when `!client`? The previous way would have ensured the backend data stayed in sync with the expected reality I would have thought? Whereas now I expect there is more chance of them being weirdly out of sync.

src/hooks/video/agora/useAgoraCamera.ts
```diff
@@ -31,9 +31,7 @@ export const useAgoraCamera: ReactHook<
   } = useShowHide();
 
   const toggleCamera = () => {
-    if (!localCameraTrack) return;
-
-    localCameraTrack.setEnabled(!isCameraOn);
+    localCameraTrack?.setEnabled(!isCameraOn);
     setIsCameraOn(!isCameraOn);
   };
```
0xdevalias: `useShowHide` already has a premade `toggle` function that is more efficient than this, uses proper `useState` update patterns, and is already `useCallback`'d

src/hooks/video/agora/useAgoraCamera.ts
```diff
@@ -68,13 +66,13 @@ export const useAgoraCamera: ReactHook<
...
+    localCameraTrack?.stop();
+    localCameraTrack?.close();
 
-    localMicrophoneTrack.stop();
-    localMicrophoneTrack.close();
+    localMicrophoneTrack?.stop();
+    localMicrophoneTrack?.close();
 
     setLocalCameraTrack(undefined);
     setLocalMicrophoneTrack(undefined);
```
0xdevalias: Is this change correct? By guarding based on `!client` it means we're never going to update `localCameraTrack`, `localMicrophoneTrack`, `setLocalCameraTrack`, `setLocalMicrophoneTrack`; so I expect they are now going to be more likely to 'fall out of sync' and cause weird bugs.

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -17,7 +17,7 @@ import {
 } from "hooks/video/agora";
 
 import AppButton from "components/atoms/Button";
-import Player, { VideoPlayerProps } from "./components/Player/Player";
+import { Player, VideoPlayerProps } from "./components/Player/Player";
 import { ControlBar } from "./components/ControlBar";
 import Audience from "./components/Audience/Audience";
 import SettingsSidebar from "./components/SettingsSidebar/SettingsSidebar";
```
0xdevalias, line 23: I can't actually comment on it inline since it wasn't modified in this PR, but a couple of lines down from here: `remotesClient`, `screenClient`, `cameraClient` are incorrectly created in global scope, and thus loaded and held in memory from the first load of the app, regardless of whether the `TalkShowStudio` / agora is even accessed/used at all.

---

> RE: https://github.com/sparkletown/sparkle/pull/1653#issuecomment-871297368 + https://github.com/sparkletown/sparkle/pull/1653#issuecomment-871305231 + https://github.com/sparkletown/sparkle/pull/1653#issuecomment-871330197
> 
> > > we should take the time to properly review/assess the changes here and ensure they won't needlessly add risk to the env/github environment when merged
> > 
> > There does seem to be a lot of... Agora stuff... left in the console now in env//github. And this is just on the register page!
> 
> @dmitrii-zolotukhin @whoaitsaimz This is the issue I called out in #tech on slack yesterday ([ref](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625029929020100)). The agora clients are being made in global scope, thus initialised on app load regardless of whether this template is even being accessed. That shouldn’t be happening. They should be done within the appropriate hook/react component only when needed.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1653#issuecomment-871783853_

0xdevalias: > Because of how the `TalkShowStudio` is creating 3 agora clients in global scope within it’s component, it is creating them + holding them in memory from initial app load, regardless of whether the user is actually using that venue/agora at all. This is something that should be removed ASAP, and could be having impacts on the first-load performance of `env/github`
>
> _Originally posted by @0xdevalias in [`#client-github` on Slack](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625029929020100)._

0xdevalias: @dmitrii-zolotukhin Please follow our standards and add the relevant commit hash to the review comment threads where it was resolved, then mark it resolved.

https://github.com/sparkletown/sparkle/pull/1653/commits/1a95556a6dfecdc953e99c7e694c1f6b703180f9

0xdevalias:
@dmitrii-zolotukhin Please follow our standards and add the relevant commit hash to the review comment threads where it was resolved, then mark it resolved.

I asked you to do this in [this slack message](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625023021018200?thread_ts=1620395110.209200&cid=C01ER4M61TL) and have asked you again in [this slack message](https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625194410053100?thread_ts=1620395110.209200&cid=C01ER4M61TL)

src/hooks/video/agora/useAgoraScreenShare.ts
dmitrii-zolotukhin: @0xdevalias at the beginning I consider in return !localScreenTrack || !localAudioTrack. But the issue was that when tapping on the button "Stop sharing" in the browser, localScreenTrack and localAudioTrack are undefined, so the streaming can't be stoped.

0xdevalias: By ‘the previous way’ I mean the literal code change shown in this diff here. Without the guard statement at all, and using the ? when calling .leave

dmitrii-zolotukhin: @0xdevalias so it's ok now, right?

src/hooks/video/agora/useAgoraScreenShare.ts
dmitrii-zolotukhin: @0xdevalias fixed in https://github.com/sparkletown/sparkle/pull/1653/commits/1a95556a6dfecdc953e99c7e694c1f6b703180f9

src/hooks/video/agora/useAgoraScreenShare.ts
dmitrii-zolotukhin: @0xdevalias fixed in https://github.com/sparkletown/sparkle/pull/1653/commits/1a95556a6dfecdc953e99c7e694c1f6b703180f9

src/hooks/video/agora/useAgoraCamera.ts
dmitrii-zolotukhin: @0xdevalias fixed in 1a95556

src/hooks/video/agora/useAgoraCamera.ts
dmitrii-zolotukhin: fixed in 24b3cd8ba03f64a21bea2d35a249cf817facc979

PR #1656, opened 2021-06-25
[env/ohbm] fix PostersProvider causing posters to re-randomize on render

goran-peoski-work:
Originally the intent was to store first randomization of posters and reused it on re-renders of `PostersProvider` but instead the check
```javascript
if (isEmptyQuery && isPostersLoaded && previousPosterVenues.length === 0) {
}
```
wasn't properly done because of a typo.


0xdevalias (approved):
Change makes sense. LGTM!

PR #1657, opened 2021-06-25
add showSchedule toggle to admin v1 + clean up legacy showLiveSchedule + a few bonus bugfixes

0xdevalias:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/409
partially (in admin v1) fixes https://github.com/sparkletown/internal-sparkle-issues/issues/740
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/715

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -193,7 +193,8 @@ const createVenueData = (data, context) => {
     owners,
     profile_questions: data.profile_questions,
     placement: { ...data.placement, state: PlacementState.SelfPlaced },
-    showLiveSchedule: data.showLiveSchedule ? data.showLiveSchedule : false,
+    // @debt find a way to share src/settings with backend functions, then use DEFAULT_SHOW_SCHEDULE here
+    showSchedule: data.showSchedule ? data.showSchedule : true,
     showChat: true,
     showRangers: data.showRangers || false,
     parentId: data.parentId,
@@ -547,8 +548,8 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     updated.roomVisibility = data.roomVisibility;
   }
 
-  if (typeof data.showLiveSchedule === "boolean") {
-    updated.showLiveSchedule = data.showLiveSchedule;
+  if (typeof data.showSchedule === "boolean") {
+    updated.showSchedule = data.showSchedule;
   }
 
   if (typeof data.showGrid === "boolean") {
@@ -670,8 +671,8 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
     updated.columns = data.columns;
   }
 
-  if (typeof data.showLiveSchedule === "boolean") {
-    updated.showLiveSchedule = data.showLiveSchedule;
+  if (typeof data.showSchedule === "boolean") {
+    updated.showSchedule = data.showSchedule;
   }
 
   if (typeof data.showBadges === "boolean") {
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -194,7 +194,8 @@ const createVenueData = (data, context) => {
     profile_questions: data.profile_questions,
     placement: { ...data.placement, state: PlacementState.SelfPlaced },
     // @debt find a way to share src/settings with backend functions, then use DEFAULT_SHOW_SCHEDULE here
-    showSchedule: data.showSchedule ? data.showSchedule : true,
+    showSchedule:
+      typeof data.showSchedule === "boolean" ? data.showSchedule : true,
     showChat: true,
     showRangers: data.showRangers || false,
     parentId: data.parentId,
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -7,7 +7,7 @@ import { faTicketAlt, faHome } from "@fortawesome/free-solid-svg-icons";
 
 import firebase from "firebase/app";
 
-import { PLAYA_VENUE_ID } from "settings";
+import { DEFAULT_SHOW_SCHEDULE, PLAYA_VENUE_ID } from "settings";
 import { IS_BURN } from "secrets";
 
 import { UpcomingEvent } from "types/UpcomingEvent";
@@ -115,6 +115,8 @@ export const NavBar: React.FC<NavBarPropsType> = ({
   const hasSovereignVenue = sovereignVenueId !== undefined;
 
   const shouldShowHomeButton = hasSovereignVenue && !isSovereignVenue;
+
+  const shouldShowSchedule = venue?.showSchedule ?? DEFAULT_SHOW_SCHEDULE;
 
   const isOnPlaya = pathname.toLowerCase() === venueInsideUrl(PLAYA_VENUE_ID);
 
@@ -330,19 +332,21 @@ export const NavBar: React.FC<NavBarPropsType> = ({
         </div>
       </header>
 
-      <div
-        className={`schedule-dropdown-backdrop ${
-          isEventScheduleVisible ? "show" : ""
-        }`}
-        onClick={hideEventSchedule}
-      >
-        <div className={navBarScheduleClassName}>
-          <NavBarSchedule
-            isVisible={isEventScheduleVisible}
-            venueId={venueId}
-          />
+      {shouldShowSchedule && (
+        <div
+          className={`schedule-dropdown-backdrop ${
+            isEventScheduleVisible ? "show" : ""
+          }`}
+          onClick={hideEventSchedule}
+        >
+          <div className={navBarScheduleClassName}>
+            <NavBarSchedule
+              isVisible={isEventScheduleVisible}
+              venueId={venueId}
+            />
+          </div>
         </div>
-      </div>
+      )}
 
       {/* @debt Remove back button from Navbar */}
       {hasBackButton && venue?.parentId && parentVenue?.name && (
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -224,14 +224,20 @@ export const NavBar: React.FC<NavBarPropsType> = ({
                   onClick={navigateToHomepage}
                 />
               )}
-              <div
-                className={`nav-party-logo ${
-                  isEventScheduleVisible && "clicked"
-                }`}
-                onClick={toggleEventSchedule}
-              >
-                {navbarTitle} <span className="schedule-text">Schedule</span>
-              </div>
+
+              {shouldShowSchedule ? (
+                <div
+                  className={`nav-party-logo ${
+                    isEventScheduleVisible && "clicked"
+                  }`}
+                  onClick={toggleEventSchedule}
+                >
+                  {navbarTitle} <span className="schedule-text">Schedule</span>
+                </div>
+              ) : (
+                <div>{navbarTitle}</div>
+              )}
+
               <VenuePartygoers venueId={venueId} />
             </div>
 
--- a/src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
+++ b/src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
@@ -1,6 +1,8 @@
 import React, { useCallback, useMemo } from "react";
 import { Modal } from "react-bootstrap";
 
+import { DEFAULT_SHOW_SCHEDULE } from "settings";
+
 import { Room, RoomType } from "types/rooms";
 import { AnyVenue, VenueEvent } from "types/venues";
 
@@ -61,6 +63,7 @@ export const RoomModal: React.FC<RoomModalProps> = ({
           room={room}
           venueName={venue.name}
           venueEvents={venueEvents}
+          showSchedule={venue.showSchedule}
         />
       </div>
     </Modal>
@@ -71,12 +74,14 @@ export interface RoomModalContentProps {
   room: Room;
   venueName: string;
   venueEvents: WithVenueId<WithId<VenueEvent>>[];
+  showSchedule?: boolean;
 }
 
 export const RoomModalContent: React.FC<RoomModalContentProps> = ({
   room,
   venueName,
   venueEvents,
+  showSchedule = DEFAULT_SHOW_SCHEDULE,
 }) => {
   const dispatch = useDispatch();
 
@@ -101,6 +106,8 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
   const enterRoomWithSound: () => void = _enterRoomWithSound;
 
   const renderedRoomEvents = useMemo(() => {
+    if (!showSchedule) return [];
+
     return venueEvents.map((event, index: number) => (
       <ScheduleItem
         // @debt Ideally event.id would always be a unique identifier, but our types suggest it
@@ -114,10 +121,9 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
         roomUrl={room.url}
       />
     ));
-  }, [enterRoomWithSound, room.url, venueEvents]);
+  }, [enterRoomWithSound, room.url, showSchedule, venueEvents]);
 
-  // @debt do we want to show/hide the schedule on RoomModal based on venue.showSchedule?
-  const hasRoomEvents = renderedRoomEvents?.length > 0;
+  const showRoomEvents = showSchedule && renderedRoomEvents.length > 0;
 
   const iconStyles = {
     backgroundImage: room.image_url ? `url(${room.image_url})` : undefined,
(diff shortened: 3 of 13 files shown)


0xdevalias:
> I've just created https://github.com/sparkletown/sparkle/pull/1657 that removes the old 'show/hide live schedule' toggle, and adds a new 'show/hide schedule' toggle in its place (within admin v1)
> 
> - Cleaned up related aspects across the platform.
> - Added some `@debt` comments to `RoomModal` to decide if we want to also hide the schedule/events from there as well.
> - Haven't considered admin v2/3 at all in this and generally expect it's likely to be OOS
> 
> If more work needs to be done on this PR, it can likely be assigned to one of the newer engineers to polish/finish off. Otherwise, IMO, it's pretty much good to go.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/409#issuecomment-868188725_

functions/venue.js
```diff
@@ -193,7 +193,8 @@ const createVenueData = (data, context) => {
     owners,
     profile_questions: data.profile_questions,
     placement: { ...data.placement, state: PlacementState.SelfPlaced },
-    showLiveSchedule: data.showLiveSchedule ? data.showLiveSchedule : false,
+    // @debt find a way to share src/settings with backend functions, then use DEFAULT_SHOW_SCHEDULE here
+    showSchedule: data.showSchedule ? data.showSchedule : true,
```
denisdimitrov, line 197: This can be simplified to:
```suggestion
    showSchedule: data.showSchedule || true,
```

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -222,14 +224,20 @@ export const NavBar: React.FC<NavBarPropsType> = ({
...
-                  isEventScheduleVisible && "clicked"
-                }`}
-                onClick={toggleEventSchedule}
-              >
-                {navbarTitle} <span className="schedule-text">Schedule</span>
-              </div>
+
+              {shouldShowSchedule ? (
+                <div
+                  className={`nav-party-logo ${
```
denisdimitrov, line 230: You can use `classNames` to align this with our latest styles.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -330,19 +338,21 @@ export const NavBar: React.FC<NavBarPropsType> = ({
...
-        onClick={hideEventSchedule}
-      >
-        <div className={navBarScheduleClassName}>
-          <NavBarSchedule
-            isVisible={isEventScheduleVisible}
-            venueId={venueId}
-          />
+      {shouldShowSchedule && (
+        <div
+          className={`schedule-dropdown-backdrop ${
```
denisdimitrov, line 343: `classNames` here as well.

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -82,6 +82,7 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
     onend: enterRoom,
   });
 
+  // @debt do we want to show/hide the schedule on RoomModal based on venue.showSchedule?
```
denisdimitrov, line 85: [NB] This is a very good question, should we ask product for input or is it fine like the way it is now?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -536,15 +537,16 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </div>
   );
 
-  const renderLiveScheduleToggle = () => (
+  const renderShowScheduleToggle = () => (
```
denisdimitrov, line 540: Should we `useMemo` this? It doesn't look like a very expensive component but still...

src/types/venues.ts
```diff
@@ -160,7 +160,6 @@ export interface BaseVenue {
     text: string;
   };
   showLearnMoreLink?: boolean;
-  liveScheduleOtherVenues?: string[];
```
denisdimitrov, line 163: It's cool that you caught this. How did you do it though? Did you search for `liveSchedule` in this file and just noticed the extra field?

denisdimitrov (dismissed):

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: I was opting for minimal touch, if you enable the ‘hide white space changes’ mode of the PR diff review you’ll see that that’s existing code, indenting just changed because of the conditional rendering I added.

Guven urgency and workload of GitHub features needed, I don’t have bandwidth to spend time polishing existing tech debt (unfortunately)

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: Probably worth asking @sparkletown/product, as then we can finish it off quickly all in one PR

src/pages/Admin/Venue/DetailsForm.tsx
0xdevalias: No real benefit. It’s a function that will always return the same JSX, so memoing it will have do nothing different. Because the function is called to produce its output rather than being passed as a handler, it doesn’t matter that it’s not a stable function reference.

it’s tech debt for sure, and a pattern we want to eliminate from the codebase. But this being admin v1 legacy, not worth spending the time to refactor.

src/types/venues.ts
0xdevalias: Haha yup! And then my IDE has a ‘find references’ thing that told me it was unused. Then I did a find across all files for the name to confirm and also couldn’t find usages.

functions/venue.js
sunny-viktoryia, line 197: is this logic correct?
what if `showSchedule` is `false`? it'll be set to `true` then. Looks like it can will never be `false` here. 🤔 

shouldn't it be
```suggestion
    showSchedule: data.showSchedule ?? true,
```

sunny-viktoryia (approved):
Looks good to me! 🙃 

Just have a doubt in one place. Please, explain if this is a desired behavior.

functions/venue.js
0xdevalias: Unfortunately the backend functions are using JavaScript rather than TypeScript at the moment (I have an outdated WIP draft PR to resolve that.. one day...) so we don't have access to using `??` I don't believe.

You're right in saying that the code here seems like a bug though.. Really what it should be is something like:
```
typeof data.showSchedule === 'boolean' ? data.showSchedule : true
```

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1657/commits/77bee1af53d58d4907b2ec8f7d1015628a388e71 + https://github.com/sparkletown/sparkle/pull/1657/commits/a07d7ead64853134ba1b8af92f2ceb7dd05bc4a6

`venue.showSchedule` turned on
image

`venue.showSchedule` turned off
image

functions/venue.js
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1657/commits/987fd5b68c808862c6e849f92e14d10391b26b14

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.scss
```diff
@@ -26,8 +26,4 @@
     margin-bottom: 10px;
     white-space: pre-wrap;
   }
-
-  .room-entry-button {
-    width: 100%;
-  }
```
0xdevalias, line 32: [note] Moved to `src/components/templates/PartyMap/components/RoomModal/RoomModal.scss` as `&__btn-enter`

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -114,13 +139,23 @@ export const RoomModalContent: React.FC<RoomModalContentProps> = ({
...
+          {showSchedule && <RoomModalOngoingEvent roomEvents={venueEvents} />}
+
+          <button
+            className="btn btn-primary room-modal__btn-enter"
+            onMouseOver={triggerAttendance}
+            onMouseOut={clearAttendance}
+            onClick={enterRoomWithSound}
+          >
+            Enter
+          </button>
```
0xdevalias, line 153: [note] This was previously in `RoomModalOngoingEvent`

src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx
```diff
@@ -68,21 +74,40 @@ export interface RoomModalContentProps {
...
+
+  // @debt do we need to keep this retainAttendance stuff (for counting feature), or is it legacy tech debt?
+  const triggerAttendance = useCallback(() => {
+    dispatch(retainAttendance(true));
+  }, [dispatch]);
+
+  // @debt do we need to keep this retainAttendance stuff (for counting feature), or is it legacy tech debt?
+  const clearAttendance = useCallback(() => {
+    dispatch(retainAttendance(false));
+  }, [dispatch]);
```
0xdevalias, line 96: [note] This was previously in `RoomModalOngoingEvent`

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -1,45 +1,28 @@
-import React, { useCallback } from "react";
```
0xdevalias, line 1: [note] All of this stuff removed from this file was moved into `src/components/templates/PartyMap/components/RoomModal/RoomModal.tsx`

src/components/templates/PartyMap/components/RoomModalOngoingEvent/RoomModalOngoingEvent.tsx
```diff
@@ -52,27 +35,19 @@ export const RoomModalOngoingEvent: React.FC<RoomModalOngoingEventProps> = ({
...
-        className="btn btn-primary room-entry-button"
-        onClick={onRoomEnter}
-      >
-        {joinButtonText ?? "Enter"}
-      </button>
     </div>
   );
 };
-
-export default RoomModalOngoingEvent;
```
0xdevalias, line 78: [note] This didn't appear to ever be used.

0xdevalias (commented):
A few notes to help speed up review.

malaikaaiyar (approved):

functions/venue.js
goran-peoski-work: Something doesn't feel right here.

What happens if `data.showSchedule` is explicitly set to `false`?
Will it continue being such after `false || true`?

I think if `true` is supposed to be a default in case the value is missing 
i.e. is `null`/`undefined`, 
the operator should be `??` or the equivalent `?:` check for older JS

goran-peoski-work: NVM, found the resolution to it https://github.com/sparkletown/sparkle/pull/1657#discussion_r659145693 although an explicit test for boolean... `typeof new Boolean()` all's well I guess, no1 should be making these wrapper objects anyway

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
```diff
@@ -67,7 +67,9 @@ export const ScheduleItem: React.FunctionComponent<PropsType> = ({
             <RenderMarkdown text={event.description} />
           </div>
         </div>
+
         {isCurrentEvent && (
+          //  @debt Does ScheduleItem__entry-room-button even exist? Can we remove this?
```
sunny-viktoryia, line 72: It does exist. It appears only for Live events.
image

But I've just tested. When you click this button it opens in the two tabs: current and a new one. 😬 
So looks like we can remove the `@debt` comment and `externalUrlAdditionalProps` from here. 🙃

sunny-viktoryia (approved):
Awesome! 🙃 

could you also fix a minor bug that is slightly related to the files you've changed? That would be great!

src/components/templates/PartyMap/components/ScheduleItem/ScheduleItem.tsx
sunny-viktoryia: also. Might be worth renaming the variable `isCurrentEvent` -> `isLiveEvent` ?

goran-peoski-work: Could there be some ambiguity  if just `isLiveEvent`?

Then maybe something like `isEventLiveNow`, `isEventOngoingATM`...

0xdevalias: @goran-peoski-work Not really. An event is live or it isn’t. That aligns to our other current naming, including related helpers.

0xdevalias: @sunny-viktoryia i’ll have to check again, but I’m sure I searched for the code and couldn’t find its definition, which is why I added the debt comment. I’ll check again.

I’ll also check the bug you mention here and fix it up too.

0xdevalias: Ok, just confirmed, I was searching for the full class name, but obviously because we use `&` I needed to search for `__entry-room-button`. Can confirm that it exists, will remove the `@debt`

0xdevalias: > But I've just tested. When you click this button it opens in the two tabs: current and a new one. 😬
> So looks like we can remove ... externalUrlAdditionalProps from here. 🙃

So this is likely happening because there is both the `onClick={onRoomEnter}` as well as the `href={roomUrl}`.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1657/commits/fe21a57ae1a329c3ef583fbd123d57a73f9dbfea by using `e.preventDefault()`

- https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault
  - > The `Event` interface's `preventDefault()` method tells the user agent that if the event does not get explicitly handled, its default action should not be taken as it normally would be. The event continues to propagate as usual, unless one of its event listeners calls `stopPropagation()` or `stopImmediatePropagation()`, either of which terminates propagation at once.
  
Basically, this says "when the `onClick` handler runs, don't also let the `<a href>`'s default run, and thus will prevent opening 2 windows.

Note that there are also a few nuances around using `<a href>` with regards to opening rooms/etc, which could impact user presence/counting. I've written more about it in https://github.com/sparkletown/sparkle/issues/1670

0xdevalias: bonus: fixed 'onend' not firing in `useCustomSound` when the sound sprite is missing from the config in https://github.com/sparkletown/sparkle/pull/1657/commits/8820af929165bc9e340a1a2b9b18965e7a3e8db3 (which just happened to occur on the room I picked to test all of this on, and was stopping me from being able to enter

0xdevalias: > Might be worth renaming the variable `isCurrentEvent` -> `isLiveEvent` ?

> Could there be some ambiguity if just isLiveEvent?
> 
> Then maybe something like isEventLiveNow, isEventOngoingATM...

fixed in https://github.com/sparkletown/sparkle/pull/1657/commits/f909b4868fa06bf9de677a88cf7363f53139886e by removing the `isCurrentEvent` prop from `ScheduleItem` entirely, and just calculating it within the component itself. Now called `isCurrentEventLive` to disambiguate it from the function from `utils/time` which is named `isEventLive`

0xdevalias: bonus fix: renamed `ScheduleItem`'s 'passively named' `onRoomEnter` prop to the 'active action named' `enterEventLocation` in https://github.com/sparkletown/sparkle/pull/1657/commits/9e445c8601c73da181b4c962d8cc13b5f8094c3c

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: [OOS] Unrelated tech debt, future cleanup.

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: [OOS] Unrelated tech debt, future cleanup.

PR #1665, opened 2021-06-25
Change the number of dates to show in the schedule

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/849

see the comment https://github.com/sparkletown/internal-sparkle-issues/issues/849#issuecomment-868837055

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -622,7 +622,7 @@ export const USER_STATUSES = [UserStatus.available, UserStatus.busy];
 // @debt probably would be better to adjust max hour based on user's display size
 export const SCHEDULE_MAX_START_HOUR = 16;
 export const SCHEDULE_HOUR_COLUMN_WIDTH_PX = 200;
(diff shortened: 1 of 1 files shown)


0xdevalias (changes requested):
See my comments in https://github.com/sparkletown/internal-sparkle-issues/issues/849#issuecomment-868851248 as to why I don’t think this will be sufficient.

0xdevalias:
@sunny-viktoryia FYI: I just pushed a draft PR (https://github.com/sparkletown/sparkle/pull/1671) that would allow us to set the end date of the 'party'/event, and have the `NavBarSchedule` not show any days beyond that. IMO it supersedes this PR  in a more generalised way. Would be great if you can have a look sometime :)

mike-lvov:
This PR is irrelevant as the event has finished and it was specifically for GitHub
//cc @sunny-viktoryia

PR #1667, opened 2021-06-25
[env/github] map SAML SSO data from Okta into user's profile data

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/767

Get data from Okta and apply it to the app.
Adds new user fields.

~~Will need to base it off `env/github`, once it's updated with the latest `staging`~~ This has been done now


src/hooks/useSAMLSignIn.ts
```diff
@@ -12,9 +59,28 @@ export const useSAMLSignIn = (samlAuthProviderId?: string) => {
...
+        if (
+          !result.additionalUserInfo?.isNewUser ||
+          !result.user?.uid ||
+          !samlProfileMappings
+        )
+          return;
+
+        const mappedProfileData = getMappedValues(samlProfileMappings, result);
+        // TODO: Make a utils sessionStorage. Move PrefillProfileData into const/enum
+        sessionStorage.setItem(
```
0xdevalias, line 79: This is probably ok for now, but eventually this would probably be done some other way. Canonically I would say redux, but given we navigate with full page loads for now that won’t work. Probably good to add an @ debt comment here.

0xdevalias: We also have a file where we aim to centralise local storage/etc things. Like a utils file.

src/pages/Account/Profile.tsx
```diff
@@ -60,6 +62,15 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
     history.push(IS_BURN ? `/enter/step3` : nextUrl);
   };
 
+  // TODO: use functions from utils/sessionStorage or similar to retrieve the data
+  const data = sessionStorage.getItem("profileData") ?? "{}";
+  const refinedData = JSON.parse(data);
+
+  const realName =
```
0xdevalias, line 69: This should probably handle the case where we only have one name or the other? Could you do [firstName,lastName].filter(isTruthy).join(“ “) or similar instead?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -180,8 +187,15 @@ const VenuePage: React.FC = () => {
     return <LoadingPage />;
   }
 
-  if (!user) {
-    return <Login venue={venue} />;
+  if (!user || isLoginLoadingShown) {
+    return (
+      <Login
+        venue={venue}
+        isLoadingShown={isLoginLoadingShown}
```
0xdevalias, line 194: Do we need to pass these controls through like this? What happens if we only show the loading state within that component itself? Why does the loading page need to be controlled out here at this component’s level?

src/types/User.ts
```diff
@@ -5,6 +5,8 @@ export interface Experience {
   column?: number;
 }
 
+export type Mapping = { name: string; path: string };
```
0xdevalias, line 8: Couldn’t we just do this as a standard object? Where we use the key for name, and the oath is the value. Rather than needing to ‘double up’ like this.

then we could simplify implementation where it’s used as well to just be Object.entries(mapping).map(([name, path]) => [name, get(foo, path)]).filter(where value not defined) etc

There might even be a direct .mapValues or similar for the entries, can’t remember if that’s standard or just a lib I used once.

src/hooks/useSAMLSignIn.ts
mike-lvov: > This is probably ok for now, but eventually this would probably be done some other way. Canonically I would say redux, but given we navigate with full page loads for now that won’t work. Probably good to add an @ debt comment here.

Yep, but not only this. I guess we would also want to handle a case, when something happens to a connection or a person closes the tab. We still want to have that data saved. My initial approach was to use URL params, but given the way the flow implemented now, it would take drastic changes to achieve it.

src/pages/VenuePage/VenuePage.tsx
mike-lvov: That's because of the initial idea. My first approach was to just create a user + update his profile based on the data provided, but it was a unachievable because of the current implementation (user auth is checked higher in the hierarchy which causes `VenuePage` removal from the tree and addition afterwards)

mike-lvov: Currently, I don't think we need it to be on the root level though, will move it to the local one

src/hooks/useSAMLSignIn.ts
0xdevalias: The user will always auth via okta though, so we should be able to always read this same info off their login session I believe. Even directly in the create profile screen. So presumably we wouldn’t need to ‘cache and read back’ it at all?

mike-lvov: Changed in https://github.com/sparkletown/sparkle/pull/1667/commits/15dfaa4a6348a87a57db3386e982e12358de521c

src/pages/Account/Profile.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/15dfaa4a6348a87a57db3386e982e12358de521c

src/pages/VenuePage/VenuePage.tsx
mike-lvov: Removed in https://github.com/sparkletown/sparkle/pull/1667/commits/15dfaa4a6348a87a57db3386e982e12358de521c

src/types/User.ts
mike-lvov: Partially done in https://github.com/sparkletown/sparkle/pull/1667/commits/15dfaa4a6348a87a57db3386e982e12358de521c

I didn't find any value in removing `reduce` from there. It does filtering + mapping per single go-through. Even though it's just 5-6 fields, I don't think mapping + filtering adds any cognitive value

src/hooks/useSAMLSignIn.ts
0xdevalias: My original thoughts with using the profile was that we could just pass the URL directly into the profile image field within firestore rather than needing explicit logic in `src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx`/etc. Obviously if they chose to upload a new image then this method would no longer automagically work. Maybe it could just be an option shown on the 'default avatars' page when we have a `githubHandle` for them. Shown alongside the other 'defaults'. And pre-selected as the main avatar to use on initial profile creation.

0xdevalias: Actually.. disregard.. I think that may be more or less exactly what you did here.. `ProfilePictureInput` != `UserProfilePicture` as I originally read it as.

src/types/User.ts
0xdevalias: Since it’s already written it’s somewhat moot, but it’s less overall boilerplate code required to implement, which is why I suggested it. Simplifies the code down dramatically. Tradeoff is that filter will run through it a second time

src/utils/profile.ts
```diff
@@ -21,3 +25,14 @@ export const updateProfileEnteredVenueIds = async (
...
+export const getMappedValues = (mappings: Mappings, source: Object) =>
+  Object.entries(mappings).reduce((acc, [name, path]) => {
+    const mappingValue: string = get(source, path);
+
+    if (!isDefined(mappingValue)) {
+      return acc;
+    }
+
+    return { ...acc, [name]: mappingValue };
+  }, {});
```
0xdevalias, line 38: This seems to show type errors for me:
image

Seems `Object.entries`/`Object.fromEntries` as I suggested in https://github.com/sparkletown/sparkle/pull/1667#discussion_r659067758 isn't typesafe: https://github.com/microsoft/TypeScript/issues/35745

I tried playing around with a few methods.. this was the best I came up with:
```typescript
export const getMappedValues = <
  TObject extends object,
  TKey extends keyof TObject
>(
  mappings: Record<string, TKey>,
  source: TObject
) => mapValues(mappings, (value) => get(source, value, undefined));
```

**Further Reading:**

- https://lodash.com/docs/4.17.15#get
- https://lodash.com/docs/4.17.15#mapValues

0xdevalias: If we end up using the new method described above, it's super generic, so belongs somewhere in general utils rather than `src/utils/profile.ts` IMO.

(though ultimately it's basically just wrapping those lodash methods anyway, so there may be no point having it extracted as a function and we could just call those methods inline where we need them?

0xdevalias: (note: my method above may not actually filter the undefined values come to think of it.. unsure if that's needed.. expecting it may not matter to filter them?)

src/types/User.ts
```diff
@@ -20,6 +20,8 @@ export interface Experience extends TalkShowStudioExperience {
   column?: number;
 }
 
+export type Mappings = Record<string, string>;
```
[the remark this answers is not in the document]
0xdevalias: If we even need it as an explicit type.. I'm somewhat inclined not to; It's sort of just adding a layer of indirection without 'adding much value' IMO.

src/pages/Account/Login.tsx
```diff
@@ -38,9 +39,18 @@ export const Login: React.FC<LoginProps> = ({
...
+    show: showLoginLoading,
+    hide: hideLoginLoading,
+  } = useShowHide(false);
+
+  const { signInWithSAML, hasSamlAuthProviderId } = useSAMLSignIn({
+    samlAuthProviderId: venue.samlAuthProviderId,
+    samlProfileMappings: venue.samlProfileMappings,
+    showLoginLoading,
+    hideLoginLoading,
+  });
```
0xdevalias, line 53: What's the point of having the `useShowHide` out here, just to pass it's values into `useSAMLSignIn`?

Surely we should just track the loading state within that provider itself, and return the appropriate status?

I would probably also destructure the keys off `venue` too:

Something like:

```suggestion
  const { samlAuthProviderId, samlProfileMappings } = venue;
  
  const { signInWithSAML, hasSamlAuthProviderId, isSAMLLoginLoading } = useSAMLSignIn({
    samlAuthProviderId,
    samlProfileMappings,
  });
```

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
     history.push(IS_BURN ? `/enter/step3` : nextUrl);
   };
 
+  const profilePrefilledData = getLocalStorageItem(
+    LocalStorageItem.prefillProfileData
+  );
```
0xdevalias, line 70: I believe we can read the provided data as claims off the firebase auth identity token always can't we? So do we even need to use `localStorage` like this?

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
         <div className="login-welcome-subtitle">
           {`Swing back and edit your profile anytime you like.`}
         </div>
-        <a
-          className="profile-picture-button Profile__summit-snap"
-          href="https://virtual.githubphotobooth.com/virtual/capture/gr99n"
-          {...externalUrlAdditionalProps}
-        >
-          Take a Summit snap
-        </a>
```
0xdevalias, line 82: [note] This is moved further down in this same file

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
             )}
-            {errors.partyName &&* errors.partyName.type === "maxLength" && (
+            {errors.partyName && errors.partyName.type === "maxLength" && (
               <span className="input-error">
                 Display name must be {DISPLAY_NAME_MAX_CHAR_COUNT} characters or
                 less
               </span>
-            )} */}
+            )}
+            {/* TODO: Proper text, namings */}
```
0xdevalias, line 120: > TODO: Proper text, namings

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+            <p className="input-info">This is your title</p>
+            <input
+              name="companyDepartment"
+              className="input-block input-centered"
+              placeholder="Your department"
+              defaultValue={profilePrefilledData?.companyDepartment}
+              ref={register()}
+              autoComplete="off"
+            />
+            <p className="input-info">This is your department</p>
```
0xdevalias, line 147: Will each of these need to have error display added to them?

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+            <a
+              className="profile-picture-button Profile__summit-snap"
+              href="https://virtual.githubphotobooth.com/virtual/capture/gr99n"
+              {...externalUrlAdditionalProps}
+            >
+              Take a Summit snap
+            </a>
```
0xdevalias, line 154: Not really a fan of hardcoding the link like this.. but it's probably fine..

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +65,98 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
             {user && (
               <ProfilePictureInput
                 venueId={venueId}
                 setValue={setValue}
+                githubHandle={githubHandle}
```
0xdevalias, line 159: Should this also just be a field we set on the user profile? Then we could just read it off `user`?

src/hooks/useSAMLSignIn.ts
```diff
@@ -12,9 +39,28 @@ export const useSAMLSignIn = (samlAuthProviderId?: string) => {
       samlAuthProviderId
     );
 
+    if (showLoginLoading) {
+      showLoginLoading();
+    }
```
0xdevalias, line 44: If we just use `useAsync` instead of `useCallback` we'll have proper loading status built in for free.

src/hooks/useSAMLSignIn.ts
```diff
@@ -12,9 +39,28 @@ export const useSAMLSignIn = (samlAuthProviderId?: string) => {
...
+          !samlProfileMappings
+        )
+          return;
+
+        const mappedProfileData = getMappedValues(samlProfileMappings, result);
+
+        setLocalStorageItem(
+          LocalStorageItem.prefillProfileData,
+          mappedProfileData
+        );
```
0xdevalias, line 62: I believe we can read the provided data as claims off the firebase auth identity token always can't we? So do we even need to use `localStorage` like this?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -132,7 +132,17 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
             <div className="profile-extras">
               {renderedProfileQuestionAnswers}
             </div>
+            {/* Proper styling + add editing possibility */}
```
0xdevalias, line 135: > Proper styling + add editing possibility

0xdevalias: `UserProfileModal` should presumably always just be static content?

`EditProfileForm` appears to be the component used for changing party name/profile picture/etc. That's where it would make sense for all of this editing to happen.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -132,7 +132,17 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
...
             <div>{renderedProfileLinks}</div>
+            {selectedUserProfile.realName && (
+              <p>{selectedUserProfile.realName}</p>
+            )}
+            {selectedUserProfile.companyDepartment && (
+              <p>{selectedUserProfile.companyDepartment}</p>
+            )}
+            {selectedUserProfile.companyTitle && (
+              <p>{selectedUserProfile.companyTitle}</p>
+            )}
```
0xdevalias, line 145: [nit] May as well destructure these like all of the other values here?

src/pages/Account/Account.scss
```diff
@@ -76,6 +76,26 @@ $page-max-width: 1240px;
   padding-top: $spacing--xxl;
```
0xdevalias, line 76: [note] I haven't really reviewed these styles at all.

src/types/venues.ts
```diff
@@ -145,6 +145,7 @@ export interface BaseVenue {
   miniAvatars?: boolean;
   adultContent?: boolean;
   samlAuthProviderId?: string;
+  samlProfileMappings?: Mappings;
```
0xdevalias, line 148: As called out elsewhere, i'm not sure that we 'gain anything' by using `Mappings` here, vs just using the underlying types directly.

0xdevalias: And if we are going to use it, it would seem that the type should live in this file, not `User`. And probably have a more specific name to communicate its purpose, eg. `SAMLProfileMappings` or similar.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -27,6 +33,7 @@ export interface ProfilePictureInputProps {
   pictureUrl: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   register: any;
+  githubHandle?: string;
```
0xdevalias, line 36: If we just added this directly to the user profile then we could read it off `user`

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -92,6 +100,15 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
   };
 
+  const githubImageSrc =
+    githubHandle && `https://github.com/${githubHandle}.png?size=120`;
+
+  useEffect(() => {
+    if (!githubImageSrc) return;
+
+    setValue("pictureUrl", githubImageSrc, true);
+  }, [githubImageSrc, setValue]);
```
0xdevalias, line 110: Keeping it together like this will make it easier to see the relation.

```suggestion
  useEffect(() => {    
    if (!githubHandle) return;

    const githubImageSrc =
    githubHandle && `https://github.com/${githubHandle}.png?size=120`;
    
    setValue("pictureUrl", githubImageSrc, true);
  }, [githubHandle, setValue]);
```

0xdevalias: Seemingly we use `githubImageSrc` later on. Though i'm confused why we need the `setValue` here then?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -107,21 +124,34 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
+    ),
+    [uploadDefaultAvatar]
+  );
+
+  const gitHubImage = useMemo(
+    () =>
+      githubImageSrc &&
+      getRenderedAvatarImageComponent(githubImageSrc, "github-avatar"),
+    [githubImageSrc, getRenderedAvatarImageComponent]
+  );
```
0xdevalias, line 149: Guard pattern

```suggestion
  const gitHubImage = useMemo(
    () => {
      if (!githubImageSrc)
      
      return getRenderedAvatarImageComponent(githubImageSrc, "github-avatar")
    },
    [githubImageSrc, getRenderedAvatarImageComponent]
  );
```

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -107,21 +124,34 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
+
+  const avatarImages = useMemo(
+    () => defaultAvatars.map(getRenderedAvatarImageComponent),
+    [defaultAvatars, getRenderedAvatarImageComponent]
+  );
```
0xdevalias, line 154: Why not combine this together in a single `useMemo`, instead of splitting it across 3 (`getRenderedAvatarImageComponent `, `gitHubImage`, `avatarImages`)?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -156,6 +186,7 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
         Feeling camera shy? Choose from one of these Summit profile avatars.
       </small>
       <div className="default-avatars-container">
+        {gitHubImage}
         {isLoading ? <div>Loading...</div> : avatarImages}
```
0xdevalias, line 190: I would probably just display these together personally. Showing one but not the others seems like a weird UX flow.

0xdevalias (changes requested):
A bunch of comments/changes/etc. I'll probably try and fix these up today to save time since I assume you're offline at the moment.

src/pages/Account/Profile.tsx
0xdevalias: This is where I believe it would live. Not 100% though, and can't really test as I don't have SSO creds:

```
firebase
  .auth()
  .currentUser?.getIdTokenResult()
  .then(
    (token) => token.claims
  );
```

src/hooks/useSAMLSignIn.ts
0xdevalias: See further discussion in https://github.com/sparkletown/sparkle/pull/1667#discussion_r659470758

src/pages/Account/Profile.tsx
0xdevalias: See also: https://github.com/sparkletown/sparkle/pull/1667#discussion_r659472381

src/pages/Account/Profile.tsx
0xdevalias: I believe these also may need/want to be added to `EditProfileForm`

src/pages/Account/Profile.tsx
0xdevalias: Mappings paths for reference:
image

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: @0xdevalias to make it the default picture

0xdevalias: Fair

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: At the point, when `ProfilePictureInput` is rendered, the user object is not yet updated and we retrieve this data from LocalStorage. In future, its value used as `partyName` and there is no sense in duping it

src/types/User.ts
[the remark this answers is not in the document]
mike-lvov: Changed in https://github.com/sparkletown/sparkle/pull/1667/commits/2d4f2617700b9bb64a4edaab5c88b7bcb223540c

src/utils/profile.ts
```diff
@@ -1,7 +1,11 @@
 import firebase from "firebase/app";
+import get from "lodash/get";
```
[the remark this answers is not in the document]
mike-lvov: not relevant with the latest changes

src/utils/profile.ts
```diff
@@ -1,7 +1,11 @@
...
 import { FirebaseReducer } from "react-redux-firebase";
-import { User } from "types/User";
+
+import { User, Mappings } from "types/User";
+
 import { isTruthy } from "./types";
+import { isDefined } from "utils/types";
```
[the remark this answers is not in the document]
mike-lvov: Not relevant with the latest changes

src/utils/profile.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/2d4f2617700b9bb64a4edaab5c88b7bcb223540c + https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

src/pages/Account/Login.tsx
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

mike-lvov: Since the logic change, the changes are drastic

src/pages/Account/Profile.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

src/utils/localStorage.ts
```diff
@@ -8,3 +8,23 @@ export const getLocalStorageToken = (venueId: string) =>
...
+export const getLocalStorageItem = (
+  itemKey: LocalStorageItem
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): any | undefined => {
+  const itemData = localStorage.getItem(itemKey);
+
+  if (itemData === null) return;
+
+  return JSON.parse(itemData);
+};
```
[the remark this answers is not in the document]
mike-lvov: Fixed the logic in https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

src/hooks/useSAMLSignIn.ts
mike-lvov: Removed in https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f but you can't use `useAsync` there since it's happening inside the function

src/hooks/useSAMLSignIn.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

src/types/venues.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/2d4f2617700b9bb64a4edaab5c88b7bcb223540c

src/pages/Account/Profile.tsx
mike-lvov: Discussion happening in https://github.com/sparkletown/sparkle/pull/1667#discussion_r659567943

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/1667/commits/8c56ac87d7d1fd51d6b29db5df212b7a3a948efc

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: Fixes in https://github.com/sparkletown/sparkle/pull/1667/commits/8c56ac87d7d1fd51d6b29db5df212b7a3a948efc

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: Fixes in https://github.com/sparkletown/sparkle/pull/1667/commits/8c56ac87d7d1fd51d6b29db5df212b7a3a948efc

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/3da671cd41b87e2e89594ea2e62bcb2fb7fde73d

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Implemented proper styling in https://github.com/sparkletown/sparkle/pull/1667/commits/3da671cd41b87e2e89594ea2e62bcb2fb7fde73d

mike-lvov: image

mike-lvov: Will add the editing, but it should not be a blocker

src/pages/Account/Profile.tsx
mike-lvov: 1) They are optional + no restrictions since they will be fetched from their Okta.
2) as per https://github.com/sparkletown/sparkle/pull/1667#discussion_r659621274 will work on this now, but it can be reviewed without that

src/pages/Account/Profile.tsx
mike-lvov: removes comment as per implicit `approval` from @whoaitsaimz as per https://doesitsparkle.slack.com/archives/C0229C1EC86/p1624842181410800?thread_ts=1624838827.397500&cid=C0229C1EC86

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: Fair. Makes sense.

src/pages/Account/Profile.tsx
0xdevalias: Just to confirm, did you test this and see that it works? I was assuming, and couldn't test it since I don't have creds.

src/pages/Account/Profile.tsx
0xdevalias: I agreed it was fine in https://github.com/sparkletown/sparkle/pull/1667#discussion_r659636177, so this can probably be resolved.

src/hooks/useSAMLSignIn.ts
0xdevalias: > but you can't use `useAsync` there since it's happening inside the function

What do you mean by that?

There are 3 different variations of `useAsync`, do none of them work here?

- https://github.com/streamich/react-use/blob/master/docs/useAsync.md
- https://github.com/streamich/react-use/blob/master/docs/useAsyncRetry.md

src/components/organisms/UserProfileModal/UserProfileModal.tsx
0xdevalias: Looks good. Sounds good.

src/pages/Account/Profile.tsx
mike-lvov: @0xdevalias yes, I did test it

src/pages/Account/Profile.tsx
mike-lvov: Using this thread to track the progress
https://github.com/sparkletown/sparkle/pull/1667#discussion_r659473073

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -107,21 +124,37 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
-    ));
-  }, [defaultAvatars, uploadDefaultAvatar]);
+    ),
+    [uploadDefaultAvatar]
+  );
+
+  const avatarImages = useMemo(() => {
+    const renderedAvatarImages = defaultAvatars.map(
+      getRenderedAvatarImageComponent
+    );
```
0xdevalias, line 147: I still don't see why we would bother keeping `getRenderedAvatarImageComponent` instead of just inlining it within `avatarImages`.

`defaultAvatars` is a `string[]`

`githubImageSrc` is a `string | undefined`

So just do something like:

```typescript
[githubAvatarImage, ...defaultAvatars].filter(isDefined).map((avatarSrc) =>
  <div
    key={avatarSrc}
    // ..snip..
```

src/components/organisms/UserProfileModal/UserProfileModal.scss
```diff
@@ -28,7 +28,7 @@ $profile-link--padding: 6px 18px;
   }
 }
 
-.question {
+.no-margin {
   margin-bottom: 0;
 }
```
0xdevalias, line 33: Can we move this definition inside the `.UserProfileModal` scope rather than being here in global scope?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -67,7 +75,7 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
 
             return (
               <React.Fragment key={question.text}>
-                <p className="light question">{question.text}</p>
+                <p className="light no-margin">{question.text}</p>
                 <h6>{questionAnswer}</h6>
               </React.Fragment>
             );
```
0xdevalias, line 81: [OOS] But the usage of `h6` here is SUPER  janky. Header tags shouldn't be used for styling like this. This breaks all sorts of accessibility rules.

0xdevalias: Actually.. bootstrap has classes that mirror the header styling. Can you change this to a div or something and use the h6 class at the very least? Should be a quick fix for it i'm guessing.

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -109,7 +117,7 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
               <div className="profile-pic">
                 {/* @debt Refactor this to use our useImage hook? Or just UserAvatar / UserProfilePicture directly? */}
                 <img
-                  src={selectedUserProfile.pictureUrl || DEFAULT_PROFILE_PIC}
+                  src={pictureUrl || DEFAULT_PROFILE_PIC}
```
0xdevalias, line 120: Should this be using `??`
```suggestion
                  src={pictureUrl ?? DEFAULT_PROFILE_PIC}
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -124,15 +132,31 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
                 />
               </div>
               <div className="profile-text">
-                <h2 className="italic">
-                  {selectedUserProfile.partyName || DEFAULT_PARTY_NAME}
-                </h2>
+                <h2 className="italic">{partyName || DEFAULT_PARTY_NAME}</h2>
```
0xdevalias, line 135:
```suggestion
                <h2 className="italic">{partyName ?? DEFAULT_PARTY_NAME}</h2>
```

src/components/organisms/UserProfileModal/UserProfileModal.tsx
```diff
@@ -124,15 +132,31 @@ export const UserProfileModal: React.FC<UserProfileModalProps> = ({
...
+                <p className="light no-margin">Department</p>
+                <h6>{companyDepartment}</h6>
+              </>
+            )}
+            {companyTitle && (
+              <>
+                <p className="light no-margin">Title</p>
+                <h6>{companyTitle}</h6>
+              </>
+            )}
```
0xdevalias, line 159: Same "don't abuse `h6` for styling" comments as above. Bootstrap has a class that mirrors the heading styles I believe. At the very least use that.

src/hooks/useSAMLSignIn.ts
```diff
@@ -1,8 +1,23 @@
...
+  signInWithSAML: () => void;
+}
+
+export const useSAMLSignIn: ReactHook<
+  UseSAMLSignInProps,
+  UseSAMLSignInReturn
+> = ({ samlAuthProviderId }) => {
   const hasSamlAuthProviderId = samlAuthProviderId !== undefined;
 
   const signInWithSAML = useCallback(() => {
```
0xdevalias, line 23: Use one of the `useAsync` variants here as called out elsewhere in previous review thread.

src/pages/Account/Login.tsx
```diff
@@ -38,9 +39,15 @@ export const Login: React.FC<LoginProps> = ({
...
-  );
+  const {
+    isShown: isLoginLoadingShown,
+    // show: showLoginLoading,
+    // hide: hideLoginLoading,
+  } = useShowHide(false);
+
+  const { signInWithSAML, hasSamlAuthProviderId } = useSAMLSignIn({
+    samlAuthProviderId: venue.samlAuthProviderId,
+  });
```
0xdevalias, line 50: What's the need/point of `useShowHide` here now? Use one of the `useAsync` variants within `useSAMLSignIn` instead.

src/pages/Account/Profile.tsx
```diff
@@ -10,20 +13,25 @@ import { DISPLAY_NAME_MAX_CHAR_COUNT, DEFAULT_VENUE } from "settings";
...
 import { externalUrlAdditionalProps } from "utils/url";
 
 import { useVenueId } from "hooks/useVenueId";
 import { useUser } from "hooks/useUser";
+import { useConnectCurrentVenueNG } from "hooks/useConnectCurrentVenueNG";
 
 import { updateUserProfile } from "./helpers";
 
 import { ProfilePictureInput } from "components/molecules/ProfilePictureInput";
+import { LoadingPage } from "components/molecules/LoadingPage";
```
0xdevalias, line 26: Alphabetic
```suggestion
import { LoadingPage } from "components/molecules/LoadingPage";
import { ProfilePictureInput } from "components/molecules/ProfilePictureInput";
```

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +76,114 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+            path,
+            undefined
+          );
+        })
+      );
+  }, [samlProfileMappings]);
+
+  const githubHandle = samlPrefillData?.githubName;
+
+  const realName = [samlPrefillData?.firstName, samlPrefillData?.lastName]
```
0xdevalias, line 103:
```suggestion
  const {
    githubName: githubHandle,
    firstName,
    lastName,
    companyTitle,
    companyDepartment,
  } = samlPrefillData ?? {};

  const realName = [firstName, lastName]
```

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +76,114 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
-        <a
-          className="profile-picture-button Profile__summit-snap"
-          href="https://virtual.githubphotobooth.com/virtual/capture/gr99n"
-          {...externalUrlAdditionalProps}
-        >
-          Take a Summit snap
-        </a>
         <form onSubmit={handleSubmit(onSubmit)} className="form">
           <div className="input-group profile-form">
             {/* @debt - this input is hidden because for github's event they fetch the names from an OKTA, and for the demo we set the default name to Mona */}
```
0xdevalias, line 123: The input is no longer hidden, so this seems invalid now?
```suggestion
```

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +76,114 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+              defaultValue={realName}
+              ref={register()}
+              autoComplete="off"
+            />
+            <p className="input-info">This is your real name</p>
+            <input
+              name="companyTitle"
+              className="input-block input-centered"
+              placeholder="Your title"
+              defaultValue={samlPrefillData?.companyTitle}
```
0xdevalias, line 161:
```suggestion
              defaultValue={companyTitle}
```

src/pages/Account/Profile.tsx
```diff
@@ -61,57 +76,114 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
+              ref={register()}
+              autoComplete="off"
+            />
+            <p className="input-info">This is your title</p>
+            <input
+              name="companyDepartment"
+              className="input-block input-centered"
+              placeholder="Your department"
+              defaultValue={samlPrefillData?.companyDepartment}
```
0xdevalias, line 170:
```suggestion
              defaultValue={companyDepartment}
```

src/utils/profile.ts
```diff
@@ -1,6 +1,8 @@
 import firebase from "firebase/app";
 import { FirebaseReducer } from "react-redux-firebase";
+
 import { User } from "types/User";
+
 import { isTruthy } from "./types";
```
0xdevalias, line 6:
```suggestion
import { isTruthy } from "utils/types";
```

0xdevalias (changes requested):
Few more changes needed here.

src/types/User.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/2d4f2617700b9bb64a4edaab5c88b7bcb223540c + https://github.com/sparkletown/sparkle/pull/1667/commits/c77cd255fb597da0dd2611bc66089e77dbe0e59f

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Why are they junky though? Note that this is for `env/github` only

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: @0xdevalias note that this is targeted to `GitHub` and probably we won't use these fields anywhere except this

src/hooks/useSAMLSignIn.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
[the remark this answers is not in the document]
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/components/organisms/UserProfileModal/UserProfileModal.scss
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/hooks/useSAMLSignIn.ts
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/pages/Account/Login.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/36f9ee2f4cfc1f5342c9ef7c0843ba5452393d9f

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Not sure which bootstrap you are referring to, but it looks like here the h1 tags are also used.
https://mdbootstrap.com/docs/standard/navigation/headers/

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: I can't find what you are referring to here either
https://hackerthemes.com/bootstrap-cheatsheet/

mike-lvov:
Done the editing 
image

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
```diff
@@ -99,6 +103,33 @@ export const EditProfileForm: React.FunctionComponent<EditProfileFormProps> = ({
...
+        />
+
+        <div className="EditProfileForm__edit-input-title">Title</div>
+        <InputField
+          containerClassName="EditProfileForm__input-container-name"
+          inputClassName="EditProfileForm__input-name"
+          name="companyTitle"
+          placeholder="Your title"
+          ref={register()}
+        />
```
0xdevalias, line 131: Instead of using `div` for the ‘titles’ of these edit fields, can you use `label`? And if so, can you nest the `InputField` within them? That way it will use proper semantic html/accessibility/etc

src/components/organisms/UserProfileModal/UserProfileModal.tsx
0xdevalias: 'Bootstrap' refers to https://getbootstrap.com/

- https://getbootstrap.com/docs/4.0/content/typography/#headings
  - > `.h1` through `.h6` classes are also available, for when you want to match the font styling of a heading but cannot use the associated HTML element.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
0xdevalias: Why did you use the `index` as part of the `key`?

src/components/organisms/ProfileModal/EditProfileForm/EditProfileForm.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/b1a496133c781e522058ff4474875191d420b179

src/components/organisms/UserProfileModal/UserProfileModal.tsx
0xdevalias: See link + explicit references above in https://github.com/sparkletown/sparkle/pull/1667#discussion_r659707405

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
mike-lvov: @0xdevalias legacy logic

src/hooks/useSAMLSignIn.ts
0xdevalias: I think you'll need to `await` or `return` that `firebase.auth.etc` `Promise` for this to work properly?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/1667/commits/41dc9b23e005d6755989c6684484c7568675d1d9

Out of curiousity, hoes does it effect the `all sorts of accessibility rules`?

src/components/organisms/UserProfileModal/UserProfileModal.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/commit/41dc9b23e005d6755989c6684484c7568675d1d9

src/components/organisms/UserProfileModal/UserProfileModal.tsx
0xdevalias: I've explained this way back in the past. But basically headings have a very specific and defined semantic meaning in HTML structure, that is used by screen readers/etc.

This was from a quick google, there may be other/better resources:

- https://www.w3.org/WAI/tutorials/page-structure/headings/

0xdevalias (approved):
I believe this comment still needs addressing: https://github.com/sparkletown/sparkle/pull/1667#discussion_r659710497

But aside from that, I think this looks good to go.

Note: I've only code reviewed, not clickthrough/etc. But if you're happy with how it looks/functions, LGTM! :shipit:

PR #1671, opened 2021-06-27
add optional end_utc_seconds to Venue + use it to limit days shown in NavBarSchedule

0xdevalias:
supercedes https://github.com/sparkletown/sparkle/pull/1665

partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/849

When `end_utc_seconds` is present, it will be used to limit the days shown on `NavBarSchedule` to the greater of `SCHEDULE_SHOW_DAYS_AHEAD` or the remaining days until `end_utc_seconds`.

When there are no 'day tabs' to show (eg. `end_utc_seconds` has already passed), then a different UI is rendered, explaining that the scheduled events have ended.

Also fixes a number of bugs related to how 'time shifted' dates are handled in `NavBarSchedule`, so that the schedule properly advances each day when `start_utc_seconds` is configured. Previously it looks like it would have always stayed static with the first 'day tab' of the schedule always being `start_utc_seconds`, rather than being 'today'. This would have meant that a party/event that had `start_utc_seconds` configured, and lasted longer than `SCHEDULE_SHOW_DAYS_AHEAD`, would have it's schedule cut off before the end of the event.

Note that I explicitly chose to use `end_utc_seconds` to match the existing `start_utc_seconds`, even though neither fit our naming patterns. I consider that to be future tech-debt to clean up together, and OOS of this PR. The 'risk'/'impact' of not having them aligned is 'worse' than the cost of them not matching our naming patterns.

## Notes on testing

- You can use https://www.epochconverter.com/ to easily get the required UTC seconds value
- There is no UI for configuring the Venue's `end_utc_seconds` in this PR, so you'll need to set it directly within the firestore DB
- I don't believe the `NavBarSchedule` live-updates when it's venue config changes, so you may need to do a full page reload between updating the venue config before being able to see the output change

### DIFF
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -3,7 +3,7 @@ import {
   addDays,
   format,
   fromUnixTime,
-  isToday,
+  isBefore,
   startOfDay,
   startOfToday,
 } from "date-fns";
@@ -78,13 +78,21 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
 
   const scheduledStartDate = sovereignVenue?.start_utc_seconds;
 
-  const firstDayOfSchedule = useMemo(() => {
-    return scheduledStartDate
+  const { firstDayOfSchedule, isScheduleTimeshifted } = useMemo(() => {
+    const today = startOfToday();
+    const firstDayOfEvent = scheduledStartDate
       ? startOfDay(fromUnixTime(scheduledStartDate))
-      : startOfToday();
-  }, [scheduledStartDate]);
+      : today;
 
-  const isScheduleTimeshifted = !isToday(firstDayOfSchedule);
+    const isScheduleTimeshifted = isBefore(today, firstDayOfEvent);
+
+    const firstDayOfSchedule = isScheduleTimeshifted ? firstDayOfEvent : today;
+
+    return {
+      firstDayOfSchedule,
+      isScheduleTimeshifted,
+    };
+  }, [scheduledStartDate]);
 
   const {
     isEventsLoading,
@@ -178,11 +186,11 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
   const hasSavedEvents = schedule.personalEvents.length > 0;
 
   const downloadPersonalEventsCalendar = useCallback(() => {
-    const dayStart = addDays(startOfToday(), selectedDayIndex);
+    const startOfSelectedDay = addDays(firstDayOfSchedule, selectedDayIndex);
     const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
       .map(
         prepareForSchedule({
-          day: dayStart,
+          day: startOfSelectedDay,
           usersEvents: userEventIds,
           isForCalendarFile: true,
         })
@@ -193,7 +201,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
       calendar: createCalendar({ events: allPersonalEvents }),
       calendarName: `${PLATFORM_BRAND_NAME}_Personal`,
     });
-  }, [relatedVenueEvents, userEventIds, selectedDayIndex]);
+  }, [firstDayOfSchedule, selectedDayIndex, relatedVenueEvents, userEventIds]);
 
   const downloadAllEventsCalendar = useCallback(() => {
     downloadCalendar({
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useMemo, useState } from "react";
 import {
   addDays,
+  differenceInCalendarDays,
   format,
   fromUnixTime,
   isBefore,
@@ -27,6 +28,7 @@ import {
 import { WithVenueId } from "utils/id";
 import { range } from "utils/range";
 import { formatDateRelativeToNow } from "utils/time";
+import { isDefined } from "utils/types";
 
 import { useRelatedVenues } from "hooks/useRelatedVenues";
 import { useUser } from "hooks/useUser";
@@ -77,6 +79,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
   });
 
   const scheduledStartDate = sovereignVenue?.start_utc_seconds;
+  const scheduledEndDate = sovereignVenue?.end_utc_seconds;
 
   const { firstDayOfSchedule, isScheduleTimeshifted } = useMemo(() => {
     const today = startOfToday();
@@ -94,6 +97,11 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     };
   }, [scheduledStartDate]);
 
+  const lastDayOfSchedule = useMemo(() => {
+    return scheduledEndDate
+      ? startOfDay(fromUnixTime(scheduledEndDate))
+      : undefined;
+  }, [scheduledEndDate]);
   const {
     isEventsLoading,
     events: relatedVenueEvents = emptyRelatedEvents,
@@ -105,7 +113,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
 
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
-  const weekdays = useMemo(() => {
+  const renderedScheduleDayTabs = useMemo(() => {
     const formatDayLabel = (day: Date | number) => {
       if (isScheduleTimeshifted) {
         return format(day, "E, LLL d");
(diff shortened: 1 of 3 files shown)


src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -77,15 +79,29 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+    const isScheduleTimeshifted = isBefore(today, firstDayOfEvent);
 
-  const isScheduleTimeshifted = !isToday(firstDayOfSchedule);
+    const firstDayOfSchedule = isScheduleTimeshifted ? firstDayOfEvent : today;
 
+    return {
+      firstDayOfSchedule,
+      isScheduleTimeshifted,
+    };
+  }, [scheduledStartDate]);
```
0xdevalias, line 98: [note] Previously `firstDayOfSchedule` would always be set to the date of `scheduledStartDate` (aka `start_utc_seconds`) when configured. This meant that the schedule would never advance forward each day, and would be able to show at most `SCHEDULE_SHOW_DAYS_AHEAD` days before being unable to show 'today' anymore.

The logic used for `isScheduleTimeshifted` also seemed strange/likely buggy.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -77,15 +79,29 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+
+  const lastDayOfSchedule = useMemo(() => {
+    return scheduledEndDate
+      ? startOfDay(fromUnixTime(scheduledEndDate))
+      : undefined;
+  }, [scheduledEndDate]);
```
0xdevalias, line 104: [note] It doesn't matter that we used `startOfDay` here, as we only care about the date portion since we use `differenceInCalendarDays`; so the time within the day doesn't really matter.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,7 +113,9 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
 
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
-  const weekdays = useMemo(() => {
+  // TODO: extract a new useMemo'd  const startOfSelectedDay = addDays(firstDayOfSchedule, selectedDayIndex);
+  //  Then reuse that in all the places that are currently duplicating this logic
```
0xdevalias, line 117: [TODO] extract a new `useMemo`'d  `const startOfSelectedDay = addDays(firstDayOfSchedule, selectedDayIndex);` then reuse that in all the places that are currently duplicating this logic

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -108,7 +126,22 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+            startOfToday()
+          )
+        )
+      : undefined;
+
+    const scheduleDaysToShowCount = isDefined(daysTillScheduleEnded)
+      ? Math.min(daysTillScheduleEnded, SCHEDULE_MAX_DAYS_AHEAD)
+      : SCHEDULE_MAX_DAYS_AHEAD;
+
+    return range(scheduleDaysToShowCount).map((dayIndex) => {
```
0xdevalias, line 144: [note] This is the logic for determining how many days are left till `lastDayOfSchedule`, and showing either the remaining days, or `SCHEDULE_MAX_DAYS_AHEAD` if there are many days to go (or if `end_utc_seconds` isn't defined)

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -143,7 +181,8 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     const daysEvents = relatedVenueEvents
       .filter(
         isScheduleTimeshifted
-          ? isEventWithinDate(startOfSelectedDay)
+          ? // @debt do we need this ternary here? Can we just always use one of isEventWithinDate / isEventWithinDateAndNotFinished ?
+            isEventWithinDate(startOfSelectedDay)
           : isEventWithinDateAndNotFinished(startOfSelectedDay)
       )
```
0xdevalias, line 187: [debt] do we need this ternary here? Can we just always use one of `isEventWithinDate` / `isEventWithinDateAndNotFinished` ?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -178,11 +217,12 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+    const startOfSelectedDay = addDays(firstDayOfSchedule, selectedDayIndex);
+    // @debt this seems to only download events for the currently selected day, yet the UI implies as though it will download all of my events
     const allPersonalEvents: PersonalizedVenueEvent[] = relatedVenueEvents
       .map(
         prepareForSchedule({
-          day: dayStart,
+          day: startOfSelectedDay,
           usersEvents: userEventIds,
           isForCalendarFile: true,
         })
```
0xdevalias, line 228: [debt] this seems to only download events for the currently selected day, yet the UI implies as though it will download all of my events

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -214,6 +254,8 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
           {hasSavedEvents && (
             <Button
               onClick={downloadPersonalEventsCalendar}
+              // @debt downloadPersonalEventsCalendar seems to only download events for the currently selected day,
+              //  yet the UI here implies as though it will download all of my events
               customClass="NavBarSchedule__download-schedule-btn"
             >
               Download your schedule
```
0xdevalias, line 261: [debt] `downloadPersonalEventsCalendar` seems to only download events for the currently selected day, yet the UI here implies as though it will download all of my events

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -228,9 +270,20 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          <ul className="NavBarSchedule__weekdays">
+            {renderedScheduleDayTabs}
+          </ul>
+
+          <Schedule isLoading={isLoadingSchedule} {...schedule} />
+        </>
+      ) : (
+        // TODO: Improve how this displays
+        <div>Scheduled events have ended</div>
+      )}
```
0xdevalias, line 286: [TODO] refactor this to be cleaner + follow our standards better + improve how "Scheduled events have ended" displays

0xdevalias (commented):
Notes to assist when reviewing

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
goran-peoski-work: What happens when  `today` is greater than the last day of the event? Any UI/UX concerns about accessing historical data?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
goran-peoski-work:
```
 isDefined(daysTillScheduleEnded)
      ? Math.min(daysTillScheduleEnded, SCHEDULE_MAX_DAYS_AHEAD)
      : SCHEDULE_MAX_DAYS_AHEAD;
```
Seems like an excessive logic for not setting `daysTillScheduleEnded = Infinity` since `Math.min(Infinity, X ) -> X`

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: @sparkletown/product and/or @sunny-viktoryia is probably best to weigh in here, but my understanding is that the feature as originally implemented/designed would ‘move forward’ and not show past days. I believe this got broken with some open source contributed work.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: I have no idea where you’re getting ‘infinity’ from here? But I assume you’re making an incorrect assumption about the value of SCHEDULE_MAX_DAYS_AHEAS, which is set to something like 7

0xdevalias: If you have a suggestion for how to simplify it and maintain the existing functionality, feel free to suggest.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -108,7 +126,22 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          differenceInCalendarDays(
+            addDays(lastDayOfSchedule, 1),
+            startOfToday()
+          )
+        )
+      : undefined;
+
+    const scheduleDaysToShowCount = isDefined(daysTillScheduleEnded)
+      ? Math.min(daysTillScheduleEnded, SCHEDULE_MAX_DAYS_AHEAD)
+      : SCHEDULE_MAX_DAYS_AHEAD;
```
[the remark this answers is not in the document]
0xdevalias: Not sure I follow, but I refer back to my previous comment in https://github.com/sparkletown/sparkle/pull/1671/files/699628ada951b2fb0cacbabcf33ee10436db4901#r659387807 that states that there is no use of infinity here at all, so I think you're going down the wrong path with your line of thinking.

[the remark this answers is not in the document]
goran-peoski-work: You're not sure you follow, but you think I'm going down the wrong path with my line of thinking... 😕

From the link you shared:
> I have no idea where you’re getting ‘infinity’ from here?

Yes.

So I will try to explain my explanation from above.

Where does `Infinity` come from?
Math: if x is a natural number, x is always less than infinity.

There is nothing special about the value `undefined` in the above context. It just as easily can be `Symbol('not the number you were looking for')` and still have the same 5 lines of checking a moniker value before doing the operation. You can also replace it with `Infinity` and still have the same 5 lines.

One thing the value `Infinity` has going for it, though, is it's type of `Number`. So comparing numbers gives:
sparkle-infinity-01

whoaitsaimz:
@goran-peoski-work the issue was closed but this PR is still open. Looks like there was a lot of discussion and work done here. Do we need to close this PR or should it be finalized / reviewed / merged?

goran-peoski-work:
@whoaitsaimz I think this one was superseded by a PR merged one or two weeks ago, just can't remember the number. You can just check staging if there is any need of improvement and open new issue since the schedule has been reworked and any continuing work on this PR may just take more time merging conflicts than actually solving any lingering issue at hand.

PR #1469, opened 2021-05-28
add Talk Show Studio - NEW venue type

kakarda653:
Working on feature FULL Talk Show Studio - NEW venue type - screen sharing + host + panel + audience  
- added new room type screenshare (working name)
- added the template for that room
- installed agora sdk
- implemented the streaming from camera and microphone
- implemented the streaming from screen with the sound 
- made the ability to disconnect devices and screen sharing by clicking the button

url for testing: http://localhost:3000/in/agoratest

closes sparkletown/internal-sparkle-issues#748

PR #1673, opened 2021-06-27
add secure dynamic token generation for agora video

0xdevalias:
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/748

See references/overview in 🔒 https://github.com/sparkletown/internal-sparkle-issues/issues/748#issuecomment-868132831

This will be used as part of the `TalkShowStudio` (added in https://github.com/sparkletown/sparkle/pull/1469), specifically where it uses Agora for it's video features instead of Twilio.

## TODO

- [x] Implement token generation as a backend function
- [x] Add the frontend `api/*` layer binding to the token generation function
- [x] Configure functions config for `staging`
  - `npx firebase --project staging functions:config:set agora.app_id=REDACTED agora.app_certificate=REDACTED`
- [x] Enable Agora 'co-host token authentication' in Admin console as described in https://github.com/AgoraIO/Tools/issues/83#issuecomment-869149777
  - Docs on how to do this: https://docs.agora.io/en/Interactive%20Broadcast/faq/token_cohost
  - This will also require changes to the app logic (which is OOS of this PR): https://docs.agora.io/en/Interactive%20Broadcast/faq/token_cohost#2-change-the-app-logic
- [x] Anything else?

### OOS

This PR is focussed on the backend functionality for fetching the Agora token, not for 'wiring it up' into the rest of the frontend app.

There will also be specific changes required to support the 'secure co-host' mode as per:

- https://docs.agora.io/en/Interactive%20Broadcast/faq/token_cohost
- https://docs.agora.io/en/Agora%20Platform/faq/host_set_role
  - Note: we likely wouldn't use the Agora RTM SDK as suggested here, as we can just 'communicate' via our own firestore database

Here is an example snippet for how you can fetch the token:

```typescript
  import React, { useEffect } from "react";
  import { useAsync } from "react-use";

  import { getAgoraToken } from "api/video";

  const {
    loading: isTokenLoading,
    error: tokenError,
    value: token,
  } = useAsync(async () => {
    // Note: This will fail if the user isn't logged in

    // channelName will be the 'video channel/room' that everyone connects to. This should be something secure/unguessable (UUID?)
    return getAgoraToken({ channelName: "fooChannel" });
  }, []);

  useEffect(() => {
    console.log("agoraToken", token);
  }, [token]);
```

## Further Reading

- https://docs.agora.io/en/Agora%20Platform/token#Token
  - > Token is a dynamic key generated by App ID, App Certificate, user ID, channel name, token expiration timestamp, and other information. For scenarios requiring high security, such as the production environment, Agora recommends using a token for authentication.
  - > If you are using an earlier version of one of these products, you can upgrade your product version or use Channel Keys for authentication.
  - > If you upgrade your product from a version that supports channel keys to a version that supports tokens, please refer to the Token Migration Guide.
    - https://docs.agora.io/en/Agora%20Platform/channel_key
      - https://docs.agora.io/en/Agora%20Platform/channel_key#step-4-use-a-channel-key
        - > The Channel Key is based on the App Certificate, App ID, Channel Name, Current Timestamp, Client User ID, and Lifespan Timestamp
        - > The Channel Key expires after a certain period of time. Your application must call `renewChannelKey` when a timeout occurs. The `onError` or `didOccurError` callback returns `ERR_CHANNEL_KEY_EXPIRED` (`109`).
  - https://docs.agora.io/en/Agora%20Platform/token#3-generate-a-token
    - > Tokens are generated on your app server. Agora provides an open-source AgoraDynamicKey repository in GitHub, which enables you to generate tokens on your server with programming languages such as C++, Java, Python, PHP, Ruby, Node.js, and Go. For detailed descriptions on the repository and how to generate a token with the sample code, see Generate a Token.
- https://docs.agora.io/en/Agora%20Platform/token_server
- https://docs.agora.io/en/Interactive%20Broadcast/token_server?platform=Web
- https://docs.agora.io/en/Video/token_server?platform=Web#token-generator-libraries
- https://github.com/AgoraIO/Tools/tree/master/DynamicKey/AgoraDynamicKey/nodejs
  - `npm i agora-access-token`
  - > IMPORTANT! Build token with either the uid or with the user account.
- https://github.com/AgoraIO-Community/TokenServer-nodejs
  - > One-Click deploy Agora sample token server
- https://www.section.io/engineering-education/agora-express-token-server/#generating-authentication-tokens
  - https://github.com/zolomohan/agora-express-token-server
- https://docs.agora.io/en/Agora%20Platform/faq/string
  - > How can I use string user names?
- https://stackoverflow.com/questions/54689020/create-a-channel-in-agora-io
  - > To make a video call Agora provides the channel concept. If you are starting a video call, you have to specify a channel name and the token for authentication. The other person also needs to join on the same channel to talk to each other.
  - > A Channel name can be any string within 64byte.
- https://agoraio-community.github.io/AgoraWebSDK-NG/docs/en/basic_call
  - > `token`: A token that identifies the role and privilege of the user if your project enables the App Certificate. See Use a token for authentication for details.
    - https://docs.agora.io/en/Agora%20Platform/token#Token

### DIFF
--- a/functions/auth.js
+++ b/functions/auth.js
@@ -9,10 +9,12 @@ const { assertValidUrl, assertValidVenueId } = require("./src/utils/assert");
 const { createOAuth2Client } = require("./src/utils/auth");
 const { getJson, postJson } = require("./src/utils/fetch");
 
+// @debt refactor lowercaseFirstChar into utils/* (or maybe remove it entirely..?)
 // Case-insensitive first character for iDevices
 const lowercaseFirstChar = (password) =>
   password.charAt(0).toLowerCase() + password.substring(1);
 
+// @debt refactor passwordsMatch into utils/*
 exports.passwordsMatch = (submittedPassword, actualPassword) =>
   submittedPassword.trim() === actualPassword.trim() ||
   lowercaseFirstChar(submittedPassword.trim()) ===
--- a/functions/auth.js
+++ b/functions/auth.js
@@ -9,19 +9,6 @@ const { assertValidUrl, assertValidVenueId } = require("./src/utils/assert");
 const { createOAuth2Client } = require("./src/utils/auth");
 const { getJson, postJson } = require("./src/utils/fetch");
 
-const PROJECT_ID = functions.config().project.id;
-
-const checkAuth = (context) => {
-  if (!context.auth || !context.auth.token) {
-    throw new functions.https.HttpsError("unauthenticated", "Please log in");
-  }
-
-  if (context.auth.token.aud !== PROJECT_ID) {
-    throw new functions.https.HttpsError("permission-denied", "Token invalid");
-  }
-};
-exports.checkAuth = checkAuth;
-
 // Case-insensitive first character for iDevices
 const lowercaseFirstChar = (password) =>
   password.charAt(0).toLowerCase() + password.substring(1);
--- a/functions/src/utils/agora.js
+++ b/functions/src/utils/agora.js
@@ -16,29 +16,6 @@ const getExpirationTime = () => {
   return currentTimestamp + expirationTimeInSeconds;
 };
 
-/**
- * @deprecated use generateAgoraTokenForAccount where possible so that we can use strings as account, rather than a numeric uid
- */
-const generateAgoraTokenForUid = ({ channelName, uid, role }) => {
-  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
-  if (typeof channelName !== "string" && channelName.length <= 0)
-    throw new Error("channelName must be a string");
-
-  if (typeof uid !== "number") throw new Error("uid must be a number");
-
-  if (![RtcRole.PUBLISHER, RtcRole.SUBSCRIBER].includes(role))
-    throw new Error("role must be a valid value from the RtcRole enum");
-
-  return RtcTokenBuilder.buildTokenWithUid(
-    appId,
-    appCertificate,
-    channelName,
-    uid,
-    role,
-    getExpirationTime()
-  );
-};
-
 const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
   // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
   if (typeof channelName !== "string" && channelName.length <= 0)
@@ -60,12 +37,5 @@ const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
 };
 
 exports.getExpirationTime = getExpirationTime;
-
-/**
- * @deprecated use generateAgoraTokenForAccount where possible so that we can use strings as account, rather than a numeric uid
- */
-exports.generateAgoraTokenForUid = generateAgoraTokenForUid;
-
 exports.generateAgoraTokenForAccount = generateAgoraTokenForAccount;
-
 exports.RtcRole = RtcRole;
--- a/functions/src/utils/agora.js
+++ b/functions/src/utils/agora.js
@@ -21,7 +21,8 @@ const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
   if (typeof channelName !== "string" && channelName.length <= 0)
     throw new Error("channelName must be a string");
 
-  if (typeof account !== "string") throw new Error("account must be a string");
+  if (typeof account !== "string" && account.length <= 0)
+    throw new Error("account must be a string");
 
   if (![RtcRole.PUBLISHER, RtcRole.SUBSCRIBER].includes(role))
     throw new Error("role must be a valid value from the RtcRole enum");
--- a/functions/src/utils/agora.js
+++ b/functions/src/utils/agora.js
@@ -1,0 +1,71 @@
+const functions = require("firebase-functions");
+
+const { RtcTokenBuilder, RtcRole } = require("agora-access-token");
+
+const AGORA_CONFIG = functions.config().agora;
+
+const appId = AGORA_CONFIG.app_id;
+const appCertificate = AGORA_CONFIG.app_certificate;
+
+// @debt should we move this into AGORA_CONFIG, or maybe venue config in firestore or similar?
+const expirationTimeInSeconds = 3600;
+
+const getExpirationTime = () => {
+  const currentTimestamp = Math.floor(Date.now() / 1000);
+
+  return currentTimestamp + expirationTimeInSeconds;
+};
+
+/**
+ * @deprecated use generateAgoraTokenForAccount where possible so that we can use strings as account, rather than a numeric uid
+ */
+const generateAgoraTokenForUid = ({ channelName, uid, role }) => {
+  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
+  if (typeof channelName !== "string" && channelName.length <= 0)
+    throw new Error("channelName must be a string");
+
+  if (typeof uid !== "number") throw new Error("uid must be a number");
+
+  if (![RtcRole.PUBLISHER, RtcRole.SUBSCRIBER].includes(role))
+    throw new Error("role must be a valid value from the RtcRole enum");
+
+  return RtcTokenBuilder.buildTokenWithUid(
+    appId,
+    appCertificate,
+    channelName,
+    uid,
+    role,
+    getExpirationTime()
+  );
+};
+
+const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
+  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
+  if (typeof channelName !== "string" && channelName.length <= 0)
+    throw new Error("channelName must be a string");
+
+  if (typeof account !== "string") throw new Error("account must be a string");
+
+  if (![RtcRole.PUBLISHER, RtcRole.SUBSCRIBER].includes(role))
+    throw new Error("role must be a valid value from the RtcRole enum");
+
+  return RtcTokenBuilder.buildTokenWithAccount(
+    appId,
+    appCertificate,
+    channelName,
+    account,
+    role,
+    getExpirationTime()
+  );
+};
+
+exports.getExpirationTime = getExpirationTime;
+
+/**
+ * @deprecated use generateAgoraTokenForAccount where possible so that we can use strings as account, rather than a numeric uid
+ */
+exports.generateAgoraTokenForUid = generateAgoraTokenForUid;
+
+exports.generateAgoraTokenForAccount = generateAgoraTokenForAccount;
+
+exports.RtcRole = RtcRole;
--- a/functions/src/utils/assert.js
+++ b/functions/src/utils/assert.js
@@ -1,7 +1,20 @@
+const functions = require("firebase-functions");
 const { HttpsError } = require("firebase-functions/lib/providers/https");
 
 const { isValidUrl } = require("./url");
 const { checkIfValidVenueId } = require("./venue");
+
+const PROJECT_ID = functions.config().project.id;
+
+const assertValidAuth = (context) => {
+  if (!context.auth || !context.auth.token) {
+    throw new HttpsError("unauthenticated", "Please log in");
+  }
+
+  if (context.auth.token.aud !== PROJECT_ID) {
+    throw new HttpsError("permission-denied", "Token invalid");
+  }
+};
 
 const assertValidVenueId = (venueId, paramName = "venueId") => {
   if (!checkIfValidVenueId(venueId)) {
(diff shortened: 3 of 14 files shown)


functions/auth.js
```diff
@@ -9,23 +9,12 @@ const { assertValidUrl, assertValidVenueId } = require("./src/utils/assert");
...
-  if (!context.auth || !context.auth.token) {
-    throw new functions.https.HttpsError("unauthenticated", "Please log in");
-  }
-
-  if (context.auth.token.aud !== PROJECT_ID) {
-    throw new functions.https.HttpsError("permission-denied", "Token invalid");
-  }
-};
-exports.checkAuth = checkAuth;
-
```
0xdevalias, line 24: [note] moved to `functions/src/utils/assert.js`

functions/src/utils/assert.js
```diff
@@ -1,8 +1,21 @@
...
+const assertValidAuth = (context) => {
+  if (!context.auth || !context.auth.token) {
+    throw new HttpsError("unauthenticated", "Please log in");
+  }
+
+  if (context.auth.token.aud !== PROJECT_ID) {
+    throw new HttpsError("permission-denied", "Token invalid");
+  }
+};
+
```
0xdevalias, line 18: [note] moved from `functions/auth.js`

functions/src/utils/assert.js
```diff
@@ -18,5 +31,11 @@ const assertValidUrl = (url, paramName = "url") => {
...
 };
 
+exports.assertValidAuth = assertValidAuth;
 exports.assertValidVenueId = assertValidVenueId;
 exports.assertValidUrl = assertValidUrl;
+
+/**
+ * @deprecated use assertValidAuth export/naming instead
+ */
+exports.checkAuth = assertValidAuth;
```
0xdevalias, line 41: [note] This is so I don't have to rename all of the other usages across the backend functions, which is OOS of this PR and can be done in a future cleanup.

functions/src/utils/twilio.js
```diff
@@ -0,0 +1,28 @@
...
+  const token = generateTwilioToken();
+
+  token.addGrant(videoGrant);
+  token.identity = identity;
+
+  return token;
+};
+
+exports.generateTwilioToken = generateTwilioToken;
+exports.twilioVideoToken = twilioVideoToken;
```
0xdevalias, line 28: [note] This was moved out of `functions/video.js`

functions/video.js
```diff
@@ -1,46 +1,57 @@
...
-};
-
-const videoToken = (identity, room) => {
-  const videoGrant = new VideoGrant({ room });
-  const token = generateToken();
-  token.addGrant(videoGrant);
-  token.identity = identity;
-  return token;
-};
-
```
0xdevalias, line 25: [note] This was moved to `functions/src/utils/twilio.js`

functions/video.js
```diff
@@ -1,46 +1,57 @@
...
-exports.getToken = functions.https.onCall((data, context) => {
-  if (!context.auth || !context.auth.token) {
-    throw new functions.https.HttpsError("unauthenticated", "Please log in");
-  }
+const { RtcRole, generateAgoraTokenForAccount } = require("./src/utils/agora");
+const { assertValidAuth } = require("./src/utils/assert");
+const { twilioVideoToken } = require("./src/utils/twilio");
+
+// @debt either remove data.identity entirely, or validate that it matches the context.auth.uid
+//   (once checking that this won't break anything in the app)
```
0xdevalias, line 9: > either remove data.identity entirely, or validate that it matches the `context.auth.uid` (once checking that this won't break anything in the app)

[OOS] Unsure of the impact will have to the rest of the app, and I don't want to delay this PR based on that

functions/video.js
```diff
@@ -1,46 +1,57 @@
...
-    throw new functions.https.HttpsError("permission-denied", "Token invalid");
+  if (!data || !data.identity || !data.room) {
+    throw new HttpsError("invalid-argument", "identity or room data missing");
   }
 
-  if (data && data.identity && data.room) {
-    const token = videoToken(data.identity, data.room);
-    return {
-      token: token.toJwt(),
-    };
```
0xdevalias, line 39: [note] `assertValidAuth` handles the `context.auth` checks that used to be inline in this function

functions/video.js
```diff
@@ -1,46 +1,57 @@
...
+
+  return {
+    token: token.toJwt(),
+  };
+});
+
+exports.getAgoraToken = functions.https.onCall((data, context) => {
+  assertValidAuth(context);
+
+  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
```
0xdevalias, line 27: > we should enforce a stricter security requirement on channelName. Maybe use UUIDs?

[OOS] This is probably OOS for this PR, as it would require deeper integration with the rest of the frontend code, which isn't available on `staging` currently

functions/video.js
```diff
@@ -1,46 +1,57 @@
...
+  //   See my discovery + explanation + solution at:
+  //     https://github.com/AgoraIO/Tools/issues/83#issuecomment-869149777
+  //   tl;dr We need to enable 'Co-Host token authentication' in the Agora admin console
+  //     In a live streaming channel, when an audience member applies to co-host, you can use a token to authenticate whether the user can publish a stream. This feature is co-host token authentication.
+  //   We need to check against firebase to ensure that the user requesting the host permissions is actually allowed to have them.
+  const token = generateAgoraTokenForAccount({
+    channelName: data.channelName,
+    account: context.auth.uid,
+    role: RtcRole.PUBLISHER,
+  });
```
0xdevalias, line 52: [OOS] This is probably OOS for this PR, as it would require deeper integration with the rest of the frontend code, which isn't available on `staging` currently

src/api/video.ts
```diff
@@ -1,38 +1,60 @@
 import Bugsnag from "@bugsnag/js";
 import firebase from "firebase/app";
 
-export interface GetVideoTokenProps {
+export interface GetTwilioVideoTokenProps {
   userId: string;
   roomName: string;
-  onError?: (msg: string) => void;
-  onFinish?: () => void;
```
0xdevalias, line 8: [note] Refactored to remove this legacy pattern. There's no point to this sort of callback when we are using `Promise`. Nothing was using these callbacks in the codebase anyway.

functions/src/utils/agora.js
```diff
@@ -0,0 +1,41 @@
...
+
+const { RtcTokenBuilder, RtcRole } = require("agora-access-token");
+
+const AGORA_CONFIG = functions.config().agora;
+
+const appId = AGORA_CONFIG.app_id;
+const appCertificate = AGORA_CONFIG.app_certificate;
+
+// @debt should we move this into AGORA_CONFIG, or maybe venue config in firestore or similar?
+const expirationTimeInSeconds = 3600;
```
0xdevalias, line 11: > should we move this into AGORA_CONFIG, or maybe venue config in firestore or similar?

[note] Open to thoughts/suggestions on what we do here. For this PR it's probably good enough as is. Need to read into it a bit more to see what happens when they expire, how renewal is handled, and if it will disrupt an ongoing video call if it expires part way through. So probably OOS to change in this PR.

functions/src/utils/agora.js
```diff
@@ -0,0 +1,41 @@
...
+const getExpirationTime = () => {
+  const currentTimestamp = Math.floor(Date.now() / 1000);
+
+  return currentTimestamp + expirationTimeInSeconds;
+};
+
+const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
+  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
+  if (typeof channelName !== "string" && channelName.length <= 0)
+    throw new Error("channelName must be a string");
```
0xdevalias, line 22: > we should enforce a stricter security requirement on channelName. Maybe use UUIDs?

[OOS] This is probably OOS for this PR, as it would require deeper integration with the rest of the frontend code, which isn't available on `staging` currently

0xdevalias (commented):
Some notes to assist during review.

functions/src/utils/agora.js
yurii-lubynets, line 11: IMHO the `AGORA_CONFIG` should contain only Agora console-related stuff. And `expirationTimeInSeconds` is used for our own purposes. It's fine to set it in `src/settings.ts`

functions/src/utils/agora.js
```diff
@@ -0,0 +1,41 @@
...
+const expirationTimeInSeconds = 3600;
+
+const getExpirationTime = () => {
+  const currentTimestamp = Math.floor(Date.now() / 1000);
+
+  return currentTimestamp + expirationTimeInSeconds;
+};
+
+const generateAgoraTokenForAccount = ({ channelName, account, role }) => {
+  // @debt we should enforce a stricter security requirement on channelName. Maybe use UUIDs?
```
yurii-lubynets, line 20: UUID is the best option. But right now it's ok to have just a string

src/components/organisms/Room/Room.tsx
```diff
@@ -59,18 +65,14 @@ const Room: React.FC<RoomProps> = ({
...
-        room: roomName,
-      });
-      setToken(response.data.token);
-    })();
+    if (!user) return;
+
+    getTwilioVideoToken({
+      userId: user.uid,
+      roomName,
+    }).then(setToken);
```
yurii-lubynets, line 75: nice catch!

functions/src/utils/agora.js
0xdevalias: Backend functions don’t have access to src/settings

yurii-lubynets (approved):
Thank you @0xdevalias That looks efficient and sensible!
Anyway we need more reviewers to make sure it coverages all requirements

functions/src/utils/agora.js
sunny-viktoryia: > we should enforce a stricter security requirement on channelName. Maybe use UUIDs?

(another option) I guess channel name can be also generated using some hash function based on venueId.

functions/src/utils/agora.js
```diff
@@ -0,0 +1,41 @@
...
+  if (typeof channelName !== "string" && channelName.length <= 0)
+    throw new Error("channelName must be a string");
+
+  if (typeof account !== "string") throw new Error("account must be a string");
```
sunny-viktoryia, line 24: [NB] just to be consistent we could also check if `account` is not empty ( as we do for `channelName` few lines above)
```suggestion
  if (typeof account !== "string" && account.length <= 0)
    throw new Error("account must be a string");
```

sunny-viktoryia (approved):
Looks good to me. 🙃 

tested locally to get the token using your code example from the description. The token is received.

functions/src/utils/agora.js
0xdevalias: That's roughly close to what we do currently with Twilio, but i'm not super happy about it. If we do use a hash function, it would need to use secret data within the hash, which would need to be unguessable (since our code is open source, and venueId is a known value, 'security through obscurity' wouldn't be enough here.

Since we would need to store a secret value in firestore/etc anyways, we may as well just make that the whole value we use IMO.

0xdevalias: (For continuity, see also this thread: https://github.com/sparkletown/sparkle/pull/1673#discussion_r659327633 on the same topic)

functions/src/utils/agora.js
0xdevalias: For continuity, see also this thread https://github.com/sparkletown/sparkle/pull/1673#discussion_r659323441 on the same topic

functions/src/utils/agora.js
0xdevalias: Sounds like a good idea :)

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1673/commits/7176004c307d345508fb782d7713f4f0684cc672

functions/src/utils/agora.js
sunny-viktoryia: aha, I see. Makes sense. yep, forgot a little that we're open source. 😄

0xdevalias: Hehe. Even when not open source, security through obscurity isn't a great idea. Me and my friends have made a lot of bug bounty (ethical hacking) money from projects/companies that have relied on less than secure implementations :p

sunny-viktoryia: hahaha woooooow! 😲 Super cooool!

Where did you get all this security knowledge? Any resource that comes to your mind first(for newbies)? Or it's just accumulated by your experience? I remember you were at a security conference some time ago. Sounds like a really captivating topic to learn deeper.

0xdevalias: Mostly a lifetime of interest/being 'jacked in' to the internet + an insatiable curiosity for wanting to know how things work, and then how to make them work how I want them to rather than how they're meant to. Then i've bounced back and forth between dev and security as I burn out/get bored of one or the other 😛 

Hit me up on Slack/etc about it sometime when deadlines aren't as pressing and let's chat a bit more. I can likely dig up some good starting points/etc for you, depending on what level of interest/how deep you want to go/etc. :)

sunny-viktoryia: THAT
WOULD
BE
AWESOMEEE!!! 

Appreciate your will to help. Thanks!!! ❤️

PR #1676, opened 2021-06-27
fix AdminEventModal (admin v1) to show the correct errors for the 'room' field

malaikaaiyar:
Uses room validation for the room field instead of host validation

(note: there is no validation schema for the room; should it be a required field or something?)


0xdevalias:
> (note: there is no validation schema for the room; should it be a required field or something?)

Off the top of my head, it’s not required since events can be made that aren’t linked to rooms.

src/pages/Admin/AdminEventModal.tsx
```diff
@@ -214,8 +214,8 @@ const AdminEventModal: React.FunctionComponent<PropsType> = ({
                 ref={register}
                 value={roomName}
               />
-              {errors.host && (
-                <span className="input-error">{errors.host.message}</span>
+              {errors.room && (
```
sunny-viktoryia, line 217: Looks like `room` does not have any validation requirements. So we can remove this block `{errors.room && ...}` fully. 
But ideally would be to add a validation rule allowing only existent room names. But probably it's out of scope of this PR.

sunny-viktoryia (approved):
Approving this but if you have capacity would be awesome if you could add validation for `room` field (see my comment). 🙃

src/pages/Admin/AdminEventModal.tsx
malaikaaiyar: I'll merge this and make a new PR for that

PR #1678, opened 2021-06-28
[env/github] Replace back to label with back

alexey-petrov:
As per https://github.com/sparkletown/internal-sparkle-issues/issues/855
* Replaced `back to {href}` with `back` label for github only

### DIFF
--- a/src/pages/VenuePage/TemplateWrapper.tsx
+++ b/src/pages/VenuePage/TemplateWrapper.tsx
@@ -87,7 +87,7 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
             className="btn btn-primary"
             onClick={() => history.goBack()}
           >
-            Go Back
+            Back
           </button>
         </p>
       );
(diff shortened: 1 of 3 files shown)


0xdevalias (approved):
LGTM

0xdevalias (changes requested):
This definitely isn't all of the locations where this back text exists in the platform. I know for a fact at least the NavBar/similar also has it's own. And there may be more than that as well.

image

May be other locations, didn't check deeply.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -87,7 +87,7 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
             className="btn btn-primary"
             onClick={() => history.goBack()}
           >
-            Go Back
+            Back
```
alexey-petrov, line 90: as I can see this also pops up when user enters a venue, but it's a little different than the other places
guess it should be changed as well tho

alexey-petrov:
> This definitely isn't all of the locations where this back text exists in the platform. I know for a fact at least the NavBar/similar also has it's own. And there may be more than that as well.
> 
> image
> 
> May be other locations, didn't check deeply.

should be fixed now, thanks

yurii-lubynets (approved):

PR #1685, opened 2021-06-28
fix for venue chat name change when redirect back from room to venue map

alexey-petrov:
Fix for venue chat as per https://github.com/sparkletown/internal-sparkle-issues/issues/824
* Fixed venue chat name change when going back from room to venue map

closes https://github.com/sparkletown/internal-sparkle-issues/issues/824

Patch described by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/824#issuecomment-863848152


0xdevalias (approved):
[CRO] Code changes here look fine. I haven't clicked through and tested this, so if you're happy that it resolves the issue, then LGTM!

PR #1690, opened 2021-06-28
env/ohbm: RF revert soichi and revert-the-revert-of-Glen

yarikoptic:
At 285 attendees still be claimed to reside on the platform ATM (may be just an opened tab somewhere in a browser) I thought may be it would be better to stop them query every minute for a list of users as was needed to be done for OHBM 
 in #1630 ?  To say the truth I am not sure if it would make positive or negative impact on the number of requests and thus cost of still running OHBM sparkle instance but just couldn't stay silent seeing $s going wasted ;)

FWIW, tested locally - seems to work ok.


mike-lvov:
@yarikoptic 
Any difference from the one created a week ago?
https://github.com/sparkletown/sparkle/pull/1639

mike-lvov:
Something is different as per lines added/deleted
Mine: 
image

Yours:

yarikoptic:
I could check later whenever get to the laptop, some things are still tricky on the phone ;-)

mike-lvov:
@yarikoptic I would expect it to only have the revert of this commit: https://github.com/sparkletown/sparkle/commit/4cc90ba719f2a51b543b1593c7f94e9ced1259a1, but it has 2 commits inside. Is there any explicit intention for this?

yarikoptic:
the main goal was to revert @soichih's patch since it places instance on guaranteed periodic pull of updates.

Here is btw the diff between the 'backup' and reverts:
```
(git)lena:~/proj/misc/sparkle/sparkle[rf-revert-soichi]git
$> git diff origin/env/ohbm-backup..
diff --git a/src/hooks/posters.tsx b/src/hooks/posters.tsx
index 356121671..796934777 100644
--- a/src/hooks/posters.tsx
+++ b/src/hooks/posters.tsx
@@ -120,7 +120,7 @@ export const PostersProvider: React.FC<PostersProviderProps> = ({
   useEffect(() => {
     // only save the shuffled poster venues state if the previous ones were empty
     // additional logic can be added to (re-/in-)validate based on time, userId or other parameters
-    if (isEmptyQuery && isPostersLoaded && previousPosterVenues.length !== 0) {
+    if (isEmptyQuery && isPostersLoaded && previousPosterVenues.length === 0) {
     }
   }, [
diff --git a/src/settings.ts b/src/settings.ts
index 51d33d4fd..17812f80c 100644
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -11,6 +11,7 @@ import {
 } from "secrets";
 import { VenueTemplate } from "types/venues";
 import { RoomType } from "types/rooms";
+import { FIVE_MINUTES_MS } from "utils/time";
 import { UserStatus } from "types/User";
 
 import sparkleNavLogo from "assets/icons/sparkle-nav-logo.png";
@@ -89,8 +90,9 @@ export const DUST_STORM_TEXT_2 =
 // How often to refresh events schedule
 export const REFETCH_SCHEDULE_MS = 10 * 60 * 1000; // 10 mins
 
+// @debt FIVE_MINUTES_MS is deprecated; use utils/time or date-fns functions instead
 // How often to update location for counting
-export const LOC_UPDATE_FREQ_MS = 120 * 60 * 1000; // 120 mins
+export const LOC_UPDATE_FREQ_MS = FIVE_MINUTES_MS;
 
 // How often to increment user's timespent
 export const LOCATION_INCREMENT_SECONDS = 10;
```
didn't look in greater detail yet...

yarikoptic:
ok, the diff in posters.tsx is due to a4ea41f3af1ac89b2abcef7759308e92a25f403b which was a fix after the merge of soichi's fix  to fix randomization of the posters.  Was not sure how 2nd one came about, so decided to just redo the reversion and then I think we cross our  paths in github somehow -- I think you did the same reversion like
```
git revert -m 1 ec37dad678f451d2037b5d54331329add5e584f9
```
and pushed it, and I also force pushed that into my branch, but then how did this PR got closed by "me" not sure.

anyways -- we are were I thought we should arrive, which is good.  Let's hope it would reduce the $pendings per day which seems to remain substantial.

0xdevalias:
Duplicate of #1639

0xdevalias:
> I thought may be it would be better to stop them query every minute for a list of users as was needed to be done for OHBM
> in #1630 ? To say the truth I am not sure if it would make positive or negative impact on the number of requests and thus cost of still running OHBM sparkle instance but just couldn't stay silent seeing $s going wasted ;)

@yarikoptic Thanks for this PR! To provide some context, I am almost certain that this will have a positive impact on the requests/costs.

Without going too deeply into the specifics, our firebase costs have been relatively low/stable for across most environments we run. OHBM spiked in cost _significantly_, and from the interim exploration I have done to find out 'why', it seems to point back to @soichih's #1630, though there will almost certainly be some element of it that is just 'increased load' due to OHBM being a larger event in general (and having some 'unique specifics' that differ from previous events we have run, eg. having 1000+ related venues, etc)

In particular, I believe the way that #1630 is doing a full query every minute or so per user is loading vastly more data than the previous 'live update' method would have. The live update method will load the data the first time, but then only receive the tiny 'snapshot updates' each time. So if a particular user's record hasn't changed then no data will need to be sent down again, and it won't count as a 'read' against the firestore DB.

While I haven't looked too deeply at the code in #1630, in skimming it just now I see the following:

- `src/index.tsx` uses `setInterval` to `dispatch` `CacheActionTypes.RELOAD_USER_CACHE` every `60` seconds. Given this is so high up in the code, this will run while a user is connected to the platform in any way, even if they aren't logged in.
- `src/store/reducers/Cache.ts` handles 'reducing' the various redux actions that were added here. It violates one of the core principles that reducers should be pure functions, and calls async side effect code to load the entire `users` collection from firebase; then violates Redux's "never directly mutate the state, only return an updated state" to update it with the fetched records. This should have used `redux-thunk` or similar patterns for handling async things in redux. This will end up loading a LOT of unneeded data each time it's called, and is almost certainly the reason this drove up costs by several orders of magnitude. A hacky optimisation would have been to only query for records that had changed since the last time the query was run or similar, which should have been a 1-2 line change, and likely would have eliminated most of the excess costs incurred here from ever needing to have been incurred.
- `useUserInvalidateCache` in `useUser` queries the `users` collection in firestore to retrieve a certain user's document, and then updates the cache with that record
  - This appears to be used in `src/components/templates/Audience/Audience.tsx` / `src/components/templates/PartyMap/components/Map/Map.tsx` to handle the 'update user's seat position more quickly' issue

yarikoptic:
> In particular, I believe the way that #1630 is doing a full query every minute or so per user is loading vastly more data than the previous 'live update' method would have. The live update method will load the data the first time, but then only receive the tiny 'snapshot updates' each time. So if a particular user's record hasn't changed then no data will need to be sent down again, and it won't count as a 'read' against the firestore DB.

FWIW, from looking at the cost breakdown it looked like the most of it is not from amount (in size) of reads but the number of them. So, if users "don't move", or don't have multiple "competing" tabs open, there indeed should no longer be large number of reads.  But, since after the deploy of "this" revert they still might be running old codebase -- wouldn't they still query per each minute? is there a way to force them "to reload"?

0xdevalias:
> from looking at the cost breakdown it looked like the most of it is not from amount (in size) of reads but the number of them

@yarikoptic Correct. Each time @soichih's code runs each minute to update the users, it reads ever single user (eg. many reads), instead of the old way using the `.onSnapshot` listener which will only ever accumulate new reads if the user has actually changed since.

---

> But, since after the deploy of "this" revert they still might be running old codebase -- wouldn't they still query per each minute? is there a way to force them "to reload"?

@yarikoptic That's a great question.. and not one that I necessarily even have an answer for tbh..

If they were on `/v/:venueId` then it would be using `VenueLandingPage`, and we could configure `venue.config.redirectUrl` to try and force them to another page/site, which given it uses `window.location.hostname`, would probably be possible to at least get them off the site.

- https://github.com/sparkletown/sparkle/blob/staging/src/pages/VenueLandingPage/VenueLandingPage.tsx#L71-L76

But I expect anyone left on an idle tab will be in `/in/:venueId`, which renders `VenuePage`, which renders `TemplateWrapper`; which can be forced to redirect people when `venue.template` is one of `zoomroom`, `performancevenue`, or `artcar`; and `venue.zoomUrl` is configured:

- https://github.com/sparkletown/sparkle/blob/staging/src/pages/VenuePage/TemplateWrapper.tsx#L74-L93

That could be done in the firestore database, but since they could be on literally any venue.. 1000+ changes would need to be made. And that's assuming they're still receiving snapshot updates to even get any changes we make it the database 'pushed through' to them still.

We could do something hacky like changing the access rules to disallow the frontend from being able to read from the `users` collection at all. That should halt the reads.. but will also render the entire site broken/inaccessible, and we presumably wouldn't be able to just change the rule for a short time, as the 'update intervals' will still continue running for anyone with a tab still open, which means as soon as we changed the rule back, they would just keep trying to fetch the data.

We might be able to invalidate all of the current auth sessions, which may end up forcing users to re-login before they would be able to query again/etc. If that works.. then they would presumably (maybe) be forced to have a full page reload before they can do anything?

We could potentially delete the entire 'web app' config within firebase's project settings and re-create it, which would change the api key and similar used to access the site, which theoretically would prevent older code/clients from accessing it still, assuming it doesn't 'hold open' legacy connections/etc. This would require a decent chunk of effort to make the site work again though, as we would need to go and change all the config in our CI, re-deploy code changes to pick up the new configuration, etc.

PR #1692, opened 2021-06-29
[env/github] change default url to /v/register instead of sparkespaces.com

malaikaaiyar:
partly fixes https://github.com/sparkletown/internal-sparkle-issues/issues/868

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -18,7 +18,7 @@ import sparkleNavLogo from "assets/icons/sparkle-nav-logo.png";
 import defaultMapIcon from "assets/icons/default-map-icon.png";
 import sparkleverseLogo from "assets/images/sparkleverse-logo.png";
 
-export const SPARKLE_HOMEPAGE_URL = "https://sparklespaces.com/";
+export const SPARKLE_HOMEPAGE_URL = "/v/register/";
 export const SPARKLE_TERMS_AND_CONDITIONS_URL =
   "https://sparklespaces.com/terms-of-use/";
 export const SPARKLE_PRIVACY_POLICY =


src/settings.ts
```diff
@@ -18,7 +18,7 @@ import sparkleNavLogo from "assets/icons/sparkle-nav-logo.png";
 import defaultMapIcon from "assets/icons/default-map-icon.png";
 import sparkleverseLogo from "assets/images/sparkleverse-logo.png";
 
-export const SPARKLE_HOMEPAGE_URL = "https://sparklespaces.com/";
+export const SPARKLE_HOMEPAGE_URL = "/v/register/";
```
denisdimitrov, line 21: This will actually redirect to a venue page and consider `register` as a venue. Which will fail if it doesn't exist in the ohbm database. 
https://ohbm.sparkle.space/v/register/

There is also being used only in a single place:
```tsx
export const HOMEPAGE_URL = IS_BURN
  ? SPARKLEVERSE_HOMEPAGE_URL
  : SPARKLE_HOMEPAGE_URL;
```
I don't really remember what this `IS_BURN` flag is but make sure you are changing the correct `const`.

denisdimitrov (changes requested):

src/settings.ts
denisdimitrov: You can feel free to re-request review if the venue `register` already exists and everything is fine with the redirect.

malaikaaiyar: `register` does exist as a venue. This is the change that github asked for.

jonboutelle (approved):
Approved

whoaitsaimz:
Can we also create a ticket or note that on Tuesday July 12th at 7am PST it would be super helpful to have this point to the main map `github-summit.com/in/explore`   rather than register page?

0xdevalias:
@whoaitsaimz Make a backlog issue about this on our internal issue board. This is a closed public PR, definitely not the place for tracking 'future' needs.

PR #1698, opened 2021-06-29
[env/github] remove profile links from profile

malaikaaiyar:
no issue created for this - removes profile links from the profile modal


jonboutelle (approved):
This seems safe only because it's mostly deleting code. Not crazy about approving but gonna do it. @0xdevalias can roast me in a couple hrs when he wakes up.

jonboutelle (approved):
This seems safe only because it's mostly deleting code. Not crazy about approving but gonna do it. @0xdevalias can roast me in a couple hrs when he wakes up.

PR #1693, opened 2021-06-29
[env/github] add purple border around octocat login button

malaikaaiyar:
partly fixes https://github.com/sparkletown/internal-sparkle-issues/issues/868

Screen Shot 2021-06-29 at 10 57 02 AM

PR #1700, opened 2021-06-29
[env/github] update registration page again

malaikaaiyar:
based on @whoaitsaimz's comments in https://github.com/sparkletown/sparkle/pull/1693 (https://github.com/sparkletown/sparkle/pull/1693#issuecomment-870869218 + https://github.com/sparkletown/sparkle/pull/1693#issuecomment-870871005)

partly fixes sparkletown/internal-sparkle-issues#868

Screen Shot 2021-06-29 at 10 26 36 PM
Screen Shot 2021-06-29 at 10 26 44 PM

### DIFF
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -1,25 +1,15 @@
-import React, {
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { useFirebase } from "react-redux-firebase";
-import { useAsync } from "react-use";
 import { UserInfo } from "firebase/app";
 import { FirebaseStorage } from "@firebase/storage-types";
+import { externalUrlAdditionalProps } from "utils/url";
 
 import {
   ACCEPTED_IMAGE_TYPES,
-  DEFAULT_AVATARS,
   MAX_AVATAR_IMAGE_FILE_SIZE_BYTES,
 } from "settings";
 
 import { resizeFile } from "utils/image";
-import { isDefined } from "utils/types";
-
-import { useSovereignVenueId } from "hooks/useSovereignVenueId";
 
 import "./ProfilePictureInput.scss";
 
@@ -50,24 +40,6 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
   const [error, setError] = useState("");
   const firebase = useFirebase();
   const uploadRef = useRef<HTMLInputElement>(null);
-
-  const { sovereignVenueId, isSovereignVenueIdLoading } = useSovereignVenueId({
-    venueId,
-  });
-
-  const {
-    value: customAvatars,
-    loading: isLoadingCustomAvatars,
-  } = useAsync(async () => {
-    if (!sovereignVenueId) return;
-
-    const storageRef = firebase.storage().ref();
-    const list = await storageRef
-      .child(`/assets/avatars/${sovereignVenueId}`)
-      .listAll();
-
-    return Promise.all(list.items.map((item) => item.getDownloadURL()));
-  }, [firebase, sovereignVenueId]);
 
   const uploadPicture = async (profilePictureRef: Reference, file: File) => {
     setIsPictureUploading(true);
@@ -111,41 +83,6 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
     setValue("pictureUrl", githubImageSrc, true);
   }, [githubImageSrc, setValue, pictureUrl]);
 
-  const uploadDefaultAvatar = useCallback(
-    async (avatar: string) => {
-      setValue("pictureUrl", avatar, true);
-    },
-    [setValue]
-  );
-
-  const isLoading =
-    (isSovereignVenueIdLoading || isLoadingCustomAvatars) &&
-    (customAvatars !== undefined || error !== undefined);
-
-  const defaultAvatars = customAvatars?.length
-    ? customAvatars
-    : DEFAULT_AVATARS;
-
-  const avatarImages = useMemo(
-    () =>
-      [githubImageSrc, ...defaultAvatars]
-        .filter(isDefined)
-        .map((avatarSrc, index) => (
-          <div
-            key={`${avatarSrc}-${index}`}
-            className="profile-picture-preview-container"
-            onClick={() => uploadDefaultAvatar(avatarSrc)}
-          >
-            <img
-              src={avatarSrc}
-              className="profile-icon profile-picture-preview"
-              alt={`default avatar ${index}`}
-            />
-          </div>
-        )),
-    [defaultAvatars, githubImageSrc, uploadDefaultAvatar]
-  );
-
   return (
     <div className="profile-picture-upload-form">
       <div
@@ -153,34 +90,39 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
         onClick={() => uploadRef.current?.click()}
       >
         <img
-          src={pictureUrl || "/default-profile-pic.png"}
+          src={pictureUrl || "/avatars/default-octocat-3.png"}
           className="profile-icon profile-picture-preview"
           alt="your profile"
         />
       </div>
-      <input
-        type="file"
-        id="profile-picture-input"
-        name="profilePicture"
-        onChange={handleFileChange}
-        accept={ACCEPTED_IMAGE_TYPES}
-        className="profile-picture-input"
-        ref={uploadRef}
-      />
-      <label htmlFor="profile-picture-input" className="profile-picture-button">
-        Upload your profile pic
-      </label>
+      <p className="profile-picture-options">
+        Want to change your profile photo? Take a{" "}
+        <a
+          className="button--a"
+          href="https://virtual.githubphotobooth.com/virtual/capture/gr99n"
+          {...externalUrlAdditionalProps}
+        >
+          Summit snap
+        </a>{" "}
+        or{" "}
+        <input
+          type="file"
+          id="profile-picture-input"
+          name="profilePicture"
+          onChange={handleFileChange}
+          accept={ACCEPTED_IMAGE_TYPES}
+          className="profile-picture-input"
+          ref={uploadRef}
+        />
+        <label htmlFor="profile-picture-input" className="button--a">
+          upload a photo.
+        </label>
+      </p>
       {errors.pictureUrl && errors.pictureUrl.type === "required" && (
         <span className="input-error">Profile picture is required</span>
       )}
       {isPictureUploading && <small>Picture uploading...</small>}
       {error && <small>Error uploading: {error}</small>}
-      <small>
-        Feeling camera shy? Choose from one of these Summit profile avatars.
-      </small>
-      <div className="default-avatars-container">
-        {isLoading ? <div>Loading...</div> : avatarImages}
-      </div>
       <input
         name="pictureUrl"
         className="profile-picture-input"
--- a/src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
+++ b/src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
@@ -19,13 +19,4 @@
   .login-text {
     text-align: center;
   }
-
-  .login-button {
-    color: white !important;
-    text-decoration: underline;
-
-    &:hover {
-      cursor: pointer;
-    }
-  }
 }
--- a/src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
+++ b/src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
@@ -16,13 +16,16 @@
     text-align: center;
   }
 
+  .login-text {
+    text-align: center;
+  }
+
   .login-button {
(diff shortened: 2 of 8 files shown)


src/components/organisms/AuthenticationModal/InitialForm/InitialForm.tsx
```diff
@@ -21,12 +21,12 @@ export const InitialForm: FC<InitialFormProps> = ({
...
-        className="btn btn-block btn-centered login-button"
-        onClick={displayLoginForm}
-      >
-        Log in for non-Okta users
-      </div>
+      <p className="login-text">
+        No Okta?{" "}
+        <span className="login-button" onClick={displayLoginForm}>
+          Login here
+        </span>
```
0xdevalias, line 28: Pretty sure we have explicit styling for `button` that allows displaying as a text link for this purpose. Using `span` here isn't really valid/semantic/accessible.

src/pages/Account/Login.scss
```diff
@@ -6,9 +6,17 @@
     flex-flow: column;
     align-items: center;
     margin-bottom: $spacing--md;
-    padding: $spacing--lg 0;
+    padding: $spacing--lg $spacing--lg;
     box-shadow: inset 0 0 0 3px #6f43ff;
     border-radius: 0.25rem;
+    font-size: 19px;
+    transition: all 400ms $transition-function;
+    font-weight: 500;
```
0xdevalias, line 14: We have variables for `font-size` / `font-weight`, etc

src/pages/Account/Login.scss
```diff
@@ -6,9 +6,17 @@
...
+
+    &:hover {
+      background-color: darken($color: $primary, $amount: 20%);
+      cursor: pointer;
+    }
```
0xdevalias, line 19: Why are we needing to add `cursor: pointer;` here?

src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
```diff
@@ -16,13 +16,16 @@
...
   .login-button {
-    padding: $spacing--lg $spacing--xl;
-    box-shadow: inset 0 0 0 3px $primary !important;
     color: white !important;
+    text-decoration: underline;
 
     &:hover {
-      background-color: darken($color: $primary, $amount: 20%);
+      cursor: pointer;
     }
```
0xdevalias, line 29: Why are we needing to add `cursor: pointer;` here?

src/pages/Account/Login.scss
```diff
@@ -18,13 +26,6 @@
     margin-bottom: $spacing--xs;
   }
 
-  &__quick-login-icon {
-    &:hover {
-      cursor: pointer;
-      opacity: 0.8;
-    }
-  }
```
0xdevalias, line 26: Why were these styles removed from here? Should just be commented out + an `@debt` comment about this being an environment specific hack at most. Though would likely be find to just leave them in place presumably?

src/pages/Account/Login.tsx
```diff
@@ -92,30 +92,46 @@ export const Login: React.FC<LoginProps> = ({
...
+                  <FontAwesomeIcon
+                    className="Login__quick-login-icon"
+                    icon={faGithub}
+                    size="4x"
+                    title="SAML SSO login"
+                  />
+                </div>
+              </div>
+            )}
+          </>
```
0xdevalias, line 134: I'm not sure what the intent of this change is.. but looking at the diff here.. seems.. wrong?

0xdevalias: The `hasCustomAuthConnect`section  has nothing to do with Okta/SAML login.. so none of those changes should have been made here.

0xdevalias: Not sure why the 'quick login with' was moved/duplicated either. This should presumably remain in the `hasAlternativeLogins` block.

0xdevalias (changes requested):
See comments/questions/changes/etc.

src/pages/Account/Login.tsx
whoaitsaimz: The onClick handler was not applied to the whole containing div previously, so effectively only the icon itself was clickable

whoaitsaimz: Just a thought -- and this should not be a blocker -- but if we have variables like {customAuthName}  then we shouldn't have to hard-code the word _Okta_ in here, right? That would make this code more usable for other login flows...

whoaitsaimz:
I checked on https://co-reality-staging--preview-pr-1700-jctypaza.web.app/in/bootstrap  but we don't have Okta turned on I guess. I was able to verify that the standard email login works but don't have visibility of the Okta button in this preview

src/pages/Account/Login.scss
malaikaaiyar: So that it looks like the whole div is a button

src/components/organisms/AuthenticationModal/InitialForm/InitialForm.scss
malaikaaiyar: So that it's obvious that you can click on the login text- otherwise it turns into the text cursor and it looks unclickable

src/pages/Account/Login.tsx
malaikaaiyar: The intent of this change was to make the whole purple box div clickable rather than just the octocat icon. Since there's a different onclick function depending on the conditions, I had to duplicate the whole div with the "quick login with" etc.

src/pages/Account/Login.scss
malaikaaiyar: They just weren't needed anymore. The point of this PR was to fix some styles and make the whole okta box clickable rather than the icon, so having styles for the quick login icon's hover is not right. I can comment it out and write a comment though.

src/pages/Account/Login.tsx
malaikaaiyar: For some reason the diff looks weird? don't know what's going on there, but I basically deleted the existing div and replaced it with the two conditional statements

malaikaaiyar:
I've fixed all the issues except the last comment which needs discussion

malaikaaiyar:
I've just added changes from @whoaitsaimz's comments in https://github.com/sparkletown/internal-sparkle-issues/issues/821

src/pages/Account/Login.scss
```diff
@@ -2,29 +2,39 @@
...
+    transition: all 400ms $transition-function;
+    font-weight: $font-weight--500;
+
+    &:hover {
+      background-color: darken($color: $primary, $amount: 20%);
+      cursor: pointer;
+    }
   }
 
+  // @debt removed because rather than the icon being a button, the whole login box is supposed to be a button
```
alexey-petrov, line 23: should we remove it instead of commenting out?

malaikaaiyar: @0xdevalias asked me to comment it out instead

src/pages/Account/Login.tsx
malaikaaiyar: @whoaitsaimz and I have decided to remove this change for now and move on with the other things

malaikaaiyar:
I deleted the large changes that needed discussion/re-review by @0xdevalias (making the entire okta box clickable rather than just the icon)

jonboutelle (approved):
LGTM and we outta time. Lets gooooo!

src/pages/Account/Login.scss
0xdevalias: If this was the main `staging`/etc code branch + this was a permanent change then removing would absolutely make sense. But given this is environment specific hacks, I'd rather us comment it out + leave comments with reasoning why so that when we try and diff the environment against the main codebase again, it will be really obvious what was just a hack vs what might be useful to bring back across.

src/pages/Account/Login.tsx
0xdevalias: > if we have variables like {customAuthName} then we shouldn't have to hard-code the word Okta in here, right? That would make this code more usable for other login flows...

@whoaitsaimz That's the plan, but the okta SSO stuff doesn't actually use the `customAuth` stuff yet (which was my more generic solution written for OHBM's SSO, but GitHub's SSO wasn't 'rolled into it')

0xdevalias:
> I've just added changes from @whoaitsaimz's comments in [sparkletown/internal-sparkle-issues#821](https://github.com/sparkletown/internal-sparkle-issues/issues/821)

@malaikaaiyar Please link to the specific comments, not just the issue as a whole. It's ambiguous as to which comments you're referring to here.

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -51,24 +41,6 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
-    if (!sovereignVenueId) return;
-
-    const storageRef = firebase.storage().ref();
-    const list = await storageRef
-      .child(`/assets/avatars/${sovereignVenueId}`)
-      .listAll();
-
-    return Promise.all(list.items.map((item) => item.getDownloadURL()));
-  }, [firebase, sovereignVenueId]);
-
```
0xdevalias, line 71: Why was all of this removed?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -111,76 +83,46 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
-          >
-            <img
-              src={avatarSrc}
-              className="profile-icon profile-picture-preview"
-              alt={`default avatar ${index}`}
-            />
-          </div>
-        )),
-    [defaultAvatars, githubImageSrc, uploadDefaultAvatar]
-  );
```
0xdevalias, line 147: Why was all of this removed?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -111,76 +83,46 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
         <span className="input-error">Profile picture is required</span>
       )}
       {isPictureUploading && <small>Picture uploading...</small>}
       {error && <small>Error uploading: {error}</small>}
-      <small>
-        Feeling camera shy? Choose from one of these Summit profile avatars.
-      </small>
-      <div className="default-avatars-container">
-        {isLoading ? <div>Loading...</div> : avatarImages}
-      </div>
```
0xdevalias, line 183: Why was all of this removed?

src/pages/Account/Profile.tsx
```diff
@@ -115,11 +114,11 @@ const Profile: React.FunctionComponent<PropsType> = ({ location }) => {
...
     <div className="Profile">
       <h2 className="login-welcome-title">
-        Hey, {githubHandle}. We’re so glad you’re here! Upload or take a photo
-        and share your Summit snap here.
+        Hey, {githubHandle}. We’re so glad you’re here!
       </h2>
       <div className="login-welcome-subtitle">
-        {`Swing back and edit your profile anytime you like.`}
+        We&apos;ve pre-populated your profile with details from Okta - please{" "}
+        <br /> review and click Next if it all looks good.
```
0xdevalias, line 121: `<br />` Shouldn't be used here like this.. If you want the text to 'flow' to a new line, then the 'container' it is in shouldn't be as wide.

0xdevalias (commented):
Post-merge review comments

whoaitsaimz:
https://github.com/sparkletown/sparkle/pull/1700#discussion_r661893880 large chunks were removed because for this environment, goal was to de-emphasize the alternate profile photo options-- since we were able to pull default photos by handle, they wanted these to be prominent and the buttons were converted to text links instead.  This is an environment-specific patch only.

whoaitsaimz:
https://github.com/sparkletown/sparkle/pull/1700#discussion_r661894561 100% and this page layout needs some CSS work in general.

0xdevalias:
@whoaitsaimz You know you can reply inline in those comment threads right..?

whoaitsaimz:
Nope, I think because it's already been merged, I don't see that option

image

0xdevalias:
@whoaitsaimz Thats.. strange. As they’re still accessible for me. 
3A5B1ADD-5A51-43D1-A75D-AF27966AA942

whoaitsaimz:
🤷🏻‍♀️ Lol

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
malaikaaiyar: we had to get rid of the custom avatars

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
malaikaaiyar: we had to get rid of the custom avatars

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
malaikaaiyar: same as above (we had to get rid of the default avatars)

src/pages/Account/Profile.tsx
malaikaaiyar: @mobilevinay suggested this so I went with it. I could make a new PR for this, but I think the registration page is already done being used (as in everyone has already registered for the event so they won't go to this page, but I might be wrong)

PR #1643, opened 2021-06-24
Allow camera and microphone for embeddable (v2)

whoaitsaimz:
closes sparkletown/internal-sparkle-issues#662

Trying again with proper base and fresh commits, will close previous PR that had conflicts

PR #1701, opened 2021-06-29
fix IFRAME_ALLOW by adding semicolons after both statements

whoaitsaimz:
Following up on https://github.com/sparkletown/sparkle/pull/1643

Closes https://github.com/sparkletown/internal-sparkle-issues/issues/662

Unsure why a backtick is used instead of quotation marks -- (for the variable thing you added, Glenn?) Just checking

image

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -156,8 +156,8 @@ export const ACCEPTED_IMAGE_TYPES =
 export const VALID_URL_PROTOCOLS = ["http:", "https:"];
 
 export const IFRAME_ALLOW =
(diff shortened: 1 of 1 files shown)


0xdevalias:
> Unsure why a backtick is used instead of quotation marks -- (for the variable thing you added, Glenn?) Just checking

@whoaitsaimz The backticks are used to provide variable substitution within the text using the ${} syntax I used there. Without them, it would just show that as a literal string.

0xdevalias (approved):
Looks good, LGTM!

whoaitsaimz:
It worked! Thank you for the support on this one -- winning!!

image

PR #1703, opened 2021-07-01
refactor lastSeenAt / lastSeenIn out of main User document + associated refactors

0xdevalias:
Bringing back the changes from https://github.com/sparkletown/sparkle/pull/1623 since it was re-targeted into `env/ohbm` rather than the main codebase.

See also the review comments raised in https://github.com/sparkletown/sparkle/pull/1623#pullrequestreview-690375645

---

**See also:**

- https://github.com/sparkletown/sparkle/pull/1708
- https://github.com/sparkletown/sparkle/pull/1714

---

## TODO

- [x] Merge `staging` into this branch + fix any merge conflicts.
- [x] We need the updated version of `react-redux` to properly fix memoisation in `useSelector` from this PR:
  - https://github.com/sparkletown/sparkle/pull/1714

### Potentially OOS

- [ ] Implement manual redux patterns for user fetching instead of using `useFirestoreConnect` (**note:** We could likely do this in a new followup PR to land the fixes already included here sooner..)
  - [ ] Implement redux boilerplate/patterns using Redux Toolkit to store the user data 'broken down' into it's appropriate 'chunks' for more efficient access/updates
  - [ ] Fetch user data by using the firestore SDK's `.onSnapshot` listener directly
- Should we also refactor `enteredVenueIds` out of the main `User` type like we did with `lastSeenIn` / `lastSeenAt`?
- Should we look at refactoring how the `data` object on the user is used (and maybe splitting it off the main user object, at least on the frontend)?
- Should we look at refactoring how the `myPersonalizedSchedule` object on the user is used (and maybe splitting it off the main user object, at least on the frontend)? I feel like that may belong in `usersPrivate` more than it does in `users`?

## Previous Context

> Opening this draft PR super early to track progress.
> 
> Aim is to refactor `lastSeenIn` + `lastSeenAt` off the main `User` object/types to prevent it breaking memoisation across the platform when updated.
> 
> This is suspect to be one of the core contributors to performance issues, particularly as the number of online users increases.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1623_

---

> Profiling, `UserAvatar` (Within `VenueChat`) seems to be ranking as the slowest/most often needing to run/etc
> 
> <details><summary>Screenshots</summary>
> 
> image
> image
> 
> </details>
> 
> It seems like commenting out `useRecentWorldUsers` + `isOnline` removes this issue:
> 
> <details><summary>Diff + new screenshots</summary>
> 
> ```diff
> diff --git a/src/components/atoms/UserAvatar/UserAvatar.tsx b/src/components/atoms/UserAvatar/UserAvatar.tsx
> index 42889c54..6f235a94 100644
> --- a/src/components/atoms/UserAvatar/UserAvatar.tsx
> +++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
> @@ -1,10 +1,11 @@
> -import React, { useMemo } from "react";
> +// import React, { useMemo } from "react";
> +import React from "react";
>  import classNames from "classnames";
>  
>  import { DEFAULT_PARTY_NAME, DEFAULT_PROFILE_IMAGE } from "settings";
>  
>  import { User, UsernameVisibility } from "types/User";
> -import { useRecentWorldUsers } from "hooks/users";
> +// import { useRecentWorldUsers } from "hooks/users";
>  
>  import { WithId } from "utils/id";
>  
> @@ -30,7 +31,7 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
>    showStatus,
>    large,
>  }) => {
> -  const { recentWorldUsers } = useRecentWorldUsers();
> +  // const { recentWorldUsers } = useRecentWorldUsers();
>  
>    const avatarSrc: string = user?.anonMode
>      ? DEFAULT_PROFILE_IMAGE
> @@ -45,10 +46,11 @@ export const UserAvatar: React.FC<UserAvatarProps> = ({
>      "UserAvatar--large": large,
>    });
>  
> -  const isOnline = useMemo(
> -    () => recentWorldUsers.find((worldUser) => worldUser.id === user?.id),
> -    [user, recentWorldUsers]
> -  );
> +  // const isOnline = useMemo(
> +  //   () => recentWorldUsers.find((worldUser) => worldUser.id === user?.id),
> +  //   [user, recentWorldUsers]
> +  // );
> +  const isOnline = true;
>  
>    const status = user?.status;
> ```
> 
> image
> 
> image
> 
> </details>
> 
> Based on that profiler's result, next areas to look at from what I see in that profiling would be:
> 
> - `src/components/molecules/VenuePartygoers/VenuePartygoers.tsx`
>   - calls `useRecentWorldUsers` (`src/hooks/users.ts`)
>   - only seems to care about the `.length` though.. so could likely use a more specific selector to save it being called too often
> - `src/components/templates/PartyMap/components/RoomAttendance.tsx`
>   - calls `src/hooks/useRoom.ts`
>   - which calls `useRecentLocationUsers` (`src/hooks/users.ts`)
>   - it seems to only need to slice off up to `maxVisible` users, which defaults to `2` via `DEFAULT_ROOM_ATTENDANCE_LIMIT`
> - `src/components/templates/PartyMap/components/Map/MapRoom.tsx`
>   - calls `src/hooks/useRoom.ts`
>   - which calls `useRecentLocationUsers` (`src/hooks/users.ts`)
>   - this seems to only be used to make `hasRecentRoomUsers`, and then that is used to help decide whether the `maproom--always-show-label` class should be applied, which basically decides whether the block containing `RoomAttendance` should be displayed via CSS (it's only conditionally rendered based on `isCovertRoom`)
> - etc
> 
> My browser tab still seems to be doing a LOT of work though.. so not sure if some of the changes have just moved the problem to somewhere this particular profiler can't 'see it'.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1623#issuecomment-866301427_

---

> I can't tell if this has helped at all.. figured I would capture it here in case. My react profiling looks slightly different, but that could just be due to differing load now, not necessarily this change:
> 
> ```diff
> diff --git a/src/index.tsx b/src/index.tsx
> index 4dac6bf2..7ce8b2dc 100644
> --- a/src/index.tsx
> +++ b/src/index.tsx
> @@ -125,8 +125,8 @@ const store = createStore(
>    initialState,
>    composeWithDevTools(
>      applyMiddleware(
> -      thunkMiddleware,
> -      LogRocket.reduxMiddleware() // logrocket needs to be last
> +      thunkMiddleware
> +      // LogRocket.reduxMiddleware() // logrocket needs to be last
>      )
>    )
>  );
> ```
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1623#issuecomment-866316974_

---

> In my latest React Profiler run i'm seeing a LOT of entries related to `EventModal` (being rendered from `ScheduleEvent`). It looks like it might be rendered for every single entry on the schedule. Not sure that that is ideal. We should just have a single modal that is loaded with the appropriate data if and when an item is clicked on. That's kind of OOS of this specific PR.. but figured I would point it out.
> 
> <details><summary>Screenshots</summary>
> 
> image
> 
> image
> 
> </details>
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1623#issuecomment-866319231_


denisdimitrov:
@0xdevalias This PR is still a draft but it's also in the list of `Review required` PRs in the epic. Should I review this or wait a bit?

0xdevalias:
> @0xdevalias This PR is still a draft but it's also in the list of `Review required` PRs in the epic. Should I review this or wait a bit?

@denisdimitrov Ready to be bulk reviewed now, may need a little polish still, but that's not a blocker. @sunny-viktoryia was going to review it, so check with her to make sure you don't double up.

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -44,169 +42,189 @@ const firestoreUpdate = (doc: string, update: any) => {
...
+    TableComponent,
+    isSeatedAtTable,
+    onJoinClicked,
+    recentVenueUsers,
+    tableLocked,
+    tables,
+    venueName,
+  ]);
+
+  if (!isRecentVenueUsersLoaded) return <>Loading...</>;
```
denisdimitrov, line 191: We should probably use `<LoadingPage/>` or `<Loading/>` here.

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -44,169 +42,189 @@ const firestoreUpdate = (doc: string, update: any) => {
...
+  const renderedTables = useMemo(() => {
+    if (isSeatedAtTable) return;
+
+    return tables.map((table: Table, index: number) => (
+      <TableComponent
+        key={table.reference}
+        experienceName={venueName}
+        users={recentVenueUsers}
+        table={table}
+        tableLocked={tableLocked}
```
denisdimitrov, line 175: [NB] This name is miss-leading, it indicates that this is a boolean but it's a function instead. This whole component and it's probably probably need to be refactored. Maybe add tech debt for it.

src/components/templates/Playa/Playa.tsx
```diff
@@ -600,11 +588,13 @@ const Playa = () => {
           backgroundImage={venue?.mapBackgroundImageUrl}
         />
         {venues?.filter(isPlaced).map((v, idx) => {
-          const usersInVenue = recentVenueUsers.filter(
-            (partygoer) =>
-              partygoer.lastSeenIn?.[v.name] >
-              (nowMs - LOC_UPDATE_FREQ_MS * 2) / 1000
-          );
+          // @debt This isn't strictly correct here.. but this is an unused legacy template soon to be deleted, so we don't mind
+          const usersInVenue = recentVenueUsers;
```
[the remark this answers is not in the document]
denisdimitrov: Ignore my comment, just noticed that this is a legacy component.

src/types/User.ts
```diff
@@ -49,3 +67,83 @@ export enum UsernameVisibility {
...
+
+// export const userWithLocationToUser = (
+//   user: WithId<UserWithLocation>
+// ): WithId<User> =>
+//   withId(UserSchema.validateSync(user, { stripUnknown: true }), user.id);
+
+// @debt Not sure if the validations are too 'heavyweight' for this, but object destructuring seemed to work
+//  here, whereas the validations seemed to hang my browser tab. There might also be something wrong with the
+//  validation rules leading to infinite recursion or similar?
+export const userWithLocationToUser = (
```
denisdimitrov, line 143: Is this the correct file for this helper function? Also it seems to destructure the user and return `userWithoutLocation` but the name is `userWithLocationToUser`

src/utils/selectors.ts
```diff
@@ -1,4 +1,5 @@
 import { FirebaseReducer } from "react-redux-firebase";
+import { mapValues } from "lodash";
```
denisdimitrov, line 2: [OOS] I suggest removing `lodash` as dependency and adding only the required functions as modules. The library is 1.4mb large according to the npmjs https://www.npmjs.com/package/lodash

Adding only the functions that we need will be much less IMO, we need 4 or 5 functions only.
https://www.npmjs.com/package/lodash.chunk
https://www.npmjs.com/package/lodash.isequal
https://www.npmjs.com/package/lodash.throttle
etc...

src/utils/selectors.ts
```diff
@@ -7,13 +8,13 @@ import { Experience } from "types/Firestore";
...
 import { SparkleSelector } from "types/SparkleSelector";
-import { User } from "types/User";
+import { User, UserWithLocation, userWithLocationToUser } from "types/User";
 import { AnyVenue, PosterPageVenue, VenueEvent } from "types/venues";
 import { ScreeningRoomVideo } from "types/screeningRoom";
 
 import { SovereignVenueState } from "store/reducers/SovereignVenue";
 
-import { WithId } from "utils/id";
+import { withId, WithId } from "utils/id";
```
denisdimitrov, line 17: The name of this function should probably start with a verb to make it more obvious that this adds an Id to the passed object.

src/types/User.ts
```diff
@@ -49,3 +67,83 @@ export enum UsernameVisibility {
...
+    mirrorVideo: Yup.boolean(),
+    status: Yup.string(),
+    data: UserExperienceDataSchema,
+    myPersonalizedSchedule: MyPersonalizedScheduleSchema,
+    enteredVenueIds: Yup.array().of(Yup.string().required()),
+
+    // Legacy?
+    room: Yup.string(),
+    video: VideoStateSchema.notRequired(),
+    kidsMode: Yup.boolean(),
```
denisdimitrov, line 130: These are legacy AF. Can you please add tech debt?

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -145,19 +146,21 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
...
-      </ReactionsProvider>
-    </RelatedVenuesProvider>
+    <WorldUsersProvider venueId={venue.id}>
+      <RelatedVenuesProvider venueId={venue.id}>
+        <ReactionsProvider venueId={venue.id}>
+          <WithNavigationBar
+            fullscreen={fullscreen}
+            hasBackButton={hasBackButton}
+          >
+            <AnnouncementMessage message={venue.bannerMessage} />
```
denisdimitrov, line 156: [NB] Context is cool and declarative but this is starting to look like 'wrapper hell', and that's exactly one of the benefits of react hooks, they can avoid this. Just noting. No change request required.

denisdimitrov (dismissed):
Mostly small things, some improvements and comments, one or two questions as well.

src/components/templates/Playa/Playa.tsx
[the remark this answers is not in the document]
0xdevalias: It wasn't intended to make anything immutable, it pretty much solely was being used to avoid refactoring/renaming any of the latter things, since I ripped out the `const usersInVenue = recentVenueUsers.filter`

src/hooks/users.ts
```diff
@@ -1,75 +1,53 @@
...
+import { withId, WithId } from "utils/id";
 import { normalizeTimestampToMilliseconds } from "utils/time";
 
-import { worldUsersByIdSelector, worldUsersSelector } from "utils/selectors";
+import {
+  worldUsersByIdSelector,
+  worldUsersByIdWithoutLocationSelector,
+  worldUsersSelector,
+} from "utils/selectors";
+import { useWorldUsersContext } from "./users/useWorldUsers";
```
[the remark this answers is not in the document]
0xdevalias: Not sure that is true. But the import path should probably read `hooks/users/useWorldUsers` rather than using the `./` syntax.

src/hooks/users.ts
```diff
@@ -1,75 +1,53 @@
...
-export const useWorldUsers = (): {
-  worldUsers: readonly WithId<User>[];
-  isWorldUsersLoaded: boolean;
-} => {
-  useConnectWorldUsers();
+export { useWorldUsers } from "./users/useWorldUsers";
 
-  const selectedWorldUsers = useSelector(worldUsersSelector);
+const noUsers: WithId<User>[] = [];
+const noUsersById: Record<string, WithId<User>> = {};
```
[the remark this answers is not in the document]
0xdevalias: > Can we start adding these in a more global/reusable file instead?

If we did that, then they would have the same reference for potentially unrelated things. Not sure if that would matter in reality or not.. but it might. Aside from that edge case, I don't have any strong feels against moving them to a common location.

> How about we use closures instead, that way it will be initialized only when the function is called.

What do you mean by this?

src/hooks/users/useWorldUsers.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  // @debt we use lodash's isEqual here to do a deep compare to prevent re-renders, but ideally we wouldn't need to
+  const selectedWorldUsers = useSelector(
+    worldUsersWithoutLocationSelector,
+    isEqual
+  );
+
+  return useMemo(
+    () => ({
+      worldUsers: selectedWorldUsers ?? [],
```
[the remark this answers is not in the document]
0xdevalias: If we keep the `useMemo` here then it's not really needed to extract it into another variable. If we get rid of the `useMemo` then we would need to follow the same pattern to avoid creating a new array every time the hook is called.

I don't think there is any benefit to the `useMemo` here since it would only be of use if we were passing on the entire return object without de-structuring, which we practically never do. `selectedWorldUsers` will already be memo'd via `useSelector`.

So yeah, seems like a decent idea.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -63,6 +64,8 @@ const VenuePage: React.FC = () => {
   // const [isAccessDenied, setIsAccessDenied] = useState(false);
 
   const { user, profile } = useUser();
+  const { userLocation } = useWorldUserLocation(user?.uid);
+  const { lastSeenIn: userLastSeenIn } = userLocation ?? {};
```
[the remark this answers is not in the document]
0xdevalias: I believe I explicitly chose not to do that, as I want `userLocation` to be `undefined` if it doesn't exist, so as not to 'hide' that fact. This syntax here is just a shorthand method for destructuring.

I could also have written it like this, but I tend to find the pattern overly verbose in most cases (this is a bit of an exception since i'm also renaming the variable, usually it's even shorter still):

```typescript
// Method used here
const { lastSeenIn: userLastSeenIn } = userLocation ?? {};

// Non-destructuring version
 const userLastSeenIn = userLocation?.lastSeenIn;

// ----------

// More canonical example
const { lastSeenIn } = userLocation ?? {}; 

// The more verbose version of it that I really dislike. I basically never want to be typing the same variable name out twice when I could be destructuring it instead.
const lastSeenIn = userLocation?.lastSeenIn; 
```

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: Agreed. Would have to look closer at the specifics to decide which makes most sense, or which can provide the least disruptive feel for the user.

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: Yup, this is part of the bundle of legacy tech debt that needs cleaning up. I think this is related to jazz bar from memory? That whole template/related is... a mess.. to say the least.

src/types/User.ts
0xdevalias: That's what the `// Legacy?` was for, but i'll make them more explicitly `@debt` and probably even `@deprecated` on the main object.

src/types/User.ts
0xdevalias: I put it here in a similar way to how we have the tiny `withId` helper alongside the `WithId` type.

My thinking was that having it 'close' to the types made most sense (and I think originally it was also using the validations from this file, so felt even more like it made sense)

I can see arguments for why it should be in somewhere like `utils/` or similar as well.. i'm kind of split 50/50 on it tbh.

---

> Also it seems to destructure the user and return `userWithoutLocation` but the name is `userWithLocationToUser`

The naming is correct, if not a little lacking in verbosity. It takes a `userWithLocation` and converts it `To` a `User`; thus `userWithLocation` `To` `User`.

This PR refactors the base `User` type to remove the location details from it, thus only the `UserWithLocation` type has them.

src/types/User.ts
```diff
@@ -49,3 +67,83 @@ export enum UsernameVisibility {
...
+//   user: WithId<UserWithLocation>
+// ): WithId<User> =>
+//   withId(UserSchema.validateSync(user, { stripUnknown: true }), user.id);
+
+// @debt Not sure if the validations are too 'heavyweight' for this, but object destructuring seemed to work
+//  here, whereas the validations seemed to hang my browser tab. There might also be something wrong with the
+//  validation rules leading to infinite recursion or similar?
+export const userWithLocationToUser = (
+  user: WithId<UserWithLocation>
+): WithId<User> => {
```
0xdevalias, line 145: This function doesn't actually require anything from `WithId`, so it should probably be made to support the basic types, or the wrapping types, and pass through the result in the same way. That would make it more robust.

src/utils/selectors.ts
0xdevalias: Importing the individual npm packages like that is a legacy method that's only really useful for libraries/environments that aren't using bundling/tree shaking/etc.

---

@goran-peoski-work Put up https://github.com/sparkletown/sparkle/pull/1715 for replacing `lodash` with `lodash-es`, which is a more modern/compatible way of solving this (and with less manual effort required) when using bundlers/tree shaking/etc.

---

I've also been looking into a few further optimisations as well lately, since even if we don't include `lodash` ourselves, a good chunk of our dependencies do, so it will end up in our bundle regardless. A few notes/resources off the top of my head:

- https://github.com/gsoft-inc/craco
  - Needed for customising the babel/webpack/etc config's with ejecting from `create-react-app`
- https://github.com/lodash/babel-plugin-lodash
  - > Modular Lodash builds without the hassle.
  - > A simple transform to cherry-pick Lodash modules so you don’t have to. Combine with lodash-webpack-plugin for even smaller cherry-picked builds!
- https://github.com/lodash/lodash-webpack-plugin
  - > Create smaller Lodash builds by replacing feature sets of modules with noop, identity, or simpler alternatives. This plugin complements `babel-plugin-lodash` by shrinking its cherry-picked builds even further!
  - > DISCLAIMER: Using this plugin without enabling the proper feature sets may cause lodash functions to behave in unexpected ways. Methods may appear to work, however they might return incorrect results.
- https://github.com/GoogleChromeLabs/webpack-libs-optimizations
  - > Using a library in your webpack project? Use these tips to make your bundle smaller!
  - https://github.com/GoogleChromeLabs/webpack-libs-optimizations/issues/21#issuecomment-756622581
    - Concludes that with webpack 4, using `lodash` is actually better than `lodash-es` when combined with `babel-plugin-lodash` & `lodash-webpack-plugin`
  - https://github.com/GoogleChromeLabs/webpack-libs-optimizations/issues/25
    - Has further discussions/nuances on why that might not be the best idea/potential alternatives.
    - https://github.com/GoogleChromeLabs/webpack-libs-optimizations/issues/25#issuecomment-874370372
      - My thoughts/comments, including pointing out that `create-react-app` currently still uses webpack v4, which means the points made in https://github.com/GoogleChromeLabs/webpack-libs-optimizations/issues/21#issuecomment-756622581 about tree shaking performance likely still apply to us

0xdevalias: Also cross-posted the above in https://github.com/sparkletown/sparkle/pull/1715#issuecomment-874371412 for reference

src/utils/selectors.ts
0xdevalias: This function has existed for ages and has been used all over the codebase. It's named to match the type that it converts things to, and to 'flow' in a sentence when used (though would do so even better if it was a curried function).

eg. (somewhat contrived examples to show the usage, may not be strictly perfectly valid code)

```typescript
// Currently
const userWithId = withId(user, userId);

const users: User[] = /* etc */;
const usersWithIds = users.map(user => withId(user, userId))

// -----

// If it was written as a curried function
const userWithId = withId(userId)(user);

const users: User[] = /* etc */;
const usersWithIds = users.map(withId(userId))
```

---

Do you have a specific suggestion for a relevant verb to be used for it + arguments for it? It hasn't seemed to be an issue so far, and was introduced back in november apparently:

> commit db093d3b6305f082ccdea171c00149c6731a5756
> Author: Glenn 'devalias' Grant
> Date:   Sat Nov 21 22:25:29 2020 +1100
> 
>     add generic `withId()` helper

0xdevalias: Definitely OOS of this PR in any case.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: What's your suggestion for the alternative? This is the canonical way that React uses context providers. Hooks can't help/change this in any way that I am aware of. At the 'usage site' we use the hooks for React's context API.

src/hooks/users.ts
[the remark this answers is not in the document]
denisdimitrov: >>How about we use closures instead, that way it will be initialized only when the function is called.

>What do you mean by this?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures
https://www.w3schools.com/js/js_function_closures.asp

Maybe it's better to initialize the value only when it's needed (in the function itself), otherwise it will be always created globally, even if that function is not called and that value is not actually used or needed.

src/pages/VenuePage/TemplateWrapper.tsx
denisdimitrov: We have redux, why not use it instead of context? It has the same benefits, avoiding props drilling and global access to the store.

src/components/molecules/TablesUserList/TablesUserList.tsx
denisdimitrov: > I think this is related to jazz bar from memory? That whole template/related is... a mess.. to say the least.

Yes.

src/components/templates/Playa/AvatarPartygoers.tsx
```diff
@@ -15,7 +15,8 @@ const AvatarPartygoers: React.FC<PropsType> = ({
   state,
   useProfilePicture,
 }) => {
-  const { recentVenueUsers } = useRecentVenueUsers();
+  // @debt we don't have access to venueName here (but is legacy Playa template code to be removed so we don't mind), this is basically a noop
+  const { recentVenueUsers } = useRecentVenueUsers({ venueName: undefined });
```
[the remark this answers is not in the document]
0xdevalias: Mostly because I used the `ReactHook` helper type and wanted to write proper input/output types for it. And I think also because I wanted to make it super explicit that it was a `venueName`, not a `venueId`, and to have that obvious when it's refactored as well, as per this `@debt` comment:

image

src/hooks/users.ts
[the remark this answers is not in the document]
0xdevalias: I believe this got fixed in the process of the refactors I merged in from `staging` in  https://github.com/sparkletown/sparkle/pull/1703/commits/cf9b5dcc7042c86556f11a5004680e39f6ec7be7

src/hooks/users.ts
```diff
@@ -1,75 +1,53 @@
...
-import { WithId } from "utils/id";
+import { withId, WithId } from "utils/id";
 import { normalizeTimestampToMilliseconds } from "utils/time";
 
-import { worldUsersByIdSelector, worldUsersSelector } from "utils/selectors";
+import {
+  worldUsersByIdSelector,
+  worldUsersByIdWithoutLocationSelector,
+  worldUsersSelector,
+} from "utils/selectors";
```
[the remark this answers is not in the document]
0xdevalias: I believe this got fixed in the process of the refactors I merged in from `staging` in  https://github.com/sparkletown/sparkle/pull/1703/commits/cf9b5dcc7042c86556f11a5004680e39f6ec7be7

src/hooks/users.ts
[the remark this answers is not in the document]
0xdevalias: If this was for some super heavyweight thing I might agree with using a lazy initialisation pattern. But IMO in doing so we're likely going to create excess bloat and complicate what the compiler/browser may just optimise out for us anyway. I don't think it's worth spending the time thinking about this tbh.

src/hooks/users/useWorldUsers.tsx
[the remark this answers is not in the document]
0xdevalias: fixed in c957697a068fcfd596de060d38d9db000e5b8ec3

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
0xdevalias: Marking this as resolved based on the above.

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: It's rendered within a bunch of other JSX, so definitely `<Loading/>`

0xdevalias: fixed in 247bc9ac730dcc377ff5e95e5343deaa0f9a1d4e

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: OOS

src/types/User.ts
0xdevalias: @denisdimitrov IMO no changes are required here, so marking this as resolved.

src/utils/selectors.ts
0xdevalias: This is implemented in https://github.com/sparkletown/sparkle/pull/1735

src/utils/selectors.ts
0xdevalias: Marking this as resolved as no action required here IMO.

src/pages/VenuePage/TemplateWrapper.tsx
0xdevalias: @denisdimitrov What does context have to do with props drilling..? You know that redux is built on top of React's context API's right...?

There are things that make sense in Redux. There are other things that are way harder if not impossible to do with it. Some choices not to use it so far have been boilerplate / time / existing patterns related. I want to move more to redux as our core pattern in future, but that doesn't make context a bad choice.

I'm still failing to see what your actual concerns/issues with it are..?

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -44,169 +42,189 @@ const firestoreUpdate = (doc: string, update: any) => {
     });
 };
 
-const defaultTables = [...Array(TABLES)].map((_, i: number) => createTable(i));
+const defaultTables = [...Array(DEFAULT_TABLE_COUNT)].map((_, i: number) =>
```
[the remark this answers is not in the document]
0xdevalias: Might be a good idea, but it's OOS / NB for this PR, it's existing legacy code in a component that needs a total refactor anyway.

src/types/User.ts
0xdevalias: fixed in b3440d05433547c7516d6113f5446f2cc10a6b3f

0xdevalias:
@denisdimitrov @sunny-viktoryia @mike-lvov This should be ready for re-review. Would be good for people to do a little clickthrough testing on it to ensure it seems stable/doesn't break things as I haven't had a chance to tonight after merging on/fixing recent things. This is probably the highest priority performance related PR we have currently, so would be good to get it prioritised/merged ASAP once we're sure it's stable. I will then likely followup with a further enhancement to this to improve things a bit more tomorrow/when I get a chance to.

src/hooks/users/useRecentLocationUsers.ts
```diff
@@ -1,10 +1,17 @@
...
+import { WithId } from "utils/id";
+import { worldUsersSelector } from "utils/selectors";
 import { normalizeTimestampToMilliseconds } from "utils/time";
 
+import { useSelector } from "hooks/useSelector";
+import { useConnectWorldUsers } from "hooks/users";
 import { useUserLastSeenThreshold } from "hooks/useUserLastSeenThreshold";
 
-import { useWorldUsers } from "./useWorldUsers";
+const noUsers: WithId<User>[] = [];
```
mike-lvov, line 14:
```suggestion
const noUsers: never[] = [];
```
?

src/components/molecules/TablesUserList/index.ts
```diff
@@ -1 +1,6 @@
+export { TablesUserList } from "./TablesUserList";
+
+/**
+ * @deprecated use named export instead
+ */
 export { default } from "./TablesUserList";
```
mike-lvov, line 6: It's only used in 3 files, maybe it worth changing it there?
image

src/hooks/users/useConnectWorldUsers.ts
```diff
@@ -1,29 +1,14 @@
...
-  });
+export const useConnectWorldUsers = () => {
+  // We mostly use this here to ensure that the WorldUsersProvider has definitely been connected
+  useWorldUsersContext();
+
+  useEffect(() => {
+    console.log(
+      "useConnectWorldUsers is practically a noop at the moment, all the real work happens in WorldUsersProvider in src/hooks/users/useWorldUsers.tsx"
+    );
+  }, []);
```
mike-lvov, line 13: It's only used in 2 places. Doesn't it make sense just to use context there?
image

src/hooks/users/useRecentLocationUsers.ts
```diff
@@ -14,22 +21,37 @@ import { useWorldUsers } from "./useWorldUsers";
...
+      .filter(
+        (user) =>
+          user.lastSeenIn?.[locationName] &&
+          normalizeTimestampToMilliseconds(user.lastSeenIn[locationName]) >
+            lastSeenThreshold
+      )
+      .map(userWithLocationToUser);
+
+    return { recentLocationUsers, isWorldUsersLoaded };
+  }, isEqual);
```
mike-lvov, line 51: Since the firestore connect was removed, Do we have users in the redux state? Given that we handle it using the context

src/hooks/users/useRecentWorldUsers.ts
```diff
@@ -1,30 +1,44 @@
...
 import { WithId } from "utils/id";
+import { worldUsersSelector } from "utils/selectors";
 import { normalizeTimestampToMilliseconds } from "utils/time";
 
+import { useSelector } from "hooks/useSelector";
+import { useConnectWorldUsers } from "hooks/users";
 import { useUserLastSeenThreshold } from "hooks/useUserLastSeenThreshold";
 
-import { useWorldUsers } from "./useWorldUsers";
+const noUsers: WithId<User>[] = [];
```
mike-lvov, line 14:
```suggestion
const noUsers: never[] = [];
```

src/hooks/users/useRecentWorldUsers.ts
```diff
@@ -1,30 +1,44 @@
...
-    () => ({
-      recentWorldUsers: worldUsers.filter(
+  const { recentWorldUsers, isWorldUsersLoaded } = useSelector((state) => {
+    const worldUsers = worldUsersSelector(state);
+    const isWorldUsersLoaded = isLoaded(worldUsers);
+
+    if (!worldUsers) return { recentWorldUsers: noUsers, isWorldUsersLoaded };
+
+    const recentWorldUsers = worldUsers
+      .filter(
```
mike-lvov, line 31: Same question as above. How do we have the users in the redux state, given that we don't fetch them

src/hooks/users/useWorldUsers.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+import { WithId } from "utils/id";
+import { worldUsersWithoutLocationSelector } from "utils/selectors";
+import { isDefined } from "utils/types";
+
+import { isLoaded, useFirestoreConnect } from "hooks/useFirestoreConnect";
+import { useSelector } from "hooks/useSelector";
+import { useSovereignVenueId } from "hooks/useSovereignVenueId";
+
+const noUsers: WithId<User>[] = [];
```
mike-lvov, line 14:
```suggestion
const noUsers: never[] = [];
```

src/hooks/users/useWorldUsers.tsx
```diff
@@ -0,0 +1,123 @@
...
+  venueId,
+  children,
+}) => {
+  const {
+    sovereignVenueId,
+    isSovereignVenueIdLoading,
+    errorMsg: sovereignVenueIdError,
+  } = useSovereignVenueId({
+    venueId,
+  });
```
mike-lvov, line 42: probably has conflicts with the recent sovereignVenue changes

src/hooks/users/useWorldUsers.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const shouldConnect =
```
mike-lvov, line 44: Not sure if it was obvious for me what this var is about until I got into the logic. Probably should be renamed

src/hooks/users/useWorldUsers.tsx
```diff
@@ -0,0 +1,123 @@
...
+);
+
+export interface WorldUsersProviderProps {
+  venueId?: string;
+}
+
+export const WorldUsersProvider: React.FC<WorldUsersProviderProps> = ({
+  venueId,
+  children,
+}) => {
```
mike-lvov, line 35: I don't understnad what do we gain by having this in the Context. It basically does what `useConnectUsers` did with no additions?

src/hooks/users/useWorldUsersById.ts
```diff
@@ -1,30 +1,33 @@
...
-import { worldUsersByIdSelector } from "utils/selectors";
+import { WithId } from "utils/id";
+import { worldUsersByIdWithoutLocationSelector } from "utils/selectors";
 
 import { isLoaded } from "hooks/useFirestoreConnect";
 import { useSelector } from "hooks/useSelector";
 
 import { useConnectWorldUsers } from "./useConnectWorldUsers";
 
+const noUsersById: Record<string, WithId<User>> = {};
```
mike-lvov, line 13: This type is misleading. It basically assumes that the empty object has users. (when you try to get noUsersById["randomString"])

src/utils/selectors.ts
```diff
@@ -58,19 +59,33 @@ export const currentVenueSelectorData: SparkleSelector<AnyVenue | undefined> = (
...
-export const worldUsersSelector: SparkleSelector<WithId<User>[] | undefined> = (
-  state
-) => state.firestore.ordered.worldUsers;
+export const worldUsersSelector: SparkleSelector<
+  WithId<UserWithLocation>[] | undefined
+> = (state) => state.firestore.ordered.worldUsers;
+
+export const worldUsersWithoutLocationSelector: SparkleSelector<
+  WithId<User>[] | undefined
+> = (state) => worldUsersSelector(state)?.map(userWithLocationToUser);
```
mike-lvov, line 68: the name `userWithLocationToUser` implies that user will have location?

src/utils/selectors.ts
```diff
@@ -58,19 +59,33 @@ export const currentVenueSelectorData: SparkleSelector<AnyVenue | undefined> = (
...
+  Record<string, UserWithLocation> | undefined
 > = (state) => state.firestore.data.worldUsers;
 
+export const worldUsersByIdWithoutLocationSelector: SparkleSelector<
+  Record<string, WithId<User>> | undefined
+> = (state) => {
+  const worldUsersById = worldUsersByIdSelector(state);
+
+  return mapValues(worldUsersById, (user, userId) =>
+    userWithLocationToUser(withId(user, userId))
```
mike-lvov, line 85: Same here

mike-lvov (changes requested):
I don't think I understand the reason to move users into the Context/Provider?

src/hooks/users/useRecentLocationUsers.ts
0xdevalias: Nope, we want it to have the correct type.

src/components/molecules/TablesUserList/index.ts
0xdevalias: I'll call this NB, but probably happy to do so given it will be an easy change.

src/hooks/users/useConnectWorldUsers.ts
0xdevalias: Probably. This was mostly a placeholder as I was working through this + before I was sure what was going to actually be needed still.

src/hooks/users/useRecentLocationUsers.ts
0xdevalias: I'd have to check my code again, but i'm pretty sure the firestore connect still exists within the context provider doesn't it?

src/hooks/users/useRecentWorldUsers.ts
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1703#discussion_r666009006, nope, because we want it to have the correct type.

src/hooks/users/useRecentWorldUsers.ts
0xdevalias: See my answer in https://github.com/sparkletown/sparkle/pull/1703#discussion_r666010508, but I believe the context provider still uses redux.

src/hooks/users/useWorldUsers.tsx
0xdevalias: As per https://github.com/sparkletown/sparkle/pull/1703#discussion_r666009006, nope, because we want it to have the correct type.

src/hooks/users/useWorldUsers.tsx
0xdevalias: Yeah, makes sense. Though should also be a fairly trivial merge to get it fixed up.

src/hooks/users/useWorldUsers.tsx
0xdevalias: Given the logic is like literally the next bit of code.. is that such a bad thing?

It seems super straightforward and obvious to me. It controls whether `useFirestoreConnect` `shouldConnect`?

Do you have suggestions for better names?

src/hooks/users/useWorldUsers.tsx
0xdevalias: It also provides us explicit guarantees that the data has actually been connected, because it will throw errors otherwise.

This will also likely become more relevant in future followup PRs where I rip out `react-redux-firebase` and connect using the firebase SDK directly.

src/hooks/users/useWorldUsersById.ts
0xdevalias: Yup. Should be `Partial<Record<string, WithId<User>>>`, but this is the whole broken/legacy tech debt aspect of `useWorldUsersById` / `useWorldUsersByIdWorkaround` if you scroll down a bit more in that same file. Which I don't want to touch in this PR as it's going to require changing a LOT of code across the codebase.

src/utils/selectors.ts
0xdevalias: See my explanation to @denisdimitrov about this in https://github.com/sparkletown/sparkle/pull/1703#discussion_r664153565

it takes a `userWithLocation` and makes it a `user`.

src/utils/selectors.ts
0xdevalias: See my explanation to @denisdimitrov about this in https://github.com/sparkletown/sparkle/pull/1703#discussion_r664153565

it takes a `userWithLocation` and makes it a `user`.

src/hooks/users/useRecentLocationUsers.ts
mike-lvov: Yep, it does. It's just unconventional way of utilizing Context for our codebase so I had questions

src/hooks/users/useWorldUsers.tsx
mike-lvov: Not a big deal, just sharing my experience. totally NB. I don't have the suggestions though

src/hooks/users/useRecentLocationUsers.ts
0xdevalias: See further explanations of the 'why' of this in https://github.com/sparkletown/sparkle/pull/1703#discussion_r666014344

0xdevalias: Marking this as resolved. Feel free to reopen if needs be.

src/hooks/users/useRecentWorldUsers.ts
0xdevalias: Marking this as resolved, reopen if needs be.

src/hooks/users/useWorldUsers.tsx
0xdevalias: Fair fair. Will call this 'wont fix' then, unless someone brings up a better name/etc.

mike-lvov (approved):
Looks good as the initial approach for this problem, but I guess it needs a few follow up PRs as it can be insufficient to handle the same amount of users we had during the OHBM event. My conclusion is not based on the actual data, that's just my gut feeling

src/utils/selectors.ts
mike-lvov: > See my explanation to @denisdimitrov about this in #1703 (comment)
> 
> it takes a userWithLocation and makes it a user.

Given that it brought attention of both me and @denisdimitrov, it can be a good sign that the naming is suboptimal

0xdevalias: It's a fairly common pattern to use for 'converter' type functions. eg.

https://date-fns.org/v2.22.1/docs/hoursToSeconds

[OOS/NB] I'm open to suggestions for better naming.

0xdevalias:
> Looks good as the initial approach for this problem, but I guess it needs a few follow up PRs as it can be insufficient to handle the same amount of users we had during the OHBM event. My conclusion is not based on the actual data, that's just my gut feeling

@mike-lvov That's kind of my gut feel at the moment as well, but i'm not actually sure if that will even be true or not. It really depends how much 'browser effort' will need to go into those `isEqual` / `shallowEqual` checks. The problem before was that because we weren't using those + the bug in `react-redux` that broke memoisation, the app was having cascading flow on effects causing re-renders. My theory is that the main part of the performance issues was in all of that 'flow on' stuff, so the theory is that even if these bits end up doing more work than they might otherwise need to 'at the top', we probably won't see the bigger 'flow on' impact.

That said, my plans for followup include replacing the calls to `react-redux-firebase` / `redux-firestore` with just a raw `.onSnapshot` listener using the firebase SDK directly + using `redux-toolkit` to make some nice 'boilerplate free' redux patterns to 'split the data' into relevant chunks at the time we store it in redux. That would then allow us to be more selective in what data we are selecting/updating, and likely remove (or greatly reduce) the need for deep compares/etc at the `useSelector` level of things.

src/components/molecules/TablesUserList/index.ts
0xdevalias: fixed in 961b90e0fe910b273a48ad52c520b92132a3b7a2

src/hooks/users/useConnectWorldUsers.ts
0xdevalias: fixed in 985301f16c477e29959e12a6301f71a8901cc5d2

src/hooks/users/useWorldUsers.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1703/commits/ef36347c3d89ff3877cb51ee7d81e5a175b22040

src/hooks/users/useWorldUsers.tsx
0xdevalias: Explained this a bit more in https://github.com/sparkletown/sparkle/pull/1703#issuecomment-876312854

PR #1706, opened 2021-07-01
add ability to hide full or locked tables

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/851


0xdevalias (changes requested):
We have tables in more locations than just the jazz bar. This should be applied to all of them. And utils/shared components extracted to support this feature in all locations without duplicating the code.

Further specific review comments to come when I’m on a computer.

sunny-viktoryia:
> We have tables in more locations than just the jazz bar. This should be applied to all of them. 

@yurii-lubynets have you fixed that and is the PR ready for the review?

yurii-lubynets:
@sunny-viktoryia 
The PR is ready for review. The feature is available for Jazzbar, Conversation space templates as per comment here https://github.com/sparkletown/internal-sparkle-issues/issues/851#issuecomment-873308110

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -24,14 +25,15 @@ export const ConversationSpace: React.FunctionComponent = () => {
   const venue = useSelector(currentVenueSelectorData);
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const defaultTables = venue?.config?.tables ?? TABLES;
+  const [tables, setTables] = useState(defaultTables);
```
denisdimitrov, line 29: `useState(defaultTables);` This default value will be set only once on the initial render, is this what we want? Are you sure that this won't create a problem later on.

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -88,6 +90,16 @@ export const ConversationSpace: React.FunctionComponent = () => {
               )}
             </div>
           </div>
+          {!seatedAtTable && (
+            <TablesControlBar
+              className="control-bar"
+              defaultTables={defaultTables}
+              venue={venue}
+              users={recentVenueUsers}
+              updateTables={setTables}
```
denisdimitrov, line 99: This is a callback and should be named something like `onUpdateTables`

denisdimitrov: Actually, by looking at the implementation of `TablesControlBar` a better name would be `onToggleTables`

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+        Hide full/locked tables
+      </label>
+      <input
+        type="checkbox"
+        name="toggle-tables"
+        id="chk-toggle-tables"
+        defaultChecked={false}
+        onClick={toggleTables}
+      />
+    </div>
```
denisdimitrov, line 69: You should try and use our already existing atom component `Toggler`.

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+  isChecked,
+  venue,
+  users,
+  updateTables,
+  className,
+}) => {
+  const experience = useSelector(experienceSelector);
+
+  // not locked and not full tables
+  const freeTables = defaultTables
```
denisdimitrov, line 35: You should `useMemo` this to avoid extra calculations on re-rendering.

src/components/molecules/TablesControlBar/TablesControlBar.scss
```diff
@@ -0,0 +1,8 @@
+@import "scss/constants.scss";
+
+$control-bar-height: 40px;
+
+.TablesControlBar {
+  height: $control-bar-height;
+  margin-top: 13px;
```
denisdimitrov, line 7: Can you use our global constants here for this value?

denisdimitrov (changes requested):

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: IMO I should store `defaultTables` and toggle them by filtering (if they are locked or full). I assume yes, it's ok to set them on the initial render.
Please clarify what problem it could create later?

src/components/molecules/TablesControlBar/TablesControlBar.tsx
yurii-lubynets: According to design, the toggle component should be a checkbox and `Toggler` looks quite differ

https://www.figma.com/file/UFAJMGn4ucPtQNe2NC55iq/Venues?node-id=4%3A315

src/components/molecules/TablesControlBar/TablesControlBar.scss
yurii-lubynets: Can't find any relevant var in global constants for the value of `13px`. Anytime `13px` is used in the app it's just set in .scss file

src/components/molecules/TablesControlBar/TablesControlBar.tsx
yurii-lubynets: fixed in f6a6e31bf722054e762d6e00c4ad34ed6e248279

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: fixed in f6a6e31bf722054e762d6e00c4ad34ed6e248279

src/components/molecules/TablesControlBar/TablesControlBar.tsx
denisdimitrov: Then how about our `Checkbox` component?

src/components/templates/ConversationSpace/ConversationSpace.tsx
denisdimitrov: Well, for instance, if the prop updates and `defaultTables` changes, this won't re-trigger itself and the default values will be the same.

src/components/molecules/TablesControlBar/TablesControlBar.tsx
yurii-lubynets: That looks nice! Fixed in 868c63875c76829e1b6136de6a9549909e245321

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: Well, I've fixed the edge case when the `defaultTables` are changed in 876c4407b981917b01639e7768d6a3a9d6ef75e8

src/components/molecules/TablesControlBar/TablesControlBar.scss
0xdevalias: Is there a margin constant that is close to it? Do we know why it's specifically 13px?

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -24,14 +25,17 @@ export const ConversationSpace: React.FunctionComponent = () => {
   const venue = useSelector(currentVenueSelectorData);
   const { recentVenueUsers } = useRecentVenueUsers();
 
+  const defaultTables = venue?.config?.tables ?? TABLES;
+  const [tables, setTables] = useState(defaultTables);
+
+  useEffect(() => setTables(defaultTables), [defaultTables]);
```
denisdimitrov, line 31: There is duplicated logic between this component and JazzTab. They both have `useState` `defaultTables` and `useEffect` instead of accepting `defaultTables` is it possible to make it accept `venue.config.tables` and move this whole logic into the `TablesControlBar` That way we won't have to do this twice.

denisdimitrov: @yurii-lubynets ^ Not really a change request, more of a comment.

yurii-lubynets: Actually,  `tables` and `jazzbarTables` are used for JazzTab and ConversationSpace components, that's why extracting them into `TablesControlBar` could be redundant

src/components/molecules/TablesControlBar/TablesControlBar.scss
yurii-lubynets: Yeah, I'm using `$spacing--md` (12px) instead of. Well, 13px was in Figma design, and I hope it's fine to use 12px

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+      defaultTables
+        .filter((table) => !experience?.tables[table.title]?.locked)
+        .filter(
+          (table) =>
+            !users.filter(
+              (u) => u.data?.[venue.name]?.table === table.reference
+            ).length
+        ),
+    [defaultTables, experience?.tables, users, venue.name]
+  );
```
sunny-viktoryia, line 48: moreover I believe this logic should be moved out of the `TableControlBar`.
Here we just define if the checkbox is checked or not, and then trigger the updates in the parent component and let it decided what to do with that information.

if you want not to repeat the logic. You can move that filtering function to some utils. 

<hr />

no need to run `filter` twice. You can do all the logic in the first one.

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -91,6 +95,16 @@ export const ConversationSpace: React.FunctionComponent = () => {
...
             </div>
           </div>
+          {!seatedAtTable && (
+            <TablesControlBar
+              className="control-bar"
+              defaultTables={defaultTables}
+              venue={venue}
+              users={recentVenueUsers}
+              onToggleTables={setTables}
+              isChecked={tables !== defaultTables}
```
sunny-viktoryia, line 105: hmmm, not sure if we should use that approach to understand if the input was toggled. Can we use `useShowHide` hook in the component instead and rewrite the logic?
```
tables !== defaultTables
```

sunny-viktoryia (changes requested):
Haven't reviewed it fully because I feel that the logic of toggling should be changed.

`<TableControlBar />` should only notify parent component that the show-hide input was toggled and the parent component should handle that information.

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  isChecked,
+  venue,
+  users,
+  onToggleTables,
+  className,
+}) => {
+  const experience = useSelector(experienceSelector);
+
+  // not locked and not full tables
+  const freeTables = useMemo(
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 5aa6329bd90551d39634b475d84adbb35e5e663a

src/components/molecules/TablesControlBar/TablesControlBar.tsx
yurii-lubynets: Thank you for idea!
fixed in 5aa6329bd90551d39634b475d84adbb35e5e663a

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: fixed in 5aa6329bd90551d39634b475d84adbb35e5e663a

yurii-lubynets:
@sunny-viktoryia 
Thanks for the review. Feel free to make re-review

sunny-viktoryia:
@yurii-lubynets can you check if you have same bug? haven't looked at the reason of it in the code.

Pay attention to the "Table 10", which is not locked but it gets hidden when I toggle on.
recording (3)

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -249,6 +264,12 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
                     />
                   )} */}
                 </div>
+              ) : (
+                <TablesControlBar
+                  className="control-bar"
+                  handleChecked={onToggleTables}
+                  isChecked={isChecked}
+                />
```
sunny-viktoryia, line 272: `!shouldShowReactions` is not equal to `!seatedAtTable` which should be the right thing to check. As far as I understand reactions can be turned off in the venue settings as well and in this case you'll see `TablesControlBar`, which is not correct. 
so, please, fix that.

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -27,13 +30,25 @@ export const ConversationSpace: React.FunctionComponent = () => {
   const venue = useSelector(currentVenueSelectorData);
   const { recentVenueUsers } = useRecentVenueUsers({ venueName: venue?.name });
 
+  const { isShown: isChecked, toggle: toggleTables } = useShowHide();
```
sunny-viktoryia, line 33: `isChecked` is not a self-explanatory name for such a complex component. -> `areNotAvailableTablesShown` ?
`toggleTables` -> maybe `toggleNotAvailableTables` ?

yurii-lubynets:
> @yurii-lubynets can you check if you have same bug? haven't looked at the reason of it in the code.
> 
> Pay attention to the "Table 10", which is not locked but it gets hidden when I toggle on.
> recording (3)

fixed in c5a7a8087f30e7a96a228fee019016e7dabbf61b

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -27,13 +30,25 @@ export const ConversationSpace: React.FunctionComponent = () => {
...
+
+  const defaultTables = venue?.config?.tables ?? TABLES;
+  const { availableTables } = useTables({ defaultTables });
+
+  const [tables, setTables] = useState(defaultTables);
+
+  useEffect(() => setTables(defaultTables), [defaultTables]);
```
sunny-viktoryia, line 40: useTables + useState + useEffect + onToggleTables -> the logic seem to be overcomplicated.

```ts
  const { isShown: areNotAvailableTablesShown, toggle: toggleNotAvailableTables } = useShowHide();

  const tablesToShow = useMemo(() => (venue?.config?.tables ?? TABLES).filter(table => {
    return areNotAvailableTablesShown || !(isLockedTable(table) || isFullTable(table))
  })), [...])
```
the names of varialbes can be better I suppose.

then you can remove `onToggleTables`

and in `<TableConrolBar handleChecked={toggleNotAvailableTables} .../>`

sunny-viktoryia: that's just a draft solution.

same for JazzBar (and consider refactoring and moving the logic to some utils/table to reuse it)

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+
+import classNames from "classnames";
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

src/components/molecules/TablesControlBar/TablesControlBar.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { Checkbox } from "components/atoms/Checkbox";
+
+import "./TablesControlBar.scss";
+
+export interface TablesControlBarProps {
+  isChecked: boolean;
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

src/components/templates/ConversationSpace/ConversationSpace.scss
```diff
@@ -254,6 +254,10 @@
     grid-gap: 1rem;
   }
 
+  .control-bar {
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
yurii-lubynets: fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: Thank you for help!
fixed in e8298bec7e35a4e40e7175b130015db2080f52e5

sunny-viktoryia (commented):
a new round of review from me

the styles seem to get broken a bit.
image

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -40,6 +43,15 @@ export const ConversationSpace: React.FC<ConversationSpaceProps> = ({
 
   const { recentVenueUsers } = useRecentVenueUsers({ venueName: venue?.name });
 
+  const {
+    isShown: showAvailableTables,
+    toggle: togglAvailableTables,
+  } = useShowHide();
+
+  const defaultTables = venue?.config?.tables ?? TABLES;
```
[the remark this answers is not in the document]
0xdevalias: No point in useMemo’ing this.

src/hooks/useTables.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface UseTablesProps {
+  defaultTables: Table[];
+  showAvailableTables: boolean;
+}
+
+export interface UseTablesData {
+  tablesToShow: Table[];
+}
+
+export const useTables: ReactHook<UseTablesProps, UseTablesData> = ({
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 83f0efe0fcdefb2d55d3bcd22e6d3564ad047267

src/hooks/useTables.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { currentVenueSelectorData } from "utils/selectors";
+import { experienceSelector } from "utils/selectors";
+import { getUserExperience } from "utils/user";
+
+import { useSelector } from "hooks/useSelector";
+import { useRecentVenueUsers } from "hooks/users";
+
+export interface UseTablesProps {
+  defaultTables: Table[];
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 83f0efe0fcdefb2d55d3bcd22e6d3564ad047267

src/hooks/useTables.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface UseTablesData {
+  tablesToShow: Table[];
+}
+
+export const useTables: ReactHook<UseTablesProps, UseTablesData> = ({
+  defaultTables,
+  showAvailableTables,
+}) => {
+  const venue = useSelector(currentVenueSelectorData);
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 83f0efe0fcdefb2d55d3bcd22e6d3564ad047267

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -60,7 +60,8 @@ export interface TablesUserListProps {
   venueName: string;
   setSeatedAtTable: (value: string) => void;
   seatedAtTable: string;
-  customTables?: Table[];
+  customTables: Table[];
```
yurii-lubynets, line 63: checked `<TablesUserList/>` usage and the `customTables` are always required

src/pages/FriendShipPage/FriendShipPage.tsx
```diff
@@ -54,6 +54,7 @@ export const FriendShipPage: React.FunctionComponent = () => {
               customTables={FRIENDSHIP_CUSTOM_TABLES}
               leaveText="Return to Isle of Friends"
               joinMessage={true}
+              showAvailableTables={false}
```
yurii-lubynets, line 57: set as `false` to align with the latest changes and to render all tables

yurii-lubynets:
@sunny-viktoryia 
- removed unused custom hook `useAvailableTables`
- refactored `<TablesUserList/>` component (discussed previously)
- updated styles and minor improvements

Please review

whoaitsaimz:
Or @denisdimitrov please

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -191,7 +215,14 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
   const renderedTables = useMemo(() => {
     if (isSeatedAtTable) return;
 
-    return tables.map((table: Table, index: number) => (
+    // @debt refactor table properties into one place
+    const tablesToShow = showAvailableTables
+      ? tables.filter(
+          (table) => !(isFullTable(table) || isLockedTable(table.title))
```
denisdimitrov, line 221: `isLockedTable(table.title)` should be using `table.reference`. I'm pretty sure this won't work if you edit the table and change the title.

denisdimitrov (changes requested):

src/components/molecules/TablesUserList/TablesUserList.tsx
yurii-lubynets: fixed in 5107070527cdd346077ab84f3249ad327e1ab5d7

src/components/templates/ConversationSpace/ConversationSpace.tsx
```diff
@@ -40,6 +42,11 @@ export const ConversationSpace: React.FC<ConversationSpaceProps> = ({
 
   const { recentVenueUsers } = useRecentVenueUsers({ venueName: venue?.name });
 
+  const {
+    isShown: showAvailableTables,
```
sunny-viktoryia, line 46: same comment about the property hame; maybe `showOnlyAvailableTables` is better?
`toggleTablesVisibility` ?

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -53,6 +54,11 @@ interface JazzProps {
 const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
   const { recentVenueUsers } = useRecentVenueUsers({ venueName: venue.name });
 
+  const {
+    isShown: showAvailableTables,
+    toggle: togglAvailableTables,
```
sunny-viktoryia, line 59: comment about the names of the variables.

sunny-viktoryia (commented):
few more comments from me. The logic seems correct now. The majority of the comments are about the variable names. 

Although I have raised concern here about the performance. Not sure if it's a good idea to iterate through the online users several times (looks like we do that for all the tables twice, would be great if we could that only once if possible).

I believe we could create a key-value object which would store the `{ [table.reference]: usersAtTableArray }` and then could use that information where needed: 
* if the table is empty
* if the table is full (twice)

so would be great if @yurii-lubynets could do the refactoring of that part (if you have the capacity for that). Not sure if it has a huge performance impact though.

_[dreaming]_ definitely OOS, in an ideal Sparkle ✨  would be cool to be able to run emulators for each venue type to test the performance and see how our changes impact that. mmmmmm.

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: `toggleTablesVisibility` is fine for me

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -60,7 +60,8 @@ export interface TablesUserListProps {
   venueName: string;
   setSeatedAtTable: (value: string) => void;
   seatedAtTable: string;
-  customTables?: Table[];
+  customTables: Table[];
+  showAvailableTables: boolean;
```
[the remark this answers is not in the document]
yurii-lubynets: Let's freeze variables naming for this PR. I assume the `showOnlyAvailableTables` and `toggleTablesVisibility` look perfect and suitable!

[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -71,6 +72,7 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
   setSeatedAtTable,
   seatedAtTable,
   customTables,
+  showAvailableTables,
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -130,8 +132,29 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
     [recentVenueUsers, user, venueName, videoRoom]
   );
 
+  // @debt refactor table properties into one place
+  const isLockedTable = useCallback(
+    (table: string) => isTruthy(experience?.tables?.[table]?.locked),
+    [experience?.tables]
+  );
+
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -140,9 +163,9 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
       if (!areUsersAtTable) return false;
 
       // Locked state is in the experience record
-      return isTruthy(experience?.tables?.[table]?.locked);
+      return isLockedTable(table);
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -140,9 +163,9 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
...
     },
-    [experience?.tables, recentVenueUsers, venueName]
+    [recentVenueUsers, venueName, isLockedTable]
```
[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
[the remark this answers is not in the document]
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/templates/ConversationSpace/ConversationSpace.tsx
yurii-lubynets: fixed in 172c71750e0f23a6675546856b40288b3521892a

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -191,7 +215,14 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
...
+    const tablesToShow = showAvailableTables
+      ? tables.filter(
+          (table) => !(isFullTable(table) || isLockedTable(table.title))
+        )
+      : tables;
+
+    return tablesToShow.map((table: Table, index: number) => (
       <TableComponent
         key={table.reference}
         experienceName={venueName}
```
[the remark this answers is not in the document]
yurii-lubynets: refactoring `TableComponent` is definitely OOS of this PR. I left a comment about it here 0ea9df24c902bbbb16b10efebf8daec4395c3ab3

yurii-lubynets:
> few more comments from me. The logic seems correct now. The majority of the comments are about the variable names.
> 
> Although I have raised concern here about the performance. Not sure if it's a good idea to iterate through the online users several times (looks like we do that for all the tables twice, would be great if we could that only once if possible).
> 
> I believe we could create a key-value object which would store the `{ [table.reference]: usersAtTableArray }` and then could use that information where needed:
> 
> * if the table is empty
> * if the table is full (twice)
> 
> so would be great if @yurii-lubynets could do the refactoring of that part (if you have the capacity for that). Not sure if it has a huge performance impact though.
> 
> _[dreaming]_ definitely OOS, in an ideal Sparkle ✨ would be cool to be able to run emulators for each venue type to test the performance and see how our changes impact that. mmmmmm.

@sunny-viktoryia 
have refactored `<TablesUserList/>` to store the `{ [table.reference]: usersAtTableArray }` and use this information where needed. Please review

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -130,19 +133,41 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
...
+      const numberOfSeatsLeft =
+        table.capacity &&
+        table.capacity - usersSeatedAtTables[table.reference].length;
+      return numberOfSeatsLeft === 0;
+    },
+    [usersSeatedAtTables]
+  );
+
   const tableLocked = useCallback(
     (table: string) => {
```
sunny-viktoryia, line 163: `table` -> `tableReference` ?

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -130,19 +133,41 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
     [recentVenueUsers, user, venueName, videoRoom]
   );
 
+  const itemsToObjectByRefReducer = useCallback(
```
sunny-viktoryia, line 136: `itemsToObjectByRefReducer` -> `usersAtTableReducer` ?

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -130,19 +133,41 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
...
+    (obj: Record<string, WithId<User>[]>, table: Table) => ({
```
sunny-viktoryia, line 137: `obj`-> `acc` ? (accumulator)

sunny-viktoryia (commented):
suuuuuper clooooooooose!! 🙃 
just a couple of variable name changes and should be good to go,

src/components/molecules/TablesUserList/TablesUserList.tsx
denisdimitrov: Looks like this has a return value, maybe it should be using `useMemo` instead.

sunny-viktoryia: @denisdimitrov this is a function that is used in `arr.reduce()` a bit later - it takes a table and accumulate the result in the accumulator. I believe we can leave `useCallback` here. what do you think?

0xdevalias: @sunny-viktoryia Is this function 'closing around' internal state? If so, `useCallback` seems fine. If not, then should be extracted out of the component body into a `utils/*` function or similar.

src/components/molecules/TablesUserList/TablesUserList.tsx
0xdevalias: `obj` is probably fine here given the name of the reducer also calls it 'object'. I try and aim to name it in a way that it would make the most sense to a human.

src/components/molecules/TablesUserList/TablesUserList.tsx
yurii-lubynets: fixed in 46db62e1cd0e7d70ba136b3235574076c0ca5d6d

src/components/molecules/TablesUserList/TablesUserList.tsx
yurii-lubynets: left as `obj`

src/components/molecules/TablesUserList/TablesUserList.tsx
sunny-viktoryia: @0xdevalias, the function uses `recentVenueUsers`  and `venueName`. So i'd leave it here.

denisdimitrov: I think the correct way would be to use `useMemo` because it will memo the result, not the function. That way if the dep array doesn't change the value will be returned without the calculations. However this is nitpicking and I don't have a strong opinion about it.

sunny-viktoryia: @denisdimitrov hmm but it does not have the result by the moment of its creation (if you do not mean the function as the result). It has a function which will produce the result once you pass the arguments.
we could use `useMemo` but it would look like something like this:
```js
useMemo(() => (obj, table) => ({ ... result }), [dependencies])
```
and the current version is:
```js
useCallback((obj, table) => ({ ... result }), [dependencies])
```

sunny-viktoryia (approved):
mmmm, finally, approving!! Looks good to merge. 🙃 🎉 

Thanks for your patience and for being open to fix my requested changes. We did it! It looks so much better to my eye than the 1st version! 😍

src/components/molecules/TablesUserList/TablesUserList.tsx
denisdimitrov: @sunny-viktoryia That's true and makes sense. The function that's using this should be using memo, and it is. Looks good to me. :)

sunny-viktoryia: @denisdimitrov cool! glad we've sorted that out. 🙃

PR #1712, opened 2021-07-03
fix NavSearchBar to filter out rooms that aren't enabled (and events in those rooms) from search results

0xdevalias:
Fixed `NavSearchBar` to filter out rooms that aren't enabled (and events in those rooms) from search results.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/147

**Note:** I haven't tested these changes in a clickthrough, so that should definitely be done before merging this PR.

### DIFF
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -70,7 +70,13 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     () =>
       relatedVenues
         .flatMap((venue) => venue.rooms ?? [])
-        .filter((room) => room),
+        .filter((room) => {
+          if (isDefined(room.type) && COVERT_ROOM_TYPES.includes(room.type)) {
+            return false;
+          }
+
+          return room.isEnabled;
+        }),
     [relatedVenues]
   );
 
@@ -96,12 +102,7 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     /* @debt we really shouldn't be using the index as part of the key here, it's unstable.. but rooms don't have a unique identifier */
     return (
       relatedRooms
-        ?.filter(
-          (room) =>
-            (!room.type || !COVERT_ROOM_TYPES.includes(room.type)) &&
-            room.isEnabled &&
-            room.title.toLowerCase().includes(searchQuery)
-        )
+        .filter((room) => room.title.toLowerCase().includes(searchQuery))
         .map((room, index) => (
           <NavSearchResult
             key={`room-${room.title}-${index}`}
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -87,6 +87,7 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
         ?.filter(
           (room) =>
             (!room.type || !COVERT_ROOM_TYPES.includes(room.type)) &&
+            room.isEnabled &&
             room.title.toLowerCase().includes(searchQuery)
         )
         .map((room, index) => (
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -152,11 +152,15 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
 
     return relatedEvents
       .filter((event) => {
-        const isRoomEnabled = isDefined(event.room)
-          ? relatedRoomsByTitle[event.room]?.isEnabled ?? true
-          : true;
+        const isEventRoomEnabledAndExists =
+          isDefined(event.room) && event.room !== ""
+            ? isDefined(relatedRoomsByTitle[event.room])
+            : true;
 
-        return isRoomEnabled && event.name.toLowerCase().includes(searchQuery);
+        return (
+          isEventRoomEnabledAndExists &&
+          event.name.toLowerCase().includes(searchQuery)
+        );
       })
       .map((event) => {
         const imageUrl =
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -74,6 +74,18 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     [relatedVenues]
   );
 
+  const relatedRoomsByTitle = useMemo<Partial<Record<string, Room>>>(
+    () =>
+      relatedRooms.reduce(
+        (relatedRoomsByTitle, room) => ({
+          ...relatedRoomsByTitle,
+          [room.title]: room,
+        }),
+        {}
+      ),
+    [relatedRooms]
+  );
+
   const { isEventsLoading, events: relatedEvents } = useVenueEvents({
     venueIds: relatedVenueIds,
   });
@@ -138,7 +150,13 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     if (!searchQuery) return [];
 
     return relatedEvents
-      .filter((event) => event.name.toLowerCase().includes(searchQuery))
+      .filter((event) => {
+        const isRoomEnabled = isDefined(event.room)
+          ? relatedRoomsByTitle[event.room]?.isEnabled ?? true
+          : true;
+
+        return isRoomEnabled && event.name.toLowerCase().includes(searchQuery);
+      })
       .map((event) => {
         const imageUrl =
           relatedRooms.find((room) => room.title === event.room)?.image_url ??
@@ -161,7 +179,14 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
           />
         );
       });
-  }, [searchQuery, relatedEvents, relatedRooms, relatedVenues, clearSearch]);
+  }, [
+    searchQuery,
+    relatedEvents,
+    relatedRoomsByTitle,
+    relatedRooms,
+    relatedVenues,
+    clearSearch,
+  ]);
 
   const numberOfSearchResults =
     foundRooms.length + foundEvents.length + foundUsers.length;
--- a/src/components/molecules/NavSearchBar/NavSearchBar.tsx
+++ b/src/components/molecules/NavSearchBar/NavSearchBar.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useState, ChangeEvent, useMemo } from "react";
 import classNames from "classnames";
-import { isEqual } from "lodash";
+import { isEqual, reduce } from "lodash";
 
 import { faSearch } from "@fortawesome/free-solid-svg-icons";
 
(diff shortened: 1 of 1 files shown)


src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -137,7 +151,13 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     if (!searchQuery) return [];
 
     return relatedEvents
-      .filter((event) => event.name.toLowerCase().includes(searchQuery))
+      .filter((event) => {
+        const isRoomEnabled = isDefined(event.room)
+          ? relatedRoomsByTitle[event.room]?.isEnabled ?? true
+          : true;
```
sunny-viktoryia, line 157: current version does not exclude events that are related to disabled venues because:
1) room field can be an empty string and it means that event is attached to a venue (looks like we need to change something in events creating/editing)
2) `relatedRoomsByTitle[event.room]?.isEnabled ?? true` this condition was `true` for the events whose rooms were not added to `relatedRoomsByTitle` ( `elatedRoomsByTitle[event.room]?.isEnabled` === `undefined`)

looks like you have already filtered rooms by `isEnabled` so we can just check here if the room exists in the list.
```suggestion
        const isRoomEnabled =
          isDefined(event.room) && event.room !== ""
            ? isDefined(relatedRoomsByTitle[event.room])
            : true;
```

sunny-viktoryia (changes requested):
I have clicked through. Looks like room filtering works fine, but there is an issue with the events. (see my comment)

Also, I think we should also hide these events/rooms in Schedule, but probably it's OOS of this PR.

\+ a question: do we need to hide events that are related directly to venues when the room that targets to that venue is disabled? 🤔 (or will there be a capability to hide venues/events in Admin V3?)

0xdevalias:
> + a question: do we need to hide events that are related directly to venues when the room that targets to that venue is disabled? 🤔 (or will there be a capability to hide venues/events in Admin V3?)

@sunny-viktoryia I opted to explicitly not touch that edgecase, as there is technically no way for a venue itself to be 'turned off', only rooms can be. And there could be 2 different partymaps that both refer to the same venue, one with a room enabled, the other with it disabled.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: > 2) `relatedRoomsByTitle[event.room]?.isEnabled ?? true` this condition was `true` for the events whose rooms were not added to `relatedRoomsByTitle` ( `relatedRoomsByTitle[event.room]?.isEnabled` === `undefined`)

If I remember correctly, I believe that's an explicit choice I made to make the logic work. The name of the variable isn't the best probably.. but I believe the logic here should be "if there is a room and it is enabled, or if the event is not linked to a room, keep it"

0xdevalias: Not sure the suggested changes are correct to make.

0xdevalias: > room field can be an empty string and it means that event is attached to a venue 

So off the top of my head we need to basically handle:

- event without room (eg. linked to venue)
- event with undefined room (this is probably the same thing as above)
- event with room set to `""` (this is probably the same thing as above)
- event with an explicit room, and that room `isEnabled`

---

> (looks like we need to change something in events creating/editing)

This is part of our legacy tech debt patterns of avoiding using `undefined` and similar on backend functions. Definitely something to be changed in future, but well and truly OOS of this issue. It'll be a big/painful refactor I believe.

0xdevalias: > looks like you have already filtered rooms by `isEnabled` so we can just check here if the room exists in the list.

If that's the case, then the variable name and such may need to change here? I'd need to see the code in context to be 100%.

alexey-petrov: @sunny-viktoryia @0xdevalias fixed event find logic in 3948212 as per Viktoria's suggestion, did really help to hide events that are based in disabled rooms
All the other cases that @0xdevalias described are covered, so that should be fine
User can still point an event to a disabled/non-existing venue via room url, but don't think that we can fix it unless we re do the event create/edit mode as per @sunny-viktoryia comment

alexey-petrov:
@sunny-viktoryia feel free to re-review

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -152,11 +152,15 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
 
     return relatedEvents
       .filter((event) => {
-        const isRoomEnabled = isDefined(event.room)
-          ? relatedRoomsByTitle[event.room]?.isEnabled ?? true
-          : true;
+        const isEventRoomEnabledAndExists =
```
sunny-viktoryia, line 155: not sure if your chosen name is 100% valid.
confused about `...AndExists`: for me sounds wrong here.

I'd better rename:
`relatedRooms` -> `enabledRelatedRooms`
`relatedRoomsByTitle` -> `enabledRelatedRoomsByTitle`
`isEventRoomEnabledAndExists` -> `isEventRoomEnabled`

sunny-viktoryia (approved):
Looks good to go! 🙃 

would be wonderful to fix variable names before the merge (see my comment).

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -70,10 +70,28 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
...
+    () =>
+      relatedRooms.reduce(
+        (relatedRoomsByTitle, room) => ({
+          ...relatedRoomsByTitle,
+          [room.title]: room,
+        }),
+        {}
+      ),
+    [relatedRooms]
+  );
```
denisdimitrov, line 93: I suppose this is transforming array into an object. Can we use lodash here?

0xdevalias: We might be able to.. but why would we? Reduce is a pretty straightforward thing to do.

alexey-petrov: sure, used lodash in a3708be

src/components/molecules/NavSearchBar/NavSearchBar.tsx
alexey-petrov: done a3708be

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: > User can still point an event to a disabled/non-existing venue via room url, but don't think that we can fix it unless we re do the event create/edit mode as per @sunny-viktoryia comment

That's fine a fine edgecase to leave out for now.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: @alexey-petrov Why did you import `reduce` from `lodash` and replace our existing code there? There should be no need to use a helper from lodash, just use the standard built in `.reduce` functionality the language provides.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: @alexey-petrov see my comments in https://github.com/sparkletown/sparkle/pull/1712#discussion_r665838282 No idea why we used `lodash` when the old code was already using the built in language feature `.reduce` already, and should have continued to be used.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useCallback, useState, ChangeEvent, useMemo } from "react";
 import classNames from "classnames";
-import { isEqual } from "lodash";
+import { isEqual, reduce } from "lodash";
```
0xdevalias, line 3: There's no need to use `reduce` from `lodash`. The language already has built in functionality for this: `theArray.reduce`. This needs to be reverted and fixed in a followup PR.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -66,14 +66,33 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
...
+      reduce(
+        enabledRelatedRooms,
+        (enabledRelatedRoomsByTitle, room) => ({
+          ...enabledRelatedRoomsByTitle,
+          [room.title]: room,
+        }),
+        {}
+      ),
+    [enabledRelatedRooms]
+  );
```
0xdevalias, line 94: There's no need to use `reduce` from `lodash`. The language already has built in functionality for this: `theArray.reduce`. This needs to be reverted and fixed in a followup PR.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: @alexey-petrov see my comments in https://github.com/sparkletown/sparkle/pull/1712#discussion_r665838282 No idea why we used `lodash` when the old code was already using the built in language feature `.reduce` already, and should have continued to be used.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -137,10 +152,20 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
...
+      .filter((event) => {
+        const isEventRoomEnabled =
+          isDefined(event.room) && event.room !== ""
+            ? isDefined(enabledRelatedRoomsByTitle[event.room])
+            : true;
+
+        return (
+          isEventRoomEnabled && event.name.toLowerCase().includes(searchQuery)
+        );
+      })
```
0xdevalias, line 164: This might be more efficient to extract as a `useMemo`'d `relatedEventsInEnabledRooms` or something like that.. (except for the `event.name.toLowerCase().includes(searchQuery)` part which should stay here)

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -137,10 +152,20 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
...
       .map((event) => {
         const imageUrl =
-          relatedRooms.find((room) => room.title === event.room)?.image_url ??
+          enabledRelatedRooms.find((room) => room.title === event.room)
+            ?.image_url ??
```
0xdevalias, line 168: This should make use of `enabledRelatedRoomsByTitle[event.room]` rather than re-searching through them all again.

0xdevalias (commented):
[Retroactive Review] A few things to be fixed in a followup PR @alexey-petrov

(cc // @sunny-viktoryia @denisdimitrov FYI)

src/components/molecules/NavSearchBar/NavSearchBar.tsx
alexey-petrov: I guess that the intention to use `lodash` was that it's faster, at least it's what I found using benchmarks
though not sure it's the idea that @denisdimitrov had originally

PR #1713, opened 2021-07-03
[sub-PR] VenuePage load performance improvements

goran-peoski-work:
Sub-PR expanding on the removal of fetching all venues across the codebase #1620
with some changes for improved loading performance specific to VenuePage.
Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/839 .

There are 3 main changes:
- venues fetching from PR #1620
- ~BE function for earlier venue and events fetch~ (no longer relevant)
- lazy loading of components / chunking of code bundles.

No longer relevant due to higher request count and negligible performance impact:

> There are two methods of fetching venue and events data in VenuePage:
> - pre, using `firebase.functions().httpsCallable()` to call BE function
> - post, using original `useConnectCurrentVenue()`.
> 
> Using only pre- or post- method of getting data is OK,
> but using them in tandem provides fallback and opportunities:
> 
> - not being locked behind authentication, so data can be preloaded / cached
> - deterministically fetch venue and events info first before all other.
> 

Comparisons of VenuePage loading performance:
sparkle-venue-page-01

`React.lazy` loading components lessens the initial bundle size and blocking of subsequent requests
 and moves those chunks after the data pull and preload of party map assets.

`<Suspense>` could use simplified div elements with probably inline styles
to provide the illusion of the components being there before actual load,
but that's OOS to performance and will not help reduce time to first interaction.
That may also require inlining css and fonts.

`<Loading>` component (or similar generic one) not used as fallback because
the border aligned components show up all lined up at the center of the screen.

There might be OOS possibility for lazy loaded party map and rooms to have a fallback
that may reduce time to interactive (e.g. plain icon for rooms, no background...).

Chunks before lazy load being used:
```
File sizes after gzip:

  757.68 KB         build\static\js\2.ad1efc2b.chunk.js
  133.44 KB (+4 B)  build\static\css\main.f23b4b18.chunk.css
  129.27 KB         build\static\js\main.5b1778a9.chunk.js
  9.48 KB           build\static\js\3.a560293f.chunk.js
  1.86 KB           build\static\css\2.ca756f56.chunk.css
  1.17 KB           build\static\js\runtime-main.3a521cfb.js
```

Chunks after lazy load being used:
```
  585.24 KB (+1 B)   build\static\js\2.83a9b1ba.chunk.js
  95.58 KB (+150 B)  build\static\js\main.a37d074d.chunk.js
  93.4 KB            build\static\css\main.0e2ae69a.chunk.css
  93.28 KB           build\static\js\6.885ddae2.chunk.js
  82.73 KB           build\static\js\3.c51093e0.chunk.js
  38.52 KB (-3 B)    build\static\css\4.a9e8e7ee.chunk.css
  36.53 KB (+2 B)    build\static\js\4.fc3fca00.chunk.js
  9.49 KB            build\static\js\10.689d2aaf.chunk.js
  5.49 KB (+2 B)     build\static\js\5.b40e9828.chunk.js
  2.49 KB (-1 B)     build\static\js\8.b1d97aab.chunk.js
  2.49 KB            build\static\css\5.696a8c40.chunk.css
  1.87 KB            build\static\css\6.ca756f56.chunk.css
  1.82 KB (-2 B)     build\static\js\7.1e19dfc6.chunk.js
  1.61 KB (-1 B)     build\static\js\runtime-main.8b2e4b11.js
  746 B              build\static\js\9.32c6098f.chunk.js
  534 B              build\static\css\8.c94bd734.chunk.css
  394 B              build\static\css\9.9c93afa3.chunk.css
  335 B              build\static\css\7.fab4d1ec.chunk.css
```

Layouts of the chunks
sparkle-venue-page-02


0xdevalias:
@goran-peoski-work Should this PR be targeting #1620 rather than staging? Since it seems to be sub-PR.

goran-peoski-work:
> @goran-peoski-work Should this PR be targeting #1620 rather than staging? Since it seems to be sub-PR.

Yeah, it may be. Right now as a draft PR I only use it as a deployment method. Will figure out where when/if it matures to be a proper PR.

0xdevalias (changes requested):
image

I expect you may need to rebase and prune out some merges from staging/similar..? There seem to be a whole bunch of unrelated changes in this PR.

0xdevalias:
> `React.lazy` loading components lessens the initial bundle size and blocking of subsequent requests
> and moves those chunks after the data pull and preload of party map assets.
> 
> `<Suspense>` could use simplified div elements with probably inline styles
> to provide the illusion of the components being there before actual load,
> but that's OOS to performance and will not help reduce time to first interaction.
> That may also require inlining css and fonts.

@goran-peoski-work What was lazy loaded/where?

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -192,15 +196,11 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
...
-            {foundRooms}
-            {foundEvents}
-            {foundUsers}
-          </div>
-        )}
+        <div className="NavSearchBar__search-results">
+          {isRoomsLoading ? <Loading /> : foundRooms}
+          {isEventsLoading ? <Loading /> : foundEvents}
+          {foundUsers}
+        </div>
```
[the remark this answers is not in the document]
goran-peoski-work: Forgot I had changed this. I don't think it will add to performance. Originally only displayed one spinner until both rooms and events were available results. Now it displays both spinners, then one, then none. Should probably revert it to the original.

public/index.html
```diff
@@ -68,13 +68,15 @@
...
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <title>Sparkle</title>
   </head>
 
   <body>
     <noscript>You need to enable JavaScript to use Sparkle.</noscript>
-    <div id="root"></div>
+    <div id="root">
+      Looks like you haven't got our latest code loaded yet. Please wait while we fetch this for the first time.
+    </div>
```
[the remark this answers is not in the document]
goran-peoski-work: It is OOS, and I had reverted it in subsequent commit https://github.com/sparkletown/sparkle/pull/1713/commits/3233d1267578632235f91d057f27f4f4cb7b90a9

public/index.html
```diff
@@ -68,13 +68,15 @@
     <link
       href="https://fonts.googleapis.com/css2?family=Rubik:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700&display=swap"
       rel="stylesheet"
-    />
+  />
```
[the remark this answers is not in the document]
goran-peoski-work: IDE autoformat, was reverted in subsequent cleanup commit https://github.com/sparkletown/sparkle/pull/1713/commits/3233d1267578632235f91d057f27f4f4cb7b90a9

src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -16,7 +16,7 @@ import { VenueAdminPage } from "pages/Admin/Venue/VenueAdminPage";
 
 import { VenueEntrancePage } from "pages/VenueEntrancePage";
 import { VenueLandingPage } from "pages/VenueLandingPage";
-import { VenuePage } from "pages/VenuePage";
+import { VenuePage } from "pages/VenuePage/VenuePage";
```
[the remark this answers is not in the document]
goran-peoski-work: I had removed the extra `index.tsx`, a single line file

`export { VenuePage } from "./VenuePage";`

goran-peoski-work:
> @goran-peoski-work What was lazy loaded/where?

Used `React.lazy()` and `<Suspense>` for components

imported by `VenuePage.tsx`:

-  `CountDown`
-  `LoadingPage`
-  `TemplateWrapper`

imported by `TemplateWrapper.tsx`:
- `PartyMap`
- `ChatSidebar`
- `UserProfileModal`
- `WithNavigationBar`
- `AnnouncementMessage`

goran-peoski-work:
> What happens to a preloaded asset that can’t be cached?

As it turns out, it will either be used by the browser in the first few seconds while it builds up the DOM or there will be a console message that the asset was preloaded but wasn't used in a timely manner.

This has helped in testing which assets are necessary (party map background, room images) and which aren't (global font, app icon).

> preloading is only going to be of any value when the asset is cacheable.
From the above, it turns out, browsers determine preloading to be viable only if it speeds up displaying the page, minimizing re-rendering and shortening the time for first interaction by user regardless of caching.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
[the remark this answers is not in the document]
goran-peoski-work: Reverted in bdb277f6eb41e3a42e9b3b1383c20c03e2e59adc

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -62,7 +62,11 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
   const [selectedEvent, setSelectedEvent] = useState<WithVenueId<VenueEvent>>();
   const hideEventModal = useCallback(() => setSelectedEvent(undefined), []);
 
-  const { isLoading, relatedVenues, relatedVenueIds } = useRelatedVenues({
+  const {
+    isLoading: isRoomsLoading,
```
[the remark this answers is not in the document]
goran-peoski-work: Reverted in bdb277f6eb41e3a42e9b3b1383c20c03e2e59adc no performance need to change NavSearchBar

sofisparkle:
So @mike-lvov slimmed this baby down, @sunny-viktoryia is working on the review

mike-lvov:
@goran-peoski-work @sofisparkle And reason to not merge it into my #1620 ?

0xdevalias:
> @goran-peoski-work @sofisparkle And reason to not merge it into my #1620 and review it as one?

None that i'm aware of.

functions/venue.js
```diff
@@ -2,6 +2,7 @@ const admin = require("firebase-admin");
 const functions = require("firebase-functions");
 const { HttpsError } = require("firebase-functions/lib/providers/https");
 
+const { assertValidVenueId } = require("./src/utils/assert");
 const { checkAuth } = require("./src/utils/assert");
```
0xdevalias, line 6:
```suggestion
const { assertValidVenueId, checkAuth } = require("./src/utils/assert");
```

functions/venue.js
```diff
@@ -543,6 +544,41 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
...
+    .get();
+
+  return {
+    venue: Object.assign(venueDocumentSnapshot.data(), {
+      id: venueDocumentSnapshot.id,
+    }),
+    events: eventsQuerySnapshot.docs.map((doc) =>
+      Object.assign(doc.data(), { id: doc.id })
+    ),
+  };
```
0xdevalias, line 579: Why was this changed from my original implementation? And why is it using `Object.assign` here? We use ES6+ spread syntax.

I don't think bundling the events here is a good idea. The whole point of this function (if even still needed) is to get the base venue config as soon as possible, not the events. And further.. the events probably shouldn't be getting exposed without auth anyways.

If you're going to keep this function, please go basic to the original basic type I had it as explicitly.

functions/venue.js
```diff
@@ -543,6 +544,41 @@ exports.toggleDustStorm = functions.https.onCall(async (_data, context) => {
   }
 });
 
+exports.getVenue = functions.https.onCall(async (data) => {
```
0xdevalias, line 547: Is this even needed anymore after the other changes? No point having a hacky 'fallback' if the standard way works fine.

package.json
```diff
@@ -53,6 +54,7 @@
     "react-dnd": "^11.1.3",
     "react-dnd-html5-backend": "^11.1.3",
     "react-dom": "^16.13.1",
+    "react-helmet": "^6.1.0",
```
0xdevalias, line 57: [note] This is used to preload various assets I believe.

src/pages/Admin/Admin.tsx
```diff
@@ -456,22 +459,24 @@ const Admin: React.FC = () => {
...
-            <VenueList selectedVenueId={venueId} roomIndex={queryRoomIndex} />
-          </div>
-          <div className="page-container-adminpanel">
-            {venueId ? (
-              <VenueDetails venueId={venueId} roomIndex={queryRoomIndex} />
-            ) : (
-              <>Select a venue to see its details</>
-            )}
+    <RelatedVenuesProvider venueId={venueId}>
+      <WithNavigationBar fullscreen>
```
0xdevalias, line 463: What do we need related venues for here in the admin panel? I wouldn't expect it should be needed at all.

My guess is that this was added due to something in `WithNavigationBar` needing it. But depending on what that is, I expect it ALSO probably has no place being in the admin panel. The navigation bar in the admin panel should be SUPER slim/basic, and I know @denisdimitrov had a PR for refactoring this to slim it right down. No idea if that ever got merged or if it's stuck in 'PR rot hell'.

src/pages/Admin/VenueList/VenueList.tsx
```diff
@@ -28,30 +30,34 @@ const VenueList: React.FC<VenueListProps> = ({
...
-            )}
-          </li>
-        ))}
+        {isRelatedVenuesLoading ? (
+          <Loading />
+        ) : (
+          relatedVenues.map((venue, index) => (
+            <li
+              key={index}
+              className={`${selectedVenueId === venue.id ? "selected" : ""} ${
```
0xdevalias, line 39: This whole chunk should be extracted into a `useMemo`'d section, then should use our 'guard pattern' to return `<Loading />` when it's not yet ready.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -15,18 +16,39 @@ import { ConversationSpace } from "components/templates/ConversationSpace";
...
+  import("components/organisms/WithNavigationBar").then((m) => ({
+    default: m.WithNavigationBar,
+  }))
+);
+
+const AnnouncementMessage = React.lazy(() =>
+  import("components/molecules/AnnouncementMessage").then((m) => ({
+    default: m.AnnouncementMessage,
+  }))
+);
```
0xdevalias, line 51: Why were these particular components split with lazy loading? I'm not sure that they all make sense to.

Further, there will be conflicts with a bunch of these and the work I did on bundle splitting in:

- https://github.com/sparkletown/sparkle/pull/1737

So I would potentially suggest reverting these changes.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -57,7 +79,11 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
 
     case VenueTemplate.partymap:
     case VenueTemplate.themecamp:
-      template = <PartyMap venue={venue} />;
+      template = (
+        <Suspense fallback={<></>}>
+          <PartyMap venue={venue} />
+        </Suspense>
+      );
```
0xdevalias, line 86: As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -144,14 +170,27 @@ const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
...
+            <Suspense fallback={<></>}>
+              <ChatSidebar venue={venue} />
+            </Suspense>
+            <Suspense fallback={<></>}>
+              <UserProfileModal venue={venue} />
+            </Suspense>
+          </WithNavigationBar>
+        </Suspense>
+      </ReactionsProvider>
+    </RelatedVenuesProvider>
```
0xdevalias, line 193: As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -38,25 +42,105 @@ import { useVenueId } from "hooks/useVenueId";
...
+  import("components/molecules/CountDown").then((m) => ({
+    default: m.CountDown,
+  }))
+);
+const LoadingPage = React.lazy(() =>
+  import("components/molecules/LoadingPage/LoadingPage").then((m) => ({
+    default: m.LoadingPage,
+  }))
+);
+const TemplateWrapper = React.lazy(() => import("./TemplateWrapper"));
```
0xdevalias, line 61: Lazy loading `LoadingPage` doesn't seem to make a lot of sense to me..?

Also, as per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -38,25 +42,105 @@ import { useVenueId } from "hooks/useVenueId";
...
+        (room) =>
+          room?.image_url && (
+            <link
+              key={room?.image_url}
+              href={room.image_url}
+              rel="preload"
+              as="image"
+            />
+          )
+      )
```
0xdevalias, line 93: If this stays it would need to be `useMemo`'d or similar.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -38,25 +42,105 @@ import { useVenueId } from "hooks/useVenueId";
...
+    }
+  </Helmet>
+);
```
0xdevalias, line 96: Do we have any hard evidence that this extra complexity has much if any real world impact? And even moreso, that would have impact beyond what just generally setting the cache headers appropriately on the map image (and all images loaded from firebase storage)?

How does preload interact with non-cacheable images? Does it have any actual impact, or will it just have to fetch the asset again regardless?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -38,25 +42,105 @@ import { useVenueId } from "hooks/useVenueId";
...
+
 export const VenuePage: React.FC = () => {
   const venueId = useVenueId();
+  const {
+    venue: preVenue,
+    events: preEvents,
+    error: preError,
+    isLoaded: preLoaded,
+  } = usePreloaded(venueId);
+
```
0xdevalias, line 143: If this is going to be kept, then it needs to be moved out of this component file to somewhere in `hooks/*`.

As mentioned in the backend function file itself, adding events in here defeats the purpose that this function was originally created for (to get the venue config in the quickest and most minimal way possible). I don't believe we have any need to fetch the events this early. I also believe we don't want to be making a function that exposes events to unauthed users for security/privacy reasons.

After the 'slimming down' of our standard data fetching methods, do we even need this at all anymore? A fallback 'in case' isn't a good reason to take on extra code complexity/etc.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -66,12 +150,22 @@ export const VenuePage: React.FC = () => {
 
   // @debt Remove this once we replace currentVenue with currentVenueNG or similar across all descendant components
   useConnectCurrentVenue();
-  const venue = useSelector(currentVenueSelector);
-  const venueRequestStatus = useSelector(isCurrentVenueRequestedSelector);
+  const postVenue = useSelector(currentVenueSelector);
+  const postVenueLoaded = useSelector(isCurrentVenueRequestedSelector);
```
0xdevalias, line 154: Why are these renamed? There's no point having both of this and the 'fetch via backend function' code. I suggest we just stick to this code as we already had it if it's fast enough now that the excess data fetching has been slimmed down.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -66,12 +150,22 @@ export const VenuePage: React.FC = () => {
...
+
+  if (preError) {
+    console.warn("VenuePage()", "preload failed:", preError.message);
+  }
+
+  const venue = (preError ? undefined : preVenue) ?? postVenue;
+  const isVenueLoaded = (preError ? undefined : preLoaded) ?? postVenueLoaded;
+
+  const events = (preError ? undefined : preEvents) ?? postEvents;
+  const isEventLoaded = (preError ? undefined : preLoaded) ?? postEventsLoaded;
```
0xdevalias, line 168: Why are these renamed? There's no point having both of this and the 'fetch via backend function' code. I suggest we just stick to this code as we already had it if it's fast enough now that the excess data fetching has been slimmed down.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -84,11 +178,7 @@ export const VenuePage: React.FC = () => {
   const venueName = venue?.name ?? "";
   const venueTemplate = venue?.template;
 
-  const event = currentEvent?.[0];
-
-  const { isRelatedVenuesLoading } = useRelatedVenues({
-    currentVenueId: venueId,
-  });
```
0xdevalias, line 91: Why was this removed? (and/or what was it used for in the first place?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -173,20 +263,44 @@ export const VenuePage: React.FC = () => {
...
   }
 
-  if (!venue || !venueId) {
-    return <LoadingPage />;
+  if (!venueId) {
+    // this should be happening only if invalid url param
+    // no sense in displaying loading page, another message or action is needed
+    // @debt use/display proper message/action when !venueId
+    return null;
+  }
```
0xdevalias, line 275: Makes sense, but this shouldn't be `@debt`, it should be done in this PR, if even only as a hacky 1 liner 'black screen error'

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -173,20 +263,44 @@ export const VenuePage: React.FC = () => {
...
+
+  if (!venue) {
+    return (
+      <Suspense fallback={<></>}>
+        <LoadingPage />
+      </Suspense>
+    );
   }
```
0xdevalias, line 283: Lazy loading `LoadingPage` and wrapping it with `Suspense` here doesn't make any sense.

As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -173,20 +263,44 @@ export const VenuePage: React.FC = () => {
...
 
   if (!user) {
-    return <Login venue={venue} />;
+    return (
+      <>
+        <Preload venue={venue} />
+        <Login venue={venue} />
+      </>
+    );
   }
```
0xdevalias, line 292: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -173,20 +263,44 @@ export const VenuePage: React.FC = () => {
...
-    return <LoadingPage />;
+    return (
+      <>
+        <Preload venue={venue} />
+        {
+          // is it really necessary to display loading page here?
+          // <LoadingPage />
+        }
+      </>
+    );
```
0xdevalias, line 303: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -204,12 +318,25 @@ export const VenuePage: React.FC = () => {
...
-    if (eventRequestStatus && !event) {
-      return <>This event does not exist</>;
+    if (isEventLoaded && !event) {
+      return (
+        <>
+          <Preload venue={venue} />
+          This event does not exist
+        </>
+      );
     }
```
0xdevalias, line 328: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -204,12 +318,25 @@ export const VenuePage: React.FC = () => {
...
-      return <LoadingPage />;
+      // considering there is prior !venue check, venue should exist at this point
+      return (
+        <>
+          <Preload venue={venue} />
+          <Suspense fallback={<></>}>
+            <LoadingPage />
+          </Suspense>
+        </>
+      );
```
0xdevalias, line 339: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

0xdevalias: Lazy loading `LoadingPage` and wrapping it with `Suspense` here doesn't make any sense.

As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -204,12 +318,25 @@ export const VenuePage: React.FC = () => {
...
+        <>
+          <Preload venue={venue} />
+          This event does not exist
+        </>
+      );
     }
 
     if (!event || !venue || !userPurchaseHistoryRequestStatus) {
-      return <LoadingPage />;
+      // considering there is prior !venue check, venue should exist at this point
```
0xdevalias, line 331: Can probably be removed then?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -224,21 +351,41 @@ export const VenuePage: React.FC = () => {
...
+          <Suspense fallback={<></>}>
+            <CountDown
+              startUtcSeconds={event.start_utc_seconds}
+              textBeforeCountdown="Bar opens in"
+            />
+          </Suspense>
+        </>
       );
     }
   }
```
0xdevalias, line 365: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

0xdevalias: How much do we gain from lazy loading `CountDown` here? I would have thought that it's going to be so small that the extra bundle split wouldn't be worthwhile.

As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -224,21 +351,41 @@ export const VenuePage: React.FC = () => {
...
+  if (!user) {
+    return (
+      <>
+        <Preload venue={venue} />
+        <Suspense fallback={<></>}>
+          <LoadingPage />
+        </Suspense>
+      </>
+    );
   }
```
0xdevalias, line 377: If it's not needed, then remove it?

0xdevalias: At this stage i'm not convinced that `Preload` is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

0xdevalias: Lazy loading `LoadingPage` and wrapping it with `Suspense` here doesn't make any sense.

As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -224,21 +351,41 @@ export const VenuePage: React.FC = () => {
...
 
-  return <TemplateWrapper venue={venue} />;
+  return (
+    <>
+      <Preload venue={venue} />
+      <Suspense fallback={<></>}>
+        <TemplateWrapper venue={venue} />
+      </Suspense>
+    </>
+  );
```
0xdevalias, line 390: As per previous comment, this should probably be reverted due to conflicts with:

- https://github.com/sparkletown/sparkle/pull/1737

src/pages/VenuePage/index.ts
```diff
@@ -1 +0,0 @@
-export { VenuePage } from "./VenuePage";
```
0xdevalias, line 1: Why was this removed? This should be kept..

0xdevalias (changes requested):
I have a whole lot of questions and thoughts about things in here that don't seem to make sense. To the point that by the time I got to the end of the review I was wondering if we actually want to keep anything here.

So might be worth holding off on merging it back into @mike-lvov's PR until we sort out those aspects of it all so as not to clutter that PR.

public/index.html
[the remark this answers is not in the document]
0xdevalias: It would still definitely make sense to sync with @sparkletown/product with and get some actual proper content to display there though (OOS of this PR, but as a priority)

src/components/organisms/AppRouter/AppRouter.tsx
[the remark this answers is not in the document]
0xdevalias: > I had removed the extra `index.tsx`, a single line file

Why? That's our standard way of doing things, and you'll see it literally everywhere across the codebase..

package.json
0xdevalias: Though I wonder if there is a simpler way we can do it without needing to bring in a whole new dependency?

src/pages/Admin/Admin.tsx
0xdevalias: Apparently it's maybe for this? https://github.com/sparkletown/sparkle/pull/1713#discussion_r665029238

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: It will be easier if https://github.com/sparkletown/sparkle/pull/1737 lands in staging, we do a merge from there I guess.

> At this stage i'm not convinced that Preload is going to have any meaningful impact on things; and if we do decide to keep it, it shouldn't end up duplicated in so many places.

At start, I had the preload component wrap the entire `VenuePage`, but I made it wrap the inner components just so we have the flexibility remove any unnecessary during PR review. 

For `CountDown`, I can run a test with or without it, see if there's a difference. Will add the finding here.

src/pages/VenuePage/index.ts
goran-peoski-work: What is the rationale for a single line `index.ts` re-exporting from another file?

src/components/organisms/AppRouter/AppRouter.tsx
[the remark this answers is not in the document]
goran-peoski-work: > everywhere

TBH, had noticed it in only 2-3 places. Will revert.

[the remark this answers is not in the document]
goran-peoski-work: Done in 263857b00a054c005f3b98163e85beb37c656619

functions/venue.js
goran-peoski-work: Will deal with https://github.com/sparkletown/sparkle/pull/1713#discussion_r665027700 first, then if we decide to keep it, will do proper adjustments.

The rationale was to get any data that isn't specific to user (venue and events) and see if I could match it to the data retrieved with the original method, to keep it in parity of sort.

>  the events probably shouldn't be getting exposed without auth anyways.
Might be prudent if we get a definitive Yes/No on this, should we proceed. How sensitive is the data in that collection?

functions/venue.js
goran-peoski-work: Agree on the "if the standard way works fine" I'm just not ready to call that if case true.  Some pros/cons I've noticed:


PRO: Data for event does deterministically come first (Firestore connect may decide to send it in different order)
PRO: Fallback data for venue, getting it through 2 different means and using the first available data

**CON: No significant measurable performance benefits (on localhost)**

CON: Hard to memoize AJAX call without it becoming stale, memoized can be useful only on first render.
CON If not memoized, request will go more than once.

src/pages/VenuePage/index.ts
0xdevalias: Have a look across the codebase at every place we do this as our standard pattern and let me know if you still need it to be explained further.

package.json
goran-peoski-work: Helmet library added for the specific preloading reason in this PR, but in general it allows React components to add tags to the `<head>` part of the document.

functions/venue.js
0xdevalias: > CON: Hard to memoize AJAX call without it becoming stale, memoized can be useful only on first render.

Is this because we fetch it once rather than subscribing to live updates? If so, makes sense.

package.json
0xdevalias: Yup, i'm aware of the library and its usages. I'm just not sure that we need it. I expect we can do the preload stuff in a smaller/simpler way using hooks + JS APIs or similar.

src/pages/VenuePage/VenuePage.tsx
0xdevalias: > It will be easier if #1737 lands in staging, we do a merge from there I guess.

@mike-lvov Is looking at / merging all of those things now, so should be done soon I expect.

0xdevalias: > For `CountDown`, I can run a test with or without it, see if there's a difference. Will add the finding here.

Sounds good.

functions/venue.js
goran-peoski-work: Browser network tab has shown more than one request.
Sure one is right at the start, but re-render would trigger it again, and by that time the Firestore connect data should already be available (if not triggering the render)

package.json
goran-peoski-work: I mentioned the usage for completeness in this discussion. Still, if there's a way to escape  the irony of adding library to reduce bundle size, I'm in.

> I expect we can do the preload stuff in a smaller/simpler way using hooks + JS APIs or similar.

I can look into that, had noticed a `setTitle` hook being used, so I guess it should be similar.

src/pages/Admin/VenueList/VenueList.tsx
goran-peoski-work: Got example of which the "guard pattern" is? I can't seem to connect the term with the appropriate refactor.

src/pages/Admin/Admin.tsx
goran-peoski-work: I think the reason was https://github.com/sparkletown/sparkle/pull/1620/files#r655862063 . Can't remember clearly, I think the hook was there, but needed the provider

src/pages/Admin/VenueList/VenueList.tsx
0xdevalias: Where we have a if (!foo) return type of like at the start of a block/function/etc that ‘shortcuts early’ or ‘guards’ the rest of the function/block. Followed by a single empty line of white space.

As opposed to nesting everything within it/etc.

functions/venue.js
0xdevalias: Makes sense

src/pages/Admin/VenueList/VenueList.tsx
goran-peoski-work: Oh, "fail fast", "early return", those terms I've used, and I've heard some one respond once with "we don't use that defensive programming".  As "guard pattern" on the other hand,  I thought of TS type guards only.

That aside, re reasoning was to allow some of the components to not be blocked by loading. like
```
  <Link to="/admin-ng/venue/creation" className="btn btn-primary">
    Create a venue
  </Link>
  ```
  so a memo block might also work.

File: functions/venue.js (hunk 8 above)
Comment (goran-peoski-work) [reply functions/venue.js]: Removed the BE function and the code in `VenuePage` that was calling it in 0da80c02d45c99456fdaff937409dc563ca3fdb4.
Reason being, too many requests, not enough performance gain.

File: package.json (hunk 9 above)
Comment (goran-peoski-work) [reply package.json]: Using hook instead of `react-helmet` in e9da1d5788e915f6de13a7aa0315601017c75b73. 

Code is simplified, more in line with the rest app code and no extra library included i.e. no extra KBs in bundles.

File: src/pages/VenuePage/VenuePage.tsx (hunk 16 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Superseded by e9da1d5788e915f6de13a7aa0315601017c75b73 .

Hook named `usePreloadAssets` with `useEffect` inside dealing with `document.head` using plain DOM functions. Not sure if/how `useMemo` will remain relevant.

File: functions/venue.js (hunk 7 above)
Comment (goran-peoski-work) [reply functions/venue.js]: Removed entire function in 0da80c02d45c99456fdaff937409dc563ca3fdb4

File: public/index.html (hunk 2 above)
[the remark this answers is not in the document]
Comment (goran-peoski-work) [reply public/index.html]: Because of the mechanics how Lighthouse and similar tools measure first content load, just by removing that sentence, the metric jumps from 0.5 to over 3 seconds.

 Elected to keep simple "Loading..." text for "performance" reasons that will mimic the Loading screen (will look like unstyled version I guess).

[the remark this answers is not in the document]
Comment (goran-peoski-work) [reply public/index.html]: Full text message still OOS, but having any text in there is within scope I guess.

File: src/pages/VenuePage/VenuePage.tsx (hunk 16 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Can use confirmation if no more changes are needed before closing this discussion

File: src/pages/VenuePage/VenuePage.tsx (hunk 17 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Superseded with hook `usePreloadAssets` at e9da1d5788e915f6de13a7aa0315601017c75b73 as per https://github.com/sparkletown/sparkle/pull/1713#discussion_r665881498 which simplifies code a bit.

On the impact side, tested with venues having half a dozen rooms, no browser messages complaining assets got loaded but weren't used for the room images. Using the hook should also not have issues with similar number of rooms.

 Haven't tested on anything like 20, 50, but it's a reasonable assumption that many requests might have impact.

File: src/pages/VenuePage/VenuePage.tsx (hunk 18 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Removed entire logic and replaced with hook (`src/hooks/usePreloadAssets.ts`) that doesn't use Helmet library in e9da1d5788e915f6de13a7aa0315601017c75b73

Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Removed extra requests by removing the BE function in 0da80c02d45c99456fdaff937409dc563ca3fdb4
No more extra fallback data.

File: src/pages/VenuePage/VenuePage.tsx (hunk 22 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx:197]: Asked product for proper message, can always duplicate the original Loading... if we need something quick.

File: src/pages/VenuePage/index.ts (hunk 32 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/index.ts]: Not questioning the rationale of following a standard, even if implicit. Asking for reasoning behind having `index.ts` re-export in terms of performance impact with Lazy loading of components and routes.

Comment (goran-peoski-work) [reply src/pages/VenuePage/index.ts]: Restored file in 263857b00a054c005f3b98163e85beb37c656619

**Comment by sunny-viktoryia** [comment]
@goran-peoski-work, please, update the branch from the base one and resolve the conflicts.

And as a reminder: if you fix any of the comments from the previous review, add a comment to the discussion with the commit URL where it was fixed, and then resolve the discussion, so that it could be visible if the PR is ready for the next review cycle.

(also, removing @sofisparkle from the reviewers list so that you could add her again when the PR is ready again)

File: src/pages/VenuePage/VenuePage.tsx (hunk 29 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: `<CountDown>` didn't show noticeable difference in performance between eager and lazy loading in normal venue flow. didn't know how to simulate the event starting soon condition, assuming it's a high load case as well..

File: src/pages/VenuePage/VenuePage.tsx (hunk 30 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Removed the second `!user`  in https://github.com/sparkletown/sparkle/pull/1713/commits/b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e.

First one is enough.

File: src/pages/VenuePage/VenuePage.tsx (hunk 31 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Preload is now done with a hook for `VenuePage` and removing the lazy load from `TemplateWrapper` showed increase of _Cumulative Layout Shift_ metric from the range of **0.01** - **0.02** to the higher order of magnitude **0.1** .

I'm assuming this is because eager `TemplateWrapper`  shows up too early and resizes  after the map background gets fully loaded.

Electing to keep it lazy in here.

Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Other lazy loaded components (especially in `TemplateWrapper`) are with b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e in line with staging now.

File: src/pages/VenuePage/VenuePage.tsx (hunk 28 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: done 6e1ca9a6086f3a008b5e180b5c0f676ffc041046

File: src/pages/VenuePage/TemplateWrapper.tsx (hunk 12 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/TemplateWrapper.tsx]: Since aforementioned PR is merged in staging, b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e should bring those component in line with it.

File: src/pages/VenuePage/TemplateWrapper.tsx (hunk 14 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/TemplateWrapper.tsx]: Since aforementioned PR is merged in staging, b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e should bring those component in line with it.

File: src/pages/VenuePage/VenuePage.tsx (hunk 19 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: For testing purposes. Performance tests back up claim we don't need new way of fetching. Restored original logic and naming with 0da80c02d45c99456fdaff937409dc563ca3fdb4

File: src/pages/VenuePage/VenuePage.tsx (hunk 20 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Same as https://github.com/sparkletown/sparkle/pull/1713#discussion_r666013755

File: src/pages/VenuePage/VenuePage.tsx (hunk 21 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Got added as part of a quick and dirty merge in GitHub UI that broke the code but allowed @mike-lvov to fix the rebase issue of https://github.com/sparkletown/sparkle/pull/1620 being behind staging.

Removed afterwards in 7489e0216306ef845d2e58085515fe0df99b1c27

File: src/pages/VenuePage/VenuePage.tsx (hunk 22 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx:210]: OOS for performance reasons, UX issue to be resolved in latter PR. Reverting to original behavior in 6e1ca9a6086f3a008b5e180b5c0f676ffc041046 with added comment
>// if !venueID is true loading page might display indefinitely, another message or action may be appropriate

File: src/pages/VenuePage/VenuePage.tsx (hunk 24 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Superseded by single hook in `VenuePage` component e9da1d5788e915f6de13a7aa0315601017c75b73

File: src/pages/VenuePage/TemplateWrapper.tsx (hunk 13 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/TemplateWrapper.tsx]: Done in b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e

File: src/pages/VenuePage/VenuePage.tsx (hunk 15 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Removed lazy load in b7ec26dc13ee3c39f4d061fe3565ff7cd02b0f3e

File: src/pages/VenuePage/VenuePage.tsx (hunk 23 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: No longer lazy loaded, removed `Suspense` wrapper f57638d5e088c8412a2d7eaf65a3a30942f29641

File: src/pages/VenuePage/VenuePage.tsx (hunk 25 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Replaced with hook in e9da1d5788e915f6de13a7aa0315601017c75b73 . Preloading does show impact in the _Cumulative Layout Shift_ metric at order of a magnitude:

- with `usePreloadAssets()` hook, cumulative layout shift is in the range of **0.01** and **0.02** (good)
- without the hook, cumulative layout shift is above **0.1** (not good, not terrible)

File: src/pages/VenuePage/VenuePage.tsx (hunk 26 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: same as https://github.com/sparkletown/sparkle/pull/1713#discussion_r666071927

File: src/pages/VenuePage/VenuePage.tsx (hunk 27 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: `Suspense` removed in 44282c08246eebc36473b0ce7cc4ae0e0f9482f7

Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Preloading still in place with a hook, due to https://github.com/sparkletown/sparkle/pull/1713#discussion_r666071927

**Comment by sofisparkle** [comment]
Also fyi, I don't have any skills to actually review hence, just asking to
be cc'd in the comments!

Sofi Lee-Henson of sparkle <http://www.sparklespaces.com/>

File: src/pages/VenuePage/VenuePage.tsx (hunk 29 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: Removed `Suspense` in https://github.com/sparkletown/sparkle/pull/1713/commits/12ac10cb717639f2354bc2b94a8a56640bde1d4b

**Comment by goran-peoski-work** [comment]
This PR is able to be squashed and merged, no conflicts to parent https://github.com/sparkletown/sparkle/pull/1620. But since  it got merge from staging and force pushed, to just see the changes relevant to this PR, you can look at branch

https://github.com/sparkletown/sparkle/commits/rnd/venue-page-load-performance

**Comment by goran-peoski-work** [comment]
@sofisparkle a ping here, for a review..

File: src/pages/VenuePage/VenuePage.tsx (hunk 17 above)
Comment (goran-peoski-work) [reply src/pages/VenuePage/VenuePage.tsx]: If there are no objections, I can resolve this conversation, otherwise, as alternative, could open PR to fix any possible issue rising from large number of rooms i.e. preload assets.

File: package.json (hunk 9 above)
Comment (0xdevalias) [reply package.json]: Nice!

File: src/pages/Admin/Admin.tsx (hunk 10 above)
Comment (0xdevalias) [reply src/pages/Admin/Admin.tsx:465]: Makes sense.

File: src/pages/VenuePage/VenuePage.tsx (hunk 16 above)
Comment (0xdevalias) [reply src/pages/VenuePage/VenuePage.tsx]: Sounds fine. If I have any more specific comments i'll make them inline in a new thread on next review.

**Comment by sunny-viktoryia** [comment]
@goran-peoski-work running your branch locally. Looks like I have errors in my console.
image

**Comment by goran-peoski-work** [comment]
@sunny-viktoryia Correct, changes introduced in staging, pulled into parent PR compounded in this sub-PR

> related to the parent PR?

Will see if I can work on fixing them here.

File: src/pages/VenuePage/VenuePage.tsx [hunk 33]
```diff
@@ -89,10 +112,6 @@ export const VenuePage: React.FC = () => {
 
   const event = currentEvent?.[0];
 
-  const { isLoading } = useRelatedVenues({
-    currentVenueId: venueId,
-  });
-
```
sunny-viktoryia, line 95: haven't checked other comments. Can you explain why is that changed? Don't we need the `isLoading` anymore here?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -181,6 +200,7 @@ export const VenuePage: React.FC = () => {
   }
 
   if (!venue || !venueId) {
+    // if !venueId is true loading page might display indefinitely, another message or action may be appropriate
```
sunny-viktoryia, line 203: have you had that behavior? looks like `venueId` is taken from the url. And if there is no `venueId` it gets redirected to https://sparklespaces.com/ (at least this is how it works for me)

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -72,6 +86,15 @@ export const VenuePage: React.FC = () => {
...
   const venueRequestStatus = useSelector(isCurrentVenueRequestedSelector);
 
+  usePreloadAssets(
+    [
+      venue?.mapBackgroundImageUrl,
+      ...(venue?.rooms ?? []).map((room) => room?.image_url),
+    ]
+      .filter(isDefined)
+      .map((url) => ({ url }))
+  );
```
sunny-viktoryia, line 96: you should `useMemo` here as it created a new array every time and will trigger `useEffect` of the hook.
```suggestion
  const assetsToPreload = useMemo(
    () =>
      [
        venue?.mapBackgroundImageUrl,
        ...(venue?.rooms ?? []).map((room) => room?.image_url),
      ]
        .filter(isDefined)
        .map((url) => ({ url })),
    [venue]
  );

  usePreloadAssets(assetsToPreload);
```

src/hooks/usePreloadAssets.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useRef } from "react";
+import { isDefined } from "utils/types";
```
sunny-viktoryia, line 2:
```suggestion
import { useEffect, useRef } from "react";

import { isDefined } from "utils/types";
```

sunny-viktoryia (changes requested):
requesting few changes. Once fixed should be good to be merged. 🙃

src/pages/VenuePage/VenuePage.tsx
0xdevalias: @goran-peoski-work Can you please prefix that comment with `@debt`

src/pages/VenuePage/VenuePage.tsx
0xdevalias: This is likely just because the map becomes available earlier, which we can solve better by properly setting the expected width/height on it ahead of time.

OOS of this PR though.

src/pages/VenuePage/VenuePage.tsx
0xdevalias: Can you please show the explicit stats on what having this lazy loaded or not does for our bundle metrics as per the screenshots/etc I provided in my 'split' PRs.

I'm still not convinced that this is going to be of any major benefit, and IMO the fact that the CLS changes is more of a side effect of a race condition for a problem we can fix better in future by explicitly setting the appropriate size on the map image/container ahead of time.

src/pages/VenuePage/index.ts
0xdevalias: > Asking for reasoning behind having index.ts re-export in terms of performance impact with Lazy loading of components and routes.

In terms of performance impact, there should be zero impact/difference/etc, particularly given all of the code is bundled together, etc. This is more of a dev/coding standard side of things than a runtime relevant thing.

0xdevalias:
> This PR is able to be squashed and merged, no conflicts to parent #1620

@goran-peoski-work FYI, when merging this to #1620 don't squash merge it. We don't squash merge sub-PRs.

0xdevalias:
> Will see if I can work on fixing them here.

@goran-peoski-work Definitely don't fix them in this PR. OOS. Standalone fix to staging is the correct place, then can update this branch with staging once it's merged (assuming it merges first)

src/pages/VenuePage/VenuePage.tsx
0xdevalias: @sunny-viktoryia This is kind of a guard for a situation that is unlikely if not impossible in our current setup, but could potentially become an issue if the routes/implementation we use to get the `venueId`/etc changes.

There are some patterns I used in my recent `src/pages/Account/Questions.tsx` refactor that solve this need in a better way. I linked them on slack to @goran-peoski-work, it may be worth just adding that here to solve this potential edge case once and for all?

src/pages/VenuePage/VenuePage.tsx
0xdevalias: I haven't looked at the definition of `usePreloadAssets`, but why do we `.map((url) => ({ url })),` map to an object with `url` rather than just a `string[]`?

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: This component doesn't use the related venues, so it makes no sense to show loading screen for children components. They can use `useRelatedVenues()` themselves the data and the flag and display their own loading message or icon.

goran-peoski-work:
> Definitely don't fix them in this PR. OOS.

Some of those staging -> PR -> sub-PR changes overwrote the work in `VenuePage` this sub-PR was supposed to do. I dealt with those errors in merge.

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: Maybe next PR, a smaller change, faster to review. I'm hesitant of adding more things in this one., that might be OOS.

I put a comment that would be able to pinpoint to go back and replace with proper message/action, once/if product/tech decide course.

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: `usePreloadAssets()` allows for preloading different kinds of assets, just defaults to `image`

```
export interface PreloadAsset {
  url: string;
  as?: string;
  type?: string;
}
```

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: The map and rooms are the first things user interacts with so loading those images sooner helps , but the most notable difference is the cumulative layout shift, it goes above the threshold of 0.1 seconds (the next one is 0.25)

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: done https://github.com/sparkletown/sparkle/pull/1713/commits/923a6499ed118c310a50b27bd029f207cdad8701

goran-peoski-work: done https://github.com/sparkletown/sparkle/pull/1713/commits/923a6499ed118c310a50b27bd029f207cdad8701

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: > IMO the fact that the CLS changes is more of a side effect of a race condition for a problem we can fix better in future by explicitly setting the appropriate size on the map image/container ahead of time.

Maybe side effect of lazy loading, perceived load is still improved with it. Should there be PR in the future that improves upon it (don't know if I'd call async loading of assets race condition due to connotation) that PR can revert the lazy loading of components as well.

This PR is about load performance of `VenuePage` so if it is lazy or not, once it gets loaded, that performance is of scope here.

goran-peoski-work: Here is the table I've compiled
sparkle-preload-assets-01

src/pages/VenuePage/VenuePage.tsx
goran-peoski-work: Accepted changes https://github.com/sparkletown/sparkle/pull/1713/commits/660c49d3e8c725ff2f62c860c8b9c86b326cf451

sunny-viktoryia (approved):
Looks good to me. 🎉

src/pages/VenuePage/VenuePage.tsx
0xdevalias: My point here is about not adding arbitrary bundle splits without proving their relevance and showing the impact they have..

PR #1721, opened 2021-07-05
refactor ChatMessageBox to extract EmojiPicker component + lazy load it

0xdevalias:
Refactored `ChatMessageBox` to extract `EmojiPicker` component + lazy load it with `React.lazy` / `Suspense`.

---

## Results

While the total bundle size remains basically the same, we've split the `emoji-mart` dependencies into a separate chunk, that will be lazy loaded when the user attempts to use the emoji picker in the chat.

In this recording, we first load most of the website at normal speed, then artificially slow the network speed down to 'Slow 3G' to demonstrate better. We open the chat sidebar, and then click on the emoji chooser. It's only once we first click this that the dependency is lazy loaded, which we can see by the 'loading spinner'. Once this has been loaded once, the emoji chooser can be opened and closed without needing to re-fetch the dependency. This chunk will also be individually cached, so that user's should only ever have to re-download it when the `emoji-mart` dependency changes.

https://user-images.githubusercontent.com/753891/124449552-eaf32b00-ddc6-11eb-9a30-56d8b5540410.mov

### Before Refactor

```
Compiled successfully.

File sizes after gzip:

  808.44 KB  build/static/js/2.9accd122.chunk.js
  133.45 KB  build/static/css/main.518bab14.chunk.css
  129.21 KB  build/static/js/main.651919bf.chunk.js
  9.49 KB    build/static/js/3.aca06453.chunk.js
  1.17 KB    build/static/js/runtime-main.b20c0baa.js
```

image

image

### After Refactor

```
Compiled successfully.

File sizes after gzip:

  715.33 KB (-93.11 KB)  build/static/js/2.5144d5eb.chunk.js
  133.45 KB (-1 B)       build/static/css/main.b3e46173.chunk.css
  129.34 KB (+142 B)     build/static/js/main.eae025ed.chunk.js
  92.87 KB (+83.38 KB)   build/static/js/3.25aaa1c1.chunk.js
  9.48 KB                build/static/js/4.2934e0fe.chunk.js
  1.18 KB (+11 B)        build/static/js/runtime-main.0bbf9a51.js
```

image

image


src/components/molecules/EmojiPicker/EmojiPicker.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import "emoji-mart/css/emoji-mart.css";
+
+const Picker = lazy(() =>
+  tracePromise("EmojiPicker::lazy-import::emoji-mart", () =>
+    import("emoji-mart").then(({ Picker }) => ({
+      default: Picker,
+    }))
+  )
+);
```
denisdimitrov, line 17: Is it possible to transform the lazy load logic into a HOC or hook and just reuse it?

0xdevalias: Probably OOS of this PR.

I tried playing around with extracting a util/similar for it but was running into some painful if not impossible aspects with the typing.

My ideal would be to pass in just the name of the module and the thing we want to extract from it, but that didn’t seem to work.

It might be possible to provide a () => import(‘foo’) and then a key to extract from that.. shall give it another try this morning and see how that turns out.

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
```diff
@@ -85,8 +86,8 @@ export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
   } = useShowHide();
 
   const addEmoji = useCallback(
-    (emoji: BaseEmoji) => {
-      if (emoji.native) {
+    (emoji: EmojiData) => {
+      if ("native" in emoji && emoji.native) {
```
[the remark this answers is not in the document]
0xdevalias: Because TypeScript says no.

emoji is a union of 2 types, and only one of them has native defined on it. So this is a way of narrowing the types and telling typescript to only run this on the type that has it. The second part of the check is probably irrelevant I think, but it’s the logic that existed previously so didn’t want to deal with figuring out any potential nuances of changing it.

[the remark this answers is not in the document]
0xdevalias: Note that i'm pretty certain I tried `if (emoji?.native)` before figuring out this method, and it doesn't work because the 2 types within `EmojiData` don't both have the `native` field as part of a common 'base type' (which is what we do in a few cases, and is how we have tended to work around/avoid a situation like this in our own code, rightly or wrongly)

src/components/molecules/EmojiPicker/EmojiPicker.tsx
0xdevalias: Well.. here is my attempt..

```typescript
import { ComponentType, LazyExoticComponent, lazy } from "react";

import { tracePromise } from "./performance";

export const lazyImportComponent = <
  T extends Record<string, ComponentType<U>>,
  U
>(
  lazyImport: () => Promise<T>,
  component: keyof T,
  traceName: string
): LazyExoticComponent<T[keyof T]> =>
  lazy(() =>
    tracePromise(traceName, () =>
      lazyImport().then((module) => {
        if (!(component in module))
          throw new Error("component not found in module");

        return { default: module[component] };
      })
    )
  );
```

Which should be usable like this:

```typescript
import { lazyImportComponent } from "utils/import";

const Picker = lazyImportComponent(() => import("emoji-mart"), "Picker", "TODO");
```

Except TypeScript says no.. so if you can figure out solving this.. we can do it:
image

I believe the issue is in this part of it most likely:

```typescript
T extends Record<string, ComponentType<U>>,
```

0xdevalias: Slack thread for figuring it out in future. Calling it OOS of this PR.

- https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625569313087900

denisdimitrov (approved):
LGTM!

src/components/molecules/EmojiPicker/EmojiPicker.tsx
denisdimitrov: > Except TypeScript says no.. so if you can figure out solving this.. we can do it:

I can give it a try tonight, if I have some free time.

goran-peoski-work: > Is it possible to transform the lazy load logic into a HOC or hook and just reuse it?

The first time I tried it, it worked. I committed it to a branch. Afterwards, I merged it into another branch and suddenly the same code was throwing compiler errors.

Last time I had seen something similar was an Angular router (same problem basically, lazy loading) that did static analysis of the code, verbatim, regardless the dynamic result is correct.

Agree on OOS

0xdevalias: Theoretically figured it out today. Will test if it works on this PR, and if so, add it in:

```typescript
// TODO: move this to types/utilities
// This is copied from an answer at
//   https://stackoverflow.com/questions/54520676/in-typescript-how-to-get-the-keys-of-an-object-type-whose-values-are-of-a-given
// This uses TypeScript mapped types
//   https://www.typescriptlang.org/docs/handbook/2/mapped-types.htm
//   https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#mapped-types
// The -? removes the optional property
//   https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers
// It then maps to a conditional type
//   https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html
// That checks if the type of T[K] extends from the type of V
//   If it does, then it maps to K, otherwise it maps to never
// Finally it uses a lookup type
//   https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#keyof-and-lookup-types
export type KeysMatching<T, V> = {
  [K in keyof T]-?: T[K] extends V ? K : never;
}[keyof T];

export const lazyImportComponent = <
  T extends Record<string, unknown>,
  U extends KeysMatching<T, ComponentType<unknown>>
>(
  lazyImport: () => Promise<T>,
  componentName: U
): LazyExoticComponent<ComponentType<T[U]>> =>
  lazy(() =>
    lazyImport().then((module) => {
      if (!(componentName in module))
        throw new Error("component not found in module");

      return { default: module[componentName] };
    })
  );

export const lazyImportComponentWithTrace = <
  T extends Record<string, unknown>,
  U extends KeysMatching<T, ComponentType<unknown>>
>(
  lazyImport: () => Promise<T>,
  componentName: U,
  traceName: string
): LazyExoticComponent<ComponentType<T[U]>> =>
  lazyImportComponent(
    () => tracePromise(traceName, () => lazyImport()),
    componentName
  );
```

0xdevalias: Nope.. still having issues, not sure why, but I expect it's probably saying that there are no exports that match `ComponentType<unknown>`. So.. probably a future problem to deal with in a standalone PR:

image

PR #1724, opened 2021-07-05
remove mixpanel, logrocket, and stripe

ermiaSparkle:

You can find updated information:

**Before**

Screenshot from 2021-07-05 12-36-53

**After**
Screenshot from 2021-07-05 14-40-33


0xdevalias (changes requested):
@ermiaSparkle Can you please do this as separate PRs for each of mixpanel, logrocket and stripe. We don't want them all together here in a single PR.

Also, why did you delete some files, but comment things out in other places?

ermiaSparkle:
@0xdevalias 
Should I delete them or comment them?

PR #1725, opened 2021-07-05
refactor useSovereignVenueId into useSovereignVenue

dmitrii-zolotukhin:
- [x] Make the change to store the venue in redux.

- [x]  Rename all sovereignVenueId actions/stores/etc into sovereignVenue.

- [x]  Replace all the places, where this hook is used to use the new version of the hook. It should return both sovereignVenueId and sovereignVenue

Fixed sparkletown/internal-sparkle-issues#882


src/hooks/useRelatedVenues.tsx
```diff
@@ -10,15 +10,15 @@ import { WithId } from "utils/id";
...
 const emptyArray: never[] = [];
 
 export interface RelatedVenuesContextState {
   isLoading: boolean;
   isError: boolean;
 
-  isSovereignVenueIdLoading: boolean;
+  isSovereignVenueLoading: boolean;
   sovereignVenue?: WithId<AnyVenue>;
   sovereignVenueId?: string;
```
mike-lvov, line 23: Please, also refactor RelatedVenues to not fetch `sovereignVenue` twice.

image

You should pass sovereignVenue there

mike-lvov (changes requested):

src/hooks/useRelatedVenues.tsx
dmitrii-zolotukhin: @mike-lvov fixed in c0aa08769265d8bf04074054cae05bac8a0bd2db

src/hooks/useRelatedVenues.tsx
```diff
@@ -59,19 +60,17 @@ export const RelatedVenuesProvider: React.FC<RelatedVenuesProviderProps> = ({
...
       {
-        attributes: {
-          sovereignVenueId: sovereignVenueId,
-        },
+        attributes: { sovereignVenueId },
         withDebugLog: true,
       }
     );
-  }, [sovereignVenueId]);
+  }, [sovereignVenueId, sovereignVenue]);
```
mike-lvov, line 73: remove sovereignVenueId from the deps and use only sovereignVenue please

mike-lvov (changes requested):

src/hooks/useRelatedVenues.tsx
dmitrii-zolotukhin: @mike-lvov i wanted to do it in that way but there appears an error. Do I need to correct the types tracePromise?
Screenshot 2021-07-05 at 18 09 00

mike-lvov: could you not export sovereignVenueId and use the id inside since it's anyway used in the fn

src/hooks/useRelatedVenues.tsx
0xdevalias: That depends on the logic of your hook actually.. because my code here will work regardless of the venueId provided. Whereas unless you changed the logic, you cache the very first sovereign venue ever used and never change it.

src/hooks/useRelatedVenues.tsx
0xdevalias: Why do we want to change this? Fetch related venues should only receive the id, the attributes should only have the id, this particular snippet shouldn’t have the full venue in it at all..

src/api/venue.ts
```diff
@@ -156,10 +156,8 @@ export const fetchDescendantVenues = async (
 };
 
 export const fetchRelatedVenues = async (
-  venueId: string
+  sovereignVenue: WithId<AnyVenue>
```
0xdevalias, line 159: This shouldn’t be changed like this

src/hooks/useRelatedVenues.tsx
0xdevalias: Also, why would we change fetch related venues from being something generic that can be provided any venueId and work properly to requiring a full sovereign venue object to be passed into it? Seems like degraded functionality/not a great plan.

src/hooks/useRelatedVenues.tsx
0xdevalias: The solution to that typing bug is to read the id off the sovereignVenue

src/hooks/useSovereignVenue.ts
```diff
@@ -0,0 +1,78 @@
...
+  UseSovereignVenueProps,
+  UseSovereignVenueData
+> = ({ venueId }) => {
+  const dispatch = useDispatch();
+
+  const {
+    venue: sovereignVenue,
+    isLoading: isSovereignVenueLoading,
+    errorMsg,
+  } = useSelector(sovereignVenueSelector);
```
0xdevalias, line 42: Add , isEqual) to the end of this useSelector, otherwise it won’t memo correctly due to the deeply nested structure of venues

0xdevalias: Lodash’s isEqual

src/hooks/useSovereignVenue.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  useEffect(() => {
+    // NOTE: Force to fetch it only once
```
0xdevalias, line 45: OOS, but this is something I don’t really like.. and we should really find a better solution for.

we want to only fetch it once for a ‘known venueId’ or any related venues. But if we ‘step outside’ that set of related venues, we absolutely need to find the new one.

This will cause problems when we switch to using only router based navigation and need to step across the boundary of related venues, which we already do in env/github

src/store/reducers/SovereignVenue.ts
```diff
@@ -3,8 +3,12 @@ import {
   SovereignVenueActionTypes,
 } from "store/actions/SovereignVenue";
 
+import { AnyVenue } from "types/venues";
+
+import { WithId } from "utils/id";
+
 export type SovereignVenueState = {
-  id?: string;
+  venue?: WithId<AnyVenue>;
```
0xdevalias, line 11:
```suggestion
  SovereignVenue?: WithId<AnyVenue>;
```

src/hooks/useRelatedVenues.tsx
mike-lvov: > That depends on the logic of your hook actually.. because my code here will work regardless of the venueId provided. Whereas unless you changed the logic, you cache the very first sovereign venue ever used and never change it.
> 
> So changing this like that is probably a regression in functionality.

Disagree. If sovereignVenue is not working correctly, the app will be broken anyway. But with this change results in the performance gains by not fetching the sovereignVenue. Since you are using `await` there, it won't start fetching the descendants before fetching the sovereignVenue

> Also, why would we change fetch related venues from being something generic that can be provided any venueId and work properly to requiring a full sovereign venue object to be passed into it? Seems like degraded functionality/not a great plan.

It's only used in relatedVenues. I guess with sovereignVenue available, we can just use fetching the descendants though. 

> Whereas unless you changed the logic, you cache the very first sovereign venue ever used and never change it.

But you cache it too? You only fetch once given that it's in the context?

src/hooks/useRelatedVenues.tsx
mike-lvov: > Why do we want to change this? Fetch related venues should only receive the id, the attributes should only have the id, this particular snippet shouldn’t have the full venue in it at all..

My intention was not to double fetch the sovereignVenue since we already have it in place. It would improve the performance especially for the lower-level venues

src/hooks/useRelatedVenues.tsx
0xdevalias: > But you cache it too? You only fetch once given that it's in the context?

Then that should also be fixed.

src/hooks/useRelatedVenues.tsx
0xdevalias: This change makes what is currently an all in one wrapped up good generic function, dependendent on having an entire venue object available to it and passed as a prop.. that's not a good way to do things.. to say the least.

src/hooks/useRelatedVenues.tsx
mike-lvov: Given the above, my suggestion would be to revert the `fetchRelatedVenues` prev. functionality and to use `fetchDescendantVenues` in the `RelatedVenuesProvider`

0xdevalias: > my suggestion would be to revert to the fetchRelatedVenues prev. functionality

What does that mean?

mike-lvov: @0xdevalias 
image

src/api/venue.ts
mike-lvov: https://github.com/sparkletown/sparkle/pull/1725#discussion_r664265175

src/hooks/useRelatedVenues.tsx
0xdevalias: Makes sense.

mike-lvov: So, you agree with using `fetchDescendantVenues` in the `RelatedVenuesProvider`?

0xdevalias: Seems fine to me, so long as it's properly combined and the array object from it memo'd/etc (not constructed inline in the function in a way that will change reference every time it's called)

mike-lvov: @0xdevalias IMO it'll be combined in the useAsync fn to prevent any reference changes

0xdevalias: Sounds good.

0xdevalias: Actually.. that risks re-running the `useAsync` and fetching everything all over again unless we're super careful.. I would probably suggest not including it there. Using deeply nested objects in react's deps is prime for issues unless everything is done perfectly.

src/hooks/useRelatedVenues.tsx
dmitrii-zolotukhin: @0xdevalias @mike-lvov fixed in 98b0388a5f7e0fb23799f214e5b588b02eaf1fa0. If I understood correctly, I needed to set sovereignVenueId instead of sovereignVenue.id  in tracePromise function, I did so.

src/api/venue.ts
dmitrii-zolotukhin: @0xdevalias @mike-lvov fixed in 98b0388

src/hooks/useSovereignVenue.ts
dmitrii-zolotukhin: @0xdevalias fixed in 98b0388

src/hooks/useSovereignVenue.ts
dmitrii-zolotukhin: @0xdevalias fixed in 98b0388

src/store/reducers/SovereignVenue.ts
dmitrii-zolotukhin: @0xdevalias fixed in 98b0388

src/hooks/useRelatedVenues.tsx
dmitrii-zolotukhin: @0xdevalias @mike-lvov fixed in 98b0388

0xdevalias: @dmitrii-zolotukhin Wasn't @mike-lvov 's suggestion to reuse the sovereign venue from the hook, not to fetch it again here? As currently implemented in your recent change here it's the same problem as before, but made worse through more verbose/inefficient code.

dmitrii-zolotukhin: @0xdevalias @mike-lvov fixed in 2bcf61bbd342470f31bc588e68410eeb6a247708

mike-lvov: @dmitrii-zolotukhin nope, not it. I'll explain it to you in a call

src/hooks/useSovereignVenue.ts
mike-lvov: What was fixed in the commit? @dmitrii-zolotukhin

mike-lvov (approved):
LGTM

PR #1726, opened 2021-07-05
refactor RenderMarkdown to extract RenderMarkdownInner component + lazy load it

0xdevalias:
Refactored `RenderMarkdown` to extract `RenderMarkdownInner` component + lazy load it with `React.lazy` / `Suspense`.

This has the effect of splitting all of the markdown rendering dependencies/plugins/etc (eg. `react-markdown`, `remark-gfm`, `remark-emoji`, `remark-external-links`, `rehype-sanitize`) into their own bundle chunk.

I pass through the raw `text` as the fallback for `Suspense`, so even before the markdown rendering chunk is loaded, the text can still be viewed in its raw form, allowing the site to still be usable even while the chunk is loading.

---

partially fixes sparkletown/internal-sparkle-issues#839

---

## Results

The overall bundle size only increases ever so slightly (from `3.49 MB` to `3.5 MB`), but we end up splitting off a new chunk of `202.07 KB`, reducing chunk 2 from `2.98 MB` to `2.78 MB`.

### Before Refactor

```
Compiled successfully.

File sizes after gzip:

  757.37 KB  build/static/js/2.0ff36f07.chunk.js
  133.44 KB  build/static/css/main.19c8cac9.chunk.css
  129.2 KB   build/static/js/main.409bffc0.chunk.js
  9.48 KB    build/static/js/3.a560293f.chunk.js
  1.17 KB    build/static/js/runtime-main.3a521cfb.js
```

image

image

### After Refactor

```
Compiled successfully.

File sizes after gzip:

  698.71 KB (-58.66 KB)  build/static/js/2.813c8d58.chunk.js
  133.44 KB (+1 B)       build/static/css/main.6c272471.chunk.css
  131 KB (+1.8 KB)       build/static/js/main.8a2ac429.chunk.js
  58.69 KB (+49.2 KB)    build/static/js/3.e0d513cf.chunk.js
  9.48 KB                build/static/js/5.1e631716.chunk.js
  1.19 KB (+24 B)        build/static/js/runtime-main.b2055cca.js
  549 B                  build/static/js/4.a824b5b7.chunk.js
```

image

image

image

### DIFF
--- a/src/components/organisms/RenderMarkdown/RenderMarkdown.tsx
+++ b/src/components/organisms/RenderMarkdown/RenderMarkdown.tsx
@@ -1,24 +1,14 @@
-import React from "react";
+import React, { Suspense, lazy } from "react";
 
-import ReactMarkdown from "react-markdown";
-import glm from "remark-gfm";
-import emoji from "remark-emoji";
-import externalLinks from "remark-external-links";
-import sanitize from "rehype-sanitize";
+import { tracePromise } from "utils/performance";
 
-import {
-  MARKDOWN_BASIC_FORMATTING_TAGS,
-  MARKDOWN_HEADING_TAGS,
-  MARKDOWN_IMAGE_TAGS,
-  MARKDOWN_LINK_TAGS,
-  MARKDOWN_LIST_TAGS,
-  MARKDOWN_PRE_CODE_TAGS,
-} from "settings";
-
-import { isTruthy } from "utils/types";
-
-const REMARK_PLUGINS = [glm, emoji, externalLinks];
-const REHYPE_PLUGINS = [sanitize];
+const RenderMarkdownInner = lazy(() =>
+  tracePromise("RenderMarkdown::lazy-import::RenderMarkdownInner", () =>
+    import("./RenderMarkdownInner").then(({ RenderMarkdownInner }) => ({
+      default: RenderMarkdownInner,
+    }))
+  )
+);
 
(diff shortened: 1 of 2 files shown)


mike-lvov (approved):
Looks good. Out of curiosity, why is this a separate PR, not into a batch of turning imports into lazy loads?

0xdevalias:
> Out of curiosity, why is this a separate PR, not into a batch of turning imports into lazy loads?

@mike-lvov Essentially to minimize the size/scope of PRs to make them more 'individually chunked' and easier to review/roll back/etc. I also wasn't sure just how far I was going to go down that path, so figured a couple of PRs would be less 'bad' than a big PR that never gets finished/etc, or requires excess work to split out unrelated/unwanted things.

0xdevalias:
> Derp 🤦🏻 That's me rushing the change where this was introduced with (see #1726) a 🧠💤 and getting confused between the api of `RenderMarkdownInner` (which matches `RenderMarkdown`'s), and the API of `ReactMarkdown` (which passes text in the `children`.
> 
> The worst part is that I had it right originally and then changed it last second because I thought I had it wrong 😅 Should have trusted the type system saying it was fine..
> 
> Thanks for fixing this up @mike-lvov !
> 
_Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/pull/1747#pullrequestreview-701508151_

PR #1727, opened 2021-07-05
improve accessibility concerns in NavBar, NavBarSchedule, and MapRoom

sabrinaorban:
partially fixes sparkletown/internal-sparkle-issues#887

- replaced divs with buttons
- removed default style of buttons
- added style for button:focus to match the hover style

Preview link for visibility: https://co-reality-staging--preview-pr-1727-3q22ybbt.web.app/


src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -227,6 +241,9 @@ export const NavBar: React.FC<NavBarPropsType> = ({
 
               {shouldShowSchedule ? (
                 <div
+                  role="button"
+                  tabIndex={0}
+                  onKeyPress={(e) => handleKeyPress(e, toggleEventSchedule)}
```
0xdevalias, line 246: We shouldn’t need to explicitly handle key presses here if we use the correct base types that naturally support these navigations. Eg. Button

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -356,7 +375,11 @@ export const NavBar: React.FC<NavBarPropsType> = ({
 
       {/* @debt Remove back button from Navbar */}
       {hasBackButton && venue?.parentId && parentVenue?.name && (
-        <div className="back-map-btn" onClick={backToParentVenue}>
+        <div className="back-map-btn"
+        role="button"
+        tabIndex={0}
+        onKeyPress={(e) => handleKeyPress(e, backToParentVenue)}
```
0xdevalias, line 381: We shouldn’t need to explicitly handle key presses here if we use the correct base types that naturally support these navigations. Eg. Button

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,8 +114,24 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          e.preventDefault();
+          setSelectedDayIndex(dayIndex)
+        }
+       }
+
       return (
         <li
+          role="button"
+          tabIndex={0}
+          onKeyPress={(e) => handleKeyPress(e)}
```
0xdevalias, line 134: We shouldn’t need to explicitly handle key presses here if we use the correct base types that naturally support these navigations. Eg. Button within the li

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -114,8 +114,24 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+        if (enterOrSpace) {
+          e.preventDefault();
+          setSelectedDayIndex(dayIndex)
+        }
+       }
+
       return (
         <li
+          role="button"
+          tabIndex={0}
```
0xdevalias, line 133: Not sure this tabindex is going to be correct? Why setting it to 0 here?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -202,6 +218,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     });
   }, [relatedVenueEvents]);
 
+
```
0xdevalias, line 221: Excess space, please remove

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -65,6 +65,7 @@ export const MapRoom: React.FC<MapRoomProps> = ({
       !isCovertRoom && venue.roomVisibility === RoomVisibility.count,
   });
 
+
```
0xdevalias, line 68: Excess space, please remove

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -88,11 +89,28 @@ export const MapRoom: React.FC<MapRoomProps> = ({
     selectRoom();
   }, [play, selectRoom]);
 
+
```
0xdevalias, line 92: Excess space, please remove

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -88,11 +89,28 @@ export const MapRoom: React.FC<MapRoomProps> = ({
...
+ }
+
   return (
     <div
+      role="button"
+      tabIndex={0}
       className={containerClasses}
       style={roomInlineStyles}
       onClick={isCovertRoom ? noop : selectRoomWithSound}
+      onKeyPress={isCovertRoom ? noop: (e) => handleKeyPress(e, selectRoomWithSound)}
```
0xdevalias, line 113: We shouldn’t need to explicitly handle key presses here if we use the correct base types that naturally support these navigations. Eg. Button

0xdevalias (changes requested):
See my piecemeal review comments starting from https://github.com/sparkletown/sparkle/pull/1727#pullrequestreview-699430335

oleksii-front:
@0xdevalias merged with staging and ready for the re-review.

denisdimitrov:
Something looks wrong but I can't really find it in the review. Some of the colors and spacings are messed up in the Schedule.

Staging:  https://staging.sparkle.space/in/devaliashacksville
staging

This feature branch: https://co-reality-staging--preview-pr-1727-3q22ybbt.web.app/in/devaliashacksville
branch

0xdevalias:
@sabrinaorban @oleksii-front FYI: As discussed in engineering standup just now, this PR has some conflicts that will need fixing up.

jonboutelle:
@sabrinaorban can you take this over the finish line today? Client has mentioned it as a P1.

whoaitsaimz (commented):
> &:focus {
      outline: none;
    }
    
To my knowledge this is NOT good practice when accessibility is a concern. Please do not do this, outlines help w/accessibility via keyboard nav (even though it's less pretty)

src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -12,9 +12,17 @@ $label-visible: flex;
   flex-direction: column;
   align-items: center;
 
+  appearance: none;
+  -webkit-appearance: none;
```
0xdevalias, line 16: Why using vendor prefixed styles here?

src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -12,9 +12,17 @@ $label-visible: flex;
...
+  background: none;
+  border: none;
```
0xdevalias, line 18: Why are these being set to `none`?

src/components/templates/PartyMap/components/Map/MapRoom.scss
```diff
@@ -12,9 +12,17 @@ $label-visible: flex;
...
+  color: white;
+
   transition: transform 600ms $transition-function;
 
-  &:hover {
+  &:hover,
+  &:focus {
+    outline: none;
```
0xdevalias, line 25: Pretty sure removing `outline` like this creates an a11y issue, not resolves it.

src/scss/global.scss
```diff
@@ -772,3 +772,9 @@ span.link {
 .flex-one {
   flex: 1;
 }
+
+.no-outline-on-focus {
+  &:focus {
+    outline: none;
+  }
+}
```
0xdevalias, line 780: Why was this added? Removing outline on focus creates accessibility issues, not fixes them..

src/components/organisms/RadioModal/RadioModal.tsx
```diff
@@ -46,9 +46,11 @@ export const RadioModal: React.FunctionComponent<RadioModalPropsType> = ({
   );
 
   return (
-    <div className="radio-modal-container">
+    <div role="dialog" className="radio-modal-container">
       {!isRadioPlaying ? (
-        <Button onClick={handleEnableButtonClick}>Enable radio</Button>
+        <Button aria-label="Enable radio" onClick={handleEnableButtonClick}>
+          Enable radio
+        </Button>
```
0xdevalias, line 53: Do we need to use an `aria-label` when the button already wraps it's text? I would assume that is fine as is from an a11y standpoint.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -122,7 +122,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
             setSelectedDayIndex(dayIndex);
           }}
         >
-          {formatDayLabel(day)}
+          <button>{formatDayLabel(day)}</button>
```
0xdevalias, line 125: Why is this a raw `button`? This doesn't seem right. Presumably the `onClick` handler should be on this button, etc.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -68,11 +70,31 @@ $weekday--margin-right: 10px;
...
+      -webkit-appearance: none;
+      border: none;
+      background: inherit;
+      color: $white;
+
+      &:focus {
+        color: $white;
+        opacity: 1;
+        font-weight: bold;
+        outline: none;
```
0xdevalias, line 84: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -68,11 +70,31 @@ $weekday--margin-right: 10px;
...
+      }
+    }
+
     &--active {
       opacity: 1;
       background-color: $primary;
       text-shadow: 0 2px 4px opaque-black(0.3);
       box-shadow: 0 3px 6px opaque-black(0.2);
+      border: none;
```
0xdevalias, line 93: Why removing border here?

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -68,11 +70,31 @@ $weekday--margin-right: 10px;
     border-radius: $border-radius--xl;
     opacity: 0.8;
 
+    button {
+      appearance: none;
+      -webkit-appearance: none;
+      border: none;
```
0xdevalias, line 76: Why using vendor prefixed styles here? Why setting these things to `none`?

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -55,8 +55,10 @@ $weekday--margin-right: 10px;
     text-shadow: 0 2px 4px opaque-black(0.3);
     box-shadow: 0 3px 6px opaque-black(0.2);
 
-    &:hover {
+    &:hover,
+    &:focus {
       opacity: 1;
+      outline: none;
```
0xdevalias, line 61: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -94,6 +94,7 @@ export const EventModal: React.FC<EventModalProps> = ({
 
   return (
     <Modal show={show} onHide={onHide} className="EventModal">
+      <Modal.Header className="EventModal__close" closeButton />
```
0xdevalias, line 97: Should this be added to all of the `Modal`s? I only see it on 1, I would expect to see a LOT of this to fix it across the platform.

src/components/organisms/EventModal/EventModal.scss
```diff
@@ -15,6 +15,12 @@ $EventModal--padding-top: 40px;
     background-color: $dark-grey;
   }
 
+  &__close {
+    background: transparent;
+    border: none;
```
0xdevalias, line 20: Why setting `border: none` here?

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -33,24 +33,41 @@ export const ChatSidebar: React.FC<ChatSidebarProps> = ({ venue }) => {
...
+    }
+  );
 
-  const privateChatTabStyles = classNames("chat-sidebar__tab", {
-    "chat-sidebar__tab--selected":
-      chatSettings.openedChatType === ChatTypes.PRIVATE_CHAT,
-  });
+  const privateChatTabStyles = classNames(
+    "chat-sidebar__tab",
+    "no-outline-on-focus",
```
0xdevalias, line 57: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -33,24 +33,41 @@ export const ChatSidebar: React.FC<ChatSidebarProps> = ({ venue }) => {
...
     "chat-sidebar--expanded": isExpanded,
   });
 
-  const venueChatTabStyles = classNames("chat-sidebar__tab", {
-    "chat-sidebar__tab--selected":
-      chatSettings.openedChatType === ChatTypes.VENUE_CHAT,
-  });
+  const venueChatTabStyles = classNames(
+    "chat-sidebar__tab",
+    "no-outline-on-focus",
```
0xdevalias, line 49: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -60,6 +64,8 @@ $header-height: 45px;
   }
 
   &__controller {
+    color: white;
+    border: none;
```
0xdevalias, line 68: Why removing border here?

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -35,6 +35,8 @@ $header-height: 45px;
   }
 
   &__tab {
+    background: transparent;
+    border: none;
```
0xdevalias, line 39: By removing border + why do we need to set transparent background here?

src/components/molecules/NavBar/playa.scss
```diff
@@ -79,6 +79,7 @@ $sand: #937c63;
         height: 40px;
         width: 40px;
         background-color: rgba($black, 0.4);
+        border: none;
```
0xdevalias, line 82: Why removing border here?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -288,8 +289,8 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
 
                 {showSoundCloudRadio && (
                   <S.RadioTrigger>
-                    <div
-                      className={`profile-icon navbar-link-radio ${
+                    <button
+                      className={`profile-icon navbar-link-radio no-outline-on-focus ${
```
0xdevalias, line 293: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -278,8 +279,8 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
                     rootClose={true}
                     defaultShow={showRadioOverlay}
                   >
-                    <div
-                      className={`profile-icon navbar-link-radio ${
+                    <button
+                      className={`profile-icon navbar-link-radio no-outline-on-focus ${
```
0xdevalias, line 283: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -209,14 +209,15 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
               )}
 
               {shouldShowSchedule ? (
-                <div
+                <button
+                  aria-label="Schedule"
```
0xdevalias, line 213: Should the label not be `navbarTitle` and "Schedule" given that's what the visible text is?

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -164,15 +164,25 @@ $border-radius: 28px;
...
+    outline: none;
+    border: none;
     background-color: rgba($white, 0.14);
     &:after {
       opacity: 0.7;
     }
   }
+  &:focus {
+    outline: none;
+  }
```
0xdevalias, line 185: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -164,15 +164,25 @@ $border-radius: 28px;
   white-space: nowrap;
   text-overflow: ellipsis;
   transition: transform 400ms $transition-function;
+  border: none;
+  appearance: none;
+  -webkit-appearance: none;
```
0xdevalias, line 169: Why are these being removed? Why using vendor prefixed styles?

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -164,15 +164,25 @@ $border-radius: 28px;
...
+  color: white;
```
0xdevalias, line 170: Use our variables?
```suggestion
  color: $white;
```

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -164,15 +164,25 @@ $border-radius: 28px;
...
   img {
     height: 40px;
   }
-  &:hover {
+  &:hover,
+  &:focus {
+    outline: none;
+    border: none;
```
0xdevalias, line 177: Why settings these to `none`?

Don't remove `outline`, it causes a11y issues not fixes them.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
```diff
@@ -33,16 +33,21 @@ export const AnnouncementMessage: React.FC<AnnouncementMessageProps> = ({
...
         centered: !isExpanded,
       })}
     >
       <div className="announcement-message">
         <RenderMarkdown text={message} />
       </div>
-      <span className="close-button" onClick={hideAnnouncement}>
+      <button
+        aria-label="Close announcement message"
+        className="close-button no-outline-on-focus"
```
0xdevalias, line 46: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -24,11 +24,15 @@ $message-margin: $spacing + $close-icon-width + 5px;
   text-align: center;
 
   .close-button {
+    background: transparent;
+    border: none;
```
0xdevalias, line 28: Why no border + transparent background?

src/components/molecules/AnnouncementMessage/AnnouncementMessage.scss
```diff
@@ -24,11 +24,15 @@ $message-margin: $spacing + $close-icon-width + 5px;
...
+    color: white;
```
0xdevalias, line 29: Use our variables?

```suggestion
    color: $white;
```

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -75,13 +75,13 @@ export const ChatMessage: React.FC<ChatProps> = ({
...
 
-          <div className="ChatMessage__reply-icon">
+          <button className="ChatMessage__reply-icon">
             <FontAwesomeIcon
               icon={faReply}
               size="sm"
               onClick={selectThisThread}
             />
-          </div>
+          </button>
```
0xdevalias, line 84: On other buttons like this you've added `aria-label` and similar. Why not here? Can we please be consistent in the changes we're making so that we ensure they are equally applied in all locations.

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -115,10 +117,16 @@ $reply-icon-radius: 20px;
...
+    transition: opacity 100ms;
+    border: none;
 
     &:hover {
       background-color: $secondary--lightest;
     }
+
+    &:focus {
+      outline: none;
+    }
```
0xdevalias, line 129: Don't remove `outline`, it causes a11y issues not fixes them.

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -96,7 +96,9 @@ $reply-icon-radius: 20px;
   }
 
   &__reply-icon {
-    display: none;
+    color: white;
```
0xdevalias, line 99: Why do we need to explicitly set this color here anyway?

```suggestion
    color: $white;
```

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -96,7 +96,9 @@ $reply-icon-radius: 20px;
...
+    display: flex;
+    opacity: 0;
```
0xdevalias, line 101: Why changed from `display: none`? Why using `opacity` here?

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -115,10 +117,16 @@ $reply-icon-radius: 20px;
 
     border-radius: $border-radius--max;
     background-color: $secondary--light;
+    transition: opacity 100ms;
+    border: none;
```
0xdevalias, line 121: Why transitioning `opacity`? Why not using the standard transition function variable? Why border none?

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -19,7 +19,7 @@ $reply-icon-radius: 20px;
 
   &:hover {
     .ChatMessage__reply-icon {
-      display: flex;
+      opacity: 1;
     }
```
0xdevalias, line 23: Why was this changed from `display: flex` to `opacity`?

0xdevalias (changes requested):
This almost certainly needs another full review once the questions/comments/etc here are fixed up. But here's my quick 'blast through' of questions/comments/changes/etc.

src/components/templates/PartyMap/components/Map/MapRoom.scss
whoaitsaimz: Please do NOT remove _outline_ anywhere. It's not good for accessibility.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
oleksii-front: It was done by @sabrinaorban . On my iPad it not affects the UI. Removed this style.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
oleksii-front: Same inheritance issue after changing tag

src/components/templates/PartyMap/components/Map/MapRoom.scss
oleksii-front: Same inheritance issue after changing tag

src/components/templates/PartyMap/components/Map/MapRoom.scss
oleksii-front: removed

src/scss/global.scss
oleksii-front: removed

src/components/templates/PartyMap/components/Map/MapRoom.scss
oleksii-front: In this case item zooms when it is in focus. I back all outlines across the project. But in my opinion that it is not right. iOS accessibility mode draws system-level outlines. Default outlines only useful by navigating using "tab" button.

oleksii-front (commented):
I used this tool to check the result https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd/related

src/components/organisms/RadioModal/RadioModal.tsx
oleksii-front: image

src/components/atoms/ChatMessage/ChatMessage.scss
0xdevalias: But surely we have it set to display: none in that instance because it explicitly shouldn’t be focusable, no?

src/components/molecules/NavBar/NavBar.tsx
0xdevalias: I believe they are also going to read the default text on the button though. I’m pretty sure this attribute only needs to be added when the standard text doesn’t exist/work (eg. When it’s an icon only button)

src/components/organisms/RadioModal/RadioModal.tsx
0xdevalias: That’s talking about an entirely different button though, and that chat button has no text on it because it’s just an image icon. So that makes sense. But in this case the button already seems to have text within it. So we shouldn’t have any need to duplicate that text again in the aria-label

src/settings.ts
```diff
@@ -94,6 +94,16 @@ export const REFETCH_SCHEDULE_MS = 10 * 60 * 1000; // 10 mins
 // How often to update location for counting
 export const LOC_UPDATE_FREQ_MS = FIVE_MINUTES_MS;
 
+export const WEEKDAYS = [
```
0xdevalias, line 97: Why was this added here? We use date-fns for our date handling.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -122,7 +126,12 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
             setSelectedDayIndex(dayIndex);
           }}
         >
-          {formatDayLabel(day)}
+          <button
+            aria-label={WEEKDAYS[day.getDay()]}
```
0xdevalias, line 130: There should be no need for an aria label here given the button already has text on it. And if we do explicitly need an aria-label then it should be using the same text and method of calculating it that we are using for the visible button

oleksii-front: The visible test is short version. I provided long version as a label.

src/components/organisms/RadioModal/RadioModal.tsx
mike-lvov: @oleksii-front ^?

src/components/atoms/ChatMessage/ChatMessage.scss
mike-lvov: @0xdevalias I guess it's correct. People with accessibility issues will not know of the existence of this button, if it's display: none

src/components/atoms/ChatMessage/ChatMessage.tsx
```diff
@@ -75,13 +75,16 @@ export const ChatMessage: React.FC<ChatProps> = ({
...
+            aria-label={replyButtonText}
+            className="ChatMessage__reply-icon"
+          >
             <FontAwesomeIcon
               icon={faReply}
               size="sm"
               onClick={selectThisThread}
             />
-          </div>
+          </button>
```
mike-lvov, line 87: move the onClick to the button

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
```diff
@@ -33,16 +33,22 @@ export const AnnouncementMessage: React.FC<AnnouncementMessageProps> = ({
...
   return (
     <div
+      aria-labelledby="announcement-container-message"
+      role="dialog"
       className={classNames("announcement-container", {
         centered: !isExpanded,
       })}
     >
-      <div className="announcement-message">
+      <div className="announcement-message" id="announcement-container-message">
```
mike-lvov, line 42: Any reason to add id?

src/components/molecules/NavBar/NavBar.tsx
mike-lvov: > I believe they are also going to read the default text on the button though. I’m pretty sure this attribute only needs to be added when the standard text doesn’t exist/work (eg. When it’s an icon only button

@oleksii-front have you checked if it's true?

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -68,11 +70,31 @@ $weekday--margin-right: 10px;
...
     &--active {
       opacity: 1;
       background-color: $primary;
       text-shadow: 0 2px 4px opaque-black(0.3);
       box-shadow: 0 3px 6px opaque-black(0.2);
+      border: none;
+
+      button:focus {
+        font-weight: inherit;
+      }
```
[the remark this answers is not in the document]
mike-lvov: @oleksii-front ^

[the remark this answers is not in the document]
mike-lvov: @oleksii-front ^

src/utils/time.ts
```diff
@@ -247,6 +247,10 @@ export const normalizeTimestampToMilliseconds = (timestamp: number) => {
     : timestamp * ONE_SECOND_IN_MILLISECONDS;
 };
 
+export const getWeekday = (date: Date | number) => {
+  return format(date, "EEEE");
+};
+
```
mike-lvov, line 253: should be deleted?

mike-lvov (changes requested):
Questions/ideas

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: There's a whole bunch of logic that goes into determining what is shown here, based on whether the schedule is timeshifted/etc. We need to ensure whatever we change/add here is going to align with that in all cases; without needlessly increasing the scope of work to maintain changes/avoid bugs.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -122,7 +122,12 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
             setSelectedDayIndex(dayIndex);
           }}
         >
-          {formatDayLabel(day)}
+          <button
+            aria-label={getWeekday(day)}
+            className="NavBarSchedule__weekday-button"
+          >
+            {formatDayLabel(day)}
+          </button>
```
denisdimitrov, line 130: Do you really need this button? The `onClick` event is passed to the parent node `li`. Is this going to have any impact at all?

src/components/molecules/NavBar/playa.scss
denisdimitrov: Using `none` and `inherit` isn't very popular/best practice and there is a reason for it. Compare staging to your branch and you will see the differences.

Just some examples:

This branch poll:
branch-poll

Staging poll:
staging-poll


This branch schedule:
branch-schedule


Staging schedule:
staging-schedule


I'm pretty sure these visual issues are coming from all of the inheritance and `none` values. This should be investigated and fixed before merging.

These 2 are the quick ones that I noticed, I saw that you change MapRoom's styles too, it might be visually affected as well.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -288,7 +289,7 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
 
                 {showSoundCloudRadio && (
                   <S.RadioTrigger>
-                    <div
+                    <button
```
denisdimitrov, line 292: I keep seeing `buttons` without any onClick prop what's the benefit here?

denisdimitrov (dismissed):

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
mike-lvov: @denisdimitrov the button is for accessibility purposes

src/components/atoms/ChatMessage/ChatMessage.tsx
oleksii-front: Done

src/components/molecules/NavBar/NavBar.tsx
mike-lvov: @denisdimitrov buttons should have the underlying onClicks. The benefit is so that the markup is aligned with the functionality. If it can be clicked, it should be a button

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
oleksii-front: I use aria-lableby to target this element.

src/utils/time.ts
oleksii-front: no.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
oleksii-front: yes. The focus works for buttons. Not for li's
image

src/components/molecules/NavBar/NavBar.tsx
oleksii-front: https://github.com/sparkletown/sparkle/pull/1727/files/b708780c5afa1dc9f135af4bc00a12d914d47e59#diff-d2dbbd9743869736d76cf73512c615856022b7a3ee80ec9b63bed997c40f7e37R296

src/components/molecules/NavBar/playa.scss
oleksii-front: @denisdimitrov  I got the same issue after pulling the latest 'staging'. Continue work on that.

src/components/molecules/NavBar/NavBar.tsx
oleksii-front: @mike-lvov It's true. But in this case the content is whole 'events' dialog. This change forces to read this button right.

src/components/organisms/RadioModal/RadioModal.tsx
oleksii-front: @mike-lvov @0xdevalias  agree.

denisdimitrov:
@oleksii-front There are still visual differences in the schedule. The subtitle has a weird margin left, the subtitle is not visible, and the message for no scheduled events is missing.

branch
staging

denisdimitrov:
@oleksii-front You should also add an event and test that as well. Looks like there is missing texts in the events section too.

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
mike-lvov: @oleksii-front could you explain it, please?

mike-lvov:
@oleksii-front were the visual effects fixed? Could you please provide the screenshots?

oleksii-front:
> @oleksii-front You should also add an event and test that as well. Looks like there are missing texts in the events section too.
> 
> 

@mike-lvov @denisdimitrov  here is how it looks now
image

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -122,7 +122,12 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
             setSelectedDayIndex(dayIndex);
           }}
         >
-          {formatDayLabel(day)}
+          <button
+            aria-label={getWeekday(day)}
```
denisdimitrov, line 126: Why not use `formatDayLabel(day)`? It already exists and it transforms the first two days into `Today` and `Tomorrow`

oleksii-front: Done

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
oleksii-front: @mike-lvov go to example 4 please https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
denisdimitrov: +1. Remove `getWeekday` it's no longer used.

oleksii-front: removed

dmitrii-zolotukhin:
@denisdimitrov Do I need to change these buttons too?

Снимок экрана 2021-07-19 в 15 07 39

dmitrii-zolotukhin:
@denisdimitrov How can I create new event for venue in app?

src/components/templates/PartyMap/components/Map/MapRoom.scss
0xdevalias: ‘Navigating with tab button’ is literally the point of accessibility..

src/components/atoms/ChatMessage/ChatMessage.scss
0xdevalias: Isn’t that the point though? And then presumably it’s set to a visible state when it is meant to be known about, to both sighted and screen reader based users?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
0xdevalias: But the onClick’s should be moved to the buttons, and not be on the li’s

src/components/molecules/AnnouncementMessage/AnnouncementMessage.tsx
0xdevalias: This is one of the very few sotuations where we actually need to use an id in react/etc.

In using an id though, we need to ensure it’s unique within the rendered application. I haven’t looked at the code here, but that’s something we need to ensure

src/utils/time.ts
0xdevalias: I’m kind of on @mike-lvov’s side with this one as well. Why does this need to stay? Seems wrong to me.

0xdevalias:
> @denisdimitrov Do I need to change these buttons too?
> 
> Снимок экрана 2021-07-19 в 15 07 39

Yup, makes sense to me. Basically anything that should be clickable or button like should be keyboard navigable. The simplest way to get that in most cases is to use a button element

0xdevalias:
> @denisdimitrov How can I create new event for venue in app?

Admin v1

dmitrii-zolotukhin:
> @oleksii-front There are still visual differences in the schedule. The subtitle has a weird margin left, the description is not visible, and the message for no scheduled events is missing.
> 
> branch
> staging

@denisdimitrov it's displaying correctly now as on the second

dmitrii-zolotukhin:
> @oleksii-front You should also add an event and test that as well. Looks like there are missing texts in the events section too.
> 
> 

@denisdimitrov also checked it and it's displaying correctly too

mike-lvov (approved):
Looks good

src/components/templates/PartyMap/components/Map/MapRoom.scss
whoaitsaimz: 100% with @0xdevalias.  I repeat -- please do NOT "back all outlines across the project" @oleksii-front

PR #1735, opened 2021-07-06
add CRACO + dedupe lodash imports from our bundle + prune unused lodash functions

0xdevalias:
## Overview

We do 3 main things here:

Add CRACO to allow customising `create-react-app` config without ejecting:

- https://github.com/gsoft-inc/craco
  - > Create React App Configuration Override is an easy and comprehensible configuration layer for `create-react-app`.
  - > Get all the benefits of `create-react-app` and customization without using 'eject' by adding a single `craco.config.js` file at the root of your application and customize your `eslint`, `babel`, `postcss` configurations and many more.

Use Webpack's `resolve.alias` functionality to de-duplicate `lodash` across all of our dependencies/our dependencies dependencies so that we only bundle exactly 1 version (which is then pruned by `babel-plugin-lodash`, see below)

- https://webpack.js.org/configuration/resolve/#resolvealias

Add `babel-plugin-lodash` to rewrite lodash imports to support better tree shaking, and thus allow unused functions to be pruned from our bundle automagically by webpack:

- https://github.com/lodash/babel-plugin-lodash

We could have chosen to use `lodash-webpack-plugin` to make things slightly smaller again, but it comes with some risks of potentially weirdly breaking behaviour, so it seemed not worthwhile to use here now:

- https://github.com/lodash/lodash-webpack-plugin

---

partially fixes sparkletown/internal-sparkle-issues#839

supercedes https://github.com/sparkletown/sparkle/pull/1715

---

## Results

We had 4 versions of `lodash` in our bundle, which has now been reduced to just a single version. The overall bundle size decreased from `3.49 MB` to `3.33 MB` by de-duplicating these + pruning the remaining functions to only keep the ones in use.

### Before Refactor

```
Compiled successfully.

File sizes after gzip:

  757.37 KB  build/static/js/2.0ff36f07.chunk.js
  133.44 KB  build/static/css/main.a7f79f4f.chunk.css
  129.2 KB   build/static/js/main.409bffc0.chunk.js
  9.48 KB    build/static/js/3.a560293f.chunk.js
  1.17 KB    build/static/js/runtime-main.3a521cfb.js
```

image

image

### After Refactor

```
Compiled successfully.

File sizes after gzip:

  705.61 KB (-51.76 KB)  build/static/js/2.09f96c93.chunk.js
  133.44 KB (+2 B)       build/static/css/main.54f8269b.chunk.css
  129.23 KB (+29 B)      build/static/js/main.22700bb8.chunk.js
  9.49 KB (+3 B)         build/static/js/3.60793b91.chunk.js
  1.17 KB                build/static/js/runtime-main.22afab2d.js
```

image

image


craco.config.js
```diff
@@ -0,0 +1,18 @@
...
+    plugins: [
+      // https://github.com/lodash/babel-plugin-lodash
+      ["lodash"],
+    ],
+  },
+  webpack: {
+    alias: {
+      // De-duplicate lodash in our bundle by ensuring all versions resolve to the same dependency
+      lodash: path.resolve(__dirname, "node_modules/lodash"),
+      "lodash-es": path.resolve(__dirname, "node_modules/lodash"),
```
goran-peoski-work, line 15: Any significant difference if the same dependency resolves to the ES version of Lodash instead of the CJS?
I'm guessing the biggest reason for the original is 3rd party libraries depending on it.

mike-lvov (approved):
LGTM

craco.config.js
0xdevalias: The lodash babel plugin is explicitly written to work on `lodash`, not `lodash-es`. And `lodash-es` doesn't contain `lodash/fp` which is needed for some things.

The end result is that the lodash babel plugin cuts everything down to only including what is actually needed/used, so it's like using `lodash-es` but better.

PR #1736, opened 2021-07-06
Alexey petrov/sort schedule rooms alphabetically

alexey-petrov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/884

* Added sorting for schedule room list
* Exception is 'Main Stage' room ( github related, always displays first in the list right after the 'my daily schedule' ( bookmarked events ) )

### DIFF
--- a/src/components/molecules/Schedule/Schedule.tsx
+++ b/src/components/molecules/Schedule/Schedule.tsx
@@ -66,10 +66,23 @@ export const Schedule: React.FC<ScheduleProps> = ({
     [scheduleStartHour, scheduleDate]
   );
 
+  const sortedEvents = useMemo(
+    () =>
+      locatedEvents.sort((a, b) => {
+        const nameA =
+          a.location.roomTitle ?? a.location.venueName ?? a.location.venueId;
+        const nameB =
+          b.location.roomTitle ?? b.location.venueName ?? b.location.venueId;
+
+        return nameA.localeCompare(nameB);
+      }),
+    [locatedEvents]
+  );
+
   // pairs (venueId, roomTitle) are unique because they are grouped earlier (see NavBarSchedule#schedule)
   const roomCells = useMemo(
     () =>
-      locatedEvents.map(({ location, events }) => (
+      sortedEvents.map(({ location, events }) => (
         <div
           key={`RoomCell-${location.venueId}-${location.roomTitle}`}
           className="Schedule__room"
@@ -82,7 +95,7 @@ export const Schedule: React.FC<ScheduleProps> = ({
           </span>
         </div>
       )),
-    [locatedEvents]
+    [sortedEvents]
   );
 
   const hoursRow = useMemo(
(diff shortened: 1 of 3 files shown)


yurii-lubynets (approved):
LGTM

src/settings.ts
```diff
@@ -87,6 +87,8 @@ export const DUST_STORM_TEXT_1 = `A dust storm is ripping across the ${PLAYA_VEN
 export const DUST_STORM_TEXT_2 =
   "Your only option is to seek shelter in a nearby venue!";
 
+export const GITHUB_MAIN_STAGE_NAME = "Main Stage";
```
sunny-viktoryia, line 90: if this PR is for `staging`, than the logic related to `GITHUB_MAIN_STAGE_NAME` should be removed as it's a GitHub specific request.
if this PR is for `env/github`, than the base branch should be changed.

sunny-viktoryia:
@alexey-petrov, please link your PRs to the issues they close for visibility.
looks like @malaikaaiyar has created a duplicate PR... https://github.com/sparkletown/sparkle/pull/1740

src/settings.ts
alexey-petrov: created another PR for env/github specifically https://github.com/sparkletown/sparkle/pull/1741
thx

alexey-petrov:
> @alexey-petrov, please link your PRs to the issues they close for visibility.
> looks like @malaikaaiyar has created a duplicate PR... #1740
> 
> you should use closing words for that and then add the link to the issue (I made it for you right now)
> image
> 
> then the PR gets linked to the issue
> image
> image

got it, thanks

src/utils/schedule.ts
```diff
@@ -0,0 +1,12 @@
+import { LocationEvents } from "types/venues";
+
+export const sortScheduleRoomsAlphabetically = (rooms: LocationEvents[]) => {
```
0xdevalias, line 3: Instead of a standalone function like this, this standalone function should only be the inner function passed to the .sort. Wrapping the rest in here like this makes it harder to use.

0xdevalias: Also.. would have to check docs/etc, but it looks like @malaikaaiyar ’s implementation may be better/more robust than this?


https://github.com/sparkletown/sparkle/pull/1740/files

src/components/molecules/Schedule/Schedule.tsx
```diff
@@ -66,10 +67,15 @@ export const Schedule: React.FC<ScheduleProps> = ({
     [scheduleStartHour, scheduleDate]
   );
 
+  const sortedEvents = useMemo(
```
0xdevalias, line 70: There shouldn’t be any need to separate this out into a new memo like this is there? Can we not just use .sort directly at the source of locatedEvents? (After refactoring the sort function as suggested in this review)

src/components/molecules/Schedule/Schedule.tsx
```diff
@@ -29,6 +29,7 @@ import { ScheduleRoomEvents } from "components/molecules/ScheduleRoomEvents";
 import { calcStartPosition } from "./utils";
 
 import "./Schedule.scss";
+import { sortScheduleRoomsAlphabetically } from "utils/schedule";
```
0xdevalias, line 32: This isn’t sorted with the other related imports

0xdevalias (commented):
Comment/changes

PR #1742, opened 2021-07-07
[env/github] remove auth-getCustomAuthConfig

ermiaSparkle:
remove auth-getCustomAuthConfig as https://github.com/sparkletown/internal-sparkle-issues/issues/839#issuecomment-873360254 in this comment suggested we don't need it for Github


mike-lvov:
Not sure if it's the best way to approach it. I don't think the code itself results in loading problems. May be should just comment out the Usage in `Login`? @0xdevalias wdyt?

ermiaSparkle:
@0xdevalias

0xdevalias:
> I don't think the code itself results in loading problems. May be should just comment out the Usage in `Login`?

@ermiaSparkle @mike-lvov Probably just comment it out + add `@debt` comments explaining why it was commented out like we have done for some of the other things being removed/hacked out of `env/github`

jonboutelle:
Can we drop this now that github is past? Verify with @ermiaSparkle at Monday sync-up

0xdevalias:
> Can we drop this now that github is past? Verify with @ermiaSparkle at Monday sync-up

@jonboutelle I think we can probably close this. It was proposed as an environment specific hack, probably wasn’t going to have a huge amount of impact on first load speed, and definitely not something we want to remove from the main codebase.

0xdevalias (changes requested):
I don't think this is a change we care about making anymore, so this PR can probably be closed.

PR #1744, opened 2021-07-07
fixed alphabetical sorting if number is used

alexey-petrov:
fix for https://github.com/sparkletown/internal-sparkle-issues/issues/884#issuecomment-875165878


yurii-lubynets (approved):
Nice catch! LGTM

PR #1747, opened 2021-07-07
Fixes chat display

mike-lvov:
Fixes chat display
closes https://github.com/sparkletown/internal-sparkle-issues/issues/892


jonboutelle (approved):
Awesome!

0xdevalias (commented):
Derp 🤦🏻 That's me rushing the change where this was introduced with a 🧠💤 and getting confused between the api of `RenderMarkdownInner` (which matches `RenderMarkdown`'s), and the API of `ReactMarkdown` (which passes text in the `children`.

The worst part is that I had it right originally and then changed it last second because I thought I had it wrong 😅 Should have trusted the type system saying it was fine..

Thanks for fixing this up!

PR #1751, opened 2021-07-07
hide glow from Auditorium v1 iframe

whoaitsaimz:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/817


jonboutelle (approved):
CSS: it has what plants crave

0xdevalias:
@whoaitsaimz Please never name branches with `env/` or similar. That is specifically reserved for branches that are literally our environments, and it's making this PR unmergeable due to our branch protection rules.

whoaitsaimz:
Aha -- thank you. That explains why it was having an issue being deployed, perhaps. What's a good format for a name for a branch that's specific to this environment? `whoaitsaimz/github-featurename`?

0xdevalias:
So long as it doesn't start with `env/` then it doesn't really matter. I follow the format `devalias/something-descriptive-here` personally, and that's kind of our recommended standard so that it's easy to see who's branches are who's, etc.

There's no need to specify that it's for a specific environment in the branch name on my/our side of things, only if it helps you. The branch that was branched from originally + the branch the PR is opened against to merge back into is what matters the most here.

0xdevalias:
@whoaitsaimz However, please prefix your PR titles with `[env/github]` to make it obvious it's not destined for the main `staging`/etc branches:

image

0xdevalias:
@whoaitsaimz @jonboutelle This build failed and broke deployment to `env/github`. Was this actually run/tested locally before being put up/approved?

- https://app.circleci.com/pipelines/github/sparkletown/sparkle/7734/workflows/ad87f82c-648f-46a2-8a8b-7c15a4f7ff63/jobs/12026

image

whoaitsaimz:
I'll revert right now and change border-radius to 0 instead of adding new variable to global

whoaitsaimz:
+ the answer is, no, I still need to connect to `sparkle-github` firebase from local, etc

0xdevalias:
> I'll revert right now and change border-radius to 0 instead of adding new variable to global

@whoaitsaimz It looks like you used `$border-radius--none` but never actually defined it in `scss/constants` and it doesn't seem to exist there.. So presumably you need to actually define it?

0xdevalias:
> & the answer is, no, I still need to connect to `sparkle-github` firebase from local, etc

@whoaitsaimz Changes to styling/etc like this that exist on core platform features can be run against `staging` to test/develop/etc. You just wont have access to the database/venues/etc itself from that environment, but you can just as easily create/use an auditorium in `staging` to test your fixes against that. And that's how you should be doing a majority if not literally all of your dev work. Connecting a local dev environment through to production is SUPER risky and could end up destroying it if you do something wrong. It's one of the reasons we don't tend to give most of our engineers access to the production environment backend, etc.

whoaitsaimz:
> It looks like you used $border-radius--none but never actually defined it in scss/constants and it doesn't seem to exist there.. So presumably you need to actually define it?

yes and did that but decided not to try to change global.scss in this PR -- just wanted to keep things in customOverride

However, I thought that if the variable didn't exist it would just be a 0... not break the whole thing. My bad

whoaitsaimz:
>Connecting a local dev environment through to production is SUPER risky and could end up destroying it if you do something wrong. 

Aha. Ok will try again to get this setup this evening (when I am not bogged down in all day meetings and have actual headspace to do deep work)

PR #1753, opened 2021-07-08
[env/github] fix hide glow from Auditorium v1 iframe + hide glow from reactions as well

whoaitsaimz:
Fix build errors introduced by https://github.com/sparkletown/sparkle/pull/1751 as per https://github.com/sparkletown/sparkle/pull/1751#issuecomment-876011919

closes sparkletown/internal-sparkle-issues#817


0xdevalias:
@whoaitsaimz This PR has conflicts, presumably because you didn't branch off an updated `env/github` again, but just pushed your old branch with more commits.

image

src/scss/customOverrides.scss
```diff
@@ -10,3 +11,20 @@
...
 }
+
+// hide glow from Auditorium v1 iframe
+.audience-container {
+  .audience-overlay {
+    .video-container {
+      .reaction-container {
+        box-shadow: none;
+      }
+      .video {
```
0xdevalias, line 22: Spacing between sibling blocks for readability:
```suggestion
      }

      .video {
```

0xdevalias:
> Your old code had `border: 1px black;`, the new code doesn't. Is that an explicit choice to remove it? Or an error of omission?
> 
> image

Looking at https://github.com/sparkletown/sparkle/pull/1753/commits/f4d7be77d861b5953d86b87add53e0691d6675ee i'm guessing it was intended.

whoaitsaimz:
> > Your old code had `border: 1px black;`, the new code doesn't. Is that an explicit choice to remove it? Or an error of omission?

Yes intended b/c we don't need to override it.  It's part of the original code I copied. I took it out so it's really only doing what's necessary in the override. With a black background for the page, doesn't make much difference here either way.

0xdevalias:
> Yes intended b/c we don't need to override it. It's part of the original code I copied. I took it out so it's really only doing what's necessary in the override. With a black background for the page, doesn't make much difference here either way.

Cool, makes sense.

0xdevalias (approved):
LGTM! :shipit:

PR #1719, opened 2021-07-05
refactor AppRouter to split routes into subrouters

0xdevalias:
**Note:** This PR depends on the following being merged first, so cannot be merged until that is done:

- [x] https://github.com/sparkletown/sparkle/pull/1718

---

While small on it's own, this PR sets up for some future refactor work.

- refactor `/enter/*` routes from `AppRouter` -> `EnterSubrouter`
- refactor `/account/*` routes from `AppRouter` -> `AccountSubrouter`
- refactor `/admin/*` routes from `AppRouter` -> `AdminV1Subrouter`
- refactor `/admin-ng/*` routes from `AppRouter` -> `AdminNGSubrouter`
- cleanup `/sparkleverse` redirect + fix logic
- cleanup default redirect + fix logic
- cleanup import sort order
- refactor `VenuePage` to use explicitly named export

PR #1754, opened 2021-07-08
fix 'black screen of death' infinite loading during new account onboarding flow + related tech debt

0xdevalias:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/886

Also fixes 'totally black screen after page navigation' caused by having nested `BrowserRouter`s issue introduced in https://github.com/sparkletown/sparkle/pull/1719

I also cleaned up a bunch of tech debt in the 'onboarding screens' in general, so might fix some other bugs in the process of that too.

The fixes here based on original investigation in https://github.com/sparkletown/internal-sparkle-issues/issues/886#issuecomment-876098881:

> RE: https://github.com/sparkletown/internal-sparkle-issues/issues/886#issuecomment-875987764
> 
> > Looks like I can replicate this seemingly consistently by just trying to access the following:
> > 
> > * `/account/questions?venueId=register`
> 
> `src/pages/Account/Questions.tsx` uses `useConnectCurrentVenue()` and `const venue = useSelector(currentVenueSelectorData) as AnyVenue;` to get the venue data.
> 
> `src/hooks/useConnectCurrentVenue.ts` uses `const venueId = useVenueId();` to read the `venueId` from the URL.
> 
> `src/hooks/useVenueId.ts` currently uses `const { venueId } = useParams() as ParamsTypes;` to read the params.. but I feel like this was a refactor/change we made ages back.. and I believe this won't actually read from the query string in the way that it's trying to here since `useParams` relates to React-Router's routes/etc:
> 
> - https://reactrouter.com/web/api/Hooks/useparams
> 
> In this instance, it's not using a `:venueId` section in the route, so it will never extract a `venueId`, which means `useConnectCurrentVenue` will never run to fetch the data.
> 
> So really, this has likely been broken like this for ages, and it only sort of works 'accidentally' because in the normal flow one of the previous 'steps' would have already loaded the venue data. So this bug _should_ be an edgecase that only happens when navigating to the URL directly (or as @jonboutelle theorised, refreshing the page once already at that URL, eg. when it loads so slowly that you assume it's broken)
> 
> The simplest 'triage fix' here would likely be to make `useVenueId` also check the URL query string params and use that as a fallback when `useParams` doesn't return the right value. We can access this with something like:
> 
> - https://reactrouter.com/web/api/Hooks/uselocation
> - https://reactrouter.com/web/api/location
> 
> I have some local changes that start to do this. Will polish them and either paste a diff here, or make a PR for it.
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/internal-sparkle-issues/issues/886#issuecomment-876098881_


src/pages/Account/CodeOfConduct.tsx
```diff
@@ -91,73 +127,82 @@ const CodeOfConduct: React.FunctionComponent<PropsType> = ({ location }) => {
...
   }
 
   const codeOfConductQuestions = IS_BURN
     ? BURN_CODE_OF_CONDUCT_QUESTIONS
     : venue.code_of_conduct_questions;
 
   return (
-    <div className="page-container code-of-conduct-container">
-      <div className="code-of-conduct-form">
+    <div className="CodeOfConduct page-container">
```
mike-lvov, line 147: do we need the page-container?

src/pages/Account/Profile.tsx
```diff
@@ -2,42 +2,36 @@ import React from "react";
...
-    venueName ??
-    getQueryParameters(window.location.search)?.venueId?.toString() ??
-    DEFAULT_VENUE;
-  const { returnUrl } = getQueryParameters(window.location.search);
+  const { user, userWithId } = useUser();
+
+  const venueId = useVenueId() ?? DEFAULT_VENUE;
+
+  const returnUrl: string | undefined =
+    useSearchParam("returnUrl") ?? undefined;
```
mike-lvov, line 34: can it return null?

mike-lvov (approved):
Did you test it?

src/pages/Account/CodeOfConduct.tsx
0xdevalias: Yeah, most of the styling still comes from it.

src/pages/Account/Profile.tsx
0xdevalias: It does by default, and I _guess_ it can since `??` handles `null` as well. But given we tend to use `undefined` I wanted to align it to our standards as soon as possible.

0xdevalias:
> Did you test it?

@mike-lvov Yup, a fair bit last night while developing it/etc. I feel like it's pretty robust. But also 🧠💤 at the end of a 17hr day.. so.. 🤷🏻 😅

PR #1755, opened 2021-07-08
[env/github] fix auditorium v1 glow + badges location hiding

whoaitsaimz:
Hopefully actually closes https://github.com/sparkletown/internal-sparkle-issues/issues/817

~~Added !important to all overrides. A horrible way to do this.~~  Updated by malaika w / specifishity ;-) 

Riff off of https://github.com/sparkletown/sparkle/pull/1753

### DIFF
--- a/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
+++ b/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
@@ -110,10 +110,7 @@ $all-section-previews-bottom-spacing: 50px;
   }
 
   &__iframe {
-    width: 100%;
     height: 100%;
-    border: thin $black;
-    border-radius: $border-radius--lg;
-    box-shadow: $iframe-box-shadow;
+    width: 100%;
   }
 }
--- a/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
+++ b/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
@@ -110,7 +110,10 @@ $all-section-previews-bottom-spacing: 50px;
   }
 
   &__iframe {
-    height: 100%;
     width: 100%;
+    height: 100%;
+    border: thin $black;
+    border-radius: $border-radius--lg;
+    box-shadow: $iframe-box-shadow;
   }
 }
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -90,8 +90,6 @@ $spacing--lg: 16px;
 $spacing--xl: 24px;
 $spacing--xxl: 30px;
 
-$iframe-box-shadow: 2px 2px $spacing--lg $white;
-
 $submit-button-diameter: 44px;
 
 $font-size--xxs: 10px;
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -90,6 +90,8 @@ $spacing--lg: 16px;
 $spacing--xl: 24px;
 $spacing--xxl: 30px;
 
+$iframe-box-shadow: 2px 2px $spacing--lg $white;
+
 $submit-button-diameter: 44px;
 
 $font-size--xxs: 10px;
--- a/src/scss/customOverrides.scss
+++ b/src/scss/customOverrides.scss
@@ -17,13 +17,13 @@
   .audience-overlay {
     .video-container {
       .reaction-container {
-        box-shadow: none;
+        box-shadow: none !important;
       }
 
       .video {
         .frame {
-          border-radius: 0;
-          box-shadow: none;
+          border-radius: 0 !important;
+          box-shadow: none !important;
         }
       }
     }
--- a/src/scss/customOverrides.scss
+++ b/src/scss/customOverrides.scss
@@ -3,12 +3,12 @@
 // This file exists for custom styles to deal with environment specific hacks/overrides
 
 // hide time on site and number of venues visited box (when badges are turned on)
-.Badges {
+div.Badges {
   &__visits {
-    display: none !important;
+    display: none;
   }
   &__visit {
-    display: none !important;
+    display: none;
   }
 }
 
--- a/src/scss/customOverrides.scss
+++ b/src/scss/customOverrides.scss
@@ -29,10 +29,3 @@ div.Badges {
     }
   }
 }
-
-// hide glow in Auditorium v2 iframe
-iframe.AllSectionPreviews__iframe,
-iframe.Section__iframe {
-  border-radius: 0;
-  box-shadow: none;
-}
--- a/src/scss/customOverrides.scss
+++ b/src/scss/customOverrides.scss
@@ -29,3 +29,10 @@ div.Badges {
     }
   }
 }
+
+// hide glow in Auditorium v2 iframe
+iframe.AllSectionPreviews__iframe,
+iframe.Section__iframe {
+  border-radius: 0;
+  box-shadow: none;
+}
--- a/src/scss/customOverrides.scss
+++ b/src/scss/customOverrides.scss
@@ -5,10 +5,10 @@
 // hide time on site and number of venues visited box (when badges are turned on)
 .Badges {
   &__visits {
-    display: none;
+    display: none !important;
   }
   &__visit {
-    display: none;
+    display: none !important;
   }
 }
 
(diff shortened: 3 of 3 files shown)


0xdevalias:
@whoaitsaimz Have you actually tested this change this time? I'd rather not deploy another PR for this without it being actually tested first.. this has been a source of much distraction today.

malaikaaiyar:
I tested it out and if you sit in a section there is still a glow

Screen Shot 2021-07-08 at 10 14 41 PM

In the sections view, it looks like this:
Screen Shot 2021-07-08 at 10 15 19 PM

whoaitsaimz:
The problem is my customOverrides.scss file is not taking priority. @malaikaaiyar would love your help w/ completing this one

My "hide time on venue & venues visited" no longer works either which makes me think there may have been a change to the code / loading order or something else that reduced priority of this file

whoaitsaimz:
Glenn linked us to https://specifishity.com/ and it's a great reference.

However, when specificity is the same, order of loading DOES matter.

Ref: https://css-tricks.com/precedence-css-order-css-matters/

So I suggest, again, that we check where & how customOverrides.scss is being compiled relative to Auditorium.scss

This way we don't have to use !important flags -- the "nuclear" option, on every line, which is just silly.

If that can't be done this weekend, then please verify and merge existing PR so we can get this quickly finalized.

malaikaaiyar:
Okay. I will take this on and complete it today. Thanks for the start though!

malaikaaiyar:
This is ready for review. I've gotten rid of all the !important and made the css selectors specific enough to take priority

0xdevalias:
> However, when specificity is the same, order of loading DOES matter.

> So I suggest, again, that we check where & how customOverrides.scss is being compiled relative to Auditorium.scss

@whoaitsaimz As I have told you elsewhere (but adding it here for completeness); it's correct that order has an impact, but relying on order is a brittle antipattern that doesn't work in modern web apps that use bundled assets. It's unstable, unreliable, and shouldn't be used. Specificity is the only true way to resolve this, which is why I linked you to those resources.

src/components/templates/Auditorium/components/Section/Section.scss
```diff
@@ -48,8 +48,6 @@ $spacing: calc(1 * min(var(--section-seat-size), var(--section-seat-size-min)));
     width: 100%;
     height: 100%;
     border: thin $black;
-    border-radius: $border-radius--lg;
-    box-shadow: $iframe-box-shadow;
```
0xdevalias, line 52: Why were these removed from here? I believe the intent was for all the changes to be made within `src/scss/customOverrides.scss` was it not?

src/scss/customOverrides.scss
```diff
@@ -5,23 +5,23 @@
...
 .Badges {
   &__visits {
-    display: none;
+    display: none !important;
   }
   &__visit {
-    display: none;
+    display: none !important;
   }
 }
```
0xdevalias, line 13: These still seem to use `!important`, which goes against what you suggested to be the case in https://github.com/sparkletown/sparkle/pull/1755#issuecomment-877652052 ? Whys that?

0xdevalias (changes requested):
Questions/comments.

src/components/templates/Auditorium/components/Section/Section.scss
malaikaaiyar: Oh I thought this was just for auditorium v1 changes. I'll put that in customOverrides then

src/scss/customOverrides.scss
malaikaaiyar: fixed in d414cb523b6cf736e1693468c5beeee0686bb8b4

src/components/templates/Auditorium/components/Section/Section.scss
malaikaaiyar: Those styles were removed from `AllSectionPreviews.scss` which is the other auditorium v2 component, so I thought that I should follow that. I'll add both back to their files and then 'remove' them in customOverrides

src/scss/customOverrides.scss
malaikaaiyar: fixed in c129a927c7a43aec64d1228a2bf25f6194528f40

src/components/templates/Auditorium/components/Section/Section.scss
malaikaaiyar: added all overriding styles to customOverrides in c129a927c7a43aec64d1228a2bf25f6194528f40

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
```diff
@@ -110,7 +110,10 @@ $all-section-previews-bottom-spacing: 50px;
   }
 
   &__iframe {
-    height: 100%;
     width: 100%;
+    height: 100%;
+    border: thin $black;
+    border-radius: $border-radius--lg;
+    box-shadow: $iframe-box-shadow;
   }
```
0xdevalias, line 118: Hrmm.. still not sure this is right? The only file I expect to see any changes from on this PR should be in `src/scss/customOverrides.scss`?

src/components/templates/Auditorium/components/Section/Section.scss
```diff
@@ -7,8 +7,6 @@ $section-seat-size: $avatarSize;
 $section-seat-size-min: $avatarSizeMin;
 $section-seat-spacing: 3px;
 
-$iframe-box-shadow: 2px 2px $spacing--lg $white;
-
```
0xdevalias, line 11: Hrmm.. still not sure this is right? The only file I expect to see any changes from on this PR should be in `src/scss/customOverrides.scss`?

src/scss/constants.scss
```diff
@@ -90,6 +90,8 @@ $spacing--lg: 16px;
 $spacing--xl: 24px;
 $spacing--xxl: 30px;
 
+$iframe-box-shadow: 2px 2px $spacing--lg $white;
+
```
0xdevalias, line 94: Hrmm.. still not sure this is right? The only file I expect to see any changes from on this PR should be in `src/scss/customOverrides.scss`?

0xdevalias (changes requested):
Looks like a few more unrelated changes sneaking in here.

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss
malaikaaiyar: @0xdevalias, the changes are not unrelated. In a past pr, someone (i think @whoaitsaimz) deleted the styles from auditorium v2 so that it doesn't have the glow. I re-added the glow styles to the components (needed to share the iframe-box-shadow through constants.scss) and "re-removed" it through customOverrides.scss.

0xdevalias: Are you sure they just weren't removed from the code in general? (eg. by @mike-lvov)

Searching for `border-radius: $border-radius--lg;` I don't see it in this file on `staging` either.

0xdevalias: https://github.com/sparkletown/sparkle/blob/staging/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss#L108-L115

And according to the history:

- https://github.com/sparkletown/sparkle/commits/staging/src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.scss

This file hasn't been touched in `staging` since it was merged by @mike-lvov in

- https://github.com/sparkletown/sparkle/pull/1251

malaikaaiyar: You're right. I'll make sure this only touches v1 stuff and `customOverrides.scss`

malaikaaiyar: I basically just reverted the last commit

malaikaaiyar: fixed in c78545c

whoaitsaimz: Correct, I did not change ANYTHING in any file other than customOverrides.scss ;-)

sunny-viktoryia (approved):
Approving to unblock the PR. :)
@alexey-petrov has tested and confirmed that it works.

alexey-petrov:
rebased & tested
seems to be working properly
image

whoaitsaimz:
Thank you!!!

PR #1757, opened 2021-07-08
add WorldUsersProvider to admin v1

denisdimitrov:
Adds the worlds users provider to admin v1 to fix 💣 This will need to be added in v3 as well but there are multiple pages there and needs more thought.

### DIFF
--- a/src/pages/Admin/Admin.tsx
+++ b/src/pages/Admin/Admin.tsx
@@ -52,7 +52,7 @@ import { useSelector } from "hooks/useSelector";
 import { useShowHide } from "hooks/useShowHide";
 import { useUser } from "hooks/useUser";
 import { useVenueId } from "hooks/useVenueId";
-import { WorldUsersProvider } from "hooks/users/useWorldUsers";
+import { WorldUsersProvider } from "hooks/users";
 
 import { PlayaContainer } from "pages/Account/Venue/VenueMapEdition";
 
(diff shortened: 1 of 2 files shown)


src/pages/Admin/Admin.tsx
```diff
@@ -52,6 +52,7 @@ import { useSelector } from "hooks/useSelector";
 import { useShowHide } from "hooks/useShowHide";
 import { useUser } from "hooks/useUser";
 import { useVenueId } from "hooks/useVenueId";
+import { WorldUsersProvider } from "hooks/users/useWorldUsers";
```
0xdevalias, line 55: Can you please add an export of `WorldUsersProvider` to `src/hooks/users/index.ts` then import it like this:

```suggestion
import { WorldUsersProvider } from "hooks/users";
```

0xdevalias (approved):
Small tweak that would be nice to do, otherwise LGTM! :shipit:

sunny-viktoryia (approved):
Approving. 🙃 

THANK YOU SO MUCH FOR QUICK FIX! ❤️

src/pages/Admin/Admin.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1757/commits/26699fa3ad119469e8647bf1b17ea10b4d76eade

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1757/commits/26699fa3ad119469e8647bf1b17ea10b4d76eade

PR #1769, opened 2021-07-09
fix Error: <WorldUsersProvider/> not found in a more holistic way

goran-peoski-work:
Should fix Error: `<WorldUsersProvider/>` not found in Admin components that was introduced in https://github.com/sparkletown/sparkle/pull/1703. One such issue is _Edit Room_ in Admin v1 panel.

Partial fix to https://github.com/sparkletown/internal-sparkle-issues/issues/886

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/904

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/905

Reverts and supercedes https://github.com/sparkletown/sparkle/pull/1757


0xdevalias (changes requested):
We probably don't want to add this in to `src/components/organisms/WithNavigationBar/WithNavigationBar.tsx` I don't think, but should be wrapping at a higher level (eg. potentially in the admin subrouter, or possibly could be around all of the routes maybe..? Not 100%

goran-peoski-work:
Possible edge case bug, by @0xdevalias :
> because admin lets you jump around between all sorts of venues, not just related venues. And since the `venueId` is passed into it, and that used to get the sovereign venue, but the sovereign venue cached forever currently, if we switch out of a related venue it will return the wrong users.

OOS for this fix, should be addressed/superseded/rendered moot by latter Admin development PRs

goran-peoski-work:
@sofisparkle Ready for review

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -10,26 +10,31 @@ import VenueWizard_v2 from "pages/Admin/Venue/VenueWizard/VenueWizard";
 import { AdminAdvancedSettings } from "pages/AdminAdvancedSettings";
 
 import { AdminVenueView } from "components/organisms/AdminVenueView";
+import { WorldUsersProvider } from "hooks/users";
+import { useVenueId } from "hooks/useVenueId";
```
0xdevalias, line 14: Sort `hooks/*` imports before the `pages/*`

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -10,26 +10,31 @@ import VenueWizard_v2 from "pages/Admin/Venue/VenueWizard/VenueWizard";
...
+  const venueId = useVenueId();
   return (
-    <Switch>
-      {/* Admin V1 */}
-      <Route path="/admin/venue/rooms/:venueId" component={RoomsForm} />
-      <Route path="/admin/venue/creation" component={VenueWizard} />
-      <Route path="/admin/venue/edit/:venueId" component={VenueWizard} />
-      <Route path="/admin/:venueId" component={Admin} />
-      <Route path="/admin" component={Admin} />
+    <WorldUsersProvider venueId={venueId}>
```
0xdevalias, line 19: This is how I was thinking of doing it.. but what I wasn't necessarily sure about was what `venueId` would actually end up being in this instance, since we're reading it from the 'routers path', yet that isn't defined yet in theory, since it happens in the `Route`s below.

Can you do some hacky console logs or similar and tell me what it returns when eg. on `/admin-ng` vs when on `/admin-ng/edit/:venueId` for example.

src/pages/Admin/Admin.tsx
```diff
@@ -456,24 +455,22 @@ const Admin: React.FC = () => {
   }
 
   return (
-    <WorldUsersProvider venueId={venueId}>
```
0xdevalias, line 459: [note] Removed from here because we provide it higher up in `src/components/organisms/AppRouter/AdminSubrouter.tsx` now.

0xdevalias (changes requested):
A tiny cleanup + a question about how it actually works.

Also, can you please check if it’s going to be needed for any of the other routes in `AppRouter`.

I know for a fact that `VenueAdminPage` (`/in/:venueId/admin`) needs it, as that’s the one that broke the banner admin (and it’s not part of the admin subrouter), but I expect there will be other places too.

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: The not having `:venueId` is the reason why I wanted to try wrapping only the components with those routes (and proved more work for less benefit),
but in general, we can just let the provider do the job only on existing id.

Will test.

goran-peoski-work: Provider already has a safeguard for missing `venueId` among other things:
```
  const shouldConnect =
    !isSovereignVenueLoading &&
    isDefined(sovereignVenueId) &&
    isDefined(venueId);

  // TODO: refactor this to not use useFirestoreConnect
  useFirestoreConnect(() => {
    if (!shouldConnect) return [];
```

Can't find edge case where missing id would display console errors.
Testing the `/admin-ng` and `admin` without it in parameters correctly displays message that a venue should be selected or just recovers it from the subsequent navigation URL which contains it.

goran-peoski-work:
@sofisparkle ready for review

0xdevalias:
@goran-peoski-work Please follow our standards for responding to PR review comments as per https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1625826204174500?thread_ts=1620395110.209200&cid=C01ER4M61TL

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -58,3 +59,12 @@ export const VenueAdminPage: React.FC = () => {
     </>
   );
 };
+
+export const VenueAdminPage = () => {
+  const venueId = useVenueId();
+  return (
+    <RelatedVenuesProvider venueId={venueId}>
```
goran-peoski-work, line 66: **Note**: It may look like OOS for the title of this PR, but it does help fix access to the `VenueAdminPage` in a sense there isn't BSOD and/or error.

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: This is what I was requesting above:

Using this hacky console log:
```diff
diff --git a/src/components/organisms/AppRouter/AdminSubrouter.tsx b/src/components/organisms/AppRouter/AdminSubrouter.tsx
index 30d7793b4..e7913b567 100644
--- a/src/components/organisms/AppRouter/AdminSubrouter.tsx
+++ b/src/components/organisms/AppRouter/AdminSubrouter.tsx
@@ -16,6 +16,9 @@ import { AdminVenueView } from "components/organisms/AdminVenueView";
 
 export const AdminSubrouter: React.FC = () => {
   const venueId = useVenueId();
+
+  console.log(venueId, window.location);
+
   return (
     <WorldUsersProvider venueId={venueId}>
       <Switch>

```

This is when accessing `/admin-ng`:

image

This is when accessing `/admin-ng/venue/devaliashacksville`:
image

Note that in both cases the `venueId` is `undefined`, which is invalid. So this isn't going to work for us.

0xdevalias: This will solve it though:

```diff
diff --git a/src/components/organisms/AppRouter/AdminSubrouter.tsx b/src/components/organisms/AppRouter/AdminSubrouter.tsx
index 30d7793b4..18e60d7e2 100644
--- a/src/components/organisms/AppRouter/AdminSubrouter.tsx
+++ b/src/components/organisms/AppRouter/AdminSubrouter.tsx
@@ -15,27 +15,69 @@ import { AdminAdvancedSettings } from "pages/AdminAdvancedSettings";
 import { AdminVenueView } from "components/organisms/AdminVenueView";
 
 export const AdminSubrouter: React.FC = () => {
-  const venueId = useVenueId();
   return (
-    <WorldUsersProvider venueId={venueId}>
-      <Switch>
-        {/* Admin V1 */}
-        <Route path="/admin/venue/rooms/:venueId" component={RoomsForm} />
-        <Route path="/admin/venue/creation" component={VenueWizard} />
-        <Route path="/admin/venue/edit/:venueId" component={VenueWizard} />
-        <Route path="/admin/:venueId" component={Admin} />
-        <Route path="/admin" component={Admin} />
-
-        {/* Admin V2/3/NG */}
-        <Route path="/admin-ng/venue/:venueId?" component={AdminVenueView} />
-        <Route
-          path="/admin-ng/advanced-settings/:venueId?"
-          component={AdminAdvancedSettings}
-        />
-        <Route path="/admin-ng/venue/creation" component={VenueWizard_v2} />
-        <Route path="/admin-ng/edit/:venueId" component={VenueWizard_v2} />
-        <Route path="/admin-ng" component={Admin_v2} />
-      </Switch>
-    </WorldUsersProvider>
+    <Switch>
+      {/* Admin V1 */}
+      <Route path="/admin/venue/rooms/:venueId">
+        <WithWorldUsersProviderForVenueId>
+          <RoomsForm />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin/venue/creation">
+        <WithWorldUsersProviderForVenueId>
+          <VenueWizard />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin/venue/edit/:venueId">
+        <WithWorldUsersProviderForVenueId>
+          <VenueWizard />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin/:venueId">
+        <WithWorldUsersProviderForVenueId>
+          <Admin />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin">
+        <WithWorldUsersProviderForVenueId>
+          <Admin />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+
+      {/* Admin V2/3/NG */}
+      <Route path="/admin-ng/venue/:venueId?">
+        <WithWorldUsersProviderForVenueId>
+          <AdminVenueView />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin-ng/advanced-settings/:venueId?">
+        <WithWorldUsersProviderForVenueId>
+          <AdminAdvancedSettings />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin-ng/venue/creation">
+        <WithWorldUsersProviderForVenueId>
+          <VenueWizard_v2 />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin-ng/edit/:venueId">
+        <WithWorldUsersProviderForVenueId>
+          <VenueWizard_v2 />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+      <Route path="/admin-ng">
+        <WithWorldUsersProviderForVenueId>
+          <Admin_v2 />
+        </WithWorldUsersProviderForVenueId>
+      </Route>
+    </Switch>
   );
 };
+
+const WithWorldUsersProviderForVenueId: React.FC = ({ children }) => {
+  const venueId = useVenueId();
+
+  console.log(venueId, window.location);
+
+  return <WorldUsersProvider venueId={venueId}>{children}</WorldUsersProvider>;
+};
```

0xdevalias: Something like this could also potentially work:

```typescript
const RouteWithWorldUsersProvider: React.FC<RouteWithWorldUsersProviderProps> = ({
  path,
  children,
}) => {
  const venueId = useVenueId();

  console.log(venueId, window.location);

  return (
    <Route {...{ path }}>
      <WorldUsersProvider venueId={venueId}>{children}</WorldUsersProvider>
    </Route>
  );
};
```

Then used like this:
```
{/*<Route path="/admin-ng" component={Admin_v2} />*/}
<RouteWithWorldUsersProvider path="/admin-ng">
  <Admin_v2 />
</RouteWithWorldUsersProvider>
```

src/pages/Admin/Venue/VenueAdminPage.tsx
0xdevalias: I changed the PR title to be more generic anyway:

> fix Error: `<WorldUsersProvider/>` not found in a more holistic way

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -58,3 +59,12 @@ export const VenueAdminPage: React.FC = () => {
...
+      <VenueAdminPageContents />
+    </RelatedVenuesProvider>
+  );
+};
```
0xdevalias, line 70: Why do we need to extract a new wrapper component here? Does `VenueAdminPage` itself cause the error? Or it's children? If the latter, just include it in the JSX within `VenueAdminPage`.

0xdevalias: But since we need to make changes to fix things as per https://github.com/sparkletown/sparkle/pull/1769#discussion_r666850256, it would make sense to follow that same pattern and just use that pattern to handle this at the `AppRouter` layer.

0xdevalias (changes requested):
Looks like the current implementation won't work unfortunately. Details + solutions in https://github.com/sparkletown/sparkle/pull/1769#discussion_r666850256

src/pages/Admin/Venue/VenueAdminPage.tsx
goran-peoski-work: `VenueAdminPage` uses `useConnectCurrentVenueNG(venueId)`  itself, so it needs provider to wrap it.

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: K, maybe it looks less flexible, but it does provide us with opportunities to:

- keep it DRY in case of boilerplate for all the routes
- allow us to include logic that may add one or more providers based on props (boolean flags I guess)

goran-peoski-work: And clean up the component at https://github.com/sparkletown/sparkle/pull/1769#discussion_r666860767

0xdevalias: https://github.com/sparkletown/sparkle/pull/1769#discussion_r666854028 Would seem the better choice to me

src/pages/Admin/Venue/VenueAdminPage.tsx
0xdevalias: That hook has nothing to do with the world users provider..?

mike-lvov:
This needs to be fixes as per https://github.com/sparkletown/sparkle/pull/1769#pullrequestreview-702928315

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Added wrapper component named `Provided`  https://github.com/sparkletown/sparkle/pull/1769/commits/5efe4611ba289afe281b83ccf1106504ed793b68

src/pages/Admin/Venue/VenueAdminPage.tsx
goran-peoski-work: reverted in 5efe4611ba289afe281b83ccf1106504ed793b68

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Sorted dd68f14446c2d665067f190cc17f3cb4eb01ec2a

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Had to add exception for one route (at aa35a4db8b00596596a6fb5b9e8edf3c0a1839ab) , didn't delve deeper into the cause

```
<Route
        path="/admin/venue/creation"
        component={
          // for some state related reasons, Provided fails this VenueWizard
          // and using function here will cause the error
          // so WithNavigationBar wraps the RelatedVenuesProvider for it
          VenueWizard
        }
      />
```

0xdevalias: I'm pretty sure that's going to have the exact same issues as pointed out in https://github.com/sparkletown/sparkle/pull/1769#discussion_r666850256 since we need the path from `Route` to be able to know how to extract the `:venueId` from it. Therefore we can't use any form of 'parent wrapper' to do it.

0xdevalias: As implemented in https://github.com/sparkletown/sparkle/commit/5efe4611ba289afe281b83ccf1106504ed793b68, i'm pretty sure that's going to have the exact same issues as pointed out in https://github.com/sparkletown/sparkle/pull/1769#discussion_r666850256 since we need the path from `Route` to be able to know how to extract the `:venueId` from it. Therefore we can't use any form of 'parent wrapper' to do it.

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -5,31 +5,100 @@ import Admin from "pages/Admin/Admin";
 import { RoomsForm } from "pages/Admin/Venue/Rooms/RoomsForm";
 import { VenueWizard } from "pages/Admin/Venue/VenueWizard";
 
-import Admin_v2 from "pages/Admin/Admin_v2";
-import VenueWizard_v2 from "pages/Admin/Venue/VenueWizard/VenueWizard";
+import AdminV2 from "pages/Admin/Admin_v2";
+import VenueWizardV2 from "pages/Admin/Venue/VenueWizard/VenueWizard";
```
0xdevalias, line 9: Why did we rename these imports here?

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -5,31 +5,100 @@ import Admin from "pages/Admin/Admin";
...
 import { AdminAdvancedSettings } from "pages/AdminAdvancedSettings";
 
 import { AdminVenueView } from "components/organisms/AdminVenueView";
+import { Provided } from "components/organisms/AppRouter/Provided";
```
0xdevalias, line 13: Since we're already in `components/organisms/AppRouter` this can just be a `./` import
```suggestion

import { Provided } from "./Provided";
```

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -5,31 +5,100 @@ import Admin from "pages/Admin/Admin";
...
+          </Provided>
+        )}
+      />
+      <Route
+        path="/admin-ng"
+        component={() => (
+          <Provided withWorldUsers>
+            <AdminV2 />
+          </Provided>
+        )}
```
0xdevalias, line 100: This `component={() => (` syntax is invalid and will cause constant re-renders.

The 'nested' method I used in my described solution is the canonically best/correct way to do this.

**Further Reading:**

- https://reactrouter.com/web/api/Route/route-render-methods
  - > **The recommended method of rendering something with a `<Route>` is to use children elements**, as shown above. There are, however, a few **other methods** you can use to render something with a `<Route>`. These are provided mostly for **supporting apps that were built with earlier versions of the router before hooks were introduced.**
  - https://reactrouter.com/web/api/Route/component
    - > When you use component (instead of render or children, below) the router uses `React.createElement` to create a new React element from the given component. That means **if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component.** When using an inline function for inline rendering, use the `render` or the `children` prop (below).
- https://reactrouter.com/web/api/Route/render-func
  - > This allows for convenient inline rendering and wrapping **without the undesired remounting explained above.**

src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -82,10 +83,31 @@ export const AppRouter: React.FC = () => {
...
+            )}
+          />
+          <Route
+            path="/in/:venueId"
+            component={() => (
+              <Provided withWorldUsers withRelatedVenues>
+                <VenuePage />
+              </Provided>
+            )}
+          />
```
0xdevalias, line 110: This `component={() => (` syntax is invalid and will cause constant re-renders.

The 'nested' method I used in my described solution is the canonically best/correct way to do this.

**Further Reading:**

- https://reactrouter.com/web/api/Route/route-render-methods
  - > **The recommended method of rendering something with a `<Route>` is to use children elements**, as shown above. There are, however, a few **other methods** you can use to render something with a `<Route>`. These are provided mostly for **supporting apps that were built with earlier versions of the router before hooks were introduced.**
  - https://reactrouter.com/web/api/Route/component
    - > When you use component (instead of render or children, below) the router uses `React.createElement` to create a new React element from the given component. That means **if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component.** When using an inline function for inline rendering, use the `render` or the `children` prop (below).
- https://reactrouter.com/web/api/Route/render-func
  - > This allows for convenient inline rendering and wrapping **without the undesired remounting explained above.**

src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -97,15 +119,15 @@ export const AppRouter: React.FC = () => {
...
             }}
           />
 
           <Route
             path="/"
-            render={() => {
+            component={() => {
               window.location.href = DEFAULT_REDIRECT_URL;
               return <LoadingPage />;
             }}
```
0xdevalias, line 133: Why were these changed? They were correct as the were before, and this change makes them broken and invalid.

0xdevalias: **Further Reading:**

- https://reactrouter.com/web/api/Route/route-render-methods
  - > **The recommended method of rendering something with a `<Route>` is to use children elements**, as shown above. There are, however, a few **other methods** you can use to render something with a `<Route>`. These are provided mostly for **supporting apps that were built with earlier versions of the router before hooks were introduced.**
  - https://reactrouter.com/web/api/Route/component
    - > When you use component (instead of render or children, below) the router uses `React.createElement` to create a new React element from the given component. That means **if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component.** When using an inline function for inline rendering, use the `render` or the `children` prop (below).
- https://reactrouter.com/web/api/Route/render-func
  - > This allows for convenient inline rendering and wrapping **without the undesired remounting explained above.**

src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+
+import { useVenueId } from "hooks/useVenueId";
+import { WorldUsersProvider } from "hooks/users";
+import { RelatedVenuesProvider } from "hooks/useRelatedVenues";
+
+interface ProvidedProps {
```
0xdevalias, line 7:
```suggestion
export interface ProvidedProps {
```

src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -0,0 +1,44 @@
...
+  if (withRelatedVenues) {
+    return (
+      <RelatedVenuesProvider venueId={venueId}>
+        {children}
+      </RelatedVenuesProvider>
+    );
+  }
+
+  return <>{children}</>;
+};
```
0xdevalias, line 44: The logic here can be simplified by making a 'noop' style wrapper component to use in the cases where we don't need them all. I'll make some changes and push them.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -52,9 +51,9 @@ import SecretPasswordForm from "components/molecules/SecretPasswordForm";
...
 
 export interface VenueLandingPageProps {
-  venue: Firestore["data"]["currentVenue"];
+  venue?: Firestore["data"]["currentVenue"];
   venueEvents?: Firestore["ordered"]["venueEvents"];
-  venueRequestStatus: Firestore["status"]["requested"]["currentVenue"];
+  venueRequestStatus?: Firestore["status"]["requested"]["currentVenue"];
   purchaseHistory?: Firestore["ordered"]["userPurchaseHistory"];
   venueId?: string;
 }
```
0xdevalias, line 59: Why were these types changed here?

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -5,31 +5,100 @@ import Admin from "pages/Admin/Admin";
...
+      />
+      <Route
+        path="/admin/venue/creation"
+        component={
+          // for some state related reasons, Provided fails this VenueWizard
+          // and using function here will cause the error
+          // so WithNavigationBar wraps the RelatedVenuesProvider for it
+          VenueWizard
+        }
+      />
```
0xdevalias, line 35: What were the issues/etc observed with this that made it 'fail'? Specifics of the errors, etc?

src/hooks/users/useWorldUserLocation.tsx
```diff
@@ -6,15 +6,15 @@ import { withId, WithId } from "utils/id";
 import { worldUsersByIdSelector } from "utils/selectors";
 
 import { useSelector } from "hooks/useSelector";
+import { useWorldUsersContext } from "hooks/users/useWorldUsers";
 
 // import { useWorldUsersContext } from "./useWorldUsers";
```
0xdevalias, line 11: There was an existing commented import for this here.
```suggestion
import { useWorldUsersContext } from "./useWorldUsers";
```

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -155,267 +154,256 @@ export const VenueLandingPage: React.FunctionComponent<VenueLandingPageProps> =
...
 
   return (
-    <WorldUsersProvider venueId={venueId}>
-      <WithNavigationBar>
-        <div className="container venue-entrance-experience-container">
-          <div
-            className="header"
-            style={{
-              background: `linear-gradient(
+    <WithNavigationBar>
```
0xdevalias, line 157: We could potentially/probably even roll `WithNavigationBar` into `Provided` I think.

0xdevalias (changes requested):
Questions/comments/changes/etc.

I'll likely be cloning this branch and working on them to expedite getting this ready/merged.

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: There was browser console wrnings about having `_` underscore in the name.
So to reduce that chatter and since they're already named locally to `AdminSubrouter.tsx` anyway, 
I just removed that one char to make them camelCase

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: In general, having two providers, one higher and one lower in the component tree, will there be issues of hooks getting the data from the closest ancestor (the lower level provider) or some other errors?

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: > The 'nested' method I used in my described solution is the canonically best/correct way to do this.

Had spent more than an hour trying to figure out why it's either _TypeScript_ complaining for the type or the router in the browser throwing _functions aren't components_ errors.

I did try `render` instead of `component` prop in the `Route`, that should reduce re-renders, but went with `component` instead due to lack of time to properly test the differences.

But I'm open to solution that would create something akin to the original suggestion i.e. some `<ProvidedRoute>` that will closely mimic the `<Route>` interface

src/components/organisms/AppRouter/Provided.tsx
goran-peoski-work: My original intent was to do some "clever" rebinding of the variable i.e.

```
let wrapper = component;
if(case1)
  wrapper = <Provider>{wrapper}</Provider}
if(case2
// etc.
```

But due to time constraints,, I aborted the attempt and stuck with the dumbest but working solution

src/pages/VenueLandingPage/VenueLandingPage.tsx
goran-peoski-work: I removed the provider from `WithNavigationBar` and worked in most cases, but did trip some OOS bug https://github.com/sparkletown/sparkle/pull/1769/commits/aa35a4db8b00596596a6fb5b9e8edf3c0a1839ab
so I left it there, as an exception until another PR fixes that issue.

My guess is something to do with https://github.com/facebook/react/issues/18178 or, more specifically to the case, some child component should use `useEffect()` when passing event handler. Can't remember the specific names, but can be reproduced by following the comment in the code https://github.com/sparkletown/sparkle/commit/aa35a4db8b00596596a6fb5b9e8edf3c0a1839ab.

src/pages/VenueLandingPage/VenueLandingPage.tsx
goran-peoski-work: Didn't allow for `<VenueLandingPage>` to be used without props, even though those props weren't being passed in through JSX code but retrieved inside the component.

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Didn't note them at the time. Will add the provider to get exact error message.

goran-peoski-work: Steps to reproduce: 

1. replace the `component` prop with:
```
        component={() => (
          <Provided withWorldUsers withRelatedVenues>
            <VenueWizard />
          </Provided>
        )}
```
2. head to URL ending in `/admin/venue/creation?page=1`
3. pick a space type, and click `Create your space` button

Error in browser console:
```
Warning: Cannot update a component (`WithNavigationBar`) while rendering a different component (`DetailsForm`). To locate the bad setState() call inside `DetailsForm`, follow the stack trace as described in https://fb.me/setstate-in-render
    in DetailsForm (at VenueWizard.tsx:133)
    in div (at WithNavigationBar.tsx:46)
    in WithNavigationBar (at VenueWizard.tsx:143)
    in VenueWizardCreate (at VenueWizard.tsx:61)
    in VenueWizard (at AdminSubrouter.tsx:31)
    in RelatedVenuesProvider (at Provided.tsx:22)
    in WorldUsersProvider (at Provided.tsx:21)
    in Provided (at AdminSubrouter.tsx:30)
    in component (created by Context.Consumer)
    in Route (at AdminSubrouter.tsx:27)
    in Switch (at AdminSubrouter.tsx:17)
    in AdminSubrouter (created by Context.Consumer)
    in Route (at AppRouter.tsx:76)
    in Switch (at AppRouter.tsx:73)
    in Suspense (at AppRouter.tsx:72)
    in Router (created by BrowserRouter)
    in BrowserRouter (at AppRouter.tsx:71)
    in AppRouter (at src/index.tsx:283)
```

goran-peoski-work: Same or similar error will show up with or without `RelatedVenuesProvider` in `WithNavigationBar`

src/components/organisms/AppRouter/AppRouter.tsx
goran-peoski-work: > Why were these changed? They were correct as the were before, and this change makes them broken and invalid.

Typo. I had converted multiple `component` to `render`, but reverted them. This one shouldn't have.

src/components/organisms/AppRouter/AppRouter.tsx
goran-peoski-work: Went with the first working solution, same as https://github.com/sparkletown/sparkle/pull/1769/files#r667278276

src/components/organisms/AppRouter/AppRouter.tsx
goran-peoski-work: Reverted 1ba4f769cbdd73764c103e8cde1ba2a26f33f772

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: The 'lowest' provider will 'win' and will result in duplicated data/etc if stored in the provider itself I believe.

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: > There was browser console wrnings about having _ underscore in the name.

I've literally never seen that anywhere. Can you please reproduce and then provide screenshot/further details.

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: > Started with the "canonical solution". Had spent more than an hour trying to figure out why it's either TypeScript complaining for the type or the router in the browser throwing functions aren't components errors.

Are you able to reproduce that now? And if so, can you please show me the exact code that would cause that?

0xdevalias: > I'm open to solution that would create something akin to the original suggestion i.e. some `<ProvidedRoute>` that will closely mimic the `<Route>` interface

This was my secondary solution that I decided was a worse idea than my primary solution.

src/components/organisms/AppRouter/Provided.tsx
0xdevalias: The 'clever' pattern I was thinking of is basically this: 

https://github.com/sparkletown/sparkle/blob/staging/src/index.tsx#L234-L236

Something like this (untested, but should give the gist):

```typescript
const MaybeWorldUsersProvider = withWorldUsers ? WorldUsersProvider : React.Fragment;
const MaybeRelatedVenuesProvider = withRelatedVenues ? RelatedVenuesProvider : React.Fragment;

return (
  <MaybeWorldUsersProvider venueId={venueId}>
    <MaybeRelatedVenuesProvider venueId={venueId}>
      {children}
    </MaybeRelatedVenuesProvider>
  </MaybeWorldUsersProvider>
);
```

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: Is the type correct at all then? If we're not meant to be passing it props, those props should be removed from it's signature?

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: And that error doesn't show up when you don't wrap it like this? That sounds like an error within that component itself rather than an issue with how we're doing things here.

I ran into similar at some point during my refactoring of the 'account onboarding' flow. Can't remember exactly, but I think the places where it happened it was due to code trying to change navigation/etc in the 'main body' of the component (which is invalid React), rather than using a `useEffect` or similar to wrap it.

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: See my response in https://github.com/sparkletown/sparkle/pull/1769#discussion_r667284025

But basically I believe it's in line with this:

> Some child component should use `useEffect()` when passing event handler.

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: Strange.. I wonder if this is something new, or if i've just never noticed it before:

image

Either way, change seems fine.

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1769/commits/c58428242104ab1c0becf206460260f3c346a754

This doesn't seem to throw errors, and is the exact solution I described originally, so not sure what you were doing to get the errors?

src/components/organisms/AppRouter/AppRouter.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1769/commits/c58428242104ab1c0becf206460260f3c346a754

src/components/organisms/AppRouter/Provided.tsx
0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1769/commits/a037b8c6644fbb481981e2a84327fe2fe53e0ec9

Mostly followed my plan from https://github.com/sparkletown/sparkle/pull/1769#discussion_r667283279 but added some type safety as well.

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: This component is only used in `AppRouter` which doesn't pass any props to it. The component also never actually reads any props passed to it.

fixed by removing the type completely in f3ed798dfc0ba72a98457f68b714dbdb966a5ed7

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: Replicated following instructions in https://github.com/sparkletown/sparkle/pull/1769#discussion_r667280484

---

> And that error doesn't show up when you don't wrap it like this? 

Seems the error doesn't show up given the currently committed syntax (which is strange and a little confusing to me.. but anyway..

Looking further into this now.

0xdevalias: Ok, pretty sure I have found it:

`src/pages/Admin/Venue/DetailsForm.tsx:272` calls the `previous()` function in the body of the component.

This is defined at `src/pages/Admin/Venue/VenueWizard.tsx:130` and uses `history.push` to navigate to a new page.

image


I can wrap it in `useEffect` properly and the error goes away, but need to dig into the expect logic of this section of the admin panel more, because once I 'fix' that, the rest of the flow seems broken since I always get sent back to that first menu page because `state.templatePage` doesn't ever seem to be defined for me.

0xdevalias: > once I 'fix' that, the rest of the flow seems broken since I always get sent back to that first menu page because `state.templatePage` doesn't ever seem to be defined for me.

Ok, got this figured out, and it's basically a case of the exact problem I called out in https://github.com/sparkletown/sparkle/pull/1769#discussion_r667270203 where using the `component={() => (` syntax  would cause issues because it will always unmount/re-mount the component.

So in this case, the `state` is empty because it's been unmounted and remounted.

But by properly changing this `Route` to use the 'nested' style + wrapping the navigation in a `useEffect` within the component itself, it seems to fix it.

0xdevalias: fixed in https://github.com/sparkletown/sparkle/pull/1769/commits/cc0930efb0a06d0a1a6ba22e43128973f681b427 + https://github.com/sparkletown/sparkle/pull/1769/commits/4fd34816ca35db8046e256cda95b6e842bce3684

src/pages/VenueLandingPage/VenueLandingPage.tsx
0xdevalias: > We could potentially/probably even roll `WithNavigationBar` into Provided I think.

Calling this OOS for this PR. Might be something interesting to look at doing in future though.

0xdevalias:
@ermiaSparkle Why did you do this?

image

ermiaSparkle:
@0xdevalias 
Soooooooooooooooooo sorry
I was going to apply his changes to mine. not mine to his 
:((((((((((((((((((((((((((((((((((((((((((
How can I fix it?

0xdevalias:
> How can I fix it?

@ermiaSparkle I've pulled the branch and rebased it to remove your commits, then force pushed it, so should be fixed now.

image

src/components/organisms/AppRouter/AdminSubrouter.tsx
0xdevalias: I've looked at the code that exists now and it seems to work fine. Resolving this.

src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -82,10 +83,23 @@ export const AppRouter: React.FC = () => {
           {/* @debt The /login route doesn't work since we added non-defaulted props to the Login component */}
           {/*<Route path="/login" component={Login} />*/}
 
-          <Route path="/v/:venueId" component={VenueLandingPage} />
+          <Route path="/v/:venueId">
+            <Provided withWorldUsers withRelatedVenues>
+              <VenueLandingPage />
+            </Provided>
+          </Route>
+          {/* TODO: change this to the same 'nested Provided' pattern we are using everywhere else here? */}
```
0xdevalias, line 91:
```suggestion
```

0xdevalias (approved):
LGTM! :shipit: 🚄

src/pages/VenueLandingPage/VenueLandingPage.tsx
goran-peoski-work: K, reason enough to resolve this convo.

PR #1772, opened 2021-07-09
remove hardcoded user statuses

denisdimitrov:
- Removes the hard coded user statuses from the admin and status dropdown
- Handle an edge case, where the user status can be outdated and shown incorrectly due to database changes. Comment added in the code.
- Fix a bug where user statuses were cut if the list was too long.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/720#issuecomment-876860333
Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/720#issuecomment-876858552

### DIFF
--- a/src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
+++ b/src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
@@ -1,11 +1,9 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useEffect } from "react";
 import { Dropdown, DropdownButton } from "react-bootstrap";
 
-import { USER_STATUSES } from "settings";
+import { UserStatus } from "types/User";
 
 import { useProfileStatus } from "hooks/useProfileStatus";
-
-import { UserStatus } from "types/User";
 
 import "./UserStatusDropdown.scss";
 
@@ -14,18 +12,28 @@ export interface UserStatusDropdownProps {
 }
 
 export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
-  userStatuses,
+  userStatuses = [],
 }) => {
   const { status, changeUserStatus } = useProfileStatus();
 
-  const allUserStatuses = useMemo(
-    () => (userStatuses ? [...USER_STATUSES, ...userStatuses] : USER_STATUSES),
-    [userStatuses]
-  );
+  // This will check if the user status from the database exists in the venue user statuses and if it doesn't, it will fallback to the first one from the list.
+  useEffect(() => {
+    const hasUserStatuses = !!userStatuses.length;
+
+    if (!status || !hasUserStatuses) return;
+
+    const statusTexts = userStatuses.map((userStatus) => userStatus.status);
+
+    const defaultUserStatus = userStatuses[0].status;
+
+    if (!statusTexts.includes(status)) {
+      changeUserStatus(defaultUserStatus);
+    }
+  }, [changeUserStatus, status, userStatuses]);
 
   const userStatusDropdownOptions = useMemo(
     () =>
-      allUserStatuses.map((userStatus) => (
+      userStatuses.map((userStatus) => (
         <Dropdown.Item
           key={userStatus.status}
           onClick={() => changeUserStatus(userStatus.status)}
@@ -33,7 +41,7 @@ export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
           {userStatus.status}
         </Dropdown.Item>
       )),
-    [allUserStatuses, changeUserStatus]
+    [userStatuses, changeUserStatus]
   );
 
   return (
--- a/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
+++ b/src/components/organisms/ProfileModal/ProfilePopoverContent.scss
@@ -16,6 +16,7 @@ $profile-popover--margin-top: 12px;
   overflow: auto;
   padding: $spacing--md;
   background: $black;
+  height: 100%;
   background-image: linear-gradient(
     0deg,
     rgba($white, 0.05) 0%,
--- a/src/hooks/useVenueUserStatuses.ts
+++ b/src/hooks/useVenueUserStatuses.ts
@@ -1,8 +1,4 @@
-import {
-  ONLINE_USER_STATUS,
-  DEFAULT_SHOW_USER_STATUSES,
-  USER_STATUSES,
-} from "settings";
+import { DEFAULT_USER_STATUS, DEFAULT_SHOW_USER_STATUSES } from "settings";
 
 import { User } from "types/User";
 
@@ -15,9 +11,7 @@ export const useVenueUserStatuses = (venueId?: string, user?: WithId<User>) => {
 
   const venueStatuses = sovereignVenue?.userStatuses ?? [];
 
-  const statuses = [...USER_STATUSES, ...venueStatuses];
-
-  const userStatus = statuses.find(
+  const userStatus = venueStatuses.find(
     (userStatus) => userStatus.status === user?.status
   );
 
@@ -25,6 +19,6 @@ export const useVenueUserStatuses = (venueId?: string, user?: WithId<User>) => {
     venueUserStatuses: venueStatuses,
     isStatusEnabledForVenue:
       sovereignVenue?.showUserStatus ?? DEFAULT_SHOW_USER_STATUSES,
-    userStatus: userStatus ?? ONLINE_USER_STATUS,
+    userStatus: userStatus ?? DEFAULT_USER_STATUS,
   };
 };
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -628,17 +628,10 @@ export const SEARCH_DEBOUNCE_TIME = 200; // ms
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 48;
 export const DEFAULT_DISPLAYED_VIDEO_PREVIEW_COUNT = 12;
 
-export const ONLINE_USER_STATUS = {
+export const DEFAULT_USER_STATUS = {
   status: DefaultUserStatus.online,
   color: "#53E52A",
 };
-
-export const BUSY_USER_STATUS = {
-  status: DefaultUserStatus.busy,
-  color: "#F44336",
-};
-
-export const USER_STATUSES = [ONLINE_USER_STATUS, BUSY_USER_STATUS];
 
 // SCHEDULE
 export const DEFAULT_SHOW_SCHEDULE = true;
--- a/src/components/molecules/NavBar/NavBar.scss
+++ b/src/components/molecules/NavBar/NavBar.scss
@@ -249,4 +249,8 @@ $border-radius: 28px;
       margin-right: $spacing--sm;
     }
   }
+
+  &__profile-popover {
+    height: 100%;
+  }
 }
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -54,7 +54,7 @@ const TicketsPopover: React.FC<{ futureUpcoming: UpcomingEvent[] }> = (
 
 const ProfilePopover = (
   <Popover id="profile-popover">
-    <Popover.Content>
+    <Popover.Content className="NavBar__profile-popover">
       <ProfilePopoverContent />
     </Popover.Content>
   </Popover>
--- a/src/components/molecules/UserStatusManager/UserStatusManager.tsx
+++ b/src/components/molecules/UserStatusManager/UserStatusManager.tsx
@@ -1,8 +1,6 @@
 import React, { useCallback, useMemo } from "react";
 import { Button } from "react-bootstrap";
 
-import { USER_STATUSES } from "settings";
-
 import { UserStatus } from "types/User";
 
 import { Checkbox } from "components/atoms/Checkbox";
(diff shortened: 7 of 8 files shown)


src/components/organisms/ProfileModal/ProfilePopoverContent.scss
```diff
@@ -16,6 +16,7 @@ $profile-popover--margin-top: 12px;
   overflow: auto;
   padding: $spacing--md;
   background: $black;
+  height: 100%;
```
sunny-viktoryia, line 19: can you explain why was that added? can't see the difference if I remove it. 🙈

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -1,39 +1,47 @@
...
 }) => {
   const { status, changeUserStatus } = useProfileStatus();
 
-  const allUserStatuses = useMemo(
-    () => (userStatuses ? [...USER_STATUSES, ...userStatuses] : USER_STATUSES),
-    [userStatuses]
-  );
+  // This will check if the user status from the database exists in the venue user statuses and if it doesn't, it will fallback to the first one from the list.
+  useEffect(() => {
+    const hasUserStatuses = !!userStatuses.length;
```
sunny-viktoryia, line 21: I believe we can move this check to `src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx`
and render `UserStatusDropdown` only when the statuses are enabled and the length of the statuses array > 0.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -1,39 +1,47 @@
...
 
 export interface UserStatusDropdownProps {
   userStatuses?: UserStatus[];
 }
 
 export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
-  userStatuses,
+  userStatuses = [],
 }) => {
   const { status, changeUserStatus } = useProfileStatus();
```
sunny-viktoryia, line 17: can we use `useVenueUserStatuses` hook here for `status`? looks like it knows what to do when there is no existent status. (and no need of the `useEffect` then). What do you think? 🤔

sunny-viktoryia (commented):
a few questions and suggestions in this first-pass review. 🙃

sunny-viktoryia:
@denisdimitrov, also, I believe it's OOS of this PR, but could you create one more PR to display the text of the user statuses as it was designed.

At least in 2 places.
Chat
image

User Profile Modal
image


check the designs in the issue: https://github.com/sparkletown/internal-sparkle-issues/issues/720

denisdimitrov:
@sunny-viktoryia 
>@denisdimitrov, also, I believe it's OOS of this PR, but could you create one more PR to display the text of the user statuses as it was designed.

Yes, I'd love to. I'll create a follow up PR and that will mark the issue for user statuses as done.

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
denisdimitrov: @sunny-viktoryia If you add too many user statuses it breaks, you can find the bug in the description of this PR.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
denisdimitrov: I'll check and apply it if this is true :)

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
denisdimitrov: Actually, it makes sense. I will see if I can replace it with the other one and remove `useProfileStatus` because this is the only place it's being used.

src/settings.ts
```diff
@@ -628,18 +628,11 @@ export const SEARCH_DEBOUNCE_TIME = 200; // ms
 export const DEFAULT_DISPLAYED_POSTER_PREVIEW_COUNT = 48;
 export const DEFAULT_DISPLAYED_VIDEO_PREVIEW_COUNT = 12;
 
-export const ONLINE_USER_STATUS = {
+export const DEFAULT_USER_STATUS = {
   status: DefaultUserStatus.online,
```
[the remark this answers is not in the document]
denisdimitrov: Yeah, I will remove it and hardcode the string in here.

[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1772/commits/b2fbb77490f27b007d651eb4d75d2e62d3fb15e6

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1772/commits/b2fbb77490f27b007d651eb4d75d2e62d3fb15e6

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
denisdimitrov: The result is the same :/ Both hooks only get the status from the database, but if that status was set 5-10 days ago and the statuses have changed, the hook is still required and needs to check if that status exists.

denisdimitrov: I can move `changeUserStatus` in `useVenueUserStatuses` hook and remove `useProfileStatus` but the result would be the same. It's your call. I don't mind refactoring it in here.

sunny-viktoryia: @denisdimitrov I believe both hooks serve similar purposes. So I'd merge the functionality. Can we do that?

denisdimitrov: @sunny-viktoryia I'll merge it and commit in about 10 minutes, are you going to be around to approve this?

sunny-viktoryia: yes, I'll be here.

denisdimitrov: Unless github decides that this is eng-lead worthy PR and doesn't accept your approve 😄

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/1772/commits/80aec3855dd428a02639aa6009e990f5fc8e837e

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
```diff
@@ -150,9 +150,9 @@ export const UserInformationContent: React.FunctionComponent<UserInformationCont
           >
             {email}
           </div>
-          {sovereignVenue?.showUserStatus && (
+          {sovereignVenue?.showUserStatus && sovereignVenue?.userStatuses && (
```
sunny-viktoryia, line 153: looks like `<UserStatusDropdown />` breaks when `userStatuses` list is empty (length === 0).
maybe you were right adding this check to the `< UserStatusDropdown />` component;

so we need to check both cases: `userStatuses is undefined`, `userStatuses is []`

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
denisdimitrov: The user statuses were broken after that commit so I had to add some additional changes as well.

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
```diff
@@ -1,46 +1,52 @@
...
-    () => (userStatuses ? [...USER_STATUSES, ...userStatuses] : USER_STATUSES),
-    [userStatuses]
-  );
+  // This will check if the user status from the database exists in the venue user statuses and if it doesn't, it will fallback to the first one from the list.
+  useEffect(() => {
+    if (!userStatus) return;
+
+    const statusTexts = userStatuses.map((userStatus) => userStatus.status);
+
+    const defaultUserStatus = userStatuses[0].status;
```
sunny-viktoryia, line 25: I'm curious if `userStatuses[0].status` is a default status we should display... 
maybe we'll better use default status from `settings` ? because the 1st status in the `userStatuses` array is not necessarily what we need.
image
am I a noob then?.. 😐😅 

I guess it'd be more correct to say that I'm just online (our default status).

If we agree on that, seems that we don't need `defaultUserStatus` const and can do 
```
       changeUserStatus(userStatus);
```

as the `useVenueUserStatuses` hook will do that for us.

sunny-viktoryia (commented):
suuuuper cloooose. A couple of comments here. 🙃

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
whoaitsaimz: Great callout. I can change the order and make "here for it" the first option and default.  I think that status [0] should def be default rather than "online" being hard coded

whoaitsaimz: Tried to change the order (by adding new one, re-typing over 1, deleting first position) and getting this unhelpful error lol 

image

whoaitsaimz: (this is the very bottom of the page, I cant scroll any more and there is nothing about the error displayed even in html)

whoaitsaimz: I went back in, changed last status to `here for it 2`, changed status 1 to 'here for it`, saved, it worked. Then went back in, changed last status to `a n00b!` and saved, and it worked. So I'm unblocked here

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
whoaitsaimz: @denisdimitrov and @sunny-viktoryia what's up here? Can we finalize plz?

src/components/atoms/UserStatusDropdown/UserStatusDropdown.tsx
whoaitsaimz: @denisdimitrov and @sunny-viktoryia  That's not a problem you need to fix for us at the code level. If the first two statuses are not hard-coded, as an admin I can decide for myself what the first / default status should be. I'm going to call this resolved.

whoaitsaimz: (cc: @jonboutelle fyi)

src/components/organisms/ProfileModal/UserInformationContent/UserInformationContent.tsx
denisdimitrov: > looks like <UserStatusDropdown /> breaks when userStatuses list is empty (length === 0).
maybe you were right adding this check to the < UserStatusDropdown /> component;

so we need to check both cases: userStatuses is undefined, userStatuses is []

It's probably because I'm checking if `sovereignVenue?.userStatuses` exists. Which will pass as `true` even if the array is empty. I think this should work If I check for the length instead.

denisdimitrov: @sunny-viktoryia Fixed in: https://github.com/sparkletown/sparkle/pull/1772/commits/832b2e1e6e77cfefb88accaa67edcb3cc3389988

src/hooks/useVenueUserStatuses.ts
```diff
@@ -1,30 +1,46 @@
...
 
   return {
+    changeUserStatus,
     venueUserStatuses: venueStatuses,
     isStatusEnabledForVenue:
       sovereignVenue?.showUserStatus ?? DEFAULT_SHOW_USER_STATUSES,
-    userStatus: userStatus ?? ONLINE_USER_STATUS,
+    userStatus: userStatus ?? {
+      status: profile?.status ?? "",
+      color: DEFAULT_USER_STATUS.color,
```
sunny-viktoryia, line 43: seems like we need to use here `venueStatuses[0].color` as well as per discussion: https://github.com/sparkletown/sparkle/pull/1772#discussion_r667032778.
```suggestion
      color: venueStatuses[0]?.color ?? DEFAULT_USER_STATUS.color,
```

sunny-viktoryia (approved):
Looks good. 🙃 

@denisdimitrov, please, create a new PR which would show the text of user statuses to consider the issue closed.

src/hooks/useVenueUserStatuses.ts
denisdimitrov: Correct. I just pushed the change in https://github.com/sparkletown/sparkle/pull/1772/commits/f61564ac56ee8bec1ee7997723fd34c47c21636b

sunny-viktoryia: @denisdimitrov oo, looking at your changes. Maybe it would be better to do something like this?
```
userStatus: userStatus ?? venueStatuses[0] ?? DEFAULT_USER_STATUS;
```

denisdimitrov: Just saw the comment, looks like a valid solution. We might actually be able to remove that useEffect that checks if the profile exists.

PR #1775, opened 2021-07-09
refactor legacy checkbox toggles to use our standard Toggler component

goran-peoski-work:
Should fix https://github.com/sparkletown/internal-sparkle-issues/issues/900

~~NB: mechanical replacement, not tested (yet). If there's a change to be made, will be quicker to provide a code suggestion.~~

PR #1737, opened 2021-07-06
refactor AppRouter / VenuePage / TemplateWrapper / WithNavigationBar to lazy load routes and split bundle

0xdevalias:
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/839

## Details

<details><summary>More Details</summary>

### Baseline

TODO

### After splitting `AdminSubrouter`

```
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  738.56 KB (-18.9 KB)   build/static/js/2.32bc8a33.chunk.js
  130.45 KB (-2.98 KB)   build/static/css/main.1927d0e4.chunk.css
  108.65 KB (-19.49 KB)  build/static/js/main.84a23342.chunk.js
  27.11 KB               build/static/js/4.40f15263.chunk.js
  18.54 KB (+9.05 KB)    build/static/js/3.143b9912.chunk.js
  9.48 KB                build/static/js/5.d046ae47.chunk.js
  6.19 KB                build/static/css/4.d59db4d6.chunk.css
  1.86 KB                build/static/css/2.ca756f56.chunk.css
  1.51 KB (+344 B)       build/static/js/runtime-main.8e06ef61.js
```

image

### After splitting `EnterSubrouter`

```
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  735.16 KB (-3.4 KB)  build/static/js/2.e484f69c.chunk.js
  130.44 KB (-13 B)    build/static/css/main.35e1552e.chunk.css
  105.26 KB (-3.4 KB)  build/static/js/main.c54421c1.chunk.js
  27.11 KB (-4 B)      build/static/js/4.8484ef04.chunk.js
  18.53 KB (-4 B)      build/static/js/3.b70ca74f.chunk.js
  9.48 KB              build/static/js/7.91034cbe.chunk.js
  6.19 KB              build/static/css/4.d59db4d6.chunk.css
  4.85 KB (-4.63 KB)   build/static/js/5.966421c7.chunk.js
  4.39 KB              build/static/js/6.df76cb4d.chunk.js
  1.86 KB              build/static/css/2.ca756f56.chunk.css
  1.53 KB (+25 B)      build/static/js/runtime-main.cbf23057.js
```

image

### After splitting `AccountSubrouter`

```
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  721.4 KB (+702.87 KB)  build/static/js/3.b6e5c5fa.chunk.js
[...]

PR #1781, opened 2021-07-10
quick fix for toggle/etc styles in DetailsForm in admin (v1)

0xdevalias:
`DetailsForm` seems to rely on some styles that are currently defined in `Admin.scss` (or another file with similar styles that were defined in 'global scope'). When we split our bundle in https://github.com/sparkletown/sparkle/pull/1737 (or a similar PR around that time), these styles seemed to be lost in the process.

This PR is a quick hacky fix to import the styles from `Admin.scss`, and adds some `@debt` notes to properly clean this up and separate the needed styles to the correct location in future.

fixes https://github.com/sparkletown/internal-sparkle-issues/issues/900

---

Note that there is also another PR (see below) that will fix this + also clean up associated tech debt, but given it's larger in scope, I figured getting a quick fix out sooner was better to unbreak the admin tooling:

- https://github.com/sparkletown/sparkle/pull/1775

### DIFF
--- a/src/pages/Admin/Venue/DetailsForm.tsx
+++ b/src/pages/Admin/Venue/DetailsForm.tsx
@@ -73,6 +73,9 @@ import { WizardPage } from "./VenueWizard";
 import QuestionInput from "./QuestionInput";
 import EntranceInput from "./EntranceInput";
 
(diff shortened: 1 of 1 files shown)


sunny-viktoryia (approved):
Approving 🙃

PR #1787, opened 2021-07-10
[env/github] update schedule event modal

malaikaaiyar:
partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/895

looks like this now

Screen Shot 2021-07-10 at 5 58 49 PM


0xdevalias (approved):
LGTM!

PR #1788, opened 2021-07-10
Add configuration for a new environment

sunny-viktoryia:
partially closes https://github.com/sparkletown/internal-sparkle-issues/issues/808


0xdevalias (approved):
Changes in the code LGTM. Assuming this covers all the files/etc that need changing (which I believe it does), then :shipit:

PR #1797, opened 2021-07-12
Use EmptyProvider instead of React.Fragment in Provided

goran-peoski-work:
In order not to have console errors in browsers when using `React.Fragment` like in the image,
a default no-operation `EmptyProvider` wraps `children` in `Provided`

sparkle-admin-flashes-03

### DIFF
--- a/src/components/organisms/AppRouter/Provided.tsx
+++ b/src/components/organisms/AppRouter/Provided.tsx
@@ -7,6 +7,13 @@ import {
   RelatedVenuesProviderProps,
 } from "hooks/useRelatedVenues";
 
+interface EmptyProviderProps {
+  venueId?: string;
+}
+const EmptyProvider: React.FC<EmptyProviderProps> = ({ children }) => {
+  return <>{children}</>;
+};
+
 export interface ProvidedProps {
   withWorldUsers?: boolean;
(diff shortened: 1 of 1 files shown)


src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -7,6 +7,13 @@ import {
   RelatedVenuesProviderProps,
 } from "hooks/useRelatedVenues";
 
+interface EmptyProviderProps {
```
0xdevalias, line 10:
```suggestion
export interface EmptyProviderProps {
```

src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -7,6 +7,13 @@ import {
...
+  venueId?: string;
+}
+const EmptyProvider: React.FC<EmptyProviderProps> = ({ children }) => {
```
0xdevalias, line 13:
```suggestion
}

const EmptyProvider: React.FC<EmptyProviderProps> = ({ children }) => {
```

0xdevalias (approved):
2 small fixes (see earlier suggestions) then this should be good to go.

PR #1799, opened 2021-07-12
fix admin v1 edit venue functionality

denisdimitrov:
When editing a venue with `parentId` the function updates the sovereign venue with the values from the current one, because they were provided with spread operator. Now the sovereign venue data is being passed instead.

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -636,17 +636,17 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
 
   // @debt this is missing from updateVenue_v2, why is that? Do we need it there/here?
   //   I expect this may be legacy functionality related to the Playa template?
-  if (
-    !data.placement.state ||
-    data.placement.state === PlacementState.SelfPlaced
-  ) {
-    updated.placement = {
-      ...data.placement,
-      state: PlacementState.SelfPlaced,
-    };
-  } else if (data.placementRequests) {
-    updated.placementRequests = data.placementRequests;
-  }
+  // if (
+  //   !data.placement.state ||
+  //   data.placement.state === PlacementState.SelfPlaced
+  // ) {
+  //   updated.placement = {
+  //     ...data.placement,
+  //     state: PlacementState.SelfPlaced,
+  //   };
+  // } else if (data.placementRequests) {
+  //   updated.placementRequests = data.placementRequests;
+  // }
 
   // @debt the logic here differs from updateVenue_v2, which only sets this field when data.showGrid is a boolean
   if (data.columns) {
--- a/src/pages/Admin/Venue/DetailsForm.tsx
+++ b/src/pages/Admin/Venue/DetailsForm.tsx
@@ -183,14 +183,24 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
             user
           );
 
-          if (sovereignVenueId)
+          //@debt Create separate function that updates the userStatuses separately by venue id.
+          if (sovereignVenueId && sovereignVenue)
             await updateVenue(
               {
-                ...(vals as VenueInput),
+                name: sovereignVenue.name,
+                subtitle:
+                  sovereignVenue.config?.landingPageConfig.subtitle ?? "",
+                description:
+                  sovereignVenue.config?.landingPageConfig.description ?? "",
+                adultContent: sovereignVenue.adultContent ?? false,
+                profile_questions: sovereignVenue.profile_questions,
+                code_of_conduct_questions:
+                  sovereignVenue.code_of_conduct_questions,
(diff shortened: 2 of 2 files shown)


functions/venue.js
```diff
@@ -636,17 +636,17 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
...
+  //   !data.placement.state ||
+  //   data.placement.state === PlacementState.SelfPlaced
+  // ) {
+  //   updated.placement = {
+  //     ...data.placement,
+  //     state: PlacementState.SelfPlaced,
+  //   };
+  // } else if (data.placementRequests) {
+  //   updated.placementRequests = data.placementRequests;
+  // }
```
mike-lvov, line 649: Why was this commented out?

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -183,14 +183,24 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
...
+                subtitle:
+                  sovereignVenue.config?.landingPageConfig.subtitle ?? "",
+                description:
+                  sovereignVenue.config?.landingPageConfig.description ?? "",
+                adultContent: sovereignVenue.adultContent ?? false,
+                profile_questions: sovereignVenue.profile_questions,
+                code_of_conduct_questions:
+                  sovereignVenue.code_of_conduct_questions,
                 id: sovereignVenueId,
                 parentId: sovereignVenue?.parentId,
```
mike-lvov, line 200: sovereignVenue can't have a parentId

denisdimitrov:
@dmitrii-zolotukhin Please check this PR for the fix I provided and if you think there is something else that needs to be added feel free to give your input.

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -183,14 +183,24 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
...
-                ...(vals as VenueInput),
+                name: sovereignVenue.name,
+                subtitle:
+                  sovereignVenue.config?.landingPageConfig.subtitle ?? "",
+                description:
+                  sovereignVenue.config?.landingPageConfig.description ?? "",
+                adultContent: sovereignVenue.adultContent ?? false,
+                profile_questions: sovereignVenue.profile_questions,
+                code_of_conduct_questions:
+                  sovereignVenue.code_of_conduct_questions,
```
mike-lvov, line 198: Why does it update sovereignVenue, not the venue?

mike-lvov (commented):
A few questions

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: User statuses are stored in the sovereignVenue that way the whole world can share them. This is why after I update the current venue I need to update the sovereignVenue with the userStatuses only.

functions/venue.js
denisdimitrov: This is playa legacy code and it blows up for some of the venues that were created from the admin v3

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: Good point. I will change it.

src/pages/Admin/Venue/DetailsForm.tsx
mike-lvov: Just to be clear. What if you don't pass the field here? Will it be erased?

mike-lvov: checked the logic, shouldn't do it. Only for the specific fields such as this
image

mike-lvov: checked the logic, shouldn't do it. Only for the specific fields such as this
image

mike-lvov (approved):
Looks a bit hacky, but will work

denisdimitrov:
@mike-lvov I just removed the `parentId` from the sovereignVenue now it should work properly.

src/pages/Admin/Venue/DetailsForm.tsx
denisdimitrov: @mike-lvov It's the types that actually force me to pass these fields, I can make them optional. If that's okey.

functions/venue.js
0xdevalias: If we haven't already can we please `@debt` it to explain that.

denisdimitrov: @0xdevalias There is debt about it, explaining that this is playa legacy code. We should remove these in the near future.

PR #1800, opened 2021-07-12
Fixed sovereignVenue background 

dmitrii-zolotukhin:
- [x] SovereignVenue background does not change when we set new background for subVenue


mike-lvov:
duplicates https://github.com/sparkletown/sparkle/pull/1799

PR #1802, opened 2021-07-12
hide back button in NavBar within admin panel

goran-peoski-work:
Admin v1 shows the `BackButton` in the `NavBar` which isn't supposed to.

This PR adds `hasBackButton={false}` for the Admin v1 use cases.

image

### DIFF
--- a/src/components/molecules/NavBar/NavBar.tsx
+++ b/src/components/molecules/NavBar/NavBar.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useMemo, useRef, useCallback } from "react";
-import { useHistory, useRouteMatch } from "react-router-dom";
+import { useHistory } from "react-router-dom";
 import { OverlayTrigger, Popover } from "react-bootstrap";
 
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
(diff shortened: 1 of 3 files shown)


src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -101,6 +101,9 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
 
   const isOnPlaya = pathname.toLowerCase() === venueInsideUrl(PLAYA_VENUE_ID);
 
+  // @debt remove when back button is removed from NavBar and/or admin v1 is no longer used
+  const isInAdmin = useRouteMatch("/admin/");
```
goran-peoski-work, line 105: [NB] Not using a constant in `settings.ts` as to contain  the `@debt` for this particular case in this particular line/file and not spread it further.

mike-lvov (dismissed):

0xdevalias (changes requested):
This is super hacky. `NavBar` already has a prop `hasBackButton` which defaults to true. In admin, that prop should be set to false.

src/components/molecules/NavBar/NavBar.tsx
goran-peoski-work: reverted change in bbcc9c48252441cff5298c00280c10ac2c3eef35

0xdevalias (approved):
Haven't checked if this is all the places, but code diff looks good now. If you're happy that it covers all the things, then LGTM!

mike-lvov:
The alternative for this PR that popped into my head is to take the BackButton from NavBar and only render it in TemplateWrapper

goran-peoski-work:
> to take the BackButton from NavBar

This is already mentioned in `@debt` comment above the use of it, but felt like being more conservative ATM would be more prudent. Just addressing the bug in hand.

PR #1807, opened 2021-07-12
[env/github] refactor AgoraRTC clients from global scope to a useMemo within TalkShowStudio

dmitrii-zolotukhin:
- [x] remove the agora client’s from the global scope


src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -47,9 +32,19 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
   const stage = useStage();
   const { userId, profile } = useUser();
   const currentVenue = useSelector(currentVenueSelectorData);
-  const remoteUsers = useAgoraRemotes({ client: remotesClient });
   const isRequestToJoinStageEnabled = venue.requestToJoinStage;
 
+  const [remotesClient, screenClient, cameraClient] = React.useMemo(() => {
```
0xdevalias, line 37:
```suggestion
  const [remotesClient, screenClient, cameraClient] = useMemo(() => {
```

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -47,9 +32,19 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
...
+    return Array.from({ length: 3 }).map(() =>
+      AgoraRTC.createClient({
+        codec: "h264",
+        mode: "rtc",
+      })
+    );
```
0xdevalias, line 43: Do these clients need to be 'closed' or otherwise marked as being no longer needed in some form? If so, should that be happening in the 'cleanup' return of this `useEffect` (it probably should be)

0xdevalias (dismissed):
Questions/changes.

dmitrii-zolotukhin:
@0xdevalias for testing that I needed to partially fixed the Talk Show Studio functional, so can I pushed there venue's fix together with comments fixes?

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
```diff
@@ -47,9 +32,19 @@ export const TalkShowStudio: FC<TalkShowStudioProps> = ({ venue }) => {
   const stage = useStage();
   const { userId, profile } = useUser();
   const currentVenue = useSelector(currentVenueSelectorData);
-  const remoteUsers = useAgoraRemotes({ client: remotesClient });
   const isRequestToJoinStageEnabled = venue.requestToJoinStage;
 
+  const [remotesClient, screenClient, cameraClient] = React.useMemo(() => {
+    return Array.from({ length: 3 }).map(() =>
```
0xdevalias, line 38: This `Array.from({ length: 3 })` is neat by the way 🎉 . I was going to call it out in my review as probably being invalid, but then when I went and looked it up and thought about it for a while, it made sense:

> array-like objects (**objects with a length property** and indexed elements); or

I also learned that `Array.from` has a built in `mapFn` param that I also never knew existed.

> `Array.from()` has an optional parameter `mapFn`, which allows you to execute a `map()` function on each element of the array being created.
> 
> More clearly, `Array.from(obj, mapFn, thisArg)`
> has the same result as `Array.from(obj).map(mapFn, thisArg)`,
> except that it does not create an intermediate array.

(cc // @sunny-viktoryia FYI: Another contender for our 'quickest way to make an array' benchmarking challenge :p )

sunny-viktoryia: wooow! Super cool! haven't heard about `mapFn` either.
hahahha I've just found our discussion about that benchmarking https://github.com/sparkletown/sparkle/pull/1307/files#r608258420

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
dmitrii-zolotukhin: @0xdevalias fixed in efa8913141c5d517382f8f25178d4967c902a3df

src/components/templates/TalkShowStudio/TalkShowStudio.tsx
dmitrii-zolotukhin: @0xdevalias fixed in efa8913141c5d517382f8f25178d4967c902a3df

src/hooks/video/agora/useAgoraRemotes.ts
```diff
@@ -29,7 +29,16 @@ export const useAgoraRemotes: ReactHook<
...
+    if (!client) {
+      setClient(
+        AgoraRTC.createClient({
+          codec: "h264",
+          mode: "rtc",
+        })
+      );
+
+      return;
+    }
```
mike-lvov, line 41: Can this cause a loop? What if there is an error in setting the client?

src/hooks/video/agora/useAgoraScreenShare.ts
```diff
@@ -68,24 +66,33 @@ export const useAgoraScreenShare: ReactHook<
...
     await client?.leave();
   }, [client, stopShare]);
 
   useEffect(() => {
+    setClient(
+      AgoraRTC.createClient({
+        codec: "h264",
+        mode: "rtc",
+      })
+    );
```
mike-lvov, line 83: same

src/hooks/video/agora/useAgoraCamera.ts
```diff
@@ -75,19 +76,29 @@ export const useAgoraCamera: ReactHook<
...
     await client?.leave();
   }, [client, localCameraTrack, localMicrophoneTrack]);
 
   useEffect(() => {
+    setClient(
+      AgoraRTC.createClient({
+        codec: "h264",
+        mode: "rtc",
+      })
+    );
```
mike-lvov, line 89: same

mike-lvov (commented):
A few questions

src/hooks/video/agora/useAgoraRemotes.ts
dmitrii-zolotukhin: @mike-lvov there shouldn't be the loop as we get the object with client in any case. As a variant, I can put here the creating of client in useEffect with empty dependencies. After the calling to createClient function need to check if we have client or not, if not so use Bugsnag. I didn't find any other extra API, only this one but if I understand correctly we are talking about the connection of the client to the channel and all info is sending to the servers of Agora. https://docs.agora.io/en/All/API%20Reference/web/modules/agorartc.logger.html#enablelogupload

src/hooks/video/agora/useAgoraScreenShare.ts
dmitrii-zolotukhin: @mike-lvov 
As a variant, after the calling to `createClient` function checking if we have client or not, if not so use Bugsnag. I didn't find any other extra API, only this one but if I understand correctly we are talking about the connection of the client to the channel and all info is sending to the servers of Agora. https://docs.agora.io/en/All/API%20Reference/web/modules/agorartc.logger.html#enablelogupload

src/hooks/video/agora/useAgoraCamera.ts
dmitrii-zolotukhin: @mike-lvov As a variant, after the calling to createClient function checking if we have client or not, if not so use Bugsnag. I didn't find any other extra API, only this one but if I understand correctly we are talking about the connection of the client to the channel and all info is sending to the servers of Agora. https://docs.agora.io/en/All/API%20Reference/web/modules/agorartc.logger.html#enablelogupload

mike-lvov (approved):
LGTM

PR #1809, opened 2021-07-12
Adds conditional/lazy loading for long lists

mike-lvov:
Adds conditional/lazy loading for long lists

### DIFF
--- a/src/components/atoms/ChatMessage/ChatMessage.scss
+++ b/src/components/atoms/ChatMessage/ChatMessage.scss
@@ -17,6 +17,8 @@ $reply-icon-radius: 20px;
   flex-direction: column;
   margin-bottom: $spacing--lg;
 
+  content-visibility: auto;
+
   &:hover {
     .ChatMessage__reply-icon {
       display: flex;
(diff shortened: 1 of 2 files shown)


denisdimitrov (approved):
LGTM!

src/components/atoms/ChatMessage/ChatMessage.scss
```diff
@@ -17,6 +17,8 @@ $reply-icon-radius: 20px;
   flex-direction: column;
   margin-bottom: $spacing--lg;
 
+  content-visibility: auto;
```
0xdevalias, line 20: @mike-lvov FYI: This seems to cause performance issues and pins my GPU at 50%+ processing even when just idling. Commenting out this line removes the issues. See further details in this thread:

- https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1626669763475800

Before commenting out the line:
image

Each of those little marks zoomed way in:
image

Zooming out a little we can see it runs every 15ms or so:
image

After commenting out the line, when GPU usage returns to normal/barely registering:
image

PR #1811, opened 2021-07-12
[env/github] Remove online users list

mike-lvov:
Remove online users list as per 
https://doesitsparkle.slack.com/archives/C01700P4QCT/p1626105523145800?thread_ts=1625727975.139200&cid=C01700P4QCT & https://doesitsparkle.slack.com/archives/C01700P4QCT/p1626105477145600?thread_ts=1625727975.139200&cid=C01700P4QCT


sunny-viktoryia (approved):
Approving. 

But will it affect the performance significantly? as we still use `onlineUsers` in the component which I suspect causes the issue.

Also, I'd suggest **commenting** not needed functionality instead of **removing** it (when we it it for environment customization purposes). I believe it'd simplify the deployment process later.

0xdevalias:
> Also, I'd suggest **commenting** not needed functionality instead of **removing** it (when we it it for environment customization purposes). I believe it'd simplify the deployment process later.

@sunny-viktoryia Yup, 100%. I’ve been commenting + adding a debt comment explaining why.

PR #1815, opened 2021-07-12
fix event modal url: remove all the spaces in the URL

sunny-viktoryia:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/916


yurii-lubynets (approved):
LGTM

PR #1819, opened 2021-07-13
[env/github][scripts] Script user reports

ermiaSparkle:
temporal script to export users reports

### DIFF
--- a/scripts/get-venue-users-reports.ts
+++ b/scripts/get-venue-users-reports.ts
@@ -1,66 +1,35 @@
 #!/usr/bin/env node -r esm -r ts-node/register
-import { initFirebaseAdminApp, makeScriptUsage } from "./lib/helpers";
+
+import { resolve } from "path";
+
 import admin from "firebase-admin";
-import * as path from "path";
-import * as fs from "fs";
-import { FormatterOptionsArgs, Row, writeToStream } from "@fast-csv/format";
+import { writeToPath } from "@fast-csv/format";
 
-type CsvFileOpts = {
-  headers: string[];
-  path: string;
-};
-
-class CsvFile {
-  static write(
-    stream: NodeJS.WritableStream,
-    rows: Row[],
-    options: FormatterOptionsArgs<Row, Row>
-  ): Promise<void> {
-    return new Promise((res, rej) => {
-      writeToStream(stream, rows, options)
-        .on("error", (err: Error) => rej(err))
-        .on("finish", () => res());
-    });
-  }
-
-  private readonly headers: string[];
-
-  private readonly path: string;
-
-  private readonly writeOpts: FormatterOptionsArgs<Row, Row>;
-
-  constructor(opts: CsvFileOpts) {
-    this.headers = opts.headers;
-    this.path = opts.path;
-    this.writeOpts = { headers: this.headers, quote: true };
-  }
-
-  create(rows: Row[]): Promise<void> {
-    return CsvFile.write(fs.createWriteStream(this.path), rows, {
-      ...this.writeOpts,
-    });
-  }
-}
+import { initFirebaseAdminApp, makeScriptUsage } from "./lib/helpers";
 
 const usage = makeScriptUsage({
-  description: "Bulk upload events from a spreadsheet ",
+  description: "Generate a CSV report of all users on the platform environment",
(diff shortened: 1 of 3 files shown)


scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env node -r esm -r ts-node/register
```
0xdevalias, line 1:
```suggestion
#!/usr/bin/env node -r esm -r ts-node/register

```

scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.path = opts.path;
+    this.writeOpts = { headers: this.headers, quote: true };
+  }
+
+  create(rows: Row[]): Promise<void> {
+    return CsvFile.write(fs.createWriteStream(this.path), rows, {
+      ...this.writeOpts,
+    });
+  }
+}
```
0xdevalias, line 43: This seems super convoluted/unnecessary.

scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
...
+    "Last Sign In",
+  ];
+
+  const csvFile = new CsvFile({
+    path: path.resolve(
+      __dirname,
+      "users-report-" + new Date().getTime() + ".csv"
+    ),
+    headers: headers,
+  });
```
0xdevalias, line 99: Why do we need a whole class object here? We can just resolve the path to a `const`.

scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  while (nextPageToken) {
+    const { users, pageToken } = await admin
+      .auth()
+      .listUsers(1000, nextPageToken);
+    allUsers.push(...users);
+    nextPageToken = pageToken;
+  }
+
+  let headers = [
```
0xdevalias, line 79:
```suggestion
  const headers = [
```

scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
...
+  allUsers.push(...users);
+  nextPageToken = pageToken;
+
+  while (nextPageToken) {
+    const { users, pageToken } = await admin
+      .auth()
+      .listUsers(1000, nextPageToken);
+    allUsers.push(...users);
+    nextPageToken = pageToken;
+  }
```
0xdevalias, line 77: We shouldn't be reading all users within the auth database like this. There are better patterns in `scripts/get-badges.ts` script for how to handle this and only get the explicit needed users.

0xdevalias: Why are you listing `10` users at first, and then `1000`?

scripts/get-venue-users-reports.ts
```diff
@@ -0,0 +1,136 @@
...
+      [headers[5]]: Q1,
+      [headers[6]]: Q2,
+      [headers[7]]: Q3,
+      [headers[8]]: Q4,
+      [headers[9]]: Q5,
+      [headers[10]]: lastSignInTime
+        ? new Date(lastSignInTime).toISOString()
+        : "ٔٔNever",
+    });
+  });
```
0xdevalias, line 130: Instead of using `let` for `rows`, then `.forEach` iterating over the results from firebase and then mutating `rows`; you should just be following our standard patterns of using `.map`.

That way there is no need for mutations, and the code is simpler/cleaner/easier to cognitively follow.

0xdevalias (changes requested):
Questions/comments/changes.

---

**Edit:**

> @ermiaSparkle I refactored and cleaned up this script in https://github.com/sparkletown/sparkle/pull/1819/commits/4ad2f6f3d5ac7d59ffc95ef5941991b76737b023
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1819#issuecomment-878869506_

0xdevalias:
@ermiaSparkle I refactored and cleaned up this script in https://github.com/sparkletown/sparkle/pull/1819/commits/4ad2f6f3d5ac7d59ffc95ef5941991b76737b023

mike-lvov:
@ermiaSparkle do we still need this or I can close it?

PR #1820, opened 2021-07-13
Stop event propagation of ChatMessageInfo delete icon

goran-peoski-work:


src/components/atoms/ChatMessageInfo/ChatMessageInfo.tsx
```diff
@@ -33,6 +33,14 @@ export const ChatMessageInfo: React.FC<ChatMessageInfoProps> = ({
     openUserProfileModal(author);
   }, [openUserProfileModal, author]);
 
+  const handleDelete = useCallback(
+    (e: SyntheticEvent) => {
+      e.stopPropagation();
```
sunny-viktoryia, line 38: we try to avoid using `e.stopPropagation()`. we tend to use `e.target.closest()` instead where possible. (you can find the examples in the schedule - `ScheduleEvent` click)

goran-peoski-work: done 074df636ec32a2e30c4560e4f98709dc5cbd6227

sunny-viktoryia (approved):
Looks good! Thanks for the quick fix! ❤️

PR #1822, opened 2021-07-13
fixed thread reply state reset after sending a message

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/918
Thread reply state now resets after sending a message to a thread

### DIFF
--- a/src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
+++ b/src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
@@ -25,6 +25,7 @@ export interface ChatMessageBoxProps {
   sendThreadReply: SendChatReply;
   unselectOption: () => void;
   isQuestion?: boolean;
+  closeThread: () => void;
 }
 
 export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
@@ -32,6 +33,7 @@ export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
   sendMessage,
   sendThreadReply,
   unselectOption,
+  closeThread,
   isQuestion = false,
 }) => {
   const hasChosenThread = selectedThread !== undefined;
@@ -77,6 +79,7 @@ export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
     sendThreadReply({ replyText: message, threadId: selectedThread.id });
     reset();
     unselectOption();
+    closeThread();
   });
 
   const {
--- a/src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
+++ b/src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
@@ -7,7 +7,7 @@ import { faPaperPlane, faSmile } from "@fortawesome/free-solid-svg-icons";
 
 import { CHAT_MESSAGE_TIMEOUT } from "settings";
 
-import { MessageToDisplay, SendChatReply, SendMessage } from "types/chat";
+import { MessageToDisplay, SendChatReplyProps, SendMessage } from "types/chat";
 
 import { WithId } from "utils/id";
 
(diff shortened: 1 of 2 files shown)


src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
```diff
@@ -25,13 +25,15 @@ export interface ChatMessageBoxProps {
...
   isQuestion?: boolean;
+  closeThread: () => void;
 }
 
 export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
   selectedThread,
   sendMessage,
   sendThreadReply,
   unselectOption,
+  closeThread,
```
denisdimitrov, line 36: React's naming pattern for callbacks is with `on...` prefix. Modals have `onHide`, buttons have `onClick`, `onMouseEnter` etc... We should be following the same pattern when it comes to our components as well. If you rename this to `onReplyToThread` and I see the usage of the component I will not only know what it does but where it does it.

This shows what the prop does.
`<ChatMessageBox closeThread={closeThread}`

This shows where the prop is used for + what it does because of the function.
`<ChatMessageBox onReplyToThread={closeThread}`

@mike-lvov @0xdevalias Some input here? Do you agree with me.

denisdimitrov (changes requested):

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
0xdevalias: @denisdimitrov This is passing through an explicit action isn't it? This seems to align with the existing API here: `sendMessage, sendThreadReply, unselectOption`

0xdevalias: I haven't looked too deeply into this code, but I would probably have expected the `sendMessage` / `unselectOption` / etc to have been composed into a single handler 'higher up' in the app probably, rather than passing them all through just to compose them again within a handler here. But that's probably more of a refactor than we should be getting into there.

denisdimitrov: > This is passing through an explicit action isn't it?

Yes, it is. But usually naming it with `on` prefix will also indicate where and how this action is actually used.

denisdimitrov: If we create `ChatMessageButton` it should accept `onClick` not `sendMessage` prop. This should apply for all of our components not just buttons and modals.

denisdimitrov: >I haven't looked too deeply into this code, but I would probably have expected the sendMessage / unselectOption / etc to have been composed into a single handler 'higher up' in the app probably

This is exactly what I mean, if the component itself accepts `onReply` then you will be 'forced' to create this handler higher and pass only a single function.

alexey-petrov: @denisdimitrov refactored this part in 0a87b37

alexey-petrov:
@denisdimitrov please re-review

denisdimitrov:
@alexey-petrov Looks good. I dismissed my change request on purpose because it was just a conversation.

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
```diff
@@ -74,9 +74,9 @@ export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
     if (!selectedThread) return;
 
     setMessageSending(true);
-    sendThreadReply({ replyText: message, threadId: selectedThread.id });
     reset();
-    unselectOption();
+    onReplyToThread({ replyText: message, threadId: selectedThread.id });
+    reset();
```
denisdimitrov, line 79: No need for reset. It's already called 2 lines above.

denisdimitrov (changes requested):

denisdimitrov (approved):
LGTM!

PR #1823, opened 2021-07-13
render LoadingPage while we wait for the redirect to complete in TemplateWrapper instead of showing invalid error

0xdevalias:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/661

Previously while waiting for the redirect, `TemplateWrapper` would 'fall through' and show the 'no venue' error, even when it eventually got there. We now `return <LoadingPage />` after starting the redirect to prevent this.


denisdimitrov (approved):
LGTM! Tested it locally, looks good and better than staging.

PR #1824, opened 2021-07-13
refactor VenueChat and related for performance

0xdevalias:
Cleans up some tech debt, makes more effective use of memoing in `useSelector` in places, and uses `React.memo` to create 'pure' functional components.

This was all driven from using `whyDidYouRender` + the react performance profiler + `useDebugWarnLog` to track down and isolate the areas needing to be fixed up.

### DIFF
--- a/src/hooks/useRoles.ts
+++ b/src/hooks/useRoles.ts
@@ -4,20 +4,24 @@ import { useSelector } from "./useSelector";
 
 export const useRoles = () => {
   const { user } = useUser();
+
   useFirestoreConnect({
     collection: "roles",
     where: [["users", "array-contains", user?.uid || ""]],
     storeAs: "userRoles",
   });
+
   useFirestoreConnect({
     collection: "roles",
     where: [["allowAll", "==", true]],
     storeAs: "allowAllRoles",
   });
-  const { userRoles, allowAllRoles } = useSelector((state) => ({
-    userRoles: state.firestore.data.userRoles,
-    allowAllRoles: state.firestore.data.allowAllRoles,
-  }));
+
+  const userRoles = useSelector((state) => state.firestore.data.userRoles);
+
+  const allowAllRoles = useSelector(
+    (state) => state.firestore.data.allowAllRoles
+  );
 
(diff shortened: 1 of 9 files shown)


src/hooks/useRoles.ts
```diff
@@ -1,48 +1,60 @@
...
+    return {
+      isUserRolesLoaded: isLoaded(userRoles),
+      userRoles: rolesForUser,
+
+      isAllowAllRolesLoaded: isLoaded(allowAllRoles),
+      allowAllRoles: rolesForAll,
+
+      isRolesLoaded: isLoaded(userRoles) && isLoaded(allowAllRoles),
+      roles: combinedRoles,
+    };
```
mike-lvov, line 49: This can actually break some of the logic. As far as I could understand, it would return userRoles/allowAllRoles as undefined if it was undefined and as an empty array if it was null. (NOTE: null means no data, undefined means data not loaded in Firestore language)

mike-lvov: The new change will return empty array for both `undefined` and `null`

0xdevalias: Yeah, but I went and looked around at all the usages and IMO nowhere even used it. Would be good for you to check and confirm whether you agree with that.

mike-lvov: Confirming that it won't break anything. A bit weird piece in 
image

@0xdevalias 
NOTE: It seems like all the places where it's used just needs `isAdmin` boolean

0xdevalias: > NOTE: It seems like all the places where it's used just needs isAdmin boolean

Yeah, that's pretty much what I observed as well.

0xdevalias: That `useRoles` + `useIsAdminUser` is probably people not understanding the hooks, as I believe they both read from the exact same place (though via separate firestore queries)

mike-lvov: Okay, then I suppose it's OOS

mike-lvov (approved):
LGTM

PR #1828, opened 2021-07-13
[env/github] changed footer url link to sparklespaces

alexey-petrov:
Slack convo: https://doesitsparkle.slack.com/archives/C0229C1EC86/p1626075892397900

[env/github] changed footer url link to sparklespaces

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -18,7 +18,7 @@ import sparkleNavLogo from "assets/icons/sparkle-nav-logo.png";
 import defaultMapIcon from "assets/icons/default-map-icon.png";
 import sparkleverseLogo from "assets/images/sparkleverse-logo.png";
 
-export const SPARKLE_HOMEPAGE_URL = "/in/explore/";
+export const SPARKLE_HOMEPAGE_URL = "https://sparklespaces.com/";
 export const SPARKLE_TERMS_AND_CONDITIONS_URL =
   "https://sparklespaces.com/terms-of-use/";
 export const SPARKLE_PRIVACY_POLICY =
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -18,7 +18,7 @@ import sparkleNavLogo from "assets/icons/sparkle-nav-logo.png";
 import defaultMapIcon from "assets/icons/default-map-icon.png";
 import sparkleverseLogo from "assets/images/sparkleverse-logo.png";
 
-export const SPARKLE_HOMEPAGE_URL = "/in/explore/";
+export const SPARKLE_HOMEPAGE_URL = "https://sparklespaces.com/";
```
0xdevalias, line 21: This isn't the correct fix here. We want a seperate new const to be used for the footer so that all of the other places this variable are used don't get changed.

0xdevalias (changes requested):

src/settings.ts
alexey-petrov: `SPARKLE_HOMEPAGE_URL` is not used anywhere in the app except this file

denisdimitrov: @alexey-petrov Yes, but it's used for `HOMEPAGE_URL` which is used in a couple of places. The footer component should be using a separate const, I think that's what Glenn is trying to say.

0xdevalias: > The footer component should be using a separate const, I think that's what Glenn is trying to say.

Correct. As per the context/discussion on slack where this task was first started from.

alexey-petrov: @denisdimitrov done c69590c

0xdevalias (approved):
LGTM! :shipit:

whoaitsaimz:
Probably needs to be done to staging, not just one environment

PR #1829, opened 2021-07-13
Add function that calls addAdmin() on user creation

goran-peoski-work:
Part of https://github.com/sparkletown/internal-sparkle-issues/issues/747
Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/874 with minimalistic changes to the application i.e. considers Creator = admin .

A function named `autoAdminOnRegister ` will execute each  time new user is added/authenticated to the system using
```javascript
functions.auth.user().onCreate()
```
as a trigger and will execute `addAdmin()` that adds the user id under the `roles/admin/users` array to enable admin tools to the newly registered user.

Note:  this is an opt-in  with a `flag.autoadmin` set in the firebase environment containing truthy value that can be retrieved through `functions.config().flag.autoadmin`.


functions/venue.js
```diff
@@ -434,6 +434,13 @@ const removeAdmin = async (adminId) => {
     });
 };
 
+exports.createUser = functions.auth.user().onCreate(async (user) => {
```
denisdimitrov, line 437: This is not being used anywhere in our code, you should add documentation about how this works and why it's here, otherwise someone will see no usage tomorrow and remove it.

functions/venue.js
```diff
@@ -434,6 +434,13 @@ const removeAdmin = async (adminId) => {
...
+  const flag = functions.config().flag;
+  if (flag && flag.autoadmin) {
+    await addAdmin(user.uid);
+  }
+});
```
0xdevalias, line 442: We probably don't want to be doing this in `functions/venue.js` as it's not really related to venues. It's more of an 'auth' type thing, so should be in that file.

The name of the function doesn't make sense, we're not creating the user, so shouldn't call it `createUser`. Maybe something like `autoMakeUserAdmin` or similar.

It's probably worth noting this in a comment near this function:

> A Cloud Functions event is not triggered when a user signs in for the first time using a custom token.

You've made changes to the `functions.config()`, yet not included any of those details within this PR. There's a `*.example` file in `functions/` that will need to be updated as well.

Most of the other backend code here seems to read the functions.config() as a `const` outside of the handler itself. Any particular reason you chose to do it this way?

0xdevalias (changes requested):
Questions/comments/changes.

functions/venue.js
0xdevalias: https://firebase.google.com/docs/functions/auth-events

@denisdimitrov This is a different type of function to most of the backend ones. There will be no usage in our code as it is an event trigger.

functions/venue.js
goran-peoski-work: > Any particular reason you chose to do it this way?

in case of a missing setting `flag.autoadmin`, should not be necessary once I update `*.example`

0xdevalias: And to preempt one of the likely next questions here, `addAdmin` and `removeAdmin` should be refactored into a new location. Possibly `functions/src/api/roles`. Then imported back into `functions/venue.js` for the things that need it there, and then also used from `functions/auth.js` or wherever this trigger function ends up when moved as per my comments above.

0xdevalias: It would also be good to add some logging output here, so we have some records of what is happening.

```javascript
const functions = require("firebase-functions");

// ..snip..

functions.logger.log("Foo:", bar);
```

At a minimum we would want something that mentions the userId and explains that they are being made an admin. Possibly also log with the email address to make it easier to search for?

functions/venue.js
denisdimitrov: Yes, I read the description of the PR. But it's not documented in the code itself. And when you see something that's not being used anywhere, you remove it....

goran-peoski-work: Added comment in 0774226691b386bb6d2e7e37fa6dc140838def6b

functions/venue.js
goran-peoski-work: Added adjustments in 0774226691b386bb6d2e7e37fa6dc140838def6b

functions/src/utils/user.js
```diff
@@ -0,0 +1,17 @@
+const admin = require("firebase-admin");
+
+/** Add a user to the list of admins
+ *
+ * @param {string} newAdminId
+ */
+const addAdmin = async (newAdminId) => {
```
goran-peoski-work, line 7: [NB] moved `addAdmin` to separate utility file as to not create direct dependency between `auth.js` and `venue.js` files that both `require()` it

functions/venue.js
```diff
@@ -406,20 +407,6 @@ const dataOrUpdateKey = (data, updated, key) =>
     typeof updated[key] !== "undefined" &&
     updated[key]);
 
-/** Add a user to the list of admins
- *
- * @param {string} newAdminId
- */
-const addAdmin = async (newAdminId) => {
```
goran-peoski-work, line 413: [NB] moved addAdmin to separate utility file as to not create direct dependency between auth.js and venue.js files that both require() it

functions/auth.js
```diff
@@ -225,3 +226,25 @@ exports.connectI4AOAuthHandler = functions.https.onRequest(async (req, res) => {
...
+ *    - A user signs in for the first time using a federated identity provider.
+ *    - The developer creates an account using the Firebase Admin SDK.
+ *    - A user signs in to a new anonymous auth session for the first time.
+ *
+ *  NOTE: A Cloud Functions event is not triggered when a user signs in for the first time using a custom token.
+ *
+ *  @see https://firebase.google.com/docs/functions/auth-events
+ */
+exports.autoAdminOnRegister = functions.auth.user().onCreate(async (user) => {
+  const flag = functions.config().flag || {};
```
goran-peoski-work, line 245: [NB] using guard ` || {} ` as to not force older environments retrofit to have a flag object until necessary in their respective cases

functions/.runtimeconfig.json.example
```diff
@@ -14,5 +14,8 @@
   "agora": {
     "app_id": "",
     "app_certificate": ""
+  },
+  "flag": {
+    "autoadmin": ""
   }
```
goran-peoski-work, line 20: An example of default/empty `flag` object that can be used to switch the _Auto Admin On Register_ functionality per Firestore environment. Any truthy value should suffice.

functions/venue.js
0xdevalias: I don't see the logging? https://github.com/sparkletown/sparkle/pull/1829#discussion_r668753678

0xdevalias: Nor did you follow this one: https://github.com/sparkletown/sparkle/pull/1829#discussion_r668750203

functions/src/utils/user.js
0xdevalias: Please re-read my original request in https://github.com/sparkletown/sparkle/pull/1829#discussion_r668750203 about moving both `addAdmin` and `removeAdmin` into `api/roles`

0xdevalias (changes requested):
Please re-read my original comments/requests in https://github.com/sparkletown/sparkle/pull/1829#discussion_r668750203 + https://github.com/sparkletown/sparkle/pull/1829#discussion_r668753678 etc that weren't done.

functions/src/utils/user.js
goran-peoski-work: moved 3b1e532e09e585f60dd5b0c919666db539bb8f73

functions/venue.js
goran-peoski-work: moved functions 3b1e532e09e585f60dd5b0c919666db539bb8f73 and added logging 3c7148def4b25bdfe8609c4a4cec9b3b3ca91019

goran-peoski-work: added email to log output c005a930ce964aed5763d95dd152e05ba3859833

mike-lvov (approved):
The functions were moved to api/roles. The logger fn was added

PR #1831, opened 2021-07-13
fixed readme links

alexey-petrov:
fixed readme links


sunny-viktoryia:
@alexey-petrov there is a PR created by @0xdevalias which does pretty similar things. https://github.com/sparkletown/sparkle/pull/1818

0xdevalias:
> there is a PR created by @0xdevalias which does pretty similar things. #1818

@alexey-petrov @sunny-viktoryia 💯 Why would we duplicate it rather than just getting that merged properly?

PR #1833, opened 2021-07-13
refactor user update fetching for performance + introduce new redux patterns

0xdevalias:
This PR:

- introduces new Redux patterns using Redux Toolkit + Redux Toolkit Query + cleans up our redux implementation a bit
- removes the legacy `react-redux-firebase` / `redux-firestore` (aka: `useFirestoreConnect`) implementation of `worldUsers`
- introduces a new Redux Toolkit Query based implementation of `worldUsers` (called `worldUsersApi`) + refactor our existing hooks to use that instead
  - note: there are likely future optimisations/tuning that we can do based on this to improve how the app reads data from this new store
- refactored the `User` type a bit to add `BaseUser` + ensure that `UserLocation` attributes can only ever be included on a `UserWithLocation`, not a basic `User`
- and maybe some other little bits and pieces along the way ([see commits](https://github.com/sparkletown/sparkle/pull/1833/commits))


src/index.tsx
```diff
@@ -108,30 +96,6 @@ const rrfConfig = {
...
-  composeWithDevTools(
-    applyMiddleware(
-      thunkMiddleware,
-      LogRocket.reduxMiddleware() // logrocket needs to be last
-    )
-  )
-);
-
-export type AppDispatch = typeof store.dispatch;
-
```
0xdevalias, line 134: [note] Moved into `src/store/store.ts`

package.json
```diff
@@ -70,7 +70,6 @@
     "redux": "^4.1.0",
     "redux-devtools-extension": "^2.13.8",
     "redux-firestore": "sparkletown/redux-firestore#fix/introduce-batch-store-updates-on-addition",
-    "redux-thunk": "^2.3.0",
```
0xdevalias, line 73: [note] No longer needed as a direct dependency because `@reduxjs/toolkit` handles it for us.

src/hooks/useDispatch.ts
```diff
@@ -1,9 +1,6 @@
...
-import { ThunkDispatch } from "redux-thunk";
-import { RootActions } from "store/actions";
-import { RootState } from "index";
 
-export const useDispatch = () =>
-  _useDispatch<ThunkDispatch<RootState, void, RootActions>>();
-export type Dispatch = ReturnType<typeof useDispatch>;
+import { AppDispatch } from "store";
+
+export const useDispatch = () => _useDispatch<AppDispatch>();
```
0xdevalias, line 6: [note] Refactored to use properly exported types rather than needing to manually reconstruct them here ourselves.

src/store/store.ts
```diff
@@ -0,0 +1,109 @@
+import { configureStore, combineReducers, Reducer } from "@reduxjs/toolkit";
```
0xdevalias, line 1: [note] A lot of this `src/store/store.ts` file was just refactored here from `src/index.tsx` + updated to use the syntax/helpers/etc that come with `@reduxjs/toolkit`

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,112 @@
...
+        // TODO: use api/* layer for this?
+        const usersCollectionRef = firebase.firestore().collection("users");
+
+        const worldUsersQuery = usersCollectionRef.where(
+          "enteredVenueIds",
+          "array-contains-any",
+          relatedLocationIds
+        );
+
+        const unsubscribeListener = worldUsersQuery.onSnapshot((snapshot) => {
```
0xdevalias, line 39: [note] This `worldUsersApi` is where all of the 'connect to firestore snapshot listeners'/etc stuff is happening now. It makes use of redux toolkit query to abstract/encapsulate a lot of the boilerplate/etc we would otherwise need to write.

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,112 @@
...
+  }),
+});
+
+// TODO: https://redux-toolkit.js.org/rtk-query/api/created-api/hooks
+// TODO: https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#skiptoken
+export const {
+  useWorldUsersQuery,
+  useLazyWorldUsersQuery,
+  usePrefetch,
+} = worldUsersApi;
```
0xdevalias, line 112: [note] Note that redux toolkit query even generates React hooks for us to use.

src/hooks/users/useWorldUsers.tsx
```diff
@@ -46,24 +49,37 @@ export const WorldUsersProvider: React.FC<WorldUsersProviderProps> = ({
...
+  //       where: ["enteredVenueIds", "array-contains-any", relatedLocationIds],
+  //       storeAs: "worldUsers",
+  //     },
+  //   ];
+  // });
+
+  const worldUsersQueryResult = useWorldUsersQuery(
+    shouldConnect && venueId ? { relatedLocationIds: [venueId] } : skipToken
+    // { selectFromResult: (result) => result?.data?.userIds }
+  );
```
0xdevalias, line 74: [note] While this is obviously fairly placeholder/MVP for now.. this is basically how we can connect to the `worldUsersApi` data we defined earlier in `src/store/api/worldUsers.ts`

0xdevalias (commented):
Just a few notes to lay out the main areas/parts of this PR so far and give a bit of an overview of things.

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,112 @@
...
+
+              case "removed":
+                // TODO
+                // updateCachedData((draft) => {
+                //   // draft.push(JSON.parse(event.data));
+                // });
+                break;
+            }
+          });
+        });
```
0xdevalias, line 96: [note] The `draft` param of `updateCachedData` uses Immer for efficiently handling immutable updates in a simple way. You can read more about it and how it works here:

- https://immerjs.github.io/immer/

src/hooks/users/useWorldUsers.tsx
```diff
@@ -41,44 +41,38 @@ export const WorldUsersProvider: React.FC<WorldUsersProviderProps> = ({
...
+    return [sovereignVenueId, venueId];
+  }, [isSovereignVenueLoading, sovereignVenueId, venueId]);
+
+  // TODO: should we refactor this to use useWorldUsersQuerySubscription rather than connecting it to receive data updates?
+  //   It seems that useWorldUsersQuerySubscription doesn't return originalArgs, so we can't use it in the same way
+  //   as we are currently doing things.. but we could potentially just pass the same args object into the hook as what we pass
+  //   down for the descendant code to use.. which would work around that 'limitation'
+  // TODO: https://redux-toolkit.js.org/rtk-query/usage/queries#selecting-data-from-a-query-result
+  const { originalArgs: worldUsersApiArgs } = useWorldUsersQuery(
+    relatedLocationIds ? { relatedLocationIds } : skipToken,
```
alexey-petrov, line 56: can we move this ternary expression to a const and use here?

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,193 @@
...
+        // TODO: we can also return error and/or meta here as well as data. Do we need to be able to?
+        // error: undefined,
+        // meta: undefined,
+      }),
+      async onCacheEntryAdded(
+        { relatedLocationIds },
+        { updateCachedData, cacheEntryRemoved }
+      ) {
+        // TODO: use api/* layer for this?
+        const usersCollectionRef = firebase.firestore().collection("users");
```
alexey-petrov, line 63: doesn't seem to be using a ref, so the const name is a bit misleading? might want to rename it

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,193 @@
...
+          "enteredVenueIds",
+          "array-contains-any",
+          relatedLocationIds
+        );
+
+        const unsubscribeListener = worldUsersQuery.onSnapshot((snapshot) => {
+          // TODO: implement this properly + cleanup this placeholder stuff
+          const added = snapshot
+            .docChanges()
+            .filter((docChange) => docChange.type === "added").length;
```
alexey-petrov, line 75: do we want to introduce a helper function for `added`, `removed` and `modified` ? code seems to repeat there
also, maybe move 'added', 'removed' etc strings to consts somewhere?

src/store/api/worldUsers.ts
0xdevalias: It's using a firestore collection ref, not a react ref. This is fine, and matches our naming patterns across the rest of the codebase.

src/store/api/worldUsers.ts
0xdevalias: Are you talking about in these `.filter` functions? Or..? If so, they are placeholder to be removed before this is merged properly. If the latter, can you please expand on your thoughts for what the helper function would entail.

src/store/store.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import { worldUsersApi } from "./api/worldUsers";
+import { MiscReducers, VenueTemplateReducers } from "./reducers";
+
+export const rootReducer = combineReducers({
+  firebase: firebaseReducer as Reducer<FirebaseReducer.Reducer<User>>,
+  firestore: firestoreReducer as Reducer<Firestore>,
+  ...VenueTemplateReducers,
+  ...MiscReducers,
+  [worldUsersApi.reducerPath]: worldUsersApi.reducer,
```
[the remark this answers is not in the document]
0xdevalias: I chose to put it below so that it would 'win' if there were any clashes, but in reality, there should never be allowed to be any clashes here, and if there are, we will have much bigger problems. I don't feel like we need to do anything here.

src/store/api/worldUsers.ts
alexey-petrov: yep, talking about .filter ones and moving shared logic `snapshot.docChanges().filter` into some func that would be reused for these `added`, `removed` etc consts

0xdevalias: So these specific filters will be removed. They're just debug code. Not sure that we gain anything by making the `"added"`/etc consts, TypeScript already restricts them to only valid values, and these come from the library itself, so no reason for us to make a new enum/etc.

If the lib exports an enum I could potentially use that here, but I think it's probably fine as is.

src/hooks/users/useWorldUsers.tsx
0xdevalias: For what purpose? Just moving it out of the params, or?

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,231 @@
...
+            //
+            // if (change.doc.id === currentUserId) {
+            //   updateCachedData((draft) => {
+            //     processUserDocChange(draft)(change);
+            //   });
+            // } else {
+            queuedChanges.push(change);
+            // }
+          });
+        });
```
0xdevalias, line 133: [note] Here is where we handle the main snapshot processing, with some placeholder boilerplate for ensuring that we immediately process any actions related to the current user so that we don't need to hack around cache invalidation/end up with stale data when taking some actions on the platform (eg. taking a seat)

Basically, for each change from `snapshot.docChanges()` (usually seems to only be a single change per snapshot for user updates) we just `.push` that change into the mutable `queuedChanges` array, which will then be processed later by the `processQueuedChanges` we defined earlier.

When implemented, if the change is to the current user's document, then we will just immediately process it with `processUserDocChange` within Immer's `updateCachedData` instead of adding it to `queuedChanges`.

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,231 @@
...
+        await cacheEntryRemoved;
+
+        // Unsubscribe the firestore query snapshot listener
+        unsubscribeListener();
+
+        // Clear the interval for processing the queued changes
+        clearInterval(processQueuedChangesIntervalId);
+
+        // Make sure we process any last remaining queued changes
+        processQueuedChanges();
```
0xdevalias, line 145: [note] Here we ensure that we clear the interval + drain any remaining items in the `queuedChanges` array by calling `processQueuedChanges` one last time.

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,231 @@
...
+      //   more safety.
+      draft.worldUsers.push(userWithoutLocation);
+      draft.worldUsersById[userId] = userWithoutLocation;
+      draft.worldUserLocationsById[userId] = userLocation;
+
+      return;
+    }
+
+    case "modified": {
+      const existingUserIndex = draft.worldUsers.findIndex(
```
0xdevalias, line 187: [note] `processUserDocChange` handles the actual logic for making updates to the datastore based on the 'change' data received previously. This makes use of Immer's `draft`, that is created earlier with `updateCachedData`.

0xdevalias (commented):
Some notes/overview of the 'snapshot change queued batcher' I recently implemented.

package.json
denisdimitrov: Nice 💯

src/store/api/worldUsers.ts
```diff
@@ -0,0 +1,231 @@
...
+        // TODO: move this interval into a proper const/config value somewhere
+        const processQueuedChangesIntervalId = setInterval(
+          processQueuedChanges,
+          5000
+        );
+
+        const worldUsersQuery = firebase
+          .firestore()
+          .collection("users")
+          .where("enteredVenueIds", "array-contains-any", relatedLocationIds);
```
denisdimitrov, line 84: How often is this going to be called? Is `array-contains-any` going to work properly if `relatedLocationIds` has more than 10 entries?

0xdevalias: What do you mean by ‘how many times will it be called’?

Nope, it would max out at 10, but I think the naming is sort of legacy as I think we generally work on the assumption that all users will have entered the sovereignVenue.

I’m honestly not sure if we even need to have the current venue id in there either; but I didn’t want to change the existing logic too much here.

With the proper ‘world architecture’ refactor I expect this should be a lot more obvious/straightforward.

mike-lvov (approved):
LGTM. Will require some additional updates in the future, but it's already better than the Soichi fix

src/store/api/worldUsers.ts
```diff
@@ -96,46 +96,22 @@ export const worldUsersApi = createApi({
...
-            .docChanges()
-            .filter((docChange) => docChange.type === "removed").length;
+        const unsubscribeListener = worldUsersQuery.onSnapshot((snapshot) => {
+          const snapshotSize = snapshot.size;
 
-          const modified = snapshot
-            .docChanges()
-            .filter((docChange) => docChange.type === "modified").length;
+          // NOTE: Don't delay the update of the first load.
+          if (snapshotSize > 1 && !hasLoadedDataInitially) {
```
0xdevalias, line 105: Why > 1 here? Should it be > 0?

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -59,13 +60,7 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
...
   });
 
-  // @debt this seems to cause performance issues. It should be possible to use worldUserLocationsById
-  //   from useWorldUserLocation / similar, plus some 'is recent user?' logic to directly look up the individual
-  //   user in an efficient manner.
-  const isOnline = useMemo(
-    () => recentWorldUsers.find((worldUser) => worldUser.id === user?.id),
-    [user, recentWorldUsers]
-  );
+  const isOnline = useMemo(() => !isEmpty(userLastSeenIn), [userLastSeenIn]);
```
0xdevalias, line 63: You could push this logic back into the useWorldUserLocation hook so that it can be used elsewhere. It could also potentially make sense to have some ‘is recent user’ sort of logic in there as well potentially?

src/store/api/worldUsers.ts
mike-lvov: > Why > 1 here?
It's because of how Firebase sends the data. The first snapshot always contains 1 item, which is your own profile, which you most probably fetch when you get the auth data.

> Should it be > 0?
If I do this, then the next batch which is `all the other users` will be delayed

src/components/atoms/UserAvatar/UserAvatar.tsx
mike-lvov: > You could push this logic back into the useWorldUserLocation hook so that it can be used elsewhere.
Makes sense, will do in the next performance updates as well as addressing the other `TODOs`.

> It could also potentially make sense to have some ‘is recent user’ sort of logic in there as well potentially?
I'll just expose `isOnline` from `useWorldUserLocation`

PR #1835, opened 2021-07-13
chore: adds a couple of allow options so we can embed VR

owenniblock:
### Motivation

Thinking about adding some VR stuff to a venue. Maybe some embedded Youtube VR, maybe some Hubs.

Hoping this will enable the embeds we need.


owenniblock:
Since the summit is now over - would it make sense for me to move this PR to point to a different branch? 🙂

I'm not sure if this is something you'd want to include - if not: let me know and I'll close this PR.

mike-lvov:
@owenniblock could you please create this PR based off the `staging`? There are two ways: change the base of the current PR and removing all of the irrelevant commits coming from the production or create a brand new one.

owenniblock:
I closed this - but it looks like the underlying settings code has changed in the `staging` branch 🤷 so I'm not 100% sure where the best place to put this is now (I no longer see the advanced iframe options) I guess I'll park this for now and maybe re-open if I get some spare time to test changes against the new branch.

PR #1838, opened 2021-07-14
Create NavSearchList component that uses react-window library

goran-peoski-work:
Probable fix for https://github.com/sparkletown/internal-sparkle-issues/issues/923
Part of https://github.com/sparkletown/internal-sparkle-issues/issues/747

In order to display only the necessary search results and lighten the load of requests the browsers have to make, a library `react-window` is used inside new component `<NavSearchList>` that calculates per the inline `style` pixels which items are (or soon to be) visible and only adds those elements to the DOM.

This has the benefit of a sort of lazy loading elements that should not be visible but would contain `` elements with assets the browser might load regardless.

The inline style quirks of `react-window` necessitate dynamic style calculation, so care should be taken when updating `.scss` files to keep in sync with `NavSearchList`.

14 results to be displayed:
sparkle-frozen-search-02

11 DOM elements created:
sparkle-frozen-search-01


package.json
```diff
@@ -66,6 +66,7 @@
     "react-switch": "^5.0.1",
     "react-use": "^17.2.3",
     "react-webcam": "^5.2.0",
+    "react-window": "^1.8.6",
```
goran-peoski-work, line 69: Added library solves the issue at the expense of some inline styles used.

Not super ergonomic, so it may add a bit of debt that can be repaid by searching for another.

Combined with `@types/react-window` adds under 900kB https://packagephobia.com/result?p=react-window@1.8.6

mike-lvov:
NOTE: This approach can also be used for Online Users and Chat Messages

src/components/molecules/NavSearchBar/NavSearchBar.scss
```diff
@@ -56,10 +56,4 @@
   &__search-results-number {
     font-weight: 500;
   }
-
-  &__search-results {
-    @include scrollbar;
-    overflow-y: auto;
-    max-height: $search-results-height--max;
-  }
```
goran-peoski-work, line 64: Moved to `./NavSearchList.scss`

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -194,8 +190,10 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
     clearSearch,
   ]);
 
-  const numberOfSearchResults =
-    foundRooms.length + foundEvents.length + foundUsers.length;
+  const foundItems = useMemo<NavSearchListItem[]>(
+    () => [...foundRooms, ...foundEvents, ...foundUsers],
+    [foundRooms, foundEvents, foundUsers]
+  );
```
goran-peoski-work, line 196: [NB] The memos now produce only data object, not JSX elements. That's being moved to `<NavSearchList>`

src/components/molecules/NavSearchBar/NavSearchList.scss
```diff
@@ -0,0 +1,10 @@
+@import "scss/constants.scss";
+@import "./navSearchBarConstants.scss";
+
+.NavSearchList {
+  &__list {
+    @include scrollbar;
+    overflow-y: auto;
+    max-height: $search-results-height--max;
+  }
```
goran-peoski-work, line 9: Moved from `./NavSearchBar.scss` and renamed to keep in line with the component that contains the style

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export interface NavSearchListProps {
+  items?: NavSearchListItem[];
+}
+
+export const NavSearchList: React.FC<NavSearchListProps> = ({ items = [] }) => {
+  const children = items.map(
+    ({ key, ...props }) => <NavSearchResult key={key} {...props} />,
+    [items]
+  );
```
goran-peoski-work, line 37: [NB] Opted not to include the wrapper `<div>` into the memoized components so it doesn't possibly trip some `<FixedSizeList>` logic.

 In theory it's possible to try and do more optimizations, but I guess that time might be worth spending finding other library or just reverse engineering the functionality in our own and ditching `react-window`

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { FixedSizeList } from "react-window";
+
+import "./NavSearchList.scss";
+
+import {
+  NavSearchResultProps,
+  NavSearchResult,
+} from "components/molecules/NavSearchBar/NavSearchResult";
+
+// @debt refactor this into .scss or find more ergonomic component with same desired functionality as FixedSizeList
```
goran-peoski-work, line 11: Marked with comment `@debt` due to `<FixedSizeList>` depending on inline styles to do the calculations required.

This function basically incapsulates the necessary props preparation based on the `.length` of the items

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+    0 === length
+      ? 40 // $search-dropdown-height--empty; @see navSearchBarConstants.scss
+      : Math.min(
+          480, // $search-dropdown-height--max; @see NavSearchBar.scss
+          48 * length // 2 * $search-result-padding--vertical + $search-result-image--size; @see navSearchBarConstants.scss
+        ),
+  width: "18.125rem", // $search-dropdown-width; @see navSearchBarConstants.scss
+});
+
+export type NavSearchListItem = NavSearchResultProps & { key: Key };
```
goran-peoski-work, line 27: [NB] Due to the diverse nature of the results, it's better to have the `key` generation logic done at the source of the results. and just reuse the value here.

package.json
0xdevalias: https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists lists the following:

- https://react-window.now.sh/
- https://bvaughn.github.io/react-virtualized/

0xdevalias: https://github.com/bvaughn/react-virtualized#a-word-about-react-window

> If you're considering adding react-virtualized to a project, take a look at react-window as a possible lighter-weight alternative. Learn more about how the two libraries compare here.

https://github.com/bvaughn/react-window#how-is-react-window-different-from-react-virtualized

0xdevalias: This also sounds interesting:

- https://github.com/bvaughn/react-window-infinite-loader

0xdevalias: Also looks like an interesting read: https://www.smashingmagazine.com/2020/03/infinite-scroll-lazy-image-loading-react/

goran-peoski-work: I went through all of those links before I settled to just using `react-window` out of 2 resons:

- They build on top of older versions  or the current version of `react-window` to tailor for specific cases, nothing game changing
- Their interface depends on the same inline styles used by `react-window`

So, I opted for the minimal and encapsulated way I could use `react-window` in case might find another or roll out our own that takes into account the `.scss` styles e.g. by inspecting the DOM element itself and/or preset `var(--iables)`

goran-peoski-work: All that related to the library, https://www.smashingmagazine.com/2020/03/infinite-scroll-lazy-image-loading-react/ is in line to what I was considering might replace `react-window` as a more robust solution

src/components/molecules/NavSearchBar/NavSearchList.tsx
0xdevalias: > but I guess that time might be worth spending finding other library or just reverse engineering the functionality in our own and ditching `react-window`

As per https://github.com/sparkletown/sparkle/pull/1838#discussion_r671356391

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: Let's revert this back to how it was previously.

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  itemSize: 48, // 2 * $search-result-padding--vertical + $search-result-image--size; @see navSearchBarConstants.scss
+  height:
+    0 === length
+      ? 40 // $search-dropdown-height--empty; @see navSearchBarConstants.scss
+      : Math.min(
+          480, // $search-dropdown-height--max; @see NavSearchBar.scss
+          48 * length // 2 * $search-result-padding--vertical + $search-result-image--size; @see navSearchBarConstants.scss
+        ),
+  width: "18.125rem", // $search-dropdown-width; @see navSearchBarConstants.scss
+});
```
0xdevalias, line 25: Docs: https://react-window.vercel.app/#/api/FixedSizeList

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export type NavSearchListItem = NavSearchResultProps & { key: Key };
+
+export interface NavSearchListProps {
+  items?: NavSearchListItem[];
+}
+
+export const NavSearchList: React.FC<NavSearchListProps> = ({ items = [] }) => {
+  const children = items.map(
```
0xdevalias, line 34: 'children' has a very explicit and defined usage in React (for elements nested within the JSX tags of this component), we shouldn't overload that terminology:

```suggestion
  const renderedSearchResults = items.map(
```

src/components/molecules/NavSearchBar/NavSearchList.tsx
0xdevalias, line 37: What's the point of `.map`'ing this data here like this? Would it not make more sense to just do it within the JSX block below, only actually processing the relevant items as needed, rather than the entire list always.
```suggestion
```

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <div
+          style={style}
+          key={items[index].key}
+          className="NavSearchList__item"
+        >
+          {children[index]}
+        </div>
+      )}
+    </FixedSizeList>
+  );
```
0xdevalias, line 54:
```suggestion
  return (
    <FixedSizeList
      className="NavSearchList NavSearchList__list"
      {...propsFor(items.length)}
    >
      {({ index, style }) => {
        const {key, ...searchResultProps} = items[index];
        
        return (
          <div
            key={key}
            className="NavSearchList__item"
            style={style}
          >
            <NavSearchResult {...searchResultProps} />
          </div>
        )
      }}
    </FixedSizeList>
  );
```

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import {
+  NavSearchResultProps,
+  NavSearchResult,
+} from "components/molecules/NavSearchBar/NavSearchResult";
+
+// @debt refactor this into .scss or find more ergonomic component with same desired functionality as FixedSizeList
+// values are used here due to quirks of the FixedSizeList component
+// NOTE: keep in sync with the .scss files mentioned in comments
+const propsFor = (length: number) => ({
```
0xdevalias, line 14:
```suggestion
const getFixedSizeListProps = (length: number) => ({
```

src/components/molecules/NavSearchBar/NavSearchList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  itemCount: length,
+  itemSize: 48, // 2 * $search-result-padding--vertical + $search-result-image--size; @see navSearchBarConstants.scss
+  height:
+    0 === length
```
0xdevalias, line 18: Thing you're testing on the left. Value it's being tested against on the right. ("If length is 0", not "If 0 is length")
```suggestion
    length === 0
```

src/components/molecules/NavSearchBar/NavSearchList.tsx
0xdevalias: Is there any reason we need the wrapping div? Could we just directly render the `NavSearchResult` here?

Why applying both `NavSearchList` and `NavSearchList__list` to the container here? If they are the same thing, then it should all be in `NavSearchList` and we should just delete `NavSearchList__list`

src/components/molecules/NavSearchBar/NavSearchBar.tsx
0xdevalias: I'm also not sure that there's much/any real value of memoising this combined array here, based on how it's used in the latter component.

0xdevalias (changes requested):
Questions/comments/changes.

src/components/molecules/NavSearchBar/NavSearchList.tsx
goran-peoski-work: Why name a non getter functions with get? Shouldn't be more `calculateFixedSizeListProps`?

src/components/molecules/NavSearchBar/NavSearchList.tsx
goran-peoski-work: I add `NavSearchList` at the top level element of the component so that non-local style selectors can refer to children components e.g. `.NavSearchResult p` (or any element in `children`) while the `&_list` and `&_item` refer to known elements of the component itself.

The wrapping `div` is so that `react-window` will work its inline styles without interfering with the styles needed for the item components. Should `react-window` be replaced with something like https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API that `div` should be removed.

When I return to doing the required changes to this PR,  I will make an attempt of replacing with `IntersectionObserver` thus rendering the `div` addition and similar decisions moot.

src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -27,12 +27,14 @@ import { useRelatedVenues } from "hooks/useRelatedVenues";
...
 import { Loading } from "components/molecules/Loading";
+import {
+  NavSearchList,
+  NavSearchListItem,
+} from "components/molecules/NavSearchBar/NavSearchList";
 import { InputField } from "components/atoms/InputField";
 
 import navDropdownCloseIcon from "assets/icons/nav-dropdown-close.png";
 
-import { NavSearchResult } from "./NavSearchResult";
```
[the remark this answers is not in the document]
goran-peoski-work: IDE logic, good catch.

PR #1843, opened 2021-07-14
[env/github] Fixed removing parentId

dmitrii-zolotukhin:
- [x] Fixed: You cannot remove parentId from admin v1 when editing venue

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -346,7 +346,7 @@ const createBaseUpdateVenueData = (data, updated) => {
     updated.bannerMessage = data.bannerMessage;
   }
 
-  if (data.parentId) {
+  if (typeof data.parentId === "string") {
     updated.parentId = data.parentId;
   }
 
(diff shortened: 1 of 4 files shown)


denisdimitrov:
@dmitrii-zolotukhin We should add this to `staging` as well.

denisdimitrov:
There is still missing code in this PR. Removing the `parentId` is not enough because the sovereign venue would still stay the same.

dmitrii-zolotukhin:
> @dmitrii-zolotukhin We should add this to `staging` as well.

@denisdimitrov done in https://github.com/sparkletown/sparkle/pull/1851

dmitrii-zolotukhin:
When I was working on that, I also faced with some problems, which I described below:
1. sovereignVenue doesn't reset when you delete it from input in the venue editing form - fixed
2. After resetting in redux was incorrect sovereignVenue - fixed
3. Checking on existing of sovereignVenue in DB - fixed
4. Reset of variable formError when entering new value in parentId input after appearing of error from the point 3 - fixed
5. Situation when a user enters value in parentId input equal to id of venue itself (in other words - venue eId) - fixed

mike-lvov:
is the duplicate of https://github.com/sparkletown/sparkle/pull/1851

PR #1844, opened 2021-07-14
[env/github] increase loc update for user location

mike-lvov:
increase loc update for user location


src/settings.ts
```diff
@@ -1,4 +1,6 @@
 import { CSSProperties } from "react";
+import hoursToMilliseconds from "date-fns/hoursToMilliseconds";
+import minutesToMilliseconds from "date-fns/minutesToMilliseconds";
```
0xdevalias, line 3:
```suggestion
import { minutesToMilliseconds, hoursToMilliseconds } from "date-fns";
```

0xdevalias (approved):
1 tiny change, then LGTM!

PR #1845, opened 2021-07-14
[env/github] nuclear users cache fix

mike-lvov:
Adds cache

### DIFF
--- a/src/components/organisms/WithNavigationBar/WithNavigationBar.tsx
+++ b/src/components/organisms/WithNavigationBar/WithNavigationBar.tsx
@@ -8,7 +8,18 @@ import { RelatedVenuesProvider } from "hooks/useRelatedVenues";
 import { Footer } from "components/molecules/Footer";
 import { Loading } from "components/molecules/Loading";
 
+import { store } from "index";
+import { CacheActionTypes } from "store/actions/Cache";
+
 import "./WithNavigationBar.scss";
+
+//load users every 60 seconda
+setInterval(() => {
+  store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+}, 1000 * 60);
+//initial loading of users
+store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+console.log("Dispatching", CacheActionTypes.RELOAD_USER_CACHE);
 
 const NavBar = lazy(() =>
   tracePromise("WithNavigationBar::lazy-import::NavBar", () =>
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -42,6 +42,7 @@ import {
 import { FIREBASE_CONFIG } from "settings";
 
 import { VenueTemplateReducers, MiscReducers } from "store/reducers";
+import { CacheActionTypes } from "store/actions/Cache";
 
 import * as serviceWorker from "./serviceWorker";
 import { activatePolyFills } from "./polyfills";
@@ -149,6 +150,13 @@ if (BUGSNAG_API_KEY) {
     "env/github",
     "env/summit-hack",
   ];
+
+  //load users every 60 seconda
+  setInterval(() => {
+    store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  }, 1000 * 60);
+  //initial loading of users
+  store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
 
   const releaseStage = () => {
     if (
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -42,7 +42,7 @@ import {
 import { FIREBASE_CONFIG } from "settings";
 
 import { VenueTemplateReducers, MiscReducers } from "store/reducers";
-import { CacheActionTypes } from "store/actions/Cache";
+// import { CacheActionTypes } from "store/actions/Cache";
 
 import * as serviceWorker from "./serviceWorker";
 import { activatePolyFills } from "./polyfills";
@@ -103,7 +103,7 @@ const rootReducer = combineReducers({
 export type RootState = ReturnType<typeof rootReducer>;
 
 const initialState = {};
-const store = createStore(
+export const store = createStore(
   rootReducer,
   initialState,
   composeWithDevTools(applyMiddleware(thunkMiddleware))
@@ -151,12 +151,13 @@ if (BUGSNAG_API_KEY) {
     "env/summit-hack",
   ];
 
-  //load users every 60 seconda
-  setInterval(() => {
-    store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
-  }, 1000 * 60);
-  //initial loading of users
-  store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  // //load users every 60 seconda
+  // setInterval(() => {
+  //   store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  // }, 1000 * 60);
+  // //initial loading of users
+  // store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  // console.log("Dispatching", CacheActionTypes.RELOAD_USER_CACHE);
 
   const releaseStage = () => {
     if (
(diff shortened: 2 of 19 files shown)


0xdevalias (approved):
Skimmed the code, seems ok for a quick nuclear fix I think. Nothing jumped out at me at least.

0xdevalias:
> Note that this PR should 'turn off the location updates at the source', but relies on everyone (or majority of) running the latest code (basically 'herd immunity' 😛). So will need to refresh the browser page, etc:
> 
> - https://github.com/sparkletown/sparkle/pull/1846
> 
> ---
> 
> FYI: 'Nuclear triage PR' for switching from 'live streamed updates' to 'poll every minute or so' (while dramatically increasing the running costs) is on the following PR, still seems to be some issues with it which is why it isn't merged yet:
> 
> - https://github.com/sparkletown/sparkle/pull/1845
> - https://github.com/sparkletown/sparkle/pull/1847 (followup fix)
> 
> And for completeness, here is the PR I've been working on to completely rip out/refactor the user location updates to a (hopefully) more efficient/performant system rather than our legacy `react-redux-firebase` integration (which I got pretty close to having finished but ran out of time before the livestream tonight.. 😓):
> 
> - https://github.com/sparkletown/sparkle/pull/1833
> 
> _Originally posted by @0xdevalias in https://github.com/sparkletown/sparkle/issues/1846#issuecomment-880054356_

src/components/organisms/WithNavigationBar/WithNavigationBar.tsx
```diff
@@ -8,8 +8,19 @@ import { RelatedVenuesProvider } from "hooks/useRelatedVenues";
...
+
 import "./WithNavigationBar.scss";
 
+//load users every 60 seconda
+setInterval(() => {
+  store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+}, 1000 * 60);
+//initial loading of users
+store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+console.log("Dispatching", CacheActionTypes.RELOAD_USER_CACHE);
```
0xdevalias, line 22: Polling logic runs here.

src/index.tsx
```diff
@@ -150,6 +151,14 @@ if (BUGSNAG_API_KEY) {
...
   ];
 
+  // //load users every 60 seconda
+  // setInterval(() => {
+  //   store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  // }, 1000 * 60);
+  // //initial loading of users
+  // store.dispatch({ type: CacheActionTypes.RELOAD_USER_CACHE });
+  // console.log("Dispatching", CacheActionTypes.RELOAD_USER_CACHE);
+
```
0xdevalias, line 161: Seemingly was previously intended to be here.

src/store/reducers/Cache.ts
```diff
@@ -0,0 +1,70 @@
...
+          //console.log(u.id);
+
+          if (u.id !== action.payload.id) return u; //don't change
+          return action.payload.user;
+        }),
+      };
+    default:
+      return state;
+  }
+};
```
0xdevalias, line 70: Here's the reducer logic

0xdevalias: This query will load literally all users within the user collection every single time we poll, regardless of whether those users have changed or not:

```typescript
firebase
  .firestore()
  .collection("users")
  .get()
  .then((snapshot) => {
```

Firebase queries allow us to only fetch data that has changed since X time, and/or 'after' a certain snapshot, etc. This would dramatically reduce the amount of data that needs to be loaded each time, and thus will reduce bandwidth usage + also reduce the cost of reads from firebase.

0xdevalias (commented):
Some notes to potentially iteratively improve this polling fix.

src/store/reducers/Cache.ts
0xdevalias: It might be possible to use query cursors or similar https://firebase.google.com/docs/firestore/query-data/query-cursors ?

0xdevalias: @denisdimitrov is going to look into this and assess our options here

denisdimitrov: > It might be possible to use query cursors or similar https://firebase.google.com/docs/firestore/query-data/query-cursors ?

@0xdevalias @mike-lvov Looks like this is mainly used for pagination purposes and won't do the job. It's not flexible enough and doesn't provide the necessary options. Basically, it allows you to add "offset" and "limit" to your query.

I think we should be providing the world users to this action and the reducer should be using only those users and cache their locations. The list of user will be less that fetching all users but it still doesn't change the fact nor fix the problem that unnecessary users will be cache without actual change in their location. I will keep investigating and reading about other possible solutions.

Looks like there are value events that might be useful, currently reading about them: https://firebase.google.com/docs/database/admin/retrieve-data#section-event-types

0xdevalias: Mm, I wasn't sure if that link I provided specifically would work, but yeah, makes sense as to why it wouldn't as you said.

My original thought (not backed up by any specific links) was to just query for 'give me things that have changed since I last queried'. So I assumed there may be a timestamp type of metadata thing we could potentially query on or similar.

0xdevalias: > Looks like there are value events that might be useful, currently reading about them: https://firebase.google.com/docs/database/admin/retrieve-data#section-event-types

@denisdimitrov Which particular part of this? That's for 'realtime database' not firestore, but it also just seems to be equivalent to our standard 'snapshot listener' connection that this PR rips out due to the issues we were having with it.

denisdimitrov: > @denisdimitrov Which particular part of this? That's for 'realtime database' not firestore, but it also just seems to be equivalent to our standard 'snapshot listener' connection that this PR rips out due to the issues we were having with it.

@0xdevalias You are right, this won't  do the job either. 

I have an idea but I might be talking BS. We can fetch world users once (by their user locations) and then use a to populate an array with the changed users, that way the next time the interval triggers it will have to fetch and update only that list of users. Wdyt?

0xdevalias: > and then use a to populate an array with the changed users,

Is this missing something? How will you tell it to fetch only a list of users? And how will you know which ones to query?

On this environment basically every user that exists is going to be a 'related user' anyway, so this wouldn't reduce the number we need to get I don't believe.

denisdimitrov: Yup, there was a typo. We can use a listener, onSnapshot or something like that which can populate the user id in the new array, and then we fetch only the users from that array, by bulk id query.

0xdevalias: With bulk id query we can only do 10 per query I believe, so that would be 100+ queries based on the number of users on the environment.

What is the snapshot/userIds being used for here? Just to detect users which have had updates? The snapshot sends the whole document each time anyway, so this would be creating extra double work, and I don't believe would solve the issue we're having at all.

0xdevalias: In my 'proper fix' PR i'm looking into writing a 'queued throttle' queue/cache to batch snapshot updates together, and then process them every X timeframe; since my observation is that it's the 'constant drip' of updates that causes more of the issue than making a bunch of changes at once.

denisdimitrov: >With bulk id query we can only do 10 per query I believe, so that would be 100+ queries based on the number of users on the environment.

That's true, but it will run only once every minute. It should be much better than live updates.

>What is the snapshot/userIds being used for here? Just to detect users which have had updates?

Yes, that's the idea

> The snapshot sends the whole document each time anyway, so this would be creating extra double work, and I don't believe would solve the issue we're having at all.

Fair enough. I'll search for something else.

P.S. Looking at your WIP fix PR atm: https://github.com/sparkletown/sparkle/pull/1833/files?file-filters%5B%5D=.ts#diff-ae206d6bb7c21f941d94450ed2c3e15ec38e85d6898407bfe91550fadc78b6cbR117-R172
Will let you know if I think of something else.

0xdevalias: You can see the overview I wrote about my 'queued batcher' commit here: https://github.com/sparkletown/sparkle/pull/1833#pullrequestreview-707134935

PR #1846, opened 2021-07-14
[env/github] hack out updateCurrentLocationData to reduce user location update spam

0xdevalias:
`updateCurrentLocationData` runs on an interval, which is fine when only a couple of users, but the cumulative effect of many 100's-1000's of users seems to grind the platform to a halt at times.

Let's hack this location update interval out for now.

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -2,7 +2,7 @@ import React, { Suspense, lazy, useEffect } from "react";
 import { Redirect } from "react-router-dom";
 import { useTitle } from "react-use";
 
-import { LOC_UPDATE_FREQ_MS, PLATFORM_BRAND_NAME } from "settings";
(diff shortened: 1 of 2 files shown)


mike-lvov (approved):

0xdevalias:
Note that this PR should 'turn off the location updates at the source', but relies on everyone (or majority of) running the latest code (basically 'herd immunity' 😛). So will need to refresh the browser page, etc.

---

FYI: 'Nuclear triage PR' for switching from 'live streamed updates' to 'poll every minute or so' (while dramatically increasing the running costs) is on the following PR, still seems to be some issues with it which is why it isn't merged yet:

- https://github.com/sparkletown/sparkle/pull/1845

And for completeness, here is the PR I've been working on to completely rip out/refactor the user location updates to a (hopefully) more efficient/performant system rather than our legacy `react-redux-firebase` integration (which I got pretty close to having finished but ran out of time before the livestream tonight.. 😓):

- https://github.com/sparkletown/sparkle/pull/1833

PR #1850, opened 2021-07-15
[env/github] Set default profile photo on Okta login

goran-peoski-work:
Possible fix for corner cases of https://github.com/sparkletown/internal-sparkle-issues/issues/927

Initially a default avatar image was shown when there was a missing `githubHandle` and there wasn't one already set as `pictureUrl`  . This gave wrong impression the avatar image was already pre-selected/uploaded.

Now `<ProfilePictureInput>` sets the actual value of `githubImageSrc` (and by extension `pictureUrl` when missing ) to be the default `/avatars/default-octocat-3.png`

There is still one edge case remaining, an invalid `githubHandle` that may produce invalid avatar URL e.g. `https://github.com/invalid-github-handle.png?size=120`. But this should be visible to user and not misinform there is a pre-set avatar so they can take the appropriate action.


yurii-lubynets (approved):
Nice catch about a default avatar image! LGTM

PR #1853, opened 2021-07-16
Fix RangeError Invalid interval

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/693


src/utils/event.ts
```diff
@@ -48,8 +49,15 @@ export const hasEventFinished = (event: VenueEvent) =>
...
-  addMinutes(eventStartTime(event), event.duration_minutes);
+export const eventEndTime = (event: VenueEvent) => {
+  const endTime = addMinutes(eventStartTime(event), event.duration_minutes);
+
+  if (isValid(endTime)) {
+    return endTime;
+  }
+
+  return eventStartTime(event);
+};
```
yurii-lubynets, line 60: well, the main problem I've investigated was the "Invalid date" of the `addMinutes` method.

The actual range of times supported by ECMAScript Date objects is slightly smaller https://262.ecma-international.org/5.1/#sec-15.9.1.1
and `eventEndTime` could be "Invalid date" as its result is later than `new Date(8640000000000000)`.

That's why I've added a fallback -> if the `endTime` is invalid then set it as `eventStartTime`
The interval is empty [eventStartTime, eventStartTime]

yurii-lubynets: This one is applied for events created by the script and don't provide any restrictions for user input

PR #1854, opened 2021-07-16
add firestore rule to only allow users read for authenticated users

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/929


firestore.rules
```diff
@@ -18,7 +18,7 @@ service cloud.firestore {
       allow write: if request.auth.uid != null;
     }
     match /users/{userId} {
-      allow read: if true;
+      allow read: if request.auth.uid != null;
```
0xdevalias, line 21: @mike-lvov Have you tested what impacts this has to functionality on the site (if any). I noticed weird things break in ways I wouldn’t have expected when I tried changing some of the other rules recently.

Also, there are other rules here that should be similarly fixed up at the same time, so should really be fixing those as well. Let’s clean all of this up properly in one fell swoop.

0xdevalias (changes requested):

firestore.rules
mike-lvov: @0xdevalias I didn't do the testing, but will do it before merging it in

mike-lvov: Did the testing

jonboutelle (approved):
Lets GO!

0xdevalias (commented):
> Also, there are other rules here that should be similarly fixed up at the same time, so should really be fixing those as well. Let’s clean all of this up properly in one fell swoop.

@mike-lvov @jonboutelle @whoaitsaimz @sparkletown/product @sparkletown/engineering @sparkletown/security Looks like this aspect of my previous review comments wasn’t addressed. Which means that every other rule that is set to ‘true’ will still suffer from this same pattern of vulnerability.

firestore.rules
0xdevalias: See https://github.com/sparkletown/sparkle/pull/1854#pullrequestreview-712273168

PR #1856, opened 2021-07-16
implement Soichi polling hack

mike-lvov:
implements Soichi polling hack as per https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1626440525387800


src/api/users.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { WithId } from "utils/id";
+
+export const fetchUsersRecord = async (): Promise<
+  Record<string, WithId<UserWithLocation>>
+> =>
+  firebase
+    .firestore()
+    .collection("users")
+    .get()
```
0xdevalias, line 13: This will still fetch literally all users every time, which is hugely inefficient. We want to query for only new changes since we last updated.

mike-lvov:
Is unrelevant not

PR #1859, opened 2021-07-19
Change venue creation URL to avoid clash with other route

goran-peoski-work:
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/879

Original URL `/admin-ng/venue/creation` would have been picked up by 
```
<Route path="/admin-ng/venue/:venueId?">
```
misinterpreting the `/creation` segment as a `venueId`.

Now URL is renamed to be picked up by
```
<Route path="/admin-ng/create/venue">
```
for parity with
```
<Route path="/admin-ng/edit/:venueId">
```


denisdimitrov (approved):
LGTM!

PR #1862, opened 2021-07-20
fix room toggle for hidden label

denisdimitrov:
The field used in the types and conditional rendering is called `isLabelHidden` but the toggle (useForm) in the admin was called `isLabeled`. You can see that in the previous PR for the feature: https://github.com/sparkletown/sparkle/pull/1632

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/914

P.S. People should test their stuff before merging. 🔥


yurii-lubynets (approved):
Nice catch!

sofisparkle:
Doesn't seem to have worked on 2.sparkle see issue

PR #1863, opened 2021-07-20
[env/github] improve accessibility concerns in NavBar, NavBarSchedule, and MapRoom

dmitrii-zolotukhin:
- replaced divs with buttons
- removed default style of buttons
- added style for button:focus to match the hover style

fixes for sparkletown/internal-sparkle-issues#887

### DIFF
--- a/src/components/organisms/ChatSidebar/ChatSidebar.scss
+++ b/src/components/organisms/ChatSidebar/ChatSidebar.scss
@@ -60,7 +60,6 @@ $header-height: 45px;
   }
 
   &__controller {
-    outline: none !important;
     border-width: 0;
 
     display: flex;
@@ -80,10 +79,6 @@ $header-height: 45px;
     transform: translateX(-100%);
 
     &:hover {
-      background-color: $lighter-intermediate-grey;
-    }
-
-    &:focus {
       background-color: $lighter-intermediate-grey;
     }
 
--- a/src/components/organisms/ChatSidebar/ChatSidebar.scss
+++ b/src/components/organisms/ChatSidebar/ChatSidebar.scss
@@ -61,7 +61,6 @@ $header-height: 45px;
 
   &__controller {
     border-width: 0;
-
     display: flex;
     justify-content: center;
     align-items: center;
@@ -73,7 +72,6 @@ $header-height: 45px;
     display: flex;
     align-items: center;
     justify-content: center;
-
     position: absolute;
     top: 0;
     transform: translateX(-100%);
--- a/src/components/organisms/ChatSidebar/ChatSidebar.scss
+++ b/src/components/organisms/ChatSidebar/ChatSidebar.scss
@@ -60,6 +60,9 @@ $header-height: 45px;
   }
 
   &__controller {
+    outline: none !important;
+    border-width: 0;
+
     display: flex;
     justify-content: center;
     align-items: center;
@@ -77,6 +80,10 @@ $header-height: 45px;
     transform: translateX(-100%);
 
     &:hover {
+      background-color: $lighter-intermediate-grey;
+    }
+
+    &:focus {
       background-color: $lighter-intermediate-grey;
     }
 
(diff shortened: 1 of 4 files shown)


src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -60,6 +60,9 @@ $header-height: 45px;
   }
 
   &__controller {
+    outline: none !important;
```
yurii-lubynets, line 63: is there any way to avoid `!important`. Maybe it's better to increase CSS specificity via `div.`?

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -51,7 +51,11 @@ export const _ChatSidebar: React.FC<ChatSidebarProps> = ({ venue }) => {
   return (
     <div className={containerStyles}>
       <div className="chat-sidebar__header">
-        <div className="chat-sidebar__controller" onClick={toggleSidebar}>
+        <button
+          className="chat-sidebar__controller"
+          onClick={toggleSidebar}
+          type={"button"}
```
yurii-lubynets, line 57:
```suggestion
          type="button"
```

src/pages/Account/Login.scss
```diff
@@ -29,6 +29,15 @@
   }
 
   &__alternative-logins {
+    outline: none !important;
```
yurii-lubynets, line 32: Could we avoid `!important`?

yurii-lubynets (commented):
Nice work about accessibility! Some comments left

src/components/organisms/ChatSidebar/ChatSidebar.scss
whoaitsaimz: Why are we doing this at all??!!  I've asked to keep outlines in place.  I see that there is a focus background-color applied, but our goal here is accessibility, and this is counter to that goal.

https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

src/pages/Account/Login.scss
whoaitsaimz: Why are we taking away outline at all? As I've mentioned before, **outline** is a built-in accessibility feature, and we should not be removing it, even if it looks better.

This should not be a blocking comment, we need this merged asap, but please consider removing this line entirely @dmitrii-zolotukhin 

https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

src/components/organisms/ChatSidebar/ChatSidebar.scss
sofisparkle: 👆

src/pages/Account/Login.scss
dmitrii-zolotukhin: @yurii-lubynets @whoaitsaimz fixed in 5f6ea55285f3f28e51ae012db3290ac4694d5c03

src/components/organisms/ChatSidebar/ChatSidebar.tsx
dmitrii-zolotukhin: @yurii-lubynets fixed in 5f6ea55285f3f28e51ae012db3290ac4694d5c03

src/components/organisms/ChatSidebar/ChatSidebar.scss
dmitrii-zolotukhin: @yurii-lubynets @whoaitsaimz fixed in 5f6ea55285f3f28e51ae012db3290ac4694d5c03

yurii-lubynets (approved):
LGTM

mike-lvov (changes requested):
Why is it pointed at `env/github`? If there are any accessibilty additions, just point them at staging. I haven't noticed anything github-specific

dmitrii-zolotukhin:
@mike-lvov because Glenn asked to do this change not only against staging but against env/github too, look here https://github.com/sparkletown/internal-sparkle-issues/issues/887#issuecomment-873343095 please, I've already said you in slack

mike-lvov:
@dmitrii-zolotukhin it was said that there are specific things that only live in the `env/github`, but this PR doesn't include them

mike-lvov:
is a duplicate of https://github.com/sparkletown/sparkle/pull/1727

PR #1865, opened 2021-07-20
fixed user avatar load error fallback

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/926

Fixed broken avatars when expanding event user list

### DIFF
--- a/src/components/atoms/UserAvatar/UserAvatar.tsx
+++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useState, useEffect } from "react";
 import classNames from "classnames";
 import { isEqual } from "lodash";
 
@@ -11,11 +11,9 @@ import { WithId } from "utils/id";
 import { useRecentWorldUsers } from "hooks/users";
 import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 import { useVenueId } from "hooks/useVenueId";
+import { useImage } from "hooks/useImage";
 
 import "./UserAvatar.scss";
-import { useImage } from "hooks/useImage";
-import { useState } from "react";
-import { useEffect } from "react";
 
 export type UserAvatarSize = "small" | "medium" | "large" | "full";
 
--- a/src/components/atoms/UserAvatar/UserAvatar.tsx
+++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
@@ -13,6 +13,9 @@ import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 import { useVenueId } from "hooks/useVenueId";
 
 import "./UserAvatar.scss";
+import { useImage } from "hooks/useImage";
+import { useState } from "react";
+import { useEffect } from "react";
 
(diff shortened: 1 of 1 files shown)


src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -97,7 +112,7 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
       {showNametag && <div className={nametagClasses}>{user?.partyName}</div>}
       <img
         className={imageClasses}
-        src={avatarSrc}
+        src={userAvatar}
```
denisdimitrov, line 115: The way I see it, you can just add the `useImage` value here, and get rid of the `useEffect` and `useState`. It has a fallback src of `DEFAULT_PROFILE_IMAGE`
```suggestion
        src={pictureUrl}
```

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -13,6 +13,9 @@ import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 import { useVenueId } from "hooks/useVenueId";
 
 import "./UserAvatar.scss";
+import { useImage } from "hooks/useImage";
+import { useState } from "react";
+import { useEffect } from "react";
```
denisdimitrov, line 18: React imports should be always first and inline. However, after my other comment, these will probably go away.

denisdimitrov (changes requested):

src/components/atoms/UserAvatar/UserAvatar.tsx
alexey-petrov: @denisdimitrov @mike-lvov `useImage` returns fallback img with a delay ( that also depends on the amount of visitors ) after I press "See more" to expand the list of user avatars, so I had to use internal state to have a fallback ready before the load happens and then updating the sate with new img value if we load it successfully
I agree that the stateless version is smoother, but it just doesn't work the way we need it to

src/components/atoms/UserAvatar/UserAvatar.tsx
alexey-petrov: fixed 0d6d78b

src/components/atoms/UserAvatar/UserAvatar.tsx
goran-peoski-work: @alexey-petrov Have you tried alternative approach of encapsulating the logic into a new component, an `<Image src=... fallback=...>` one to replace ``?

denisdimitrov: You should add some documentation about this and why it's written in the way it is. Otherwise someone like me, might change or remove this tomorrow because it looks weird.

ermiaSparkle: fixed in 0d6d78b4fd9051c1d8004354d1f29b70c1ba8171 and 8ac293678fe368d40a8ca067a62b6f5d5c178d3a

denisdimitrov (approved):
LGTM!

PR #1867, opened 2021-07-20
remove column/row admin v1 toggles for non-grid venues

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/935


yurii-lubynets (approved):
LGTM

PR #1869, opened 2021-07-21
Add ButtonNG

goran-peoski-work:
Original `Button` component suffers from legacy debt that also hampers the ability to style new components due to having inline styles by way of: `export const Button = styled.button<ButtonProps>` and is heavily styled by `global.scss`. 

This makes it require `!important` style overrides in cases like 
```scss
  .create-account-button {
    box-shadow: inset 0 0 0 3px $primary !important;
    background-color: $primary !important;
    color: white !important;
```
especially evident with the different style of the Admin NG buttons.

In order not to break existing uses of `<Button>` a new one `<ButtonNG>` is created that has the styles conforming to the established practice of BEM that make it easily (re)styled if necessary as well as props with types like
```ts
export type ButtonType = "button" | "reset" | "submit";
export type ButtonVariant = "primary" | "secondary" | "white" | "dark";
export type ButtonIconSize = "1x" | "2x" | "3x";
```
 that try to simplify the use of established UI elements in https://www.figma.com/file/WvrLkcfGAbyUvRl29AwvdQ/Components?node-id=14%3A553.

Side by side comparison:
sparkle-button-ng-02


denisdimitrov:
@goran-peoski-work How much effort would it take to replace all of the current buttons with this one instead, and just remove the old one.

denisdimitrov:
Maybe not every single `<Button/>` `<button/>` but at least the atom one that's using `styled-components`

goran-peoski-work:
@denisdimitrov not sure how much effort it will take since I'm not aware of all the instances, and more importantly the issues it may create.  Currently many elements depend on the `global.scss` styles, so there might be need for overrides here and there.

My suggestion is to have them side by side, start replacing the old with new and update/upgrade the new one with some edge cases we may encounter.

src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -0,0 +1,121 @@
+@import "scss/constants.scss";
+
+.ButtonNG {
+  display: inline-block;
+
+  padding: 0.75em;
+  margin: $spacing--xs;
+  width: 100%;
+  max-width: 275px;
```
yurii-lubynets, line 9: Could we store `275px` as a variable?

src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -0,0 +1,121 @@
...
+
+  border-radius: $border-radius--xl;
+  border: none;
+  background-image: none;
+
+  font-style: normal;
+  font-size: 1rem;
+  font-weight: 500;
```
yurii-lubynets, line 17:
```suggestion
  font-weight: $font-weight--500;
```

src/components/atoms/ButtonNG/ButtonNG.tsx
```diff
@@ -0,0 +1,124 @@
...
+    "ButtonNG ButtonNG__button": !isLink,
+    "ButtonNG--disabled": disabled,
+    "ButtonNG--enabled": !disabled,
+    "ButtonNG--loading": loading,
+    [`ButtonNG--icon-only ButtonNG--${iconSize}`]: iconOnly,
+    [`ButtonNG--icon-text`]: !iconOnly,
+    [`ButtonNG--${gradient}`]: gradient && !disabled,
+    [`ButtonNG--${variant}`]: variant && !disabled,
+    [className]: className,
+  });
```
yurii-lubynets, line 65:
```suggestion
  const parentClasses = classNames(className, {
    "ButtonNG ButtonNG__link": isLink,
    "ButtonNG ButtonNG__button": !isLink,
    "ButtonNG--disabled": disabled,
    "ButtonNG--enabled": !disabled,
    "ButtonNG--loading": loading,
    [`ButtonNG--icon-only ButtonNG--${iconSize}`]: iconOnly,
    [`ButtonNG--icon-text`]: !iconOnly,
    [`ButtonNG--${gradient}`]: gradient && !disabled,
    [`ButtonNG--${variant}`]: variant && !disabled,
  });
```

src/components/atoms/ButtonNG/ButtonNG.tsx
```diff
@@ -0,0 +1,124 @@
...
+  type?: ButtonType;
+  variant?: ButtonVariant;
+  iconOnly?: boolean;
+  iconName?: IconProp;
+  iconSize?: ButtonIconSize;
+}
+
+export const ButtonNG: React.FC<ButtonProps> = ({
+  children,
+  className = "",
```
yurii-lubynets, line 39: please name it more clearly, like `containerClassName`

yurii-lubynets (commented):
Cool! 
It looks like an efficient way of reusable `Button`. Some comment left

src/components/atoms/ButtonNG/ButtonNG.tsx
goran-peoski-work: Why is `containerClassName` clear?

src/components/atoms/ButtonNG/ButtonNG.scss
goran-peoski-work: What's the rationale for it?

Few questions that may help me understand it better:

- Why if it's used only once in one place?
- Doesn't property name `max-width` describes its intent already?
- If that's the only variable substitution  what makes it different from `width` or `padding`?

src/components/atoms/ButtonNG/ButtonNG.scss
goran-peoski-work: Let's hope some day `$font-weight--500` doesn't become the value of `800`

src/components/atoms/ButtonNG/ButtonNG.scss
yurii-lubynets: Well, it looks useful to store the main measurements of the component view at the head of a file (like in other component styles files).
Paddings are used just for displaying the block and it's fine to define them at a place. 
And `100%` is a general measure. Otherwise, it could be settled in `scss/constants`

src/components/atoms/ButtonNG/ButtonNG.tsx
goran-peoski-work: No explanation provided why `[className]: className,` should be removed.

yurii-lubynets: `classNames` function is already accept `className` as the first argument. It's tastier to add `classNames(className, {...}` rather then `[className]: className`

WDYT?

src/components/atoms/ButtonNG/ButtonNG.scss
goran-peoski-work: K, but since we're starting down this road, I'm starting to reconsider if `max-width` is all we need, there should probably be `min-width` and few others...

goran-peoski-work: I'd put those down as OOS and so that @sparkletown/design may have some starting point to provide more input.

src/components/atoms/ButtonNG/ButtonNG.tsx
yurii-lubynets: discussed and left as a `className`

src/components/atoms/ButtonNG/ButtonNG.tsx
goran-peoski-work: The issue is this: If you didn't point it out to me, I wouldn't have seen it.

That's the reason why I usually just don't mix string + object properties but either go with one or the other.

goran-peoski-work: If no objection, I will leave it as is.

src/components/atoms/ButtonNG/ButtonNG.scss
goran-peoski-work: Added variable @ 187a97d35ba7dea92fa0730f70886a2dc50a5aec

denisdimitrov (changes requested):

src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -0,0 +1,123 @@
+@import "scss/constants.scss";
+
+$max-allowed-button-width: 275px;
+
+.ButtonNG {
+  display: inline-block;
+
+  padding: 0.75em;
```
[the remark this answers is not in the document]
goran-peoski-work: The spacing constants are... constant, to `px` while the height of a button isn't set in `px` and is relative to the font height so the use of `em` makes the padding relative to it.

src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -0,0 +1,123 @@
...
+  margin: $spacing--xs;
+  width: 100%;
+  max-width: $max-allowed-button-width;
+
+  border-radius: $border-radius--xl;
+  border: none;
+  background-image: none;
+
+  font-style: normal;
+  font-size: 1rem;
```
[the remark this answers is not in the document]
goran-peoski-work: The spacing constants are... constant, to `px` while the height of a button isn't set in `px` and is relative to the font height so the use of `rem` makes the padding relative to it. `rem` uses as a sort of soft reset of style before additional is applied

src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -0,0 +1,123 @@
...
+  &:disabled,
+  &:disabled:hover {
+    color: opaque-white(0.4);
+    background: opaque-white(0.1);
+    backdrop-filter: blur(10px);
+    box-shadow: 0 2px 4px opaque-black(0.25);
+    cursor: default;
+  }
+
+  @mixin square-size($size: 1) {
```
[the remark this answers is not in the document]
goran-peoski-work: I agree Dan Abramov considers them harmful. I don't necessarily agree with everything Dan Abramov has to say for all cases that might apply.

A mixin constrained to only `.ButtonNG` defined and scoped just before its use is as explicit and reductant of complexity an "implicit dependency snowballing complexity" can get.

src/components/atoms/ButtonNG/ButtonNG.tsx
```diff
@@ -0,0 +1,124 @@
...
+          size={iconSize}
+          className="ButtonNG__icon"
+        />
+      </button>
+    );
+  }
+
+  if (isLink) {
+    return (
+      <Link
```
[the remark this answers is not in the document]
goran-peoski-work: I opted to keep a minimal viable button that can be a drop-in replacement, so everything is neatly tucked into one tsx (+scss) file until there's a need for extra complexity by the demands of specific uses.

In any case, `ButtonNG` should serve as a façade if new specialized button components stem out of it and used locally only by it and not externally by all other components. This can ensure consistency as an outside component API and also reduce complexity by having the button façade as a switchboard for the few specialized buttons.

You don't need to "know" they're both used together because they're also both used apart.

src/components/atoms/ButtonNG/ButtonNG.scss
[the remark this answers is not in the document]
goran-peoski-work: Did a bit of testing, at first glance it appears the padding plays more role on width then height

At browser zoom: 125%
sparkle-button-padding-02

But then, a side by side comparison of `0.75em` and `12px` doesn't show any difference.
sparkle-button-padding-03

That might be a consequence of generally setting `font-size` at `16px` and/or browsers scaling everything relative to pixels.

There is difference, however, when removing the `1rem` size of the button's font. It falls back to parent `font-size` which at certain places the button is used can be different, but then again, the button styles can easily be added to from parent components

Example without and with 
```css

.wrapper > .ButtonNG {
  display: inline-flex;
  align-content: center;
  justify-content: center;
  align-items: center;
}
```

sparkle-button-ng-05

src/components/atoms/ButtonNG/ButtonNG.scss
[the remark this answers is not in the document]
goran-peoski-work: see https://github.com/sparkletown/sparkle/pull/1869/files/dce1c9428fdb4835f85d88d8772c7ddc130861f8#r682510816

src/components/atoms/ButtonNG/ButtonNG.tsx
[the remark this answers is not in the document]
goran-peoski-work: Example for the flags:
- `isLink` - signals form i.e. "display as a link"
- `linkTo` - signals function i.e. "where do I point to" 

It does look a bit confusing, but some of these are added just for parity with existing `Button` to facilitate easier transition.

I tried a shortcut of the form `const isLink = !! linkTo` but that might clash with a disabled link where `  to={disabled ? "#" : linkTo}`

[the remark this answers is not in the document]
goran-peoski-work: I'd leave them as they are now and figure out a possible refactor after all usages are known.

src/components/atoms/ButtonNG/ButtonNG.scss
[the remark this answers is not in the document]
goran-peoski-work: `1rem` removed as it didn't alter style significantly 1479e4cfaada475c17ac2eb834deb58534bec223

src/components/atoms/ButtonNG/ButtonNG.scss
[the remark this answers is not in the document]
goran-peoski-work: Removed `@mixin` definitions from `ButtonNG.scss` @ 1479e4cfaada475c17ac2eb834deb58534bec223

src/components/atoms/ButtonNG/ButtonNG.scss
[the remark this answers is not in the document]
goran-peoski-work: replaced the `em` values with `px` values from `constants.scss` @ 1479e4cfaada475c17ac2eb834deb58534bec223

denisdimitrov (approved):
LGTM!

PR #1870, opened 2021-07-21
Set UserAvatar image size in pixels, not percents

goran-peoski-work:
Fix issue where `<UserAvatar>` would display oval (not circular) image due to using `%` instead of `px` in SCSS.
Suplemental to https://github.com/sparkletown/sparkle/pull/1263

While https://github.com/sparkletown/sparkle/pull/1263 tries to adjust for the image source (`src` prop) itself not being square size or even missing, this PR deals **only** with the `` element size not being square when using percents.

BEFORE:
sparkle-user-avatar-04

AFTER:
sparkle-user-avatar-03


src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -30,14 +35,12 @@
     bottom: -1px;
     right: -1px;
     border-radius: 50%;
-    height: 8px;
-    width: 8px;
     z-index: z(user-avatar-status-indicator);
     display: block;
+    @include square-size(8px);
```
mike-lvov, line 40: I think the indicator should be dependent of the container size and should be defined in %.

goran-peoski-work: The % is exactly what causes the issue, even when no modifier is provided.

goran-peoski-work: But, let's see if there is some reasonable `px` for all 3 modifiers

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,9 +1,15 @@
 @import "scss/constants";
 
 .UserAvatar {
```
denisdimitrov, line 3: This component and it's styles are used in a lot of places and this change might break something. Please check every single usage and make sure it doesn't break anything visually. Chats, navbar, profile modals, etc...

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: @mike-lvov considering: 

```scss
  &__status-indicator {    

    // .....

    @include square-size(8px);

    &--large {
      @include square-size(12px);
    }
  }
```


The status  indicator for all but `--large` is `8px` . Do you think there is a necessity of defining value for `--small` and `--medium` other than the default?

mike-lvov: @goran-peoski-work I don't think I understand what you mean by 

> The % is exactly what causes the issue, even when no modifier is provided

Setting the % value for the `indicator` should not cause any issue or am I missing something?

goran-peoski-work: Well, the difference of 4 pixels - default at 8px, only set to 12px when size set to large... There wasn't any benefit of playing with % in that interval, but I did rework how the size is set, defaulting to the current a.k.a. legacy if not set.

```scss
  $avatar-sizes-map: (
    // NOTE: legacy not added because it is the default
    small: 25px,
    medium: 40px,
    large: 54px,
    full: 100%
  );
  $indicator-sizes-map: (
    // NOTE: legacy not added because it is the default
    small: 8px,
    medium: 10px,
    large: 12px,
    full: 25%
  );
```

goran-peoski-work: NOTE: the legacy behavior is still set at 8px, not scaling with %, and will have to get some one who knows the possibility of that causing issues to point it out for me to test.

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Checked usages, and generally doesn't break things. In the one place scaling is preferred i.e. the party map, it did show inconsistency, so I've made a `size` prop that will default to this behavior when not explicitly set otherwise. b40bc48846f23c0b8cebb8e9f7f5941bfc5b29bf

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: this is the current state b40bc48846f23c0b8cebb8e9f7f5941bfc5b29bf

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: same as https://github.com/sparkletown/sparkle/pull/1870/files#r674052990

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -14,15 +14,16 @@ import { useVenueId } from "hooks/useVenueId";
...
   user?: WithId<User>;
   containerClassName?: string;
   imageClassName?: string;
   showNametag?: UsernameVisibility;
   showStatus?: boolean;
   onClick?: () => void;
-  large?: boolean;
-  medium?: boolean;
+  size?: UserAvatarSize;
 }
```
goran-peoski-work, line 27: Setting a `size` prop with limited set of values helps not bloat the `UserAvatarProps` with many mutually exclusive booleans, keeps the possibility of being expanded, but also limits the variations to a handful for consistency and maintainability.

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -33,8 +34,7 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
   showNametag,
   onClick,
   showStatus,
-  large,
-  medium,
+  size = "legacy",
```
goran-peoski-work, line 37: To not break existing uses of `<UserAvatar>` where no size is explicitly set, the value of `"legacy"` is used as a no-op keeping the original defaults set for that use.

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -56,8 +56,7 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
 
   const containerClasses = classNames("UserAvatar", containerClassName, {
     "UserAvatar--clickable": onClick !== undefined,
-    "UserAvatar--large": large,
-    "UserAvatar--medium": medium,
+    [`UserAvatar--${size}`]: size,
```
goran-peoski-work, line 59: Size is limited to a handful values, it shouldn't be a problem either being undefined or `"legacy"`, `"small"` etc.

src/components/atoms/UserAvatar/UserAvatar.tsx
mike-lvov: @goran-peoski-work I don't think we should add `legacy`. Let's just leave the default as it used to be, but add it to our map, which should equal to medium

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: @goran-peoski-work just setting the size of indicator to 25% should simplify the thing, I don't understand what's the stopping point here. It will automatically be adjusted to any size changes & if a person will set the size via the classes for ex.

src/components/atoms/UserAvatar/UserAvatar.tsx
mike-lvov: https://github.com/sparkletown/sparkle/pull/1870#discussion_r674119797

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: Btw, the current `legacy` doesn't look right
image

goran-peoski-work: > I don't understand what's the stopping point here.

@mike-lvov OOS?

An issue about the UserAvatar not being square is morphing to issue about the indicator. And I'd be OK fixing it if it was all about simply about slapping a % and saying it's done. 

It is not a static proportion, it is granular.
At `25px` and at `48px` the indicator is both `8px`, that's `32%` and `16%` respectively, while for the large size - `22%`.

So instead of us here trying to figure out what looks right in which cases, I suggest we ask design to give us proper guide on that, we create separate PR for the indicator and until then, leave the legacy indicator as it already was and just fix the avatar size in this PR while providing a manageable interface for future addition/expansion.

There are after all 2 other PRs relating to different aspects of the avatar, so one more shouldn't be that of a problem.

src/components/atoms/UserAvatar/UserAvatar.tsx
goran-peoski-work: 'legacy' is the word for default. the default is already left as it used to be, I just didn't want to use that label in case there's new default

src/components/atoms/UserAvatar/UserAvatar.tsx
goran-peoski-work: > is used as a no-op keeping the original defaults set 

Will resolve it here and leave the linked comment the place to continue

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: @goran-peoski-work okay, I agree that it's a design issue

src/components/atoms/UserAvatar/UserAvatar.tsx
mike-lvov: @goran-peoski-work the `default` should be not passed value, so there should be no string `default`/`legacy`. Not passing a size option means it's the default. What do you mean by `new default`?

denisdimitrov: `legacy` does sound strange for a prop called `size`. I might be wrong here, but looks like there was no default value before, why do we need it set to `legacy` now?

goran-peoski-work: removed legacy value @ 76220e8c2aab97e736091f15773d4ae204e4ddd7

mike-lvov (approved):
LGTM

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -49,19 +65,11 @@
...
-  }
-
-  &--large {
-    height: 54px;
-    width: 54px;
+  @each $modifier, $size in $avatar-sizes-map {
+    &--#{$modifier},
+    &--#{$modifier} img {
+      @include square-size($size);
+    }
```
denisdimitrov, line 72: The syntax is shorter but the readability is not better. We should choose one and keep it consistent, I haven't seen this syntax in the project so far. And we need to communicate these things before applying them. If I start adding my own standards and styles with every PR you guys will hate me by the end of the week 😆

goran-peoski-work: >  I haven't seen this syntax in the project so far. 

https://github.com/sparkletown/sparkle/blob/staging/src/scss/constants.scss contains mixins, functions, interpollation...

Standard SCSS syntax really https://sass-lang.com/documentation 
and my view is if we don't use it as intended to make few things DRY and robust, might as well move to back to CSS.


> And we need to communicate these things before applying them.

PR review process is communication, PR merge is application, so we're good I think.

> you guys will hate me by the end of the week

Not really. Not a fan of setting things in stone.

sabrinaorban:
My safe formula for the avatar sizes would be:
- for avatars up to 55px, the indicator should be  28% of the avatar size, and the indicator border should be 12% of the indicator size
- for avatars bigger than 55px, if it’s ever the case we are safe with a 16% indicator size
- I don’t know all the cases where we use all these avatar sizes to check how it looks on the platform but I tested it on a design file and it seems to work alright
- the indicator border is extra, it makes the indicator stand out in an elegant way. it would be nice if the border could have the color of the background behind the avatar
Screenshot 2021-07-22 at 18 16 39

PR #1871, opened 2021-07-21
Add subroutes for different tabs of AdminVenueView

goran-peoski-work:
In order to have functionality of the kind:
> As an admin, I can add experiences to each of my spaces (takes you to the timings tab) P0

(from issue https://github.com/sparkletown/internal-sparkle-issues/issues/683)
there is a need of more precisely targeting the `AdminVenueView`'s tabs.

~Added library 5.5kB (https://bundlephobia.com/package/react-router-bootstrap@0.25.0) library `react-router-bootstrap` that allows React Router to properly route from Bootstrap elements such as the tabs in Admin NG:~
👆
No longer the case. Using plain old React Router with its own `<Link>`s to produce the same result.

sparkle-admin-ng-subroutes-02
sparkle-admin-ng-subroutes-01


src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -34,7 +34,7 @@ export const AdminVenueCard: React.FC<AdminVenueCardProps> = ({ venue }) => (
     <Link className="AdminVenueCard__button" to={adminNGVenueUrl(venue.id)}>
       Manage Space
     </Link>
-    <Link className="AdminVenueCard__button" to={adminNGSettigsUrl(venue.id)}>
+    <Link className="AdminVenueCard__button" to={adminNGSettingsUrl(venue.id)}>
```
goran-peoski-work, line 37: Typo fix along the way.

src/components/molecules/AdminVenueCard/AdminVenueCard.tsx
```diff
@@ -6,7 +6,7 @@ import { AnyVenue } from "types/venues";
 import { DEFAULT_VENUE_LOGO } from "settings";
 
 import { WithId } from "utils/id";
-import { adminNGSettigsUrl, adminNGVenueUrl } from "utils/url";
+import { adminNGSettingsUrl, adminNGVenueUrl } from "utils/url";
```
goran-peoski-work, line 9: Typo fix along the way.

src/utils/url.ts
```diff
@@ -11,12 +11,13 @@ export const venueInsideUrl = (venueId: string) => {
...
-export const adminNGVenueUrl = (venueId?: string) =>
-  generatePath("/admin-ng/venue/:venueId?", {
+export const adminNGVenueUrl = (venueId?: string, selectedTab?: string) =>
+  generatePath(`/admin-ng/venue/:venueId?/:selectedTab?`, {
     venueId: venueId,
+    selectedTab: selectedTab,
   });
 
-export const adminNGSettigsUrl = (venueId?: string) =>
+export const adminNGSettingsUrl = (venueId?: string) =>
```
goran-peoski-work, line 20: Typo fix along the way.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -36,22 +44,21 @@ export const AdminVenueView: React.FC = () => {
...
+          key={key}
+          className={classNames("AdminVenueView__tab", {
+            "AdminVenueView__tab--selected": selectedTab === key,
+          })}
+          eventKey={key}
+          href={adminNGVenueUrl(venueId, key)}
+        >
+          {label}
+        </Nav.Link>
+      </LinkContainer>
```
goran-peoski-work, line 59: NB:  `LinkContainer` irons out the inconsistencies between Bootstrap's `Nav.Link` and React Router's `Link` and provides the same interface/props as the React Router one.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -17,17 +20,22 @@ export enum AdminVenueTab {
...
 const adminVenueTabLabelMap: Readonly<Record<AdminVenueTab, String>> = {
   [AdminVenueTab.spaces]: "Spaces",
   [AdminVenueTab.timing]: "Timing",
   [AdminVenueTab.run]: "Run",
 };
 
-const DEFAULT_TAB = AdminVenueTab.spaces;
-
 export const AdminVenueView: React.FC = () => {
-  const venueId = useVenueId();
```
goran-peoski-work, line 29: NB: `useVenueId()` functionality is already covered by `useParams()` and the only difference of handling legacy `?venueId=` query parameters shouldn't be an issue for the new Admin NG

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -36,22 +44,21 @@ export const AdminVenueView: React.FC = () => {
...
-        className={classNames("AdminVenueView__tab", {
-          "AdminVenueView__tab--selected": selectedTab === key,
-        })}
-        eventKey={key}
-      >
-        {text}
-      </Nav.Link>
+    return Object.entries(adminVenueTabLabelMap).map(([key, label]) => (
+      <LinkContainer key={key} to={adminNGVenueUrl(venueId, key)}>
+        <Nav.Link
```
denisdimitrov, line 49: You can replace this with react-router's `Link` component. That way you won't have to add additional dependency `react-router-bootstrap`. We work very hard to get rid of as many deps as we can to lower the bundle size.

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -50,7 +50,7 @@ export const AdminSubrouter: React.FC = () => {
 
       {/* Admin V2/3/NG */}
 
-      <Route path="/admin-ng/venue/:venueId?">
+      <Route path="/admin-ng/venue/:venueId?/:selectedTab?">
```
denisdimitrov, line 53: You will have to do the same thing for advanced settings screen as well. To keep it consistent and allow subroute navigation, there's 90% change they are going to need it there as well.

denisdimitrov (changes requested):

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: @denisdimitrov That was the first thing I tried, hence the library: https://github.com/sparkletown/sparkle/pull/1871#discussion_r673808672

src/components/organisms/AdminVenueView/AdminVenueView.tsx
denisdimitrov: If you replace `Nav.Link` with `Link` There would be no boostrap in this chunk of code, hence to need for the library, am I missing something else? What are these inconsistencies ?

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: @denisdimitrov The question is:

Do we add 5kB more for a library only used in the Admin or do we stop using certain Bootstrap components that don't play well with React Router (does full page load in this case)?


I can do the necessary adjustments in both cases.

denisdimitrov: > Do we add 5kB more for a library

If that's the case, just keep it as it is.

goran-peoski-work: Removed dependency and `Nav.Link`s at ba2dd1d5395275e3a9cc516552882a888407661e as per https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1626860925271700?thread_ts=1626860598.271300&cid=C01H9D1P1PE

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Done for advanced settings as well in  ba2dd1d5395275e3a9cc516552882a888407661e

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: My bad, misread that `Nav.Link`.

goran-peoski-work: Removed dependency and `Nav.Link`s at ba2dd1d5395275e3a9cc516552882a888407661e as per https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1626860925271700?thread_ts=1626860598.271300&cid=C01H9D1P1PE

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -74,7 +74,7 @@ export const AdminSubrouter: React.FC = () => {
         </Provided>
       </Route>
 
-      <Route path="/admin-ng">
+      <Route path={["/admin-ng", "/admin-ng/*"]}>
```
goran-peoski-work, line 77: NB: a catch all route for anything under `/admin-ng` that didn't get handled by previous routes.

src/pages/Admin/Admin_v2.tsx
```diff
@@ -34,6 +35,11 @@ const Admin_v2: React.FC = () => {
 
   const { isAdminUser } = useIsAdminUser(user?.uid);
 
+  useEffect(() => {
+    // when URL is a catch all, replace it with the canonical form
+    window.history.replaceState(null, "", adminNGRootUrl());
+  });
+
```
goran-peoski-work, line 42: NB: This code cleans up the URL in case of a catch all route hit i.e. will replace URLs like `/admin-ng/asfdsafsdafsdf` with `/admin-ng`

src/utils/url.ts
```diff
@@ -11,15 +11,21 @@ export const venueInsideUrl = (venueId: string) => {
...
+  venueId?: string,
+  selectedTab?: string
+) =>
+  segment
+    ? generatePath(`/admin-ng/:segment?/:venueId?/:selectedTab?`, {
+        segment: segment,
+        venueId: venueId,
+        selectedTab: selectedTab,
+      })
+    : generatePath("/admin-ng");
```
goran-peoski-work, line 24: NB: keeps the code DRY

src/utils/url.ts
```diff
@@ -11,15 +11,21 @@ export const venueInsideUrl = (venueId: string) => {
...
+
+export const adminNGRootUrl = createAdminUrlHelperFor("");
```
denisdimitrov, line 26: The parameter can be optional and there would be no need for this empty string.

src/pages/Admin/Admin_v2.tsx
denisdimitrov: `react-router` should have something like this to avoid manual logic in every single screen. I've used it before but I can't really recall the name of the component.

package-lock.json
```diff
@@ -23472,6 +23482,14 @@
         }
       }
     },
+    "react-router-bootstrap": {
```
denisdimitrov, line 23485: These need to be removed.

src/utils/url.ts
```diff
@@ -11,15 +11,21 @@ export const venueInsideUrl = (venueId: string) => {
...
-  });
-
-export const adminNGSettigsUrl = (venueId?: string) =>
-  generatePath("/admin-ng/advanced-settings/:venueId?", {
-    venueId: venueId,
-  });
+const createAdminUrlHelperFor = (segment: string) => (
+  venueId?: string,
+  selectedTab?: string
+) =>
```
denisdimitrov, line 17: The first param `segment` is called in the constants written here, but the second pair of params are called from the view. This can be confusing and hard to track, you can achieve the same result by using a single pair of params to improve the readability.

Instead of having shadow params like this: 
```tsx
export const adminNGVenueUrl = createAdminUrlHelperFor("venue");
```
You can have it like this: 
```tsx
export const adminNGVenueUrl = (venueId, selectedTab) => createAdminUrlHelperFor(venueId, selectedTab, "venue");
```

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
```diff
@@ -77,25 +80,24 @@ export const AdminAdvancedSettings: React.FC = () => {
...
+        </div>
       </div>
       {selectedTab === AdminAdvancedTab.basicInfo && <VenueWizard />}
       {selectedTab === AdminAdvancedTab.entranceExperience && (
         <EntranceExperience
           // @debt Venue_v2 has different structure than AnyVenue, 1 of them should be deprecated.
           venue={venue as Venue_v2}
           sovereignVenue={sovereignVenue}
-          onSave={selectDefaultTab}
+          onSave={() => history.push(adminNGSettingsUrl(venueId))}
```
denisdimitrov, line 93: This should be a function with useCallback

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -36,22 +44,25 @@ export const AdminVenueView: React.FC = () => {
...
-  }, [selectedTab]);
+  }, [selectedTab, venueId]);
 
-  const selectTab = useCallback((tab: string) => {
-    setSelectedTab(tab as AdminVenueTab);
-  }, []);
+  useEffect(() => {
+    if (venueId) return;
+    // when URL is invalid, land on the root one for the admin
+    history.push(adminNGRootUrl());
```
denisdimitrov, line 64: How is this going to work, if my internet is slow and the venueId is undefined until, the request is complete. This should be somehow handled from the router dep

src/pages/Admin/Admin_v2.tsx
goran-peoski-work: I think it's basically putting a route with no path at the end under the `<Switch>` .
But from what I remember for actual replacing the history:

`const history  = useHistory();` 

was noted in some StackOverflow posts as "not always works, use native `window.history`"

src/utils/url.ts
goran-peoski-work: Could, just looked a bit like:

`createAdminUrlHelperFor()` - create admin URL helper for ...
`createAdminUrlHelperFor("")`  - create admin URL helper for empty path

🤷 either way is fine by me, so unless specifically asking for the change, I'll leave it as it is

src/utils/url.ts
goran-peoski-work: It's not pairs of params on a single function.
It's a higher order function, a function that creates a helper function.

Intention can more easily be determined by looking at

```ts
const createAdminUrlHelperFor = (segment: string) => {
  if (segment) {
    return (venueId?: string, selectedTab?: string) =>
      generatePath(`/admin-ng/:segment?/:venueId?/:selectedTab?`, {
        segment: segment,
        venueId: venueId,
        selectedTab: selectedTab,
      });
  } else {
    return () => generatePath("/admin-ng");
  }
};
```

So, in that regard, the following one seems to have "shadow" params, while the first one is simply a plain old standard JS closure.
```ts
export const adminNGVenueUrl = (venueId, selectedTab) => createAdminUrlHelperFor(venueId, selectedTab, "venue");
```

package-lock.json
goran-peoski-work: done 2199a2bde48e2d20f259bd6bda36305267215426

src/pages/AdminAdvancedSettings/AdminAdvancedSettings.tsx
goran-peoski-work: done f01366cd31785f24e9ad3565b0c8068e881e9a4e

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: The pains of hooks made me move it before the 
```ts
  if (!isCurrentVenueLoaded) {
    return <LoadingPage />;
  }
```

which should still execute before any `useEffect()` does, as effects scheduled after rendering.

But still, if `venueId` is undefined, that's since the start of the request.
That value is retrieved by `useParams()` and if `undefined`,  no loading will ever start in the first place.

goran-peoski-work: This kind of logic could potentially be pulled up to the router level, but that's just putting internal children logic into the parent. Could still work with some generic router handling though, one that doesn't account for children specifics.

denisdimitrov: I think you should talk to product about these things before implementing them, have you or did you add this as your own personal decision/preference? I don't think it's the wrong thing to do, but it might be confusing for someone to copy/paste a link and then get redirected to `/admin-ng/`. We also had this venue doesn't exist page but i'm not sure if that was for the template wrapper only.

src/utils/url.ts
denisdimitrov: >It's not pairs of params on a single function.
It's a higher order function, a function that creates a helper function.

I know it's a HOF I was just using the `pair` terminology, to make you understand which ones I was talking about. And I know why it has `shadow` params, that's my argument. Why add this HOF complexity when you can make it a normal function?

goran-peoski-work: @denisdimitrov "Why add this HOF complexity when you can make it a normal function?" - please define normal function.

goran-peoski-work: NOTE: comments explaining what something is are provided for future reference to anyone that might need the context and are in no way reflection of the current participants' knowledge.

goran-peoski-work: Simplified in 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168. There was no practical benefit of conflating the one root URL as an exception to the other two.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: Removed the automatic re-direct as OOS @ 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168 and will create separate PR once @sparkletown/product is being consulted.

src/utils/url.ts
goran-peoski-work: Replaced with a nullary function for brevity and type signature @ 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168

src/utils/url.ts
goran-peoski-work: Refactored 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168

src/pages/Admin/Admin_v2.tsx
goran-peoski-work: Removed in 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168

src/components/organisms/AppRouter/AdminSubrouter.tsx
goran-peoski-work: Removed in 2c33b6b8de126d7a4550fdd2ca5b20e1ad2b5168

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -43,26 +49,35 @@ export const AdminVenueView: React.FC = () => {
...
 
-  const selectTab = useCallback((tab: string) => {
-    setSelectedTab(tab as AdminVenueTab);
-  }, []);
+  const gotoHome = useCallback(() => history.push(adminNGRootUrl()), [history]);
 
-  const selectTiming = useCallback(
-    () => setSelectedTab(AdminVenueTab.timing),
-    []
+  const gotoSpaces = useCallback(
```
denisdimitrov, line 68: `goToSpaces` same for the rest of the functions. You can turn it up a notch and change it to `navigateToSpaces` that way the name indicates history change.

src/components/organisms/AdminVenueView/components/Timing/Timing.tsx
```diff
@@ -32,30 +34,36 @@ export const Timing: React.FC<TimingProps> = ({
...
-              // href="/admin-ng"
-              className="Timing__left-bottombar-btngrey Timing__btn"
-            >
-              <FontAwesomeIcon
-                icon={faHome}
-                className="edit-button__icon"
-                size="lg"
-              />
-            </button>
+            {onClickHome && (
```
denisdimitrov, line 37: Rendering content depending on callback prop seems weird, can we just change `onClickHome` to be required prop instead? WDYT?

src/components/organisms/AdminVenueView/components/Timing/Timing.tsx
```diff
@@ -32,30 +34,36 @@ export const Timing: React.FC<TimingProps> = ({
...
+              <button
+                className="Timing__left-bottombar-btngrey Timing__btn"
+                onClick={onClickBack}
+              >
+                Back
+              </button>
+            )}
+            {onClickNext && (
+              <button
+                className="Timing__left-bottombar-btnprimary Timing__btn"
```
denisdimitrov, line 61: btnprimary and btngrey sound like modifiers and not elements, can you look into this BEM naming and improve it a bit?

denisdimitrov (changes requested):

src/components/organisms/AdminVenueView/components/Timing/Timing.tsx
goran-peoski-work: The reason is to have same props for all 3 tabs i.e. the components being displayed, that way any one of them may or may not have a home/back/next buttons, but they will not have a reference if they're first or last in the flow. The concern of the flow will rest in `Spaces` component and how it arranges them.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: renamed @ a1bdfd1e5370a2260e69737437e6fe83d99dbf16. 

also `navigateToRun` 😄

src/components/organisms/AdminVenueView/components/Timing/Timing.tsx
goran-peoski-work: Open enhancement issue assigned to me. 

I didn't want to go far OOS by replacing anything with `ButtonNG`, and after https://github.com/sparkletown/sparkle/pull/1950 lands there will be prior use in the `Spaces` tab to just copy from there and improve

denisdimitrov (approved):
LGTM!

P.S. We should probably do the same thing for advanced settings in a future PR because navigation with tabs will most likely be required for that screen as well.

PR #1881, opened 2021-07-22
update Accessibility of Profile picture picker

ermiaSparkle:
related https://github.com/sparkletown/internal-sparkle-issues/issues/887

Update the profile picture page

### DIFF
--- a/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
+++ b/src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
@@ -95,8 +95,14 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
     setValue("pictureUrl", pictureUrlRef, true);
   };
 
+  const uploadProfilePic = useCallback((event) => {
+    event.preventDefault();
+    uploadRef.current?.click();
+  }, []);
+
   const uploadDefaultAvatar = useCallback(
-    async (avatar: string) => {
+    async (event, avatar: string) => {
+      event.preventDefault();
       setValue("pictureUrl", avatar, true);
     },
     [setValue]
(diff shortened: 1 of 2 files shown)


src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -112,17 +118,17 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
+        onClick={(event) => uploadDefaultAvatar(event, avatar)}
       >
         <img
           src={avatar}
           className="profile-icon profile-picture-preview"
           alt={`default avatar ${index}`}
         />
-      </div>
+      </button>
     ));
```
mike-lvov, line 132: Why do you need to prevent the default behaviour?

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -95,8 +95,14 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
     setValue("pictureUrl", pictureUrlRef, true);
   };
 
+  const uploadProfilePic = useCallback((event) => {
+    event.preventDefault();
+    uploadRef.current?.click();
+  }, []);
```
mike-lvov, line 101: Why do you want to stop propagating the click?

ermiaSparkle: Because it will submit if we use the keyboard. Am I right? The button trigger the Form submit system.

mike-lvov: Ah, okay, yeah, makes sense

mike-lvov (approved):
What about this? Do we need it?
https://github.com/sparkletown/sparkle/pull/1863

mike-lvov:
@ermiaSparkle his changes should be applied to staging. If you could review those, it would be great.

PR #1886, opened 2021-07-22
Fix import bug in ImageCollectionInput

malaikaaiyar:
A different PR of mine had a bug- forgot to import a setting. This fixes it


mike-lvov (approved):
LGTM

PR #1891, opened 2021-07-23
Make UserAvatar indicator size a percent of parent's

goran-peoski-work:
Relates to https://github.com/sparkletown/sparkle/pull/1870#discussion_r674575304

Spec  from https://github.com/sparkletown/sparkle/pull/1870#issuecomment-885047195
>- for avatars up to 55px, the indicator should be  28% of the avatar size, and the indicator border should be 12% of the indicator size
>- for avatars bigger than 55px, if it’s ever the case we are safe with a 16% indicator size
>- the indicator border is extra, it makes the indicator stand out in an elegant way. it would be nice if the border could have the color of the background behind the avatar

From Slack (https://doesitsparkle.slack.com/archives/C01700P4QCT/p1626949623174200):
> min border width: 0.8px;
> max border width: 2px;


src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,28 @@
...
+  // order of elements:
+  // $modifier - one of UserAvatarSize
+  // $avatar-size - set in pixels or percent for larger than 54px
+  // $indicator-size - set in percent of parent
+  // $indicator-border-width - calculated as a percent of indicator size (fraction of a fraction of avatar size)
+  $avatar-sizes-enum: (
+    (small, 25px, 28%, 0.12 * 0.28 * 25px),
+    (medium, 40px, 28%, 0.12 * 0.28 * 40px),
+    (large, 54px, 28%, 0.12 * 0.28 * 54px),
+    (full, max(55px, 100%), 16%, 0.12 * 0.16 * 55px)
```
alexey-petrov, line 24: where do we get all these pixel values? probably would be worthwhile to move them to constants

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
```diff
@@ -15,7 +15,7 @@ $profile-popover--margin-top: 12px;
   position: relative;
   overflow: auto;
   padding: $spacing--md;
-  background: $black;
+  @include background-color($black);
```
alexey-petrov, line 18: is there any advantage using @include instead of pure background there?

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Those are the constants

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
goran-peoski-work: yes

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,34 @@
...
+  $default-indicator-size: 28%;
   $default-avatar-size: 25px;
   $default-image-size: 100%;
-  $default-indicator-size: 8px;
+
+  // indicator border is 12% of indicator size which itself is some % of avatar size
+  // limited to min: 0.8px and max: 2px
+  @function indicator-border-width($indicator-size, $avatar-size) {
+    @return clamp(0.8px, 0.12 * #{$indicator-size} * #{$avatar-size}, 2px);
+  }
```
goran-peoski-work, line 19: Using function makes it easier to understand why some numbers get multiplied as well as a bit more robustness.

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,34 @@
...
+  // $modifier - one of UserAvatarSize
+  // $avatar-size - set in pixels or percent for larger than 54px
+  // $indicator-size - set in percent of parent
+  // $indicator-border-width - calculated as a percent of indicator size
+  $avatar-sizes-enum: (
+    (small, 25px, 28%, indicator-border-width(0.28, 25px)),
+    (medium, 40px, 28%, indicator-border-width(0.28, 40px)),
+    (large, 54px, 28%, indicator-border-width(0.28, 54px)),
+    (full, max(55px, 100%), 16%, indicator-border-width(0.16, 108px))
+  );
```
goran-peoski-work, line 31: This is basically an enumeration of related constants as per design spec.

The names `small`, `medium` and `large` are in line with pre-existing classes for `UserAvatar`, but if there's a need of the larger (than 54px) sizes to be codified as well, I can suggest moving to a Bootstrap-like nomenclature:

- Extra small (xs)
- Small (sm)
- Medium (md)
- Large (lg)
- Extra large (xl)
- Extra extra large (xxl)

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,34 @@
...
+  // order of elements:
+  // $modifier - one of UserAvatarSize
+  // $avatar-size - set in pixels or percent for larger than 54px
+  // $indicator-size - set in percent of parent
+  // $indicator-border-width - calculated as a percent of indicator size
+  $avatar-sizes-enum: (
+    (small, 25px, 28%, indicator-border-width(0.28, 25px)),
+    (medium, 40px, 28%, indicator-border-width(0.28, 40px)),
+    (large, 54px, 28%, indicator-border-width(0.28, 54px)),
+    (full, max(55px, 100%), 16%, indicator-border-width(0.16, 108px))
```
goran-peoski-work, line 30: Border width in CSS doesn't support setting it in % so a fixed value of 2 * 54px is used to calculate it.

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -48,13 +59,13 @@
...
 
-    @each $modifier, $size in $indicator-sizes-map {
-      &--#{$modifier} {
-        @include square-size($size);
-      }
-    }
+    // NOTE: for the illusion of transparency, the border's style and color should be set by @include background-color() mixin in the appropriate ancestor
+    border-style: var(--UserAvatar__status-indicator--border-style);
+    border-color: var(--UserAvatar__status-indicator--border-color);
+    border-width: $default-indicator-border-width;
```
goran-peoski-work, line 68: There is a limitation of the ability to have transparent border that will also preclude the painting of its parent, so the border is set to a color that is set on the ancestor. Example:

Parent`div` contains component `UserAvatar` `div` which has status indicator `div`.
In order for the indicator to have the same color border as the parent, it should be passed down from the parent as a CSS --variable.

The use of the BEM style in the  variable names is deliberate in order to exclude clashes with possible other variables as well as to directly indicate by which component (or element in it) should be used.

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -65,10 +76,28 @@
...
+      border-width: $indicator-border-width;
+
+      // NOTE: variable is just template for DRY, the actual calc() is performed in the browser
+      // calculated out of both circles sizes and offset for the bottom/right side of the indicator
+      $indicator-position-offset: calc(
+        #{$avatar-size} *
+          #{$indicator-offset-factor} -
+          #{$indicator-size} *
+          #{$indicator-radius-factor}
+      );
```
goran-peoski-work, line 98: SCSS is a templating engine, so in this case it will not pre-calculate the value but pass it to the browser to use the `calc()` function on the substituted values.

This is used because browsers have more flexibility when it comes to mixing calculations of pixels and percents that aren't available to the templating engine  during compile time.

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -73,7 +73,6 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
   const statusIndicatorClasses = classNames("UserAvatar__status-indicator", {
     "UserAvatar__status-indicator--online": isOnline,
     [`UserAvatar__status-indicator--${status}`]: isOnline && status,
-    [`UserAvatar__status-indicator--${size}`]: size,
```
goran-peoski-work, line 76: NOTE: not necessary anymore. It's more robust to put the size modifier at the top component level and have all descendants react accordingly.

src/components/molecules/NavBar/NavBar.scss
```diff
@@ -3,7 +3,7 @@
 $border-radius: 28px;
 
 .navbar-container {
-  background: $black;
+  @include background-color($black);
```
goran-peoski-work, line 6: NOTE: function sets the needed CSS --variables for the descendant indicator border to preclude the direct parent's paint.

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -12,13 +12,13 @@ $header-height: 45px;
   right: 0;
   width: $chat-sidebar-width;
   height: calc(100vh - 96px);
-  background-color: $saturated-black;
+  @include background-color($saturated-black);
```
goran-peoski-work, line 15: NOTE: function sets the needed CSS --variables for the descendant indicator border to preclude the direct parent's paint.

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -12,13 +12,13 @@ $header-height: 45px;
...
   z-index: z(sidebar);
 
   &--expanded {
     transform: translateX(0);
     transition: transform 0.3s;
-    background-color: rgba($saturated-black, 0.9);
+    @include background-color(rgba($saturated-black, 0.9));
```
goran-peoski-work, line 21: NOTE: function sets the needed CSS --variables for the descendant indicator border to preclude the direct parent's paint.

src/components/organisms/ProfileModal/ProfilePopoverContent.scss
goran-peoski-work: NOTE: function sets the needed CSS --variables for the descendant indicator border to preclude the direct parent's paint.

src/scss/constants.scss
```diff
@@ -248,3 +248,9 @@ $z-layers: (
   max-height: $max;
   max-width: $max;
 }
+
+@mixin background-color($color) {
+  background-color: $color;
+  --UserAvatar__status-indicator--border-style: solid;
+  --UserAvatar__status-indicator--border-color: #{$color};
+}
```
goran-peoski-work, line 256: This allows ancestors of `UserAvatar` to not only set their background color, but also provide color for the indicator border inside the `UserAvatar`

goran-peoski-work (commented):
Notes on the PR code

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,34 @@
...
-
-  // NOTE: parent and img mismatch due to legacy reasons
+  // @see https://math.stackexchange.com/questions/377234/circle-in-square-calculate-distance-from-squares-corner-to-circles-perimeter
+  $square-root-of-2: 1.41421356237;
+  $indicator-offset-factor: ($square-root-of-2 - 1) / 2;
+  $indicator-radius-factor: $square-root-of-2 / 2;
+
+  // indicator border should be 12% of the indicator size, but borders don't accept % so using px for default
+  $default-indicator-border-width: 2px;
+  $default-indicator-size: 28%;
```
mike-lvov, line 11: Is there a reason for these consts to be inside the class instead of being at the top of the file?

src/components/atoms/UserAvatar/UserAvatar.scss
```diff
@@ -1,23 +1,34 @@
...
   $default-avatar-size: 25px;
   $default-image-size: 100%;
-  $default-indicator-size: 8px;
+
+  // indicator border is 12% of indicator size which itself is some % of avatar size
+  // limited to min: 0.8px and max: 2px
+  @function indicator-border-width($indicator-size, $avatar-size) {
+    @return clamp(0.8px, 0.12 * #{$indicator-size} * #{$avatar-size}, 2px);
```
mike-lvov, line 18: should we move the 12% into a const so that it's not a magical number?

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: Good suggestion. This is aligned with our consts too. I don't think we need it now, but good to know for the future

src/scss/constants.scss
mike-lvov: We shouldn't have the background and user avatar tied up. Let's have them separate. Any reason to have the border-color same as the background?

mike-lvov (changes requested):

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Scoping. 

They are relevant only to that component so the `.UserAvatar {}` block envelopes everything that should be only part of its scope. 

Note: they're still technically "at the top" of the file, since there are only 4 lines ahead.

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Calculation is explained in the comment. The function itself serves to not bleed those literals out.

src/scss/constants.scss
goran-peoski-work: Yes https://github.com/sparkletown/sparkle/pull/1891#discussion_r679379948

goran-peoski-work:
> Any reason to have the border-color same as the background?

Explanation: https://github.com/sparkletown/sparkle/pull/1891#discussion_r679379948

Note: using a `@mixin background-color($color)` serves as a one line drop-in replacement of `background: $color` when needed. 

Technically it can be done with different mixin or no mixin at all, it will just require adding one or two more lines to places that have `background: $color` and then keep track to sync two properties whenever there are changes.

This mixin provides that syncronization.

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: Still worth moving the const outside for better readibility

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: 1) Why do you have to inherit border-style if it should always be solid
2) Why don't we just use the padding if the border is same color as the background

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: We don't need scoping for the scss variables as they are complied so please move them to the top as the rest of the codebase

mike-lvov: Since the scss variables are not complied into the css and are processed on the module level, please just move them to the top of the file to match the code style of the rest of the codebase

goran-peoski-work: I will move them to be consistent with the style even though module scope isn't required.

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: > if it should always be solid

If the border is set as solid from the start, but no ancestor provides appropriate color, the indicator border will be out of sync.

The setting of the `border-style` property serves as a kind of show/hide switch but I also wanted to leave the possibility of the `border-color` mixin to use different value (or even property) as the switch, provided there's a need of variety.

>  Why don't we just use the padding if the border is same color as the background

A > B > C > D > E
E has to have the border color of A's background. Padding will expose whatever D has painted, not even necessarily D's background.

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Will move them. I don't think it's worth, but also don't find it worth providing arguments one way or another. 
If there are more changes of the same kind needed, a code suggestion would be most welcome.

goran-peoski-work: change made in f70c61dd2eaf3b9d9db27d41da9590ab95cb1fde

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: done f70c61dd2eaf3b9d9db27d41da9590ab95cb1fde

src/scss/constants.scss
mike-lvov: Continuing discussion about this matter in the comment linked above^

mike-lvov:
@goran-peoski-work it doesn't look right
image

mike-lvov:
Also, when you open the popup with the user(Click on the avatar in the chat), it seems that it doesn't reuse the UserAvatar component. Could you please make sure that we don't have avatars rendered other than UserAvatar component for consistency reasons?

image

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: @mike-lvov Is there going to be a "theme" background variable for each possible place `<UserAvatar>` can be used?
@sabrinaorban Is it OK to not show indicator border for the time being?

goran-peoski-work:
> @goran-peoski-work it doesn't look right
> image

@mike-lvov please define "looks right"

Note: Most of `<UserAvatar>` usages don't provide a setting for size (no prop saying `small`, `medium`, `large`) thus they default to the "old style" of calculating avatar element sizes. This might be the cause for the issue.

goran-peoski-work:
> Also, when you open the popup with the user(Click on the avatar in the chat), it seems that it doesn't reuse the UserAvatar component. Could you please make sure that we don't have avatars rendered other than UserAvatar component for consistency reasons?
> 

[OOS] to the issue of indicator size. Possible connection to https://github.com/sparkletown/sparkle/pull/1263. Recommend open separate PR for setting usage of `UserAvatar`, including making the `size` prop mandatory for all of said uses to avoid possibilities like https://github.com/sparkletown/sparkle/pull/1891#issuecomment-891519294

src/components/atoms/UserAvatar/UserAvatar.scss
mike-lvov: @goran-peoski-work There probably will be not more than 2-3 themes. Light/dark theme implies that it should be sufficient to have either dark or light circle border(in any place of usage). So, let's remove that `background mixin` and merge this PR

goran-peoski-work: >  Light/dark theme implies that it should be sufficient to have either dark or light circle border(in any place of usage).

I don't see where the transparent indicator border implies usage of themes in any way. Just because there are white and black background in the example doesn't mean whichever parent the `<UserAvatar>` is put on will have a theme background, thus the question:

> Is there going to be a "theme" background variable for each possible place <UserAvatar> can be used?

@mike-lvov  How can this be ensured?

> let's remove that background mixin and merge this PR

Still waiting @sabrinaorban input to make sure it's OK to not have an indicator border.

sabrinaorban: Yes, we can give up the avatar border!

mike-lvov:
> > @goran-peoski-work it doesn't look right
> > image
> 
> @mike-lvov please define "looks right" and provide steps to reproduce
> 
> Note: Most of `<UserAvatar>` usages don't provide a setting for size (no prop saying `small`, `medium`, `large`) thus they default to the "old style" of calculating avatar element sizes. This might be the cause for the issue.

As far as I understand, if the size prop is absent, we apply the default styles. My concern is that the default styles we have are not consistent with this. As you can see, the border is too big

image

goran-peoski-work:
> As far as I understand, if the size prop is absent, we apply the default styles. My concern is that the default styles we have are not consistent with this. As you can see, the border is too big
> 
> As you can see, the border is too big

NOTE: Instead of replying to the comment above I ended up destroying it then tried to restore a mangled version of it.

Reply:

> As you can see, the border is too big

There are at least two assumptions made here.

Assumption I see the same thing you expect me to see i.e. we perceive it the same way.
Assumption "too big" means the same to both of us without any measurement such as pixels or ratios to put it in perspective.
Anyway, after sparkletown/sparkle#1891 (comment) the border will be not shown (with the minimal change of just removing the mixin use).

@mike-lvov as per sparkletown/sparkle#1891 (comment) can you please create a PR assigned to me for updating that one use and adding proper size to all uses of UserAvatar as it will be easier if we don't rely on the old behavior in the future (i.e. make the size prop mandatory)

src/components/atoms/UserAvatar/UserAvatar.scss
goran-peoski-work: Removed mixin @ 8315836c8183882047216e853ad29c66567e0d2a Border will no longer be visible while the width calculation is still present for future use when appropriate color solution can be produced.

PR #1895, opened 2021-07-23
fixed external link redirect

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/946

* Fixed incorrect event redirect to non-working url
* If room has an external URL, then an event that points to that room will redirect directly to that URL

### DIFF
--- a/src/components/organisms/EventModal/EventModal.tsx
+++ b/src/components/organisms/EventModal/EventModal.tsx
@@ -71,7 +71,8 @@ export const EventModal: React.FC<EventModalProps> = ({
       eventVenue?.rooms?.filter(
         (venueRoom) => getUrlParamFromString(venueRoom.title) === room
       ) || [];
-// @debt fix this hack on 24.07.2021
+
+    // @debt fix this hack on 24.07.2021
     if (selectedRoom) {
       openUrl(selectedRoom?.url);
 
--- a/src/components/organisms/EventModal/EventModal.tsx
+++ b/src/components/organisms/EventModal/EventModal.tsx
@@ -45,12 +45,10 @@ export const EventModal: React.FC<EventModalProps> = ({
     () =>
       eventVenue?.rooms?.find((room) => {
         const { room: eventRoom = "" } = event;
-
         const noTrailSlashUrl = getUrlWithoutTrailingSlash(room.url);
 
         const [roomName] = getLastUrlParam(noTrailSlashUrl);
         const roomUrlParam = getUrlParamFromString(eventRoom);
-
         return roomUrlParam.endsWith(`${roomName}`);
       }),
     [eventVenue, event]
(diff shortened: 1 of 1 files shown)


src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -63,12 +61,21 @@ export const EventModal: React.FC<EventModalProps> = ({
...
+    const [selectedRoom] =
+      eventVenue?.rooms?.filter(
+        (venueRoom) => getUrlParamFromString(venueRoom.title) === room
+      ) || [];
+
+    if (selectedRoom) {
+      openUrl(selectedRoom?.url);
+
+      return;
+    }
```
mike-lvov, line 78: Is it possible to just reuse `useRoom` here? It looks very hacky

alexey-petrov: useRoom is already there ( 57th row ) , but it doesn't work for us in this case
we need to check both event and room instead of just room as done on map

mike-lvov: we should still use `enterRoom` in this case as it handles the case if the URL is not a venue

image

mike-lvov:
```
const selectedRoom =
      eventVenue?.rooms?.find(
        (venueRoom) => getUrlParamFromString(venueRoom.title) === room
      );
```

alexey-petrov:
Comments will be fixed in the next related PR 
Closing it to deploy in time

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -63,12 +61,21 @@ export const EventModal: React.FC<EventModalProps> = ({
 
   const eventLocationToDisplay =
     (event.room || eventVenue?.name) ?? event.venueId;
-
   const goToEventLocation = () => {
     onHide();
 
     const { room = "" } = event;
     const roomUrlParam = getUrlParamFromString(room);
+    const [selectedRoom] =
```
mike-lvov, line 69:
```suggestion
    const roomUrlParam = getUrlParamFromString(room);
    // @debt fix this hack on 24.07.2021
    const [selectedRoom] =
```

src/components/organisms/EventModal/EventModal.tsx
```diff
@@ -63,12 +61,21 @@ export const EventModal: React.FC<EventModalProps> = ({
...
+      eventVenue?.rooms?.filter(
+        (venueRoom) => getUrlParamFromString(venueRoom.title) === room
+      ) || [];
+
+    if (selectedRoom) {
```
mike-lvov, line 74:
```suggestion
// @debt fix this hack on 24.07.2021
    if (selectedRoom) {
```

mike-lvov (approved):
Approving this hack per @alexey-petrov request. Will be done properly tomorrow

PR #1901, opened 2021-07-24
Wrapping into button

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/887


whoaitsaimz:
I dont think this accomplished what was intended.. I still can't tab to the button and then "enter" to open Okta login, so it's still not accessible

PR #1906, opened 2021-07-26
Restored edit/delete event buttons

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/953

Restored edit/delete logic for events

### DIFF
--- a/src/pages/Admin/VenueEventDetails.tsx
+++ b/src/pages/Admin/VenueEventDetails.tsx
@@ -14,7 +14,8 @@ export interface VenueEventDetailsProps {
   setEditedEvent: Function | undefined;
   setShowCreateEventModal: Function;
   setShowDeleteEventModal: Function;
-  className: string;
+  className?: string;
+  isEditable?: boolean;
 }
 
 const VenueEventDetails = ({
@@ -22,12 +23,12 @@ const VenueEventDetails = ({
   setEditedEvent,
   setShowCreateEventModal,
   setShowDeleteEventModal,
-  className,
+  className = "",
+  isEditable = false,
 }: VenueEventDetailsProps) => {
   const startTime = formatTimeLocalised(eventStartTime(venueEvent));
   const endTime = formatTimeLocalised(eventEndTime(venueEvent));
   const startDay = format(eventStartTime(venueEvent), "EEEE LLLL do");
-  const isEditable = !className;
 
   return (
     <div className={className}>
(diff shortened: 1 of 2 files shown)


src/pages/Admin/VenueEventDetails.tsx
```diff
@@ -27,6 +27,7 @@ const VenueEventDetails = ({
   const startTime = formatTimeLocalised(eventStartTime(venueEvent));
   const endTime = formatTimeLocalised(eventEndTime(venueEvent));
   const startDay = format(eventStartTime(venueEvent), "EEEE LLLL do");
+  const isEditable = !className;
```
mike-lvov, line 30: This logic seems hacky

src/pages/Admin/VenueEventDetails.tsx
```diff
@@ -49,6 +50,34 @@ const VenueEventDetails = ({
...
+              <button
+                role="link"
+                className="btn btn-primary"
+                onClick={() => {
+                  setEditedEvent && setEditedEvent(venueEvent);
+                  setShowCreateEventModal(true);
+                }}
+              >
+                Edit
+              </button>
```
mike-lvov, line 66: Where was this taken from?

src/pages/Admin/VenueEventDetails.tsx
alexey-petrov: This is what we had initially when the edit/delete was in place 
Gonna check it

src/pages/Admin/VenueEventDetails.tsx
alexey-petrov: It's the way it was done before edit/delete was deleted

src/pages/Admin/VenueEventDetails.tsx
alexey-petrov: fixed f68b41e

mike-lvov (approved):
@alexey-petrov is this component only used in one place?

alexey-petrov:
> @alexey-petrov is this component only used in one place?

Nope, it's used in 3 files

mike-lvov:
@alexey-petrov did you check the logic there. Is it possible that this PR breaks the usage in that 3 files?

alexey-petrov:
> @alexey-petrov did you check the logic there. Is it possible that this PR breaks the usage in that 3 files?

Not possible, the `classname` check there was exclusive for 1 exact file where `classname` prop was explicitly empty
So by replacing `classname={''}` with `isEditable` we simply make this logic obvious

src/pages/Admin/VenueEventDetails.tsx
```diff
@@ -14,15 +14,17 @@ export interface VenueEventDetailsProps {
...
+  isEditable?: boolean;
 }
 
 const VenueEventDetails = ({
   venueEvent,
   setEditedEvent,
   setShowCreateEventModal,
   setShowDeleteEventModal,
-  className,
+  className = "",
```
0xdevalias, line 26: className can probably be left undefined rather than defaulted to an empty string. At least as far as the DOM elements are concerned

PR #1907, opened 2021-07-26
[env/github] replace rko image with new transparent one

malaikaaiyar:
looks like this now:

Screen Shot 2021-07-26 at 12 01 43 PM


mike-lvov (approved):
LGTM

PR #1909, opened 2021-07-26
Feature/chat infinite scroll

dmitrymarchuk:
wrong PR, will create another one


package.json
```diff
@@ -55,6 +55,7 @@
     "react-dom": "^16.13.1",
     "react-hook-form": "^5.7.2",
     "react-image-file-resizer": "^0.3.8",
+    "react-infinite-scroll-component": "^6.1.0",
```
alexey-petrov, line 58: just want to make sure that we're using the best option here
is this lib well-optimized, frequently updated and doesn't add too much weight to the bundle?

src/components/molecules/Chatbox/Chatbox.tsx
```diff
@@ -24,6 +30,7 @@ import { ChatboxThreadControls } from "./components/ChatboxThreadControls";
 import { ChatboxOptionsControls } from "./components/ChatboxOptionsControls";
 
 import "./Chatbox.scss";
+import InfiniteScroll from "react-infinite-scroll-component";
```
alexey-petrov, line 33: that should be added to the lib imports in this file, move it to the top

src/components/molecules/Chatbox/Chatbox.tsx
```diff
@@ -34,14 +41,20 @@ export interface ChatboxProps {
...
+
+const _ChatBox: React.FC<ChatboxProps> = ({
   messages,
   venue,
   sendMessage,
   sendThreadReply,
   deleteMessage,
   displayPoll: isDisplayedPoll,
 }) => {
+  console.assert(messages.length > 0);
```
alexey-petrov, line 54: do we still need it here?

PR #1917, opened 2021-07-26
make sure landing page join event button always shows

malaikaaiyar:
don't know why the logic to only show the 'join event' button if there are no events exists, but this pr gets rid of that so the join button is always there


mike-lvov (approved):
LGTM

PR #1155, opened 2021-02-11
add generic Twilio VideoParticipant component + useParticipantState hook

0xdevalias:
**Note:** These changes were cherry picked out of my 'clean up the legacy playa templates' branch, and made into a standalone PR to simplify review + discoverability.

This component can be used to replace all of the existing `LocalParticipant`, `RemoteParticipant`, `Participant`, etc components.

I believe this also fixes https://github.com/sparkletown/internal-sparkle-issues/issues/456 (need to confirm)

This is one part of the changes that fix https://github.com/sparkletown/internal-sparkle-issues/issues/463

Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/341

Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/639

Partially fixes https://github.com/sparkletown/internal-sparkle-issues/issues/645

## TODO

- [ ] Refactor `src/components/organisms/Room/VideoErrorModal.tsx` into `src/components/organisms/Video`
- [ ] Refactor `src/components/organisms/Room/VideoErrorModal.scss` into `src/components/organisms/Video`
- [ ] Refactor `src/components/templates/FireBarrel/useVideo.tsx` into `src/hooks/twilio.ts`
  - **Note:** @mike-lvov did this + some related cleanup in https://github.com/sparkletown/sparkle/pull/1299#discussion_r623416528 in commit https://github.com/sparkletown/sparkle/commit/f11e9e2afc468857ecb95202bd490d3b2e37592b for that PR. Make sure we align/don't override one another.
- [ ] Refactor all of the existing components to use this, including cleaning up any legacy features (eg. bartender, 'kick user', etc)
- [ ] Remove all of the old components that are no longer relevant
- [ ] Remove any old styling that is no longer relevant
- [ ] Remove any old assets (images, etc) that are no longer relevant
- [ ] Check that this seems to work correctly in all of the venues that use it
  - [ ] Jazz Bar / Music Bar
  - [ ] Auditorium
  - [ ] Fire Barrel
  - [ ] etc?
- [ ] Confirm that this actually fixes https://github.com/sparkletown/internal-sparkle-issues/issues/456 (and if not, make sure it does, or remove these 'closing keywords')
- [ ] Anything else?

## OOS

- Refactoring all of the various components that use twilio video to use the `useVideo` hook

PR #1919, opened 2021-07-27
remove legacy Playa template + related components

0xdevalias:
FYI: This was my WIP branch cleaning up the remaining aspects of the Playa template and similar (there were a number of previous PRs on this similar topic that each tackled smaller chunks of this work if I remember correctly)

It stalled when I realised I had to unify the disparate Twilio video components in https://github.com/sparkletown/sparkle/commit/1b0361b1525bcdb02284d0c6654f2954946fb65f, which I then decided would make more sense to separate out into a standalone PR, which was done in:

- https://github.com/sparkletown/sparkle/pull/1155

I haven't looked at this branch for quite some time, so I couldn't say perfectly what state it's in.. but my memory is that there is probably so more work that was required still to refactor/clean up the old legacy Twilio video state things and 'wire them up' to use the new generic one.

I believe there is also still a bunch of work remaining to identify / cleanup the remainder of the components/styles/etc related to the Playa template and related legacy features. Unfortunately, due to how the styles were written back then (in global scope/reused across the app in haphazard ways), it's not as easy as just blindly deleting all of the related files/components, as there is a fairly high chance that will break some other part of the app in a subtle/non-obvious way. That's why I was going through methodically checking/removing things as I went within this PR, as can be seen from the individual 'focused' commits.

I also found this local stash, which I created a patch for that seems to be my 'local state' when i was last working on this, including a bunch of changes that hadn't yet been 'cleaned up and committed' + a whole bunch of notes/questions/etc. I would suggest applying/working with the content within this patch before merging in `staging`/dealing with any merge conflicts there, as that will make your life a whole lot easier than trying to do it afterwards IMO. Hopefully it's useful to whoever ends up finishing off and landing this work:

- [patch-devalias-cleanup-legacy-playa-local-stash.diff.txt](https://github.com/sparkletown/sparkle/files/6882572/patch-devalias-cleanup-legacy-playa-local-stash.diff.txt)

Once the current chaos of events and related features is over, I would suggest that it would be highly valuable to put some time to finishing off this PR and landing it, as it will remove a TON of legacy code that is currently choking up the platform, and needlessly complicating new engineer onboarding/etc through providing bad examples of code patterns/etc.

(cc // @sparkletown/engineering @jonboutelle @sparkletown/product @sparkleed FYI)


0xdevalias:
Note that while this looks like a lot of conflicting files, in reality, most of those 'conflicts' are because the file is likely deleted in this PR. So I would advise looking closely at that before deciding that 'there is too much drift' in this PR and throwing it away, as it is likely actually far less effort to solve the merge conflicts than it will be to re-do the work done here in a safe/sane way:

image

whoaitsaimz:
Hey @sofisparkle and @jonboutelle it is probably worth letting Nordbeaver / our engineers take a look at this work to remove outdated code as a cleanup exercise before lots of new custom BM-related code is merged in

cc @Sparkle-T  @Sparkle-Nat

mike-lvov (approved):
LGTM

PR #1921, opened 2021-07-27
BM Admin Adjustments: available venues, orders & descriptions

yurii-lubynets:
closes: https://github.com/sparkletown/internal-sparkle-issues/issues/955


yurii-lubynets:
closes as the incorrect base. Will be updated only for `env/burn` https://github.com/sparkletown/sparkle/pull/1927

PR #1933, opened 2021-07-30
fixed table row number calculation + fixed table update capacity value

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/811


src/api/table.ts
```diff
@@ -18,11 +18,14 @@ export const updateVenueTable = async ({
   subtitle,
   capacity,
 }: UpdateVenueTableProps) => {
+  const updatedCapacity = parseInt(capacity.toString(), 10);
```
denisdimitrov, line 21: You are converting `capacity` to string and then you are parsing it back to number, why is that?

src/api/table.ts
```diff
@@ -18,11 +18,14 @@ export const updateVenueTable = async ({
...
+  const tableColumns = tableOfUser?.columns ?? 1;
```
denisdimitrov, line 22: Why is the default `1`? This should be imported from the `settings.ts`, if it doesn't exist there, then we should add it.

src/api/table.ts
```diff
@@ -18,11 +18,14 @@ export const updateVenueTable = async ({
...
   const updatedTable = {
     ...tableOfUser,
     title,
     subtitle,
-    capacity,
+    capacity: updatedCapacity,
+    rows: Math.ceil(updatedCapacity / tableColumns),
```
denisdimitrov, line 28: Why `.ceil` and not `.round`? If `updatedCapacity / tableColumns` is `99.01` it will be rounded to `100` instead of `99` where `.round` will use the closer value.

denisdimitrov (changes requested):

src/api/table.ts
alexey-petrov: the issue is that venue update type has the capacity listed as "string" type, when it should be a number
changing `capacity` type would require a lot of complementary changes throughout the application, because venue entity is used everywhere, thus I decided to implement this little hack

src/api/table.ts
alexey-petrov: we should have integers that big because 10 is the max available capacity here
but I've replaced it with `round` anyway in 2bd967a

src/api/table.ts
alexey-petrov: replaced by MIN_TABLE_CAPACITY, fixed in 2bd967a

src/api/table.ts
sunny-viktoryia: can't we write something like this: 
```suggestion
  const updatedCapacity = parseInt(capacity, 10);
```
<hr />

if you both agree on having it as is, can you @alexey-petrov add a `@debt` comment here that would describe why it's needed so that we could fix that during the next code clean-up?

alexey-petrov: I've added @debt comment there, so that we know the source of the issue
3aac5ee

ermiaSparkle:
@alexey-petrov 
Please check this one!

alexey-petrov:
@ermiaSparkle really thought we've merged it long time ago
should be alright to be approved & merged tho
@sparkletown/engineering

src/api/table.ts
goran-peoski-work: @alexey-petrov safest way of converting to string is the function `String()` (note, no `new` keyword in front of it). 
Something like `String( x ?? '' )` will not break with error if `Symbol` is provided and will make `null` and `undefined` empty strings ( or whichever default like `'10'` ) unlike the possibility of `null.toString()`, should data be 🤷 unexpected of type.

Either way, an update to the props is welcome, something like `capacity: number | string;` to closely reflect the reality of the issue

PR #1940, opened 2021-08-04
Admin V3 - Fixed header text alignment and missing icons

dmitrii-zolotukhin:
fixed sparkletown/internal-sparkle-issues#988

### DIFF
--- a/src/components/organisms/AdminVenueView/AdminVenueView.scss
+++ b/src/components/organisms/AdminVenueView/AdminVenueView.scss
@@ -19,35 +19,34 @@
     width: 100%;
     flex-direction: row;
   }
-}
 
-.ViewTab {
-  position: relative;
-  flex-grow: 1;
-  text-align: center;
-  align-items: center;
-  cursor: pointer;
-  color: rgba($white, 0.6);
-  text-decoration: none;
+  &__tab {
+    position: relative;
+    flex-grow: 1;
+    text-align: center;
+    cursor: pointer;
+    color: rgba($white, 0.6);
+    text-decoration: none;
 
-  &:hover {
-    color: $white;
+    &:hover {
+      color: $white;
+    }
+
+    &--selected {
+      color: $white;
+      background-color: $primary;
+    }
+
+    @media (max-width: $media-breakpoint--lg) {
+      width: 50%;
+    }
+
+    @media (max-width: $media-breakpoint--md) {
+      width: 100%;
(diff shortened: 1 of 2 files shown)


goran-peoski-work:
Possible duplicate https://github.com/sparkletown/sparkle/pull/1871

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -41,15 +44,29 @@ export const AdminVenueView: React.FC = () => {
     currentVenue: venue,
   } = useConnectCurrentVenueNG(venueId);
 
+  const getTabIcon = (tab: string) => {
```
denisdimitrov, line 47:
```suggestion
  const getTabIcon = (tab: AdminVenueTab) => {
```

src/components/organisms/AdminVenueView/AdminVenueView.scss
```diff
@@ -18,34 +18,36 @@
...
-
-    &:hover {
-      color: $white;
-    }
-
-    &--selected {
-      color: $white;
-      background-color: $primary;
-    }
+.ViewTab {
```
denisdimitrov, line 24: All of the styles should be following BEM and should be inside the container/wrapper styles, because this might override or collide with another style with the same name, imported from another .scss file.

This is also 'hidden' and people won't understand that it exists as a className, if we create a component ViewTab and it has this classname it will collide.

Moving it back inside then wrapper with `__` prefix would be the way to go to keep it encapsulated and name specific.

denisdimitrov (changes requested):

src/components/organisms/AdminVenueView/AdminVenueView.scss
dmitrii-zolotukhin: @denisdimitrov fixed in 10746811a3fd6df74bf802630125e13acc06a33a

src/components/organisms/AdminVenueView/AdminVenueView.tsx
dmitrii-zolotukhin: @denisdimitrov fixed in 76c5255b0a01256e6e7d7dc76f687b10e1f76ee1

denisdimitrov (approved):
LGTM!

denisdimitrov (approved):
LGTM!

PR #1941, opened 2021-08-04
[SCHEDULE V3] Changed timeline logic + styles

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/987

* updated styles according to https://www.figma.com/file/92gx2sTMvCc2W0NhgAG7pd/SparkleVerse-UI?node-id=58%3A4116
* Moved full schedule download to the bottom
* timeline dates with no events are now disabled
* max/min timeline range is now dependant on existing events


src/utils/schedule.ts
```diff
@@ -10,3 +10,19 @@ export const sortScheduleRoomsAlphabetically = (rooms: LocationEvents[]) => {
...
+    case 1:
+      return `${day}st`;
+    case 2:
+      return `${day}nd`;
+    case 3:
+      return `${day}rd`;
+    default:
+      return `${day}th`;
+  }
+};
```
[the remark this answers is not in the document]
alexey-petrov: removed function in 05827d5, thanks!

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+
+      return resultObj;
+    },
+    { minDate: 0, maxDate: 0 }
+  );
+  const milisecondStart = new Date(secondsToMilliseconds(minDate)).getTime();
+  const milisecondFinish = new Date(secondsToMilliseconds(maxDate)).getTime();
+  const dayDifference = Math.round(
+    millisecondsToHours(milisecondFinish - milisecondStart) / 24
+  );
```
sunny-viktoryia, line 134: can we use `differenceInDays` to calculate the number of days and `fromUnixTime` to convert minDate and maxDate from seconds to the date object? 🙃 

so that it would look like:
```js
const dayDifference = differenceInDays(fromUnixTime(maxDate), fromUnixTime(minDate));
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
-          formatOtherDate: (dateOrTimestamp) => format(dateOrTimestamp, "E"),
+          formatOtherDate: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
+          formatTomorrow: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
         });
       }
     };
 
-    return range(SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
+    return range(dayDifference).map((dayIndex) => {
       const day = addDays(firstDayOfSchedule, dayIndex);
```
sunny-viktoryia, line 149: I think `firstDayOfSchedule` should be calculated based on events information:  max(minDate, today)
and then we can get rid of `isScheduleTimeshifted` logic. Because we'll show only the event dates that have at least something.

sunny-viktoryia (commented):
some comments / suggestions. 🙃

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+      if (
+        event.start_utc_seconds + minutesToSeconds(event.duration_minutes) >
+        resultObj.maxDate
+      ) {
+        const secondsLength = minutesToSeconds(event.duration_minutes);
+
+        resultObj.maxDate = event.start_utc_seconds + secondsLength;
+      }
+
+      return resultObj;
```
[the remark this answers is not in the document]
alexey-petrov: fixed 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+        resultObj.minDate =
+          resultObj.minDate === 0 ? event.start_utc_seconds : resultObj.minDate;
+
+        const secondsLength = minutesToSeconds(event.duration_minutes);
+        resultObj.maxDate =
+          resultObj.maxDate === 0
+            ? event.start_utc_seconds + secondsLength
+            : resultObj.maxDate;
+      }
+
```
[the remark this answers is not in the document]
alexey-petrov: fixed in 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+        resultObj.maxDate
+      ) {
+        const secondsLength = minutesToSeconds(event.duration_minutes);
+
+        resultObj.maxDate = event.start_utc_seconds + secondsLength;
+      }
+
+      return resultObj;
+    },
+    { minDate: 0, maxDate: 0 }
```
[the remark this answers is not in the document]
alexey-petrov: removed that complicated func in 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -89,21 +97,66 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
 
-    return range(SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
+    return range(dayDifference).map((dayIndex) => {
       const day = addDays(firstDayOfSchedule, dayIndex);
+
+      const daysWithEvents = relatedVenueEvents.filter(
+        isScheduleTimeshifted
+          ? isEventWithinDate(day)
+          : isEventWithinDateAndNotFinished(day)
+      );
```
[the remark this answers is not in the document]
alexey-petrov: fixed 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -77,15 +89,25 @@ $weekday--margin-right: 10px;
...
-      border: none;
+    }
+
+    &--disabled {
+      pointer-events: none;
+      cursor: default;
+
+      button {
+        cursor: default;
+        color: rgba(255, 255, 255, 0.383);
```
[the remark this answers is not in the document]
alexey-petrov: fixed 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -77,15 +89,25 @@ $weekday--margin-right: 10px;
...
       text-shadow: 0 2px 4px opaque-black(0.3);
       box-shadow: 0 3px 6px opaque-black(0.2);
-      border: none;
+    }
+
+    &--disabled {
+      pointer-events: none;
+      cursor: default;
+
+      button {
```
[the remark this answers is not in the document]
alexey-petrov: fixed 92a0ade

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -139,7 +139,7 @@ export const NavBar: React.FC<NavBarPropsType> = ({ hasBackButton = true }) => {
     return false;
   }, [radioFirstPlayStateLoaded]);
 
-  const [isEventScheduleVisible, setEventScheduleVisible] = useState(false);
+  const [isEventScheduleVisible, setEventScheduleVisible] = useState(true);
```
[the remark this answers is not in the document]
alexey-petrov: fixed 92a0ade

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -70,6 +81,7 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
 
   const scheduledStartDate = sovereignVenue?.start_utc_seconds;
 
+  // @debt: probably will need to be re-calculated based on minDate instead of startOfDay.Check later
```
sunny-viktoryia, line 84: image

looks like something got broken a bit. Both the dates are disabled although there are events. 😬

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -84,26 +96,63 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
   } = useVenueEvents({
     venueIds: relatedVenueIds,
   });
-
   const isLoadingSchedule = isLoading || isEventsLoading;
 
   const [selectedDayIndex, setSelectedDayIndex] = useState(0);
 
+  const minDate = useMemo(() => {
+    if (relatedVenueEvents.length) {
```
sunny-viktoryia, line 104: I believe we should we okay without that check, we can add an extra argument to Math.min(..array, SAFE_VALUE) if needed.

sunny-viktoryia (commented):
mmm looks better. 😍 
a bit more comments and suggestions from me. 🙃

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -84,26 +96,63 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+      return Math.min(
+        ...relatedVenueEvents.map((event) =>
+          secondsToMilliseconds(
+            event.start_utc_seconds + minutesToSeconds(event.duration_minutes)
+          ) > startOfToday().getTime()
+            ? event.start_utc_seconds
+            : Number.MAX_SAFE_INTEGER
```
sunny-viktoryia, line 111: and also, maybe instead of this check we should better create a const `liveAndFutureEvents` (instead of `relatedVenueEvents`) - and iterate through them everywhere? as we won't need the events that have already ended.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -35,25 +39,36 @@ $weekday--margin-right: 10px;
...
+    align-items: center;
+    background-color: opaque-black(0.9);
+    height: $schedule-btn--height;
     padding-left: $NavBarSchedule--padding;
 
     margin: $spacing--md 0;
   }
 
-  &__download-schedule-btn {
+  & .NavBarSchedule__download-schedule-btn {
```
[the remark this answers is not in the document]
alexey-petrov: we need to overwrite default btn styles here

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -56,6 +64,9 @@ export interface NavBarScheduleProps {
   venueId: string;
 }
 
+const minRangeValue = 0;
+const maxRangeValue = 1;
+
```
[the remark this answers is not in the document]
alexey-petrov: I've changed the logic of using those vars a bit, including `math.min` approach
but it's still needed so that we have correct range values for related events

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: should be good now, fixed in 365daad

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: moved it into a separate var 365daad

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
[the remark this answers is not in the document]
sunny-viktoryia: aha, I see. Can you add a `@debt` comment to it then describing why it's this way?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -84,26 +97,69 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          formatOtherDate: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
+          formatTomorrow: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
         });
       }
     };
 
-    return range(SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(firstDayOfSchedule, dayIndex);
+    return range(dayDifference).map((dayIndex) => {
+      const day = addDays(todaysDate, dayIndex);
```
sunny-viktoryia, line 154: i don't think it's correct.
let's say we have events from 1 to 2 of September, dayDifference = 2
but instead we'll show only today, tomorrow (which is not correct - these dates do not have events).

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -84,26 +97,69 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          : Number.MAX_SAFE_INTEGER
+      ),
+    [relatedVenueEvents]
+  );
+
+  const minDate = useMemo(
+    () =>
+      liveAndFutureEvents.length
+        ? Math.min(...liveAndFutureEvents)
+        : minRangeValue,
```
sunny-viktoryia, line 120: can we write like this ?
```suggestion
      Math.min(...liveAndFutureEvents, minRangeValue)
```

sunny-viktoryia (commented):
comments and suggestions. 🙃

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: fixed in 85038ed

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: code block fixed in 85038ed

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -1,22 +1,25 @@
...
-import { SCHEDULE_SHOW_DAYS_AHEAD } from "settings";
-
 import {
   LocationEvents,
   PersonalizedVenueEvent,
   VenueEvent,
 } from "types/venues";
 
+import { PLATFORM_BRAND_NAME } from "settings";
+
```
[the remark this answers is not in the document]
alexey-petrov: order fixed in a52f7df

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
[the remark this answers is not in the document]
alexey-petrov: yep, added in 2c14dee

src/utils/event.ts
```diff
@@ -88,3 +88,11 @@ export const eventTimeComparator = (a: VenueEvent, b: VenueEvent) => {
...
   return a.duration_minutes - b.duration_minutes;
 };
+
+export const getEventDayRange = (daysInBetween: number) => {
+  if (daysInBetween === 2) {
+    return daysInBetween + 1;
+  }
+
+  return daysInBetween + 2;
+};
```
sunny-viktoryia, line 98: can you explain the meaning of these magic numbers? and give the meaningful names or add a comment describing the logic?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +110,88 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
   } = useShowHide(false);
 
+  const liveAndFutureEvents = useMemo(
+    () =>
+      relatedVenueEvents.filter(
+        (event) =>
+          secondsToMilliseconds(
+            event.start_utc_seconds + minutesToSeconds(event.duration_minutes)
+          ) > startOfToday().getTime()
+      ),
```
sunny-viktoryia, line 120: can't we just use `relatedVenueEvents.filter(isEventLiveOrFuture)` ?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +110,88 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+    : millisecondsToSeconds(new Date(secondsToMilliseconds(minDate)).getTime());
+
+  const maxDate = useMemo(
+    () =>
+      Math.max(
+        ...liveAndFutureEvents.map(
+          (event) =>
+            event.start_utc_seconds + minutesToSeconds(event.duration_minutes)
+        ),
+        firstRangeDayTimeValue + 1
```
sunny-viktoryia, line 156: why +1?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +110,88 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+        : minRangeValue,
+    [firstLiveEvent, liveEventsMinimalStartValue]
+  );
+
+  const firstRangeDayTimeValue = isDateLessOrEqualsToday({
+    valueSource: secondsToMilliseconds(minDate),
+    valueTarget: millisecondsToSeconds(startOfToday().getTime()),
+  })
+    ? millisecondsToSeconds(todaysDate.getTime())
+    : millisecondsToSeconds(new Date(secondsToMilliseconds(minDate)).getTime());
```
sunny-viktoryia, line 147: can't we write something like `getUnixTime(max(minDate, startOfToday()))` ?
where `getUnixTime` and `max` are from `date-fns`.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +110,88 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
 
-    return range(SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(firstDayOfSchedule, dayIndex);
+    return range(dayDifference).map((dayIndex) => {
+      const firstScheduleDate = isDateLessOrEqualsToday({
+        valueSource: secondsToMilliseconds(minDate),
+        valueTarget: millisecondsToSeconds(startOfToday().getTime()),
+      })
+        ? todaysDate
+        : new Date(secondsToMilliseconds(minDate));
```
sunny-viktoryia, line 185: looks like we do pretty same calculations for `firstRangeDatimeValue`. I think we can reuse one of them.

sunny-viktoryia (commented):
comments / suggestions

src/utils/event.ts
alexey-petrov: yeah, basically `differenceInDays` returns days in between of the target timeline, so we need to add certain amount of days that depends on the day difference result, thus letting us have proper timeline

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: + 1 is needed to form a `daysInBetween` timeline and mitigate possible range error

sunny-viktoryia: could you add a comment, describing that?

sunny-viktoryia:
i think we should move toggle below the timeline. + looks like we need to fix the styles a bit - there is a half-transparent space on the left of the black timeline.
image

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -18,24 +21,30 @@ import {
...
+
+// Disabled as per designs. Up for deletion if confirmied not necessary
+// import { ScheduleVenueDescription } from "components/molecules/ScheduleVenueDescription";
+
+import { Button } from "components/atoms/Button";
 
 import { prepareForSchedule } from "./utils";
 
 import "./NavBarSchedule.scss";
+import { millisecondsToSeconds } from "date-fns/esm";
```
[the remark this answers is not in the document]
alexey-petrov: fixed in b41ef61

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -172,30 +247,32 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
   return (
     <div className={containerClasses}>
-      {venueId && <ScheduleVenueDescription venueId={venueId} />}
-      {/* {!isLoadingSchedule && (
+      {/* Disabled as per designs. Up for deletion if confirmied not necessary */}
+      {/* {venueId && <ScheduleVenueDescription venueId={venueId} />} */}
+      {!isLoadingSchedule && (
         <div className="NavBarSchedule__download-buttons">
-          {hasSavedEvents && (
+          {/* Enable when Schedule V3 bookmarked events are ready */}
```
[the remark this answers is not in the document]
alexey-petrov: refactored and enabled bookmarked schedule download logic b41ef61

src/utils/time.ts
```diff
@@ -251,3 +253,13 @@ export const getDayInterval = (date: Date | number) => ({
...
+
+export const isDateLessOrEqualsToday = ({
+  valueSource,
+  valueTarget,
+}: {
+  valueSource: number;
+  valueTarget: number;
+}) => {
+  return max([valueSource, valueTarget]) <= startOfToday();
+};
```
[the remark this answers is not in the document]
alexey-petrov: renamed arguments there b41ef61

src/utils/event.ts
alexey-petrov: added comments in b41ef61

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: yep, fixed b41ef61

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: yep, added comment in b41ef61

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: will look like `getUnixTime(
    max([
      new Date(secondsToMilliseconds(minDate)).getTime(),
      todaysDate.getTime(),
    ])
  );`
b41ef61

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: yep, fixed b41ef61

alexey-petrov:
> i think we should move toggle below the timeline. + looks like we need to fix the styles a bit - there is a half-transparent space on the left of the black timeline.
> image

Fixed styles + moved it down in b41ef61

src/utils/event.ts
```diff
@@ -88,3 +88,13 @@ export const eventTimeComparator = (a: VenueEvent, b: VenueEvent) => {
...
 };
+
+export const getEventDayRange = (daysInBetween: number) => {
+  if (daysInBetween === 2) {
+    // add 1 day to daysInBetween to form a full timeline, if we already have exactly 2 days difference
+    return daysInBetween + 1;
+  }
+
+  // add 2 days to daysInBetween to form a full timeline
+  return daysInBetween + 2;
```
sunny-viktoryia, line 99: why do we need 2 days here?.. shouldn't we always add just +1?

see here: the last date is disabled because it does not have events. 
image

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -133,100 +220,100 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
   return (
     <div className={containerClasses}>
-      {venueId && <ScheduleVenueDescription venueId={venueId} />}
-      {/* {!isLoadingSchedule && (
+      {/* Disabled as per designs. Up for deletion if confirmied not necessary */}
+      {/* {venueId && <ScheduleVenueDescription venueId={venueId} />} */}
+      {!isLoadingSchedule && (
         <div className="NavBarSchedule__download-buttons">
-          {hasSavedEvents && (
+          {isShowPersonalDownloadBtn ? (
```
sunny-viktoryia, line 292: shouldn't we better use `&&` here?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +112,93 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          ? new Date(liveEventsMinimalStartValue).getTime()
+          : millisecondsToSeconds(todaysDateTime)
+        : minRangeValue,
+    [firstLiveEvent, liveEventsMinimalStartValue]
+  );
+
+  const firstRangeDayTimeValue = isDateLessOrEqualsToday({
+    dateValue: secondsToMilliseconds(minDate),
+    targetDateValue: millisecondsToSeconds(startOfToday().getTime()),
+  });
```
sunny-viktoryia, line 149: `isDateLessOrEqualsToday` hints that the result of the function is `boolean` because of `is` in the beginning, though the const name is `firstRangeDayTimeValue` which does not sound like a boolean, please change one or the other.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +112,93 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+
+  const firstRangeDateInSeconds = getUnixTime(
+    max([
+      new Date(secondsToMilliseconds(minDate)).getTime(),
+      todaysDate.getTime(),
+    ])
```
sunny-viktoryia, line 155: still can't get why can't you just compare dates without all of these transforms date -> milisseconds -> date -> to milliseconds. Looks like a loooooot of unnecessary work. 
`max` should accept a date array, you don't have to parse everything to ms, so `minDate` (if it's date) and startOfToday() should be enough.

am I wrong?.. I still can't get why it looks so complicated. Maybe I'm losing something, please, explain..

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
sunny-viktoryia: moreover, the name `isDateLessOrEqualsToday` does not sound like the function that takes 2 arguments.
for me the functionality of it, based on name is:
(date) => boolean
if date is later than today, then it's `true`, otherwise `false`.
although, we have some strange arguments like "targetDateValue". What is target here (based on the function name), not sure...

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +112,93 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          formatOtherDate: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
+          formatTomorrow: (dateOrTimestamp) => format(dateOrTimestamp, "do"),
         });
       }
     };
 
-    return range(SCHEDULE_SHOW_DAYS_AHEAD).map((dayIndex) => {
-      const day = addDays(firstDayOfSchedule, dayIndex);
+    return range(dayDifference).map((dayIndex) => {
+      const firstScheduleDate = firstRangeDayTimeValue
```
sunny-viktoryia, line 190: is `firstRangeDayTimeValue` going to be undefined or null in any case? or it's just zero-check? if so, could it be more explicit? `firstDayTimeValue === 0` ?

src/utils/time.ts
```diff
@@ -251,3 +253,13 @@ export const getDayInterval = (date: Date | number) => ({
   start: startOfDay(date),
   end: endOfDay(date),
 });
+
+export const isDateLessOrEqualsToday = ({
```
sunny-viktoryia, line 257: please, check my another comment about that function. I find confusing the name of it and a set of arguments which should be passed.
maybe we can create a function that checks only one date? not multiple.

sunny-viktoryia (commented):
the ui looks good (one small bug I noticed: see the comment).

but the code still looks overcomplicated. Although, I guess if it can't get better tomorrow, I'll merge this one and will do the follow-up PR which will clean-up the logic. I feel that we have too many unnecessary date transformations. date-fns is a pretty convinient library, no need to invent bicycles. 🙃

src/utils/time.ts
alexey-petrov: renamed function in 60ff2eb

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: not going to be undefined, it returns boolean in all cases, but I've renamed it in 60ff2eb, so that the naming is more explicit

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: you're right, sorry
fixed in 60ff2eb

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: renamed func + variable, so it's more explicit what it does 60ff2eb

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: yep, converted it to boolean upon assignation 60ff2eb

src/utils/event.ts
alexey-petrov: fixed this logic in 60ff2eb
we still need 2 days, but only if there's 1 event and it's within current date, in that case we're adding 2 days
otherwise 1 day is enough to correctly display all dates

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -59,6 +72,10 @@ export interface NavBarScheduleProps {
   venueId: string;
 }
 
+const minRangeValue = 0;
+const todaysDate = new Date();
+const todaysDateTime = new Date().getTime();
```
sunny-viktoryia, line 77: ideally, these constants should not be stored this way. Actually, todasDate can change if a user enters the platform at 23:58 and stays till the next day - todaysDate and the schedule should be updated.

let's mark it as OOS of the PR and will fix that in the follow-up PR, because otherwise it feels like we'll never merge this one.. 🙈

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +114,95 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     toggle: togglePersonalisedSchedule,
   } = useShowHide(false);
 
+  const liveAndFutureEvents = useMemo(
+    () => relatedVenueEvents.filter(isEventLiveOrFuture),
+    [relatedVenueEvents]
+  );
```
sunny-viktoryia, line 120: I think would be better to add here a `.map` prepareForSchedule, what do you think?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +114,95 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+
+  const firstLiveEvent = liveAndFutureEvents.find(
+    (event) => event.start_utc_seconds === liveEventsMinimalStartValue
+  );
+
+  const minDate = useMemo(
+    () =>
+      firstLiveEvent
+        ? isEventLiveOrFuture(firstLiveEvent)
```
sunny-viktoryia, line 141: is this check needed? sounds like `firstLiveEvent` should always be "live or future" according to the const name and liveAndFutureEvents array.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +114,95 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+          ? new Date(liveEventsMinimalStartValue).getTime()
```
sunny-viktoryia, line 142: use `fromUnixTime` here, please. 
https://date-fns.org/v2.23.0/docs/fromUnixTime

should be something like:
```
fromUnixTime(liveEventsMinimalStartValue)
```

and maybe we should rename liveEventsMinimalStartValue -> liveEventsMinUtcSecs (or Seconds)

sunny-viktoryia (commented):
a few more comments. I still don't like several places in the code here. But I'm already close to agree to merge the changes to have them in staging and to do the clean-up afterwards.
Looks like the ui is fixed. So should be okay to have it earlier in staging for all to have it tested.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: I've used `getunixtime` there as it's more convenient 58b5cf2

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: nope, removed it in 58b5cf2
thanks

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: added map there and removed in cases below 58b5cf2

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: removed `todaysDateTime` completely and will remove `todaysDate` in follow up PR 58b5cf2

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +113,99 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+  const isShowPersonalDownloadBtn = hasSavedEvents && showPersonalisedSchedule;
+  const liveEventsMinimalStartValue = Math.min(
+    ...liveAndFutureEvents.map((event) => event.start_utc_seconds)
+  );
+
+  const firstLiveEvent = liveAndFutureEvents.find(
+    (event) => event.start_utc_seconds === liveEventsMinimalStartValue
+  );
+
+  const minDate = useMemo(
```
sunny-viktoryia, line 145: the name of variable is a bit confusing. Let's rename it to minDateUtcSeconds.
and then we can create a separate 
```js
const minDate = useMemo(() => fromUnixTime(minDateUtcSeconds), [minDateUtcSeconds]);
```

actually, you can check if we ever need utcSeconds. Maybe it's more comfortable to store date object instead?

sunny-viktoryia (commented):
:) one more review.

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -133,105 +227,99 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
-      .sort(eventTimeComparator)
-      .map(
-        prepareForSchedule({
-          usersEvents: userEventIds,
-        })
-      );
-
+    const firstScheduleDate = isMinDateWithinToday
+      ? todaysDate
+      : new Date(secondsToMilliseconds(minDate));
```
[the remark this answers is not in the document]
alexey-petrov: removed duplicate 12ae094

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: renamed that variable in 12ae094
we still need the utcseconds value for some functions down below tho

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -97,21 +113,99 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
 
+  const liveAndFutureEvents = useMemo(
+    () =>
+      [
+        ...relatedVenueEvents,
+        ...relatedVenueEvents,
+        ...relatedVenueEvents,
+        ...relatedVenueEvents,
+        ...relatedVenueEvents,
+      ]
```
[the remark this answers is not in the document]
alexey-petrov: yep, thanks
removed 12ae094

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -6,13 +6,53 @@ $NavBarSchedule--padding: 20px;
...
+}
+
 .NavBarSchedule {
   @include scrollbar;
   position: fixed;
   z-index: z(navbar-schedule);
   top: 0;
   left: 0;
-  height: 100%;
+  height: 91%;
```
sunny-viktoryia, line 55: on my screen I have an empty space between the schedule and the line with the download buttons.
image

can you give a fixed height ($shedule-footer-length) to that footer and then use calc here? 
```
height: calc(100% - $shedule-footer-length);
```
or something similar.

sunny-viktoryia (commented):
let's fix that ui bug and then merge the PR.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
alexey-petrov: nvm fixed in 0deea73

PR #1944, opened 2021-08-04
Remove SCSS warning by quoting ButtonNG variants' names

goran-peoski-work:
Avoids the warning by using quoted strings like
```scss
"white"
```
instead of unqoted values like
```scss
white
```

```powershell
WARNING: You probably don't mean to use the color value white in interpolation here.
It may end up represented as white, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "white").
If you really want to use the color value here, use '"" + $modifier'.

   â•·
98 â”‚     &--#{$modifier} {
   â”‚          ^^^^^^^^^
   â•µ
    src\components\atoms\ButtonNG\ButtonNG.scss 98:10  root stylesheet
```


sunny-viktoryia (approved):
Looks good to me. 🙃

PR #1945, opened 2021-08-04
BM - fix parentId for newly created venues

denisdimitrov:
The parent id had `/` in it and this will cause problems for newly created venues because the parent id should be the id of the venue with a slash.

closes https://github.com/sparkletown/internal-sparkle-issues/issues/994


yurii-lubynets (approved):
LGTM

PR #1948, opened 2021-08-05
Admin V3 -- fix hover on maps defaults

ermiaSparkle:
when we hover it moves the maps so it fixed in this PR.

closes https://github.com/sparkletown/internal-sparkle-issues/issues/982

### DIFF
--- a/src/pages/Admin/BackgroundSelect/BackgroundSelect.scss
+++ b/src/pages/Admin/BackgroundSelect/BackgroundSelect.scss
@@ -1,5 +1,7 @@
 @import "scss/constants";
 
+$map-default-size: 50px;
+
 .BackgroundSelect {
   display: flex;
   justify-content: center;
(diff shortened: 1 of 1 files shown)


src/pages/Admin/BackgroundSelect/BackgroundSelect.scss
```diff
@@ -39,7 +39,7 @@
 
     &:hover {
       border: $border-radius--xxs solid $white;
-      padding-bottom: 60px - $border-radius--xxs;
+      padding-bottom: 50px - $border-radius--xxs;
```
denisdimitrov, line 42: This 50px value and the one used a couple of lines above should be converted into a constant to avoid problems like this.

denisdimitrov (changes requested):

denisdimitrov (approved):
LGTM!

PR #1950, opened 2021-08-06
Admin V3 Spaces - add descriptions to each venue type

goran-peoski-work:
Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/984

Makes the Admin v3 Spaces tab more in line with provided design spec, specifically in the area of displaying info when/and adding new rooms.


functions/venue.js
```diff
@@ -476,23 +476,36 @@ exports.createVenue_v2 = functions.https.onCall(async (data, context) => {
...
   }
 
-  admin.firestore().collection("venues").doc(venueId).update({ rooms });
+  result.writeResult = await admin
+    .firestore()
+    .collection("venues")
+    .doc(venueId)
+    .update({ rooms: result.rooms });
+
+  return result;
```
goran-peoski-work, line 508: NOTE: None of the few places `upsertRoom` was used were handling any return value, so it's safe to add one now for future use.

src/components/organisms/AdminVenueView/components/Spaces/Spaces.scss
```diff
@@ -94,69 +95,13 @@ $icon-size: 20px;
...
-    align-items: center;
-    padding: $spacing--md $spacing--lg;
-    margin: 0 5px;
-    transition: all 400ms $transition-function;
-    background-color: $primary;
-
-    &:hover {
-      background-color: lighten($color: $primary, $amount: 10);
-    }
+    width: 40%;
```
goran-peoski-work, line 100: NOTE: This value is determined as a rule of thumb and it's OOS to this PR to try and generalize all different Admin panels' sizes at this time, although some stubs already exist in `constsnts.scss` as in 

```scss
$map-sidebar-width: 20%;
$chat-sidebar-width: 360px;
$navbar-height: 66px;
$footer-height: 30px;
$chat-input-height: 114px;
$announcement-banner-height: 50px;
$admin-spaces-sidebar-width: 20%;
```

src/components/organisms/AdminVenueView/components/Spaces/Spaces.tsx
```diff
@@ -146,17 +92,31 @@ export const Spaces: React.FC<SpacesProps> = ({ venue, onClickNext }) => {
...
+        venue.rooms?.[data?.roomIndex] ??
+        (data?.rooms)[data?.roomIndex] ??
+        data?.provided?.room;
+      if (!created) return;
+
+      setSelectedRoom(created);
+    },
+    [setSelectedRoom, venue.rooms]
+  );
+
```
goran-peoski-work, line 107: The idea behind this is to work out which is the newly added room from the most relevant source of the information and fall back to the least one. In this case that's the `venue.rooms` that would generally be in-sync through a connect hook.

src/components/organisms/EventModal/EventModal.scss
```diff
@@ -12,7 +11,7 @@ $EventModal--padding-top: 40px;
     padding: $EventModal--padding-top $spacing--xl $spacing--xl;
     display: flex;
     flex-direction: column;
-    background-color: $dark-grey;
+    background-color: $background--modal;
```
goran-peoski-work, line 14: NOTE: just unifying the modal background color under a single constant named for its purpose

src/scss/constants.scss
```diff
@@ -94,15 +96,24 @@ $spacing--xxl: 30px;
 
 $submit-button-diameter: 44px;
 
-$font-size--xxs: 10px;
+// Relative font sizes
+$font-size--document: 1rem;
+$font-size--parent: 1em;
```
goran-peoski-work, line 101: These are useful for resetting font sizes or keeping them relative in child elements as well as relative margins and paddings that make sense for cases when text needs to be moved  to a certain distance (`1em`, `2em` etc.)

goran-peoski-work (commented):
Some notes on the PR code

src/components/molecules/VenueRoomItem/RoomAddModal.scss
```diff
@@ -0,0 +1,31 @@
+@import "scss/constants.scss";
+
+.RoomAddModal {
+  .modal-content {
+    padding: $spacing--xl;
+    background: $background--modal;
+    box-shadow: 0 14px 30px rgba(0, 0, 0, 0.5);
```
denisdimitrov, line 7:
```suggestion
    box-shadow: 0 14px 30px opaque-black(0.5)
```

src/settings.ts
```diff
@@ -305,6 +330,41 @@ export const BURN_VENUE_TEMPLATES: Array<Template> = [
...
 ];
 
+// @debt Refactor this constant into types/templates or similar?
+// NOTE: The Map object holds key-value pairs and remembers the original insertion order of the keys
+// Same can be accomplished by a regular object if created through Object.fromEntries()
+export const BURN_VENUE_TEMPLATES_MAP: Map<VenueTemplate, Template> = new Map(
+  BURN_VENUE_TEMPLATES.map((item) => [item.template, item])
+);
+
+export const SPACES_VENUES_AND_ROOMS: VenueRoom[] = [
```
denisdimitrov, line 340: Can you simplify this name? Maybe VENUE_ROOMS or VENUE_SPACES

src/scss/constants.scss
```diff
@@ -94,15 +96,24 @@ $spacing--xxl: 30px;
...
+
+// Absolute font sizes
+$font-size--2s: 10px;
```
denisdimitrov, line 104: The other names are using `xs` `xxs` `xxxs` same with large. Let's keep it consistent

src/settings.ts
```diff
@@ -305,6 +330,41 @@ export const BURN_VENUE_TEMPLATES: Array<Template> = [
   },
 ];
 
+// @debt Refactor this constant into types/templates or similar?
+// NOTE: The Map object holds key-value pairs and remembers the original insertion order of the keys
+// Same can be accomplished by a regular object if created through Object.fromEntries()
+export const BURN_VENUE_TEMPLATES_MAP: Map<VenueTemplate, Template> = new Map(
```
denisdimitrov, line 336: The only benefit to use Map here would be to pass object as key which is not needed, the key is string and this can be simplify to plain object instead.

denisdimitrov (changes requested):

functions/venue.js
```diff
@@ -476,23 +476,36 @@ exports.createVenue_v2 = functions.https.onCall(async (data, context) => {
...
 
   if (!doc || !doc.exists) {
     throw new HttpsError("not-found", `Venue ${venueId} not found`);
   }
-  const docData = doc.data();
-  let rooms = docData.rooms;
 
-  if (typeof roomIndex !== "number") {
-    rooms = [...rooms, room];
+  result.rooms = doc.data().rooms ?? [];
```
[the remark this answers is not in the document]
goran-peoski-work: I will take a look at it again. Few notes:

- I had one working version assigning the results differently It just added few more lines of code, didn't really change anything.
- The result of `doc.data()` isn't being mutated, it's just the removal of `docData` variable that makes it look like there's an issue.
- I'm unsure though about removing the `if` check, since I haven't gone that extensively through all the uses and the possible values therein. I'd leave it as is
- The idea of `result` being on top is to have visible cue the input params i.e. `data` are also attached to the returning result.

functions/venue.js
```diff
@@ -476,23 +476,36 @@ exports.createVenue_v2 = functions.https.onCall(async (data, context) => {
...
+
+  if (typeof roomIndex === "number") {
```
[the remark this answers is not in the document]
goran-peoski-work: > Can you add some documentation here, why is the `else` adding the room at the end when the index is missing?

The only non-speculative documentation I can add is `it used to be like that`, but your guess is pretty much in line to mine.
I'm not going deeper into how this particular `upsert` works as that requires investigation into something OOS and for the purpose of this PR it's good enough to keep the original logic weather it's OK or not.

functions/venue.js
```diff
@@ -476,23 +476,36 @@ exports.createVenue_v2 = functions.https.onCall(async (data, context) => {
 
 exports.upsertRoom = functions.https.onCall(async (data, context) => {
   checkAuth(context);
+
+  const result = {
+    provided: data,
```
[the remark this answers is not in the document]
goran-peoski-work: `provided` used since `data` is anything and overusing `data` is like overusing `this` everywhere: provides no additional context since everything at the end is data.

src/settings.ts
goran-peoski-work: `plain object` doesn't guarantee order of properties unless is created by `Object.fromEntries()` as per the comment, which at the end is as not much simpler than using `Map`.

Considering Maps guarantee ordering of entries, I disagree the only benefit would be to pass some argument to the Map.

goran-peoski-work: There is a work in progress of specifying order of properties in objects (more specifically, how they're being manipulated by different language elements) https://github.com/tc39/proposal-for-in-order, but for now, using `Map` shows a clear signal of the intent 
> The Map object holds key-value pairs and remembers the original insertion order of the keys. 
--
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/scss/constants.scss
goran-peoski-work: Having all names being same length is consistency as well as a an open way forward should there be need of more variables that could end up named like `$font-size--xxxxxl`, especially since the original is already in doubt:
```scss
$font-size--xl: 20px;
// Not sure this is the best naming for these
$font-size--xxl: 30px;
$font-size--xxxl: 50px;
```

src/settings.ts
goran-peoski-work: The issue is of already simplified names in `settings.ts` and in types in general having `venueRooms` but representing different concepts.

There is a need of one name representing both room and venue, like `SPACES_PORTALS` if portal could be superset of venue and room. But as of now, there are few settings already dealing with venues/rooms/templates and arrays of in that file that untangling those would require separate PR (there are a lot of `@debt` comments).

If portal is acceptable, I could use it or some other single term doing the same function.

functions/venue.js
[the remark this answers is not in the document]
goran-peoski-work: > Do you need this field?

Yes https://github.com/sparkletown/sparkle/pull/1950#discussion_r684131622
That's an already existing data that serves as fallback

functions/venue.js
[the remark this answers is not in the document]
goran-peoski-work: Here is a refactor that has the original logic, is using spread, without mutation and without using dot operator to assign to results:

```javascript
  const docRooms = doc.data().rooms ?? [];

  const index = typeof roomIndex === "number" ? roomIndex : docRooms.length - 1;

  const rooms = [
    ...docRooms.slice(0, index),
    room,
    ...docRooms.slice(index + 1),
  ];

  const writeResult = await admin
    .firestore()
    .collection("venues")
    .doc(venueId)
    .update({ rooms });

  return {
    provided: data,
    roomIndex: index,
    rooms,
    writeResult,
  };
```

@denisdimitrov if you think this falls under the "would be great" definition, give it a go and I can make the change, otherwise I'd leave it as currently is.

functions/venue.js
```diff
@@ -476,23 +476,36 @@ exports.createVenue_v2 = functions.https.onCall(async (data, context) => {
...
+    result.rooms[roomIndex] = room;
+    result.roomIndex = roomIndex;
   } else {
-    rooms[roomIndex] = room;
+    result.rooms = [...result.rooms, room];
+    result.roomIndex = result.rooms.length - 1;
   }
 
-  admin.firestore().collection("venues").doc(venueId).update({ rooms });
+  result.writeResult = await admin
```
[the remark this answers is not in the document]
goran-peoski-work: What functions are being returned?

src/scss/constants.scss
denisdimitrov: I know that they can end up extremely long but that's the way they are right now and we need to keep it consistent and the same. It's amateurish for some of the code like font sizes to be using `2s` `3s` and then the rest of the code `xxs` and `sm`

src/settings.ts
denisdimitrov: Fair enough, just saw the comment above the const.

src/settings.ts
denisdimitrov: portal is acceptable, and most of the product team already refers to these as `portals`

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: I miss-spoke, the result from the function is not used anywhere and it can be removed.

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: > provided used since data is anything and overusing data is like overusing this everywhere: provides no additional context since everything at the end is data.

I agree but you are not to decide these things on your own, all of us should be notified about changes like this one and we should all discuss it before actually going with it.

I'm constantly complaining and whining about these things because the code was a complete mess a couple of months ago. It was written by 10 different people and everyone had his own style and standards. It was like 10 different styles scattered all across the project and components.

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: If you really need this data you should use type assertion before returning it. Ex: you can see `get-badges` and the way they add type to the snapshots. `visitDoc.data() as UserVisit`

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: With the risk of breaking something, I'd say keep it as it is. But we should look into it in a refactor PR.

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: If you change `provided` to something more specific I think this would be good to go.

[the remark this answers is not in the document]
goran-peoski-work: > I agree but you are not to decide these things on your own, all of us should be notified about changes like this one and we should all discuss it before actually going with it.

PR review process is a discussion.

Do you think there should be a different place for it?
And if so, what is the threshold upon which discussion is being transferred to everyone and not constrained to a PR review (ideally visible to all of @sparkletown/engineering when a request for review is made).


> If you change provided to something more specific I think this would be good to go.

I'm open to suggestions. I'm at this moment not able to find adequate name that potentially can set a precedent for other functions like `upsertRoom`.

Is `input` specific? Something else like `options`?

functions/venue.js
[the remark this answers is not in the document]
goran-peoski-work: TL;DR: `writeResult` only has one timestamp field `writeTime` which doesn't provide much use, so I will remove it.


Attempt to clear up possible confusion:

The file name where `upsertRoom` is located is `./functions/venue.js` and by it I'm guessing it's being run server side on a Node 10 without TypeScript support.

The comparative's name is `./scripts/get-badges.ts` and by it I'm guessing a client (non-web) CLI script transpiled from TypeScript before executed.

I think it would be OOS to this PR try to put TypeScript assertions into a server side JavaScript file.


On the "type assertion" note, one is already done,
as "safe" as it can `as CreateRoomResult ` be,
by declaring the return type of 

```typescript
export const createRoom = async (
  input: RoomInput_v2,
  venueId: string,
  user: UserInfo
): Promise<CreateRoomResult> => {
```
in `./src/api/admin.ts` with the interface defined as: 
```typescript
export interface CreateRoomResult {
  data: {
    provided: {
      room: Partial<Room>;
      roomIndex: number | unknown;
      venueId: string;
    };
    roomIndex: number;
    rooms: Room[];
    writeResult: admin.firestore.WriteResult;
  };
}
```

src/scss/constants.scss
goran-peoski-work: > Every single change like this one should be discussed with the rest of the team.

define "like this" as to not flood the rest of the team with every line of code that has been changed

goran-peoski-work: For completeness sake, here is the full change: 
```scss
// Relative font sizes
$font-size--document: 1rem;
$font-size--parent: 1em;

// Absolute font sizes
$font-size--2s: 10px;
$font-size--xs: 11px;
$font-size--sm: 12px;
$font-size--md: 14px;
$font-size--lg: 16px;
$font-size--xl: 20px;
$font-size--2l: 30px;
$font-size--3l: 50px;

// @debt replace uses of these with the above
$font-size--xxs: $font-size--2s;
$font-size--xxl: $font-size--2l;
$font-size--xxxl: $font-size--3l;
```

And considering the amount of @debt comments throughout the project,
I'd guess "amateurish" wasn't that much of a blocker for positive changes in new code 
while simultaneously trying to not break the workings of the older.

functions/venue.js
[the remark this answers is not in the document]
goran-peoski-work: refactored in 1bb7aaa7c18844dca1014d68f039d8237c4bbf4d

functions/venue.js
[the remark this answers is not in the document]
goran-peoski-work: removed `writeResult` in 1bb7aaa7c18844dca1014d68f039d8237c4bbf4d

src/settings.ts
goran-peoski-work: renamed in 5d5e8b9982703f31a2fb7b8e15ad090912057cdd

src/scss/constants.scss
denisdimitrov: The border radius styles are using a different naming convention:
```tsx
$border-radius--xxs: 2px;
$border-radius--xs: 4px;
$border-radius--sm: 8px;
$border-radius--md: 12px;
$border-radius--lg: 16px;
$border-radius--xl: 24px;
$border-radius--max: 50%;
```

These should be using the same convention. If you want to improve the names we should have a discussion in Slack about the correct approach and everything should be refactored in another PR. Renaming only some of the global styles will be confusing.

functions/venue.js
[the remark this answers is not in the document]
denisdimitrov: >PR review process is a discussion.

Yes, PR discussion is in progress but I'm the only one reviewing it and the only one actually seeing this new code style. Mike and Viktoria will see this in a week and wonder why it's different and not like the rest of the code. We shouldn't be doing 'shadow' changes like this.

>Do you think there should be a different place for it?
>And if so, what is the threshold upon which discussion is being transferred to everyone and not constrained to a PR >review (ideally visible to all of @sparkletown/engineering when a request for review is made).

When introducing a new code style, it should be discussed with the team in Slack. Especially with the senior devs. They should agree on the new change.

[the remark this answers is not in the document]
goran-peoski-work: > Yes, PR discussion is in progress but I'm the only one reviewing it and the only one actually seeing this new code style.

There is no limitation to whomever can see this PR discussion, even after a PR is closed/merged/etc.

> Mike and Viktoria will see this in a week and wonder why it's different and not like the rest of the code. 

The discussion is here, a simple @ to them will suffice, or share a link to them. PR can be resolved after they have reviewed it.

> We shouldn't be doing 'shadow' changes like this.

Please define what "'shadow change" means. This discussion isn't only visible now and care has been taken to detail changes as they could be read by anyone currently and in the future for historical reference.


> When introducing a new code style, it should be discussed with the team in Slack. Especially with the senior devs. They should agree on the new change.

And this shouldn't stop PR from being created with any code style change. Discussing code style changes before there is some PR makes for an abstract conversation, this way there is a concrete example. Anyone can share a link to this discussion on Slack before PR is merged and after it is closed.

src/scss/constants.scss
goran-peoski-work: During working on a PR, some changes may be required that are on the border of the scope. There are 3 options:

1.  widening the scope of a PR which can be a slippery slope as a PR for one thing becomes a PR for changing everything
2.  keeping strictly to scope which can be a major roadblock to improving code in connected areas
3.  adding enough change, rationale and marking a line with @debt resulting in new issue/PR (possibly currently or eventually)

You're proposing an exclusive either #1 or #2 as the solutions, disregarding #3.
I have done #3.

The first discussion I propose is to determine if #3 is a way forward since the amount of @debt comments through the code base suggest it has been a viable option for a long while presumably to avoid the pitfalls of having too wide or too narrow scope.

Also relevant discussion https://github.com/sparkletown/sparkle/pull/1950#discussion_r686479953

goran-peoski-work:
Merged the month old commit directly from staging, disregarding rebase of the 127 commits in the past month https://github.com/sparkletown/sparkle/pull/1950#event-5278007610

whoaitsaimz:
Would like to review this, I think we should just use the AV1 layout for the admin for now, to get AV3 to a functional place.

goran-peoski-work:
> Would like to review this, but I think we should just direct port the AV1 layout for the admin for now

Not sure what "layout" means in the context. Admin NG has quite different layout of Admin OG  as per https://www.figma.com/file/1Lv4r46oF1ioHZ6a9JRaRR/Admin?node-id=49%3A1982

If "layout" means something else, like which templates are listed, then that can be updated.

goran-peoski-work:
FYI @whoaitsaimz the current state can be seen at https://staging.sparkle.space/admin-ng/venue/

The updated preview link for this PR is now deployed and can be viewed at https://co-reality-staging--preview-pr-1950-55jozs6g.web.app/admin-ng

Here are screenshots from the list
sparkle-venue-add-01

sparkle-venue-add-02

sparkle-venue-add-03

whoaitsaimz:
Much prefer AV1 where names and descriptions are listed out with a possible thumbnail when you click or, perhaps a rollover state. Having to click an "i" icon to understand what it is I'm adding is not great. See my comments in Issue and updated Figma designs for the week.

goran-peoski-work:
> See my comments in Issue and updated Figma designs that bring AV1 directly into AV3.

☝️ 
I saw them before commenting. It is an understandable and logical suggestion, only it is anything but direct as per https://github.com/sparkletown/internal-sparkle-issues/issues/984#issuecomment-921648413 and my thoughts bellow.

If the new UI is a problem, that's easer solved than porting the legacy code stemming from the earliest versions of the system chock-full with landmines of inputs all tucked in a single form which was and probably still will be responsible for issues like 
- https://github.com/sparkletown/internal-sparkle-issues/issues/1067
- https://github.com/sparkletown/internal-sparkle-issues/issues/1163

I do agree with this comment though https://github.com/sparkletown/internal-sparkle-issues/issues/984#issuecomment-922151654 . Ever since the Av3 alpha was first introduced the gap between start-of-the-year planning and current needs was quite noticeable.

I just think building up the missing pieces to Av3 is safer than discarding some of Av3 in favor of Av1 code.

whoaitsaimz:
@goran-peoski-work fair. And I appreciate your thoughts. Let's ensure that all the AV1 features get re-built in a thoughtful way, then.  Had a great chat with Denis today and looking forward to our team meeting tomorrow.

goran-peoski-work:
Marked as "do not merge" and "scrap material". Most of the functionality already superseded by other PRs, some styles can be reused before closing down.

goran-peoski-work:
Superseded. No plans to merge, ever. Some popup UI can be reused. Close if obsolete material.

PR #1953, opened 2021-08-09
Fix bug in admin v1 event modal

malaikaaiyar:
discussed in https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1628154814036200

makes the duration minutes field not cause errors


yurii-lubynets (approved):
LGTM! The optional field should be a good solution. Please don't merge until one more approve

malaikaaiyar:
I tested it in both admin v1 and admin v3 and everything worked as expected

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -73,7 +73,9 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
         description: data.description,
         start_utc_seconds:
           start.unix() || Math.floor(new Date().getTime() / 1000),
-        duration_minutes: data.duration_hours * 60 + data.duration_minutes,
+        duration_minutes: data.duration_minutes
+          ? data.duration_hours * 60 + data.duration_minutes
+          : data.duration_hours * 60,
```
sunny-viktoryia, line 78: maybe this way would look better? what do you think?
```suggestion
        duration_minutes: data.duration_hours * 60 + (data.duration_minutes ?? 0)
```

sunny-viktoryia (approved):
Awesome! 😍 thanks for the fix!

there is a suggestion which i believe should improve readability a bit. :)

src/components/organisms/TimingEventModal/TimingEventModal.tsx
malaikaaiyar: yeah i think that's better! i'll commit it

denisdimitrov (approved):
LGTM!

PR #1955, opened 2021-08-09
Eslint rule for using only arrow functions

dmitrymarchuk:
As we don't use `funcion` declarations and expressions I propose to enforce that in eslint.

### DIFF
--- a/src/components/molecules/Chatbox/useTriggerScrollFix.tsx
+++ b/src/components/molecules/Chatbox/useTriggerScrollFix.tsx
@@ -1,5 +1,5 @@
-import { WithId } from "../../../utils/id";
-import { MessageToDisplay } from "../../../types/chat";
+import { WithId } from "utils/id";
+import { MessageToDisplay } from "types/chat";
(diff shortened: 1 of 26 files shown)


src/components/molecules/Chatbox/useTriggerScrollFix.tsx
```diff
@@ -1,18 +1,18 @@
-import { WithId } from "../../../utils/id";
-import { MessageToDisplay } from "../../../types/chat";
+import { WithId } from "utils/id";
+import { MessageToDisplay } from "types/chat";
 import { useEffect, useRef } from "react";
```
denisdimitrov, line 3: I know this is OOS, but can you please rearrange these imports.

mike-lvov: @denisdimitrov I'm working on the automatic ordering now, so it's okay to leave it as is for now

mike-lvov (approved):
LGTM. Thanks for this improvement

PR #1970, opened 2021-08-10
Schedule3: Events grouping within the selected date

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/999


src/components/molecules/ScheduleEventListNG/ScheduleEventSubListNG.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const ScheduleEventSubListNG: React.FC<ScheduleEventSubListNGProps> = ({
+  events,
+  title,
+}) => {
+  const [showMoreEvents, setShowMoreEvents] = useState(false);
+  const toggleShowMoreEvents = useCallback(() => {
+    setShowMoreEvents(!showMoreEvents);
+  }, [showMoreEvents, setShowMoreEvents]);
```
sunny-viktoryia, line 23: let's use `useShowHide` hook here instead.

src/components/molecules/ScheduleEventListNG/ScheduleEventListNG.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const liveEvents = useMemo(() => daysEvents.filter(isEventLive), [
+    daysEvents,
+  ]);
+  const comingSoonEvents = useMemo(() => daysEvents.filter(isEventSoon), [
+    daysEvents,
+  ]);
+  const laterEvents = useMemo(() => daysEvents.filter(isEventLater), [
+    daysEvents,
+  ]);
```
sunny-viktoryia, line 33: when these constants will be re-calculated? I believe we should use some setInterval to check if the event moved from status "coming soon" to "live" (for example).

src/components/molecules/ScheduleEventListNG/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const isEventLater = (event: PersonalizedVenueEvent) =>
+  !(
+    isEventLive(event) ||
+    isEventStartingSoon(event, EVENT_STARTING_SOON_TIMEFRAME)
+  );
+
+export const isEventSoon = (event: PersonalizedVenueEvent) =>
+  !isEventLive(event) &&
+  isEventStartingSoon(event, EVENT_STARTING_SOON_TIMEFRAME);
```
sunny-viktoryia, line 15: maybe would be better to move them to utils/events too? what do you think?

sunny-viktoryia (commented):
I liked that! :) a few comments and then should be good to go.

src/components/molecules/ScheduleEventListNG/ScheduleEventSubListNG.tsx
yurii-lubynets: fixed in c7e4c1a90c7f5f57663972cbdd08ab5d051cb2d7

src/components/molecules/ScheduleEventListNG/ScheduleEventListNG.tsx
yurii-lubynets: They will be re-calculated every time `daysEvents` are changed. Please clarify what you mean about `setInterval`?

I assume `isEventLive`, `isEventSoon` and `isEventLater` is used `setInterval` under the hood

src/components/molecules/ScheduleEventListNG/utils.ts
yurii-lubynets: fixed in 91ce25851e7b49d5cc1dc28c004ef21d37727e92

src/components/molecules/ScheduleEventListNG/ScheduleEventListNG.tsx
sunny-viktoryia: have you checked that event is moving from "coming soon" to "live" list automatically once it's started? 

My suggestion about setInterval was that we should check every minute/5 minutes what's the status of events to move them to the appropriate group, and if the event is finished, hide it. does that make sense?

yurii-lubynets: Got it. fixed in f3056b41c3cf3139daa66a9d68ca53e48c33f236

src/components/molecules/ScheduleItemNG/ScheduleItemNG.tsx
```diff
@@ -142,7 +142,7 @@ export const ScheduleItemNG: React.FC<ScheduleItemNGProps> = ({ event }) => {
         <div className="ScheduleItemNG__name">{event.name}</div>
         <div className="ScheduleItemNG__place">
           <span className="ScheduleItemNG__place--location">in</span>{" "}
-          {event.room}
+          {event.room || eventVenue?.name}
```
sunny-viktoryia, line 145: thank you! 😍

sunny-viktoryia (approved):
looks good. 🙃 thanks for the ui fixes!

PR #1972, opened 2021-08-10
Add .idea directory to the project

goran-peoski-work:
Draft PR for exploring the possibility of adding `./idea` directory to the project.

 As of now, it is completely ignored through `./gitignore` and future additions of parts or all of that (sub-)tree should be more researched in more detail before making changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,48 @@ cd sparkle
 
 ### Step 2: Install Packages
 
+**Note:** Before you run the next steps, you will need to ensure you have access to the Firebase project you want to use. This access can be set up through the Firebase web UI. The following variables are required to be set up:
+
+```bash
+REACT_APP_PROJECT_ID=
+REACT_APP_API_KEY=
+REACT_APP_APP_ID=
+REACT_APP_MEASUREMENT_ID=
+```
+
 Install the platform dependencies with `npm`:
 
-**Important** `npm` v7+ is not supported, it will cause issues with our `package-lock.json`, and you may end up with the wrong dependency versions. Check your current npm version with `npm -v`.
+(**Important** `npm` v7+ is not supported, it will cause issues with our `package-lock.json`, and you may end up with the wrong dependency versions. Check your current npm version with `npm -v`.
+
+```bash
+npm install
+```
+
+Now you're ready to start the server! ✨
+
+```bash
+npm start
+```
+
+Once the server is started, your web browser will be opened at http://localhost:3000 (and then it'll be immediately redirected to https://sparklespaces.com/). You can opt out of this by adding the following line to your `.env.local`:
+
+```bash
+BROWSER=none
+```
+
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -44,5 +44,32 @@ reporting/*.pdf
...
+.idea/replstate.xml
+.idea_modules/
+cmake-build-*/
+*.iws
+out/
+atlassian-ide-plugin.xml
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
```
dmitrymarchuk, line 75: Modern IntelliJ products include its own .gitignore in the `.idea` folder where they already added all the necessary files to ignore.

```suggestion
```

dmitrymarchuk (approved):
Cleanup `.gitingore`, otherwise LGTM.

.gitignore
dmitrymarchuk: But do we really need to ignore `cmake-build`, `gradle.xml` and so on? 
I mean if IDE for some reason creates them we better know why. If we just ignore them we won't know.
And if we include Java-related ignores why not include Python, PHP, Ruby etc? 

Default `.gitignore` already in the `.idea` folder on my machine produces almost the same file list to commit as in this PR... so it's doing OK.

goran-peoski-work: > And if we include Java-related ignores why not include Python, PHP, Ruby etc?

Idea is created using Java, were it being created using Python, there would be Python references in what the IDE files save in `.idea/`.


> But do we really need to ignore cmake-build, gradle.xml and so on?

The original ignore was a single line 
```
.idea
```
and going from it to include almost anything save for
 ```
/shelf/
/workspace.xml
# Editor-based HTTP Client requests
/httpRequests/
```
is an open invitation for unforeseen consequences (HL2 pun).

So, to the question of
> do we really need to ignore `gradle.xml` at al?

 I can offer only the phrase 
> better safe than sorry

which is perfectly accomplished by the original ignore file.

Do note that `gradle.xml` can pop in at any subdirectory (or several of) under `.idea/` so one can make the assumption IDE or plugins for it can create it during the course of their work.

>  we better know why. If we just ignore them we won't know.

I'd prefer not to know why the hard way by making a mess, then cleaning it up, then adding the files to ignore.
But, as this is a draft PR and there is a branch associated to it, I can welcome you to test and provide a reason why they shouldn't be ignored.

For my part, I can only go so far as create a Gist of the common groups using the settings of "allow duplicate lines" and "allow comments and empty lines":

https://gist.github.com/goran-peoski-work/5249e1982b59573149d01b3fb73952fd

sparkle-ignore-03


On the note of the empirical data and the conclusion from it: 
> Default .gitignore already in the .idea folder on my machine produces almost the same file list to commit as in this PR... so it's doing OK.

Extrapolation of one example, some times colloquially phrased as 
> it works on my machine

better illustrated with the act of scooping a glass of water from the ocean, observing it and concluding there are no sharks in the water 🤷

PR #1977, opened 2021-08-11
Typescript-eslint-recommended rules

dmitrymarchuk:
The goal is to eventually enable all the rules in the following rulesets:
```
"plugin:@typescript-eslint/recommended",
"plugin:@typescript-eslint/recommended-requiring-type-checking"
```

The strategy to achieve this is with small steps. We first enable the ruleset, disable all the rules in it and then enable them one be one (or in small groups) with necessary fixes in separate PRs. This way PRs won't get very big and will stay focused on a couple of rules which will make it easier to spot potential problems.

In this PR I enable and fix errors for:
```
    "@typescript-eslint/no-non-null-assertion": "error",
    "@typescript-eslint/adjacent-overload-signatures": "error",
```


.eslintrc
```diff
@@ -16,15 +16,35 @@
...
+    "@typescript-eslint/no-extra-semi": "off",
+    "@typescript-eslint/no-inferrable-types": "off",
+    "@typescript-eslint/no-misused-new": "off",
+    "@typescript-eslint/no-namespace": "off",
+    "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
+    "@typescript-eslint/no-this-alias": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/prefer-as-const": "off",
+    "@typescript-eslint/prefer-namespace-keyword": "off",
+    "@typescript-eslint/triple-slash-reference": "off",
```
mike-lvov, line 42:
```suggestion
    // @debt uncomment one by one and fix them in separate PRs
    "@typescript-eslint/explicit-module-boundary-types": "off",
    "@typescript-eslint/no-non-null-assertion": "error",
    "@typescript-eslint/adjacent-overload-signatures": "error",
    "@typescript-eslint/ban-ts-comment": "warn",
    "@typescript-eslint/ban-types": "off",
    "@typescript-eslint/no-array-constructor": "off",
    "@typescript-eslint/no-empty-function": "off",
    "@typescript-eslint/no-empty-interface": "off",
    "@typescript-eslint/no-extra-non-null-assertion": "off",
    "@typescript-eslint/no-extra-semi": "off",
    "@typescript-eslint/no-inferrable-types": "off",
    "@typescript-eslint/no-misused-new": "off",
    "@typescript-eslint/no-namespace": "off",
    "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
    "@typescript-eslint/no-this-alias": "off",
    "@typescript-eslint/no-var-requires": "off",
    "@typescript-eslint/prefer-as-const": "off",
    "@typescript-eslint/prefer-namespace-keyword": "off",
    "@typescript-eslint/triple-slash-reference": "off",
```

src/utils/arrayHasElements.ts
```diff
@@ -0,0 +1,3 @@
+export const arrayHasElements = <T>(arr: T[] | undefined): arr is T[] => {
+  return !!arr && arr.length > 0;
+};
```
mike-lvov, line 3: you should reuse the lodash `isEmpty` as we already use it

src/pages/Account/Questions.tsx
```diff
@@ -49,10 +49,6 @@ export const Questions: React.FC = () => {
 
   const { register, handleSubmit, formState } = useForm<QuestionsFormData>({
     mode: "onChange",
-    // @ts-ignore @debt Figure a way to type this properly
-    defaultValues: {
-      ...userWithId,
-    },
```
mike-lvov, line 55: this will probably corrupt the logic?

src/pages/Account/CodeOfConduct.tsx
```diff
@@ -78,18 +78,10 @@ export const CodeOfConduct: React.FC = () => {
...
-    watch,
-  } = useForm<CodeOfConductFormData>({
+  const { register, handleSubmit, errors, formState, watch } = useForm<
+    CodeOfConductFormData & Record<string, boolean>
+  >({
     mode: "onChange",
-    // @ts-ignore @debt Figure a way to type this properly
-    defaultValues: {
-      ...userWithId,
-    },
```
mike-lvov, line 92: can this corrupt the logic?

src/hooks/useMousetrap.ts
```diff
@@ -54,10 +54,12 @@ export const useMousetrap: ReactHook<UseMousetrapProps, void> = ({
...
   withGlobalBind = false,
-
-  // @ts-ignore
-  bindRef,
+  ...rest
 }) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const bindRef = rest?.bindRef;
+
```
mike-lvov, line 62: What's the reason for this change?

src/pages/Account/Questions.tsx
dmitrymarchuk: From the `react-hook-form` docs (for v5)
> Note: Values defined in defaultValues will be injected into watch as defaultValue.

I cannot see usages of any fields from `WithId<User>` across the component. They only use `questions.name` fields.

Also, v7 docs explicitly state:
> defaultValues will be shallowly merged with form submission data.

But we use v5 versions and no fields from `WithId<User>` show up in `data` object if I check like this:
```
    async (data: CodeOfConductFormData) => {
      if (!user) return;

      console.log(data)
```
Which may mean that in v7 they changed the behaviour and now explicitly include that in the docs.

src/pages/Account/CodeOfConduct.tsx
dmitrymarchuk: Explanation here https://github.com/sparkletown/sparkle/pull/1977#discussion_r688321322

src/hooks/useMousetrap.ts
dmitrymarchuk: Forgot about it. Reverted back in https://github.com/sparkletown/sparkle/pull/1977/commits/fdf67dbe5bdeeaf6d08d0c682dace7fdb4e3a761

src/utils/arrayHasElements.ts
dmitrymarchuk: Fixed in https://github.com/sparkletown/sparkle/pull/1977/commits/fdf67dbe5bdeeaf6d08d0c682dace7fdb4e3a761

mike-lvov (approved):
LGTM

PR #1979, opened 2021-08-13
Do not deploy Cloud functions without changes.

dmitrymarchuk:
The idea is to compute hash over source files in `functions` folder and store it in the firebase config.
```
npx firebase functions:config:set checksum.fileshash=$COMPUTED_HASH
```

Hash is computed with the following command. That produces a single md5 hash which covers all the needed files.
```bash
find . -path ./node_modules -prune -false -o -name '*.js' -o -name '*.json' | sort | xargs md5sum | md5sum | awk '{ print $1 }'
```
Then CI can match stored hash over computed one and decide if anything in js or json files has changed and deploy functions only when needed.

I checked that changes in non `js` or `json` files in `functions` don't trigger deploy. And changes in functions code trigger deploy. I also checked that the results are deterministic which means that if a function is changed and reverted back the hash also reverts back as it was before the change.

Test deploys (when I was tinkering and testing stuff) can be checked here: https://app.circleci.com/pipelines/github/sparkletown/sparkle?branch=test-functions-deploy

**UPD:**
I also checked that when a deployment fails this setup will work as expected. That is, it will not update the stored hash and the next time deployment occurs it will recognize that there are pending changes.

As per the circleCI [docs](https://circleci.com/docs/2.0/configuration-reference/#default-shell-options):

> For jobs that run on Linux, the default value of the shell option is /bin/bash **-eo pipefail**

> -e
>>Exit immediately if a pipeline (which may consist of a single simple command), a subshell command enclosed in parentheses, or >>one of the commands executed as part of a command list enclosed by braces exits with a **non-zero status**.

And as we can [see](https://app.circleci.com/pipelines/github/sparkletown/sparkle/9181/workflows/1efffe60-577d-4c39-a428-93a3e94a0e43/jobs/13972) when a circleCI deploy-functions fails the `firebase --deploy` returns 1 and the whole step stops.

> Error: Functions did not deploy properly.
> Exited with code exit status 1
> CircleCI received exit code 1


.circleci/config.yml
```diff
@@ -173,8 +173,23 @@ jobs:
...
+            if [ "$COMPUTED_HASH" == "$STORED_HASH" ]
+            then
+              echo "Functions haven't changed. Nothing to deploy."
+            else
+              echo "Functions changed. Storing new hash..."
+              ./node_modules/.bin/firebase functions:config:set checksum.fileshash=$COMPUTED_HASH --token "$FIREBASE_TOKEN"
+
+              echo "Deploying functions..."
+              # -f, --force: delete Cloud Functions missing from the current working directory without confirmation
+              ./node_modules/.bin/firebase deploy --only functions --token "$FIREBASE_TOKEN" --force
```
mike-lvov, line 191:
```suggestion
              echo "Deploying functions..."
              # -f, --force: delete Cloud Functions missing from the current working directory without confirmation
              ./node_modules/.bin/firebase deploy --only functions --token "$FIREBASE_TOKEN" --force
              
               echo "Functions changed. Storing new hash..."
              ./node_modules/.bin/firebase functions:config:set checksum.fileshash=$COMPUTED_HASH --token "$FIREBASE_TOKEN"
```

mike-lvov (changes requested):
Save new checksum only after the functions have deployed without errors

.circleci/config.yml
dmitrymarchuk: yeah, fixed that in https://github.com/sparkletown/sparkle/pull/1979/commits/4928b6c95a93ce4df007843f75c611315c139992

mike-lvov (approved):
LGTM!

PR #1985, opened 2021-08-16
Redeploy the functions

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1018

I added new twillio config to the staging functions config. Need to redeploy it now

Redeploy the functions


sunny-viktoryia (approved):
🙃

PR #1986, opened 2021-08-16
Synchronise env/burn and env/burn-staging

sunny-viktoryia:
copy all the customizations from env/burn to env/burn-staging


yurii-lubynets (approved):
I assume it's a general update of staging. LGTM

mike-lvov (approved):
LGTM

PR #1989, opened 2021-08-16
Bring parentId field back to admin v1

sunny-viktoryia:
bring the parentId field back to admin v1 form with the default value == "playa" as per @sofisparkle's request.

Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1085

### DIFF
--- a/src/pages/Admin/Venue/DetailsForm.tsx
+++ b/src/pages/Admin/Venue/DetailsForm.tsx
@@ -156,7 +156,6 @@ export const DetailsForm: React.FC<DetailsFormProps> = ({
           await createVenue(
             {
               ...vals,
-              parentId: BM_PARENT_ID,
             } as VenueInput,
             user
           );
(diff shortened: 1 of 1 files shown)


src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -536,6 +535,38 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </>
   );
 
+  const renderParentIdInput = () => (
```
dmitrymarchuk, line 538: I don't see a point of it being inside a function. You can just inline it.

dmitrymarchuk (approved):
LGTM

src/pages/Admin/Venue/DetailsForm.tsx
sunny-viktoryia: I just did it the same way it's done in staging. This PR is just a revert of that part being removed. i guess it was done this way for better visibility. We previously had everything inline here. But the file is super huge so I guess it was the way to make it easier to read.

whoaitsaimz:
Viktoryia said 2hrs ago that this is making its way from burn-staging to burn (production).

PR #1990, opened 2021-08-16
Add event icon to the schedule

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/992

added event icon based on the room / venue it's attached to, same way like it does in the search results.
\+ renamed `PersonalizedVenueEvent` -> `ScheduledVenueEvent`
maybe would be better even: `ScheduledEvent` ? what do you think? 🤔


src/components/molecules/ScheduleItemNG/ScheduleItemNG.scss
```diff
@@ -24,6 +27,19 @@ $schedule-info-block-width: 60px;
     }
   }
 
+  &__icon {
+    flex-shrink: 0;
+    height: $event-icon--size;
+    width: $event-icon--size;
+    margin-left: $event-icon--margin;
+    margin-right: $event-icon--margin;
```
yurii-lubynets, line 35:
```suggestion
    margin: 0 $spacing--xs;
```

yurii-lubynets: could we use `$spacing--xs` to avoid extra var inside the .scss file?

yurii-lubynets (approved):
LGTM!
Nice idea about renaming for `ScheduledVenueEvent`
Some comments left.

src/components/molecules/ScheduleItemNG/ScheduleItemNG.scss
sunny-viktoryia: yeah, nice catch. 🙃 thanks!
refactored here: https://github.com/sparkletown/sparkle/pull/1990/commits/75872581539be7979b443a4afa553d1b6dc7a520

PR #1991, opened 2021-08-17
ADV3 -- add advance setting tab

ermiaSparkle:
Transfer the advance setting to tabs


denisdimitrov (changes requested):
This is not adding the navigation bar, this just moves the advanced settings inside 'manage space' screen. 

You need to wrap the screen with <WithNavigationbar/> to add the navigation bar.

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -2,7 +2,8 @@
 
 .admin-venue {
   width: 100%;
-  height: 100%;
+  padding: 8%;
```
denisdimitrov, line 5: All of our paddings are with pixel values, please keep it consistent.

src/components/organisms/AdminVenues/AdminVenues.scss
```diff
@@ -15,6 +16,24 @@
     font-weight: bolder;
   }
 
+  &__button {
+    height: 45px;
+    background-color: $primary;
+    padding: 1rem;
```
denisdimitrov, line 22: There is `%` `px` and `rem` in a single file for different padding values. Very confusing.

src/components/organisms/AdminTabsBar/AdminTabsBar.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const AdminTabsBar: React.FC<AdminTabsBarProps> = ({
+  activeNav = "",
+  children,
+}) => {
+  const { venueId } = useParams<AdminNavBarRouteParams>();
+
+  const adminTabsUrl = useCallback(
+    (key: string) => {
+      return key === AdminTabsBarTab.advanceSetting
+        ? adminNGSettingsUrl(venueId)
```
denisdimitrov, line 56: It should be using the same route, not another one.
If I can access the spaces like this:
https://co-reality-staging--preview-pr-1991-4osrrjuw.web.app/admin-ng/venue/8mars/spaces
Then I should be able to access advanced settings like this:
https://co-reality-staging--preview-pr-1991-4osrrjuw.web.app/admin-ng/venue/8mars/advanced

In the current implementation it's another route which is incorrect.

Ideally you should show this to Sofi and have a conversation about it. IMO only the "Advanced" Should be moved here and the other 2 tabs should exist in the other route.

denisdimitrov (changes requested):

ermiaSparkle:
Duplicate of #2268

PR #1992, opened 2021-08-17
AV3 -- fix rooms opacity in run tab

ermiaSparkle:
closes issue https://github.com/sparkletown/internal-sparkle-issues/issues/1022


denisdimitrov (changes requested):
The rooms should be transparent only when they are disabled.

https://github.com/sparkletown/internal-sparkle-issues/issues/1022#issuecomment-899341778

ermiaSparkle:
Not any more solution

PR #1995, opened 2021-08-17
AV3: show placeholder if map background is missing

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1024


src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
```diff
@@ -5,6 +5,24 @@
...
 
+  &__background-image {
+    width: 100%;
+  }
+
+  &__background-image-placeholder {
+    width: 100%;
+    height: 1000px;
+    border-radius: $border-radius--lg;
+    box-shadow: 0px 6px 12px rgba(0, 0, 0, 0.25);
```
denisdimitrov, line 16: opaque-black

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.tsx
```diff
@@ -302,14 +304,24 @@ export const VenueRoomsEditor: React.FC<VenueRoomsEditorProps> = ({
...
-        alt="draggable background "
-        style={{
-          width: "100%",
-          ...backgroundImageStyle,
-        }}
-        src={backgroundImage}
-      />
+      {!backgroundImage ? (
+        <div className="Container__background-image-placeholder">
+          <Centered>
```
denisdimitrov, line 309: This will introduce a lot of new and unnecessary divs. It would be easier if you move the styles in the `global.scss` and just pass them in the `className`

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.tsx
```diff
@@ -302,14 +304,24 @@ export const VenueRoomsEditor: React.FC<VenueRoomsEditorProps> = ({
...
+              Pick a background for your map{"\n"}(2000x1200px recommended size)
+            </div>
+          </Centered>
+        </div>
+      ) : (
+        <img
+          alt="draggable background "
+          className={classNames(
+            "Container__background-image",
+            backgroundImageClassName
```
denisdimitrov, line 320: No need for classNames
```tsx
className={`Container__background-image ${backgroundImageClassName}`}
```

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
```diff
@@ -5,6 +5,24 @@
...
+    background: #292c33;
```
denisdimitrov, line 17: Check if it exists in `constants.scss` if it doesn't, add it and use it.

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
```diff
@@ -5,6 +5,24 @@
     position: absolute;
   }
 
+  &__background-image {
+    width: 100%;
+  }
+
+  &__background-image-placeholder {
+    width: 100%;
+    height: 1000px;
```
denisdimitrov, line 14: Sounds like a bad idea to have this hardcoded. Can you use 100% or 100vh instead?

src/hooks/useCheckImage.ts
```diff
@@ -0,0 +1,27 @@
...
+        setError(undefined);
+        setLoading(false);
+      })
+      .catch((err) => {
+        setError(err.message);
+        setLoading(false);
+      });
+  }, [src]);
+
+  return [!error, isLoading, error];
```
denisdimitrov, line 26: It would be better if you change this to be an object, that way the order of imports won't matter, and you can import the first and the last one.

src/hooks/useCheckImage.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from "react";
+
+export const useCheckImage = (
+  src: string
+): [boolean, boolean, string | undefined] => {
+  const [isLoading, setLoading] = useState(false);
+  const [error, setError] = useState<string>();
+
+  useEffect(() => {
```
denisdimitrov, line 9: Can you use `useAsync` here instead? It provides both error and loading out of the box and you won't need the 2 states above.

https://github.com/streamich/react-use/blob/master/docs/useAsync.md

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
```diff
@@ -5,6 +5,24 @@
...
+    border-radius: $border-radius--lg;
+    box-shadow: 0px 6px 12px rgba(0, 0, 0, 0.25);
+    background: #292c33;
+  }
+
+  &__background-image-placeholder_text {
```
denisdimitrov, line 20: You can move these in `__background-image-placeholder` and get rid of this style.

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.tsx
denisdimitrov: The other way around is to pass the parent className `Container__background-image-placeholder` in here as a prop. And it would still be a single div.

denisdimitrov (changes requested):

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/f033a214d20db26a1e8d035a115ccd345272d54f

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.tsx
dmitrymarchuk: I removed the `div` inside the `Centered` component here. 
The div on line 308 here is necessary as it renders the rounded rectangle with a background as in designs.

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/f033a214d20db26a1e8d035a115ccd345272d54f

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/f033a214d20db26a1e8d035a115ccd345272d54f

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
dmitrymarchuk: Initial intention was to set to 1200px (i mistakenly used 1000) as in text: *Recommended size: 2000px / 1200px*. But yeah, `vh` is better solution. Fix in https://github.com/sparkletown/sparkle/commit/f033a214d20db26a1e8d035a115ccd345272d54f

src/hooks/useCheckImage.ts
dmitrymarchuk: Generally speaking, I'd argue that using tuples as return type from hooks is more _convenient_ as 

- you don't have to rename returned fields every time the hook is used which results in more concise usage
- the practice with tuples is widely used in react and other hooks libraries

But here we are returning 3 values and two of them are of similar type `boolean`. So I changed it to object.
Commit https://github.com/sparkletown/sparkle/commit/f033a214d20db26a1e8d035a115ccd345272d54f

src/components/organisms/AdminVenueView/components/VenueRoomsEditor/VenueRoomsEditor.scss
dmitrymarchuk: Fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/f033a214d20db26a1e8d035a115ccd345272d54f

src/hooks/useCheckImage.ts
dmitrymarchuk: Great suggestion! Fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/c2dfda2754be5b6a58f54916649a3aa6da4951be

src/hooks/useCheckImage.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useAsyncFn } from "react-use";
+
+export const useCheckImage = (
+  src: string
+): {
+  isValid: boolean;
+  isLoading: boolean;
+  error: string | undefined;
+} => {
+  const [{ loading, error }, checkImage] = useAsyncFn(async () => {
```
denisdimitrov, line 11: If you change `useAsyncFn` to `useAsync` you can get rid of the `useEffect` and and logic will run the same way.

src/components/atoms/Centered/Centered.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import classNames from "classnames";
+
+import { ContainerClassName } from "types/utility";
+
+import "./Centered.scss";
+
+export const Centered: React.FC<
```
denisdimitrov, line 8: The name sounds like a state/boolean, can we think of something else? Maybe `CenterContent`, `CenterChildren` or something else.

denisdimitrov (changes requested):

src/hooks/useCheckImage.ts
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/1995/commits/5d9989860f08448b0260105f0c054f519b00bd03

src/components/atoms/Centered/Centered.tsx
dmitrymarchuk: Haven't managed to come up with other variants. Chose `CenterContent`:) Commit https://github.com/sparkletown/sparkle/pull/1995/commits/5d9989860f08448b0260105f0c054f519b00bd03

denisdimitrov (approved):
LGTM!

PR #1996, opened 2021-08-17
added descendent venue events to sovereign venue event list

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/964


src/types/venues.ts
```diff
@@ -11,7 +11,6 @@ import { Poster } from "./posters";
 import { Quotation } from "./Quotation";
 import { Room } from "./rooms";
 import { Table } from "./Table";
-import { UpcomingEvent } from "./UpcomingEvent";
```
sunny-viktoryia, line 14: I believe everything that is related to the `UpcomingEvent` type is legacy. Can you remove that?

src/types/venues.ts
```diff
@@ -141,7 +140,7 @@ export interface BaseVenue {
   code_of_conduct_questions: Question[];
   owners: string[];
   iframeUrl?: string;
-  events?: Array<UpcomingEvent>; //@debt typing is this optional? I have a feeling this no longer exists @chris confirm
+  events?: WithId<VenueEvent>[]; //@debt typing is this optional? I have a feeling this no longer exists @chris confirm
```
sunny-viktoryia, line 143: events are not stored in venues anymore. So, please, remove that logic.

src/types/venues.ts
```diff
@@ -236,6 +235,7 @@ export interface PartyMapVenue extends BaseVenue {
   admin_password?: string;
   owners: string[];
   rooms?: Room[];
+  events: WithId<VenueEvent>[];
```
sunny-viktoryia, line 238: 1.  why it was added to PartyMap when BaseVenue has already events?
2. I believe we should better handle events separately and not mutate the object by adding events there, as we don't store the information about events in the firestore-venue-object.

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -16,13 +18,27 @@ import VenueLandingPageContent from "./VenueLandingPageContent";
...
+  const venueEvents = useSelector(venueEventsSelector);
+  const allVenueEvents = useMemo(
+    () => [
+      ...(venueEvents || []),
+      ...(sovereignVenueId === venueId
+        ? [...descendantVenues.flatMap((venue) => venue.events || [])]
+        : []),
+    ],
+    [descendantVenues, venueEvents, sovereignVenueId, venueId]
+  );
```
sunny-viktoryia, line 40: I believe you should use `useVenueEvents({
    venueIds: [currentVenueId, ...descedantVenueIds]
  })` here instead.

src/pages/Admin/EventsComponent.tsx
```diff
@@ -27,30 +28,43 @@ const EventsComponent: React.FC<EventsComponentProps> = ({
...
+              ((ev?.start_utc_seconds || 1) +
+                (ev?.duration_minutes || 2) * 60) *
+                1000 >
               Date.now()
           )
-        : events,
-    [events, filterPastEvents]
+        : allVenueEvents,
+    [filterPastEvents, allVenueEvents]
   );
```
sunny-viktoryia, line 68: i believe you can use here something from `utils/event`. For example, will `isEventLiveOrFuture` fit here?

src/hooks/useSovereignVenue.ts
```diff
@@ -44,9 +44,15 @@ export const useSovereignVenue: ReactHook<
...
-    if (!venueId || sovereignVenue || isSovereignVenueLoading || errorMsg)
+    if (
+      !venueId ||
+      stateVenueId === venueId ||
+      isSovereignVenueLoading ||
+      errorMsg
+    )
       return;
 
+    updateVenueId(venueId);
```
sunny-viktoryia, line 55: can you explain the logic in this file, please? why is it needed?

src/api/venue.ts
```diff
@@ -110,12 +112,32 @@ export const fetchDirectChildVenues = async (
...
+              return {
+                ...venue,
+                ...(rest.length
+                  ? { events: rest as WithId<VenueEvent>[] }
+                  : null),
+              };
+            })
+        );
+
+        return venueWithEvents;
```
sunny-viktoryia, line 140: looks complicated... are you extending the venue with its events here? 
not sure we should do that. Can we handle that separately?

sunny-viktoryia (commented):
partial review. Please, fix them and then will do another round. 
\+ i can't find the event I used to have in staging.
image
image

src/types/venues.ts
alexey-petrov: It's being used in `UpcomingTickets` component. Can we confirm that it's not necessary anymore?

sunny-viktoryia: 99% - `UpcomingTickets` are legacy. You can remove them. Confirming.
(clarified with @mike-lvov )

PR #2000, opened 2021-08-18
Remove unnecessary validation from DetailsForm

dmitrymarchuk:
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1027
Additional fix after PR https://github.com/sparkletown/sparkle/pull/1988


src/pages/Admin/Details/ValidationSchema.ts
```diff
@@ -3,8 +3,6 @@ import firebase from "firebase/app";
 import * as Yup from "yup";
 
 import {
-  GIF_RESIZER_URL,
```
denisdimitrov, line 6: Check if this is being used anywhere else, if it's not, you can remove it from settings.ts

denisdimitrov (changes requested):

src/pages/Admin/Details/ValidationSchema.ts
dmitrymarchuk: It's used in the old admin `DetailsForm` component's validation schema.

denisdimitrov (approved):
LGTM!

PR #2002, opened 2021-08-18
Properly handle invalid map background

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1049

I also used a new file `useCheckImage.ts` from my other PR https://github.com/sparkletown/sparkle/pull/1995 which is not merged yet. I'll handle the conflicts if any arise.


src/hooks/useMapBackground.ts
```diff
@@ -0,0 +1,16 @@
+import { DEFAULT_MAP_BACKGROUND } from "settings";
+
+import { useCheckImage } from "hooks/useCheckImage";
+
+export const useMapBackground = (
+  mapBackgroundUrl: string | undefined
```
denisdimitrov, line 6:
```suggestion
  mapBackgroundUrl?: string
```

src/hooks/useCheckImage.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useAsyncFn } from "react-use";
+
+export const useCheckImage = (
+  src: string
+): {
+  isValid: boolean;
+  isLoading: boolean;
+  error: string | undefined;
+} => {
+  const [{ loading, error }, checkImage] = useAsyncFn(async () => {
```
denisdimitrov, line 11: You can use ` useAsync` and get rid of the `useEffect` just like your other PR.

src/hooks/useMapBackground.ts
```diff
@@ -0,0 +1,16 @@
...
+): [string, boolean] => {
+  const { isValid } = useCheckImage(mapBackgroundUrl ?? "");
+
+  return [
+    isValid
+      ? mapBackgroundUrl ?? DEFAULT_MAP_BACKGROUND
+      : DEFAULT_MAP_BACKGROUND,
```
denisdimitrov, line 13: Please create a constant for this conditional for better readability.

denisdimitrov (changes requested):

src/hooks/useMapBackground.ts
dmitrymarchuk: I'd argue that is not the same: if a prop (or parameter of a function) is optional it means you may not pass it at all. If it's `type | undefined` it means that you have to pass it but it may be undefined. 
The function needs `mapBackgroundUrl` prop, but it may be `undefined`. That's not the same as invoking this function without that parameter at all.

src/hooks/useCheckImage.ts
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2002/commits/4f7bca683dd8443d753e222cf319790bf2a047ef

src/hooks/useMapBackground.ts
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2002/commits/4f7bca683dd8443d753e222cf319790bf2a047ef

src/hooks/useMapBackground.ts
denisdimitrov: Fair enough.

denisdimitrov (approved):
LGTM!

PR #2005, opened 2021-08-18
BM general loading view

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1011


src/components/molecules/LoadingPage/LoadingPage.scss
```diff
@@ -0,0 +1,61 @@
...
+    justify-content: center;
+    align-items: center;
+  }
+
+  &__logo-container {
+    position: relative;
+  }
+
+  &__playa {
+  }
```
denisdimitrov, line 28: You can remove this and it's usages.

src/components/molecules/LoadingPage/svg/diamond.svg
```diff
@@ -0,0 +1,20 @@
+<svg width="80" height="133" viewBox="0 0 80 133" fill="none" xmlns="http://www.w3.org/2000/svg">
```
denisdimitrov, line 1: All of the svg files should be moved to `assets`

src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -72,6 +72,7 @@ export const AppRouter: React.FC = () => {
     <Router basename="/">
       <Suspense fallback={<LoadingPage />}>
         <Switch>
+          <Route path="/loading" component={LoadingPage} />
```
denisdimitrov, line 75: I suppose this was added for testing/developing purposes, please remove it.

src/components/molecules/LoadingPage/LoadingPage.tsx
```diff
@@ -27,30 +36,35 @@ const quotes = IS_BURN
...
-    () => {
+    useCallback(() => {
       setQuote(quotes[Math.floor(Math.random() * quotes.length)]);
-    },
+    }, []),
     IS_BURN ? 1000 : undefined
   );
 
+  const { isShown: animate, show: startAnimation } = useShowHide();
+  //this prevents "flickering" of the animation on page load
```
denisdimitrov, line 50: This sounds like a hacky solution. Can we think of something else?

src/components/molecules/LoadingPage/LoadingPage.tsx
```diff
@@ -27,30 +36,35 @@ const quotes = IS_BURN
...
-          <span className="loading-sparkle-2" />
-          <div className="burningman-loading">
-            <div className="burningman-loading-anim" />
-          </div>
+    <div className="LoadingPage">
+      <div className="LoadingPage__wrapper">
+        <div className="LoadingPage__logo-container">
+          <DiamondSvg
+            className={classNames("LoadingPage__diamond", {
+              "LoadingPage__diamond--animation": animate,
```
denisdimitrov, line 59: The animation should be always executed, no? You can remove the conditional state and always play it.

denisdimitrov (changes requested):

src/components/molecules/LoadingPage/LoadingPage.tsx
dmitrymarchuk: So the problem is that on initial page load (for example when accessing this new route `/loading`) the LoadingPage component not just re-renders multiple times but it 
- renders
- then in some time is removed from the render tree, unrenders
- then it's rendered again. 

Which means that animation is started, then when the component is removed and quickly brought back (it is just visible in the debug/logs but not to the naked eye) animation is restarted which results in some sort of _flashing_ of the diamond logo. Then it proceeds as normal.

I thought that this initial time is important for the overall user experience of this loading page but didn't found a better solution to just delay the start of the animation.

src/components/molecules/LoadingPage/LoadingPage.tsx
dmitrymarchuk: explanation here https://github.com/sparkletown/sparkle/pull/2005#discussion_r692679172

src/components/molecules/LoadingPage/svg/diamond.svg
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2005/commits/a172c44fbf91deb2b63b290bcdc476c820b80fc1

src/components/molecules/LoadingPage/LoadingPage.scss
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2005/commits/a172c44fbf91deb2b63b290bcdc476c820b80fc1 and https://github.com/sparkletown/sparkle/pull/2005/commits/813776a5c5144414214388b313bb19a368a87dfc

src/components/organisms/AppRouter/AppRouter.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2005/commits/0f5f808e5b41a355da9aa72c5fc40f4fa7320562

src/components/molecules/LoadingPage/LoadingPage.tsx
denisdimitrov: Why is this not an issue with our current implementation? There should be a way to fix this. 300 delay is a lot and it just hides/skips the problem.

dmitrymarchuk: It's actually an issue:) The current animation flickers too but I think you just don't notice it. On my machine, I barely even notice the loading page at all.
flickering

dmitrymarchuk: The other solution is to make the `LoadingPage` render all the time. We can rename it to a `LoadingScreen` or similar, move to the top of the component tree, create a context, wrap the whole component tree in it and then make a hook e.g. `useLoading` which will accept a boolean to hide or show this loading screen. Then it should work.

denisdimitrov: I think you should show it so Sofi and let her decide, IMO we should leave it as it is and let it flicker.

dmitrymarchuk: Okay, I removed the delay in https://github.com/sparkletown/sparkle/pull/2005/commits/44c45c4b89f7f394f70398fbf980eb9330d106ec so it flickers now, but as the default animation in staging does.

I'll contact @sofisparkle and consult whether this flicker needs improvements.

denisdimitrov (approved):
LGTM!

PR #2009, opened 2021-08-19
BM: make description/subtitle not required

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1068

PR #2016, opened 2021-08-20
Make description/subtitle not required

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1068


functions/venue.js
```diff
@@ -593,11 +597,7 @@ exports.updateVenue = functions.https.onCall(async (data, context) => {
     throw new HttpsError("not-found", `Venue ${venueId} not found`);
   }
 
-  // @debt this is exactly the same as in updateVenue_v2
-  const updated = doc.data();
-
-  // @debt refactor function so it doesn't mutate the passed in updated object, but efficiently returns an updated one instead
-  createBaseUpdateVenueData(data, updated);
+  const updated = createBaseUpdateVenueData(data, doc);
```
sunny-viktoryia, line 600: not sure if that should've been changed.

`data` object is mutated as the `@debt` comment says, and that part should be refactored anyway. and if it didn't return a value, it'd be easier to read that I believe.
what do you think?

sunny-viktoryia (approved):
approving. 

but, please, check my comment. Maybe you'll find it reasonable.

side note: it's the duplicate of https://github.com/sparkletown/sparkle/pull/2009

functions/venue.js
dmitrymarchuk: > so it doesn't mutate the passed in **updated** object

I think he meant the variable with the name `updated`. The code inside `createBaseUpdateVenueData` does not mutate the `data` object.

sunny-viktoryia: ah, i see. Looks like you're right. 
okay, let's have it as you did then.

PR #2020, opened 2021-08-20
[env/burn-staging] Update chat styles

goran-peoski-work:
Fixes
- https://github.com/sparkletown/internal-sparkle-issues/issues/1048

**Before**:

sparkle-chat-style-02

**After**:

sparkle-chat-style-01


src/components/atoms/ButtonNG/ButtonNG.tsx
```diff
@@ -51,6 +51,7 @@ export const ButtonNG: React.FC<ButtonProps> = ({
   iconOnly = false,
   iconName,
   iconSize = "1x",
+  ...extraProps
```
goran-peoski-work, line 54: `extraProps` in `ButtonNG` facilitate the use of `aria-*` as well as other props normally used with buttons

src/components/molecules/ChatMessageBox/ChatMessageBox.scss
```diff
@@ -1,5 +1,7 @@
 @import "scss/constants";
 
+$emoji-button-diameter: $submit-button-diameter - $spacing--md;
```
goran-peoski-work, line 3: Due to expediency of PR a more holistic solution wasn't pursued i.e. actually making the emoji button part of the input component, possibly creating a new input component. It is OOS to this PR, but an idea for further consideration.

src/components/molecules/ChatMessageBox/ChatMessageBox.scss
```diff
@@ -11,27 +13,50 @@
...
+    flex-grow: 1;
+    margin-right: $spacing--lg;
+
+    .InputField {
+      flex-grow: 1;
+
+      input {
+        // NOTE: this removes screen artefacts of the embedded <input> inside InputField being as wide as the parent
+        margin: 1px;
+      }
```
goran-peoski-work, line 30: This could be revisited on the `InputField` component itself as artefacts due to internal `input` element are visible elsewhere.

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
```diff
@@ -100,7 +100,8 @@ export const ChatMessageBox: React.FC<ChatMessageBoxProps> = ({
 
   const placeholderValue = isQuestion ? "question" : "message";
 
-  const buttonClasses = classNames("Chatbox__submit-button", {
+  const buttonClasses = classNames({
+    "Chatbox__submit-button": true,
     "Chatbox__submit-button--question": isQuestion,
   });
```
goran-peoski-work, line 106: Simpler diff/read when only object argument is used

goran-peoski-work (commented):
Initial PR notes

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
denisdimitrov: I saw in some of your other pull requests, you using object with only `true` values instead of inline syntax `className=""` That's why we were using the old syntax when you are passing them in the first param without boolean conditions it makes it visible and obvious that it can be inline. Ex:

When you write
```tsx
const someClasses = classNames("Class_el1 Class_el2")
```
It makes it obvious that this is just `className="Class_el1 Class_el2"`

The other way around you are not noticing it and making it look more complex that it actually is. Ex:

```tsx
const someClasses = classNames({
  "Class_el1": true,
  "Class_el2": true
})
```

goran-peoski-work: Looking more complex by having one object with two lines than instead of having two parameters one of type string and other of type object?

Albeit "looking more complex" might be interpreted as subjective, my rationale in "other pull requests" is it makes it easier to diff.

denisdimitrov: In my example it's "just two lines" but this can easily grow into 5 or 6. I agree about the diff though. I'm not strongly against it but it it feels and looks weird to use a library that it's only purpose in the project is to be used for conditional class names and we are using it for reviewing benefits by passing non conditional classNames in there.

denisdimitrov (approved):
LGTM!

src/components/molecules/ChatMessageBox/ChatMessageBox.tsx
goran-peoski-work: > I'm not strongly against it but it it feels and looks weird to use a library that it's only purpose in the project is to be used for conditional class names and we are using it for reviewing benefits by passing non conditional classNames in there.
Got no disagreement with this ☝️, here are few points to note though:

1. As with previous reviews, if specifically asked to change the style, I will most likely do it
2. If specifically commented on code style choice, I may provide arguments for or against, and those arguments will not be an obstacle of following point 1.
3. On the library... Oh well, it exists specifically because of React limiting `className` to a string. So if we note a definitive guide in the future on how to generate it, it will be good. Until then, few different choices here and there may give us the insight on what to pick for the guide.

whoaitsaimz:
Just made comment in the original issue, the emoji button is now a rounded rect rather than a circle, but functionally works and looks ok so I closed the issue. Just something to note if we look back on this.

PR #2025, opened 2021-08-20
Default venue banner and logo

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/682


src/hooks/useCheckImage.ts
```diff
@@ -1,22 +1,41 @@
...
   }, [src]);
 
-  return {
-    isValid: !error,
-    isLoading: loading,
-    error: error?.message,
-  };
+  return useMemo(
+    () => ({
+      //here we consider the image valid event while it still loading
```
sunny-viktoryia, line 21: typo?
```suggestion
      // here we consider the image valid even while it's still loading
```

src/hooks/useCheckImage.ts
```diff
@@ -1,22 +1,41 @@
+import { useMemo } from "react";
 import { useAsync } from "react-use";
 
 export const useCheckImage = (
-  src: string
+  src: string | undefined
```
sunny-viktoryia, line 5: maybe better would be this way? or would you prefer it to be more explicit?
```suggestion
  src?: string
```

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -69,11 +70,14 @@ export const Map: React.FC<MapProps> = ({
...
-  const [mapBackgroundUrl] = useMapBackground(venue?.mapBackgroundImageUrl);
+  const [mapBackground] = useCheckImageWithFallback(
+    venue?.mapBackgroundImageUrl,
+    DEFAULT_MAP_BACKGROUND
+  );
 
   useEffect(() => {
     const img = new Image();
-    img.src = mapBackgroundUrl;
+    img.src = mapBackground ?? DEFAULT_MAP_BACKGROUND;
```
sunny-viktoryia, line 80: why do we need a fallback here too? isn't `mapBackground` with fallback?

sunny-viktoryia (commented):
Looks great! Super close to being merged.
A couple of nit comments and a question here.

src/hooks/useCheckImage.ts
dmitrymarchuk: This a not a completely equal substitution. With `src?: string` one can not pass an argument to `useCheckImage` at all while with `string | undefined` the argument must be passed but can be undefined. My intention is to indicate that `useCheckImage` needs an argument but it can be of unspecified value.

src/components/templates/PartyMap/components/Map/Map.tsx
dmitrymarchuk: `useCheckImageWithFallback` guards from invalid image. But it still needs to return something while the image is loading. 

For example, in the `Map.tsx` component the map background loads for a split second. While it's loading nothing is displayed. If I change the return value of `useCheckImageWithFallback` to be `string` rather than `string | undefined` I must return something while the image is loading. This introduces _flickering_ where for this split second `DEFAULT_MAP_BACKGROUND` is shown. That doesn't look nice IMO so I tried to preserve the original behavior.

Also, to improve readability I renamed the hook to be `useValidImage` and the param to `fallbackIfInvalid`.
And finally, the type was set correctly `string | undefined` but I was still returning `src ?? fallbackSrc` which resulted in flickering (forgot to fix that).
Renamed and fixed here https://github.com/sparkletown/sparkle/pull/2025/commits/5f89f4dfa98705fd30dad433d7627dd14c1c9057

src/hooks/useCheckImage.ts
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2025/commits/5f89f4dfa98705fd30dad433d7627dd14c1c9057

sunny-viktoryia (approved):
Amazing! Thank you for the refactoring, @dmitrymarchuk! 😍

PR #2026, opened 2021-08-20
Fixed UserReactions performance issue.

daniel-brown-sparkle:
When rendered in large numbers, specifically inside the Audience component, UserReactions has significant CPU cost.
This was caused because a custom CSS variable was created for *every seated user* using `useCss` which needs to be injected into the document.

image

This change fixes the problem by (1) preventing the rendering of unnecessary DOM elements, and (2) removing the custom CSS variable.

When testing on Macbook 2.6 GHz 6-Core Intel Core i7, this change drops about ~100ms from each Audience render done by `processQueuedChangesIntervalId`

See https://github.com/sparkletown/internal-sparkle-issues/issues/1082 for more context


src/components/molecules/UserReactions/UserReactions.scss
```diff
@@ -2,7 +2,7 @@
 
 $reactionOffset: calc(-1 * #{$reactionSize} / 2);
 $shoutBounceTranslateXOffset: 50%;
-$reactionTimeout: var(--user-reactions-reaction-timeout, 5s);
```
daniel-brown-sparkle, line 5: The value was always to 5s when the variable was used, which made this meaningless.

src/components/molecules/UserReactions/UserReactions.tsx
```diff
@@ -43,47 +36,41 @@ export const UserReactions: React.FC<UserReactionsProps> = ({
...
-      uniqueEmojiReactionsDataMapReducer,
-      new Map()
+  const { userUniqueEmojiReactions, userShoutout } = useMemo(() => {
+    const userUniqueEmojiReactions = Array.from(
+      userReactions
+        .reduce(uniqueEmojiReactionsDataMapReducer, new Map())
+        .values()
     );
 
-    const renderedEmojiReactions = Array.from(
```
daniel-brown-sparkle, line 52: I wanted to keep the `useMemo` clean and focused on the calculation (I doubt it's worth having it here, but still). The JSX is moved onto the main section, making this component more readable

src/components/molecules/UserReactions/UserReactions.tsx
```diff
@@ -43,47 +36,41 @@ export const UserReactions: React.FC<UserReactionsProps> = ({
...
+    <>
+      {hasReactions && (
+        <div className={containerClasses}>
+          {userUniqueEmojiReactions.map((emojiReaction) => (
+            <DisplayEmojiReaction
+              key={emojiReaction.type}
+              emojiReaction={emojiReaction}
+              isMuted={isMuted}
+            />
+          ))}
```
mike-lvov, line 67: useMemo

mike-lvov (approved):
LGTM, once fixed. Good catch. Thanks for the research & fixes

src/components/molecules/UserReactions/UserReactions.tsx
daniel-brown-sparkle: https://github.com/sparkletown/sparkle/pull/2026/commits/354d57d9db14af7421b1bfbefe400852131488c9

PR #2027, opened 2021-08-20
Beavers/firebarrels

YarRainbow:
last code with little style changes for firebarrels + some changes for user's avatar animation


mike-lvov (approved):
LGTM

PR #2037, opened 2021-08-23
Sort venues by creation date in Admin V1

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1057

This PR:
* extends BaseVenue type by adding `createdAt: number` and `updatedAt: number` fields (data is stored in UTC ms)
* saves the information of the timing when a venue was created and updated
* adds sorting options to admin v1 venue list by `A-Z`, `Z-A`, `newly created first`, `oldest first`
* shows venue creation and venue update time in admin v1

### DIFF
--- a/src/pages/Admin/Admin.tsx
+++ b/src/pages/Admin/Admin.tsx
@@ -14,6 +14,9 @@ import {
   useLocation,
   useRouteMatch,
 } from "react-router-dom";
+import { faEllipsisV } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import classNames from "classnames";
 import dayjs from "dayjs";
 import advancedFormat from "dayjs/plugin/advancedFormat";
 
@@ -41,6 +44,8 @@ import {
   canHaveEvents,
   canHavePlacement,
   canHaveSubvenues,
+  sortVenues,
+  VenueSortingOptions,
 } from "utils/venue";
 
 import { useIsAdminUser } from "hooks/roles";
@@ -83,18 +88,58 @@ const VenueList: React.FC<VenueListProps> = ({
     currentVenueId: selectedVenueId,
   });
 
+  const {
+    isShown: showSortingDropdown,
+    toggle: toggleSortingDropdown,
+  } = useShowHide();
+
+  const [currentSortingOption, setCurrentSortingOption] = useState(
+    VenueSortingOptions.az
+  );
+
+  const sortedVenues = useMemo(() => {
+    return sortVenues(ownedVenues, currentSortingOption);
+  }, [currentSortingOption, ownedVenues]);
+
+  const sortingOptions = useMemo(
+    () =>
+      Object.values(VenueSortingOptions).map((sortingOption) => (
+        <li
+          key={sortingOption}
+          className={classNames("page-container-adminsidebar__sorting-option", {
+            "page-container-adminsidebar__sorting-option--active":
+              currentSortingOption === sortingOption,
+          })}
+          onClick={() => {
+            setCurrentSortingOption(sortingOption);
+            toggleSortingDropdown();
+          }}
+        >
+          {sortingOption}
+        </li>
+      )),
+    [currentSortingOption, toggleSortingDropdown]
+  );
+
   if (isLoading) return <Loading />;
 
   return (
     <>
-      <div className="page-container-adminsidebar-title title">My Venues</div>
+      <div className="page-container-adminsidebar-title title">
+        My Venues
+        <FontAwesomeIcon
+          className="page-container-adminsidebar-title__ellipsis"
+          icon={faEllipsisV}
+          onClick={toggleSortingDropdown}
+        />
+      </div>
       <div className="page-container-adminsidebar-top">
         <Link to="/admin/venue/creation" className="btn btn-primary">
           Create a venue
         </Link>
       </div>
       <ul className="page-container-adminsidebar-venueslist">
-        {ownedVenues.map((venue) => (
+        {sortedVenues.map((venue) => (
           <li
             key={venue.id}
             className={`${selectedVenueId === venue.id ? "selected" : ""} ${
@@ -119,6 +164,14 @@ const VenueList: React.FC<VenueListProps> = ({
           </li>
         ))}
       </ul>
+
+      {showSortingDropdown && (
+        <div className="page-container-adminsidebar__sorting-dropdown">
+          <ul className="page-container-adminsidebar__sorting-list">
+            {sortingOptions}
+          </ul>
+        </div>
+      )}
     </>
   );
 };
--- a/src/pages/Admin/Admin.tsx
+++ b/src/pages/Admin/Admin.tsx
@@ -98,7 +98,7 @@ const VenueList: React.FC<VenueListProps> = ({
   );
 
   const sortedVenues = useMemo(() => {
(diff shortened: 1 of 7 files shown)


src/pages/Admin/AdminVenuePreview.tsx
```diff
@@ -133,6 +134,15 @@ export const AdminVenuePreview: React.FC<AdminVenuePreviewProps> = ({
...
           </div>
+
+          <div>
+            <span className="title">Created At:</span>
+            {venue.createdAt && format(venue.createdAt, "yyyy-MM-dd HH:mm:ss")}
+          </div>
+          <div>
+            <span className="title">Updated At:</span>
+            {venue.updatedAt && format(venue.updatedAt, "yyyy-MM-dd HH:mm:ss")}
+          </div>
```
dmitrymarchuk, line 145: Maybe it's better not to render the text "Created At:" and "Updated At:" too if there is no `venue.createdAt` and `venue.updatedAt`.

src/pages/Admin/Admin.tsx
```diff
@@ -83,18 +88,58 @@ const VenueList: React.FC<VenueListProps> = ({
...
+            setCurrentSortingOption(sortingOption);
+            toggleSortingDropdown();
+          }}
+        >
+          {sortingOption}
+        </li>
+      )),
+    [currentSortingOption, toggleSortingDropdown]
+  );
+
```
dmitrymarchuk, line 123: nitpick: if this can be extracted into a hook it would be awesome.

dmitrymarchuk (changes requested):

src/utils/venue.ts
```diff
@@ -140,3 +140,34 @@ export const withVenue = <T extends object>(
...
+        (a, b) =>
+          (a.createdAt ?? 0) - (b.createdAt ?? 0) || a.id.localeCompare(b.id)
+      );
+    case VenueSortingOptions.newestFirst:
+      return [...venueList].sort(
+        (a, b) =>
+          (b.createdAt ?? 0) - (a.createdAt ?? 0) || a.id.localeCompare(b.id)
+      );
+    default:
+      return venueList;
```
dmitrymarchuk, line 171: Maybe it's better to `console.error` or `throw Error` to indicate an invalid situation.

sunny-viktoryia: i don't think we should do that. I guess we just can extend the VenueSortingOptions enum but forget to update the function or add a "default"/"no sorting" option for example which does nothing with the list.

src/pages/Admin/AdminVenuePreview.tsx
sunny-viktoryia: hmm, i did the same as it's displayed for venues that do not have a description/tagline. The titles are still there in those cases even if the values are empty. So we should change the behavior for both cases or leave it as is. 

what do you think is better? i'd prefer to be more explicit and to show the titles.

image

dmitrymarchuk: agreed

src/utils/venue.ts
dmitrymarchuk: It's just a safe approach: if a new sorting option is added the function will stop working _silently_. Some sort of indication of an invalid behavior is needed IMO. Maybe the best option is to use can use an approach with `never` described [here](https://dev.to/ddiprose/exhaustive-switch-statement-with-typescript-26dh). A function like this is created:
```typescript
function assertUnreachable(x: never): never {
  throw new Error("Didn't expect to get here");
}
```
and you use that function in the `default` case. Then TS will not compile the code if the switch over the enum is not exhaustive. You can check the example behaviour of this on this [TS playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAsgngUXNA3gKClAwgQwM7ACMUAvFAES4GHkA0G2+wATKRVS3QxwMxuVMe5NAF80aAMYB7EHgAuUAJYgADmDkAuWImRt4SSADoOxfFBwg4AbnEAzMCAlzFM83gIAnOQFUQH4DgSABY4AEYANsAAFAAeWiDAAG7AHgCU8UkpUOiYckEeUgDuUAnFCB4FHlHkACKKACYgAOQKwDEqwE5QclJQAObACkEpwOSpNmJoeIWKcsFRympyqdkMEkzaBhDGTIQaDJjSslKRhuFSfVFNKgUSwO5Q6wRQhACETeMHUKH+OADWNkOG30yB2BGY+0whxkeBOwDOFyuNykdweT2AUGY70+UO+vwBa2BOiMvEhUKOsNO50u11u9zwjw2WI+gKhPwCBMw9WAthwYHCmi++E8Pj8AWCYUiC1U6k+IiAA): if you remove the `case MyEnum.Case3:` from switch TS shows an error.

sunny-viktoryia: good point. 🙃 thank you!
i've added the function to a separate utils file so that it could be used in other places too. check here: https://github.com/sparkletown/sparkle/pull/2037/commits/44ddb8a8f1bb94c7cec345783953888672355814

src/pages/Admin/Admin.tsx
sunny-viktoryia: the only solution I see is to create a separate component for search dropdown passing the list of sorting options there so that it could be used everywhere. But I don't have any designs for that feature, in fact, I've created everything myself without the requirements. So probably, in the future when we have a proper design would be nice to refactor that part and use a single "sorting-dropdown" component if needed.

src/utils/venue.ts
dmitrymarchuk: I dug into it a little bit more and found out that this approach works because `assertUnreachable` must have a parameter of type `never` but not the return value. Return value was specified in the article for the sake of completeness I guess.
You can check that it doesn't work at [this](https://www.typescriptlang.org/play?#code/KYOwrgtgBAsgngUXNA3gKClAwgQwM7ACMUAvFAES4GHkA0G2+wATKRVS3QxwMxuVMe5NAF80aAMYB7EHgAuUAJYgADmDkAuWImRt4SSADoOxfFBwg4AbnEAzMCAlzFM83gIAnOQFUQH4DgSABY4AEYANsAAFACUWiDAAG7AHlDomHJBHlIA7lAJeQge2R5R5AAiigAmIADkCsAAHirATlByUlAA5sAKQSnA5DE2Ymh4OYpywVHKanIxaQwSTNoGEMZMhBoMmNKyUpGG4VJdUbUq2RLA7lDLBFCEAIS1wztQof44ANY2uyv6yA2BGY20wuxkeAOwCOJzOFykVxud2AUGYz1eYPenx+DCqwFsODA4U0b3wnh8fgCwTCkViIyAA) TS playground: I removed the `MyEnum.Case3` from the switch and it still compiles.

So the way it works is like that: 
- when you exhaust all the values of an enum in a switch the compiler knows that flow will never reach a `default` clause. 
- because of that it infers the type of variable under switch statement to be `never`
- which matches the type of a parameter of the `assertUnreachable` function

[Docs](https://www.typescriptlang.org/docs/handbook/basic-types.html#never) on `never` type (I definitely learned something new🙂).

sunny-viktoryia: 🙃 haha me too! Thanks for the deep digging.
fixed here: https://github.com/sparkletown/sparkle/pull/2037/commits/b23ae046845056dd17f7c8f01318f5003aae4cfd

dmitrymarchuk (approved):
Approving

PR #2039, opened 2021-08-23
Disable deployment for some envs

mike-lvov:
As per the discussion with the product owners, we want to speed up the deployment process for Burn env


sunny-viktoryia (approved):
approving. 🙃

PR #2050, opened 2021-08-23
[env/burn-staging] Improved nav bar user search performance

daniel-brown-sparkle:
The search bar currently renders all users in the database on every keystroke. When there are many users, the UI freezes for a while. To avoid that, we now limit the number of displayed users to 100.

Before:
before

After:
after


src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -132,11 +133,18 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({ venueId }) => {
   const { worldUsers } = useWorldUsers();
   const { openUserProfileModal } = useProfileModalControls();
 
-  const foundUsers = useMemo<JSX.Element[]>(() => {
-    if (!searchQuery) return [];
+  const { foundUsers, foundUsersLength } = useMemo<{
```
daniel-brown-sparkle, line 136: I'd rather extract rendering of users to another component and have the JSX done there without a memo, but i wanted to limit the scope of the change.
e.g. 

```typescript
export const NavSearchUserResults: React.FC<{ users: WithId<User>[], onClick: () => {} }> = ({ users, onClick }) => {
  return (
    <>
      {users.map((user) => (
        <NavSearchResult
          key={`user-${user.id}`}
          title={user.partyName ?? DEFAULT_PARTY_NAME}
          user={user}
          onClick={() => onClick(0)}
        />
      ))}
    </>
  )
}

```

goran-peoski-work:
This PR is duplicate and I disagree with the solution premise. It makes undesired spec change by limiting the result number. The previous PR showed up all the results in a performant manner:
- https://github.com/sparkletown/sparkle/pull/1838

It is still in draft because of prioritization and the added work forward to using native browser API (instead of 3rd party library) to achieve the result that is pending to be done.

daniel-brown-sparkle:
> This PR is duplicate and I disagree with the solution premise. It makes undesired spec change by limiting the result number. The previous PR showed up all the results in a performant manner:
> 
> * [Create NavSearchList component that uses react-window library #1838](https://github.com/sparkletown/sparkle/pull/1838)
> 
> It is still in draft because of prioritization and the added work forward to using native browser API (instead of 3rd party library) to achieve the result that is pending to be done.

@goran-peoski-work it's a low-risk change and solves a real problem for BM. The positive impact of the search bar not causing one's computer to freeze for a few seconds obviously outweighs limiting the number of users to 100, which is probably not a problem at all.

this is a really small change which is super easy to audit and verify. Your larger change can still be added on top of this one

goran-peoski-work:
> it's a low-risk change and solves a real problem for BM. 

I can agree to this if and only if the base isn't `staging` but `env/burn-staging and not delve into the assumptions that follow it, even if they would eventually prove correct.

mike-lvov:
This is a good improvement, but a bit hacky one. I don't mind merging it into `env/burn-staging`. As per @goran-peoski-work 's comment, please set the base of the PR to be `burn-staging`

daniel-brown-sparkle:
> This is a good improvement, but a bit hacky one. I don't mind merging it into `env/burn-staging`. As per @goran-peoski-work 's comment, please set the base of the PR to be `burn-staging`

Ok. That's done

mike-lvov (approved):
LGTM as a hack for Burn event. We have a draft better solution for our main `staging` env: https://github.com/sparkletown/sparkle/pull/1838

PR #2054, opened 2021-08-24
[env/burn-staging] fix bulk room save button

denisdimitrov:
Save button was working only once because of a false check inside the function and it's memoization.

This check already exists in the component and it's disabled prop.
`useAsyncFn` also handles this automatically.


goran-peoski-work:
LGTM

goran-peoski-work (approved):

PR #2064, opened 2021-08-24
[env/burn-staging] check for eventRoom in related venues

denisdimitrov:
eventRoom logic was ran only for the event's venue Id and it wasn't checking for related venues and their rooms. Now the logic will run through all of the related venue rooms and check if that room exists.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1086

### DIFF
--- a/src/components/molecules/ScheduleItemNG/ScheduleItemNG.tsx
+++ b/src/components/molecules/ScheduleItemNG/ScheduleItemNG.tsx
@@ -40,12 +40,18 @@ export interface ScheduleItemNGProps {
 }
 
 export const ScheduleItemNG: React.FC<ScheduleItemNGProps> = ({ event }) => {
-  const { currentVenue: eventVenue } = useRelatedVenues({
+  const { currentVenue: eventVenue, relatedVenues } = useRelatedVenues({
     currentVenueId: event.venueId,
   });
+
+  const relatedVenuesRooms = relatedVenues
+    ?.map((venue) => venue.rooms ?? [])
+    .flat()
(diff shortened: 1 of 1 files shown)


src/components/molecules/ScheduleItemNG/ScheduleItemNG.tsx
```diff
@@ -40,12 +40,18 @@ export interface ScheduleItemNGProps {
...
 
 export const ScheduleItemNG: React.FC<ScheduleItemNGProps> = ({ event }) => {
-  const { currentVenue: eventVenue } = useRelatedVenues({
+  const { currentVenue: eventVenue, relatedVenues } = useRelatedVenues({
     currentVenueId: event.venueId,
   });
+
+  const relatedVenuesRooms = relatedVenues
+    ?.map((venue) => venue.rooms ?? [])
+    .flat()
```
sunny-viktoryia, line 49: i guess you can use `flatMap` here instead of `map` + `flat`

denisdimitrov: Nice catch.

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2064/commits/0b02813c33eada2fe99b1a1e3c16dd15dda5955d

sunny-viktoryia (approved):
looks good! Thanks, @denisdimitrov! 😍

PR #2069, opened 2021-08-25
fixed use hook room url

alexey-petrov:
Fixed `useRoomRecentUsersList` logic for `getRoomUrl` that crashed pages with no `rooms`


sunny-viktoryia (approved):
Horraaaaayyyy!!

PR #2115, opened 2021-08-28
Adds code whitelisting

mike-lvov:
Adds code whitelisting


sunny-viktoryia (approved):
approving, but haven't tested the functionality. 🙃

PR #2116, opened 2021-08-28
Serve resized user avatars in Chat panel 

daniel-brown-sparkle:
We now use Imgix to resize user avatars in the chat panel.
This reduces network payload for avatars by ~4-5x, improves CPU performance and UX.

This is the first iteration to serve optimized images. We'll need to follow up by serving optimized images in the AnimateMap and other places.

This introduces two debts points:
1. A duplication of avatar image sizes between SCSS and the TSX files - ideally we should have one source of truth
2. Hardocded configuration value inside settings.ts - this should come from the env

If someone has a quick win for the above, great. If not, let me know and let's follow-up separately

### DIFF
--- a/src/components/atoms/UserAvatar/UserAvatar.tsx
+++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
@@ -16,7 +16,7 @@ import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 import {
   getFirebaseStorageResizedImage,
   ImageResizeOptions,
-} from "../../../utils/image";
+} from "utils/image";
 
 import "./UserAvatar.scss";
 
--- a/src/components/atoms/UserAvatar/UserAvatar.tsx
+++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
@@ -19,7 +19,7 @@ import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 
 import "./UserAvatar.scss";
 
-export type UserAvatarSize = "small" | "medium" | "large" | "full";
+export type UserAvatarSize = "small" | "medium" | "large" | "xlarge" | "full";
 
 export interface UserAvatarProps extends ContainerClassName {
   user?: WithId<User>;
@@ -35,6 +35,7 @@ const AVATAR_SIZE_MAP: { [key in UserAvatarSize]: number | null } = {
   small: 25,
   medium: 40,
   large: 54,
+  xlarge: 100,
   full: null,
 };
 
--- a/src/components/atoms/UserAvatar/UserAvatar.tsx
+++ b/src/components/atoms/UserAvatar/UserAvatar.tsx
@@ -8,15 +8,14 @@ import { User, UsernameVisibility } from "types/User";
 import { ContainerClassName } from "types/utility";
 
 import { WithId } from "utils/id";
-
-import { useIsOnline } from "hooks/useIsOnline";
-import { useVenueId } from "hooks/useVenueId";
-import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
-
 import {
   getFirebaseStorageResizedImage,
   ImageResizeOptions,
 } from "utils/image";
+
+import { useIsOnline } from "hooks/useIsOnline";
+import { useVenueId } from "hooks/useVenueId";
+import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 
 import "./UserAvatar.scss";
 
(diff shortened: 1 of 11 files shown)


daniel-brown-sparkle:
Some performance data:

### Opening the Party Chat and then Direct Messages on the BM env (~4x improvement)
Before
image
After
image

### Quality of rendering images (a lot sharper)
Before
image
After
image

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -13,6 +13,11 @@ import { useIsOnline } from "hooks/useIsOnline";
 import { useVenueId } from "hooks/useVenueId";
 import { useVenueUserStatuses } from "hooks/useVenueUserStatuses";
 
+import {
+  getFirebaseStorageResizedImage,
+  ImageResizeOptions,
+} from "../../../utils/image";
```
daniel-brown-sparkle, line 19: Fixing this

src/hooks/useUploadProfilePictureHandler.ts
```diff
@@ -35,10 +36,9 @@ export const useUploadProfilePictureHandler = (
         file = new File([resizedImage], fileName);
       }
       const storageRef = firebase.storage().ref();
-      // TODO: add rule to forbid other users to edit a user's image
```
daniel-brown-sparkle, line 38: This is already taken care of by `storage.rules`

src/hooks/useUploadProfilePictureHandler.ts
```diff
@@ -35,10 +36,9 @@ export const useUploadProfilePictureHandler = (
...
       if (user?.uid) {
         const profilePictureRef = storageRef.child(
-          `/users/${user.uid}/${file.name}`
+          `/users/${user.uid}/${uuid()}_${file.name}`
```
mike-lvov, line 41: Any reason to add uuid here?

src/settings.ts
```diff
@@ -748,3 +748,10 @@ export const PROFILE_MODAL_EDIT_MODE_TURNING_OFF_DELAY = 130;
 export const EVENT_STARTING_SOON_TIMEFRAME = 120; // in minutes
 
 export const EVENTS_PREVIEW_LIST_LENGTH = 5;
+
+// Set these to have images uploaded to Firebase Storage served off of Imgix
+// @debt load this from an env variable. This is good enough for Burning Man but we want to have env-specific conf
+export const FIREBASE_STORAGE_IMAGES_ORIGIN =
+  "https://firebasestorage.googleapis.com/v0/b/sparkle-burn.appspot.com/o/";
+export const FIREBASE_STORAGE_IMAGES_IMGIX_URL =
+  "https://sparkle-burn-users.imgix.net/";
```
mike-lvov, line 757: We are merging this into `staging`, aren't we? It means that it'll be propagated to every env

storage.rules
```diff
@@ -2,11 +2,13 @@ rules_version = '2';
...
-        allow read
+        allow read;
     }
     match /users/{uid}/{allPaths=**} {
-        allow read
-        allow write, update: if request.auth.uid == uid
+        allow read;
+        // Allow users to create new files under their unique path.
+        // Don't allow updating files so we can cache files long-term
+        allow write: if request.auth.uid == uid && resource == null;
```
mike-lvov, line 11: will this effect the attemp to update the profile pic?

src/hooks/useUploadProfilePictureHandler.ts
daniel-brown-sparkle: Yes, so images always have unique names so we can cache them. Imagine a user uploads an image called `daniel.png` and then wants to change it and uploads another one called the same. Because of caching, the image might not get updated.

src/settings.ts
daniel-brown-sparkle: This will only impact assets stroed under ` "https://firebasestorage.googleapis.com/v0/b/sparkle-burn.appspot.com/o/";`. So no, it won't have any effect outside the `sparkle-burn` env until we take care of the `@debt` item

storage.rules
daniel-brown-sparkle: Yes. It'll block it. This is also taken care of by the uuid, so it's not strictly necessary but I think it's good to block it here as well so the Firebase raises a permission denied error.

src/hooks/useUploadProfilePictureHandler.ts
mike-lvov: Understood

daniel-brown-sparkle: Added a comment to clarify https://github.com/sparkletown/sparkle/pull/2116/commits/7d66a0f519180b544b61e50b97551e37c37baa4a

storage.rules
mike-lvov: I guess it's a really edge case. Can we just leave the `auth.uid == uid` check here?

mike-lvov (approved):
Let's gooo

PR #2117, opened 2021-08-28
[env/burn-staging] Registration Flow for BM

goran-peoski-work:
Tries to resolve
- https://github.com/sparkletown/internal-sparkle-issues/issues/1012

Large number of files added, most of them copies of pre-existing components with the added *RF suffix (for registration flow) with medium modifications.


mike-lvov (approved):
Didn't look through thoroughly, but desperately need it for the event, so merging it in

PR #2122, opened 2021-08-29
Beavers/firebarrels-shouts-multiplayer-quickmenu-fixes

YarRainbow:
- firebarrels (from data to render)
- "shoutouter" widget and playerio logic
- quickmenu with open profile and send message
- reserve multiplayer logic for first days (separated room mode)
- user name in canvas
- fix player spawn position ("dead zone" checking added)
- fix z-index for general template container
- fix wheel events


mike-lvov (approved):
LGTM

PR #2127, opened 2021-08-29
disabled schedule on venue create

alexey-petrov:
disabled schedule `whats on` btn on /admin/create route


sunny-viktoryia (approved):
THANK YOU!!! 😍 Looks good to me.

PR #2144, opened 2021-08-29
Fix npm run blowing up because of craco.config.js

goran-peoski-work:
```sh
> craco start

(node:68692) UnhandledPromiseRejectionWarning: TypeError: Cannot read p
roperty 'options' of undefined

```


goran-peoski-work:
should've been a single commit

PR #2153, opened 2021-08-29
[BM] navbar adaptive adjustments + simplified view button

sunny-viktoryia:
NavBar:
* hide time on playa and burners count on mobile
* add button "simplified view" to the navbar


mike-lvov (approved):
LGTM

PR #2162, opened 2021-08-30
fixed live attendees count

alexey-petrov:
Fixes live attendees incorrect count ( schedule attendees weren't calculated for room outside of sparkle system & for related venues )


src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -166,13 +166,18 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     firstScheduleDate,
   ]);
 
+  const relatedVenueRooms =
+    relatedVenues.flatMap((venue) => venue.rooms || []).filter((x) => !!x) ||
+    [];
+  const currentVenueRooms = currentVenue?.rooms || [];
```
denisdimitrov, line 172: `relatedVenues` should also contain the current venue, this should be unnecessary. And right now it will have duplicated rooms.

denisdimitrov (changes requested):

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -166,13 +166,18 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     firstScheduleDate,
   ]);
 
+  const relatedVenueRooms =
+    relatedVenues.flatMap((venue) => venue.rooms || []).filter((x) => !!x) ||
+    [];
```
sunny-viktoryia, line 171: can you create consts for empty arrays? to have the only created object and not to cause rerenders and recalculation of the dependant objects.

goran-peoski-work: > can you create consts for empty arrays? to have the only created object and not to cause rerenders and recalculation of the dependant objects.

There is a constant reference to an object and there is a constant object:
1. `const constantReference = {}`
2. `const constantObject = Object.freeze( {} )`

Passing the first one to subsequent renders and possible those components passing around and modifying will make the example 1 basically be a shared mutable state (quite contrary to React principles)

Can that be achieved with `Object.freeze()` or  `useMemo()` instead?

alexey-petrov: fixed emptyArrayObj

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: thanks, fixed c2ecb4a

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -166,13 +164,18 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     firstScheduleDate,
   ]);
 
+  const relatedVenueRooms =
+    relatedVenues
+      .flatMap((venue) => venue.rooms || emptyArrayObj)
+      .filter((x) => !!x) || emptyArrayObj;
```
mike-lvov, line 170:
```suggestion
      .filter(isTruthy) || emptyArrayObj;
```

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -166,13 +164,18 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+
   const roomList = scheduleNG.daysEvents.map((el) => {
     const [roomData] =
-      currentVenue?.rooms?.filter((room) => room.title === el?.room) || [];
+      relatedVenueRooms.filter((room) => room?.title === el?.room) ||
+      emptyArrayObj;
```
mike-lvov, line 175:
```suggestion
      relatedVenueRooms.filter((room) => room?.title === el?.room)
```
U'll always have an empty array based on the above logic

src/hooks/useRoomRecentUsersList.ts
```diff
@@ -33,5 +44,8 @@ export const useRoomRecentUsersList = ({
     locationNames: roomSlugs,
   });
 
-  return recentLocationUsers;
+  const filteredRecentUsers = recentLocationUsers?.filter(
+    (userLocation) => !!userLocation.length
+  );
+  return filteredRecentUsers;
```
mike-lvov, line 50: I don't think we need this check here given the way the hook works:

image

mike-lvov:
Have you checked if it works?

src/hooks/useRoomRecentUsersList.ts
```diff
@@ -22,9 +25,17 @@ export const useRoomRecentUsersList = ({
...
+        (venue) =>
+          roomUrls?.filter(
+            (url, index) =>
+              url.endsWith(venue.id) ||
+              getExternalRoomSlug({
+                roomTitle: roomList && roomList[index].title,
+                venueName: venueId,
+              })
+          ).length
       ),
```
mike-lvov, line 37: Double loop here. Very inefficient

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: can't agree, it's either we filter out rooms that don't match and have 1 room that matches or an empty array

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: fixed 7b01ef2

alexey-petrov:
> Have you checked if it works?

Yes, properly counts remote urls visited + internal venues

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -49,6 +48,8 @@ export interface NavBarScheduleProps {
   venueId: string;
 }
 
+const emptyArrayObj: [] = [];
```
goran-peoski-work, line 59: Shared mutable state danger!

alexey-petrov: removed 7cc46b5

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
alexey-petrov: removed in 7cc46b5 anyway

sunny-viktoryia (approved):
haven't looked through the code in detail. Merging as I tested it and it seems to work. Better than 0s. 😬

PR #2164, opened 2021-08-30
Default room URL: made it to point to the venue

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1120

The change affects only AdminV1 as for AdminV3 it seems unnecessary.

### DIFF
--- a/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
+++ b/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
@@ -20,6 +20,7 @@ import { ExtractProps } from "types/utility";
 import { AnyVenue, PartyMapVenue } from "types/venues";
 
 import { withId } from "utils/id";
+import { venueInsideUrl } from "utils/url";
 
 import { useQuery } from "hooks/useQuery";
 import { useUser } from "hooks/useUser";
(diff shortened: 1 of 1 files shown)


src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -135,7 +136,10 @@ const RoomInnerForm: React.FC<RoomInnerFormProps> = (props) => {
     reValidateMode: "onChange",
     validationSchema: validationSchema,
     validationContext: { editing: !!editingRoom },
-    defaultValues,
+    defaultValues: {
+      ...defaultValues,
+      url: defaultValues.url ?? venueInsideUrl(venueId),
```
mike-lvov, line 141: I don't think I understand where does venueId is coming from

mike-lvov (approved):
Approving if it was tested

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
dmitrymarchuk: On line 45 in RoomsForm component:
```typescript
const venueId = useVenueId();
```

Then it's passed to `RoomInnerForm` on line 102:
```typescript
<RoomInnerForm
        venueId={venueId}
        venue={venue}
        editingRoom={room}
        editingRoomIndex={queryRoomIndex}
      />
```

And then it gets to the `defaultValues`.

PR #2167, opened 2021-08-30
Fixed BM Flashes of 404 screen on Admin panel 

dmitrii-zolotukhin:
- [x] Remove false appearance of 404 page when going to "Add a room"

fixes for sparkletown/internal-sparkle-issues#1123

### DIFF
--- a/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
+++ b/src/pages/Admin/Venue/Rooms/RoomsForm.tsx
@@ -30,6 +30,7 @@ import { LoginRF } from "pages/RegistrationFlow/LoginRF";
 import WithNavigationBar from "components/organisms/WithNavigationBar";
 
 import { ImageInput } from "components/molecules/ImageInput";
+import { LoadingPage } from "components/molecules/LoadingPage";
 
 import { AdminRestricted } from "components/atoms/AdminRestricted";
 import { NotFound } from "components/atoms/NotFound";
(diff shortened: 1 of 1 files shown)


goran-peoski-work (changes requested):
Shouldn't have to use isLoading state in this component. It could probably use a hook like `useAsyncFn` which provides it. Example:

```ts
 const [{ loading: isLoading }, fetchVenueFromAPI ] = useAsyncFn(async () => {} );
```

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
```diff
@@ -52,30 +54,31 @@ export const RoomsForm: React.FC = () => {
...
-      const template = ALL_VENUE_TEMPLATES.find(
-        (template) => data.template === template.template
-      );
+    if (!template || !HAS_ROOMS_TEMPLATES.includes(template.template)) {
+      history.replace("/admin");
+    }
+    setVenue(data as PartyMapVenue);
+  }, [firestore, history, venueId]);
+
+  useEffect(() => {
```
denisdimitrov, line 77: You can use `useAsync` and get rid of this `useEffect`

dmitrii-zolotukhin:
> Shouldn't have to use isLoading state in this component. It could probably use a hook like `useAsyncFn` which provides it. Example:
> 
> ```ts
>  const [{ loading: isLoading }, fetchVenueFromAPI ] = useAsyncFn(async () => {} );
> ```

fixed in 7d11be598f30394d024bebabd3dcf6160fe7d912

src/pages/Admin/Venue/Rooms/RoomsForm.tsx
dmitrii-zolotukhin: @denisdimitrov fixed in 8046f962fca8b50460576d6e7fad9013ae583373

goran-peoski-work (approved):
LGTM

PR #2168, opened 2021-08-30
Fix broken functions js import

sunny-viktoryia:
fixes import; makes it relative as we do that in other files


goran-peoski-work (approved):
LGTM

PR #2176, opened 2021-08-31
ChatSidebar button enhancements

dmitrymarchuk:
UI changes: 
- Add labels to chat sidebar buttons (closes https://github.com/sparkletown/internal-sparkle-issues/issues/1128)
- Open `/admin` in a new tab instead of the same tab (Ed Cook's request on today's meeting)
- Add tooltips to buttons (also Ed Cook)

On the technical side:

- Extract the `ChatSidebarButton` component from `ChatSidebar`
- Remove the absolute positioning of each button in favor of `flex`
- Extract two helper functions `openUrlInNewTab` and `openUrlInExistingTab` from `url.ts`. Gonna make a separate PR to replace all occurrences of `window.location.open` and `window.location.href` in favor of these functions.
- Add a new component `Tooltip` which wraps Boostrap `OverlayTrggeer` and `Tooltip` component


src/components/organisms/ChatSidebar/components/ChatSidebarButton/ChatSidebarButton.scss
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  &__new-message {
+    background-color: $red;
+    @include square-size($spacing--md);
+    border: $new-message-border;
+    position: absolute;
+    border-radius: $border-radius--max;
+    right: -6px;
+    bottom: -2px;
```
dmitrymarchuk, line 38: That's for precise positioning of the red dot. This is the way it was implemented before the refactoring.
image

src/components/organisms/ChatSidebar/ChatSidebar.scss
```diff
@@ -1,6 +1,6 @@
 @import "scss/constants";
 
-$header-height: 45px;
+$header-height: 50px;
```
mike-lvov, line 3: Any reason for this increase? I reckon it can affect some component because not all css code is using $header-height const as per my knowledge

src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -62,7 +60,7 @@ export const _ChatSidebar: React.FC<ChatSidebarProps> = ({ venue }) => {
   }, [toggleSidebar]);
 
   const goToAdmin = useCallback(() => {
-    openUrl("/admin");
+    openUrlInNewTab(`${window.location.origin}/admin`);
```
mike-lvov, line 63: Why new tab?

src/hooks/chats/chatSidebar.ts
```diff
@@ -23,7 +23,7 @@ export const useChatSidebarControls = () => {
   const dispatch = useDispatch();
   const isExpanded = useSelector(chatVisibilitySelector);
   const chatSettings = useSelector(selectedChatSettingsSelector);
-  const newPrivateMessageRecived = useNumberOfUnreadChats();
+  const newPrivateMessageReceived = useNumberOfUnreadChats();
```
mike-lvov, line 26: 👍 good catch. Btw, I think there is an eslint rule/extension that tracks grammar errors in an automated manner

mike-lvov (approved):
LGTM with a few questions

src/components/organisms/ChatSidebar/ChatSidebar.scss
dmitrymarchuk: That is some extra space for those new labels. 
image

Without it buttons look squeezed.

Regarding the UI - I've adjusted not only this but a lot of other styles in the ChatSidebar component. After all that I looked through it and haven't found any style bugs.

src/components/organisms/ChatSidebar/ChatSidebar.tsx
dmitrymarchuk: Ed Cook asked for that specifically. I've just included it in this PR to speed things up.

src/hooks/chats/chatSidebar.ts
dmitrymarchuk: Ooh, that sounds tempting:) I'll update my slack comment with suggestions

src/components/organisms/ChatSidebar/ChatSidebar.tsx
goran-peoski-work: @dmitrymarchuk Could be useful if mentioned in the PR description, with possible links to source. That eliminates questions like this and possible return to re-review

PR #2177, opened 2021-08-31
Add always empty/noop values

goran-peoski-work:
A way to avoid code like this:
- https://github.com/sparkletown/sparkle/pull/2162/files#r699188622

Rationale:
- https://github.com/sparkletown/sparkle/pull/2162#discussion_r698505700

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -758,3 +758,11 @@ export const FIREBASE_STORAGE_IMAGES_ORIGIN =
   "https://firebasestorage.googleapis.com/v0/b/sparkle-burn.appspot.com/o/";
 export const FIREBASE_STORAGE_IMAGES_IMGIX_URL =
   "https://sparkle-burn-users.imgix.net/";
+
+// Helper values that can be safely used in places that might re-render but don't have useMemo/useCallback
+export const ALWAYS_EMPTY_OBJECT = {};
+Object.freeze(ALWAYS_EMPTY_OBJECT);
+export const ALWAYS_EMPTY_ARRAY = [];
+Object.freeze(ALWAYS_EMPTY_ARRAY);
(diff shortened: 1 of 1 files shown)


src/settings.ts
```diff
@@ -758,3 +758,11 @@ export const FIREBASE_STORAGE_IMAGES_ORIGIN =
...
 export const FIREBASE_STORAGE_IMAGES_IMGIX_URL =
   "https://sparkle-burn-users.imgix.net/";
+
+// Helper values that can be safely used in places that might re-render but don't have useMemo/useCallback
+export const ALWAYS_EMPTY_OBJECT = {};
+Object.freeze(ALWAYS_EMPTY_OBJECT);
+export const ALWAYS_EMPTY_ARRAY = [];
+Object.freeze(ALWAYS_EMPTY_ARRAY);
+export const ALWAYS_NOOP_FUNCTION = () => {};
+Object.freeze(ALWAYS_NOOP_FUNCTION);
```
sunny-viktoryia, line 768:
```suggestion
export const ALWAYS_EMPTY_OBJECT = Object.freeze({});
export const ALWAYS_EMPTY_ARRAY = Object.freeze([]);
export const ALWAYS_NOOP_FUNCTION = Object.freeze(() => {});
```

sunny-viktoryia (approved):
:) approving, not sure though if we'll be able to use these consts everywhere because of the typings.

src/settings.ts
goran-peoski-work: This was my original code.

> :) approving, not sure though if we'll be able to use these consts everywhere because of the typings.

☝️ 
That was the reason for the separate lines of the `Object.freeze()`.
Let TS still think they're not ReadOnly<>.

Hence, not accepting the suggestion.

PR #2178, opened 2021-08-31
[env/burn-staging] Use separate inputs for day, month, year instead of calendar input

malaikaaiyar:
~not sure if this is the desired solution or not, but it's here~

A best effort solution for buggy calendar Date of Birth check.

**Before**:

sparkle-dob-02

**After**:

sparkle-dob-01


malaikaaiyar:
we decided

malaikaaiyar (commented):
The code itself looks good to me, but logic-wise you can put in the 31st for months that don't have the 31st in them (and 29-31 for february)

goran-peoski-work:
> The code itself looks good to me, but logic-wise you can put in the 31st for months that don't have the 31st in them (and 29-31 for february)

Correct. I tried to get the code to actually check if it is invalid, it just ends up setting 31st of February as 3rd of March or something :shrug:

malaikaaiyar:
Sounds good. I would approve this, but I can't because I'm the pull request owner

sofisparkle:
This is an 18+ event, please confirm your date of birth to continue.

mike-lvov (approved):
If tested, lgtm

PR #2179, opened 2021-08-31
BM: Dim empty and not live venues on the playa

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1138


mike-lvov (approved):
LGTM

PR #2186, opened 2021-08-31
Prevent re-renders of modified venues

daniel-brown-sparkle:
CloudDataProvider was not caching modified venue data. This caused all the venues which were modified since the user loaded the page to re-render each time `worldUsersApi::processQueuedChanges` ran (i.e. every 5 seconds).

This adds an extra array search but it's something I intend to fix separately in a follow-up PR which will be a part of a bigger rewrite of this logic.


YarRainbow:
Good. Merge after then #2189 , pls
https://github.com/sparkletown/sparkle/pull/2189

daniel-brown-sparkle:
> Good. Merge after then #2189 , pls

@YarRainbow https://github.com/sparkletown/sparkle/pull/2189 is a 1000 line change. This is 10. Do we really need to wait?
thoughts @mike-lvov ?

mike-lvov:
@daniel-brown-sparkle given that @yarikoptic is the main developer of AnimateMap, I'd rather listen to his opinion regarding the better way of changing it

yarikoptic:
am I now who? ;-)

mike-lvov (approved):
LGTM

PR #2174, opened 2021-08-31
[env/burn-staging] Fix art piece aspect ratio and going over 100% width/height

goran-peoski-work:
Resolves 
- https://github.com/sparkletown/internal-sparkle-issues/issues/1132

Removed default anamorphic aspect ratio and added `max-width` and `max-height` of 100% to `ArtPiece__aspect-container`.

From comment https://github.com/sparkletown/internal-sparkle-issues/issues/1132#issuecomment-908940158:
>  -  going with @Sparkle-T insight and setting max width and max height to be 100% on the aspect ratio container;
>  - Height remains 100% to keep the layout with the video strip at the bottom in place;
>  - Removed the default aspect setting class i.e. no default aspect set;
>  - Now the value `anamorphic`, just like `16:9` previously, can be set through DB field `videoAspect` at the venue itself.

PR #2194, opened 2021-09-01
[env/burn-staging] Art Piece aspect ratio followup

goran-peoski-work:
A follow up to
- https://github.com/sparkletown/sparkle/pull/2174

For
- https://github.com/sparkletown/internal-sparkle-issues/issues/1132

This one 
- is adding full width (`width:100%`) setting through DB 
- ability for any custom setting for `aspect-ratio`
  - that isn't a worded constant like the values of `VideoAspectRatio` enum ( `16:9`,  `anamorphic`, `full-width` ) 
  - follows the values convention of https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio
- use default width of 100% when `videoAspect` venue property is missing or empty string


mike-lvov (approved):
LGTM

PR #2201, opened 2021-09-01
Fixes default banner and logo

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1077
Fixes AdminV1 as well as AdminV3

This also fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1156

### DIFF
--- a/src/components/molecules/ImageInput/ImageInput.tsx
+++ b/src/components/molecules/ImageInput/ImageInput.tsx
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
   register: ReturnType<typeof useForm>["register"];
 }
 
-export const ImageInput = React.forwardRef<HTMLInputElement, ImageInputProps>(
-  (
-    {
-      image,
-      remoteUrlInputName,
+export const ImageInput: React.FC<ImageInputProps> = ({
+  image,
+  remoteUrlInputName,
+  remoteImageUrl,
+  containerClassName,
+  imageClassName,
+  name,
+  error,
+  disabled,
+  setValue,
+  register,
+}) => {
+  const imageUrl = useMemo(
+    () =>
+      (image && image.length > 0 && URL.createObjectURL(image[0])) ||
       remoteImageUrl,
-      containerClassName,
-      imageClassName,
-      name,
-      error,
-      disabled,
-      setValue,
-      register,
+    [image, remoteImageUrl]
+  );
+
+  const {
+    loading,
+    errorMessage,
+    handleFileInputChange,
+  } = useImageInputCompression(register, error?.message, name);
+
+  const handleFileInputChangeWrapper = useCallback(
+    async (event: ChangeEvent<HTMLInputElement>) => {
+      const [url, compressedFile] = await handleFileInputChange(event);
+      if (!compressedFile || !url) return;
+
+      setValue(name, [compressedFile], false);
+
+      if (remoteUrlInputName) {
+        setValue(remoteUrlInputName, url, true);
+      }
     },
-    ref
-  ) => {
-    const imageUrl = useMemo(
-      () =>
-        (image && image.length > 0 && URL.createObjectURL(image[0])) ||
-        remoteImageUrl,
-      [image, remoteImageUrl]
-    );
+    [handleFileInputChange, name, remoteUrlInputName, setValue]
+  );
 
-    const {
-      loading,
-      errorMessage,
-      handleFileInputChange,
-    } = useImageInputCompression(register, error?.message, name);
-
-    const handleFileInputChangeWrapper = useCallback(
-      async (event: ChangeEvent<HTMLInputElement>) => {
-        const [url, compressedFile] = await handleFileInputChange(event);
-        if (!compressedFile || !url) return;
-
-        setValue(name, [compressedFile], false);
-
-        if (remoteUrlInputName) {
-          setValue(remoteUrlInputName, url, true);
-        }
-      },
-      [handleFileInputChange, name, remoteUrlInputName, setValue]
-    );
-
-    return (
-      <>
-        <div
-          className={classNames(
-            `image-input default-container`,
-            containerClassName,
-            {
-              disabled: loading,
-            }
-          )}
-        >
-          {imageUrl ? (
-            <img
-              className={`default-image ${imageClassName}`}
-              src={imageUrl}
-              alt="upload"
-            />
-          ) : (
-            <div className="centered-flex empty">
-              <h6 className=" text" style={{ fontWeight: "lighter" }}>
-                Click to upload an image
-              </h6>
-            </div>
-          )}
-          <input
-            disabled={disabled || loading}
-            type="file"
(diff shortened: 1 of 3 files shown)


src/components/molecules/ImageInput/ImageInput.tsx
```diff
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
...
+        <input
+          disabled={disabled || loading}
+          type="file"
+          accept={ACCEPTED_IMAGE_TYPES}
+          onChange={handleFileInputChangeWrapper}
+          className="default-input"
+        />
+        {loading && <ImageOverlay>processing...</ImageOverlay>}
+        {remoteUrlInputName && (
+          <input type="hidden" ref={register} name={remoteUrlInputName} />
```
dmitrymarchuk, line 99: This file looks like a lot of changes but that's mostly indentaion. 
The useful one is this:
before:
```typesctipt
<input
              type="hidden"
              ref={ref}
              name={remoteUrlInputName}
              value={remoteImageUrl}
            />
```
after
```typescript
<input type="hidden" ref={register} name={remoteUrlInputName} />
```

So I removed the `forwardingRef` from this component and passed correct `register` ref to the `input`

src/components/molecules/ImageInput/ImageInput.tsx
```diff
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
...
+  imageClassName,
+  name,
+  error,
+  disabled,
+  setValue,
+  register,
+}) => {
+  const imageUrl = useMemo(
+    () =>
+      (image && image.length > 0 && URL.createObjectURL(image[0])) ||
```
goran-peoski-work, line 41:
```suggestion
      (image.?length > 0 && URL.createObjectURL(image[0])) ||
```

src/components/molecules/ImageInput/ImageInput.tsx
```diff
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
...
+    <>
+      <div
+        className={classNames(
+          `image-input default-container`,
+          containerClassName,
+          {
+            disabled: loading,
+          }
+        )}
+      >
```
goran-peoski-work, line 76: Can you please extract this outside and add the block class as well? Something like this should do it

```tsx
    const parentClasses = classNames({
      "ImageInput image-input default-container": true,
      [containerClassName]: true,
      disabled: loading,
    });
```

src/components/molecules/ImageInput/ImageInput.tsx
```diff
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
...
+        {imageUrl ? (
+          <img
+            className={`default-image ${imageClassName}`}
```
goran-peoski-work, line 79: Same, use outside `classNames()` for composite classes. Makes it easily diffable and maintainable Can't hurt to add `ImageInput__upload-image` or similar class even if not changing the SCSS

src/components/molecules/ImageInput/ImageInput.tsx
```diff
@@ -24,94 +24,82 @@ interface ImageInputProps extends ContainerClassName {
...
-        {errorMessage && <span className="input-error">{errorMessage}</span>}
-      </>
-    );
-  }
-);
-
-ImageInput.displayName = "ImageInput";
+        ) : (
+          <div className="centered-flex empty">
+            <h6 className=" text" style={{ fontWeight: "lighter" }}>
```
goran-peoski-work, line 85: Inline styles should be avoided unless there is no good workaround. This one is an easy fix if used SCSS (there are `$font-weight--*` constants even.

src/components/molecules/ImageInput/ImageInput.tsx
goran-peoski-work: > So I removed the forwardingRef from this component and passed correct register ref to the input

I'm still not used to `forwardingRef` but I did come into issues just trying to get the `register` function from the outside, especially when the validation options should be dictated by parent component where the form resides.

Are you sure this will not cause issues down the road?

src/components/molecules/ImageInput/ImageInput.tsx
goran-peoski-work, line 85: Could probably be useful to replace things like `<h6>` that are with semantical meaning but used only for "free style reuse" with `<div>` or `<span>` having the appropriate styles set.

goran-peoski-work (changes requested):
Some code comments

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: Do not now for any possible issues, but currently, it's working as intended.

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: No, you can't do that as the type of `image?.length` now includes `undefined` and cannot be used inside a boolean expression.

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: That is just an indentation change. I didn't write any of that code. 
And it's pretty much all legacy here. I feel a very strong need to remove that just don't want to start that now.

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: the same https://github.com/sparkletown/sparkle/pull/2201#discussion_r700168039

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: explanation here https://github.com/sparkletown/sparkle/pull/2201#discussion_r700168039

src/components/molecules/ImageInput/ImageInput.tsx
dmitrymarchuk: https://github.com/sparkletown/sparkle/pull/2201#discussion_r700168039

src/components/molecules/ImageInput/ImageInput.tsx
goran-peoski-work: >  I didn't write any of that code.

This https://github.com/sparkletown/sparkle/pull/2201#discussion_r700134601 isn't a note on code you may have written, it is more like a `if you would be so kind to also do this since you're changing code in the vicinity"

I see it in the same venue as deleting some obsolete function that some one else wrote but isn't used anymore. Only in this case, it's connected with the code you write.

On the other side, if  you can't touch code some one else wrote, than all your changes are also in a code some one else wrote. So it's more like, do you think it is too far [OOS] and/or will it take too much time to do so?

goran-peoski-work (approved):
LGTM

PR #2202, opened 2021-09-01
Refactors counting and room

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/640
closes https://github.com/sparkletown/internal-sparkle-issues/issues/245

Replaced all `currentVenueSelectorData` occurences with `currentVenueSelector`(eventually we want to get rid of it too)
Changed the API/structure of room/portal. It can only open the in-sparkle venues. External links should be assigned to `zoomroom` template first.
changed the lastSeenIn object structure. It used to be an object of the following structure: { [locationName]: timestamp }. Now, it's a string containing the whole path: eg. (A > B > C): `/A/B/C/`

DO NOT MERGE UNTIL THE BURN IS OVER


src/pages/Account/helpers.ts
```diff
@@ -10,7 +10,7 @@ import { QuestionsFormData } from "./Questions";
 
 type LocationUpdateData = {
   lastSeenAt: number;
-  lastSeenIn: { [key: string]: number } | null;
```
daniel-brown-sparkle, line 13: This is not backward compatible, right? Will it break anything? You can consider changing the name of the attribute to keep backward incompatibility

src/types/User.ts
```diff
@@ -68,7 +68,7 @@ export interface UserStatus {
 }
 
 export interface UserLocation {
-  lastSeenIn: { [key: string]: number };
+  lastSeenIn: string | null;
```
daniel-brown-sparkle, line 71: Same point on backward compatibility as in `LocationUpdateData `

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -94,15 +92,13 @@ export const Map: React.FC<MapProps> = ({
     (gridPosition: GridPosition) => {
       if (!userUid) return;
 
-      setLocationData({ userId: userUid, locationName: venueName });
```
daniel-brown-sparkle, line 97: Why is this not needed anymore? Is it a duplicate?

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -132,10 +135,22 @@ export const VenuePage: React.FC = () => {
...
+    const getWholePath = async () => {
+      const { sovereignVenue, checkedVenueIds } = await fetchSovereignVenue(
+        venueId
+      );
+
+      const allVenues = [...checkedVenueIds, sovereignVenue.id].reverse();
+
+      const locationPath = wrapIntoSlashes(allVenues.join("/"));
+
+      setLocationData({ userId, locationPath });
```
daniel-brown-sparkle, line 149: This  is the only place that uses `setLocationData`, which in turn just calls `updateLocationData`.
https://github.com/sparkletown/sparkle/blob/7f8e625a2b39dd97ac7eb2d09a8b1e188da3101a/src/utils/userLocation.ts#L34-L42
Can we just call `updateLocationData` and remove `setLocationData` from the codebase? It's really confusing.

Similarly, we can remove `updateCurrentLocationData`

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -132,10 +135,22 @@ export const VenuePage: React.FC = () => {
...
+    };
 
-    setLocationData({ userId, locationName: venueName });
-  }, [userId, venueName]);
+    getWholePath();
```
daniel-brown-sparkle, line 152: Nitpick - This method call has a side effect, so I'd rename it so it's clear it's setting something, not getting something.
e.g. `setWhoPathLocationData`

src/utils/userLocation.ts
```diff
@@ -59,46 +54,16 @@ export const updateCurrentLocationData = ({
...
-  updateLocationData({ userId, newLocationData: {} });
-};
-
-export interface EnterExternalRoomProps {
-  userId: string;
-  roomUrl: string;
-  locationName: string;
-}
-
-export const enterExternalRoom = ({
```
daniel-brown-sparkle, line 82: Just making sure I understand why we don't need this anymore - is it because all external rooms links are opened from a dedicated venue page?

src/utils/userLocation.ts
```diff
@@ -59,46 +54,16 @@ export const updateCurrentLocationData = ({
...
-  userId,
-  locationName,
-  roomUrl,
-}: EnterExternalRoomProps) => {
-  setLocationData({
-    userId,
-    locationName,
-  });
-
-  logEventGoogleAnalytics({
```
daniel-brown-sparkle, line 92: Is anybody looking at these events?

daniel-brown-sparkle:
These changed will break some of the load testing script functionality https://github.com/sparkletown/sparkle/blob/8f63a2562828a960cdfa8615c58d11630ccf7cb1/scripts/load-test-user-location.ts#L104

src/components/templates/AnimateMap/hooks/useRecentLocationsUsers.ts
```diff
@@ -46,10 +46,9 @@ export const useRecentLocationsUsers = (
               worldUserLocationsById[user.id];
 
             return (
-              userLocation.lastSeenIn?.[venueName] &&
-              normalizeTimestampToMilliseconds(
-                userLocation.lastSeenIn[venueName]
-              ) > lastSeenThreshold
+              userLocation.lastSeenIn === venueName &&
```
daniel-brown-sparkle, line 49: Doesn't this need to be modified like https://github.com/sparkletown/sparkle/pull/2202/files#diff-69b65f1189b0befb788170375aded3d463c6398d8804305bab5072c7624a5e93R44 ?

src/utils/user.ts
```diff
@@ -26,17 +26,13 @@ export const getUserLocationData = ({
...
-  const userLastSeenIn =
-    Object.keys(userLocation.lastSeenIn)[0] &&
-    !!Object.keys(userLocation.lastSeenIn)[0].includes(roomName);
-  const userLastSeenLocation =
-    userLocation.lastSeenIn?.[venueName] ||
-    userLocation.lastSeenIn?.[roomName] ||
-    userLocation.lastSeenIn?.[location] ||
-    Object.values(userLocation.lastSeenIn)[0];
-  const isLocationMatch = userLastSeenIn && userLastSeenLocation;
+  const isLocationMatch =
```
daniel-brown-sparkle, line 29: Doesn't this need to be modified like https://github.com/sparkletown/sparkle/pull/2202/files#diff-69b65f1189b0befb788170375aded3d463c6398d8804305bab5072c7624a5e93R44 ?

daniel-brown-sparkle (commented):
Great work @mike-lvov ! 
Let me know if any of the comments don't make sense. I might lack some context.

daniel-brown-sparkle:
If I understand correctly, with this change a user can only be present in a single place? What happens if a user has two or more tabs open? Which venue should they be marked at? 
It may create weird bugs as the interval updating the location will override the value from the other tab every time it's running https://github.com/sparkletown/sparkle/blob/7f8e625a2b39dd97ac7eb2d09a8b1e188da3101a/src/pages/VenuePage/VenuePage.tsx#L127-L134

src/components/organisms/AdminVenueView/components/RoomCard/RoomCard.tsx
```diff
@@ -46,7 +46,7 @@ export const RoomCard: React.FC<RoomCardProps> = ({
 }) => {
   const { user } = useUser();
 
-  const { recentRoomUsers } = useRoom({ room, venueName });
+  const { recentRoomUsers } = useRoom({ room });
```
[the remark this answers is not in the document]
mike-lvov: Good catch. Fixed in https://github.com/sparkletown/sparkle/pull/2202/commits/2657ac450d34fc9c91fbfc90fee24ed74bc75844

src/components/templates/PartyMap/components/RoomAttendance.tsx
```diff
@@ -23,7 +23,7 @@ export const RoomAttendance: React.FC<RoomAttendanceProps> = ({
   room,
   maxVisible = DEFAULT_ROOM_ATTENDANCE_LIMIT,
 }) => {
-  const { recentRoomUsers } = useRoom({ room, venueName: venue.name });
+  const { recentRoomUsers } = useRoom({ room });
```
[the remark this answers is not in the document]
mike-lvov: Good catch. Fixed in https://github.com/sparkletown/sparkle/pull/2202/commits/2657ac450d34fc9c91fbfc90fee24ed74bc75844

Looks like it were all the places

src/hooks/useRoom.ts
```diff
@@ -2,25 +2,29 @@ import { useCallback, useMemo } from "react";
...
 import { useRecentLocationUsers } from "hooks/users";
 import { useUser } from "hooks/useUser";
 
 export interface UseRoomProps {
   room?: Room;
-  venueName: string;
 }
-// @debt refactor useRoom to take venueId instead of venueName
-export const useRoom = ({ room, venueName }: UseRoomProps) => {
+export const useRoom = ({ room }: UseRoomProps) => {
```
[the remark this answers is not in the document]
mike-lvov: Changed the API in https://github.com/sparkletown/sparkle/pull/2202/commits/2b09da7f92a5f5b471740fb6439f21bbab81832e. It differs a bit from your suggestion, but is an improvement

src/hooks/useRoom.ts
```diff
@@ -2,25 +2,29 @@ import { useCallback, useMemo } from "react";
...
   const { user } = useUser();
   const userId = user?.uid;
```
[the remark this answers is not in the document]
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/2202/commits/2b09da7f92a5f5b471740fb6439f21bbab81832e

src/hooks/useRoom.ts
```diff
@@ -2,25 +2,29 @@ import { useCallback, useMemo } from "react";
...
 
   const roomUrl = room?.url ?? "";
 
+  const noTrailSlashPortalUrl = roomUrl && getUrlWithoutTrailingSlash(roomUrl);
+
+  const [portalVenueId] = getLastUrlParam(noTrailSlashPortalUrl);
```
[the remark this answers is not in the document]
mike-lvov: There is venue `A` which has 10 rooms/portals(better to name them portals). Each portal leads to certain venue. portalVenueId is an id of a venue that a portal leads to.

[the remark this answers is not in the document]
mike-lvov: Good take on the parsing side. For sure, it'll be refactored into just having a field like `portalVenueId`, but I decided to keep the current version for backwards compatibility with the existant portals/rooms

src/utils/userLocation.ts
mike-lvov: Yes, correct

src/utils/userLocation.ts
mike-lvov: No, it was done for some hacky way of tracking users and has never been removed

src/pages/Account/helpers.ts
mike-lvov: Changed in https://github.com/sparkletown/sparkle/pull/2202/commits/fa9ecad2e4a8b13737c78d565396fd369a53a2c1

src/types/User.ts
mike-lvov: Changed in https://github.com/sparkletown/sparkle/pull/2202/commits/fa9ecad2e4a8b13737c78d565396fd369a53a2c1

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: No, it's legacy logic

src/pages/VenuePage/VenuePage.tsx
mike-lvov: Done in https://github.com/sparkletown/sparkle/pull/2202/commits/dd77d4b9afdba40f55a14bd7da7e623874e9e92c

src/components/templates/AnimateMap/hooks/useRecentLocationsUsers.ts
mike-lvov: Good catch. Done in https://github.com/sparkletown/sparkle/pull/2202/commits/c10b9c5940ef0d49677132d8549ebe515c2217e3

src/utils/user.ts
mike-lvov: Good catch. Done in https://github.com/sparkletown/sparkle/pull/2202/commits/c10b9c5940ef0d49677132d8549ebe515c2217e3

src/pages/VenuePage/VenuePage.tsx
mike-lvov: Renamed in https://github.com/sparkletown/sparkle/pull/2202/commits/a595fe8a8fe16360a45f0a9ee992430f5e28ac9b

mike-lvov:
> If I understand correctly, with this change a user can only be present in a single place? What happens if a user has two or more tabs open? Which venue should they be marked at?
> It may create weird bugs as the interval updating the location will override the value from the other tab every time it's running
> 
> https://github.com/sparkletown/sparkle/blob/7f8e625a2b39dd97ac7eb2d09a8b1e188da3101a/src/pages/VenuePage/VenuePage.tsx#L127-L134

It's been a status-quo for the last ~8 month.

src/hooks/useRoom.ts
```diff
@@ -1,53 +1,31 @@
...
 export interface UseRoomProps {
   room?: Room;
-  venueName: string;
 }
-// @debt refactor useRoom to take venueId instead of venueName
-export const useRoom = ({ room, venueName }: UseRoomProps) => {
-  const { user } = useUser();
-  const userId = user?.uid;
-
+export const useRoom = ({ room }: UseRoomProps) => {
```
dmitrymarchuk, line 14: I find `useRoom` hook a bit confusing because it:

- is called `useRoom` but accepts `room` as a parameter
- returns `enterRoom` which implies some control over the room entrance
- returns `portalVenueId` which conflicts with the hook name

Maybe it's better to split this into two separate hooks: one for entering the room and one for getting room (or portal) venueId and name them accordingly? It may be out of scope of this PR but it undergoes significant changes here so might as well go even further.

dmitrymarchuk (changes requested):
Some comments & questions

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -124,18 +126,30 @@ export const VenuePage: React.FC = () => {
...
     });
   }, LOC_UPDATE_FREQ_MS);
 
   // @debt refactor how user location updates works here to encapsulate in a hook or similar?
   useEffect(() => {
-    if (!userId || !venueName) return;
+    if (!userId || !venueName || !venueId) return;
+
+    const updateWholeLocationUserPath = async () => {
+      const { sovereignVenue, checkedVenueIds } = await fetchSovereignVenue(
```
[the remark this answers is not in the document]
mike-lvov: @dmitrymarchuk good suggestion. The problem with that hook is that it fetches the sovereignVenue once and doesn't update, when you change the venue.

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -124,18 +126,30 @@ export const VenuePage: React.FC = () => {
...
+        venueId
+      );
+
+      const allVenues = [...checkedVenueIds, sovereignVenue.id].reverse();
+
+      const locationPath = wrapIntoSlashes(allVenues.join("/"));
+
+      updateLocationData({ userId, newLocationPath: locationPath });
+    };
```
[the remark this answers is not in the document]
mike-lvov: @dmitrymarchuk the first one is used to update the time. The second one is used to update the location on venue change

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
dmitrymarchuk: Sooo, maybe that hook needs to be changed? It's surely OOS, just asking.

src/components/templates/PartyMap/components/Map/Map.tsx
mike-lvov: > Why was this needed beforehand and not needed now?

We needed it when we had grid on partymap. But we have never used it

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
dmitrymarchuk: That's the intention I guess? But by just looking at it `newLocationPath` is provided in each of the calls and no time at all. Time is captured inside the `updateLocationData` on every call.
Still confusing to me

src/pages/Account/helpers.ts
goran-peoski-work: @mike-lvov A bit confused here:
`userLocation?.lastVenueIdSeenIn`

Is this going to be an `id` or is it going to be a path constructed of ids with slash as separator?

If it is the second, maybe the name is misleading and would be useful to rename as `lastVenuePath`, `lastLocation` or `lastPathSeenIn` or similar

src/utils/userLocation.ts
```diff
@@ -6,99 +6,29 @@ import { useInterval } from "hooks/useInterval";
...
-export const updateCurrentLocationData = ({
-  userId,
-  profileLocationData,
-}: UpdateCurrentLocationDataProps) => {
-  const [locationName] = Object.keys(profileLocationData);
-
-  updateLocationData({
-    userId,
-    newLocationData: { [locationName]: getCurrentTimeInMilliseconds() },
+    lastSeenIn: newLocationPath,
```
goran-peoski-work, line 24: Should this renamed to the new field name?

goran-peoski-work (commented):
Some notes on the code itself. Haven't check it out and run it to test how it works.

src/types/User.ts
```diff
@@ -68,7 +68,7 @@ export interface UserStatus {
 }
 
 export interface UserLocation {
-  lastSeenIn: { [key: string]: number };
```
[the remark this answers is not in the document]
mike-lvov: Yep, it should be ignored

src/components/templates/AnimateMap/hooks/useRecentLocationsUsers.ts
```diff
@@ -35,21 +33,19 @@ export const useRecentLocationsUsers = (
...
-            return (
-              userLocation.lastSeenIn?.[venueName] &&
-              normalizeTimestampToMilliseconds(
-                userLocation.lastSeenIn[venueName]
-              ) > lastSeenThreshold
+            return userLocation?.lastVenueIdSeenIn?.includes(
+              wrapIntoSlashes(venueId)
             );
           }),
         });
```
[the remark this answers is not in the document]
mike-lvov: https://github.com/sparkletown/sparkle/pull/2202#discussion_r703503693

src/components/molecules/EventDisplay/EventDisplay.tsx
```diff
@@ -78,15 +76,13 @@ export const EventDisplay: React.FC<EventDisplayProps> = ({ event, venue }) => {
...
 const EnterRoomButton: React.FC<EnterRoomButtonProps> = ({
   room,
-  venue,
   children,
 }) => {
-  const { enterRoom } = useRoom({ room, venueName: venue.name });
+  const { enterRoom } = useRoom({ room });
 
   return <div onClick={enterRoom}>{children}</div>;
 };
```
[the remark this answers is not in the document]
mike-lvov: > It's good to strive for a single component per file, even if a simple utility one for greater separation and maintenance.

We have a bunch of such components. I don't thing there is any value in actually extracting this component as they can't exist separately. Anyway, it's OOS, given the amount of changes that are already in this PR.

> @mike-lvov Not sure why a component just to use a hook, but can't it be added to the top of EventDisplay? I don't see a loop going over all rooms, but even if that's the case, what about a useRooms or useEventRooms hook that gets all rooms for the event?

It's the logic I didn't touch as it's a bit unrelated to the suggested change

src/utils/userLocation.ts
mike-lvov: Good catch. It was because of the type inconsistencies. Fixed in https://github.com/sparkletown/sparkle/pull/2202/commits/1c84be82101f89a5bfdb27401c60043d07c0999f

src/utils/string.ts
```diff
@@ -1,3 +1,7 @@
 export const uppercaseFirstChar = (str: string) => {
   return str.charAt(0).toUpperCase() + str.substring(1);
 };
+
+export const wrapIntoSlashes = (str: string): string => {
+  return `/${str}/`;
+};
```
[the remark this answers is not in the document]
mike-lvov: https://english.stackexchange.com/questions/152759/what-preposition-should-be-used-with-the-wrap-verb

> To wrap something IN paper = to use paper to surround something completely (denotes enclosure of one thing by another).

> The most common uses of 'wrap' + 'with' are where 'wrap' is used as a noun, to denote a rolled-up sandwich, or a shawl, for example.

src/hooks/useIsOnline.ts
```diff
@@ -1,15 +1,16 @@
...
 
 export const useIsOnline = (userId?: string) => {
   const { userLocation } = useWorldUserLocation(userId);
-  const lastSeenIn = userLocation?.lastSeenIn;
+  const lastSeenIn = userLocation?.lastVenueIdSeenIn;
 
   return useMemo(
     () => ({
-      isOnline: !isEmpty(lastSeenIn),
+      isOnline: !isTruthyFilter(lastSeenIn),
```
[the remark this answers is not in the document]
mike-lvov: Good catch. Fixed in + added the deprecated mark in https://github.com/sparkletown/sparkle/pull/2202/commits/9a9019e70967d372ed7a8cda7e18e0334e80948e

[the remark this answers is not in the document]
mike-lvov: Removing the occurences + adding the new checker is OOS

src/hooks/useRoom.ts
```diff
@@ -1,53 +1,31 @@
...
-  const roomVenue = useMemo(
-    () => relatedVenues.find((venue) => roomUrl.endsWith(`/${venue.id}`)),
-    [roomUrl, relatedVenues]
-  );
-
-  // @debt we should replace externalRoomSlug with preferrably room id
-  const roomSlug = roomVenue
-    ? roomVenue.name
-    : getExternalRoomSlug({ roomTitle: room?.title, venueName });
+  const noTrailSlashPortalUrl = roomUrl && getUrlWithoutTrailingSlash(roomUrl);
```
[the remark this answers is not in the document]
mike-lvov: Marking it as OOS

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
mike-lvov: @dmitrymarchuk yes, correct. The first one is in the `useInterval`, which updates it every 5 minutes or so. The second one is triggered only on venue change

src/pages/VenuePage/VenuePage.tsx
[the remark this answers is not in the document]
mike-lvov: Yep, good point

dmitrymarchuk (approved):

PR #2209, opened 2021-09-01
 Added jukebox

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1071


src/components/molecules/Jukebox/Jukebox.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }, [messagesToDisplay, updateIframeUrl]);
+
+  const sendMessageToChat = handleSubmit(({ jukeboxMessage }) => {
+    const url = jukeboxMessage;
+    setMessageSending(true);
+    sendJukeboxMsg({ message: jukeboxMessage, url });
+    reset();
+  });
+
+  // This logic disallows users to spam into the chat. There should be a delay, between each message
```
dmitrymarchuk, line 74: The spam-prevention code here looks like it could use the concept of debouncing. Useful article: [link](https://dev.to/heyitsarpit/improve-performance-of-your-react-app-with-debouncing-build-your-own-usedebounce-hook-1fb1).

We already have a `useDebounce` hook in react-use lib but it may be it somewhat limiting, so maybe [this](https://www.npmjs.com/package/use-debounce) one suits better.

src/components/molecules/Jukebox/Jukebox.tsx
```diff
@@ -0,0 +1,133 @@
...
+          />
+
+          <div className="Jukebox__buttons">
+            <ButtonNG
+              aria-label="Send message"
+              className={submitButtonClasses}
+              type="submit"
+              disabled={isBtnDisabled}
+              variant="primary"
+              iconOnly={true}
```
dmitrymarchuk, line 123:
```suggestion
              iconOnly
```

src/hooks/jukebox.ts
```diff
@@ -0,0 +1,127 @@
...
+          });
+
+          if (!displayMessage) return undefined;
+
+          return { ...displayMessage };
+        })
+        .filter(isTruthy),
+    [userId, worldUsersById, messages]
+  );
+};
```
dmitrymarchuk, line 113: This logic looks very similar to the one in `venueChat.ts`. Messages there are also first filtered by query, then filtered by age and then partitioned by replies and transformed into `messagesToDisplay` form. Would be nice if this was deduplicated as we may change the way we deal with chat messages in the future and if this would be deduplicated it would certainly help!

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -706,6 +706,17 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
     </div>
   );
 
+  const renderJukeboxToggle = () => {
+    const isJazzbar = templateID === VenueTemplate.jazzbar;
+
+    return (
+      <div className={`toggle-room DetailsForm${isJazzbar ? "" : "--hidden"}`}>
```
dmitrymarchuk, line 713: can use `classNames` here

src/utils/selectors.ts
```diff
@@ -94,6 +95,13 @@ export const venueChatMessagesSelector: SparkleSelector<
   WithId<VenueChatMessage>[] | undefined
 > = (state) => state.firestore.ordered.venueChatMessages;
 
+export const jukeboxMessagesSelector: SparkleSelector<
+  WithId<JukeboxMessage>[] | undefined
+> = (state) => {
+  // console.log('state', state.firestore)
```
dmitrymarchuk, line 101: probably left accidentially

src/utils/url.ts
```diff
@@ -105,6 +105,22 @@ export const isValidUrl = (url: string): boolean => {
   }
 };
 
+// @debt possibly merge this and isValidUrl right above
```
dmitrymarchuk, line 108: not sure if it was moved from some other place but if this is a new function then why a @debt comment from the start?

src/utils/url.ts
```diff
@@ -105,6 +105,22 @@ export const isValidUrl = (url: string): boolean => {
...
+export const isStringAValidUrl = (urlString: string) => {
+  if (!urlString) return false;
+
+  try {
+    const url = new URL(urlString);
+
+    return VALID_URL_PROTOCOLS.includes(url.protocol);
+  } catch (e) {
+    if (e.name === "TypeError") {
```
dmitrymarchuk, line 117: e here is of type `any` and it's better not to use `name` to check for the error type but use `instanceof`

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -74,7 +83,7 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
   const jazzbarTables = venue.config?.tables ?? JAZZBAR_TABLES;
 
   const [seatedAtTable, setSeatedAtTable] = useState("");
-
+  console.log(seatedAtTable);
```
dmitrymarchuk, line 86: forgotten?

src/components/molecules/Jukebox/Jukebox.scss
```diff
@@ -0,0 +1,93 @@
+@import "scss/constants.scss";
```
dmitrymarchuk, line 1: please extract constants from literal values across this file

dmitrymarchuk (changes requested):

src/hooks/jukebox.ts
alexey-petrov: yeah, I agree that the logic is the same there, altough some pieces are missing for jukebox messages as there's not that much business logic in there
i'd say that with some of the filtering removed in d0f3e28 there's not much to deduplicate
even though they share some logic the entities are different and should be treated so

src/components/molecules/Jukebox/Jukebox.tsx
alexey-petrov: fixed d0f3e28

src/utils/selectors.ts
alexey-petrov: fixed d0f3e28

src/utils/url.ts
alexey-petrov: removed d0f3e28

src/utils/url.ts
alexey-petrov: fixed d0f3e28

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
alexey-petrov: yep, fixed d0f3e28

src/components/molecules/Jukebox/Jukebox.scss
alexey-petrov: fixed d0f3e28

src/pages/Admin/Venue/DetailsForm.tsx
alexey-petrov: fixed d0f3e28

src/components/molecules/Jukebox/Jukebox.tsx
alexey-petrov: probably gonna call it OOS, so that we can implement debounce to both chat/jukebox

PR #2212, opened 2021-09-02
Use one scrollbar style

goran-peoski-work:
Consolidation of the scrollbar styles use. Now only once set for * (which can easily be overridden in the respective components) and unifying for webkit and standard as well.

### DIFF
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -272,6 +272,18 @@ $z-layers: (
 }
 
 @mixin scrollbar {
+  // The standard properties
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width
+  scrollbar-color: opaque-white(0.3) $black;
+  scrollbar-width: thin;
+
+  // The webkit properties
+  @see https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar
+  &::-webkit-scrollbar-corner {
+    background-color: $black;
+  }
+
   &::-webkit-scrollbar-track {
     border-radius: $border-radius--xs;
     background-color: $black;
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -34,7 +34,7 @@ $light-grey--lightest: #acadb0;
 $concrete: #a3a6ab;
 $concrete--dark: #9f9fa0;
 $black: #000000;
-$white: #fff;
+$white: #ffffff;
 $dark: #1a1d24;
 $red: #bc271a;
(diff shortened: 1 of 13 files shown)


goran-peoski-work (commented):
Extra empty lines for consistency added by IDEs autoformat. The rest is mostly removing the `@include scrollbar` since by default all inherit from `*`.

src/scss/global.scss
```diff
@@ -48,29 +48,37 @@ h6 {
...
   font-size: 1.4em;
 }
+
 h5 {
   font-size: 1.2em;
 }
+
 h6 {
   font-size: 1em;
 }
```
goran-peoski-work, line 84: [OOS] All of these should probably be replaced with something like
`<div className="mod--font-10em">`
`<div className="mod--font-12em">`
...
`<div className="mod--font-20em">`

dmitrymarchuk (approved):
LGTM

src/scss/constants.scss
```diff
@@ -271,30 +271,19 @@ $z-layers: (
...
-    background-color: rgba($white, 0.3);
-  }
-
-  &::-webkit-scrollbar-thumb:hover {
-    background-color: rgba($white, 0.5);
+  // The standard properties
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
+  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width
+  scrollbar-color: opaque-white(0.3) $black;
+  scrollbar-width: thin;
```
sunny-viktoryia, line 279: is it a good idea to use them? Because it seems that it's supported only in 3% browsers now..

https://caniuse.com/?search=scrollbar-color

goran-peoski-work: They make a difference in Firefox, so as long as other browsers ignore them - good. If webkit ones decide to support them - good again, it's the same colors.

goran-peoski-work: The support so far https://caniuse.com/?search=scrollbar-

sunny-viktoryia: your solution is supported by 3,16 % of the browser
and the solution you've removed is supported by 92,02% (see https://caniuse.com/?search=webkit-scrollbar)

We could use your solution as an addition to the existing one, but to remove the old one - not sure it was a good idea.

goran-peoski-work: > your solution is supported by 3,16 % of the browser
and the solution you've removed is supported by 92,02% (

@sunny-viktoryia  what solution did I remove?

sunny-viktoryia: sorry, @goran-peoski-work, looks that I was wrong. I thought that you've just removed the old solution with `::-webkit-scrollbar`, but in fact you've just removed the duplicate.

PR #2215, opened 2021-09-02
fix embed being unchangeable bug

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1078


goran-peoski-work (approved):
LGTM

PR #2213, opened 2021-09-02
[Staging] - Flashes of 404 screen on Admin panel 

dmitrii-zolotukhin:
- [x] fixed 404 screen on Admin panel

fixes for sparkletown/internal-sparkle-issues#1123

PR #2182, opened 2021-08-31
[env/burn-staging] fix 404 sparkle logo

malaikaaiyar:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1125

PR #2219, opened 2021-09-02
[staging] Backport NotFound component with uses from BM

goran-peoski-work:
Related:

- https://github.com/sparkletown/sparkle/pull/2090 for
  - https://github.com/sparkletown/internal-sparkle-issues/issues/1034
- https://github.com/sparkletown/sparkle/pull/2097 for
  - https://github.com/sparkletown/internal-sparkle-issues/issues/1034
- https://github.com/sparkletown/sparkle/pull/2182 for
  - https://github.com/sparkletown/internal-sparkle-issues/issues/1125

Precursor to:
- https://github.com/sparkletown/sparkle/pull/2213 for
  - https://github.com/sparkletown/sparkle/pull/2167 
  - https://github.com/sparkletown/internal-sparkle-issues/issues/1123


src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -53,7 +58,12 @@ const RegisterForm: React.FunctionComponent<PropsType> = ({
   closeAuthenticationModal,
 }) => {
   const history = useHistory();
-  const venue = useSelector(venueSelector);
+  const venueFromSelector = useSelector(venueSelector);
```
denisdimitrov, line 61: This is old/legacy, you can remove it and just use the value from `useConnectCurrentVenueNG`

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -103,8 +105,12 @@ export const AdminVenueView: React.FC = () => {
...
   );
 
-  if (!isCurrentVenueLoaded) {
-    return <LoadingPage />;
+  if (!venueId) {
+    return <NotFound />;
+  }
+
+  if (!venue) {
+    return isCurrentVenueLoaded ? <NotFound /> : <LoadingPage />;
```
denisdimitrov, line 113: `<NotFound fullscreen/>` ?

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -197,10 +201,16 @@ export const NavBar: React.FC<NavBarPropsType> = ({
...
+    console.warn(
+      NavBar.name,
+      `aborted display because of missing id (${venueId}) or venue (${currentVenue})`
+    );
+    return null;
+  }
 
   // TODO: ideally this would find the top most parent of parents and use those details
-  const navbarTitle = parentVenue?.name ?? currentVenue.name;
+  const navbarTitle = parentVenue?.name ?? currentVenue?.name;
```
denisdimitrov, line 213: `currentVenue?.name` We need to check and render the code from here on downwards only if `currentVenue` exists otherwise the radios, labels, searching and a lot more won't function like it should. Either our types are off or the guard at the top is not valid.

src/pages/Account/Questions.tsx
```diff
@@ -83,13 +85,14 @@ export const Questions: React.FC = () => {
...
 
   if (!venueId) {
-    return <>Error: Missing required venueId param</>;
+    console.error(Questions.name, "Error: Missing required venueId param");
+    return <NotFound fullScreen />;
   }
 
-  if (isLoaded(venue) && !venue) {
-    return <>Error: venue not found for venueId={venueId}</>;
+  if (!venue && isLoaded(venue)) {
```
denisdimitrov, line 92: You can move the top condition in here. `!venueId || (!venue && isLoaded(venue))`

src/pages/Admin/Venue/VenueAdminPage.tsx
```diff
@@ -40,7 +42,7 @@ export const VenueAdminPage: React.FC = () => {
   }
 
   if (!venue) {
-    return <div className="admin-page-title">This venue does not exist</div>;
+    return <NotFound />;
```
denisdimitrov, line 45: `fullscreen`?

src/pages/VenueLandingPage/VenueLandingPage.tsx
```diff
@@ -40,7 +42,7 @@ export const VenueLandingPage: React.FC = () => {
   }, [venue]);
 
   if (venueRequestStatus && !venue) {
-    return <>This venue does not exist</>;
+    return <NotFound />;
```
denisdimitrov, line 45: `fullscreen`?

src/settings.ts
```diff
@@ -771,6 +771,9 @@ Object.freeze(ALWAYS_EMPTY_ARRAY);
 export const ALWAYS_NOOP_FUNCTION = () => {};
 Object.freeze(ALWAYS_NOOP_FUNCTION);
 
+// NOTE: flag for components that display Not Found screen or use some mitigation when no venueId or no venue exists
+export const ALLOW_NO_VENUE: boolean = true;
```
denisdimitrov, line 775: This is unnecessary, even if we are to add dynamic flag it will be coming from the database and not the env/settings file.

denisdimitrov: I suppose the only benefit is to use it in all of the places atm and then just change it with the database value.

src/components/atoms/NotFound/NotFound.scss
```diff
@@ -0,0 +1,55 @@
...
+  background-color: $black;
+  background-image: url(/assets/images/starfield.jpg);
+  background-size: cover;
+  background-position: center;
+
+  &--full-screen {
+    position: fixed;
+    width: 100vw;
+    height: 100vh;
+    border-radius: 0 !important;
```
denisdimitrov, line 23: Why do you need `!important` which style are you overriding?

denisdimitrov (changes requested):

src/settings.ts
goran-peoski-work: > This is unnecessary, even if we are to add dynamic flag it will be coming from the database and not the env/settings file.

It is not intended to be a dynamic flag. It was intended to be a transitional switch until all code that depends on `venue`/`venueId` existing for the `NavBar` to be visible has been merged. This switch would help those PRs turn this off in case of a problem.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
goran-peoski-work: That will cover the NavBar which is still useful for navigating to beginning or clicking the profile to log out.

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
goran-peoski-work: Not an issue for the display of `NotFound` if value comes from old selectors.

[OOS] Trying to be less disruptive with this PR so an emergent issues will be fixed quicker.
Same selector logic is used in login, just like the registration form, so updating those should be done in separate, refactoring, PR.

src/components/molecules/NavBar/NavBar.tsx
goran-peoski-work: (So the search will return 0 items 🤷)

Last time I checked, there were like 2 components that needed `venueId` but worked OK with it being empty string. 
NavBar isn't visibly broken, and if it happens with some concurrent PR, there's a handy flag that will return the previous behavior until fixed.

src/pages/Admin/Venue/VenueAdminPage.tsx
goran-peoski-work: No https://github.com/sparkletown/sparkle/pull/2219#discussion_r703458494

src/pages/VenueLandingPage/VenueLandingPage.tsx
goran-peoski-work: No https://github.com/sparkletown/sparkle/pull/2219#discussion_r703458494

src/components/atoms/NotFound/NotFound.scss
goran-peoski-work: Just checked, it appears it was only an issue for BM branch, will remove it here.

goran-peoski-work: done @ c94fdeb6b0b69f3582b9698dc433f36526269a87

src/settings.ts
goran-peoski-work: It is not intended to be a dynamic flag. It was intended to be a transitional switch until all code that depends on venue/venueId existing for the NavBar to be visible has been merged. This switch would help those PRs turn this off in case of a problem.

src/pages/Account/Questions.tsx
goran-peoski-work: [OOS] Reserving that change for the Loading component backport.

src/components/molecules/NavBar/NavBar.tsx
denisdimitrov: If it was guarded, it was guarded for a reason. Sometimes it looks stupid and simple and you are thinking I can remove this it's not a big real but then we have huge regression issues like the flashing of the 404 screen. We should either add the guard back or at least add another fallback condition with `??`

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
denisdimitrov: We should at least add `@debt` to it and explanation, otherwise it will fly away and no one will remember to fix it.

goran-peoski-work: Since now the loading and flashes fix are in staging, I can probably do this as well.

src/pages/Account/Questions.tsx
goran-peoski-work: Since now the loading and flashes fix are in staging, I can probably do this as well.

src/components/molecules/NavBar/NavBar.tsx
goran-peoski-work: > If it was guarded, it was guarded for a reason.

That's why I checked them before adding a flag to make it easily revertable. I did my due diligence. Components didn't break, NavBar allowed users to log out or go back.

goran-peoski-work: The difference is this:

No `NavBar`

vs

`NavBar` that doesn't' display title because there isn't  a venue, but does display functional components, even if the Search might not return anything when searching on a non-existent venue.

goran-peoski-work: > but then we have huge regression issues like the flashing of the 404 screen

☝️ Not huge. Not regression. That was legacy code left to do it's original job on purpose. Purpose being: smaller and faster PRs, not doing OOS changes.

goran-peoski-work: @denisdimitrov did you check where `navbarTitle` is actually used?

src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
goran-peoski-work: updated @ 72c493b88f7f0ff2349151449c4a0636634b4880

src/settings.ts
```diff
@@ -662,6 +662,9 @@ Object.freeze(ALWAYS_EMPTY_ARRAY);
...
```
sunny-viktoryia, line 666: not sure about this flag. I don't see the case so far where we might need a page without a venue.
And as it's related to the NavBar and I don't think we should show NavBar if there is no current venue, my suggestion would be to go without it until we really need that flag.

sunny-viktoryia (commented):
comments and suggestions

src/settings.ts
azder: @sunny-viktoryia I'm curious, is navbar without a title an empty navbar? since it still does show the profile image that is clickable and the logo that is also clickable

azder: > not sure about this flag. I don't see the case so far where we might need a page without a venue.

- open https://staging.sparkle.space/admin/invalid
- open https://staging.sparkle.space/admin-ng/missing

src/components/molecules/NavBar/NavBar.tsx
azder: @sunny-viktoryia have you considered the possibility both <NotFound /> and <NavBar /> are being shown?

src/components/atoms/NotFound/NotFound.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import "./NotFound.scss";
+
+export interface NotFoundProps {
+  fullScreen?: boolean;
+}
+
+export const NotFound: React.FC<NotFoundProps> = ({ fullScreen }) => {
+  const history = useHistory();
+  const navigateBack = useCallback(() => history.goBack(), [history]);
```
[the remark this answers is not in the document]
goran-peoski-work: I will try this, more concise, no need of callback

src/components/atoms/NotFound/NotFound.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const componentClasses = classNames({
+    NotFound: true,
+    "NotFound--full-screen": fullScreen,
+  });
```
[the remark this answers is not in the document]
goran-peoski-work: My rule of thumb is to not mix strings and objects as arguments. Using object here makes for separate lines to diff when adding/removing classes.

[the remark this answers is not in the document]
sunny-viktoryia: > My rule of thumb is to not mix strings and objects as arguments of `classNames`

what is this rule based on? 
As for me, the ability to mix strings and objects is the exact reason why we use `classNames`: to be able to use different variable types in one single function that will take care of the rest. And it makes code more visible, as you don't have to search for `true` values in the object, instead, you just see what class name is applied by default.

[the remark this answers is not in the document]
goran-peoski-work: > . And it makes code more visible, as you don't have to search for true values in the object, instead, you just see what class name is applied by default

This is a subjective argument based only on previous experience. Had one spent writing code one way, the code will be "more visible" to them that way.

> ability to mix strings and objects
comes at the price of not having simpler mapping a rule per row the class on the left of `:` and the value/condition on the right of it

`classNames` strenght comes from the ability to evaluate said object parameter which is more elaborate than concatenation of strings.

I hope the above answers
> what is this rule based on?

As to how we shall proceed with this, I'd require a link to the guideline so it is captured here as a reasoning for the resolution.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -245,14 +255,14 @@ export const NavBar: React.FC<NavBarPropsType> = ({
                 <div>{navbarTitle}</div>
               )}
 
-              <VenuePartygoers venueId={venueId} />
+              <VenuePartygoers venueId={venueId ?? ""} />
```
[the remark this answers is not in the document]
goran-peoski-work: Acceptable suggestion: in line with https://doesitsparkle.slack.com/archives/C01700P4QCT/p1631548197056100?thread_ts=1631516472.052000&cid=C01700P4QCT

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -245,14 +255,14 @@ export const NavBar: React.FC<NavBarPropsType> = ({
...
             </div>
 
             {!user && <NavBarLogin />}
 
             {user && (
               <div className="navbar-links">
-                <NavSearchBar venueId={venueId} />
+                <NavSearchBar venueId={venueId ?? ""} />
```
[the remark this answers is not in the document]
goran-peoski-work: Acceptable suggestion: in line with https://doesitsparkle.slack.com/archives/C01700P4QCT/p1631548197056100?thread_ts=1631516472.052000&cid=C01700P4QCT

src/settings.ts
goran-peoski-work: As per https://doesitsparkle.slack.com/archives/C01700P4QCT/p1631548197056100?thread_ts=1631516472.052000&cid=C01700P4QCT

non-empty `NavBar` shown above `NotFound` with the `ALLOW_NO_VENUE` flag removed since enough time has passed from the backport origin where it made sense to use it.

src/components/molecules/NavBar/NavBar.tsx
goran-peoski-work: Since it is not about development taking sides, merging something, product input overriding, looped product team earlier to provide desirable resolution https://doesitsparkle.slack.com/archives/C01700P4QCT/p1631548197056100?thread_ts=1631516472.052000&cid=C01700P4QCT

goran-peoski-work: Now using `hasVenue` check for individual components as per the slack thread

src/components/atoms/NotFound/NotFound.tsx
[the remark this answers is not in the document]
goran-peoski-work: done @ 95715838c6812cfa61939f0a46ed4a15ce3e6f98

src/components/molecules/NavBar/NavBar.tsx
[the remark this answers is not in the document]
goran-peoski-work: using `venueId && hasVenue` check as per slack thread @ d3d4cf50d3102d297ee504e1e6b4762bf039f2fa

src/components/molecules/NavBar/NavBar.tsx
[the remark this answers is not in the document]
goran-peoski-work: using `venueId && hasVenue` check as per slack thread @ d3d4cf50d3102d297ee504e1e6b4762bf039f2fa

src/settings.ts
goran-peoski-work: flag removed, BM is over, most PRs merged @ using `venueId && hasVenue` check as per slack thread @ d3d4cf50d3102d297ee504e1e6b4762bf039f2fa

src/components/atoms/NotFound/NotFound.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <img
+          className="NotFound__tv"
+          alt="shape indicating denied access"
+          src={TV_IMAGE}
+        />
+        <p className="NotFound__move-along-text">Nothing to see here.</p>
+        <ButtonNG
+          className="NotFound__button"
+          variant="primary"
+          onClick={navigateBack}
```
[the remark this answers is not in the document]
goran-peoski-work: done @ 95715838c6812cfa61939f0a46ed4a15ce3e6f98

src/components/atoms/NotFound/NotFound.tsx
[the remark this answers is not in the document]
denisdimitrov: What's the final verdict, what are we going to use?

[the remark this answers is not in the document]
denisdimitrov: I remember there was a Slack convo but it's hard to find anything there.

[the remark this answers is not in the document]
sunny-viktoryia: @goran-peoski-work requested a guideline in our code style docs.

All the engineers (except Goran) have chosen the option I suggested.
https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1631613346195700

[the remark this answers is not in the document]
goran-peoski-work: @denisdimitrov Your question is on point. 

@sunny-viktoryia  There is ambiguity in your goal. So let me present you the two conflated options:

1. You want me to agree the code style you presented is proper.
  - Sadly your arguments didn't support your claim strongly enough for me to agree that's the better code style. Voting doesn't help in this case. A strawman example would be if everyone voted to consider Earth is flat, I'd not find that good argument.

2. You want me to change the code as per your suggestion
 - Luckily, all you need to do is provide here the link to the code style so I can use it as an overriding principle and do the appropriate change. 
 
Thus this (from https://github.com/sparkletown/sparkle/pull/2219#discussion_r708151901) was written:
   > As to how we shall proceed with this, I'd require a link to the guideline so it is captured here as a reasoning for the resolution.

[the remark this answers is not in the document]
goran-peoski-work: FYI, the current code style: https://github.com/sparkletown/sparkle/blob/staging/docs/Questions.md

goran-peoski-work:
Superseded, no plans to merge, ever. Can be closed soon.

PR #2134, opened 2021-08-29
fixed emergency space ui

alexey-petrov:
fixed emergency space ui issues
https://github.com/sparkletown/sparkle/pull/2031

PR #2221, opened 2021-09-02
Turn off camera button

dmitrymarchuk:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1160

- Extracts new component`VideoOverlayButton` for buttons that control audio, microphone, and now camera of the video stream
- Fixes styles of `mute-container` not to include internal layout. That is handled now by a new `CameraMicrophoneControls` component
- Adds `cursor:pointer` where it was missing


src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+      iconEnabled = faVideo;
+      iconDisabled = faVideoSlash;
+      break;
+    case "audio":
+      iconEnabled = faVolumeUp;
+      iconDisabled = faVolumeMute;
+      break;
+    default:
+      assertUnreachable(variant);
+  }
```
goran-peoski-work, line 58: Switch... not the best structure in the programming languages that can do:

```ts
const {iconEnabled,IconDisabled} = OVERLAY_BUTTON_ICONS[variant];

if(!iconEnabled || !iconDisabled){
    throw Error();
}
```

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <div
+      className={classNames("VideoOverlayButton", containerClassName)}
+      onClick={toggle}
+    >
+      <FontAwesomeIcon
+        size="lg"
+        icon={enabled ? iconEnabled : iconDisabled}
+        color={enabled ? undefined : "red"}
+      />
+    </div>
```
goran-peoski-work, line 70: `ButtonNG` has props that support icons, onClick etc

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+  defaultMute: boolean;
+}
+
+export const CameraMicrophoneControls: React.FC<CameraMicrophoneControlsProps> = ({
+  participant,
+  defaultMute,
+  containerClassName,
+}) => {
+  const getChangeStateHandler = useCallback(
+    (variant: "microphone" | "camera"): ((enabled: boolean) => void) => {
```
goran-peoski-work, line 25: export type StreamInputDeviceVariant = 'microphone' | 'camera'

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+        tracks.forEach(
+          (track: AudioTrackPublication | VideoTrackPublication) => {
+            const innerTrack = track.track;
+            if (innerTrack) {
+              if (enable && "enable" in innerTrack) innerTrack.enable();
+              else if ("disable" in innerTrack) innerTrack.disable();
+            }
+          }
+        );
+      };
```
goran-peoski-work, line 41: Since no index is required, one less function with:

```ts
for( const track of tracks) {

}
```

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const changeAudioState = useCallback(getChangeStateHandler("microphone"), [
+    getChangeStateHandler,
+  ]);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const changeVideoState = useCallback(getChangeStateHandler("camera"), [
+    getChangeStateHandler,
+  ]);
```
goran-peoski-work, line 54: Why not exhaustive deps?

`getChangeStateHandler` doesn't have to return a function. `useCallback()` already solves the issue of creating multiple functions.

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  variant,
+  defaultValue,
+  onEnabledChanged,
+  containerClassName,
+}) => {
+  const [enabled, toggle] = useToggle(defaultValue);
+
+  useEffect(() => {
+    onEnabledChanged(enabled);
+  }, [enabled, onEnabledChanged]);
```
goran-peoski-work, line 39:
```suggestion
  useEffect(() => onEnabledChanged(enabled), [enabled, onEnabledChanged]);
```

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+          ? participant.audioTracks
+          : participant.videoTracks;
+
+      return (enable: boolean) => {
+        tracks.forEach(
+          (track: AudioTrackPublication | VideoTrackPublication) => {
+            const innerTrack = track.track;
+            if (innerTrack) {
+              if (enable && "enable" in innerTrack) innerTrack.enable();
+              else if ("disable" in innerTrack) innerTrack.disable();
```
goran-peoski-work, line 37: `in` traverses prototype.

Isn't `innerTrack.enable` good enough? 
```ts
const keys = Object.keys(innerTrack)
if(enable && keys.includes("enable")) //...
```

src/components/organisms/Room/Participant.tsx
```diff
@@ -97,19 +102,22 @@ const Participant: React.FC<React.PropsWithChildren<ParticipantProps>> = ({
...
-    } else {
-      const audioTrack = audioTracks[0];
-      if (audioTrack && audioRef.current) {
-        audioTrack.attach(audioRef.current);
-      }
-    }
-  }, [participant, muted, audioTracks]);
+    },
+    [audioTracks]
+  );
```
goran-peoski-work, line 120: Fail fast and reduced nesting of `if`s

```ts
if( !enable ) {
  return void audioTracks[0]?.detach();  // void operator optional, 2 lines can also serve
}
```

goran-peoski-work (changes requested):
Some code style comments/change requests.

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
dmitrymarchuk: `switch` is perfectly fine and in this case, it doesn't even need runtime error handling as it's compile-time protected.

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
dmitrymarchuk: I've checked and here it greatly changes appearance of the these buttons and requires a lot more code to bring the appearance back than to simply leave the components as they were.

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: That's a matter of choice here and I like `forEach` as a function here. The performance impact is negligible.

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/b1584560cb5a102ef1a5bf47411f29275f37b717

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: With `keys` TS complaints on `innerTrack.enable()` as with `in` it can statically check `enable` invocation.

src/components/organisms/Room/Participant.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/06234506347c70293948d6367762654d711fcc4b

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/06234506347c70293948d6367762654d711fcc4b

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/06234506347c70293948d6367762654d711fcc4b

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
goran-peoski-work: `switch` is never perfectly fine, it's a multiple `goto`s in disguise, and a half way between overengineered simpler approach or underengineered factory

Then why?

```ts
default:
      assertUnreachable(variant);
```

dmitrymarchuk: This ensures during compilation that switch exhausted all the cases.

goran-peoski-work: There is no need if you don't use `switch`. The rule of thumb I have is: 

consider `switch` as a code smell and you will almost always find better way

Example: https://github.com/sparkletown/sparkle/pull/2231#discussion_r701614999

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
goran-peoski-work: I've double checked it. You're right on the `forEach` being better choice here https://github.com/sparkletown/sparkle/pull/2231#discussion_r701616807

Just to note, my comment isn't only about performance, generally `forEach` only has one advantage, the `index` argument it provides. So in most cases I recommend `for-of`.

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
goran-peoski-work: Yep, I've checked... TS quirks... https://github.com/sparkletown/sparkle/pull/2231#discussion_r701616106

goran-peoski-work (changes requested):
Agree with most of the comments @tdmitrii-zolotukhin, still have a request to re-examine one or two points. Hope the diff of this helps https://github.com/sparkletown/sparkle/pull/2231

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
dmitrymarchuk: I see that you've done a fair amount of work to prove your point
Might as well finish this PR yourself as you see fit.

goran-peoski-work: > I see that you've done a fair amount of work to prove your point
@dmitrymarchuk It is not about proving points. It's about explaining my suggestions in clearer way, a rare occasion that involved 15 minutes of coding for the diff to explain better what words couldn't

> Might as well finish this PR yourself as you see fit.
Unfortunately, with time constraints and other issues at hand, I will not have time to build up the context of the issue in general, hence my comments are about code style, not if the PR satisfies the issue requirements in all important cases.

dmitrymarchuk: Why? With tight time constraints, you managed to find time for a separate PR to provide examples for all your suggestions here.
You clearly will find time to just check out this branch and implement this yourself.

goran-peoski-work: There is no need to continue the same conversation from slack DM here. I will summarize it and consider his issues as addressed:

I've suggested  @dmitrymarchuk get another approval and explained it is not an exception to do local coding while reviewing a PR, even if rarely I've pushed code as example (one other case is https://github.com/sparkletown/sparkle/pull/2134#issuecomment-910044232 )

Despite my rationale of this PR might be delayed due to other considerations https://github.com/sparkletown/sparkle/pull/2221#discussion_r701625967
@dmitrymarchuk's decision was to be adamant in me taking over this PR.

That being said, I was OK if this PR got approval from another and I am if another person takes it over to push it over the finish line.

dmitrymarchuk: Yeah, you are right about the delay, so after discussion with @sunny-viktoryia I will finish this as she will do another review. Thank you for your time on this.

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+          ? participant.audioTracks
+          : participant.videoTracks;
+
+      tracks.forEach((track: AudioTrackPublication | VideoTrackPublication) => {
+        const innerTrack = track.track;
+        if (innerTrack) {
+          if (enable && "enable" in innerTrack) innerTrack.enable();
+          else if ("disable" in innerTrack) innerTrack.disable();
+        }
+      });
```
sunny-viktoryia, line 41: maybe it'd look better? 
```suggestion
      tracks.forEach((track: AudioTrackPublication | VideoTrackPublication) => {
        const innerTrack = track.track;
        if (!innerTrack) return;

        if (enable && "enable" in innerTrack) innerTrack.enable();
        if (!enable && "disable" in innerTrack) innerTrack.disable();
      });
```

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <div className={classNames("CameraMicrophoneControls", containerClassName)}>
+      <VideoOverlayButton
+        variant="microphone"
+        onEnabledChanged={changeAudioState}
+        defaultValue={!defaultMute}
+      />
+      <VideoOverlayButton
+        variant="camera"
+        onEnabledChanged={changeVideoState}
+        defaultValue={true}
```
sunny-viktoryia, line 66: let's make a const in settings for default camera enabling. 🙃

src/scss/global.scss
```diff
@@ -551,12 +551,10 @@ h4 {
       bottom: 10px;
       left: 0;
       right: 0;
-      text-align: center;
-      display: flex;
-      justify-content: center;
     }
 
     .mute-other-container {
+      cursor: pointer;
```
sunny-viktoryia, line 557: still not seeing pointer on my venue (JazzBar):
http://localhost:3000/in/viktoryiasmusicbar

src/components/organisms/Room/LocalParticipant.tsx
```diff
@@ -1,9 +1,6 @@
-import React, { useEffect, useState } from "react";
```
sunny-viktoryia, line 1: in this file: let's use named export instead of default one.

src/components/organisms/Room/Participant.tsx
```diff
@@ -1,12 +1,18 @@
-import React, { useEffect, useMemo, useRef, useState } from "react";
```
sunny-viktoryia, line 1: in this file: let's use named export instead of default one.

sunny-viktoryia (commented):
Well! It works! Thanks for the refactoring. A couple of questions / suggestions / and then should be good to go!

src/components/atoms/VideoOverlayButton/VideoOverlayButton.scss
```diff
@@ -0,0 +1,3 @@
+.VideoOverlayButton {
+  user-select: none;
```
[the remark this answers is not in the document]
dmitrymarchuk: чтобы эти иконки нельзя было выделить, а то выглядит не очень
image

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import classNames from "classnames";
+
+import { ContainerClassName } from "types/utility";
+
+import { assertUnreachable } from "utils/error";
+
+import "./VideoOverlayButton.scss";
+
+export type VideoOverlayButtonVariant = "microphone" | "camera" | "audio";
```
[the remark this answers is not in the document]
dmitrymarchuk: The syntax with union time here is just more concise. The difference I know of is that enums are present in runtime but unions are only compile-time TS constructs. Which enables you i.e. enumerate over all possible enum values. Which may be needed but not at this time at least.
But on the other hand you would have to write `VideoOverlayButtonVariant.xxx` instead of just strings which seems redundant to me. What do you think?

src/components/organisms/Room/LocalParticipant.tsx
dmitrymarchuk: You mean to change `React.FC` to `FC` and first line to
```typescript
import React, { FC } from "react";
```
right?

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/0f5a75531713d54e6a9146c2351117298091020d

src/scss/global.scss
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/0f5a75531713d54e6a9146c2351117298091020d

src/components/organisms/Room/LocalParticipant.tsx
sunny-viktoryia: not really. I couldn't comment on the lines I meant (because they were not changed in your PR). 

my request was to replace 
```ts
const LocalParticipant: React.FC<ParticipantProps> = ({}) => {};

export default LocalParticipant;
```

to 
```ts
export const LocalParticipant: React.FC<ParticipantProps> = ({}) => {};
```

in the project we usually use named exports (not default).

dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/38cc1cb1569924f337d9cc44a131f843e8658c1f

src/components/organisms/Room/Participant.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/38cc1cb1569924f337d9cc44a131f843e8658c1f

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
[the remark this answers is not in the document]
sunny-viktoryia: for me, I have no problems writing more symbols like `VideoOverlayButtonVariant.xxx` as long as it gives more context to the type `VideoOverlayButtonVariant` which we use so that we could go there and check other possible values. Writing just "xxx", "yyy" doesn't give context to what we're iterating.

But I wouldn't say that I have a strong disagreement with Union Types. 
Here is the article according to which Union Types win. 
https://blog.bam.tech/developer-news/should-you-use-enums-or-union-types-in-typescript

But the main idea there, I'd say, is:
> But perhaps more importantly, keep in mind that this topic is a bit like tabs vs spaces: while it’s true that there are arguments for both, which one you’re using doesn’t matter that much in the end.

so I don't mind having union type here if you like. 🙃

src/components/atoms/VideoOverlayButton/VideoOverlayButton.scss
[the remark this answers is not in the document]
sunny-viktoryia: 👍🏼  so I guess you've commented that line accidentally now?

src/components/atoms/VideoOverlayButton/VideoOverlayButton.tsx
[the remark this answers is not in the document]
dmitrymarchuk: then keeping union here:) btw, thanks for the article!

src/components/atoms/VideoOverlayButton/VideoOverlayButton.scss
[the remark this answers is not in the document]
dmitrymarchuk: omg, yes!😟
and accidentally left a comment in russian😂
fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/63d74ada8e8d6d77a0198f0e4264c1e503b2e4ea

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useCallback } from "react";
+import classNames from "classnames";
+import Video, {
+  AudioTrackPublication,
+  VideoTrackPublication,
+} from "twilio-video";
+
+import { DEFAULT_CAMERA_ENABLED } from "../../../settings";
```
sunny-viktoryia, line 8:
```suggestion
import { DEFAULT_CAMERA_ENABLED } from "settings";
```

src/components/templates/AnimateMap/components/FirebarrelWidget/FirebarrelWidget.tsx
```diff
@@ -11,13 +11,13 @@ import { AnimateMapVenue } from "types/venues";
...
 
-import LocalParticipant from "components/organisms/Room/LocalParticipant";
+import { LocalParticipant } from "components/organisms/Room/LocalParticipant";
 import VideoErrorModal from "components/organisms/Room/VideoErrorModal";
 
 import { Button } from "components/atoms/Button";
 
 import { WithId } from "../../../../../utils/id";
-import Participant from "../../../../organisms/Room/Participant";
+import { Participant } from "../../../../organisms/Room/Participant";
```
sunny-viktoryia, line 20: please, make these imports absolute and sort them out. 🙃

sunny-viktoryia (approved):
Approving! Yay, it works!! 🙃 
please, fix the imports.

If you have the capacity now, you can do that in this PR, or create a follow up for:

> Hmm, probably OOS of this PR. But could you change styles, making it more obvious that the person has turned off his camera?
Possible solutions:
show display name (or just first letters as it's done in zoom)?
or move the icon from the bottom left corner to the center?
\+ add some borders to the video box when it's empty

image

src/components/molecules/CameraMicrophoneControls/CameraMicrophoneControls.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/34c2da4fc00fb2f75d093deb88233a7c5846ec92

src/components/templates/AnimateMap/components/FirebarrelWidget/FirebarrelWidget.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2221/commits/34c2da4fc00fb2f75d093deb88233a7c5846ec92

src/components/organisms/Room/Participant.tsx
```diff
@@ -97,37 +111,29 @@ const Participant: React.FC<React.PropsWithChildren<ParticipantProps>> = ({
...
   return (
     <div className={`col participant ${bartender ? "bartender" : ""}`}>
-      {videos}
+      {!videoEnabled && <div className="participant--video-disabled" />}
+      <video
+        ref={videoRef}
+        autoPlay={true}
+        className={profileData?.mirrorVideo ? "mirrored" : ""}
+      />
+      <audio ref={audioRef} autoPlay={true} />
```
sunny-viktoryia, line 136: why was that removed from useMemo?

sunny-viktoryia (commented):
a question here.

And could you also fix the border radiuses and backgrounds on empty videos? 
image
image

src/components/organisms/Room/Participant.tsx
dmitrymarchuk: restored in https://github.com/sparkletown/sparkle/pull/2221/commits/c8ea9c6fae402dd1cb702dba9ecf856e61232663

dmitrymarchuk:
Fixed border radiuses, made background appear properly when others disable videos. Also made mic & camera controls be hidden on others videos.

sunny-viktoryia (approved):
Awesome! Approving for the 2nd time! Thank you so much for the refactoring and styles fixes! 😍😍😍

PR #2227, opened 2021-09-02
Prevent schedule rendering when hidden

daniel-brown-sparkle:
The schedule component takes a lot of processing power to render (relatively speaking) when there are many events. We currently render it on page load even though it is hidden. It also gets re-rendered whenever `worldUsersApi` cache is updated, which has a constant negative impact on UX.

This PR prevents the schedule from rendering when it is not visible.

There are two small negative side effects;

1. There's a short loading time between opening the schedule and it shows. This is the deferred load time that used to happen when the page loaded. i.e. users wait for less time on every page load but wait for more time when they open the schedule

2. The state of the schedule doesn't persist between closing and reopening. For example, if a user opens the schedule, switches a date, and then closes and reopens the schedule, it will show the `Today` tab again, instead of the last chosen date.

The above is an acceptable tradeoff for the performance boost in my opinion. Plus, we can iterate on the component to address those while keeping it performant.

----

Currently on staging: (note the bad padding on the left and the over-run of the dialog over the footer)
image

With this PR:
image

Loading state:
image


src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
 
 $schedule-btn--height: 51px;
 
-.NavBarWrapper {
```
daniel-brown-sparkle, line 13: This is merged into `NavBarSchedule`

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
...
-}
-
 .NavBarSchedule {
   position: fixed;
   z-index: z(navbar-schedule);
-  top: 0;
+  top: $navbar-height;
   left: 0;
-  height: calc(100% - #{$schedule-btn--height});
+  bottom: $footer-height;
```
daniel-brown-sparkle, line 18: Instead of using a calculated height, I've given it a bottom value. on the burn-staging branch this is set to the height of the footer, so I used that here as well. It looks better!

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
...
   width: $NavBarSchedule--width;
 
-  padding-top: $navbar-height;
```
daniel-brown-sparkle, line 57: Since we don't have a container, I just set this as the `top` value. padding-top is not needed then

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
...
   padding-left: $NavBarSchedule--padding;
 
   background-color: opaque-black(0.8);
-  overflow: auto;
 
-  pointer-events: none;
+  pointer-events: all;
```
daniel-brown-sparkle, line 25: This is not being rendered when the Schedule is invisible, so no need for two values.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
...
   transition: all 400ms $transition-function;
 
+  display: flex;
+  flex-direction: column;
+
   &--show {
-    pointer-events: all;
     opacity: 1;
-    transform: translateY(0);
+    transform: translateX(0);
```
daniel-brown-sparkle, line 37: I have no idea how this worked before! 🤔

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -10,67 +10,36 @@ $schedule-btn--padding: 0.35em 1em;
...
+  }
+
+  &__schedule {
+    height: calc(100% - #{$schedule-btn--height});
```
daniel-brown-sparkle, line 41: Although I didn't introduce this, I would've preferred to have `flex-grow: 1` here, but it has some odd side effects. Something for another time.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -125,4 +94,30 @@ $schedule-btn--height: 51px;
...
+
+    text-shadow: 0 2px 4px opaque-black(0.3);
+    box-shadow: 0 3px 6px opaque-black(0.2);
+    padding: $schedule-btn--padding;
+
+    &:hover,
+    &:focus {
+      opacity: 1;
+    }
+  }
```
daniel-brown-sparkle, line 122: This was all just moved from `NavBarWrapper`

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -244,24 +243,40 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
...
+              </Button>
+            )}
             <Button
-              onClick={downloadPersonalEventsCalendar}
-              customClass="NavBarWrapper__download-schedule-btn"
+              onClick={downloadAllEventsCalendar}
+              customClass="NavBarSchedule__download-schedule-btn"
             >
-              Download your schedule
+              Download full schedule
```
daniel-brown-sparkle, line 261: Git shows the diff wrong. The change was nesting under an empty element, indentation and replacing `NavBarWrapper` class names with `avBarSchedule`

daniel-brown-sparkle:
@mike-lvov The schedule component is quite different on the burning man branch so I don't think you'll be able to apply this change. it's probably not worth it, but let me know if you want me to create a patch especially for that branch.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
dmitrymarchuk: `&--show` style was used to toggle opacity of the schedule before this PR. Now, when `isVisible` is `false` the `NavBarScheduleBody` is not rendered at all and there is no need for opacity change, right?

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -44,19 +44,22 @@ export interface ScheduleNGDay {
...
 interface UserWithVenueIdProps extends WithId<User> {
   venueId?: string;
   portalId?: string;
 }
 
-export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
-  isVisible,
+export const emptyPersonalizedSchedule = {};
+
+export const NavBarScheduleBody: React.FC<NavBarScheduleBodyProps> = ({
```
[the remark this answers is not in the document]
dmitrymarchuk: I suggest extracting `NavBarScheduleBody` into a separate file.

dmitrymarchuk (changes requested):
Great idea @daniel-brown-sparkle! Left some suggestions.

And I also think you can easily solve the second negative side effect by lifting up the component tree those useful pieces of state that need to be preserved between schedule openings. 
Seems that there are only two: show personalized schedule toggle and selected day.
You can move 
```typescript

  const {
    isShown: showPersonalisedSchedule,
    toggle: togglePersonalisedSchedule,
  } = useShowHide(false);
```

from newly created `NavBarScheduleBody` to `NavBarSchedule` and pass necessary props and setState handlers back to `NavBarScheduleBody`
What to you think?

PR #2233, opened 2021-09-03
[env/burn-staging] Update show only events in that venue by default 

ermiaSparkle:
Resolve issue:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1002#issuecomment-911915629

### DIFF
--- a/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
+++ b/src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
@@ -94,7 +94,9 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     currentVenueId: venueId,
   });
 
-  const [filterRelatedEvents, setFilterRelatedEvents] = useState(false);
+  const [filterRelatedEvents, setFilterRelatedEvents] = useState(
+    currentVenue?.id !== sovereignVenue?.id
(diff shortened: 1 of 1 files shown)


denisdimitrov (approved):
LGTM!

src/components/organisms/NavBarSchedule/NavBarSchedule.tsx
```diff
@@ -94,7 +94,9 @@ export const NavBarSchedule: React.FC<NavBarScheduleProps> = ({
     currentVenueId: venueId,
   });
 
-  const [filterRelatedEvents, setFilterRelatedEvents] = useState(false);
+  const [filterRelatedEvents, setFilterRelatedEvents] = useState(
+    currentVenue?.id !== sovereignVenue?.id
+  );
```
goran-peoski-work, line 99: What happens if both are `undefined` ?

ermiaSparkle: will be `false` and it will work as normal.

goran-peoski-work: K, good enough. 

I've been making the venue/venueId optional for the purpose of 404 inside admin, just so there is NavBar visible

PR #2238, opened 2021-09-03
[staging] Backport the AdminRestricted component with uses

goran-peoski-work:
Follow up to
- https://github.com/sparkletown/sparkle/pull/2068

for
- https://github.com/sparkletown/internal-sparkle-issues/issues/1033


src/components/atoms/AdminRestricted/AdminRestricted.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const firebase = useFirebase();
+  const history = useHistory();
+  const venueId = useVenueId();
+  const { userId } = useUser();
+
+  const { isAdminUser, isLoading: isCheckingRole } = useIsAdminUser(userId);
+
+  const [{ loading: isLoggingOut }, logout] = useAsyncFn(async () => {
+    await firebase.auth().signOut();
+    history.push(venueId ? venueLandingUrl(venueId) : "/");
```
denisdimitrov, line 29: If the top line fails this will still redirect, the questions are, what happens when signOut() fails? Is this expected behaviour?

src/pages/Admin/Venue/VenueWizard/VenueWizard.tsx
```diff
@@ -2,22 +2,25 @@ import React, { useReducer } from "react";
...
   const [state, dispatch] = useReducer(VenueWizardReducer, initialState);
 
-  if (venueId)
-    return (
-      <VenueWizardEdit venueId={venueId} state={state} dispatch={dispatch} />
-    );
-
-  return <VenueWizardCreate state={state} dispatch={dispatch} />;
+  return (
+    <AdminRestricted>
```
denisdimitrov, line 16: You should also add this component in admin v3, that's `Admin_v2.tsx` `AdminAdvancedSettings` and `AdminVenueView`

denisdimitrov (changes requested):

src/components/atoms/AdminRestricted/AdminRestricted.tsx
goran-peoski-work: It is the current behavior.  
Code is extracted from existing example:

```tsx
const logout = useCallback(async () => {
    await firebase.auth().signOut();

    history.push(
      //@debt seems like a legacy logic
      IS_BURN ? "/enter" : venue.id ? venueLandingUrl(venue.id) : "/"
    );
  }, [firebase, history, venue.id]);
```

As I've done light checks during development, the sign out looks like it happens on client side (probably before or not even server is notified - a library simply needs to delete a cookie or similar token).

goran-peoski-work: > If the top line fails this will still redirect,

How does that happen? A rejected value on `await` will not advance execution to the next line.

src/pages/Admin/Venue/VenueWizard/VenueWizard.tsx
goran-peoski-work: added @ 2d25e8993a1a328bdc9768cf97a372e6f1d8b58c

denisdimitrov (approved):
LGTM!

PR #2239, opened 2021-09-03
Default displayWhile to true in <Loading> to unbreak older uses

goran-peoski-work:
Fix for regression in:
- https://github.com/sparkletown/sparkle/pull/2117

Comment (https://github.com/sparkletown/sparkle/commit/c6b6346682d4f81d4e9e95591c7b6791ee03da1c#r55837037):

> this change is not backward compatible and causes <Loading /> to not render anything where it was being used before this PR. e.g ScheduleNg, EmojiPicker.

BONUS regression fix: round off square buttons in chat

**Before**:

 
sparkle-chat-squares-02

**After**:
**sparkle-chat-squares-01**:

PR #2242, opened 2021-09-03
[staging] Backport of Loading component with displayWhile defaulted to true

goran-peoski-work:
Similar to https://github.com/sparkletown/sparkle/pull/2239 only for staging and without the chat buttons update.

Preemptive addition before:
- https://github.com/sparkletown/sparkle/pull/2213
- https://github.com/sparkletown/sparkle/pull/2219
- https://github.com/sparkletown/sparkle/pull/2238


daniel-brown-sparkle:
Out of curiosity, why do we even this parameter? Wouldn't it be easier to not render the component in the first place?

goran-peoski-work:
> Out of curiosity, why do we even this parameter? Wouldn't it be easier to not render the component in the first place?
> 
> e.g. instead of
> 
> ```tsx
> <Loading displayWhile={isUpdating} />
> ```
> 
> We can do
> 
> ```tsx
> {!isUpdating && <Loading /> }
> ```

So it can bet he reverse in places with heavy 
```tsx
{a && !b || c && <>block of tags</>}
{b || d && !e && <>another block of tags</>}
{!isUpdating && <Loading />}
{a && e && <>another block of tags</>}
```
or no TSX at all.

ermiaSparkle (approved):
LGTM!

src/components/molecules/Loading/Loading.tsx
```diff
@@ -9,18 +9,19 @@ import "./Loading.scss";
 
 export interface LoadingProps extends ContainerClassName {
   label?: string;
+  displayWhile?: boolean;
```
denisdimitrov, line 12: `displayWhile` display while what? The name is confusing. This can be simplified to `show` `visible` `isLoading`

goran-peoski-work: Idea is to have `displayWhile={isLoading}` in places where `{ isLoading && <Loading> }` might be an overkill of braces. Not meant to be <Loading show /> since it's `true` by default.

PR #2244, opened 2021-09-03
Beavers/artcars

YarRainbow:
new artcars


sunny-viktoryia (approved):
approving without a deep look. All the changes are related to AnimateMap.

PR #2254, opened 2021-09-06
Add autoPlay field in Venue for embeddable types

ermiaSparkle:
Close issues:

- closes https://github.com/sparkletown/internal-sparkle-issues/issues/1117
- closes https://github.com/sparkletown/internal-sparkle-issues/issues/1161

It might fix the issue :
- https://github.com/sparkletown/internal-sparkle-issues/issues/1110


src/components/templates/Audience/Audience.tsx
```diff
@@ -150,7 +150,7 @@ export const Audience: React.FC<AudienceProps> = ({ venue }) => {
   useLayoutEffect(() => {
     if (!venue) return;
 
-    setIframeUrl(ConvertToEmbeddableUrl(venue.iframeUrl, true));
+    setIframeUrl(ConvertToEmbeddableUrl(venue.iframeUrl, venue?.autoPlay));
```
goran-peoski-work, line 153: This legacy function starts with a capital and has a boolean trap. @ermiaSparkle how much would it take to do a light change of its interface? Something like

```ts
export interface ConvertToEmbeddableUrlOptions {
  url?: string;
  autoplay?: boolean;
}

// @debt Replace the naive logic with use of new URL(url) which provides standard parsing before doing checks
export const convertToEmbeddableUrl: (
  options: ConvertToEmbeddableUrlOptions
) => string = ({ url, autoplay = true }) => {
```

and replace in the places it is invoked from?

goran-peoski-work: If it takes too much time, then at least 2 @debt lines: the one from the code example above and one to update the interface of it

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -650,6 +652,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
       {errors.iframeUrl && (
         <span className="input-error">{errors.iframeUrl.message}</span>
       )}
+      <label
+        htmlFor={"autoPlaycheckbox"}
+        className={`checkbox ${
+          !watch("autoPlay", false) && "checkbox-checked"
+        }`}
+      >
+        If you are embedding a video, do you want to disable autoplay?
```
goran-peoski-work, line 661: Is it going to be a checkmark?
Maybe not present it as an opt-in or opt-out but instead:

Autoplay your embeded video ☑️

src/pages/Admin/Venue/DetailsForm.tsx
```diff
@@ -650,6 +652,21 @@ const DetailsFormLeft: React.FC<DetailsFormLeftProps> = ({
...
+      </label>
+      <input
+        type="checkbox"
+        id={"autoPlaycheckbox"}
+        name={"autoPlay"}
+        defaultChecked={!values.autoPlay}
+        ref={register}
+      />
```
goran-peoski-work, line 669: There should be `Toggler` components in the same form which this input can copy from. Example:

```tsx
     <div className="toggle-room">
      <h4 className="italic input-header">Show Schedule</h4>
      <Toggler
        name="showSchedule"
        forwardedRef={register}
        defaultToggled={DEFAULT_SHOW_SCHEDULE}
      />
```

src/types/venues.ts
```diff
@@ -141,6 +141,7 @@ export interface BaseVenue {
   code_of_conduct_questions: Question[];
   owners: string[];
   iframeUrl?: string;
+  autoPlay?: boolean;
```
goran-peoski-work, line 144: Should this be defined in the two places `EmbeddableVenue` and `BaseVenue`?

It can either be property for all venues or just for embeddable venues. Redundant to have it in both.

Similar question with `JazzbarVenue`.

goran-peoski-work (changes requested):
Notes on the DB field, using checkbox vs Toggler and possible adjustment of the URL generator function

goran-peoski-work (changes requested):
Second (a visual) pass review.

## 1. Input

sparkle-embed-01

Idea is good. One thing though, no matter if sticking with checkbox or replacing with `Toggler`,
the need of `if` question can be avoided by actual `if` in the code to display the question only if the field is not empty (might be OOS stretch to check for further context of the input).

## 2. Output

Update is done on  `IFRAME_TEMPLATES.includes(data.template)`, yet only 3 of the `.tsx` files are updated. Are there more that need update?
Here is the array for completeness:
```ts
const IFRAME_TEMPLATES = [
  VenueTemplate.artpiece,
  VenueTemplate.audience,
  VenueTemplate.embeddable,
  VenueTemplate.firebarrel,
  VenueTemplate.jazzbar,
  VenueTemplate.performancevenue,
];
```

sunny-viktoryia:
@ermiaSparkle, if you mention the issues you're going to fix in your PR, please, use closing keywords in your PR description
for example:
> closes isssue-url

doing this way, the issue will be automatically closed once you merge your PR.

see the docs:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

src/components/templates/Audience/Audience.tsx
ermiaSparkle: fixed in 418cdcb4763615a45870bc6b1dfd26c2ec198a84

src/pages/Admin/Venue/DetailsForm.tsx
ermiaSparkle: fixed in 418cdcb4763615a45870bc6b1dfd26c2ec198a84

src/pages/Admin/Venue/DetailsForm.tsx
ermiaSparkle: fixed in 418cdcb4763615a45870bc6b1dfd26c2ec198a84

src/types/venues.ts
ermiaSparkle: fixed in 418cdcb4763615a45870bc6b1dfd26c2ec198a84

src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -194,7 +195,10 @@ const Jazz: React.FC<JazzProps> = ({ setUserList, venue }) => {
                     key="main-event"
                     title="main event"
                     className="iframe-video"
-                    src={`${iframeUrl}?autoplay=1`}
+                    src={`${convertToEmbeddableUrl({
+                      url: venue.iframeUrl,
+                      autoPlay: venue?.autoPlay,
+                    })}`}
```
goran-peoski-work, line 201: `convertToEmbeddableUrl` should already return a string, no? In case it doesn't, adding ?? "" after it should serve the same as having those extra braces for the template string
```suggestion
                    src={ convertToEmbeddableUrl({
                      url: venue.iframeUrl,
                      autoPlay: venue?.autoPlay,
                    }) ?? "" }
```

goran-peoski-work (changes requested):
One minor note about a prop

goran-peoski-work (approved):
LGTM

PR #2255, opened 2021-09-06
Fix BM External Link page 2.0 text wrapping

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1168

fixed text wrapping. see example below

Screenshot 2021-09-06 at 11 43 42

### DIFF
--- a/src/components/templates/ExternalRoom/ExternalRoom.tsx
+++ b/src/components/templates/ExternalRoom/ExternalRoom.tsx
@@ -67,7 +67,12 @@ export const ExternalRoom: React.FC<ExternalRoomProps> = ({ venue }) => {
         <>
           <div className="ExternalRoom__message">
             <div>This page should automatically open</div>
-            <a rel="noreferrer" href={redirectUrl} target="_blank">
+            <a
+              rel="noreferrer"
+              href={redirectUrl}
+              target="_blank"
+              className="ExternalRoom__link"
+            >
               {redirectUrl}
             </a>
 
--- a/src/components/templates/ExternalRoom/ExternalRoom.tsx
+++ b/src/components/templates/ExternalRoom/ExternalRoom.tsx
@@ -11,7 +11,7 @@ import {
 import { AnyVenue } from "types/venues";
 
 import { WithId } from "utils/id";
-import { openUrl } from "utils/url";
+import { getExtraLinkProps, openUrl } from "utils/url";
 
 import { useRecentVenueUsers } from "hooks/users";
 
@@ -68,10 +68,9 @@ export const ExternalRoom: React.FC<ExternalRoomProps> = ({ venue }) => {
           <div className="ExternalRoom__message">
             <div>This page should automatically open</div>
             <a
-              rel="noreferrer"
               href={redirectUrl}
-              target="_blank"
               className="ExternalRoom__link"
+              {...getExtraLinkProps(true)}
             >
               {redirectUrl}
             </a>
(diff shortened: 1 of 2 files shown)


yurii-lubynets:
@whoaitsaimz @sofisparkle 

please check the screen above to make sure the adjusted UI is good for a really long URL

goran-peoski-work (changes requested):
Requesting changes as I'm not sure the increase of width and decrease of height is the way to go. Might have helpful input from @whoaitsaimz and @sofisparkle 

Current progress (from https://burn-staging.sparkle.space/in/longurlportal):
sparkle-long-url-01


Possible avenues to take ( @yurii-lubynets or anything else you might find more useful ):

-  `<PrettyLink title={room.url} onClick={() => openRoomUrl(room.url)} />` for the long url
- `.EmergencyView { }` style on how to make the whole page not expand horizontally

sunny-viktoryia:
instead of horizontal scroll, could you make the text wrapping?
```css
overflow-wrap: break-word;
```

and I guess:
```css
max-width: 50rem
```
looks better for the modal

yurii-lubynets:
@goran-peoski-work @sunny-viktoryia 

Thanks for your input I've updated styles for the external link and for the venue description. Now it looks much better
Screenshot 2021-09-06 at 16 07 07

and the general case for standard link & description
Screenshot 2021-09-06 at 16 10 01

goran-peoski-work:
> ```css
> max-width: 50rem
> ```
> 
> looks better for the modal

I would stay clear of `rem` in this case.

For one it is dependent on the root font ( usually `16px`) while the text inside the modal might have different font size (not uncommon for links to be `0.8em`, and headers anywhere from `1em` to `2em`, so as a relative text width it might be off.

As a second thing, the width of this dialog should best reflect the page/container, depending on what's more important or how the layout is structured, so probably `%` or `vh`/`vw` or a combination of those.

In any case, the original width/height, if changed, should probably be updated in the room modal itself, for consistency (that one is surely too narrow for the info it displays).

src/components/templates/ExternalRoom/ExternalRoom.tsx
```diff
@@ -67,7 +67,12 @@ export const ExternalRoom: React.FC<ExternalRoomProps> = ({ venue }) => {
         <>
           <div className="ExternalRoom__message">
             <div>This page should automatically open</div>
-            <a rel="noreferrer" href={redirectUrl} target="_blank">
+            <a
+              rel="noreferrer"
+              href={redirectUrl}
+              target="_blank"
+              className="ExternalRoom__link"
+            >
```
goran-peoski-work, line 75:
```suggestion
              href={redirectUrl}
              className="ExternalRoom__link"
              {...getExtraLinkProps(true)}
            >
```

goran-peoski-work (changes requested):
Some visual consideration, deal with as you see fit or time allows https://github.com/sparkletown/sparkle/pull/2255#issuecomment-913657568

One change request though, `getExtraLinkProps()` isn't the best, nor best known, function, but can serve a purpose of easily looking up and changing external links @ project level.

src/components/templates/ExternalRoom/ExternalRoom.tsx
yurii-lubynets: nice catch! fixed in 0ee2698825ee60395a850a7d52819eb590cb92c8

goran-peoski-work (approved):
LGTM

PR #2262, opened 2021-09-06
Update entrance field check in VenuePage

goran-peoski-work:
Tries to resolve:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1088

by updating the check of field `entrance` from `isDefined()` to `Array.isArray(entrance) && entrance.length > 0`

Based on comment https://github.com/sparkletown/internal-sparkle-issues/issues/1088#issuecomment-913609828

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -15,7 +15,7 @@ import {
   isCurrentEventRequestedSelector,
   isCurrentVenueRequestedSelector,
 } from "utils/selectors";
-import { isDefined, isTruthy } from "utils/types";
+import { isDefined } from "utils/types";
 import { venueEntranceUrl } from "utils/url";
 import {
   clearLocationData,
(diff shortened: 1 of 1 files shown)


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -209,18 +208,17 @@ export const VenuePage: React.FC = () => {
   // if (isAccessDenied) {
   //   return <AccessDeniedModal venueId={venueId} venueName={venue.name} />;
   // }
+  const { entrance, template, hasPaidEvents } = venue;
+  const { enteredVenueIds } = profile;
+
+  const hasEntrance = Array.isArray(entrance) && entrance.length > 0;
```
goran-peoski-work, line 214: Entrance step will now only work if the field in DB is an array with at least one element.

goran-peoski-work (commented):
Initial PR comments

mike-lvov (approved):
LGTM

whoaitsaimz:
Did not fix issue. But we know the source of the issue now (an empty array in database), thanks for the detective work, team.

goran-peoski-work:
@whoaitsaimz You're right about the first part. This  PR didn't fix the cause, it's just prevention if we get in the same situation. It's a check when reading invalid data.

What we didn't discover is how the invalid data (empty array) was created. Might be manual DB change, might be some code of a form somewhere. Hopefully we'd not need to after this PR merge.

P.S. All the detective credit should go to @mike-lvov  who did 💯💪 of it.

goran-peoski-work:
@whoaitsaimz Just read https://github.com/sparkletown/internal-sparkle-issues/issues/1088#issuecomment-918896384

I stand corrected. I can safely say about 80% or more of the detective credit should have gone to you. Sorry @mike-lvov .

Agree with the conclusion, we shouldn't do more PRs for this problem.

PR #2266, opened 2021-09-07
fixed navbar stacking context

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1165

### DIFF
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -163,10 +163,10 @@ $loading-page-icon-transform-small-opacity: 0.4;
 // @debt the following locations are using styled-components, so can't use our helper function yet
 //   pages/Admin/Room/Modal/RoomModal.styles.ts (z-index: 101)
 //   components/atoms/Legend/Legend.styles.ts (z-index: 5)
-$z-layer-navbar: 31;
 $z-layer-live-schedule: 35;
 $z-layer-live-schedule-download: 25;
 $z-layer-sidebar: 30;
+$z-layer-navbar: $z-layer-sidebar + 1;
 $z-layer-left-column: 15;
 $z-layer-emergency-content: 15;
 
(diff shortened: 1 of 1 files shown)


src/scss/constants.scss
```diff
@@ -163,7 +163,7 @@ $loading-page-icon-transform-small-opacity: 0.4;
 // @debt the following locations are using styled-components, so can't use our helper function yet
 //   pages/Admin/Room/Modal/RoomModal.styles.ts (z-index: 101)
 //   components/atoms/Legend/Legend.styles.ts (z-index: 5)
-$z-layer-navbar: 30;
+$z-layer-navbar: 31;
```
yurii-lubynets, line 166: Well, the main reason why `.announcement-container-message` (announcement message) hides the `.nav-schedule .nav-schedule-clicked` (the < close button) is about their z-index. 
Right now they are equal (30) and absolutely positioned announcement message is higher. 

I've checked `$z-layer-navbar` is used only for `.navbar.navbar_playa` and it'd be a direct fix for the schedule to be higher than an announcement.

goran-peoski-work: [OOS] We might need better resolution to just updating numbers. Something like:
 `$z-layer-navbar:  $z-layer-announcement + 1;`

BTW, how hard is it to just yank the code from `RoomModal.styles.ts` and put it in `RoomModal.scss`?

goran-peoski-work: [OOS] We should also determine if we're going to put all z-index constants flat in `constants.scss` or have those values part of a single enum `z-layers` for consistency.

The first option might be more conducive to adding relative heights like the above OOS note

goran-peoski-work (approved):
LGTM with some notes for future improvements.

yurii-lubynets:
@goran-peoski-work thank you for your review. 

I've updated the announcement z-index as `$z-layer-navbar: $z-layer-sidebar + 1;` it looks much better [fixed in 42161887a642debdb6a6e035864b65dfc03b921e]. 
I suppose we are using all z-index from `constants.scss` and it's more conducive

whoaitsaimz:
This PR has only made it to burn-staging, not yet burn

PR #2267, opened 2021-09-07
add room labels to run tab in admin v3

denisdimitrov:
Run tab's bulk room edit wasn't showing the room labels and it was hard to figure out which room is which especially if they have the same logo. This now fixes the issue and displays the room title underneath the avatar.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1032

blablabla

### DIFF
--- a/src/pages/Account/Venue/VenueMapEdition/DraggableSubvenue.scss
+++ b/src/pages/Account/Venue/VenueMapEdition/DraggableSubvenue.scss
@@ -1,0 +1,18 @@
+@import "scss/constants";
+
+.DraggableSubvenue {
+  &__title {
+    position: absolute;
+    left: 50%;
+    transform: translateX(-50%);
+    display: flex;
+    align-content: center;
+    justify-content: center;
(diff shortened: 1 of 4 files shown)


src/pages/Account/Venue/VenueMapEdition/DraggableSubvenue.scss
```diff
@@ -0,0 +1,18 @@
...
+    left: 50%;
+    transform: translateX(-50%);
+    display: flex;
+    align-content: center;
+    justify-content: center;
+    background-color: opaque-black(0.5);
+    border-radius: $border-radius--xl;
+    padding: $spacing--xs $spacing--sm;
+    font-size: $font-size--sm;
+    transition: all 400ms $transition-function;
```
goran-peoski-work, line 15: [OOS] we might need to check all the project's transitions, see if there are few timings we can standardize around and put them in constants similar to `$transition-function` or maybe even as part of.

goran-peoski-work (approved):
LGTM

PR #2273, opened 2021-09-07
Fixes reload on every room entry

mike-lvov:
Fixes reload on every room entry

### DIFF
--- a/src/hooks/useRoom.ts
+++ b/src/hooks/useRoom.ts
@@ -1,4 +1,5 @@
 import { useCallback, useMemo } from "react";
+import { useHistory } from "react-router-dom";
 
 import { Room } from "types/rooms";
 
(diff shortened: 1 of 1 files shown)


sunny-viktoryia (approved):
mmmm nice. 😍 😍 😍 😍 😍 😍

src/hooks/useRoom.ts
```diff
@@ -21,6 +22,8 @@ export const useRoom = ({ room, venueName }: UseRoomProps) => {
 
   const roomUrl = room?.url ?? "";
 
+  const { push: openUrlUsingRouter } = useHistory();
```
goran-peoski-work, line 25: Could've named it `customOpenRelativeUrl`, to save us from reading ` customOpenRelativeUrl: openUrlUsingRouter` 😆

PR #2274, opened 2021-09-07
Move enteredVenueIds into UserLocation

dmitrymarchuk:
The goal of this PR is to reduce the number of unnecessary re-renders. The field `enteredVenueIds` changes frequently yet it resides inside user profile data alongside fields that are not changed frequently like username, links etc. So this PR:

- Moves `enteredVenueIds` field from `BaseUser` to `UserLocation` type.
- Adds deep comparison to `userWithoutLocation` inside `worldUsers::processUserDocChange` not to update user profile data if nothing actually changed. And since `enteredVenueIds` is moved to `userLocation` this reduces re-renders.
- Adds `console.warn` and `Bugsnag` notifications in `worldUsers.ts` various places and removed old @debt and TODO related comments.
- Removes unused schema types in `User.ts`


mike-lvov (approved):
LGTM

PR #2275, opened 2021-09-07
Add   title: values.title in RoomsForm

goran-peoski-work:
Quick fix for missing field due to PR regression


denisdimitrov (approved):
LGTM!

PR #2276, opened 2021-09-07
make title in subvenuemap optional

denisdimitrov:
The field is handled properly everywhere and it's being used only in 1 single place. This change will just make it optional.


goran-peoski-work (approved):
LGTM

PR #2279, opened 2021-09-08
add delete venue button to admin v3

denisdimitrov:
- Adds `Delete Space` button to admin v3 in the run tab
- Adds additional variant of ButtonNG called `danger`
- Replaces `venue` mentions with `space` in the delete modal.


src/components/atoms/ButtonNG/ButtonNG.scss
```diff
@@ -12,7 +12,8 @@ $variants-enum: (
   ("primary", $white, $primary--live, $primary--live-hover),
   ("secondary", $white, opaque-white(0.2), opaque-white(0.25)),
   ("white", $primary--live, $white, $white),
-  ("dark", $white, opaque-black(0.4), opaque-black(0.6))
+  ("dark", $white, opaque-black(0.4), opaque-black(0.6)),
+  ("danger", $white, $input--invalid, darken($input--invalid, 10%))
```
goran-peoski-work, line 16: Can these from `constants.scss` be used instead? 

```scss
$input--invalid: #dc3545;
$input--invalid-focus: #ff7384;
```

Unless there are updated values from Design...

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
```diff
@@ -45,7 +45,7 @@ export const RunTabView: React.FC<RunTabViewProps> = ({ venue }) => {
       </div>
       <div className="RunTabView__main">
         <div className="RunTabView__toolbar RunTabView--spacing">
-          <RunTabToolbar venueId={venueId} />
+          <RunTabToolbar venueId={venueId} venueName={venue.name} />
```
goran-peoski-work, line 48: Might as well just pass in `venue` instead separate `id` and `name`.
Otherwise `RunTabToolbar` can just use the `venueId` prop to get a hold of the venue itself.

src/pages/Admin/Admin.tsx
```diff
@@ -286,6 +286,10 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
   const match = useRouteMatch();
   const placementDivRef = useRef<HTMLDivElement>(null);
 
+  const navigateToAdmin = useCallback(() => {
+    history.push("/admin");
```
goran-peoski-work, line 290: Like there is `adminNGRootUrl`, one can be made for OG admin:

```suggestion
    history.push(adminOGRootUrl);
```

src/pages/Admin/Admin.tsx
```diff
@@ -385,8 +389,11 @@ const VenueInfoComponent: React.FC<VenueInfoComponentProps> = ({
       </div>
       <VenueDeleteModal
         show={isDeleteModalVisible}
+        onCancel={hideDeleteModal}
         onHide={hideDeleteModal}
-        venue={venue}
+        onDelete={navigateToAdmin}
+        venueId={venue.id}
+        venueName={venue.name}
```
goran-peoski-work, line 396: Similar to `RunTabToolbar` props. Why the need to pass separate `id` and `name`?  `venue={venue}` serves the need for both

src/pages/Admin/Venue/VenueDeleteModal.tsx
```diff
@@ -1,106 +1,64 @@
...
-          <>
-            <div className="input-group">
-              <span className="info">
-                WARNING: This action cannot be undone! Are you sure you wish to
-                delete {venue.name}?
-              </span>
+        {isDeleting && (
+          <div className="centered-flex">
+            <div className="spinner-border">
+              <span className="sr-only">Loading...</span>
```
goran-peoski-work, line 45: [OOS] Few of these `Loading...` divs can probably be extracted into existing or new component

src/pages/Admin/Venue/VenueDeleteModal.tsx
```diff
@@ -1,106 +1,64 @@
...
-              Close
-            </button>
           </div>
         )}
+        <div className="VenueDeleteModal__buttons">
+          <ButtonNG
+            disabled={isDeleting}
+            variant="danger"
+            onClick={deleteVenue}
+          >
```
goran-peoski-work, line 54:
```suggestion
            loading={isDeleting}
          >
```

src/pages/Admin/Venue/VenueDeleteModal.tsx
```diff
@@ -1,106 +1,64 @@
...
+            Yes, Delete
+          </ButtonNG>
+          <ButtonNG disabled={isDeleting} variant="primary" onClick={onCancel}>
+            No, Cancel
+          </ButtonNG>
+        </div>
+        {error && <span className="input-error">{error}</span>}
```
goran-peoski-work, line 61: Could use something like https://github.com/sparkletown/sparkle/tree/fix/create-venue/src/components/atoms/AutoSpan to streamline display of simple error messages.

Example:
```suggestion
        <AutoSpan variant="error">{error}</AutoSpan>
```

src/pages/Admin/Venue/VenueDeleteModal.tsx
```diff
@@ -1,106 +1,64 @@
...
 
-import { AnyVenue } from "types/venues";
-
-import { WithId } from "utils/id";
+import { ButtonNG } from "components/atoms/ButtonNG";
 
 import "./VenueDeleteModal.scss";
 
-interface PropsType {
-  venue: WithId<AnyVenue>;
```
goran-peoski-work, line 13: Why remove?

goran-peoski-work (changes requested):
Few comments

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
denisdimitrov: I don't want to pass the whole venue object because we need only 2 fields from that object and it's unnecessary.

I can use the `venueId` and get a hold of the venue if that's the approach you prefer.

src/pages/Admin/Admin.tsx
denisdimitrov: Because we don't need the whole venue, you pass only the props you need. I will save 1 line of code.

src/pages/Admin/Venue/VenueDeleteModal.tsx
denisdimitrov: https://github.com/sparkletown/sparkle/pull/2279#discussion_r705137611

src/pages/Admin/Admin.tsx
denisdimitrov: React's power is in the declarativity which makes it extremely readable.

This is much more readable and you immediately understand which fields from the venue are being used inside the component before you even open it and look at the logic inside.
```tsx
<VenueModal venueName={venue.name} venueDescription={venue.description} />
```


These two are voids of logic and you have no idea what's being passed down, you need to dive in and figure it out.
```tsx
<VenueModal venue={venue} />
```
```tsx
<VenueModal {...venue} />
```

src/components/atoms/ButtonNG/ButtonNG.scss
denisdimitrov: `$input--invalid` is already being used, the `darken` is just used for darker hover color, it's not per design, because the design doesn't have hover colour.

src/pages/Admin/Admin.tsx
denisdimitrov: + if anything changes in `venue` the component will re-render.

src/pages/Admin/Venue/VenueDeleteModal.tsx
denisdimitrov: We actually have this in a reusable component `InputField` but sadly this is not a form and we can't use it. Another option is `ErrorMessage` from `react-hook-form`

src/pages/Admin/Venue/VenueDeleteModal.tsx
denisdimitrov: There is already a spinner above the buttons and it looks weird when I add this flag to the button because there's 2 spinners on the screen. :/

src/pages/Admin/Admin.tsx
goran-peoski-work: > This is much more readable and you immediately understand which fields from the venue are being used inside the component before you even open it and look at the logic inside.

Why is the need to understand what an encapsulation does? 
Isn't the purpose of encapsulation to rid you of some concerns?

Let's imagine what `VenueInfoComponent` might say to a `VenueModal`:

1. > Here is my venue, now do what you need with it

vs

2. > Here are only 2 things, and I will not even tell you if they're from the venue, because I know what you're doing with those 2 fiel....

Just doesn't make sense 🤷 

Might be read differently if it was a `Modal`, instead of `VenueModal`, maybe it's only about a name change.

A component like that is expected to be generic, visual only without logic, so giving it only a `text` and/or just `onClick` would be more like the second approach.

src/components/atoms/ButtonNG/ButtonNG.scss
goran-peoski-work: `$input--invalid-focus` looks like it's by some earlier design, even if maybe not explicit for button.

If still going with modification of `$input--invalid` maybe lighten instead of darken, since the primary has lighter color on hover...

🤷 But doesn't really matter... Your judgement is as good as mine. Kind of unfortunate can't ping design here with @ due to missing group.

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
goran-peoski-work: Not really about preference.

Do you think the toolbar might need more info than id and name?  Maybe less?

Not even sure it needs the name only to pass it down to a component inside. This kind of tight props coupling is harder to maintain.

src/pages/Admin/Admin.tsx
goran-peoski-work: New PR is a good idea, on top of these constants, there might be a need to switch between OG/NG once v3 achieves v1 parity

src/pages/Admin/Venue/VenueDeleteModal.tsx
goran-peoski-work: Any rule saying we should keep the original spinner? 😄 Probably best left as you've made it.

Which reminds me, the button supports `aria-` props, so it will be prudent to check which might apply as we add/replace.

[OOS]
`sr-only` class should probably be added on ButtonNG for the loading state

src/pages/Admin/Venue/VenueDeleteModal.tsx
goran-peoski-work: @denisdimitrov whichever you think best eliminates the need of `input-error` class which eventually might help in reducing the legacy baggage of `globals.scss`

denisdimitrov: This is old legacy component and I don't want to waste time and resources on this. There are probably 50 instances of this className and it would be much easier to fix it in a single PR

src/pages/Admin/Admin.tsx
denisdimitrov: This component is specific for venue deletion and it's not an atom it's a molecule, it's concrete and makes sense to improve the readability and understand what props it requires.

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
denisdimitrov: Probably not, I don't see it using anything else from the venue object.

src/components/atoms/ButtonNG/ButtonNG.scss
denisdimitrov: Makes sense, i'll lighten it instead.

src/pages/Admin/Admin.tsx
denisdimitrov: > Why is the need to understand what an encapsulation does?

Because that's one of the benefits of React and it's a standard/good practice. The declarativity is powerful and you immediately understand what props are passed down.

src/pages/Admin/Venue/VenueDeleteModal.tsx
denisdimitrov: > Probably best left as you've made it.

I'll leave it as it is for now, but it can be improved, yes.

> Any rule saying we should keep the original spinner?

Not really, we can remove it but there is no design for this...

src/pages/Admin/Venue/VenueDeleteModal.tsx
goran-peoski-work: Well, a single PR is one way to go about it, provided it doesn't require widely differing approaches to replacing it and it doesn't cause merge issues to large number of PRs touching any of those 50 places.

src/pages/Admin/Admin.tsx
goran-peoski-work: I'm not debating about the existential questions of React.

I'm noting of the current PR code being between two equally valid opposing ends:

1. Have only `venueId` passed down and have their respective components determine if they need more by hook
2. Have only text and callback passed down and have the child components be unaware of venues at all.

Right now the components are half way neither way.

src/pages/Admin/Venue/VenueDeleteModal.tsx
```diff
@@ -1,106 +1,61 @@
...
-
-import { WithId } from "utils/id";
+import { ButtonNG } from "components/atoms/ButtonNG";
+import { LoadingSpinner } from "components/atoms/LoadingSpinner";
 
 import "./VenueDeleteModal.scss";
 
-interface PropsType {
-  venue: WithId<AnyVenue>;
+interface VenueDeleteModalProps {
```
goran-peoski-work, line 11:
```suggestion
export interface VenueDeleteModalProps {
```

goran-peoski-work (approved):
Approving with a caveat https://github.com/sparkletown/sparkle/pull/2279/files#r708273891

Code can be made in line with 2 different approaches or left in the middle of the road between them.

PR #2281, opened 2021-09-09
Update Counting is shown -1 difference between DMs and "Online" 

ermiaSparkle:
- closes https://github.com/sparkletown/internal-sparkle-issues/issues/1043


yurii-lubynets (changes requested):
Well, `src/components/molecules/VenuePartygoers/VenuePartygoers.tsx` uses `recentWorldUsers` to count online users number.
`src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx` uses `useOnlineUsersToDisplay` hook to count online users number. And actually it filters out self (in `PrivateChats` you are adding yourself again in `userItself`). 

It'd be great to unify this calculation and to use it across the platform

yurii-lubynets (commented):
LGTM

It's needed one more review to make sure we're aligned with the code standards

goran-peoski-work:
> LGTM
> 
> It's needed one more review to make sure we're aligned with the code standards

I can suggest you check this https://github.com/sparkletown/sparkle/pull/2281#pullrequestreview-750444841 with @mike-lvov . Might be small enough change to put in this PR or open a follow up. Either way, useful to have an answer of which function to use.

goran-peoski-work:
@yurii-lubynets @ermiaSparkle code wise it looks straightforward. logic wise... is there a use case when user is set to appear offline but is still discounted from the total?

PR #2282, opened 2021-09-09
Fix logic of download self schedule btn

ermiaSparkle:
- closes https://github.com/sparkletown/internal-sparkle-issues/issues/1076


src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -26,6 +26,7 @@ $schedule-btn--height: 51px;
     background-color: opaque-black(0.9);
     height: $schedule-btn--height;
     padding-left: $NavBarSchedule--padding;
+    transition: all 400ms $transition-function;
```
sunny-viktoryia, line 29: Thanks for that. Would be even better, is we had an extra class name for open buttons: { "NavBarWrapper__download-buttons--show": isVisible }, and give them the same `transform` properties as `.NavBarSchedule` has (to make them come together). 🙃

sunny-viktoryia (approved):
Thanks! The fix is working. 🙃 
Can I ask for a style fix as well? Check the comment, please.

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
ermiaSparkle: fixed in c1e15fe93015c6a35048aac1213b8eddf3347c36

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -26,6 +26,10 @@ $schedule-btn--height: 51px;
     background-color: opaque-black(0.9);
     height: $schedule-btn--height;
     padding-left: $NavBarSchedule--padding;
+
+    &--show {
+      transition: all 400ms $transition-function;
+    }
```
sunny-viktoryia, line 32: I meant a bit different changes.
```suggestion
    transition: all 400ms $transition-function;
    transform: translateX(-40%);

    &--show {
      transform: translateX(0);
    }
```

sunny-viktoryia: but instead of 40%, can you use a variable, that is already there, please?
`$NavBarSchedule--width`

ermiaSparkle: fixed in b29eef631b102c46b08709e69b4377cce1553125

src/components/organisms/NavBarSchedule/NavBarSchedule.scss
```diff
@@ -16,16 +16,22 @@ $schedule-btn--height: 51px;
...
     display: flex;
     align-items: center;
     justify-content: flex-end;
     background-color: opaque-black(0.9);
     height: $schedule-btn--height;
     padding-left: $NavBarSchedule--padding;
+    transform: translateX(-$NavBarSchedule--width);
+
+    &--show {
+      transition: all 400ms $transition-function;
```
sunny-viktoryia, line 32: please, move `transition` from `&--show` to a `&__download-buttons`. It'll make the transition smoother.

sunny-viktoryia (commented):
a couple of small comments. 🙃

PR #2283, opened 2021-09-09
feature flag for turning off reactions in auditorium

dmitrii-zolotukhin:
- [x] Fixed turning on reactions
close sparkletown/internal-sparkle-issues#1170

### DIFF
--- a/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
+++ b/src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
@@ -76,6 +76,12 @@ const LoginForm: React.FunctionComponent<LoginFormProps> = ({
           "validation",
           `Error checking ticket: ${error.message}`
         );
+      } else if (error.code === "auth/user-not-found") {
+        setError(
+          "backend",
+          "firebase",
+          "Account not found. Please contact the event organizer."
+        );
       } else {
         setError("backend", "firebase", error.message);
       }
--- a/src/components/templates/Audience/Audience.tsx
+++ b/src/components/templates/Audience/Audience.tsx
@@ -28,6 +28,7 @@ import { isDefined } from "utils/types";
 
 import { useDispatch } from "hooks/useDispatch";
 import { useRecentVenueUsers } from "hooks/users";
+import { useSettings } from "hooks/useSettings";
 import { useShowHide } from "hooks/useShowHide";
 import { useUser } from "hooks/useUser";
 
@@ -143,6 +144,7 @@ export const Audience: React.FC<AudienceProps> = ({ venue }) => {
   const baseRows = venue?.auditoriumRows ?? DEFAULT_AUDIENCE_ROWS_NUMBER;
 
   const { isShown: isUserAudioOn, toggle: toggleUserAudio } = useShowHide(true);
+  const { isLoaded: settingsAreLoaded, settings } = useSettings();
 
   const isUserAudioMuted = !isUserAudioOn;
 
@@ -317,6 +319,8 @@ export const Audience: React.FC<AudienceProps> = ({ venue }) => {
       seated: userSeated,
     });
 
+    const shouldShowReactions = settingsAreLoaded && settings.showReactions;
+
     // @debt This should probably be all rolled up into a single canonical component for emoji reactions/etc
     const renderReactionsContainer = () => (
       <>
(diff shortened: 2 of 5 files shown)


src/components/templates/Audience/Audience.tsx
```diff
@@ -143,6 +144,7 @@ export const Audience: React.FC<AudienceProps> = ({ venue }) => {
   const baseRows = venue?.auditoriumRows ?? DEFAULT_AUDIENCE_ROWS_NUMBER;
 
   const { isShown: isUserAudioOn, toggle: toggleUserAudio } = useShowHide(true);
+  const { isLoaded: settingsAreLoaded, settings } = useSettings();
```
denisdimitrov, line 147: `useSettings` is setting a global state in the `customizations` document which will be used for all venues on the platform. Is this behaviour expected?

goran-peoski-work: `settingsAreLoaded` -> `areSettingsLoaded`

src/components/templates/Audience/Audience.tsx
```diff
@@ -317,6 +319,8 @@ export const Audience: React.FC<AudienceProps> = ({ venue }) => {
       seated: userSeated,
     });
 
+    const shouldShowReactions = settingsAreLoaded && settings.showReactions;
```
goran-peoski-work, line 322: if settings are not loaded, this defaults `showReactions` to `false`, while so far it has been `true`. An explicit default should be used i.e.

```suggestion
    const shouldShowReactions = areSettingsLoaded ? settings.showReactions : DEFAULT_SHOW_REACTIONS;
```

goran-peoski-work (changes requested):
Logic changes from before flag is introduced. For backwards compatibility / keeping the same functionalities, when no settings are loaded, flag should default to the one before this change.

src/components/templates/Audience/Audience.tsx
sunny-viktoryia: @goran-peoski-work, I believe the logic is correct here. We need to wait until the settings are loaded, and then decide if we want to show the reactions. The default value (`DEFAULT_SHOW_REACTIONS`) is already used in the `useSettings` hook, so if nothing is specified there, DEFAULT_SHOW_REACTIONS will play. 

It's done this way to avoid flickering. We'll better show reactions a second later but by then we'll be sure that we want them to be shown.

src/components/templates/Audience/Audience.tsx
sunny-viktoryia: @denisdimitrov, yes. Our aim is to be able to hide the features that can be a bottleneck for us in terms of performance. It's done this way to be able to turn off the reactions in firebase for everyone.

sunny-viktoryia (commented):
Looks good, but we need to hide reactions in a new auditorium too.
Can you check the Sections component too, please? 🙃

src/components/templates/Audience/Audience.tsx
goran-peoski-work: > We need to wait until the settings are loaded,

Why the need?

> default value (DEFAULT_SHOW_REACTIONS) is already used in the useSettings hook, so if nothing is specified there, DEFAULT_SHOW_REACTIONS will play

@sunny-viktoryia  This means no check should be done if the settings are loaded since loaded or not, the default is provided and shouldn't be overriden by the `isLoaded` value.

src/components/templates/Audience/Audience.tsx
dmitrii-zolotukhin: fixed in 2bf823d11ad27c3992680c9ac6a4585293744884

src/components/templates/Audience/Audience.tsx
dmitrii-zolotukhin: fixed in 2bf823d11ad27c3992680c9ac6a4585293744884

src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -76,6 +76,12 @@ const LoginForm: React.FunctionComponent<LoginFormProps> = ({
           "validation",
           `Error checking ticket: ${error.message}`
         );
+      } else if (error.code === "auth/user-not-found") {
+        setError(
+          "backend",
+          "firebase",
+          "Account not found. Please contact the event organizer."
+        );
```
goran-peoski-work, line 84: Is this code misplaced from https://github.com/sparkletown/sparkle/pull/2294 ?

sunny-viktoryia:
@dmitrii-zolotukhin, the name of the PR and the attached issue are not related to actual changes. Was something mixed up here? Please, fix.

PR #2284, opened 2021-09-09
Feature/Admin v3 - Event start & end time

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/700


src/hooks/useVenueScheduleEvents.ts
```diff
@@ -99,8 +99,9 @@ const useVenueScheduleEvents = ({
     [liveAndFutureEvents, firstRangeDateInSeconds]
   );
 
+  const endScheduleDate = sovereignVenue?.end_utc_seconds;
   const daysInBetween = differenceInDays(
-    fromUnixTime(maxDate),
+    fromUnixTime(endScheduleDate || maxDate),
```
yurii-lubynets, line 104: > (currently it shows you a week from the current time, but the thinking is that the schedule will only show dates within the start and end time of the event)

https://github.com/sparkletown/internal-sparkle-issues/issues/700#issue-878702427

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,58 @@
...
+          name={`${name}_date`}
+          className="DateTimeField__container--date"
+          value={dateValue}
+          onChange={handleDateChange}
+        />
+        <input
+          type="time"
+          name={`${name}_time`}
+          value={timeValue}
+          onChange={handleTimeChange}
```
goran-peoski-work, line 53: Since these `onChange` handlers are used, I'm not sure even `name` as a prop is really needed, not the least elaborate one made with prefix. Component should probably work great with out those.

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { getDateTimeFromUtc, getUtcFromDateTime } from "utils/time";
+
+import "./DateTimeField.scss";
+
+export interface DateFieldProps {
+  title: string;
+  subTitle?: string;
+  name: string;
+  dateTimeValue?: number;
+  handleDateTimeChange: (args: number) => void;
```
goran-peoski-work, line 12: Since this is the public interface of the component, something more in line with common React style

```suggestion
  onChange: (value: number) => void;
```

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  &__subtitle {
+    font-weight: $font-weight--300;
+    color: opaque-white(0.6);
+    margin: 0;
+  }
+
+  &__container {
+    display: flex;
+    margin-top: $spacing--md;
```
goran-peoski-work, line 16: For some reason there are different margins that make the container go out of the left sidebar. I managed to fix it with 

```scss
  padding: 1em;
```

**Change**:
sparkle-padding-03

**Before**:
sparkle-padding-02

**After**:
sparkle-padding-01

src/components/organisms/AdminVenueView/components/Timing/Timing.scss
```diff
@@ -33,9 +33,13 @@ $left-section-width: 360px;
   }
```
goran-peoski-work, line 33: `.Timing` could probably use

`align-items: start;`

so the events table doesn't leave wide gap from the top of its cell.

goran-peoski-work (changes requested):
Mostly sound logic with some HTML/SCSS/BEM related notes.

src/components/organisms/AdminVenueView/components/Timing/Timing.scss
yurii-lubynets: nice catch! fixed in ecdf3dcd597340a1deea0d0f04f0bc2037c2dc65

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.scss
yurii-lubynets: fixed in ecdf3dcd597340a1deea0d0f04f0bc2037c2dc65

src/components/organisms/AdminVenueView/components/Timing/Timing.scss
```diff
@@ -33,9 +33,13 @@ $left-section-width: 360px;
...
 
   &__left {
-    &-content {
+    &--content {
       padding: 40px 30px 100px;
     }
+
+    &--header {
+      margin-bottom: 0;
+    }
```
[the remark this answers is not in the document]
yurii-lubynets: Got it. I was trying to be aligned with the existing file naming and now it looks more BEM.
fixed in ecdf3dcd597340a1deea0d0f04f0bc2037c2dc65

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.scss
goran-peoski-work: LOL, wrong place for the above comment. Fixed in `Timing.scss`

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
yurii-lubynets: You are right. removed `name` prop in 7de26631ccd91b927fc8b788c5689419ef221390

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    const dateTime = getUtcFromDateTime(`${dateValue} ${timeValue}`);
+
+    handleDateTimeChange(dateTime);
+  }, [dateValue, timeValue, handleDateTimeChange]);
+
+  return (
+    <>
+      <label className="DateTimeField__title" htmlFor={`${name}_date`}>
```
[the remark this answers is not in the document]
yurii-lubynets: Great idea. I've added `generateId` into `src/utils/id.ts` as a helper and updated `<DateTimeField/>`.
fixed in a498ad26f53ecd6235b397be872656275c9420a0

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const handleTimeChange = useCallback((e) => setTimeValue(e.target.value), []);
+
+  useEffect(() => {
+    const dateTime = getUtcFromDateTime(`${dateValue} ${timeValue}`);
+
+    onChange(dateTime);
+  }, [dateValue, timeValue, onChange]);
+
+  return (
+    <label className="DateTimeField__title" id={generateId(name)}>
```
goran-peoski-work, line 37: The pairing goes like this
- `<label htmlFor= ...`
- `<input id=...`

but that's only there for cases when `label` doesn't include `input` as child. 

Since this is the case, there is no longer need of any `id` prop.
```suggestion
    <label className="DateTimeField__title">
```

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useCallback, useEffect, useState } from "react";
+
+import { generateId } from "utils/id";
```
goran-peoski-work, line 3:
```suggestion
```

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
goran-peoski-work: From MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for):

>Note: A <label> element can have both a for attribute and a contained control element, as long as the for attribute points to the contained control element.

goran-peoski-work: Unfortunately, per https://stackoverflow.com/a/9004357, it turns out that only a single `input` can be associated to a `label` so it might be better to change it to `div`, and if necessary, wrap both `input` tags with their own `label` tag, or have them connected with 2 separate `id`/`htmlFor` pairs or follow the SO answer for an alternative

goran-peoski-work (commented):
I guess the only remaining issue is presenting the two different input fields. I have included links in the comments from MDN and SO that might help out.

src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
yurii-lubynets: thank you a lot for detailed explanation. Removed `id` and set Label as a parent for this input

yurii-lubynets: fixed in b16a405ab6403c476874cd1cb2bab1f3107abdba

goran-peoski-work:
LGTM

goran-peoski-work (approved):

PR #2287, opened 2021-09-09
Embed function refactoring

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1110
@goran-peoski-work tagging you just in case


src/utils/ConvertToEmbeddableUrl.tsx
```diff
@@ -1,5 +1,12 @@
...
-
+import {
+  FACEBOOK_EMBED_URL,
+  TWITCH_EMBED_URL,
+  TWITCH_SHORT_URL,
+  VIMEO_EMBED_URL,
+  VIMEO_SHORT_EVENT_URL,
+  YOUTUBE_EMBED_URL,
+  YOUTUBE_SHORT_URL_STRING,
+} from "../settings";
```
goran-peoski-work, line 9:
```suggestion
} from "settings";
```

src/utils/ConvertToEmbeddableUrl.tsx
```diff
@@ -8,28 +15,55 @@ export interface ConvertToEmbeddableUrlOptions {
 // @debt Replace the naive logic with use of new URL(url) which provides standard parsing before doing checks
```
goran-peoski-work, line 15:
```suggestion
```

src/utils/ConvertToEmbeddableUrl.tsx
```diff
@@ -8,28 +15,55 @@ export interface ConvertToEmbeddableUrlOptions {
...
   options: ConvertToEmbeddableUrlOptions
-) => string = ({ url, autoPlay = DEFAULT_VENUE_AUTOPLAY }) => {
-  if (url?.includes("youtube")) {
-    url = url?.replace("watch?v=", "embed/");
-  } else if (
-    url?.includes("vimeo") &&
-    !url?.includes("player") &&
-    // NOTE: If you have a scheduled live event, it gives you a different embed code
-    !url?.includes("vimeo.com/event")
+) => string = ({ url, autoPlay }) => {
```
goran-peoski-work, line 18: Can use destructuring to rename it internally `url: urlString` so it is made more clear later on:

```ts
  const urlObject = new URL(urlString);
```

or just

```ts
  const url = new URL(urlString);
```

src/utils/ConvertToEmbeddableUrl.tsx
```diff
@@ -8,28 +15,55 @@ export interface ConvertToEmbeddableUrlOptions {
...
+    return "";
+  }
+
+  let urlObj = new URL(url);
+  const { host, searchParams, pathname } = urlObj;
+  const youtubeVideoId = searchParams.get("v") || pathname.slice(1);
+  const twitchVideoId = searchParams.get("video");
+  const twitchId = searchParams.get("channel") || twitchVideoId;
+  const isTwitch = host.includes(TWITCH_SHORT_URL);
+
```
goran-peoski-work, line 29: Moving 
```ts
  if (autoPlay) {
    const autoPlayValue = isTwitch ? "true" : "1";
    url.searchParams.set("autoplay", autoPlayValue);
  } else {
    url.searchParams.delete("autoplay");
  }
```
before the other checks will help reducing the `if-else-if` nesting by fail fast i.e. just `return url;` inside the `if`

goran-peoski-work: Better yet, extract a helper function, then specialized delegate functions per case:

```ts

const withAutoPlay = (url: URL, autoPlay: "1" | "true" | false | undefined) => {
  if (autoPlay) {
    url.searchParams.set("autoplay", autoPlay);
  }
  return url.href;
};

type ConvertTwitchUrlOptions = {
  pathname: string;
  twitchId?: string;
  twitchVideoId?: string;
  autoPlay?: boolean;
};
const convertTwitchUrl: (options: ConvertTwitchUrlOptions) => string = ({
  pathname,
  twitchId,
  twitchVideoId,
  autoPlay,
}) => {
  const id = pathname.split("/")?.slice(-1)?.[0] || twitchId || "";
  const twitchUrl = new URL(TWITCH_EMBED_URL);

  if (pathname.includes("video") || twitchVideoId) {
    twitchUrl.searchParams.set("video", id);
  } else {
    twitchUrl.searchParams.set("channel", id);
  }

  twitchUrl.searchParams.set("parent", "localhost");

  return withAutoPlay(twitchUrl, autoPlay && "true");
};```

This way `convertToEmbeddableUrl()` becomes a facade that just checks which service it is called about and returns the results from the helper functions

**Review by goran-peoski-work — CHANGES_REQUESTED** [review]
So far so good. `convertToEmbeddableUrl` is  letting the browser do the parsing and conversion. It can use a bit of refactor, separation of concerns by extracting the separate use cases into their own utility functions

File: src/utils/ConvertToEmbeddableUrl.tsx [hunk 5]
```diff
@@ -8,28 +15,55 @@ export interface ConvertToEmbeddableUrlOptions {
...
+  if (
+    host?.includes(YOUTUBE_SHORT_URL_STRING) &&
+    !pathname?.includes("embed")
   ) {
-    url = url?.replace("vimeo.com/", "player.vimeo.com/video/");
+    urlObj = new URL(`${YOUTUBE_EMBED_URL}${youtubeVideoId}`);
+    const params = searchParams.toString().split("&");
```
goran-peoski-work, line 35: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams no need for `.toString()` there is `.getAll()` and similar accessors on the search parameters object

alexey-petrov: `URLSearchParams.getAll(name)` requires a `name` prop to be passed and returns all mentions of that same prop
In this piece of code I'm extracting all props from the parameters
I've been looking for an api method that would do the same but without this workaround, but failed to find one :(

src/utils/ConvertToEmbeddableUrl.tsx
alexey-petrov: thanks, fixed in ea1c3b0

src/utils/ConvertToEmbeddableUrl.tsx
alexey-petrov: fixed ea1c3b0

src/utils/ConvertToEmbeddableUrl.tsx
alexey-petrov: fixed ea1c3b0

src/utils/ConvertToEmbeddableUrl.tsx
alexey-petrov: fixed ea1c3b0

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
...
+  YOUTUBE_EMBED_URL,
+  YOUTUBE_SHORT_URL_STRING,
+} from "settings";
+
+const setParamsIteratively = (urlObject: URL, urlParams?: string[]) => {
+  urlParams?.forEach((param) => {
+    const [key, value] = param.split("=");
+    urlObject.searchParams.set(key, value);
+  });
+};
```
goran-peoski-work, line 16: I think object of the type https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams with utility methods can serve better than manually dealing with `string[]`

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
...
+    return "";
+  }
+
+  let urlObject = new URL(urlString);
+  const { host, searchParams, pathname } = urlObject;
+  const youtubeVideoId = searchParams.get("v") || pathname.slice(1);
+  const youtubeStartTime =
+    searchParams.get("start") || searchParams.get("t") || "0";
+  const twitchVideoId = searchParams.get("video");
+  const urlParameters = searchParams.toString().split("&");
```
goran-peoski-work, line 78: I think object of the type https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams with utility methods can serve better than manually dealing with `string[]`

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
...
+  FACEBOOK_EMBED_URL,
+  TWITCH_EMBED_URL,
+  TWITCH_SHORT_URL,
+  VIMEO_EMBED_URL,
+  VIMEO_SHORT_EVENT_URL,
+  YOUTUBE_EMBED_URL,
+  YOUTUBE_SHORT_URL_STRING,
+} from "settings";
+
+const setParamsIteratively = (urlObject: URL, urlParams?: string[]) => {
```
goran-peoski-work, line 11: `setParamsIteratively` -> `withParameters` kind of like `withAutoPlay`

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
+import {
```
goran-peoski-work, line 1: Why the rename from `.ts` to `.tsx`?

This utility file doesn't seem to contain any TSX syntax

tsconfig.json
```diff
@@ -23,5 +23,5 @@
   },
   "include": [
     "src"
-  ]
+, "src/utils/.tsx"  ]
```
goran-peoski-work, line 26: Why is this line added?

goran-peoski-work (changes requested):
I like the name change of the file, just puzzled by the extension change.

`convertToEmbeddableUrl` looks cleaner and more straight forward to follow the logic in it. 
More separation can be done though so that the single responsibility of it could be to delegate per different service to appropriate helpers. This may help to easy add/modify in future change requests.

There might be a possibility of re-using `URLSearchParams` for additional parsing, if possible, better leave it to the browser deal with corner cases we haven't predicted.

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
...
+    host?.includes("vimeo") &&
+    !host?.includes("player") &&
+    // NOTE: If you have a scheduled live event, it gives you a different embed code
+    !urlString?.includes(VIMEO_SHORT_EVENT_URL)
+  ) {
+    urlObject = new URL(`${VIMEO_EMBED_URL}${pathname}`);
+  }
+
+  if (pathname?.includes(FACEBOOK_EMBED_URL)) {
+    searchParams.set("mute", "0");
```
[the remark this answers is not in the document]
alexey-petrov: we don't really have an option to mute videos in venues and i'm not really certain why we had this mute option for facebook videos
also, if we introduce mute here, then users won't be able to unmute their videos ( youtube for sure ) as volume level set in embed ui would be overwritten with the parameter

tsconfig.json
alexey-petrov: not sure, didn't manually add it.
removed in 741d64c, but had to add `downlevelIteration` because we're now using iterators `urlParams.entries()`

src/utils/embeddableUrl.tsx
alexey-petrov: done 741d64c

src/utils/embeddableUrl.tsx
alexey-petrov: renamed it back to `.ts` 741d64c

src/utils/embeddableUrl.tsx
alexey-petrov: used `.entries()` there 741d64c

src/utils/embeddableUrl.tsx
alexey-petrov: done in 741d64c

src/utils/embeddableUrl.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  withAutoPlay({ urlObject, autoPlay: autoPlay && urlAutoPlayValue });
```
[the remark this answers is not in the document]
alexey-petrov: changed func order in 741d64c

tsconfig.json
goran-peoski-work: `downlevelIteration`...  🤔 Hm, interesting. Weird how there isn't a copy support on that object, but still good job 👍

src/utils/embeddableUrl.tsx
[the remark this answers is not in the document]
goran-peoski-work: I meant only as a parameter to the function, not really an issue. If there is a need in the future, it can be added

goran-peoski-work (approved):
LGTM

PR #2288, opened 2021-09-10
Removes useConnect for all venues

mike-lvov:
Removes useConnect for all venues.
It seems to be the leftovers from the introduction of `useRelatedVenues`


goran-peoski-work (approved):
LGTM

PR #2292, opened 2021-09-10
Follow up to "Move enteredVenueIds into UserLocation"

dmitrymarchuk:
Follows up https://github.com/sparkletown/sparkle/pull/2274

- Fixes a situation when `enteredVenueIds` is initially empty and then is never populated.
- Adds `enteredVenueIds?: never` to type `User`
- Updates `useUser` hook to return user location and removed `useWorldUserLocation` from `VenuePage`

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -32,7 +32,6 @@ import useConnectCurrentVenue from "hooks/useConnectCurrentVenue";
 import { useInterval } from "hooks/useInterval";
 import { useMixpanel } from "hooks/useMixpanel";
 import { usePreloadAssets } from "hooks/usePreloadAssets";
-import { useWorldUserLocation } from "hooks/users";
 import { useSelector } from "hooks/useSelector";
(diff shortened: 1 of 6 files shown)


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -71,8 +70,8 @@ export const VenuePage: React.FC = () => {
 
   // const [isAccessDenied, setIsAccessDenied] = useState(false);
 
-  const { user, profile } = useUser();
-  const { userLocation } = useWorldUserLocation(user?.uid);
+  const { user, profile, userLocation } = useUser();
+  console.log(userLocation);
```
mike-lvov, line 74: Remove

mike-lvov (approved):
ltgm

src/pages/VenuePage/VenuePage.tsx
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2292/commits/18636351f55c7dccea7213907110d8a0cec1d106

PR #2295, opened 2021-09-13
add create world concept and add worldId to newly created venues

denisdimitrov:
When creating a new space from Admin ng, a new world will be created and the worldId will be passed to the new space.

The PR also adds a bunch of new functions for world creation, update and delete. (Only create is being used now)


goran-peoski-work (changes requested):
Few comments, most notable the use of `.then().catch()` in cloud functions

functions/world.js
```diff
@@ -0,0 +1,141 @@
...
+
+      throw new HttpsError("permission-denied", `User is not an admin`);
+    })
+    .catch((err) => {
+      throw new HttpsError(
+        "internal",
+        `Error occurred obtaining world ${worldId}: ${err.toString()}`
+      );
+    });
+};
```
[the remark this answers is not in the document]
denisdimitrov: > Already async function, why .then().catch()?

For easier error handling, nothing more.

> Will the inner throw new HttpsError throws get masked by the latter throw new HttpsError( "internal" and is this the intended functionality?

It's not going to get masked, after I throw the error in `.then()` the `.catch()` will trigger and throw another 'proper' one with the old as a string: `${err.toString()}`

functions/world.js
```diff
@@ -0,0 +1,141 @@
...
+        `User is not an owner of ${worldId}`
+      );
+    })
+    .catch((err) => {
+      throw new HttpsError(
+        "internal",
+        `Error occurred obtaining world ${worldId}: ${err.toString()}`
+      );
+    });
+};
```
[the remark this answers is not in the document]
denisdimitrov: https://github.com/sparkletown/sparkle/pull/2295#discussion_r707394912

functions/world.js
```diff
@@ -0,0 +1,141 @@
...
+  await checkIsWorldOwner(worldId, context.auth.token.user_id);
+  await checkIsAdmin(context.auth.token.user_id);
+
+  const worldData = {
+    updatedAt: Date.now(),
+    ...(data.bannerImageUrl && { host: { icon: data.logoImageUrl } }),
+    ...(data.logoImageUrl && { host: { icon: data.logoImageUrl } }),
+    ...(data.rooms && { rooms: data.rooms }),
+    ...(data.code_of_conduct_questions && {
+      code_of_conduct_questions: data.code_of_conduct_questions,
```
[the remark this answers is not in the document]
denisdimitrov: These are all of the 'basic' required fields for the world, we should add user statuses as well but it will be in the future.

src/pages/Admin/Details/Form/DetailsForm.tsx
```diff
@@ -38,19 +33,19 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ dispatch, editData }) => {
...
-        else await createVenue_v2(vals as VenueInput_v2, user);
+      const world = { ...vals, id: createUrlSafeName(vals.name) };
 
-        if (vals.name) {
-          history.push(`/admin-ng/venue/${createUrlSafeName(vals.name)}`);
-        } else {
+      try {
+        if (venueId) {
+          await updateVenue_v2(world, user);
           history.push("/admin-ng");
```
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/23ab4abba1ebb0c004ae645cd979bbb571ae1105

src/pages/Admin/Details/Form/DetailsForm.tsx
```diff
@@ -38,19 +33,19 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ dispatch, editData }) => {
...
+        } else {
+          await createWorld(world, user);
+          history.push(`/admin-ng/venue/${world.id}`);
```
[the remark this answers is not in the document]
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/23ab4abba1ebb0c004ae645cd979bbb571ae1105

functions/world.js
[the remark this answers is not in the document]
denisdimitrov: Then catch have been remove and trycatch is used now
Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/23ab4abba1ebb0c004ae645cd979bbb571ae1105

functions/world.js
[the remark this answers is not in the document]
denisdimitrov: Then catch have been remove and trycatch is used now
Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/23ab4abba1ebb0c004ae645cd979bbb571ae1105

functions/world.js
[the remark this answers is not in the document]
mike-lvov: @goran-peoski-work in the current iteration they are just copied over. Eventually, we'll change the naming to be consistent + to be more precise.

src/api/admin.ts
```diff
@@ -129,6 +129,20 @@ export interface VenueInput_v2
   end_utc_seconds?: number;
 }
 
+export interface WorldProps {
+  id: string;
```
mike-lvov, line 133: I don't think the id should be here by default because it's not mirrored in the data. I think we should follow the `WithId` pattern here

src/api/admin.ts
```diff
@@ -129,6 +129,20 @@ export interface VenueInput_v2
...
+  name: string;
+  description?: string;
+  subtitle?: string;
+  bannerImageFile?: FileList;
+  bannerImageUrl?: string;
+  logoImageFile?: FileList;
+  logoImageUrl?: string;
+  rooms?: Room[];
```
mike-lvov, line 141: should be no rooms as discussed?

src/api/admin.ts
```diff
@@ -129,6 +129,20 @@ export interface VenueInput_v2
...
 
+export interface WorldProps {
+  id: string;
+  name: string;
+  description?: string;
+  subtitle?: string;
+  bannerImageFile?: FileList;
+  bannerImageUrl?: string;
+  logoImageFile?: FileList;
+  logoImageUrl?: string;
```
mike-lvov, line 140: there types are a bit mismatching based on what I've seen in `functions/world`. Why?

mike-lvov (changes requested):

src/api/admin.ts
denisdimitrov: Because these are the form values that are expected in the create function, I will create another interface for the actual object thats being returned from the database.

src/api/admin.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/563960b617e371ddfc9d465b3428bdae64fafb0d

src/api/admin.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/563960b617e371ddfc9d465b3428bdae64fafb0d

src/api/admin.ts
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2295/commits/563960b617e371ddfc9d465b3428bdae64fafb0d

functions/world.js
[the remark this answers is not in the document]
goran-peoski-work: 👍

functions/world.js
```diff
@@ -1,57 +1,59 @@
...
+
+    if (admins.users && admins.users.includes(uid)) return;
+
+    throw new HttpsError("permission-denied", `User is not an admin`);
+  } catch (error) {
+    throw new HttpsError(
+      "internal",
+      `Error occurred obtaining world ${worldId}: ${err.toString()}`
+    );
+  }
```
goran-peoski-work, line 28: Name mismatch. Also, can probably work w/o explicit `.toString()`, assuming `HttpsError` extends `Error`

```suggestion
  } catch (error) {
    throw new HttpsError(
      "internal",
      `Error occurred obtaining world ${worldId}: ${error}`
    );
  }
```

goran-peoski-work (dismissed):
Few comments and suggestions. One suggestion about name typo `error` v `err`

denisdimitrov:
@goran-peoski-work The error has been fixed

@mike-lvov World id is now passed to newly created venues from admin ng

src/api/admin.ts
```diff
@@ -12,7 +12,7 @@ import {
   VenueTemplate,
 } from "types/venues";
 
-import { WithId } from "utils/id";
+import { WithId, WithWorldId } from "utils/id";
```
goran-peoski-work, line 15: @mike-lvov @denisdimitrov Does it make sense for these utility mixins to be generics or is it easier to deal with them as type unions?

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -106,6 +106,10 @@ export const AdminVenueView: React.FC = () => {
     return <LoadingPage />;
   }
 
+  if (!venue) {
+    return null;
+  }
+
```
goran-peoski-work, line 112:
```suggestion
  if (!venue) {
    // @debt replace null with a `NotFound` component for better UX
    return null;
  }

```

src/utils/id.ts
```diff
@@ -4,6 +4,8 @@ import omit from "lodash/omit";
 
 export type WithId<T extends object> = T & { id: string };
 
+export type WithWorldId<T extends object> = T & { worldId: string };
+
```
goran-peoski-work, line 8: Transforming `WithId<T>` as an alias just for `{ id: string }` might be OOS and a lot of work.

Should `WithWorldId<T>` also follow this pattern or can it be a POC?

Example:
```ts
export type WithWorldId = { worldId: string };
export type WithId = { id: string };
export type SpaceBase = { someSpaceProps: string[]  };
export type Space = Space & WithWorldId & WithId;
```

Of course, naming can be improved, the question is if generics really necessary since the most common use of the types will be with id and with worldId

denisdimitrov: It's a good question and you should bring it up in slack or the documentation, we can decide which one is better and change it if we have to.

goran-peoski-work: done

goran-peoski-work: @denisdimitrov https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1631782683234500?thread_ts=1631780155.228400&cid=C01ER4M61TL as a reference. Probably `WithWorldId` should be removed and just the optional prop added in few places

denisdimitrov:
@goran-peoski-work Debt has been added for the null return

src/utils/id.ts
denisdimitrov: @goran-peoski-work The field `worldId` is already optional in the venue type. But I've added this generic because sometimes you have forms, for example WorldFormInput wich doesn't have worldId in it. Because the form isn't editing the worldId. This is why I've added this and used it

mike-lvov (approved):
Looks good!

PR #2304, opened 2021-09-15
Chat users denormalization

dmitrymarchuk:
This PR removes usages of `worldUsers` from chats & reactions.

- Adds `ChatUser` type which is subset of `User` and contains `partyName`, `pictureUrl`, `anonMode`.
- Replaces `from` and `to` (in prviate chats) fields to contain `ChatUser` instead of user id. This is used to show avatar & name in messages.
- Does various refactorings:
  - ensures promises are passed wherever appropriate
  - aligns `Jukebox` messages with other types of messages
  - removes now redundant `MessageForDisplay` concept

As discussed with @mike-lvov this PR:

- Does not ensure that user data stays consistent. So if a user changes their username or profile picture the change will not propagate to all the messages the user has posted.
- Does not backfill existing messages with data. Instead it filters out messages that contain strings instead of user data objects.

Those issues will be addressed in separate PRs.


src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
```diff
@@ -17,10 +21,10 @@ import { OnlineUser, PrivateChatPreview, RecipientChat } from "..";
 import "./PrivateChats.scss";
 
 export interface PrivateChatsProps {
-  recipientId?: string;
+  recipient: WithId<ChatUser> | undefined;
```
mike-lvov, line 24:
```suggestion
  recipient,: WithId<ChatUser>;
```

src/store/actions/Chat.ts
```diff
@@ -37,12 +44,12 @@ export const setVenueChatTabOpened = (): SetVenueChatTabOpenedAction => ({
 });
 
 export const setPrivateChatTabOpened = (
-  recipientId?: string
+  recipient: WithId<ChatUser> | undefined
```
mike-lvov, line 47:
```suggestion
  recipient?: WithId<ChatUser>
```

src/types/chat.ts
```diff
@@ -13,17 +13,19 @@ export enum ChatMessageType {
   poll = "poll",
 }
 
+export type ChatUser = Pick<User, "partyName" | "pictureUrl" | "anonMode">;
+
 export type BaseChatMessage = {
-  from: string;
+  from: WithId<ChatUser>;
```
mike-lvov, line 19: I lean towards renaming it to `fromUser` and `toUser`. Since we already change the structure. + rename ts_utc to timestamp

src/utils/chat.ts
```diff
@@ -60,26 +32,29 @@ export const getPreviewChatMessage = ({
...
-  isMine: myUserId === message.from,
-});
-
 export const buildMessage = <T extends ChatMessage>(
-  message: Pick<T, Exclude<keyof T, "ts_utc">>
+  from: WithId<ChatUser>,
+  message: Pick<T, Exclude<keyof T, "ts_utc" | "from">>
 ) => ({
   ...message,
+  from: pickChatUserFromUser(from),
```
mike-lvov, line 40: I guess you already pass chat user?

src/utils/chat.ts
```diff
@@ -60,26 +32,29 @@ export const getPreviewChatMessage = ({
...
   ts_utc: firebase.firestore.Timestamp.now(),
 });
 
+export const pickChatUserFromUser = (user: WithId<User>) =>
+  pick(user, "id", "partyName", "pictureUrl", "anonMode") as WithId<ChatUser>;
```
mike-lvov, line 45:
```suggestion
export const pickChatUserFromUser = (user: WithId<User>): WithId<ChatUser> =>
  pick(user, "id", "partyName", "pictureUrl", "anonMode");
```

We should really avoid usage of `as`

src/utils/chat.ts
```diff
@@ -60,26 +32,29 @@ export const getPreviewChatMessage = ({
...
+
+export const messageContainsUserObject = <T extends ChatMessage>(
+  message: WithId<T>
+) => {
+  if (isString(message.from)) return false;
```
mike-lvov, line 50:
```suggestion
  if (isString(message.from)) return false;
  
```

mike-lvov (changes requested):
Looks good. A few improvements

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
dmitrymarchuk: You meant `recipient?:`?
But that's not the same as `type | undefined` as with `?` you can completely omit the prop but with `type | undefined` you have to pass it but it can be of the `undefined` type.

src/store/actions/Chat.ts
dmitrymarchuk: the same https://github.com/sparkletown/sparkle/pull/2304#discussion_r711879752

src/utils/chat.ts
dmitrymarchuk: That's for safety. 
The type of `from` is indeed `WithId<ChatUser>` but you can pass whenever "descendant" of that type. `WithId<User>` is usually passed but I didn't want to restrict to that as we only need `ChatUser` fields from `User`.
 But the object built here is sent to Firebase so we need to ensure that only needed fields are present in the object.

src/types/chat.ts
dmitrymarchuk: renamed in 

- https://github.com/sparkletown/sparkle/pull/2304/commits/20457ad6ae3e423d0d2f636d087d6413aee448d7
- https://github.com/sparkletown/sparkle/pull/2304/commits/440a95f3fbb83da3aa5d0029fa96ef6160913e76
- https://github.com/sparkletown/sparkle/pull/2304/commits/b5d5005b7fe96d278b9f2d76a25ea4f23acb20fb

src/components/organisms/ChatSidebar/components/PrivateChats/PrivateChats.tsx
mike-lvov: Ah, is it intentianal?

mike-lvov (approved):
LGTM

PR #2310, opened 2021-09-16
Move ./src/settings.ts into ./src/settings/index.ts

goran-peoski-work:
In order to facilitate addition of many more constants into the settings file, like in the examples bellow, and to make grouping easier, this change moves the file named `settings.ts` into a directory with the same name in order to allow it to re-export settings from additional files without changes on any code that imports it.

sparkle-settings-01

sparkle-settings-02

sparkle-settings-03


denisdimitrov (approved):
LGTM!

PR #2326, opened 2021-09-21
Add linkTo capability to non-isLink ButtonNG

goran-peoski-work:
The idea for ButtonNG props is simple:

isLink - should I use <a> or should I use <button>
linkTo - no matter if <a> or <button>, upon click, navigate to the provided URL

Thus far, the case of `isLink` = `false` and `linkTo` being provided hasn't been covered properly.

This was also missing for few BM use-cases
(slack reference for one of those https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1630255107051100?thread_ts=1630253800.048900&cid=C01ER4M61TL ).


src/utils/url.ts
```diff
@@ -149,3 +149,16 @@ export const getLastUrlParam = (url: string) => {
...
 };
+
+export const resolveUrlPath: (path: string) => string = (path) => {
+  const base = window.location.href;
+  try {
+    return new URL(path, base).href;
+  } catch (error) {
+    Bugsnag.notify(new Error(error), (event) => {
+      event.severity = "info";
+      event.addMetadata("utils::url::resolveUrlPath", { path, base });
```
denisdimitrov, line 160:
```suggestion
      event.addMetadata("utils/url::resolveUrlPath", { path, base });
```

denisdimitrov (approved):
LGTM!

src/utils/url.ts
denisdimitrov: We separate the folder/file with a `/` and then `::` for the function.

src/components/atoms/ButtonNG/ButtonNG.tsx
```diff
@@ -59,6 +63,27 @@ export const ButtonNG: React.FC<ButtonProps> = ({
...
+    if (disabled) return;
+
+    if (linkTo) {
+      if (newTab) {
+        window.open(
+          linkTo,
+          externalUrlAdditionalProps.target,
+          externalUrlAdditionalProps.rel
+        );
+      } else history.push(linkTo);
```
goran-peoski-work, line 78: Not necessary to use `resolvedUrl`, since that's transformed from the originally supplied and may be an error in resolution

src/utils/url.ts
goran-peoski-work: This should go in the code style MD documents as my code is a copy from above in the same file

PR #2208, opened 2021-09-01
[staging] POC for externalized messages in components

goran-peoski-work:
A POC for writing the messages in easy to maintain files, outside the logic of components

PR #2332, opened 2021-09-21
Added tooltip component [ staging ]

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1189
Needs https://github.com/sparkletown/internal-sparkle-issues/issues/1185 (the PR is blocked until the Topbar changes are merged)


src/assets/messages/index.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_MESSAGES: Readonly<Record<string, string>> = {
```
goran-peoski-work, line 1: [NOTE]
I like the way you've structured the file since I'm trying to make a simple externalization for non-dev people to be able to edit.

The place may end up being the assets/messages (like I've tested in a draft PR https://github.com/sparkletown/sparkle/pull/2208 ) so please @alexey-petrov  do inform me when you re-start working on this., might be able to re-use things from one PR into the next.

Alternatively, for those constants, there is now the possibility of extending the settings in a more structured manner. 

Example: the settings in https://github.com/sparkletown/sparkle/blob/c853e016b491178baf06b757591e88034f053e68/src/settings/taxonomy.ts#L5-L17
are being re-exported at https://github.com/sparkletown/sparkle/blob/c853e016b491178baf06b757591e88034f053e68/src/settings/index.ts#L29

PR #2338, opened 2021-09-23
User experience refactor

dmitrymarchuk:
This PR continues to remove worldUsers from the codebase. Highlights from changes:

- Audience worldUsers code was just commented out
- code related to Twilio `Participants` now fetches user data as `Participant`s connect (relates to all venues with video, except Auditorium)
- JazzBar, ConversationSpace, FriendshipPage now use `seatedTableUsers` subcollection to show users seated at tables
- various refactorings

PR #2342, opened 2021-09-23
Removes useFirestoreConnect from relatedVenues and seatedUsers

mike-lvov:
Successor of https://github.com/sparkletown/sparkle/pull/2338/files. Instead of upgrading the firebase to v9(with breaking changes), downgrade the reactfire lib


src/utils/converters.ts
```diff
@@ -0,0 +1,30 @@
+import firebase from "firebase/app";
+
+import { AuditoriumSeatedUser } from "types/auditorium";
+import { AnyVenue } from "types/venues";
+
+import { WithId, withId } from "utils/id";
+
+export const venueConverter: firebase.firestore.FirestoreDataConverter<
```
dmitrymarchuk, line 8: That's a really useful concept here👍
Can we make generic `withIdConverter` that to deduplicate `venueConverter` and `auditoriumSeatedUserConverter`?

dmitrymarchuk (approved):
LGTM

src/utils/converters.ts
mike-lvov: I think we can and I tried to, but I need to play with types a bit

src/components/organisms/AppRouter/AdminSubrouter.tsx
```diff
@@ -12,73 +12,51 @@ import { AdminAdvancedSettings } from "pages/AdminAdvancedSettings";
...
-          <VenueWizardV2 />
-        </Provided>
+        <VenueWizardV2 />
       </Route>
 
       <Route path="/admin-ng">
-        <Provided withWorldUsers>
-          <AdminDashboard />
-        </Provided>
+        <AdminDashboard />
```
denisdimitrov, line 59: Should we make all of these inline with Route's `render` or `component` props?

denisdimitrov: However it might look inconsistent because of the other Routes rendering children wrapped with `<Provided>`

src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -1,48 +1,26 @@
...
   const { currentVenue: venue } = useConnectCurrentVenueNG(venueId);
 
-  const MaybeRelatedVenuesProvider: React.FC<RelatedVenuesProviderProps> = withRelatedVenues
-    ? RelatedVenuesProvider
-    : EmptyProvider;
-
-  const MaybeWorldUsersProvider: React.FC<WorldUsersProviderProps> = withWorldUsers
-    ? WorldUsersProvider
-    : EmptyProvider;
+  if (!withRelatedVenues) return <>{children}</>;
```
denisdimitrov, line 19: [NB] You can just `return children`

mike-lvov: I don't think I can. I tried, but 
image

src/components/organisms/AppRouter/AdminSubrouter.tsx
mike-lvov: Fixed in https://github.com/sparkletown/sparkle/pull/2342/commits/979bc63e9a268bd407d54405dddaab08c0c779b8

PR #2344, opened 2021-09-23
Replace standard default avatars x6

dmitrii-zolotukhin:
closes sparkletown/internal-sparkle-issues#1188


dmitrymarchuk (approved):
LGTM

src/settings/index.ts
```diff
@@ -552,6 +552,8 @@ export const DEFAULT_AVATARS = [
   "/avatars/default-profile-pic-2.png",
   "/avatars/default-profile-pic-3.png",
   "/avatars/default-profile-pic-4.png",
+  "/avatars/default-profile-pic-5.png",
```
denisdimitrov, line 555: These are old files these urls should be changed and the files should be moved to `src/assets`

denisdimitrov (changes requested):

src/settings/index.ts
dmitrii-zolotukhin: @denisdimitrov I don't understand what's wrong and why they are old? I've just changed avatars

and why should they be in src/assets and not in src/assets/avatars? they were there before and the other avatars are there too

denisdimitrov: @dmitrii-zolotukhin Look at the files of the PR, they are currently in the `public` folder, which is a bad practice and React will 'flash' when loading the files.

denisdimitrov: They are always going to be bundled even when not being used, if you add them to `src/assets` they will be used and imported only in the correct file, because of our lazy loading.

denisdimitrov: Right now you can use `/avatars/default-profile-pic-5.png` because they are in public, if you change them to `src` you will have to import them, you can see `settings.ts` and the usage there.

dmitrii-zolotukhin: fixed in 8e852da134b5d0818d043ba28933080e00e3094c

denisdimitrov (approved):
LGTM!

PR #2347, opened 2021-09-24
Fixes users list component

mike-lvov:
Fixes users list component. Before that it incorrectly showed the number of users in the users list component

image


sunny-viktoryia (approved):
approving. 🙃

PR #2349, opened 2021-09-24
Added mixpanel-lite, removed mixpanel-browser

dmitrii-zolotukhin:
For testing need to add REACT_APP_MIXPANEL_PROJECT_TOKEN in ENV, token can be found in mixpanel room

closes sparkletown/internal-sparkle-issues#695


src/pages/VenuePage/VenuePage.tsx
```diff
@@ -172,12 +170,12 @@ export const VenuePage: React.FC = () => {
   // @debt refactor how user location updates works here to encapsulate in a hook or similar?
   useEffect(() => {
     if (user && profile && venueId && venueTemplate) {
-      mixpanel.track("VenuePage loaded", {
+      window.mixpanel.track("VenuePage loaded", {
```
sunny-viktoryia, line 173: can we create a mixpanel service/utils with all the tracking functions to have them all in the same place?

src/index.tsx
```diff
@@ -188,7 +187,9 @@ const BugsnagErrorBoundary = BUGSNAG_API_KEY
   : React.Fragment;
 
 if (MIXPANEL_PROJECT_TOKEN) {
-  mixpanel.init(MIXPANEL_PROJECT_TOKEN, { batch_requests: true });
+  window.mixpanel.init(MIXPANEL_PROJECT_TOKEN, {
+    batch_requests: true,
+  });
```
sunny-viktoryia, line 192: I'd move all the logic related to mixpanel to a separate service. And in the places where it's needed, I'd call it something like `initMixpanel`, `trackVenueEntrance(...params)`, etc.
so that I could easily change the logic in the functions without finding all the places where the mixpanel was used.

public/index.html
```diff
@@ -30,6 +30,7 @@
 -------------------------------------------------------------------
 */
     </script>
+    <script src="%PUBLIC_URL%/mixpanel-lite.min.js"></script>
```
sunny-viktoryia, line 33: why have you chosen to add the script directly rather than installing it? 

@mike-lvov, what would you say? is it a suitable approach for us?

sunny-viktoryia (commented):
can we create a separate service for mixpanel? to have a general function that will track everything, to be able to easily switch to another tracking system when/if needed.

public/index.html
dmitrii-zolotukhin: @sunny-viktoryia because it connects thought the script's adding, that is written in readme - https://github.com/john-doherty/mixpanel-lite#mixpanel-lite

src/pages/VenuePage/VenuePage.tsx
dmitrii-zolotukhin: fixed in 45c8facc226c9f6cd04226cd497c5d3662ab2527

src/index.tsx
dmitrii-zolotukhin: fixed in 45c8facc226c9f6cd04226cd497c5d3662ab2527

src/pages/Account/Login.tsx
```diff
@@ -82,7 +83,9 @@ export const Login: React.FC<LoginProps> = ({
     setFormToDisplay("passwordReset");
   };
 
-  const redirectAfterLogin = () => {};
+  const afterUserIsLoggedIn = () => {
+    trackMixpanelEvent("Login", { worldId: venue.worldId });
```
sunny-viktoryia, line 87: can we track the id or email of the person who is logged in to be able to count unique users?

src/pages/Account/Login.tsx
```diff
@@ -137,13 +140,13 @@ export const Login: React.FC<LoginProps> = ({
...
             closeAuthenticationModal={() => null}
-            afterUserIsLoggedIn={redirectAfterLogin}
+            afterUserIsLoggedIn={afterUserIsLoggedIn}
           />
         )}
         {formToDisplay === "passwordReset" && (
           <PasswordResetForm
             displayLoginForm={displayLoginForm}
-            closeAuthenticationModal={redirectAfterLogin}
+            closeAuthenticationModal={afterUserIsLoggedIn}
```
sunny-viktoryia, line 149: should we call this method? it's not a login.. this seems wrong.

src/utils/mixpanel.ts
```diff
@@ -0,0 +1,19 @@
+import { MIXPANEL_PROJECT_TOKEN } from "secrets";
+
+export const initMixpanel = (opts?: Object) => {
+  if (!MIXPANEL_PROJECT_TOKEN) return;
```
sunny-viktoryia, line 4: can we log a warning to the console about the mixpanel not being set up correctly?
i guess it's enough to have this log in the init function only.

sunny-viktoryia (commented):
few comments / questions left 🙃

src/pages/Account/Login.tsx
dmitrii-zolotukhin: fixed in b54905562aeb508beb223e2d87710f875297d159

src/utils/mixpanel.ts
dmitrii-zolotukhin: fixed in b54905562aeb508beb223e2d87710f875297d159

ermiaSparkle:
@dmitrii-zolotukhin Close it ?

PR #2350, opened 2021-09-24
pass worldId to all of the update venue calls

denisdimitrov:
world id is now passed everywhere and the requests should work properly.


goran-peoski-work (approved):
LGTM

PR #2354, opened 2021-09-27
Removes the legacy templates

mike-lvov:
Removes the legacy templates


yurii-lubynets (commented):
LGTM. Need some more reviews

PR #2356, opened 2021-09-27
Refactors Auditorium into one-section use

mike-lvov:
If Auditorium has only one section, it should automatically redirect users from `AllSectionPreviews`

closes https://github.com/sparkletown/internal-sparkle-issues/issues/1211


sunny-viktoryia:
looks like it's related to this issue? https://github.com/sparkletown/internal-sparkle-issues/issues/1211
isn't it?

mike-lvov:
> looks like it's related to this issue? [sparkletown/internal-sparkle-issues#1211](https://github.com/sparkletown/internal-sparkle-issues/issues/1211)
> isn't it?

Pretty much yes

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
```diff
@@ -39,13 +43,16 @@ export const AllSectionPreviews: React.FC<SectionPreviewsProps> = ({
...
     auditoriumSections,
+    isAuditoriumSectionsLoaded,
     toggleFullAuditoriums,
     isFullAuditoriumsHidden,
     enterSection,
     availableSections,
   } = useAllAuditoriumSections(venue);
 
-  const sectionsCount = auditoriumSections.length;
+  const sectionsCount = venue.sectionsCount ?? 0;
```
dmitrymarchuk, line 53: Why do we have to create another source of truth here? We already fetch all sections in `useAllAuditoriumSections` so why the need in a separate variable here?

dmitrymarchuk (approved):
Approved with a question.

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
mike-lvov: Really good question. So, as you know as per https://docs.google.com/document/d/1UvFuVgE7yzJSLNSTsv86LFkkd3qJ8i3CGKupfhKv5_c/edit plan, our next task is to implement `infinite scrolling`. Given that we won't fetch `allAuditoriumSections`, we won't have access to its length. Thus this variable.

However, I think it should be moved to another place(new counters collection) given that it can be updated often & it'll affect all venues in this case. which shouldn't happen

PR #2361, opened 2021-09-28
Dangling seated users fix

dmitrymarchuk:
This PR fixes the removal of dangling seated users. It:

- fixes scheduled functions: they failed because of >500 updates in a batch
- fixes simulator to add data in recentSeatedUsers collection
- adds a script to backfill `recentSeatedUsers` for specified auditorium
- adds a function that is triggered when a user is updated inside `recentSeatedUsers` to handle overwriting data


functions/scheduled.js
```diff
@@ -22,73 +28,69 @@ const removeDanglingSeatedUsers = async () => {
...
+      return batch.commit();
+    })
+  ).then(() => console.log(`Removed ${removedUsersCount} dangling users`));
 };
 
-exports.updateSeatedUsersCountInAuditorium = functions.pubsub
-  .schedule("every 5 minutes")
-  .onRun(async () => {
+exports.updateSeatedUsersCountInAuditorium =
+  // functions.https.onCall(async () => {
```
mike-lvov, line 84: remove?

functions/scheduled.js
```diff
@@ -193,11 +257,6 @@ exports.aggregateUsersLocationsInVenue = functions.pubsub
...
       }
 
-      batch.commit().catch((error) => {
-        throw new HttpsError(
-          "internal",
-          `Commit batch of recent users of venues failed. Error: ${error}`
-        );
-      });
+      return batch.commit();
     });
```
mike-lvov, line 261: Please, leave the error there. It can be useful

src/hooks/auditorium/useAllAuditoriumSections.ts
```diff
@@ -41,6 +41,7 @@ export const useAllAuditoriumSections = (venue: WithId<AuditoriumVenue>) => {
   const isFullAuditoriumsHidden = !isFullAuditoriumsShown;
 
   const sections = useSelector(currentAuditoriumSectionsSelector);
+  console.log(sections);
```
mike-lvov, line 44: Remove it

src/hooks/useUpdateRecentSeatedUsers.ts
```diff
@@ -18,34 +19,63 @@ import { getCurrentTimeInMilliseconds } from "utils/time";
...
 
-export const useUpdateRecentSeatedUsers = <T extends VenueTemplate>(
+const updateSeatedData = async <T extends VenueTemplate>(
+  template: T,
+  venueId: string | undefined,
+  userId: string | undefined,
+  venueSpecificData: RecentSeatedUserData<T>["venueSpecificData"] | FalseyValue
+) => {
+  if (!venueSpecificData || !venueId || !userId) return;
+  console.log("updating");
```
mike-lvov, line 29: REmove this too?

src/hooks/useUpdateRecentSeatedUsers.ts
```diff
@@ -18,34 +19,63 @@ import { getCurrentTimeInMilliseconds } from "utils/time";
...
+    () => {
+      void updateSeatedData(template, venueId, userId, venueSpecificData);
+    },
+    intervalRunning ? VENUE_RECENT_SEATED_USERS_UPDATE_INTERVAL : null
+  );
+
+  useEffect(() => {
+    void updateSeatedData(template, venueId, userId, venueSpecificData);
+  }, [template, userId, venueId, venueSpecificData]);
+};
```
mike-lvov, line 65: Given that it's useEffect, you have the option to return a function from here which should clean your previous position

mike-lvov (changes requested):

functions/scheduled.js
dmitrymarchuk: Since this is an `asyn` function anything you `await` or return from it will be handled by an functions executor and all errors will show up in logs like this:
image
Maybe I am missing something, but I don't see a point of catching and rethrowing here.

functions/scheduled.js
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2361/commits/81f321955aeab6f1da1596d30eb3fd572a53fb14

src/hooks/useUpdateRecentSeatedUsers.ts
dmitrymarchuk: fixed in https://github.com/sparkletown/sparkle/pull/2361/commits/81f321955aeab6f1da1596d30eb3fd572a53fb14

src/hooks/useUpdateRecentSeatedUsers.ts
dmitrymarchuk: Good point!
We need to remove two pieces of data when user leaves their seat: `recentSeatedUsers` and the venue-specific data (`seatedSectionUsers` for auditorium and `seatedTableUsers` for jazzbar etc)
There are already hooks for removing venue-specific data in each of the venues (e.g `leaveSeat` in cleanup) but for additional safety I updated triggered functions to to remove venue-specific data when the user leaves seat too.
Commit https://github.com/sparkletown/sparkle/pull/2361/commits/82820f56e03473510aa8e855499a51e95e463eaa

functions/package.json
```diff
@@ -5,7 +5,7 @@
     "lint": "prettier --write \"./**/*.{js,jsx,ts,tsx,json,css,scss,md}\""
   },
   "engines": {
-    "node": "10"
+    "node": "14"
```
dmitrymarchuk, line 8: That's to gain an ability to use ?. optional chaining. Also this is recommended by google: [docs](https://cloud.google.com/functions/docs/concepts/nodejs-runtime).
When I finish #2351 we can change this back to 10 (if any problems arise) as TS supports optional chaining and much more.

mike-lvov (approved):
LGTM

functions/scheduled.js
mike-lvov: Sounds reasonable

PR #2364, opened 2021-09-28
Fixed space modal styles

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1167

### DIFF
--- a/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
+++ b/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
@@ -14,6 +14,10 @@ $spacing: 20px;
     text-align: center;
   }
 
+  &__wrapper {
+    padding: 0;
+  }
+
   &__title {
     font-weight: 700;
 
@@ -44,7 +48,7 @@ $spacing: 20px;
     width: 180px;
     border-radius: $border-radius;
     background-color: $black;
-    background-size: contain;
+    background-size: cover;
     background-repeat: no-repeat;
     background-position: center;
   }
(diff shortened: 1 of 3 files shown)


src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -14,6 +14,10 @@ $spacing: 20px;
     text-align: center;
   }
 
+  &__wrapper {
+    padding: 0;
+  }
```
yurii-lubynets, line 19: `.modal-content` is used a lot of times for styling. Needs to reset indents here

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -83,13 +87,5 @@ $spacing: 20px;
...
     max-height: 35vh;
     overflow-y: auto;
-    // @debt these overrides are here because ScheduleItem sets incompatible margins/etc.
-    //   We should just clean it up there instead.
-    .schedule-item-container {
-      margin: -10px;
-      padding: 10px $spacing;
-
-      border-radius: $border-radius;
-    }
```
yurii-lubynets, line 93: `<ScheduleItem/>` component is used only for `RoomModal`. There is no need to override its styles

denisdimitrov:
@yurii-lubynets Can we put padding inside the elements instead of outside? That way the scrollbar will be on the right of the modal, right now it looks slightly weird in my opinion. If this is a figma design or product wants it this way, ignore my comment.

denisdimitrov (approved):
LGTM!

PR #2372, opened 2021-09-30
Move and rename DEFAULT_SHOW_REACTIONS -> DEFAULT_SETTING_SHOW_REACTIONS

goran-peoski-work:
To be in the same place with the rest of the  `useSettings` defaults:

```ts
export const DEFAULT_SETTING_SHOW_CHAT = true;
export const DEFAULT_SETTING_ENABLE_ADMIN_V1 = true;
export const DEFAULT_SETTING_ENABLE_ADMIN_V3 = true;
export const DEFAULT_SETTING_ADMIN_VERSION = 1;
export const DEFAULT_SETTING_SHOW_REACTIONS = true;
```


yurii-lubynets (approved):
LGTM.
Now settings are consistent

PR #2376, opened 2021-09-30
add ViewingWindow venue template

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1209

copied everything from the ArtPiece venue template adding a new property `isWithParticipants`
`if (isWithParticipants)` -> show cameras on the bottom
admin side should be finished by the @denisdimitrov's team. 🐉


src/components/templates/ViewingWindow/ViewingWindow.scss
```diff
@@ -0,0 +1,132 @@
...
+        .mirrored {
+          transform: scale(-1, 1);
+        }
+
+        .profile-icon {
+          cursor: pointer;
+          position: absolute;
+          bottom: 0;
+          left: 0;
+          border-radius: 188%;
```
denisdimitrov, line 98: Can we use the max constant for border radius here?

PR #2381, opened 2021-10-01
Fix simulator to seat on proper seats

dmitrymarchuk:
Now simulator seats:

- on seats not previously occupied by simulator run
- on seats that do not overlap with video


mike-lvov (approved):
Looks good

PR #2382, opened 2021-10-01
pass parentId to newly created spaces

denisdimitrov:
- New spaces have parent id
- Room urls are lowercasing the name of the venue and remove the spaces when multiple words exist
- Remove 'theme' from venue, it's not used anywhere

### DIFF
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -307,8 +307,8 @@ const createVenueData_v2 = (data, context) => {
     rooms: [],
     createdAt: Date.now(),
     updatedAt: Date.now(),
-    parentId: data.parentId || "",
     worldId: data.worldId,
+    ...(data.parentId && { parentId: data.parentId }),
   };
 
   if (data.template === VenueTemplate.jazzbar) {
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -297,9 +297,6 @@ const createVenueData_v2 = (data, context) => {
         description: data.description,
       },
     },
-    theme: {
-      primaryColor: data.primaryColor || DEFAULT_PRIMARY_COLOR,
-    },
     host: {
       icon: data.logoImageUrl,
     },
@@ -310,6 +307,7 @@ const createVenueData_v2 = (data, context) => {
     rooms: [],
     createdAt: Date.now(),
     updatedAt: Date.now(),
+    parentId: data.parentId || "",
     worldId: data.worldId,
   };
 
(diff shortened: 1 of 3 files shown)


src/components/molecules/VenueRoomItem/VenueRoomItem.tsx
```diff
@@ -60,10 +60,13 @@ export const VenueRoomItem: React.FC<VenueRoomItemProps> = ({
 
     const roomValues = getValues();
 
-    const roomUrl = window.origin + venueInsideUrl(roomValues.venueName);
+    const roomUrl =
+      window.origin +
+      venueInsideUrl(roomValues.venueName.replace(/\W/g, "").toLowerCase());
```
yurii-lubynets, line 65: It's worth to extract `roomValues.venueName.replace(/\W/g, "").toLowerCase()` as a separated var

functions/venue.js
```diff
@@ -310,6 +307,7 @@ const createVenueData_v2 = (data, context) => {
     rooms: [],
     createdAt: Date.now(),
     updatedAt: Date.now(),
+    parentId: data.parentId || "",
```
yurii-lubynets, line 310: Do we actually need this fallback as an empty string? Looks like it's an optional prop

yurii-lubynets (commented):
LGTM. Some minor enhancements are recommended

functions/venue.js
goran-peoski-work: Optional for the form doesn't mean it should be optional for the DB, might be prudent to have that field in there even if with empty string value

src/components/molecules/VenueRoomItem/VenueRoomItem.tsx
goran-peoski-work: Or at least

```ts
const ALL_NON_WORDS = /\W/g
```

goran-peoski-work: This looks more like an already existing utility function that converts the input venue name to some canonical URL friendly form - a slug

goran-peoski-work (approved):
LGTM

functions/venue.js
denisdimitrov: For now it is, because the same function is used to create worlds.

src/components/molecules/VenueRoomItem/VenueRoomItem.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2382/commits/96aad0110710e24901a4de2ec5af7244d3b9d114

PR #2384, opened 2021-10-01
change default space icon to the template selected

denisdimitrov:
Instead of using the old default room (now portal) image, now the newly created spaces have the selected template icon as default.


yurii-lubynets (approved):
LGTM

PR #2385, opened 2021-10-01
show portal name when editing a space

denisdimitrov:
Portal name was not being passed to the edit object and that's why the name was not showing when editing a space position/size.

Before:
old-name

After:
new-name


yurii-lubynets (approved):
LGTM

PR #2386, opened 2021-10-01
add useCheckImage to admin run tab

denisdimitrov:
- When map background is provided but it doesn't exist on the url anymore, it breaks in the run tab because useCheckImage was not used. Now the problem is fixed.


yurii-lubynets (approved):
LGTM

PR #2388, opened 2021-10-01
change audience to auditorium in admin

denisdimitrov:
The value being used when creating an auditorium was `audience` which is now a legacy template. Also the default values were incorrect, being `0` instead of the section values from settings. The updateVenue function was also not taking care of the rows and columns.


yurii-lubynets (approved):
LGTM

PR #2390, opened 2021-10-04
Unified AdminDashboard style with the rest of the admin

goran-peoski-work:
Follow up PR for:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1195

The updated style:
sparkle-adash-01

The hover effect:
sparkle-adash-02


src/components/molecules/AdminSpaceCard/AdminSpaceCard.scss
```diff
@@ -53,27 +61,14 @@ $admin-venue-card-button-margin: 5px 0;
   }
 
   &__logo {
-    height: 70px;
-    width: 70px;
-    border-radius: 50%;
+    @include square-size($admin-space-card-logo-size);
```
yurii-lubynets, line 64: `$admin-space-card-logo-size: 80px;` and it had `height: 70px` previously. Is it right?

yurii-lubynets (approved):
LGTM, a minor comment left. Nice catch for `font-size--parent/root` functions!

src/components/molecules/AdminSpaceCard/AdminSpaceCard.scss
goran-peoski-work: I changed it a bit. Tried to empirically keep it looking the same even if values are different.

PR #2391, opened 2021-10-04
Fix jukebox iframe url

alexey-petrov:
Fixes jukebox url change bug

### DIFF
--- a/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
+++ b/src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
@@ -11,7 +11,6 @@ import {
 
 import { JazzbarVenue, VenueTemplate } from "types/venues";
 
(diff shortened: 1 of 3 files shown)


src/components/templates/Jazzbar/JazzTab/JazzTab.tsx
```diff
@@ -195,12 +194,7 @@ const Jazz: React.FC<JazzProps> = ({ venue }) => {
                       key="main-event"
                       title="main event"
                       className="iframe-video"
-                      src={
-                        convertToEmbeddableUrl({
-                          url: venue.iframeUrl,
-                          autoPlay: venue?.autoPlay,
-                        }) ?? ""
-                      }
+                      src={iframeUrl}
```
sunny-viktoryia, line 197: is it possible to make them play immediately? because in my venue I have to start them manually.

sunny-viktoryia (approved):
The fix looks good.
a question left.

PR #2393, opened 2021-10-04
[BM] Fix Photobooth UI issues

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1142
 * Open in same window, as we know that having more than one Sparkle window open at a time breaks functionality
 * Like What's On button, this should not appear in admin. Follow logic as what's on and chat buttons for when to display / not display
 * Blocks the info button in venues


ermiaSparkle:
We don't need any update in the BM branch

PR #2399, opened 2021-10-05
Feature/Unify colors - lights and darks

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1260

### DIFF
--- a/src/components/molecules/AdminVenueCard/AdminVenueCard.scss
+++ b/src/components/molecules/AdminVenueCard/AdminVenueCard.scss
@@ -32,7 +32,7 @@ $admin-venue-card-button-margin: 5px 0;
       background-image: linear-gradient(
         180deg,
         rgba(25, 24, 26, 0.6) 0%,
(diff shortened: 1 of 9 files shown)


yurii-lubynets:
https://github.com/sparkletown/sparkle/blob/fb60b1dd4627cd8bfa817f6aee19a7a62981b0ad/src/scss/constants.scss#L16-L74

These variables (expect basic ones like `$red`) are used across the platform as secondary (grey-ish) colors. They should be definitely simplified as secondary and relevant to it colors.
It should be defined which color needs to bed added to the palette
```
$colors-map: (
  // deprecated color, should probably be replaced with a more appropriate one
  primary--over: #f6e652,
  primary--under: #7c46fb,
  content--under: #000000,
  content--over: #ffffff,
  accent--over: #ffffff,
  accent--under: #7c46fb
    // NOTE: there is a need for danger--over and danger--under
);
```

PR #2405, opened 2021-10-06
Make both createFirestoreVenueInput functions set safe URL slug as venue name

goran-peoski-work:
Tries to resolve
- https://github.com/sparkletown/internal-sparkle-issues/issues/1252

Update on the front end that forces venue name conform to slug since currently is used just as a part of a URL.
This PR tries to make a middle-of-the-road solution between
>fix this or
> disallow user to type any spaces in venuename field and label it "URL" instead

In the way that it will allow other characters in the field, but update (strip the unwanted) before sending to back end.

### DIFF
--- a/src/api/admin.ts
+++ b/src/api/admin.ts
@@ -199,7 +199,7 @@ const createFirestoreVenueInput = async (
 ) => {
   const storageRef = firebase.storage().ref();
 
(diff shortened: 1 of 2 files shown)


denisdimitrov (approved):
LGTM!

PR #2420, opened 2021-10-12
Fix multiple worlds being created on `createWorld()` error

goran-peoski-work:
A solution for one of the bugs mentioned in https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1633708766167000?thread_ts=1633704352.165400&cid=C02DPU9KF1A
> creating world doesn't update the start form so new one gets created upon an error like one when uploading images

Part of:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1230


goran-peoski-work:
Do not merge until https://github.com/sparkletown/sparkle/pull/2420#discussion_r727066145 is resolved

whoaitsaimz:
I'm a little unclear on what this PR does or doesn't do.

We don't need a venue automatically created with a world. I'm not sure why/where that was ever specified as a thing to do.

I hope that this PR removed automatic venue creation altogether. But the discussion above and do-not-merge tag is confusing.

PR #2429, opened 2021-10-14
Add TesterRestricted component

goran-peoski-work:
This component can check user tester status and provide/deny access to children. They will simply render or not.

Resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1337

From Slack DM to Pearl:
> this was meant to provide quick and easy addition of a component that can check if users should have access to not-finished alpha, beta or any kind of components that are WIP. The component approach is reusable in the future, however, adding comprehensive user/role/permission system isn't the goal, but can update the component in the future to rely on such system

For it to work...
> you just find your user in the DB and put a field tester: true (or any truthy value in that regard, but boolean is good)

> After all that, next thing you ask to be your alpha/beta testing eyes only, not for GA, you ask in the issue we wrap the new stuff in `<TesterRestricted>` component

> In the future if we need more fine grained access to different tester restricted functionalities, we can play with that `true` value to be something more descriptive

> An example would be if a user is meant to only see demo, we can add value (instead of true) that is a string or an array of strings as to which are visible/which aren't.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,6 +201,7 @@ jobs:
       - *notify-on-fail
   deploy-hosting:
     executor: node/default
+    resource_class: "large"
     steps:
       - checkout
       - node/install-packages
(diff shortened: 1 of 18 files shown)


whoaitsaimz:
@goran-peoski-work so this is a step toward a "superadmin" user?

PR #2435, opened 2021-10-15
use adminpanel in space editor screen

denisdimitrov:
Adds AdminPanel edit space screen and removes the styled components. Rename the component to something more appropriate.


goran-peoski-work:
Just noticed `DetailsForm.types.ts` can also be removed: its two interfaces moved in the file they're being used. Regardless, might also be done in future PR dealing with all `.types.ts` remains.

PR #2439, opened 2021-10-18
Feature/toggle reactions panel w/shoutouts emojis

yurii-lubynets:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1214


whoaitsaimz:
>@whoaitsaimz
We can think about designing better visual connection between the Reaction emojis and Audible (since the former disables the latter), but for now it looks good to me.

It does not match the design provided, but if it works functionally, fine, I'll live with it for now. I'm a little confused about why this was all chunked together with a new feature being developed, and not done as laid out in the original issue spec.

PR #2446, opened 2021-10-19
Fix simulator freeze

dmitrymarchuk:
- Removes `setTimeout` logic and replaced it with loop in attempt to fix simulator freeze.
- Fix the problems with unique seat placements - grid was shared between all sections.


mike-lvov (approved):
LGTM

PR #2453, opened 2021-10-20
disable initial venue creation

denisdimitrov:
Temporary disable the initial venue creation due to possible complications. Since we want everything to be working as good as possible by the end of the day, this is the fastest route.


yurii-lubynets:
@denisdimitrov please use `||` for `/functions`

PR #2461, opened 2021-10-20
Popup bypass for clicking External Link

goran-peoski-work:
Add popup modal bypass for clicking External Link portals in the party map.

Right now External Link opens the same Join modal popup before clicking a button again to go to the correct URL (like a zoom meeting). With this PR that popup will not open but the actual URL.


goran-peoski-work:
It turns out https://github.com/sparkletown/internal-sparkle-issues/issues/1298#issuecomment-947675718 makes this PR not necessary. Closing it.

PR #2466, opened 2021-10-21
Increase circleCI instance size

colinhowe:
Temporarily increase instance size for build/deploy phase to get around
memory/slowness issues that are causing builds to fail.


colinhowe:
Yes, it supersedes #2463

PR #2474, opened 2021-10-25
[Admin] implement and use world slug instead of world id

denisdimitrov:
Use world slug in the urls and search by it in the hooks/functions.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1326


denisdimitrov:
All of the comments fixed and the slug is used only in the URLs now.

src/api/world.ts
```diff
@@ -17,7 +17,7 @@ import { isDefined } from "utils/types";
 export const createFirestoreWorldCreateInput: (
   input: WorldStartFormInput,
   user: firebase.UserInfo
```
goran-peoski-work, line 19:
```suggestion
export const createFirestoreWorldCreateInput: (
  input: WorldStartFormInput,
```

src/api/world.ts
```diff
@@ -62,7 +62,6 @@ export const createFirestoreWorldStartInput: (
   const worldUpdateData: Partial<WithId<World>> = {
     ...omit(input, Object.keys(imageInputs)),
     ...imageInputData,
-    id,
```
goran-peoski-work, line 65: What happens if `generateFirestoreId()` gets invoked because of
```ts
  // NOTE: id is needed before world is created to upload the images
  const id = input?.id ?? generateFirestoreId({ emulated: true });
```
?
Can it ever be invoked?

src/components/organisms/AdminDashboard/AdminDashboard.tsx
```diff
@@ -29,10 +30,13 @@ import "./AdminDashboard.scss";
 export const AdminDashboard: React.FC = () => {
   const { ownedVenues, isLoading } = useOwnedVenues({});
 
-  const { worldId } = useWorldEditParams();
+  const { worldSlug } = useWorldEditParams();
 
-  const venues = worldId
-    ? ownedVenues.filter((venue) => venue.worldId === worldId)
+  const { world, isLoaded } = useWorldBySlug(worldSlug);
```
goran-peoski-work, line 35: Might be easier to use if a single `useWorld()` hook exists that can inspect its options

```ts
const { world, isLoaded } = useWorld({ worldSlug });
```
```ts
const { world, isLoaded } = useWorld({ worldId });
```

src/components/organisms/AdminDashboard/AdminDashboard.tsx
```diff
@@ -29,10 +30,13 @@ import "./AdminDashboard.scss";
...
+
+  console.log(world);
```
goran-peoski-work, line 37:
```suggestion
```

src/components/organisms/AdminDashboard/AdminDashboard.tsx
```diff
@@ -72,7 +76,7 @@ export const AdminDashboard: React.FC = () => {
 
   const hasVenues = renderedPartyVenues.length > 0;
 
-  if (isLoading) {
+  if (isLoading || !isLoaded) {
```
goran-peoski-work, line 79: Maybe rename to distinguish:
```suggestion
  if (isLoadingVenues || !isLoadedWorld) {
```

src/hooks/useWorldBySlug.ts
```diff
@@ -0,0 +1,68 @@
+import { useFirestore, useFirestoreCollectionData } from "reactfire";
```
goran-peoski-work, line 1: Maybe it's best to use a single `useWorld()` hook that can distinguish search by `id` or `slug`.

src/hooks/useWorldBySlug.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (!worlds?.[0]) {
+    Bugsnag.notify(
+      `World with the following slug: ${worldSlug} does not exist.`,
+      (event) => {
+        event.severity = "warning";
+        event.addMetadata("hooks::useWorldBySlug", {
+          worldSlug,
+        });
+      }
+    );
```
goran-peoski-work, line 48: [NOTE]
Not sure if worthy of Bugsnag like multiple worlds found since it is going to be quite visible to the user the world doesn't exist, but will probably add numerous warning entries.

src/hooks/useWorldBySlug.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  if (worlds?.length > 1) {
+    Bugsnag.notify(
+      `Multiple worlds have been found with the following slug: ${worldSlug}.`,
+      (event) => {
+        event.severity = "warning";
```
goran-peoski-work, line 55: [NOTE]
Maybe not just a warning, also might be easy if there are few more "labels" attached like event.admin = true or similar

src/hooks/useWorldEdit.ts
```diff
@@ -26,7 +27,7 @@ export const useWorldEdit: (worldId?: string) => UseWorldEditResult = (
     }
   );
 
-  const isWorldLoaded = status === "success";
+  const isWorldLoaded = status !== "loading";
```
goran-peoski-work, line 30: 👍

src/pages/Admin/Details/Form/DetailsForm.tsx
```diff
@@ -52,6 +53,7 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ venue }) => {
   const { worldVenuesIds, worldParentVenues } = useWorldVenues(
     worldId ?? venue?.worldId ?? ""
   );
+  const { world, isLoaded } = useWorldEdit(worldId ?? venue?.worldId);
```
goran-peoski-work, line 56: Same comment as `useWorld()` suggestion in other places.

src/pages/Admin/Details/Form/DetailsForm.tsx
```diff
@@ -112,38 +114,44 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ venue }) => {
...
+            worldId: worldId ?? "",
+            parentId: values.parentId ?? "",
+          };
+
+          await createVenue_v2(newVenue, user);
+
+          history.push(adminWorldSpacesUrl(world?.slug));
+        }
+      } catch (e) {
+        console.error(e);
```
goran-peoski-work, line 142: This will break `const [{ error: submitError,` and really there is no need to log it here since it is the job of these lines https://github.com/sparkletown/sparkle/pull/2474/files#diff-4e9f5a0d22dd89662d1eb6e2d72a265855179bf67c408c5c350292dc9eebef24R355-R356

and an improvement coming along the way with https://github.com/sparkletown/sparkle/pull/2503

goran-peoski-work (changes requested):
Comments and suggestions.

src/components/organisms/AdminDashboard/AdminDashboard.tsx
denisdimitrov: `reactfire` is using a lot of hooks and it will be difficult to put conditions in a single hook and make it work fast, I will most likely have to fetch both worlds by Id and slug and then use the one with a result, because we can't use hooks after conditions. It won't be very good for the performance of the platform.

denisdimitrov: They also return different data, since one of the hooks is searching by the doc id and the other one with a `where` operator

src/hooks/useWorldBySlug.ts
denisdimitrov: This is only for dev purposes and getting notified on slack that this thing happened. It will be easier to trace it from there.

src/api/world.ts
denisdimitrov: I don't thinks so, id should be always there.

src/pages/Admin/Details/Form/DetailsForm.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2474/commits/810959fc143f367fa1c4c139201437a894fc3b54

src/components/organisms/AdminDashboard/AdminDashboard.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2474/commits/810959fc143f367fa1c4c139201437a894fc3b54

src/components/organisms/AdminDashboard/AdminDashboard.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2474/commits/810959fc143f367fa1c4c139201437a894fc3b54

functions/src/utils/venue.js
```diff
@@ -6,8 +6,13 @@ const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
 const getVenueId = (name) =>
   name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
 
+const generateSlug = (name) => {
+  return name.replace(/\W/g, "").toLowerCase();
+};
+
```
goran-peoski-work, line 12: These should be a single function (ideally named `generateSlug()`) using `INVALID_SLUG_CHARS_REGEX ` instead of `/\W/g`

```suggestion
const getVenueId = (name) =>
  name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
};

```

functions/src/utils/venue.js
```diff
@@ -6,8 +6,13 @@ const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
...
 const checkIfValidVenueId = (venueId) => VenueIdSchema.isValidSync(venueId);
 
 exports.VENUE_ID_REGEX = VENUE_ID_REGEX;
 exports.getVenueId = getVenueId;
+exports.generateSlug = generateSlug;
```
goran-peoski-work, line 17:
```suggestion
exports.generateSlug = getVenueId;
```

src/api/world.ts
goran-peoski-work: > Should

☝️ this is a big assumption. I will try to make another one, based on the code, before running and testing it.


Here is how the ID is being passed down in the FE:
```js
    const worldId = (
      await firebase.functions().httpsCallable("world-createWorld")(stubInput)
    )?.data;

    const fullInput = await createFirestoreWorldStartInput(
      withId(world, worldId),
      user
    );
```

And this is how this PR makes it always `undefined` :
```js
    .then(() => ({ ...worldData, id: worldDoc.id }));
```
@ https://github.com/sparkletown/sparkle/pull/2474/files#diff-c795148bf23657fc8fd09642f2b39a2789f97a4fe701fd8062eb9cd33f1c83caL89-R91

So `generateFirestoreId()` will likely always get executed and 
```js
    await firebase.functions().httpsCallable("world-updateWorld")(fullInput);
```
will try to update world other than the one created by `"world-createWorld"`, most likely with an empty string as an ID
```js
  let worldId = "";
```

I will try to test this.

src/components/organisms/AdminDashboard/AdminDashboard.tsx
goran-peoski-work: `They also return different data` <- still the type `World`, right?

So, just to double check, you say something like the following will be slow:
```ts

  const queryBySlug = worldsRef
    .where("slug", "==", worldSlug ?? "")
    .withConverter(worldConverter);
  const { data: worlds, status: statusBySlug } = useFirestoreCollectionData<WithId<World>>(queryBySlug, { initialData: undefined });

  const queryById = worldsRef.doc(worldId).withConverter(worldConverter);
  const { data: world, status: statusById } = useFirestoreDocData<WithId<World>>(queryById, { initialData: undefined });

  const isLoaded = (worldId ? statusById : statusBySlug) !== "loading";

  return {
    world: worldId ? world : worlds?.[0],
    isLoaded: isLoaded,
  };
```

I'm not sure I can make that claim without performance test, but If that's the case, I should probably make some small PR later on to make the hooks appear similar in name.

src/components/organisms/AdminDashboard/AdminDashboard.tsx
```diff
@@ -27,12 +28,14 @@ import { ButtonNG } from "components/atoms/ButtonNG";
...
 export const AdminDashboard: React.FC = () => {
-  const { ownedVenues, isLoading } = useOwnedVenues({});
+  const { ownedVenues, isLoading: isLoadingSpaces } = useOwnedVenues({});
 
-  const { worldId } = useWorldEditParams();
+  const { worldSlug } = useWorldEditParams();
 
-  const venues = worldId
-    ? ownedVenues.filter((venue) => venue.worldId === worldId)
+  const { world, isWorldLoaded } = useWorldBySlug(worldSlug);
```
goran-peoski-work, line 35:
```suggestion
  const { world, isLoaded: isWorldLoaded } = useWorldBySlug(worldSlug);
```

src/components/organisms/AdminDashboard/AdminDashboard.tsx
denisdimitrov: Yes this is how that hook is supposed to look like but there is always going to be an unnecessary search. If I'm passing down an ID the logic shouldn't search by slug, but since we can't use conditions in before hooks, it always will. And it's a bad practice to trade performance for code quality or readability. The safer approach would be to use 2 separate hooks and don't risk this.

goran-peoski-work:
Yes, I think this https://github.com/sparkletown/sparkle/pull/2474#discussion_r741118227 is a confirmed bug
sparkle-create-world-slug-01

denisdimitrov:
@goran-peoski-work All comments have been fixed an taken into consideration. I had to change the WorldNav to use worldSlug for the disable check. The validation for create world has been changed a big and now uses updateSchema as well. The buttons for Enter dashboard and Configure World have been disabled when world slug is missing because it's used in the URL and without it nothing will appear or it will be broken.

functions/src/utils/venue.js
```diff
@@ -3,6 +3,9 @@ const { VenueIdSchema } = require("../types/venue");
 const VENUE_ID_REGEX = /[a-z0-9_]{1,250}/;
 const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
 
+/**
+ * @deprecated Use generateSlug instead.
+ */
```
goran-peoski-work, line 8: Please don't use 
```js
const generateSlug = (name) => {
  return name.replace(/\W/g, "").toLowerCase();
};
```

```suggestion
// @debt rename this one to generateSlug
```

`getVenueId()` has the correct implementation as staging:

sparkle-generate-slug-01

src/api/world.ts
```diff
@@ -61,6 +61,7 @@ export const createFirestoreWorldStartInput: (
   const worldUpdateData: Partial<WithId<World>> = {
     ...omit(input, Object.keys(imageInputs)),
     ...imageInputData,
+    id,
```
goran-peoski-work: 👍

goran-peoski-work (changes requested):
- `generateSlug` is the correct name
- `.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase()` is the correct implementation

functions/src/utils/venue.js
denisdimitrov: My bad, I'll fix it/

denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2474/commits/cf4233edc137706c2ccb7d1ce357c2591cba2370

src/pages/WorldEditor/WorldEditor.tsx
```diff
@@ -41,7 +41,10 @@ export const WorldEditor: React.FC = () => {
       <WithNavigationBar hasBackButton withSchedule>
         <AdminRestricted>
           <WorldNav />
-          <WorldEditorPanel worldId={worldId} onClickHome={navigateToHome} />
+          <WorldEditorPanel
```
goran-peoski-work, line 44: [NOTE] Panel should probably not be aware how the id comes by, so translation slug -> id should probably come before the panel is used (this helps the panel being decoupled from route specifics

goran-peoski-work (approved):
LGTM

src/hooks/worlds/useWorldBySlug.ts
```diff
@@ -36,12 +36,6 @@ export const useWorldBySlug: (worldSlug?: string) => UseWorldBySlugResult = (
 
   const isWorldLoaded = status !== "loading";
 
-  if (!worlds?.[0]) {
-    throw new Error(
-      `World with the following slug: ${worldSlug} does not exist.`
-    );
-  }
```
goran-peoski-work, line 43: 🎉 FTW 🎉

PR #2503, opened 2021-10-29
Update error reporting in SpaceEditForm

goran-peoski-work:
Resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1391

Partially resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1230

There is a change to the error reporting components to support nested fields and some of the form inputs to be able to properly show the errors (old components overlaid the error text above the next ones' labels).

sparkle-embeddable-03


src/utils/filterUniqueKeys.ts
```diff
@@ -1,3 +1,4 @@
+// @debt remove, it seems it is unused
```
alexey-petrov, line 1: can we remove it now or should we do it later on?

alexey-petrov (approved):
LGTM

src/utils/filterUniqueKeys.ts
goran-peoski-work: For next PR, didn't check if/where it is used, so I didn't want to risk it.

PR #2507, opened 2021-11-01
Return firebarrel in the Add Portal but hidden

goran-peoski-work:
Partially resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1298

Based on comment (https://github.com/sparkletown/internal-sparkle-issues/issues/1298#issuecomment-955594955):
> Final thing: need to enable - but hidden - firebarrel, until we account for this in the unified table template (it will act as a one-table conversation space // no central content).

sparkle-add-portal-04


denisdimitrov (approved):
LGTM!

PR #2508, opened 2021-11-01
add world slug validation for world create and edit

denisdimitrov:
Adds world slug validation to prevent multiple worlds being created with the same world slug.

Related issue: https://github.com/sparkletown/internal-sparkle-issues/issues/1318


goran-peoski-work (approved):
LGTM

PR #2509, opened 2021-11-01
Remove underscore from available slug characters

goran-peoski-work:
Old `RegExp` allowed the underscore `_` in the generated slugs, even though usually no underscores are used in space/world names..

With this minor refactor, the function has more clear name `createSlug()` as well as having only `a` to `z` (in both lower and upper cases) and `0` through `9` as digits, be available.

Additionally, this can be updated to just turn all slug chars to lowercase.


scripts/create-jazz-bar.ts
```diff
@@ -56,7 +56,8 @@ const jazzbar: JazzbarVenue = {
   code_of_conduct_questions,
 };
 
-const venueId = name.replace(/\W/g, "").toLowerCase();
+const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
+const venueId = name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
```
colinhowe, line 60: Could we re-use the code from `utils/venue.js`?

functions/src/utils/venue.js
```diff
@@ -1,10 +1,10 @@
 const { VenueIdSchema } = require("../types/venue");
 
 const VENUE_ID_REGEX = /[a-z0-9_]{1,250}/;
+const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
```
colinhowe, line 4: Is it worth moving any slug logic into a `utils/slugs.ts` module?

src/settings/index.ts
```diff
@@ -679,3 +679,5 @@ export const VENUES_WITH_CHAT_REQUIRED = [
 export const VENUE_CHAT_MESSAGES_COUNTER_SHARDS_COUNT = 10;
 
 export const NON_EXISTENT_FIRESTORE_ID = "NON_EXISTENT_FIRESTORE_ID";
+
+export const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
```
colinhowe, line 683: This should be refactored so that we only define this constant once.

colinhowe (changes requested):
Few minor refactors, please. Mostly to remove code repetition.

I am concerned about allowing slugs to have a mix of uppercase and lowercase in them. I think that will allow for: `myworld` and `myWorld leading to two different worlds which could be a confusing experience.

We shouldn't care, but, users do strange things. I think of this in the same way as I think about DNS. You wouldn't want `www.Google.com` to lead you to a different site to `www.google.com`. Scammers would exploit it.

scripts/create-jazz-bar.ts
goran-peoski-work: No. Doing so is OOS.

One is run in the browser, another one is run in the google cloud, third one is run in the terminal.

functions/src/utils/venue.js
goran-peoski-work: Not by my account https://github.com/sparkletown/sparkle/pull/2509#discussion_r740245059

src/settings/index.ts
goran-peoski-work: Interested. Please do provide me with a working POC. https://github.com/sparkletown/sparkle/pull/2509#discussion_r740245059

scripts/create-jazz-bar.ts
colinhowe: Agreed that this is OOS.

For my own education about this project: would it be hard to extract the slug logic out to one module and have all three areas use the same module?

colinhowe:
Ah. You're right. My confusion came from the top-level comment on the PR:

> With this minor refactor, the function has more clear name createSlug() as well as having only a to z (in **both lower and upper cases**) and 0 through 9 as digits, be available.

(bold added by me).

I'm happy with this PR now.

colinhowe (approved):

scripts/create-jazz-bar.ts
goran-peoski-work: Starting from the current state: yes.
Starting with a blank slate: maybe, I don't know  (haven't configured TS to output code for one, let alone 3 environments).

goran-peoski-work:
> Ah. You're right. My confusion came from the top-level comment on the PR:

Bad wording on my part.

PR #2510, opened 2021-11-01
Fixed end time before start time bug

alexey-petrov:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1382

Prevents event bug when user sets end time before start time

### DIFF
--- a/src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
+++ b/src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from "react";
+import React, { forwardRef, useCallback, useEffect, useState } from "react";
 
 import { getDateTimeFromUtc, getUtcFromDateTime } from "utils/time";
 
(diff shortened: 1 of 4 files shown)


src/components/organisms/AdminVenueView/components/DateTimeField/DateTimeField.tsx
```diff
@@ -9,38 +9,64 @@ export interface DateFieldProps {
...
           value={dateValue}
           onChange={handleDateChange}
         />
+        <input
+          type="number"
+          value={utcValue}
+          ref={ref}
+          name={name}
+          hidden
+          readOnly
```
colinhowe, line 68: Why is this input here twice? See lines 48-55

colinhowe (changes requested):
I think there's a duplicate component that isn't neeeded.

This code only validates start time/end time on the frontend. I think we should do a change in `functions/venue.js`to validate the start time and end time there to prevent this class of error entirely.

alexey-petrov:
Closed in favor of https://github.com/sparkletown/sparkle/pull/2512

PR #2513, opened 2021-11-02
[Admin] disable world nav when creating a world

denisdimitrov:
Users could navigate to Entrance and Advanced when creating a world, which was totally breaking the UI and navigation. Users should not be able to access those tabs until the world has been created or the proper IA has been provided.

Issue: https://github.com/sparkletown/internal-sparkle-issues/issues/1409


src/pages/WorldEditor/WorldEditorStartPanel.tsx
```diff
@@ -28,9 +28,11 @@ export const WorldEditorStartPanel: React.FC<WorldEditorStartPanelProps> = ({
   const isUpdatingWorld = isLoaded && worldId && world;
   const isCreatingWorld = isLoaded && !world && !worldId;
 
+  const sidebarClassName = !worldId ? "AdminDashboard__sidebar" : "";
+
   return (
     <AdminPanel>
-      <AdminSidebar>
+      <AdminSidebar className={sidebarClassName}>
```
goran-peoski-work, line 35: This can be resolved within WorldEditorStartPanel.scss without class names injection

src/pages/WorldEditor/WorldEditor.tsx
```diff
@@ -40,7 +40,7 @@ export const WorldEditor: React.FC = () => {
     <div className="WorldEditor">
       <WithNavigationBar hasBackButton withSchedule>
         <AdminRestricted>
-          <WorldNav />
+          {worldId && <WorldNav />}
```
goran-peoski-work, line 43: Hiding and showing the navigation creates more UI issues than simply disabling the buttons. Both approaches, however, do not solve the issue of someone simply typing the URL in the address bar.

src/pages/WorldEditor/WorldEditorStartPanel.scss
```diff
@@ -7,6 +7,10 @@
     grid-template-columns: repeat(3, 1fr);
   }
 
+  &__sidebar {
+    height: calc(100% - #{$admin-ng-sidebar-footer-height--normal});
+  }
+
```
goran-peoski-work, line 13: There will be no need for re-calculating the Admin heights and breaking unified admin UI/UX (especially since almost all the panels now do have a general nav bar, an admin nav bar a side bar etc...).

Simply disabling the tabs from clicking should work without the extra effort and inconsistencies.

goran-peoski-work (changes requested):
I think there's a simpler way of solving this.

src/pages/WorldEditor/WorldEditor.tsx
denisdimitrov: That's true we need to think about that as well. Keep in mind that most of our buttons use history.goBack and opening a url directly would break those buttons and make them do nothing.

goran-peoski-work: > most of our buttons use history.goBack and opening a url directly would break those buttons and make them do nothing.

Looks like an opportunity for future PR to make the navigation more robust.

ermiaSparkle:
Users could navigate to Entrance and Advanced when creating a world, which was totally breaking the UI and navigation. Users should not be able to access those tabs until the world has been created or the proper IA has been provided.

denisdimitrov:
All of the comments have been fixed in this commit: https://github.com/sparkletown/sparkle/pull/2513/commits/cc18036b093b8c93b6e1883279d87e953770c33a

The changes have been reverted and the links are now disabled as per our discussion.

goran-peoski-work (approved):
LGTM

PR #2514, opened 2021-11-03
fix auditorium logo upload in admin v3

denisdimitrov:
The problem is that `values.image_file` is not passed down. However TS is crying because that doesn't exist in the `values` object, it's actually mutated by either the image component or useForm. That's why we can't simply pass it and we use spread operator to pass the whole thing and allow the function to get what it needs. It is a very hacky solution and now I understand why it was written this way in the other form.

Fixes: N/A


goran-peoski-work (approved):
LGTM

PR #2516, opened 2021-11-04
Remove text transforms in SCSS that capitalized all words

goran-peoski-work:
Remove 
```scss
text-transform: capitalize;
```

so that 
> for purpose of consistency: across platform, headings and most things should be written in Sentence case (first word capital, the rest lowercase) rather than Title Case.

>Design comps reflect sentence case.
>Design team has agreed that Sentence case feels more friendly and sparkly 

(from https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1635443057094400)

NB: please don't try to fix it with CSS like `:first-letter { text-transform: capitalize; }` since it may have a legitimate corner cases (e.g. try starting a sentence with iPhone), but instead try to make the actual text the proper casing.


alexey-petrov (approved):
lgtm

PR #2517, opened 2021-11-04
Remove the home button in WorldsDashboard

goran-peoski-work:
By removing the fixed footer all together.

Resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1410

sparkle-no-home-01


alexey-petrov (approved):
lgtm

PR #2519, opened 2021-11-04
[NotFound page] Hide 'log in' and 'go back' buttons

sunny-viktoryia:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1366
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1329

### DIFF
--- a/src/components/atoms/NotFound/NotFound.tsx
+++ b/src/components/atoms/NotFound/NotFound.tsx
@@ -9,6 +9,8 @@ import SPARKLEVERSE_LANDSCAPE from "assets/images/Sparkle_logo_white_yellow.png"
 
 import "./NotFound.scss";
 
+const JUST_CURRENT_PAGE_HISTORY_LENGTH = 1;
+
 export interface NotFoundProps {
   fullScreen?: boolean;
 }
@@ -34,13 +36,15 @@ export const NotFound: React.FC<NotFoundProps> = ({ fullScreen }) => {
           src={TV_IMAGE}
         />
         <p className="NotFound__move-along-text">Nothing to see here.</p>
-        <ButtonNG
-          className="NotFound__button"
-          variant="primary"
-          onClick={history.goBack}
-        >
-          Go Back
-        </ButtonNG>
+        {history.length > JUST_CURRENT_PAGE_HISTORY_LENGTH && (
+          <ButtonNG
+            className="NotFound__button"
+            variant="primary"
+            onClick={history.goBack}
+          >
+            Go Back
+          </ButtonNG>
+        )}
       </div>
     </div>
   );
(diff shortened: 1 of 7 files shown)


src/components/atoms/NotFound/NotFound.tsx
```diff
@@ -9,6 +9,8 @@ import SPARKLEVERSE_LANDSCAPE from "assets/images/Sparkle_logo_white_yellow.png"
 
 import "./NotFound.scss";
 
+const JUST_CURRENT_PAGE_HISTORY_LENGTH = 1;
```
colinhowe, line 12: No change required. Just a thought for you.

I know we love constants. In this case, I don't think this constant actually provides any value to us. It's not something we're ever going to change unless browsers change their history behaviour in a massive way.

src/components/molecules/NavBar/NavBar.tsx
```diff
@@ -60,12 +60,14 @@ export interface NavBarPropsType {
...
   withSchedule?: boolean;
   withPhotobooth?: boolean;
+  hideLoginButton?: boolean;
 }
 
 export const NavBar: React.FC<NavBarPropsType> = ({
   hasBackButton,
   withSchedule,
   withPhotobooth,
+  hideLoginButton,
```
colinhowe, line 70: Should this have a default value?

colinhowe (approved):
I've left some comments. No changes are actually required though. Approved!

src/components/atoms/NotFound/NotFound.tsx
goran-peoski-work: IMHO Constants aren't just for "change it in one place, see it reflected everywhere", they're also about providing meaningful names to values like 1, 3.14, "www.wherever.com"... 

Example, seeing   `NOT_FOUND === [].findIndex()` vs `-1 === [].findIndex()`, we know the -1 result will never change the meaning.

src/components/molecules/NavBar/NavBar.tsx
goran-peoski-work: Yes please, also, maybe in the vein of `hasLoginButton` or `withHiddenLoginButton` instead of a verb. Is it a function?

src/components/atoms/NotFound/NotFound.tsx
colinhowe: Agree with "providing meaning" in most cases. In these specific cases, I'd hope the engineer knows what that constant means in that situation without needing explanation.

colinhowe: Agree with "providing meaning" in most cases. In these specific cases, I'd hope the engineer knows what that constant means in that situation without needing explanation.

PR #2520, opened 2021-11-04
Global chat reply in thread fixes

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1359
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1360


colinhowe (approved):

src/components/molecules/Chatbox/Chatbox.tsx
```diff
@@ -72,9 +73,8 @@ const _ChatBox: React.FC<ChatboxProps> = ({
     async ({ text, threadId }) => {
       await sendThreadMessage({ text, threadId });
       unselectOption();
-      closeThread();
     },
-    [unselectOption, closeThread, sendThreadMessage]
```
mike-lvov, line 77: I still think you should leave this, because the requested behaviour was to `unselect thread` after you send a message into the thread

mike-lvov (changes requested):

src/components/molecules/Chatbox/Chatbox.tsx
sunny-viktoryia: @mike-lvov, I've moved `closeThread` to `unselectOption` function which is called here too.

mike-lvov: Ah, my bad. Looks good then!

mike-lvov (approved):
LGTM

PR #2524, opened 2021-11-04
Do not set the 1st default avatar as a profile picture on Enter click

sunny-viktoryia:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1377

explanation: https://stackoverflow.com/a/12914700


colinhowe:
I think the commit would be improved if it explained why the change is being made. 

Something like this would work:
> The reason for this change is because a button inside a form has its type implicitly set to submit. The HTML spec says that the first button or input with type="submit" is what is triggered when the user hits the enter key

colinhowe (approved):
I'd love to see a minor improvement to the commit message before this goes out, please.

Something like this would work:

> The reason for this change is because a button inside a form has its type implicitly set to submit. The HTML spec says that the first button or input with type="submit" is what is triggered when the user hits the enter key

Then the next person who comes along to this and looks at the history will know why it's like this.

sunny-viktoryia:
@colinhowe, I'll add the description to the squash commit while merging to `staging` branch, does that sound ok to you?

PR #2531, opened 2021-11-05
check world slug only on creation

denisdimitrov:
Fixes the world slug validation to work only when creating a world and not when updating a world.

Unblocks the world update.


goran-peoski-work (approved):
LGTM

PR #2534, opened 2021-11-05
Lowest effort mixpanel bug fix

goran-peoski-work:
Add `try{}catch(e){ console.error(); }` as the simplest measure of not letting mixpanel break the pages.

sparkle-error-01


sunny-viktoryia (approved):
approving. 🙃

PR #2536, opened 2021-11-08
Stop infinite objectEntries() recursion due to value being DOM object

goran-peoski-work:
**Before**:
sparkle-infinite-recursion-02

**After**:
sparkle-infinite-recursion-01


denisdimitrov (approved):
LGTM!

PR #2537, opened 2021-11-08
`World` interface move, `VenuePartygoers` small refactor, streamline of `$shades-enum` and `$alphas-enum`

goran-peoski-work:
Doesn't really resolve:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1234

since it appears to already have the desired effect.

Instead, a light refactor.

Updates the  `World` interface location and the `VenuePartygoers` component display. 
Did require a bit of light touch on the colors and shades modifiers 
(values are a bit more predictable now) :

```scss
$shades-enum: (5, 10, 15, 20, 50);
$alphas-enum: (0.1, 0.2, 0.25, 0.3, 0.4, 0.5, 0.6, 0.7, 0.75, 0.8, 0.9);
```


colinhowe (approved):
Looks good to me.

Only thought is that the commit message itself could be better. I'd probably copy the message from the PR on to the commit. Makes it easy for whoever is looking at the history later:

> A bit of light touch on the colors and shades modifiers (values are a bit more predictable now)

denisdimitrov (approved):
LGTM!

PR #2539, opened 2021-11-09
Use showSchedule defaults for world, remove from venue

goran-peoski-work:
Partially resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1240

1. When `showSchedule` is missing, it defaults to on (like when first creating or entering a world/space).
2. No longer `showSchedule` in the venue object
3. With the current `NavBar` logic, the schedule is shown in the admin, with some layout issues when expanded
sparkle-show-schedule-01


colinhowe:
@goran-peoski-work the code looks fine. As you've commened, it's a mess in the UI. Are you able to fix that? Probably just needs a CSS tweak if it's open in the admin panel.

goran-peoski-work:
@colinhowe I will leave it for a next PR should the answer not come soon. Already noted in the issue
- https://github.com/sparkletown/internal-sparkle-issues/issues/1240#issuecomment-964077898

colinhowe (approved):

functions/world.js
```diff
@@ -158,7 +158,7 @@ exports.updateWorld = functions.https.onCall(async (data, context) => {
     ...(!isNil(requiresDateOfBirth) && { requiresDateOfBirth }),
     ...(!isNil(rooms) && { rooms }),
     ...(!isNil(showNametags) && { showNametags }),
-    ...(!isNil(showSchedule) && { showSchedule }),
+    ...{ showSchedule: isNil(showSchedule) ? true : showSchedule },
```
denisdimitrov, line 161: Are we 100% sure that we want to default to `true` when `showSchedule` is `null` or `undefined`? Also if we do, we should probably use a `DEFAULT_` value here.

goran-peoski-work: Yes, that was the requirement (it kind of hinted here https://github.com/sparkletown/internal-sparkle-issues/issues/1240#issuecomment-958066249 but was explicitly said in a meeting) .

The `DEFAULT_` I think is used in the TS files, not sure if there is anything like it in the cloud functions, I mean, even the old `venue.showSchedule` used a `true` literal 🤷

goran-peoski-work: I have one or two PRs creating a new file `./src/settings/worldDefaultSettings.ts` and after we have all we need there it should be simple enough to transfer the knowledge to an analogue on the BE.

denisdimitrov (approved):
LGTM!

ermiaSparkle (approved):
QA passed and ready to merge!

PR #2543, opened 2021-11-09
introduce useWorldBySlug hook for world searching

denisdimitrov:
Adds a new hook for world searching by `worldSlug`. 

This hook will be required for the new in venue url `/in/:worldSlug/:venueSlug`

### DIFF
--- a/src/hooks/worlds/useWorldBySlug.ts
+++ b/src/hooks/worlds/useWorldBySlug.ts
@@ -1,0 +1,68 @@
+import { useFirestore, useFirestoreCollectionData } from "reactfire";
+import Bugsnag from "@bugsnag/js";
+
+import { World } from "api/admin";
+
+import { worldConverter } from "utils/converters";
+import { WithId } from "utils/id";
+
+type UseWorldEditResult = {
+  world?: WithId<World>;
+  isLoaded: boolean;
+};
+
+/**
+ * Hook which will return the world when the slug is provided.
+ * The intention is to be used on the client side, when the world slug is provided in the url.
+ * @param worldSlug
+ * @returns
+ */
+export const useWorldBySlug: (worldSlug?: string) => UseWorldEditResult = (
+  worldSlug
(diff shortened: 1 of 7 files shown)


src/hooks/worlds/useWorldBySlug.ts
```diff
@@ -0,0 +1,68 @@
...
+    {
+      initialData: undefined,
+    }
+  );
+
+  const isWorldLoaded = status !== "loading";
+
+  if (!worlds?.[0]) {
+    Bugsnag.notify(
+      `World with the following slug: ${worldSlug} does not exist.`,
```
colinhowe, line 41: I think we're going to see this a lot with bots crawling the site etc. Might be worth just throwing an error that eventually results in a 404

colinhowe (approved):
Looks good. Just a comment around error handling

goran-peoski-work:
I think this is a good opportunity to rename the other one `useWorldById()`, makes it easy for other PR's to just merge 2 lines change in 15 files because of rename.

src/hooks/worlds/useWorldBySlug.ts
goran-peoski-work: My initial reaction was mostly about spamming our Bugsnag, but I kind of let it go to first check if that causes an issue, I mean, the solution would be a simple replacement of `Bugsnag.notify()` with `console.error()` just to log it somewhere.

I don't think relying on errors unwinding stack should be the normal behavior though, but an exception to handle in a nicer way.

denisdimitrov:
@colinhowe The error is now thrown 

https://github.com/sparkletown/sparkle/pull/2543/commits/ae1cb4f71c7508b74045778ed2404e4357611e4f

@goran-peoski-work The other hook is now called `useWorldById`


https://github.com/sparkletown/sparkle/pull/2543/commits/6ef584e767fb064eeba20941eb2abc1d60495137

PR #2544, opened 2021-11-10
Add flip-switch vaiant to AdminCheckbox and fix labeless click bug

goran-peoski-work:
Previously `AdminCheckbox` would not toggle its state if label wasn't present ( `label={<></>}` was a workaround ). Now...

ON / OFF displays, no label:
sparkle-flip-switch-02

ON / OFF displays, with label above:
sparkle-flip-switch-01


alexey-petrov (approved):
LGTM

PR #2549, opened 2021-11-11
fix auditorium form buttons and sections

denisdimitrov:
- Delete button now shows indicator only on deleting, and disabled on deleting or updating
- Save button now shows indicator only on updating, and disabled on deleting or updating
- Auditorium form and it's sections will now update correctly

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1449


functions/venue.js
```diff
@@ -842,7 +842,19 @@ exports.updateVenueNG = functions.https.onCall(async (data, context) => {
   // @debt updateVenue uses checkUserIsOwner rather than checkUserIsAdminOrOwner. Should these be the same? Which is correct?
   await checkUserIsOwner(data.id, context.auth.token.user_id);
 
-  const updated = {};
+  if (!data.worldId) {
+    throw new HttpsError(
+      "not-found",
+      "World id is missing and the update can not be executed."
```
goran-peoski-work, line 848: [NOTE]
We shouldn't send user facing messages from BE, but since it is the case, I guess `Id` is the "proper English" spelling

functions/venue.js
```diff
@@ -842,7 +842,19 @@ exports.updateVenueNG = functions.https.onCall(async (data, context) => {
...
+    );
+  }
+
+  const updated = {
+    config: {
+      landingPageConfig: {},
+    },
+  };
```
goran-peoski-work, line 856: Isn't there a Lodash merge object utility we can use in this and similar cases?

goran-peoski-work (approved):
LGTM

functions/venue.js
denisdimitrov: There should be, yes. But ideally this should look like the world object with either `isNil` or `isEmpty`

functions/venue.js
denisdimitrov: I guess I can remove it, but If world id is missing this function will probably return nothing, and look weird on the client side.

goran-peoski-work: This is more like a note for a longer term consideration

functions/venue.js
goran-peoski-work: More like https://lodash.com/docs/#merge that does what the Firestore merge flag does, only inside the function, before calling the cloud API:

> like _.assign except that it recursively merges own and inherited enumerable string keyed properties of source objects into the destination object

PR #2553, opened 2021-11-12
Space setting parent selector list exclude outside spaces

alexey-petrov:
Closes a minor bug in:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1282

Fixed an error with `worldId` being undefined resulting in an empty parent list


goran-peoski-work (approved):
LGTM

PR #2558, opened 2021-11-15
Fix issues with World entrance form and actual entrance to spaces of that world

goran-peoski-work:
Resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1451

by:

1. `useArray()` hook keeps track of the data
2. the data in the code and profile questions is being updated on input's change
3. the data is now saved to DB
4. the form is now reset correctly from the data (comming from `useArray()`), not the input (comming from `useForm()`)
5. streamline entrance pages (URLs starting with `/account/` and `/e/:step/`)
6. save `convertToEmbeddableUrl()` from breaking the page on error (log in console and Bugsag instead)
7. entrance steps and their buttons handled in the same manner as the questions


colinhowe (approved):

src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
```diff
@@ -31,35 +37,93 @@ export const EntranceStepsInputFieldSet: React.FC<EntranceStepsInputFieldSetProp
...
+  const inputButtons = `${fieldset}${fieldButtons}`;
+  const inputUrl = `${fieldset}${fieldUrl}`;
+  const inputTemplate = `${fieldset}${fieldTemplate}`;
+
+  const handleRemove = useCallback(() => onRemove({ index }), [
+    onRemove,
+    index,
+  ]);
+
+  const handleChange: ChangeEventHandler<HTMLInputElement> = useCallback(
```
colinhowe, line 57: Could we extract out this code to remove the duplication?
This code is in EntranceButtonsInputFieldSet.tsx  as well. And it's in a few other places too

colinhowe (changes requested):
Minor change. Not mandatory.

colinhowe (approved):

src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
goran-peoski-work: It can be DRY-ed up or even replaced with Redux, now that ti's obvous what the problem is. For next PR though

PR #2559, opened 2021-11-15
Disable reactions audio button according to flag

alexey-petrov:
Closes:

- https://github.com/sparkletown/internal-sparkle-issues/issues/1214

Added noop + disabled styles if the `isReactionsMuted` is set to `true` in the admin
image

### DIFF
--- a/src/components/templates/Jazzbar/JazzBar/JazzBar.tsx
+++ b/src/components/templates/Jazzbar/JazzBar/JazzBar.tsx
@@ -77,6 +77,8 @@ export const JazzBar: React.FC<JazzProps> = ({ venue }) => {
     seatedAtTable && venue?.id
   );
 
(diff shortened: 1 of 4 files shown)


src/components/templates/Jazzbar/JazzBar/JazzBar.tsx
```diff
@@ -77,6 +77,8 @@ export const JazzBar: React.FC<JazzProps> = ({ venue }) => {
     seatedAtTable && venue?.id
   );
 
+  const isReactionsAudioDisabled = !venue.isReactionsMuted;
```
goran-peoski-work, line 80: [NOTE]
Not really providing any more clarity than `isAudioDisabled={!venue.isReactionsMuted}`, just adds few lines.
It is fine either way.

goran-peoski-work (approved):
LGTM.

Code wise. Didn't test drive locally.

PR #2561, opened 2021-11-16
Makes algolia search work in other envs

mike-lvov:
Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1443

Algolia search didn't work in any env except for staging because of the hardcode of the app id


sunny-viktoryia (approved):
approving. 🙃

PR #2563, opened 2021-11-17
remove chat label feature

alexey-petrov:
Closes:

- https://github.com/sparkletown/internal-sparkle-issues/issues/1422

Removed chat label feature. Rename custom labeled chat title into 'Chat' as per `step #2` - https://github.com/sparkletown/internal-sparkle-issues/issues/1418


goran-peoski-work (approved):
LGTM

PR #2564, opened 2021-11-17
Schedule bug fixes and text adjustments

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1358

It also:
* fixes appearance of the weekdays bar in case when "end schedule date" is in the past

BEFORE
image

AFTER
image


src/hooks/useVenueScheduleEvents.ts
```diff
@@ -95,7 +96,11 @@ const useVenueScheduleEvents = ({
     [liveAndFutureEvents, firstRangeDateInSeconds]
   );
 
-  const endScheduleDate = sovereignVenue?.end_utc_seconds;
+  const endScheduleDate =
+    sovereignVenue?.end_utc_seconds &&
+    isFuture(fromUnixTime(sovereignVenue.end_utc_seconds)) &&
+    sovereignVenue?.end_utc_seconds;
```
sunny-viktoryia, line 102: fixes the appearance of the weekdays' bar in the case when end_utc_seconds time has been already passed.

src/components/molecules/ScheduleEventListNG/ScheduleEventListNG.tsx
```diff
@@ -51,17 +52,20 @@ export const ScheduleEventListNG: React.FC<ScheduleEventListNGProps> = ({
       <div className="ScheduleEventListNG">
         <ScheduleEventSubListNG
           events={daysEvents}
-          title={`Events on ${formatDateRelativeToNow(scheduleDate)}`}
+          title={`Coming up ${formatDateRelativeToNow(scheduleDate, {
+            formatTomorrow: () => "tomorrow",
+            formatOtherDate: (date) => "on " + format(date, "do MMM yyyy"),
```
alexey-petrov, line 57: might as well move this `format(date, "do MMM yyyy")` into some util function that accepts `date` as an argument

src/hooks/useVenueScheduleEvents.ts
alexey-petrov: my eyes might be betraying me, but it sure looks like a duplicate 
image

alexey-petrov (changes requested):
1 suggestion and 1 duplicate to remove

src/hooks/useVenueScheduleEvents.ts
sunny-viktoryia: there is logic. 🙃 
the meaning here is that if `sovereignVenue` and `sovereignVenue.end_utc_seconds` defined and it's in future, then `endScheduleDate = end_utc_seconds`, otherwise `undefined` or `false`.
But I agree, I should better do that explicitly. 

done here: https://github.com/sparkletown/sparkle/pull/2564/commits/38330a74973238bd651ff79fe21023785e3e8112

src/components/molecules/ScheduleEventListNG/ScheduleEventListNG.tsx
sunny-viktoryia: not sure why should we wrap library functions into ours. I guess it's more obvious to use it the way it's now. Moreover, there is no extra meaning: just a needed format.

But if you insist, of course, I can do that.

alexey-petrov (approved):
LGTM

PR #2570, opened 2021-11-18
Consolidate form validation schemas into their own directory `./src/forms`

goran-peoski-work:
In order to be easier for maintenance as well as code reuse, moved some validation schemas into a separate directory. This way, something called `commonTitleSchema` can be reused with several forms having title validated. 

Similar to how 
- https://github.com/sparkletown/internal-sparkle-issues/issues/1450

requires the same validation for name on different objects (world and space).


src/forms/common.ts
```diff
@@ -0,0 +1,14 @@
+export interface SpaceSchema {
```
alexey-petrov, line 1: shouldn't this file be called `commonSpaceSchema` ?

src/forms/eventEditSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { EventInput } from "api/admin";
+
+export const eventEditSchema = Yup.object().shape<EventInput>({
+  name: Yup.string().required("Name required"),
+  description: Yup.string().required("Description required"),
+  start_date: Yup.string()
+    .required("Start date required")
+    .matches(
+      /\d{4}-\d{2}-\d{2}/,
```
alexey-petrov, line 11: probably OSS, but I think it would be good to have some sort of a `config` file with all the regex/constants etc listed in

src/forms/venueV2Schema.ts
```diff
@@ -0,0 +1,57 @@
...
+                "Must have alphanumeric characters",
+                (val: string) => createSlug(val).length > 0
+              )
+              .test(
+                "name",
+                "This venue name is already taken",
+                async (val: string) =>
+                  !val ||
+                  !(
+                    await firebase
```
alexey-petrov, line 36: we might add smth like `// @debt Replace with a function from api/venues` as done in `src/forms/worldStartSchema.ts`

alexey-petrov (commented):
A few minor things, mostly improvement suggestions

src/forms/common.ts
goran-peoski-work: It's a schema for the interface Space, it isn't a schema for a field that might happen on multiple. 
I just added the common as a moniker to shared resources between the form validation schemas/functions.

Either way, more changes can be done later, this one should probably be less limited as to not cause too many merge conflicts.

src/forms/eventEditSchema.ts
goran-peoski-work: there are already, `validationSettings.ts` (in this PR) and `dateSettings.ts` (from staging already) and more can be utilized as needed.

goran-peoski-work: This PR should help transition to more refactors later on, but for now the most important thing is to not cause too many merge conflicts (one PR for moving files, next one for moving RegExps, etc.)

src/forms/venueV2Schema.ts
goran-peoski-work: Even better, add it to
- https://github.com/sparkletown/internal-sparkle-issues/issues/1462

goran-peoski-work: Even better, add it to
- https://github.com/sparkletown/internal-sparkle-issues/issues/1462

goran-peoski-work: btw, is smth = shaking my head variation?

alexey-petrov: nah, i'm using it as `something` 
image

alexey-petrov (approved):
LGTM

src/forms/venueV2Schema.ts
goran-peoski-work: `something` reads better than mistakenly getting `smth` as `shaking my thick head` or worse 🤪

PR #2572, opened 2021-11-19
fixed user audio update on admin change

alexey-petrov:
Closes:

- https://github.com/sparkletown/internal-sparkle-issues/issues/1214

Added `useEffect` that changes client-side reactions audio status based on the venue value updated at the admin while having 2 pages open simultaneously


src/components/templates/Auditorium/components/Section/Section.tsx
```diff
@@ -33,9 +33,20 @@ export interface SectionProps {
...
+    hide: disableUserAudio,
+    show: enableUserAudio,
+  } = useShowHide(venue.isReactionsMuted ?? false);
+
+  useEffect(() => {
+    if (!venue.isReactionsMuted) {
+      disableUserAudio();
+    } else {
+      enableUserAudio();
+    }
```
goran-peoski-work, line 48: Can swap the if and else parts to remove the negation in the if

src/components/templates/Auditorium/components/Section/Section.tsx
```diff
@@ -33,9 +33,20 @@ export interface SectionProps {
...
 
-  const { isShown: isUserAudioOn, toggle: toggleUserAudio } = useShowHide(
-    venue.isReactionsMuted ?? false
-  );
+  const {
+    isShown: isUserAudioOn,
+    toggle: toggleUserAudio,
+    hide: disableUserAudio,
+    show: enableUserAudio,
+  } = useShowHide(venue.isReactionsMuted ?? false);
```
goran-peoski-work, line 41: Shouldn't there already be a `DEFAULT_` constant somewhere? There is a `worldDefaultSettings.ts` to find these faster for the world. Maybe something similar should exist for the rest.

goran-peoski-work (approved):
LGTM

src/components/templates/Auditorium/components/Section/Section.tsx
alexey-petrov: added const + imported to use here e291c35

src/components/templates/Auditorium/components/Section/Section.tsx
alexey-petrov: done e291c35

PR #2575, opened 2021-11-19
Update the SpacesDashboard with a title bar

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1459

**Before**:

sparkle-spaces-dash-02

**After**:

sparkle-spaces-dash-01


src/components/organisms/AppRouter/AdminV3Subrouter.tsx
```diff
@@ -11,21 +11,29 @@ import {
   ADMIN_V3_SPACE_SETTINGS_PARAM_URL,
   ADMIN_V3_WORLD_SPACES_PARAM_URL,
   ADMIN_V3_WORLDS_BASE_URL,
+  ADMIN_V4_SPACES_1_PARAM_URL,
```
alexey-petrov, line 14: Wait, when did we introduce admin v4 thing?
and having `spaces 1` and `spaces 2` isn't really self explanatory naming.
Maybe something like `ADMIN_V4_SPACES_WORLD_SLUG_PARAM_URL` and the other ( not sure what `/s` stands for in ``${ADMIN_V3_ROOT_URL}/w/:worldSlug/s``

alexey-petrov (approved):
A comment on variable naming. LGTM tho

src/components/organisms/AppRouter/AdminV3Subrouter.tsx
goran-peoski-work: > Wait, when did we introduce admin v4 thing?

☝️ 
I am marking anything from the IA as v4 since it is no longer about parity. This will make it distinguished from the v3 and v1 stuff.

> and having spaces 1 and spaces 2 isn't really self explanatory naming

☝️ 
True, unless you consider both to be equivalent, so you're stuck with `banana1` and `banana2` since there isn't anything really different. I'm open to any actionable suggestion. Can't find better naming than the above 🤷 


> not sure what /s stands for

☝️ 
It is the requirement here https://github.com/sparkletown/internal-sparkle-issues/issues/1459 as well as Slack DM wish from @ermiaSparkle to have 3 different URLs (one old and two new ones) point to the same page.

If you think about it:
- `/admin/w` 👈 displays all worlds
- `/admin/w/:worldSlug` 👈 displays single world
- `/admin/w/:worldslug/s/:spaceSlug` 👈 displays single space config

thus, by the above logic:
- `/admin/w/:worldSlug` 👈 displays single world which is basically all the spaces it contains
- `/admin/w/:worldSlug/s` 👈 displays all spaces under a world

PR #2582, opened 2021-11-22
Add max 50 characters check for World/Space names w/ a bit of `./src/forms` refactor

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1450

Due to quirks of how `useForm()` hook works of the old https://react-hook-form.com/ version, right now the checks are only done on form submit, and not on individual inputs' change.

1. Refactored schemas to have reusable `createNameSchema()`
2. Reused the above in
  -  `detailsSchema`, 
  - `spaceEditNGSchema`, 
  - `spaceEditSchema`, 
  - `venueDetailsCreateSchema`
  - `venueNameSchema`
  - `worldStartSchema`
3. Added ` max={COMMON_NAME_MAX_CHAR_COUNT}` to forms 
  - `WorldStartForm`
  - `DetailsForm`
  - `VenueDetailsSubForm`
 4. Delete unused duplicate schemas


src/forms/eventEditSchema.ts
```diff
@@ -8,7 +10,7 @@ export const eventEditSchema = Yup.object().shape<EventInput>({
   start_date: Yup.string()
     .required("Start date required")
     .matches(
-      /\d{4}-\d{2}-\d{2}/,
+      START_DATE_FORMAT_RE,
```
alexey-petrov, line 13: *applause*

src/settings/validationSettings.ts
```diff
@@ -1,2 +1,19 @@
...
+
+export const COMMON_STRING_MIN_CHAR_COUNT = 3;
+export const COMMON_STRING_MIN_CHAR_COUNT_RE = new RegExp(
+  `.{${COMMON_STRING_MIN_CHAR_COUNT},}`
+);
+
+export const START_DATE_FORMAT_RE = /\d{4}-\d{2}-\d{2}/;
+
+export const DEFAULT_MAP_ICON_PLACEMENT: Omit<VenuePlacement, "state"> = {
+  x: (PLAYA_WIDTH - PLAYA_VENUE_SIZE) / 2,
```
alexey-petrov, line 16: should these map settings be imported from the `playa` named file? If those settings are shared between all `map` like templates we should move them to general settings?

alexey-petrov (approved):
A question on config import. Otherwise LGTM

src/settings/validationSettings.ts
goran-peoski-work: I checked, they aren't really. Could've just as easily put them in `validationSettings.ts` but I figured it would be easier to just delete the file later once playa settings aren't necessary.

goran-peoski-work: To be more clear: most of those playa settings can be marked as deprecated and/or outright deleted. But I didn't want to spend the time to get the all clear on this PR that already changes 22 files.

Still more schemas are around the project, not inside `./src/forms` and marked in https://github.com/sparkletown/internal-sparkle-issues/issues/1462 so these playa settings can also go there.

PR #2585, opened 2021-11-23
added option none to space parent selector

alexey-petrov:
Closes:

- https://github.com/sparkletown/internal-sparkle-issues/issues/1461

Added "None" option to spaces/portals back button selector

image


functions/venue.js
```diff
@@ -828,7 +827,7 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
   const venuesRef = await admin
     .firestore()
     .collection("venues")
-    .where("slug", "==", data.slug)
+    // .where("slug", "==", data.slug)
     .where("worldId", "==", data.worldId)
```
denisdimitrov, line 831: This will not work because it will always return the same venue doc for all of the venues in that world, because of the logic below `docs[0]`

denisdimitrov (dismissed):

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -41,6 +44,11 @@ export const SpacesDropdown: React.FC<SpacesDropdownProps> = ({
   // doesn't exist on SPACE_PORTALS_ICONS_MAPPING
   const filteredPortals = omit(portals, VenueTemplate.playa);
 
+  const portalOptions = useMemo(
+    () => ({ ...filteredPortals, ...spaceNoneOption }),
```
denisdimitrov, line 48: When I've seen this `None` option on other websites, usually it's in the beginning of the dropdown, not at the end. If this is per design, dismiss my comment. Otherwise, I think we should ask product.

alexey-petrov: thanks, put `none` to the top of the list 06cbf3a

functions/venue.js
alexey-petrov: made a mistake by committing this change, thanks for noticing it! 06cbf3a

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -14,6 +14,9 @@ import { Dropdown } from "components/atoms/Dropdown";
 
 import "./SpacesDropdown.scss";
 
+const noneOptionName = "None";
+const spaceNoneOption = { none: { name: "", template: undefined } };
```
goran-peoski-work, line 18: This is a mutable object passed to a component which after few renders might or might not pass it some place else or modify it on its own.

Yes, with a close inspection one might determine that is not the case, but there is a quick and easy way of marking it as actual constant: `Object.freeze()` i.e. something like:

```suggestion
const spaceNoneOption = Object.freeze({ none: Object.freeze({ name: "", template: undefined }) } );
```
or like:
```suggestion
const spaceNoneOption = { none: { name: "", template: undefined } };
Object.freeze(spaceNoneOption);
Object.freeze(spaceNoneOption.none);
```

goran-peoski-work: The above is just a note, I don't think it merits holding back the PR, just for future consideration

alexey-petrov: done ef728b1

colinhowe (approved):
The code looks good now and I can see you've addressed the changes requested.

PR #2586, opened 2021-11-23
fixes the stretched images in spaces preview

denisdimitrov:
- An unnecessary `img` was being rendered on top of the div which had the correct styles.
- Rooms can be always selected and you don't have to click on `back` to select a new one.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1470


mike-lvov (approved):
LGTM

PR #2589, opened 2021-11-24
Separate PortalList component and add helper API functions and sub-components

goran-peoski-work:
Partially resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1466

1. The parts that deal with selecting a Portal are now extracted into separate component.
  - The component now has 2 variants `input` and `modal`
  - `modal` is used in the old place of adding a portal to an existing space
  - `input` variant can be used as an elaborate input field inside some other form
2. Some updates to the creation of portal are also made
3. For the issue above, subsequent PRs can reuse the `PortalAddModal` form logic as a copy-paste or extract into helper hook e.g. `useFormCreateSpace()`


src/settings/apiSettings.ts
```diff
@@ -0,0 +1,7 @@
+// constants for the collections, just in case some get renamed (e.g. venues->spaces)
+export const COLLECTION_SPACES = "venues";
+export const COLLECTION_WORLDS = "worlds";
```
denisdimitrov, line 3: [NB] In my opinion, it would be better if these are in another file.

src/components/molecules/PortalList/PortalList.scss
```diff
@@ -0,0 +1,4 @@
+.PortalList {
+  margin: 0;
+  padding: 0;
```
denisdimitrov, line 3: Is this overriding some default values already set by our global.scss or the library itself?

src/components/molecules/PortalAddModal/PortalAddModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+      const venueData = buildEmptyVenue(venueName, template);
+      const newSpaceSlug = createSlug(venueName);
+
+      await createVenue_v2(
+        {
+          ...venueData,
+          worldId,
+          parentId: spaceSlug, // @debt this should be parent's spaceId and if needed parentSlug: spaceSlug should be added
+          logoImageUrl: icon,
+          name: newSpaceSlug, // @debt this should be unfiltered venueName once the possible bugs of mixing slug with id are resolved
```
denisdimitrov, line 92: This is probably going to convert my name to a slug and display it in the portal name differently than expected, no?

If I add `Denis Party 123` in the portal input and create a space. This will convert it and change the name of my space to `denisparty123`

src/api/space.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  let query: firebase.firestore.Query = firebase
+    .firestore()
+    .collection(COLLECTION_SPACES)
+    .where(FIELD_SLUG, "==", spaceSlug);
+
+  if (worldId) {
+    query = query.where(FIELD_WORLD_ID, "==", worldId);
+  } else if (worldSlug) {
```
denisdimitrov, line 32: Can you move the statement inside `else if` outside below. It will improve the readability.

denisdimitrov (changes requested):
Mostly nitpicks.

src/settings/apiSettings.ts
goran-peoski-work: Maybe, couldn't figure our where exactly, so this was the most visible place. Just a preparation for the time "rename venues to spaces in the DB" comes to be worked on.

src/components/molecules/PortalList/PortalList.scss
goran-peoski-work: Should usually be overriding browser defaults.

src/components/molecules/PortalAddModal/PortalAddModal.tsx
goran-peoski-work: > If I add Denis Party 123 in the portal input and create a space. This will convert it and change the name of my space to denisparty123

Yes, this is still the current behavior. name was patched over a month ago to serve as a slug and until it is no longer used as such this prevents new bugs of appearing.

src/api/space.ts
goran-peoski-work: If both `worldId` and `worldSlug` are provided, only `worldId` is used and if neither is, error is thrown.

How can that move be done by keeping the same logic and improving readability?
I originally had fail fast checks above, didn't add much to readability and TS complained.

goran-peoski-work: How does 5fe64af99f91b6d7667cf7f18b0b390899632553 look?

denisdimitrov (approved):
LGTM!

PR #2590, opened 2021-11-24
fix updateVenue_v2 space update

alexey-petrov:
Refer to https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1637746230374000

Fixes an issue with space update using `updateVenue_v2`, which returned an error on update


goran-peoski-work (approved):
LGTM

functions/venue.js
```diff
@@ -835,7 +835,7 @@ exports.updateVenue_v2 = functions.https.onCall(async (data, context) => {
   const venueRef = venuesRef.docs && venuesRef.docs[0];
 
   if (venueRef) {
-    await venueRef.update(updated);
```
denisdimitrov, line 838: This same syntax is used in a couple of places and it might be broken there too, can you please check?

alexey-petrov: yep, updated it in the other 2 places b4cad48

PR #2593, opened 2021-11-24
fix BackButton visibility and linkability issues (take 2)

goran-peoski-work:
Resolves 
- https://github.com/sparkletown/internal-sparkle-issues/issues/1487

by refactor of `<BackButton>` so it
- doesn't show up in admin context
- uses the proper slug to go back
- compiles 3 different modes of going back from parent components via its new `variant` prop

Bonus:
- fix for space slug -> space id read from portal URLs in order to test Auditorium back button

sparkle-back-button-02
sparkle-back-button-01


denisdimitrov (approved):
LGTM!

PR #2597, opened 2021-11-25
Stop showing the horizontal scrollbar on the room modal

### LINKED ISSUE #1505: Markdown renderer needed for event modal

colinhowe:
**AV1 only**

Resolves #1505

No horizontal scrollbar anymore:
image

### DIFF
--- a/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
+++ b/src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
@@ -79,6 +79,7 @@ $spacing: 20px;
 
   &__events {
     max-height: 35vh;
+    overflow-x: hidden;
     overflow-y: auto;
     // @debt these overrides are here because ScheduleItem sets incompatible margins/etc.
     //   We should just clean it up there instead.


src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
```diff
@@ -79,6 +79,7 @@ $spacing: 20px;
 
   &__events {
     max-height: 35vh;
+    overflow-x: hidden;
```
goran-peoski-work, line 82: This could make it an issue of clipping the content that was making the horizontal scrollbar show up. 

Is the text (or whichever items are displayed in it) doing its re-flow automatically or is it clipped?

goran-peoski-work (commented):
Need more info

src/components/templates/PartyMap/components/RoomModal/RoomModal.scss
colinhowe: They get re-flowed as in the below screenshot. I would track down exactly what is causing the horizontal scroll to appear, but, we want to do the bare minimum with AV1 for the upcoming parties and then get rid of it.

image

goran-peoski-work: Sounds reasonable.

goran-peoski-work (approved):
LGTM

PR #2599, opened 2021-11-25
Update some of the copy around events->experiences

colinhowe:
Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/1491

This PR:
 * updates the copy on this modal
 * fixes the spacing on the date/time selector (minimal change as there is an upcoming issue that improves the layout of this form more)
 * increases the width of the form so that the new date/time label fits on

**Before**

**After**


goran-peoski-work:
No issue related? Maybe should add this PR to the board

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -158,7 +158,7 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
               </div>
 
               <div className="input-group">
-                <label htmlFor="name">Name your event</label>
+                <label htmlFor="name">Name your experience</label>
```
goran-peoski-work, line 161: Maybe `EVENT_TAXON` should be added on https://github.com/sparkletown/sparkle/blob/staging/src/settings/taxonomy.ts and used here and in other places.

It is currently KISS that may allow us quick search through and update later on if/when we add proper externalization of strings.

src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -16,6 +17,12 @@
     background-color: $light-grey;
   }
 
+  .model-content,
+  .modal-dialog {
+    width: 700px;
+    max-width: 700px;
+  }
+
```
goran-peoski-work, line 25: Would it be useful to add media query for narrower display? 

Most queries in the project use:
```scss

  @media (max-width: $media-breakpoint--lg) {
   }

  @media (max-width: $media-breakpoint--md) {
  }
```

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -201,26 +201,35 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
-                <label htmlFor="date">Starting time</label>
-                <label>
-                  When does this event start? Use your local time zone, it will
-                  be automatically converted for anyone visiting from around the
-                  world.
+                <label htmlFor="date">
+                  Start date and time (use your own time zone; it will be
+                  automatically localized)
                 </label>
+                <label></label>
```
goran-peoski-work, line 208: [NOTE]
This is just legacy way of "doing business". The next PR can re-use some of `SpaceTimingForm` components: 

```tsx
          <AdminDateTime
            name="end"
            register={register}
            errors={errors}
            label="Global ending time"
          />
```

There are few more props in https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/AdminDateTime/AdminDateTime.tsx#L16

goran-peoski-work (approved):
LGTM.

Can be better, but not a problem even if merged as is

src/components/organisms/TimingEventModal/TimingEventModal.scss
colinhowe: Given we're going to be doing a redesign I'm not sure it's worth putting the effort into making this work on smaller devices. Anything below 700px is going to have a terrible admin site experience at the moment regardless :(

src/components/molecules/AdminDateTime/AdminDateTime.tsx
```diff
@@ -33,9 +35,11 @@ export const AdminDateTime: React.FC<DateFieldProps> = ({
   errors,
   dateProps,
   timeProps,
+  dateFieldName,
+  timeFieldName,
 }) => {
-  const nameDate = `${name}Date`;
-  const nameTime = `${name}Time`;
+  const nameDate = dateFieldName || `${name}Date`;
+  const nameTime = timeFieldName || `${name}Time`;
```
goran-peoski-work, line 42: This makes it difficult to standardize across different forms. 

```suggestion
  const nameDate = `${name}Date`;
  const nameTime = `${name}Time`;
```

More unified approach would be for the form itself to take on the convention of `nameDate` and `nameTime` 
or just not use this component for the time being, hence the [NOTE] part, might as well have used [OOS]

> [NOTE]
This is just legacy way of "doing business". The next PR can re-use some of SpaceTimingForm components

https://github.com/sparkletown/sparkle/pull/2599#discussion_r757031336

src/components/organisms/TimingEventModal/TimingEventModal.tsx
colinhowe: https://github.com/sparkletown/sparkle/pull/2599/commits/7eb452fae9982ead81affd8ed7ad8664c6fb813e

src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -40,4 +40,25 @@
...
+  // This is to combat the use of a h2 tag in the wrong place. Longer term
+  // we probably want to NOT use AdminSection inside this form.
+  .AdminSection {
+    h2 {
+      font-weight: $font-weight--400;
+      font-size: font-size--root();
+    }
+  }
```
goran-peoski-work, line 63: Maybe not the correct time to use `AdminDateTime` component inside this form if the more specific component (form) can't be adjusted accordingly, but the more general one (AdminDateTime) must accommodate its specificities.

src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -40,4 +40,25 @@
...
+
+  .AdminDateTime__input-wrapper {
+    padding: 0;
+    input {
+      border: 0.1px solid var(--accent--over-20a);
+    }
+  }
+
+  // This is to combat the use of a h2 tag in the wrong place. Longer term
```
goran-peoski-work, line 56: Why is `h2` in the wrong place?
Does the page have no `h1`?
Does the `h2` come before or after some other `h` tags that make it inappropriate?

goran-peoski-work (changes requested):
None of the comments were blockers. Use of the taxon might have been a good addition (https://github.com/sparkletown/sparkle/pull/2599#discussion_r757027428) , but doing updates to the AdminDateTime to accomodate TimingEventModal is adding more debt than removing.

src/components/organisms/TimingEventModal/TimingEventModal.scss
colinhowe: It's basically a label in a form. Visually it looks wrong and not like the rest of the form.

goran-peoski-work (approved):
LGTM

PR #2602, opened 2021-11-26
Add full set of params to useSpaceParams

goran-peoski-work:
There are 2 slugs needed to properly address a space. This PR updates the use of `useSpaceParams()` for it.

Before:
```tsx
  const spaceSlug = useSpaceParams();
```

After:
```tsx
  const { spaceSlug, worldSlug } = useSpaceParams();
```


colinhowe (approved):
Wonderful!

PR #2604, opened 2021-11-26
Use shouldScheduleBeShown() everywhere to have same check

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1475

by making a single check `shouldScheduleBeShown()` everywhere that factors in `DEFAULT_SHOW_SCHEDULE` if the world or the value for `showSchedule` is missing.


colinhowe (approved):
Nice!

PR #2605, opened 2021-11-26
Change the layout of the experience modal

colinhowe:
Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/1513

Note: the delete button placement is temporary and will be getting moved to the list component in the near future.

New visuals:
image

### DIFF
--- a/src/components/organisms/TimingEventModal/TimingEventModal.scss
+++ b/src/components/organisms/TimingEventModal/TimingEventModal.scss
@@ -37,6 +37,7 @@
     max-width: 700px;
   }
 
+  // @debt this should be shorter, not having to compensate for `global.scss` interference. remove when no longer necessary
   &__input-group {
     position: relative;
     display: flex;
--- a/src/components/organisms/TimingEventModal/TimingEventModal.scss
+++ b/src/components/organisms/TimingEventModal/TimingEventModal.scss
@@ -17,7 +17,7 @@
     input {
       display: inline-block;
       width: auto;
-      border: 0.1px solid var(--accent--over-20a);
+      border: $border-width--slim solid var(--accent--over-20a);
       text-align: left;
     }
 
@@ -42,7 +42,7 @@
 
     &__modal-input {
       width: 100%;
-      border: 0.1px solid var(--accent--over-20a);
+      border: $border-width--slim solid var(--accent--over-20a);
       text-align: left;
 
       &--indented {
--- a/src/components/organisms/TimingEventModal/TimingEventModal.scss
+++ b/src/components/organisms/TimingEventModal/TimingEventModal.scss
@@ -37,11 +37,35 @@
     max-width: 700px;
   }
 
-  .input-group {
+  &__input-group {
     position: relative;
+    display: flex;
+    flex-direction: column;
+    margin-bottom: 0.8rem;
+
+    input.input-block {
+      margin-bottom: 0.3rem;
+    }
+
+    .input-info {
+      font-size: 0.8em;
+      opacity: 0.8;
+      align-self: center;
+    }
+
+    &.input-info {
+      align-self: center;
+    }
+
+    .input-error {
+      text-align: center;
+      color: var(--danger--under);
+      font-size: 0.9em;
+      font-weight: bold;
+    }
 
     &__modal-input {
-      width: 100%;
+      width: 100% !important;
       border: $border-width--slim solid var(--accent--over-20a);
       text-align: left;
 
--- a/src/components/organisms/TimingEventModal/TimingEventModal.scss
+++ b/src/components/organisms/TimingEventModal/TimingEventModal.scss
@@ -13,6 +13,20 @@
     }
   }
 
+  &__duration_container {
+    input {
+      display: inline-block;
+      width: auto;
+      border: 0.1px solid var(--accent--over-20a);
+      text-align: left;
+    }
+
+    label {
+      margin-left: 4px;
+      margin-right: 8px;
+    }
+  }
+
   .modal-content {
     background-color: $light-grey;
   }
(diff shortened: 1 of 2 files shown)


src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -13,6 +13,20 @@
     }
   }
 
+  &__duration_container {
+    input {
+      display: inline-block;
+      width: auto;
+      border: 0.1px solid var(--accent--over-20a);
```
goran-peoski-work, line 20: we have 
```scss
$border-width--none: 0px;
$border-width--slim: 1px;
$border-width--bold: 2px;
```

should this `0.1px` be a thing? 
If not, is it a workaround? 
If so, maybe a comment above it why it isn't simply `none`

src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -13,6 +13,20 @@
...
+      text-align: left;
+    }
+
+    label {
+      margin-left: 4px;
+      margin-right: 8px;
```
goran-peoski-work, line 26:
```suggestion
      margin-left: $spacing--xs;
      margin-right: $spacing--sm;
```

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -246,26 +243,25 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
                 </div>
               </div>
 
-              <AdminSection title="Duration">
```
goran-peoski-work, line 249: Why remove `AdminSection` which is semantic and replace it with `<div  className="input-group">` which has a style from `global.scss`?

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -279,17 +275,13 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
-                  variant="primary"
-                  type="submit"
-                >
-                  {event ? "Update" : "Create"}
-                </ButtonNG>
-
+                {/*
+                  Note: this delete button is being moved to the experience
+                  list component in the near future
+                  */}
```
goran-peoski-work, line 284:
```suggestion
                { 
                  // @debt: move this delete button to the experience list component
                 }
```

goran-peoski-work (changes requested):
Suggestions and questions

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -246,26 +243,25 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
-                      className="input-group__modal-input input-group__modal-input--indent"
-                      placeholder="minutes"
-                      ref={register}
-                    />
-                    minutes
-                  </label>
+              <div className="input-group">
+                <label>Duration</label>
+                <div className="TimingEventModal__duration_container">
+                  <input
```
alexey-petrov, line 249: We have `AdminInput` component which should fit fine and it's a general use component that we're reusing in admin v3

colinhowe: Had this discussion with Goran. I would *love* to use it. Doing so requires that I update a load of AV1 code due to field names not being what `AdminInput` expects. For now, we agreed to keep it as is.

src/components/organisms/TimingEventModal/TimingEventModal.tsx
colinhowe: It creates a H2 in the middle of form. That's not a convention in this form. I would rather update all the other divs than have this one being different. Given we're going to be redesigning all of this I'd rather not spend too much time on this kind of change.

src/components/organisms/TimingEventModal/TimingEventModal.scss
colinhowe: Existing styles. I'll update what's there to make it use the variables

src/components/organisms/TimingEventModal/TimingEventModal.tsx
goran-peoski-work: Not spending time === you don't change it in any way, even if there is h2 in the middle of the form. At least do not add `input-group`

src/components/organisms/TimingEventModal/TimingEventModal.tsx
goran-peoski-work: `<div className="input-group">` is still an issue for me. I'd simply leave the old section in place.

src/components/organisms/TimingEventModal/TimingEventModal.tsx
colinhowe: The h2 looks messy in the form with different font size etc.
image
The rest of the form uses `input-group`.

What's the issue here? Is this about trying to remove bootstrap?

goran-peoski-work: It's about trying to remove `input-group`. If you can localize the few style props to `TimingEventModal__input-group` even by just copying it from the global, that would be good enough I guess.

colinhowe: I can do that. Cheers :)

src/components/organisms/TimingEventModal/TimingEventModal.scss
```diff
@@ -23,12 +37,36 @@
     max-width: 700px;
   }
 
-  .input-group {
+  &__input-group {
```
goran-peoski-work, line 40:
```suggestion
  // @debt this should be shorter, not having to compensate for `global.scss` interference. remove when no longer necessary
  &__input-group {
```

goran-peoski-work (approved):
LGTM

I've added a `@debt` comment so we can clean it up in the future

PR #2612, opened 2021-11-29
Add venue ID to venue functions in the backend

colinhowe:
This creates a single unique and non-changing way of finding a venue in the database.

This closes https://github.com/sparkletown/internal-sparkle-issues/issues/1522

This also removes some older functions as updating them to support world ID made little sense.

### DIFF
--- a/functions/src/utils/venue.js
+++ b/functions/src/utils/venue.js
@@ -3,18 +3,11 @@ const { VenueIdSchema } = require("../types/venue");
 const VENUE_ID_REGEX = /[a-z0-9_]{1,250}/;
 const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
 
-/**
- * @deprecated Use generateSlug instead.
- */
-const getVenueId = (name) =>
-  name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
-
 const generateSlug = (name) =>
   name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
 
 const checkIfValidVenueId = (venueId) => VenueIdSchema.isValidSync(venueId);
 
 exports.VENUE_ID_REGEX = VENUE_ID_REGEX;
-exports.getVenueId = getVenueId;
 exports.generateSlug = generateSlug;
 exports.checkIfValidVenueId = checkIfValidVenueId;
--- a/functions/venue.js
+++ b/functions/venue.js
@@ -5,7 +5,7 @@ const { HttpsError } = require("firebase-functions/lib/providers/https");
 const { addAdmin, removeAdmin } = require("./src/api/roles");
 
 const { checkAuth } = require("./src/utils/assert");
-const { getVenueId, checkIfValidVenueId } = require("./src/utils/venue");
+const { checkIfValidVenueId } = require("./src/utils/venue");
 const { ROOM_TAXON } = require("./taxonomy.js");
 
 const PLAYA_VENUE_ID = "jamonline";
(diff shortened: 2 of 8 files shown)


functions/src/utils/venue.js
```diff
@@ -3,18 +3,11 @@ const { VenueIdSchema } = require("../types/venue");
 const VENUE_ID_REGEX = /[a-z0-9_]{1,250}/;
 const INVALID_SLUG_CHARS_REGEX = /[^a-zA-Z0-9]/g;
 
-/**
- * @deprecated Use generateSlug instead.
- */
-const getVenueId = (name) =>
-  name.replace(INVALID_SLUG_CHARS_REGEX, "").toLowerCase();
```
colinhowe, line 10: This function was misleading as slug is not always going to be the ID of a venue.

functions/venue.js
```diff
@@ -49,27 +49,6 @@ const VenueTemplate = {
...
-  VenueTemplate.jazzbar,
-  VenueTemplate.partymap,
-  VenueTemplate.animatemap,
-  VenueTemplate.performancevenue,
-  VenueTemplate.themecamp,
-  VenueTemplate.viewingwindow,
-  VenueTemplate.zoomroom,
-  VenueTemplate.screeningroom,
-];
-
```
colinhowe, line 72: These all became unused after I deleted the v1 create venue function.

functions/venue.js
```diff
@@ -180,117 +159,6 @@ const checkIfUserHasVoted = async (venueId, pollId, userId) => {
...
-      venueData.roomVisibility = data.roomVisibility;
-      break;
-
-    default:
-      break;
-  }
-
-  return venueData;
-};
-
```
colinhowe, line 293: Removed this function as adding support for a proper venue ID to be passed in didn't make any sense.

functions/venue.js
```diff
@@ -517,27 +385,6 @@ exports.removeVenueOwner = functions.https.onCall(async (data, context) => {
...
-  const venueId = getVenueId(data.name);
-  const venueRef = admin.firestore().collection("venues").doc(venueId);
-
-  batch.set(venueRef, venueData);
-  initializeVenueChatMessagesCounter(venueRef, batch);
-
-  await batch.commit();
-
-  return venueData;
-});
```
colinhowe, line 539: Deleting this function was easier than trying to shoe-horn in the venue ID.

src/api/admin.ts
```diff
@@ -289,6 +295,20 @@ const createFirestoreVenueInput_v2 = async (
...
+const createFirestoreVenueInput_v2 = async (
+  input: VenueInput_v2,
+  user: firebase.UserInfo
+) => {
+  // We temporarily cast the result to unknown so that we can cast to the
+  // same type with the ID added, then we add the missing property.
+  const result = ((await createFirestoreVenueInputWithoutId_v2(
+    input,
+    user
+  )) as unknown) as FirestoreVenueInput_v2;
```
colinhowe, line 307: I don't like this. I couldn't think of a cleaner way to do this though.
The return from `createFirestoreVenueInputWithoutId_v2` does not have an ID. We want to add that ID property on to both the object and the type of that object. Open to ideas on cleaner ways to do this.

src/api/admin.ts
```diff
@@ -300,10 +320,10 @@ export const createVenue = async (
 };
 
 export const createVenue_v2 = async (
-  input: WithWorldId<VenueInput_v2>,
+  input: WithWorldId<Omit<VenueInput_v2, "id">>,
```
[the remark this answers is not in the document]
colinhowe: I want the default for "doing things with venues" to require a venue ID. Then the special case is venue creation which doesn't need (or want) an ID.

denisdimitrov (approved):
LGTM!

PR #2613, opened 2021-11-29
Validate iframeUrl while editing portal/venue AdminV3

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1391


src/components/atoms/InputField/InputField.scss
```diff
@@ -17,11 +17,15 @@ $icon-width: 2.5rem;
...
-      box-shadow: 0 0 0 2px var(--danger--under-rgb);
+  &--invalid {
+    margin-bottom: 1.5rem;
+
+    .InputField__input {
+      border: 1px solid var(--danger--under);
+      &:focus {
+        border: 1px solid var(--accent--over-10a);
+        box-shadow: 0 0 0 2px var(--danger--under-rgb);
+      }
```
goran-peoski-work, line 28: Looks good. Fast and clean.

[NOTE]
There is also a possibility of replacing `InputField` with `AdminInput` so we can phase these old components out faster.

https://github.com/sparkletown/sparkle/blob/staging/src/components/molecules/AdminInput/AdminInput.tsx#L11-L16  has a simpler DX that deals with displaying the errors etc:
```tsx
           <AdminInput
              name={name}
              register={register}
              errors={errors}
             // ... etc ...
           />
```

goran-peoski-work (approved):
LGTM

goran-peoski-work:
Any issue by replacing `.includes()` with `.startsWith()`?

https://github.com/sparkletown/sparkle/pull/2613/files#diff-0556ce18a6f78a791400f295530c173c8109131fc7be1cfcb7484699fa45f621R9

src/components/atoms/InputField/InputField.scss
sunny-viktoryia: thanks for the hint. Replaced.
https://github.com/sparkletown/sparkle/pull/2613/commits/b174173485aa87b5c64ced96a909690e03ce6013

We'll need to add an error state to the AdminInput component and add red borders to it if there are errors (InputField component has them).

PR #2616, opened 2021-11-29
replaced venue id with spaceslug param

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1286

Fixed an issue with hardcoded redirect to `bootstrap` venue since `spaceSlug` was missing from the url params for new registration url

Fixed by replacing here:
image

`venueId` replaced by `spaceSlug` in the url params: 

Code of conduct questions:
image

`venueId` replaced by `spaceSlug` in the url params: 
image

### DIFF
--- a/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
+++ b/src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
@@ -164,7 +164,7 @@ const RegisterForm: React.FC<RegisterFormProps> = ({
       postRegisterCheck(auth, data);
 
       const accountProfileUrl = `/account/profile${
-        spaceId ? `?venueId=${spaceId}` : ""
+        spaceId ? `?spaceSlug=${spaceId}` : ""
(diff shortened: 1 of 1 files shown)


src/components/organisms/AuthenticationModal/RegisterForm/RegisterForm.tsx
```diff
@@ -164,7 +164,7 @@ const RegisterForm: React.FC<RegisterFormProps> = ({
       postRegisterCheck(auth, data);
 
       const accountProfileUrl = `/account/profile${
-        spaceId ? `?venueId=${spaceId}` : ""
+        spaceId ? `?spaceSlug=${spaceId}` : ""
```
goran-peoski-work, line 167: [NOTE]

This should be using one of those `url.ts` utility functions so we can avoid mess like this.

goran-peoski-work (approved):
LGTM

colinhowe:
Looks good to me!
image

PR #2624, opened 2021-11-30
Fix for portals not being addable

colinhowe:
The "add portal" button was doing nothing. This is because the space slug was
being used instead of the world slug and so the component was not being
passed a full set of props.

Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1525


alexey-petrov (approved):
LGTM

PR #2626, opened 2021-11-30
Fix for using slug instead of ID when upserting a portal

colinhowe:
In some cases the ID of the venue is not the same as the slug and this prevents a portal from being successfully created.


alexey-petrov (approved):
LGTM

PR #2628, opened 2021-11-30
Fix and unify Admin input components

goran-peoski-work:
Some bugfixes and addition of missing styles for
- `AdminInput`
- `AdminTextarea`
- `ButtonNG`

Also, their styles are now more unified and less dependent on `global.scss` that has also resulted in addition of few more standardizing `@function box-shadow...` in `constants.scss`

sparkle-input-elements-01


colinhowe (approved):
Looks good. Some nice refactoring in here.

PR #2632, opened 2021-12-01
Change layout and labels on world settings page

colinhowe:
Closes https://github.com/sparkletown/internal-sparkle-issues/issues/1498
This contains a rename of the `StartForm` and friends into `GeneralForm` etc to better match what they represent now.


src/components/organisms/WorldGeneralForm/WorldGeneralForm.tsx
```diff
@@ -165,22 +165,6 @@ export const WorldStartForm: React.FC<WorldStartFormProps> = ({
...
-          />
-        </AdminSection>
-        <AdminSection title="Describe your world">
-          <AdminInput
-            label="How should we describe this world?"
-            name="description"
-            register={register}
-            errors={errors}
-          />
-        </AdminSection>
```
colinhowe, line 183: These aren't wanted any more

src/pages/WorldEditor/WorldEditor.tsx
```diff
@@ -1,50 +1,75 @@
...
-    () => history.push(ADMIN_V3_WORLDS_BASE_URL),
-    [history]
-  );
+  const goBack = useCallback(() => {
+    if (history.length) {
+      history.goBack();
+    } else {
+      history.push(adminWorldSpacesUrl(world?.slug));
+    }
+  }, [history, world?.slug]);
```
colinhowe, line 45: This is used when the cancel button is clicked. When someone cancels we generally want to back to where we've come from. However, if someone has opened this page inside a new tab then we need to go somewhere sensible. In this case, the dashboard for the world.

src/store/reducers/WorldEdit.ts
```diff
@@ -1,11 +1,13 @@
...
 export const worldEditStartValuesReducer: (
-  state: Partial<WithOptionalWorldId<WorldStartFormInput>>,
+  state: Partial<WithOptionalWorldId<WorldGeneralFormInput>>,
   action: WorldEditActions
-) => Partial<WithOptionalWorldId<WorldStartFormInput>> = (state = {}, action) =>
-  action.type === WORLD_EDIT ? { ...action.payload } : state;
+) => Partial<WithOptionalWorldId<WorldGeneralFormInput>> = (
+  state = {},
+  action
+) => (action.type === WORLD_EDIT ? { ...action.payload } : state);
```
colinhowe, line 13: auto-linted.

src/components/molecules/WorldNav/WorldNavLabelMap.ts
```diff
@@ -3,7 +3,7 @@ import { WorldNavTab } from "types/world";
 export const WorldNavLabelMap: Readonly<
   Record<WorldNavTab, string>
 > = Object.freeze({
-  [WorldNavTab.start]: "Start",
+  [WorldNavTab.general]: "General settings",
```
goran-peoski-work, line 6: Must it have `settings` here and not after `Advanced` or even after `Entrance`? Is this a comment for the issue, not just this PR?

src/pages/WorldEditor/WorldEditor.tsx
goran-peoski-work: > When someone cancels we generally want to back to where we've come from. 

Not really or not anymore. The preferred per IA is to remove the footer (and add a back button above the tab bar).

The cancel should take the user to a sensible place. Does it though? 

It might not mean "Go Back" since the Start form now renamed General isn't just for creating a new one, but also for editing an old one, thus a use case might be to advance to the Entrance settings, not bail out to worlds list.

Since I'm seeing a back button added in the title, instead of Approve, I will add a change request to just remove the `Footer` component from this editor and move the save/cancel buttons inside their respective forms.

goran-peoski-work (changes requested):
Since there is a back button at top, Footer should probably be removed, not use Cancel as a navigation tool, but most likely as a form reset/refresh to the original from DB

src/components/molecules/WorldNav/WorldNavLabelMap.ts
colinhowe: It's a valid question. One for product team to resolve at a later date.

colinhowe:
> Since there is a back button at top, Footer should probably be removed, not use Cancel as a navigation tool, but most likely as a form reset/refresh to the original from DB

Dealth with in 9fcebb0923a085c132e7f5894a1cc836c4f00609

The scrolling is a bit janky on the page still. I want to pick that up in a separate PR as I couldn't easily get it working and I know you've been working on scrolling :)

src/components/molecules/WorldNav/WorldNavLabelMap.ts
goran-peoski-work: Ping @whoaitsaimz

src/pages/WorldEditor/WorldEditor.tsx
```diff
@@ -1,50 +1,65 @@
...
-  const navigateToHome = useCallback(
-    () => history.push(ADMIN_V3_WORLDS_BASE_URL),
-    [history]
-  );
+  if (!isLoaded) {
+    return <LoadingPage />;
+  }
+
+  const adminTitle =
+    world !== undefined ? `${world.name} settings` : "Create a new world";
```
goran-peoski-work, line 42:
```suggestion
    world ? `${world.name} settings` : "Create a new world";
```

src/pages/WorldEditor/WorldEditor.tsx
```diff
@@ -1,50 +1,65 @@
...
 
   const WorldEditorPanel = PANEL_MAP[selectedTab] ?? <></>;
 
   return (
     <div className="WorldEditor">
       <WithNavigationBar>
         <AdminRestricted>
+          <AdminTitleBar>
+            {world !== undefined && (
```
goran-peoski-work, line 51:
```suggestion
            {world && (
```

src/store/reducers/WorldEdit.ts
goran-peoski-work: Leave it to Prettier to always make code uglier.

goran-peoski-work (approved):
Approved with few comments about code style that generally shouldn't be a problem, but look like a code smell.

PR #2615, opened 2021-11-29
New Create Space Page

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1466

PR #2633, opened 2021-12-01
fix the admin sidebar footer and global footer

denisdimitrov:
- Increases the global footer z-index and the position of the admin sidebar footer.

Staging:

current

New:

new

Fixes: N/A (Saw the problem, fixed the problem)

### DIFF
--- a/src/components/organisms/AdminVenueView/components/AdminSidebarFooter/AdminSidebarFooter.scss
+++ b/src/components/organisms/AdminVenueView/components/AdminSidebarFooter/AdminSidebarFooter.scss
@@ -1,7 +1,7 @@
 @import "scss/constants.scss";
 
 .AdminSidebarFooter {
-  position: absolute;
+  position: fixed;
   bottom: $footer-height;
 
   display: flex;
--- a/src/scss/constants.scss
+++ b/src/scss/constants.scss
@@ -293,7 +293,7 @@ $z-layers: (
   // External Link
   external-message: $z-layer-external-message,
   // Unsorted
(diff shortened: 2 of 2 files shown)


goran-peoski-work:
I have a bit more comprehensive update here 
- https://github.com/sparkletown/sparkle/pull/2615/files that 
doesn't warrant making any new `position: fixed` elements.

goran-peoski-work (changes requested):
Please take a look at `constants.scss` in https://github.com/sparkletown/sparkle/pull/2615/files and if that approach of updating z-layer indices relative to each other is viable alternative to adding more fixed positioned elements.

denisdimitrov:
@goran-peoski-work  Why are you against `position: fixed`? It looks like this is exactly what we are trying to do but instead we are using `position: absolute` and trying to adapt it to act and look like `position: fixed`. `position: absolute` is more likely to break, like in the current case. Now that we have scrollbar on the page this broke, but if used `fixed` in the first place it would have still worked as intended.

goran-peoski-work:
`position: absolute` is also bad.  I was against adding it to other places because of a component that was about to be deleted https://github.com/sparkletown/sparkle/pull/2645

goran-peoski-work:
Another relevant PR:
- https://github.com/sparkletown/sparkle/pull/2635

Yes, the scrollbars still have things to desire for, but at least now there is an easier path to removing `position`, preset `height` etc.

denisdimitrov:
UI has changed and this PR is no longer valid.

PR #2635, opened 2021-12-01
Fix Admin scrollbar issues

goran-peoski-work:
Some views in Admin have scrollbar issues.

1 Cutting off to certain elements when scrolled all the way down (now no more of that):

sparkle-admin-scrolls-01

2. Double scrollbars (now only a single one displayed):

sparkle-admin-scrolls-02

### DIFF
--- a/src/components/organisms/AdminVenueView/components/AdminShowcase/AdminShowcase.tsx
+++ b/src/components/organisms/AdminVenueView/components/AdminShowcase/AdminShowcase.tsx
@@ -3,14 +3,20 @@ import classNames from "classnames";
 
 import "./AdminShowcase.scss";
 
-export interface AdminPanelProps {
+export interface AdminShowcaseProps {
   className?: string;
+  variant?: "internal-scroll" | "no-scroll";
 }
 
-export const AdminShowcase: React.FC<AdminPanelProps> = ({
+export const AdminShowcase: React.FC<AdminShowcaseProps> = ({
   className,
+  variant = "internal-scroll",
   children,
 }) => {
-  const containerClasses = classNames("AdminShowcase", className);
+  const containerClasses = classNames(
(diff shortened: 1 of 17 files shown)


src/components/organisms/AdminVenueView/components/AdminShowcase/AdminShowcase.tsx
```diff
@@ -3,14 +3,20 @@ import classNames from "classnames";
 
 import "./AdminShowcase.scss";
 
-export interface AdminPanelProps {
+export interface AdminShowcaseProps {
   className?: string;
+  variant?: "internal-scroll" | "no-scroll";
```
colinhowe, line 8: `.AdminShowcase--no-scroll` hasn't been created. I'm assuming that's intentional and it's just to make it clear to anyone looking at this what is going on.

colinhowe (approved):
I've left a question. I think it's fine, I just wanted to flag it in case there was an oversight.
Checked this out and had a play around the views. Looks fine.

src/components/organisms/AdminVenueView/components/AdminShowcase/AdminShowcase.tsx
goran-peoski-work: We've usually been adding [NOTE] and [OOS] as a shorthand for things that probably shouldn't be acted upon on the particular PR

goran-peoski-work: Right now it's intentional since  the `variant` system is still new. Some of those values should probably converge in the future, once we have more use cases and we determine similarities/differences. Also, the `?` is there to facilitate legacy uses and generally should be avoided (rectified in future PR)

PR #2641, opened 2021-12-02
remove space edit ng form and replace it with the original one

denisdimitrov:
- Removes `SpaceEditFormNG` and replaces it with the original one, `SpaceEditForm`
- Adds the missing auditorium fields to the original component

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1531


colinhowe (approved):

denisdimitrov:
Do not merge. The validation needs to be refactored after the conflict resolve.

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -60,7 +60,7 @@ export const SpacesDropdown: React.FC<SpacesDropdownProps> = ({
 
         return (
           <ReactBootstrapDropdown.Item
-            key={id}
+            key={id + name}
```
colinhowe, line 63: `id` is unique. It shouldn't need name adding to it.

colinhowe, line 63: `id` is unique. It shouldn't need `name` adding to it.
```suggestion
            key={id}
```

colinhowe (changes requested):
One minor change and then this can be merged

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
denisdimitrov: Looks like it isn't. I don't know why but React warning was showing up for duplicated keys.

colinhowe:
Superseded by https://github.com/sparkletown/sparkle/pull/2658

PR #2636, opened 2021-12-02
Add world slugs to attendee urls

colinhowe:
Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/1529

URLs before looked like: `/in/bootstrap`

This meant that if two spaces had the same slug and in different worlds we would have breakages. This PR fixes it and moves to our desired URL scheme.

Now they look like `/in/w/bootstrapaaa/s/bootstrap`.

This PR also introduces typing on `WorldSlug` and `SpaceSlug` as a few bugs have been found along the way where IDs were being used as slugs or the wrong slug was being used in the wrong place.

PR #2643, opened 2021-12-03
fixed parent space icon + update properties

alexey-petrov:
Follow up rework of parent list options: https://github.com/sparkletown/internal-sparkle-issues/issues/1461
Specifically fixes this issue: https://github.com/sparkletown/internal-sparkle-issues/issues/1461#issuecomment-984932616

Replaced `name` with `slug` to be used in the option list, as it proved to be more sustainable source of identification

### DIFF
--- a/src/components/molecules/SpaceEditForm/SpaceEditForm.tsx
+++ b/src/components/molecules/SpaceEditForm/SpaceEditForm.tsx
@@ -285,7 +285,7 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
             ({ id, worldId }) =>
               !(roomVenue?.worldId !== worldId || id === spaceIdFromPortal)
           )
-          .map((venue) => [venue.id, venue])
+          .map((venue) => [venue.slug, venue])
       ),
     [ownedVenues, roomVenue?.worldId, spaceIdFromPortal]
   );
@@ -293,8 +293,8 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
   const parentSpace = useMemo(
     () =>
       roomVenue?.parentId
-        ? ownedVenues.find(({ id }) => id === roomVenue?.parentId)
-        : { name: "" },
+        ? ownedVenues.find(({ slug }) => slug === roomVenue?.parentId)
+        : { slug: "" },
     [ownedVenues, roomVenue?.parentId]
   );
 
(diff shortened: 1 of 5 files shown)


src/components/molecules/SpaceEditForm/SpaceEditForm.tsx
```diff
@@ -285,16 +285,16 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
...
       ),
     [ownedVenues, roomVenue?.worldId, spaceIdFromPortal]
   );
 
   const parentSpace = useMemo(
     () =>
       roomVenue?.parentId
-        ? ownedVenues.find(({ id }) => id === roomVenue?.parentId)
-        : { name: "" },
+        ? ownedVenues.find(({ slug }) => slug === roomVenue?.parentId)
```
goran-peoski-work, line 296: This doesn't look OK. You can use slugs for user facing concerns, but when it comes down to background logic, it's the IDs that are the steady immutable references.

src/components/molecules/SpaceEditFormNG/SpaceEditFormNG.tsx
```diff
@@ -238,16 +238,16 @@ export const SpaceEditFormNG: React.FC<SpaceEditFormNGProps> = ({
...
       ),
     [ownedVenues, portal?.worldId, spaceIdFromPortal]
   );
 
   const parentSpace = useMemo(
     () =>
       portal?.parentId
-        ? ownedVenues.find(({ id }) => id === portal?.parentId)
-        : { name: "" },
+        ? ownedVenues.find(({ slug }) => slug === portal?.parentId)
```
goran-peoski-work, line 249: Same comment about using slug instead of ID as previous

src/pages/Admin/Details/Form/DetailsForm.tsx
```diff
@@ -295,16 +295,16 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ venue, worldId }) => {
       Object.fromEntries(
         filteredWorlds
           .filter(({ id }) => !(venueId === id))
-          .map((world) => [world.id, world])
+          .map((world) => [world.slug, world])
```
goran-peoski-work, line 298: Filtering should still be done by id than by slug, but at this point, `DetailsForm` is no longer used for creating spaces 🎉 so it's kind of a moot point

goran-peoski-work (changes requested):
Using slugs instead of names... kind of OK. Maybe slugs and names can both be shown to the user. But in the background logic, IDs are the relevant references.

alexey-petrov:
Closing in favor of https://github.com/sparkletown/sparkle/pull/2636 and any follow up for that one

PR #2644, opened 2021-12-03
styles fix for new world page

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1538

Styles fix


goran-peoski-work (approved):
Approving, but

please check how `AdminTitleBar` is used in `SpacesDashboard` and if possible duplicate that, otherwise just add  `@debt` comment that it should be used

alexey-petrov:
> 7997226

added a debt comment in 7997226

PR #2645, opened 2021-12-03
Remove AdminSidebarFooter component and add proper replacement

goran-peoski-work:
This is a bug fix:

sparkle-no-more-footer-02

But also,
🎉 No more fixed footer 🎉 .

 Buttons are now inside `AdminSidebarButtons` component that provides a bit of styles but is inside their respective forms.

sparkle-no-more-footer-01


alexey-petrov (approved):
lgtm

PR #2649, opened 2021-12-03
v1 no more

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1502

Removing files, functions, components etc. that are no longer in use.


colinhowe (approved):
I think there's an accidental change in `.gitignore`. `.txt`->`.txts`
Once that's resolved you might as well merge this and we catch things in QA.

It's pretty hard to effectively review a change like this with so many deletions without spending /days/ reviewing it.

As a general practice for changes like this I try and split things up into commits that are *purely* deletes and commits that are changing/renaming code. It makes it easier to review the meaningful changes whilst skipping past the mass deletions.

goran-peoski-work:
> commits that are purely deletes and commits that are changing/renaming code.

Most of the 11 commits in this PR are like that. However, some deletions don't work without changes where they used to be imported, so...

PR #2650, opened 2021-12-05
Spaces have their own specific settings

colinhowe:
Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/1528
This temporarily removes all the portal creation/editing from maps. There is some work coming in from @sunny-viktoryia that brings that back in as a modal.
Space specific settings are now shown for each space in the right place (e.g. click settings on a space)

image


colinhowe:
Superseded

PR #2655, opened 2021-12-06
Create/edit portals in the modal

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1497

~~PLEASE, HEEEELP! A question here: can anyone help me understand why "Override appearance" toggle triggers every time I choose any of the "appearance" options? Although, it stands still if I check other toggles;
when I click the title it triggers as well. 😐~~  
~~racking my brain trying to figure it out. I probably, missing something small, but couldn't find it so far.~~

YAY!! DONE!!! THANK YOUUU! @alexey-petrov  and @colinhowe!!! ❤️
recording (6)

***Note!*** Newly created portals will not navigate anywhere so far. As here we connect portals to the target spaces using "spaceId", it means that we'll have to finish UX, generating the needed URLs based on the space Id. (should I do that in this PR? or we'll better create a follow-up?)
Also, it might mean we'll need to write a migration script to add `spaceId` to all the portals we have.


src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -15,10 +15,15 @@ import "./SpacesDropdown.scss";
 
 const noneOptionName = "None";
 const spaceNoneOption = Object.freeze({
-  none: Object.freeze({ id: undefined, name: "", template: undefined }),
+  none: Object.freeze({ id: "", name: "", template: undefined }),
```
sunny-viktoryia, line 18: I've set an empty string here to get rid of the warning. React requires all the list items to have a key prop defined.

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -86,7 +91,7 @@ export const SpacesDropdown: React.FC<SpacesDropdownProps> = ({
       return "Select a space";
     }
 
-    const space = spaces?.[selected.name] ?? parentSpace;
+    const space = spaces?.[selected.id ?? ""] ?? parentSpace;
```
sunny-viktoryia, line 94: fixed the bug with the image display in the dropdown.

src/components/organisms/PortalAddEditForm/PortalAddEditForm.tsx
```diff
@@ -0,0 +1,268 @@
...
+          .filter(
+            ({ worldId: venueWorldId, id }) =>
+              !(venueWorldId !== world?.id || id === currentSpaceId)
+          )
+          .map((venue) => [venue.id, venue])
+      ),
+    [currentSpaceId, ownedVenues, world?.id]
+  );
+
+  // const [isOverrideAppearanceEnabled, setOverridenAppearance] = useState(false);
```
sunny-viktoryia, line 166: commented the lines with the Override appearance toggle until we resolve the issue which I mentioned in the description to the PR (or if we decide to remove that toggle at all 😬 )

src/settings/spacePortalsSettings.ts
```diff
@@ -110,7 +110,7 @@ export const SPACE_INFO_LIST: SpaceInfoListItem[] = [
     poster: PosterEmbeddable,
     description: "",
     template: VenueTemplate.embeddable,
-    hidden: false,
+    hidden: true,
```
sunny-viktoryia, line 113: Here we hide "embeddable" portals which do not require a space until we decide what to do with them. Were we going to get rid of them?

colinhowe:
PortalVisibility has a bug in it. The click event isn't being marked as handled and so the event is bubbling up and landing on the checkbox which is causing a toggle.

```
diff --git a/src/components/atoms/PortalVisibility/PortalVisibility.tsx b/src/components/atoms/PortalVisibility/PortalVisibility.tsx
index 22e73d37..a4e067ab 100644
--- a/src/components/atoms/PortalVisibility/PortalVisibility.tsx
+++ b/src/components/atoms/PortalVisibility/PortalVisibility.tsx
@@ -45,7 +45,9 @@ export const PortalVisibility: React.FC<PortalVisibilityProps> = ({
           return (
             <div
               key={label}
-              onClick={() => {
+              onClick={(ev) => {
+                ev.stopPropagation();
+                ev.preventDefault();
                 setSelected(value);
                 setValue(name, value, true);
               }}

**Review by colinhowe — APPROVED** [review]

File: src/components/organisms/PortalAddEditForm/PortalAddEditForm.tsx [hunk 5]
```diff
@@ -188,14 +188,22 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
...
   );
 
+  const parentSpace = useMemo(
+    () =>
+      portal?.spaceId
+        ? ownedVenues.find(({ id }) => id === portal?.spaceId)
+        : { name: "" },
+    [ownedVenues, portal?.spaceId]
+  );
+
```
sunny-viktoryia, line 211: @colinhowe, why was that added? I believe we don't need that, do we?

PR #2656, opened 2021-12-06
custom redirects for clients added

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1547

Adds a custom redirect for `googlecloudwest` and `iterable` spaces.

To be removed in the future as per https://github.com/sparkletown/internal-sparkle-issues/issues/1550

### DIFF
--- a/src/components/organisms/AppRouter/AppRouter.tsx
+++ b/src/components/organisms/AppRouter/AppRouter.tsx
@@ -1,5 +1,10 @@
 import React, { lazy, Suspense } from "react";
-import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
+import {
+  BrowserRouter as Router,
+  Redirect,
+  Route,
+  Switch,
+} from "react-router-dom";
 
 import {
   ACCOUNT_ROOT_URL,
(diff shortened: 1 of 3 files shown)


src/components/organisms/AppRouter/AppRouter.tsx
```diff
@@ -93,10 +103,27 @@ const EmergencyViewPage = lazy(() =>
 export const AppRouter: React.FC = () => {
   const { user } = useUser();
 
+  // @debt custom redirects that are to be removed in the future
+  // done as per: https://github.com/sparkletown/internal-sparkle-issues/issues/1547
+  const googleCloudWestWorldUrl = generateAttendeeInsideUrl({
+    worldSlug: googleCloudWestName,
+    spaceSlug: googleCloudWestName,
```
colinhowe, line 110: if you do `as WorldSlug` and `as SpaceSlug` on these and the ones below then you don't need to modify `utils/url.js`

colinhowe (changes requested):

src/components/organisms/AppRouter/AppRouter.tsx
alexey-petrov: done 639cb5f

colinhowe (approved):

PR #2661, opened 2021-12-06
Fix back to dashboard button in admin

sunny-viktoryia:
fixes back to dashboard button


sunny-viktoryia:
the changes are not correct; here is better PR https://github.com/sparkletown/sparkle/pull/2662

PR #2666, opened 2021-12-07
fix space landing page cover image

denisdimitrov:
The landing page was showing the space banner instead of the world cover image.

World is now passed from the parent via props because it's already fetched there and it doesn't have to be fetched again in the child and prolong the loading.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1532


src/pages/VenueLandingPage/VenueLandingPageContent.tsx
```diff
@@ -36,20 +37,20 @@ dayjs.extend(advancedFormat);
...
 }) => {
   const venueEvents = useSelector(venueEventsSelector);
 
   const spaceSlug = venue.slug;
 
-  const { world } = useWorldById(venue?.worldId);
-
   const [validBannerImageUrl] = useValidImage(
-    venue?.config?.landingPageConfig.bannerImageUrl,
+    world?.config?.landingPageConfig.coverImageUrl,
```
colinhowe, line 53: I think this needs to be a fallback.
So, we try the space landing cover if there is one. Otherwise, we try the world cover image.

There is another issue that @alexey-petrov is fixing for those settings not being configurable at the moment!
https://github.com/sparkletown/internal-sparkle-issues/issues/1562

colinhowe (changes requested):

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
denisdimitrov: Got it!

colinhowe (approved):

PR #2668, opened 2021-12-07
Fix splitting of events by space

colinhowe:
This also fixes the "create experience" button in this view

Resolves
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1581


goran-peoski-work (approved):
Could be refactored with cleaner code style, time permitting.

Logic looks like it's doing the correct thing.

Approving.

PR #2679, opened 2021-12-07
Disable threaded replying to DMs

colinhowe:
Resolves
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1582

Bit dirty, but, removes the threaded reply option in a minimal way.


mike-lvov (approved):
A hacky, but a quick way to solve the issue

PR #2685, opened 2021-12-08
allow portals to be moved and resized in spaces tab

denisdimitrov:
Swaps the run and spaces tab components.
Also adds some additional logic to show disabled portals in the MapPreview component

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1559


colinhowe (approved):
Looks awesome. Only small issue I found is that if you edit a portal name using the portal list below the map then the name change doesn't appear until you click a portal to move it.

Might be nice to fix if it's trivial.

Likewise, if the corners can be made so that they resize in the right direction that would be nice.

Only fix these if they'll take less than ten minutes.

colinhowe (approved):

PR #2689, opened 2021-12-08
Changes modal text. Makes modal non-dissmissable

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1589
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1476


colinhowe (approved):
Nice one

PR #2695, opened 2021-12-09
fix input autofill colors

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1579

Reproduced using Win10/Mac Chrome

Styles differ a bit in multiple components we have. Tried to match the background color based on other inputs in corresponding forms ( admin settings, timing events, user search etc )

Also, added a guard to event timing list ( used to break the page if event missed some details )

Before:
https://www.loom.com/share/6d213a11642247c29bc4e3e098e04be2

After: 
image


src/utils/event.ts
```diff
@@ -60,11 +60,21 @@ export const eventHappeningNow = (
 export const hasEventFinished = (event: VenueEvent) =>
   isAfter(Date.now(), eventEndTime(event));
 
-export const eventStartTime = (event: VenueEvent) =>
-  fromUnixTime(event.start_utc_seconds);
+export const eventStartTime = (event: VenueEvent) => {
+  if (!event?.start_utc_seconds) {
```
alexey-petrov, line 64: had some events with missing details, so had to add this guard to prevent timing page from breaking

src/utils/event.ts
```diff
@@ -60,11 +60,21 @@ export const eventHappeningNow = (
...
+    return new Date();
+  }
+
+  return fromUnixTime(event.start_utc_seconds);
+};
 
-export const eventEndTime = (event: VenueEvent) =>
-  addMinutes(eventStartTime(event), event.duration_minutes);
+export const eventEndTime = (event: VenueEvent) => {
```
alexey-petrov, line 71: had some events with missing details, so had to add this guard to prevent timing page from breaking

src/components/atoms/InputField/InputField.scss
```diff
@@ -33,6 +33,17 @@ $icon-width: 2.5rem;
...
+
+    &:-webkit-autofill,
+    &:-webkit-autofill:hover,
+    &:-webkit-autofill:focus,
+    &:-webkit-autofill:active {
+      -webkit-text-fill-color: var(--content--over);
+      -webkit-box-shadow: 0 0 0 $input-auto-fill-box-shadow-height
+        $input-auto-fill-box-shadow inset;
+      box-shadow: 0 0 0 $input-auto-fill-box-shadow-height
+        $input-auto-fill-box-shadow inset;
```
goran-peoski-work, line 45: Please make a single `box-shadow--autofilled` similar to https://github.com/sparkletown/sparkle/blob/staging/src/scss/constants.scss#L313 for reuse here and in the other components

src/components/organisms/TimingDeleteModal/TimingDeleteModal.scss
```diff
@@ -0,0 +1,14 @@
+@import "scss/constants.scss";
+
+.form {
```
goran-peoski-work, line 3: This should probably be encapsulated inside a component class, like
```suggestion
TimingDeleteModal {
  .form {
```

src/components/organisms/TimingDeleteModal/TimingDeleteModal.tsx
```diff
@@ -75,25 +77,29 @@ export const TimingDeleteModal: React.FC<TimingDeleteModalProps> = ({
 
   return (
     <Modal show={show} onHide={onHide}>
-      <div className="form-container">
+      <div className="form">
```
goran-peoski-work, line 80: For consistency with other components.

One way would be:

```suggestion
      <div className="TimingDeleteModal TimingDeleteModal__form">
```

although that leaves the top level component not the same class as this component, so

another way would be:

```suggestion
      <div className="TimingDeleteModal">
        <Modal show={show} onHide={onHide}>
          <div className="TimingDeleteModal__form">
```

src/scss/constants.scss
```diff
@@ -33,6 +33,12 @@ $lighter-intermediate-grey: #3d3a3f;
 $light-grey: #292929;
 $light-grey--lightest: #acadb0;
 
+// @debt probably needs to be replaced with var(...), but it's used in a place
+// where var wasn't working
+$input-auto-fill-box-shadow: $secondary--dark;
+$input-auto-fill-box-shadow-timeline: #434343;
+$input-auto-fill-box-shadow-height: 30px;
+
```
goran-peoski-work, line 41: Delete. Not only replaced with var, but also with a `box-shadow--autofilled` function. Please don't use `$secondary-dark` there should be a comment after that color for suitable replacement. Similar to the gray `#434343` one, there are already existing ones at `:root`

src/utils/event.ts
goran-peoski-work: Will this impact other calls of the same utility function?

goran-peoski-work: Consider this: pure function got turned into impure one.
Now none of  the places it is called from can depend on it to consistently return the same result for the same input.

A way to make it pure again is to not generate input inside, but add second options parameter with field `defaultValue` for the places that want this kind of behavior

goran-peoski-work (changes requested):
Few suggestions and comments. Let's do a little screen share about the colors when you can.

src/utils/event.ts
colinhowe: Or, just fix your events ;)

src/scss/constants.scss
alexey-petrov: removed and replaced with var() 9bff7d8

src/components/organisms/TimingDeleteModal/TimingDeleteModal.tsx
alexey-petrov: done 917110d

src/components/organisms/TimingDeleteModal/TimingDeleteModal.scss
alexey-petrov: done 917110d

src/components/atoms/InputField/InputField.scss
alexey-petrov: done 917110d

alexey-petrov:
@goran-peoski-work as for the issue described in https://github.com/sparkletown/sparkle/pull/2695#discussion_r765700244

I've decided against those changes and reverted the recently added guards. Since newer events that we create are validated to have all the necessary `date` attributes that page won't be crashing, but I've added @debt comments there, so that we can add some sort of `default` value or guard this piece of code from crashing in some other way

src/components/molecules/AdminInput/AdminInput.scss
```diff
@@ -24,10 +24,9 @@
     &:-webkit-autofill:focus,
     &:-webkit-autofill:active {
       -webkit-text-fill-color: var(--content--over);
-      -webkit-box-shadow: 0 0 0 $input-auto-fill-box-shadow-height
-        $input-auto-fill-box-shadow inset;
-      box-shadow: 0 0 0 $input-auto-fill-box-shadow-height
-        $input-auto-fill-box-shadow inset;
+      -webkit-box-shadow: box-shadow--input-shadow(var(--content--under-80a))
+        inset;
+      box-shadow: box-shadow--input-shadow(var(--content--under-80a)) inset;
```
goran-peoski-work, line 29: Kind of forgot these would still be duplicates. Don't know why override it with the prefix though, may want to reconsider  at least adding a `@debt` comment to recheck `caniuse.com` in the future and deal with it. Browsers should eventually just use the standard one.

src/scss/constants.scss
```diff
@@ -320,6 +318,10 @@ $z-layers: (
   @return 0 0 0 2px $color;
 }
 
+@function box-shadow--input-shadow($color) {
```
goran-peoski-work, line 319: Not the luckiest pick of name I think, sure it is connected to input, but shadow appearing twice in the name...

src/scss/constants.scss
```diff
@@ -35,9 +35,7 @@ $light-grey--lightest: #acadb0;
 
 // @debt probably needs to be replaced with var(...), but it's used in a place
 // where var wasn't working
-$input-auto-fill-box-shadow: $secondary--dark;
 $input-auto-fill-box-shadow-timeline: #434343;
```
goran-peoski-work: This one should be removed.

```suggestion
```

src/scss/constants.scss
```diff
@@ -35,9 +35,7 @@ $light-grey--lightest: #acadb0;
 
 // @debt probably needs to be replaced with var(...), but it's used in a place
 // where var wasn't working
```
goran-peoski-work:
```suggestion
```

goran-peoski-work (approved):
Approving, with: 

- you can add `@debt` comments if this PR is needed fast and we can rework some of the logic later
- make pretty if something can be polished in 5-10 minutes, I guess, then merge

src/utils/event.ts
```diff
@@ -60,21 +57,15 @@ export const eventHappeningNow = (
...
-  if (!event?.start_utc_seconds) {
-    return new Date();
-  }
-
-  return addMinutes(eventStartTime(event), event.duration_minutes);
-};
+// @debt: if event or event.start_utc_seconds are falsy this func crashes the page
+// probably need to add an additional `defaultValue` to return or some sort of a guard
+export const eventEndTime = (event: VenueEvent) =>
+  addMinutes(eventStartTime(event), event.duration_minutes);
```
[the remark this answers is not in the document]
alexey-petrov: changed props + types c974b57
and updated these functions usages everywhere in the codebase

src/scss/constants.scss
alexey-petrov: renamed c974b57

src/scss/constants.scss
alexey-petrov: removed

src/components/molecules/AdminInput/AdminInput.scss
alexey-petrov: removed `--webkit` ones. confirmed that everything is working fine c974b57

PR #2703, opened 2021-12-09
Fix deleting profile links

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1608

when we remove the last profile link using `remove` function of `useFieldArray`, instead of setting `[]` to the `data`, it disappears, so this hacky solution is just to set an empty list of `profileLinks` if there are no profile links.

related issue: https://github.com/react-hook-form/react-hook-form/issues/871


goran-peoski-work (approved):
LGTM.

If it works, it works. Make pretty later.

PR #2709, opened 2021-12-09
Record the worlds that a user has visited and use to determine showing world entrance

colinhowe:
Resolves
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1620

### DIFF
--- a/src/pages/VenuePage/VenuePage.tsx
+++ b/src/pages/VenuePage/VenuePage.tsx
@@ -12,7 +12,11 @@ import { VenueTemplate } from "types/venues";
 
 import { hasEventFinished, isEventStartingSoon } from "utils/event";
 import { tracePromise } from "utils/performance";
-import { isCompleteProfile, updateProfileEnteredVenueIds } from "utils/profile";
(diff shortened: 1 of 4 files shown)


goran-peoski-work (approved):
LGTM

src/pages/VenuePage/VenuePage.tsx
```diff
@@ -228,7 +250,7 @@ export const VenuePage: React.FC = () => {
   const { template, hasPaidEvents } = space;
 
   const hasEntrance = !!world?.entrance?.length;
-  const hasEntered = enteredVenueIds?.includes(spaceId);
+  const hasEntered = world?.id && enteredWorldIds?.includes(world.id);
```
goran-peoski-work, line 253: Expect to have an array with `null` or `undefined` values in it?

colinhowe: `enteredWorldIds` might not be present on the owning document. If it exists it should only have actual world IDs though :)

PR #2712, opened 2021-12-10
fix entrance step delete

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1619

* Fixed world update function to properly update `entrance` and `userStatuses` props if any of those are being removed

Post-reload state:
image


goran-peoski-work (approved):
LGTM

PR #2717, opened 2021-12-10
add workaround for missing permissions after log in

denisdimitrov:
When `history.push` is used after log in, for some strange reason the `firebase rules` doesn't consider the user logged in and `Missing or insufficient permissions` error is shown.

This is not really much of a fix, more like a "patch" and workaround. It uses `window.location.href` just like the `Join event` button does. Ideally we should investigate the problem and use the router instead.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1618


src/components/molecules/NavBar/components/NavBarLogin/NavBarLogin.tsx
```diff
@@ -1,28 +1,38 @@
...
-      ),
-    [history, worldSlug, spaceSlug]
-  );
+  const navigateToDefault = useCallback(() => {
+    const redirectUrl = generateUrl({
+      route: !hasEntrance
+        ? ATTENDEE_SPACE_INSIDE_URL
+        : ATTENDEE_SPACE_ENTRANCE_URL,
+      required: ["worldSlug", "spaceSlug"],
+      params: { worldSlug, spaceSlug, step: "1" },
```
alexey-petrov, line 27: Goran added a const `export const DEFAULT_ENTER_STEP = "1";` in `src/settings/urlSettings.ts`
https://github.com/sparkletown/sparkle/pull/2718/files#diff-4412f8f75290a96968c8717d422435457d190e4f557c135122f1ce728de4a430
we might as well do the same here and resolve conflicts later

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
```diff
@@ -64,11 +66,18 @@ const VenueLandingPageContent: React.FC<VenueLandingPageContentProps> = ({
...
-        : venueEntranceUrl(world?.slug, spaceSlug);
+    const worldSlug = world.slug;
+
+    const redirectUrl = generateUrl({
+      route:
+        user && !hasEntrance
+          ? ATTENDEE_SPACE_INSIDE_URL
+          : ATTENDEE_SPACE_ENTRANCE_URL,
+      required: ["worldSlug", "spaceSlug"],
+      params: { worldSlug, spaceSlug, step: "1" },
```
alexey-petrov, line 77: same here

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
```diff
@@ -64,11 +66,18 @@ const VenueLandingPageContent: React.FC<VenueLandingPageContentProps> = ({
...
+    });
+
```
alexey-petrov, line 79: not sure, but we might need that same comment here as well
`/**
     * Using history.push() doesn't work correctly for some strange reason and firebase returns 403 (Missing or insufficient premissions).
     * It's almost like it doesn't detect when the user logs in and the rules consider the `uid` to be `null`
     */`

alexxpetrov (approved):
LGTM
A few things to consider changing tho

src/components/molecules/NavBar/components/NavBarLogin/NavBarLogin.tsx
```diff
@@ -1,28 +1,38 @@
...
+    });
+
+    /**
+     * Using history.push() doesn't work correctly for some strange reason and firebase returns 403 (Missing or insufficient premissions).
+     * It's almost like it doesn't detect when the user logs in and the rules consider the `uid` to be `null`
```
goran-peoski-work, line 32: I think I know what the issues is. The above comment summed up is:

React doesn't refresh the state (from DB, in the Redux cache or elsewhere) before it tries to re-route, but `window.location.href =` does make the page refresh, thus the login state is updated.

In order to continue using `history.push()` we might need to see if the above assumption is correct and if so, how to force the state refresh before the route is changed. Possibly by using `async` or `.then()`.

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
denisdimitrov: debt comment has been moved to `VenuePage.tsx`

src/components/molecules/NavBar/components/NavBarLogin/NavBarLogin.tsx
denisdimitrov: The problem is in the `VenuePage.tsx` it renders login form when user is missing and content when user exists. However the firebase rules doesn't seem to detect the change only on re-render and that's why the page is now reloaded after login.

src/components/molecules/NavBar/components/NavBarLogin/NavBarLogin.tsx
denisdimitrov: Used in both places.

alexey-petrov:
What's our resolution regarding this fix? Is it still necessary? @denisdimitrov

denisdimitrov:
Stale and already fixed by the new login screens.

PR #2718, opened 2021-12-13
Tighten up entrance steps routing

goran-peoski-work:
Couldn't reproduce what is described in
- https://github.com/sparkletown/internal-sparkle-issues/issues/1613

Instead tightened up the routing connected to entrance steps:
- URLs more in line with general strategy `/e/w/mysharkyworld/s/myconversationspace/1`
- use of `generateUrl` for more explicit and robust route generation
- removed ` spaceSlug &&` check since it is legacy and not needed
- added not-a-number check for the entrance step
- a bit more consolidation work in the URL constants

### DIFF
--- a/src/pages/VenueEntrancePage/VenueEntrancePage.tsx
+++ b/src/pages/VenueEntrancePage/VenueEntrancePage.tsx
@@ -1,7 +1,10 @@
 import React, { useCallback } from "react";
 import { Redirect, useHistory, useParams } from "react-router-dom";
 
-import { ACCOUNT_PROFILE_VENUE_PARAM_URL } from "settings";
+import {
+  ACCOUNT_PROFILE_VENUE_PARAM_URL,
+  ATTENDEE_STEPPING_PARAM_URL,
+} from "settings";
 
 import {
   EntranceStepTemplate,
@@ -9,11 +12,7 @@ import {
 } from "types/EntranceStep";
 
 import { isCompleteProfile } from "utils/profile";
-import {
-  generateAttendeeInsideUrl,
-  generateUrl,
-  venueEntranceUrl,
-} from "utils/url";
+import { generateAttendeeInsideUrl, generateUrl } from "utils/url";
 
 import { useSpaceParams } from "hooks/spaces/useSpaceParams";
 import { useWorldAndSpaceBySlug } from "hooks/spaces/useWorldAndSpaceBySlug";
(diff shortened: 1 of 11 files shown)


src/pages/VenueEntrancePage/VenueEntrancePage.tsx
```diff
@@ -48,8 +47,13 @@ export const VenueEntrancePage: React.FC = () => {
 
   const proceed = useCallback(
     () =>
-      spaceSlug &&
```
goran-peoski-work, line 51: This might be one suspect for the cause of the heisenbug.

src/pages/VenueEntrancePage/VenueEntrancePage.tsx
```diff
@@ -62,14 +66,9 @@ export const VenueEntrancePage: React.FC = () => {
   }
 
   const stepConfig = world.entrance?.[step - 1];
-  if (!stepConfig) {
+  if (Number.isNaN(step) || !stepConfig) {
```
goran-peoski-work, line 69: This might be another, albeit less likely, suspect for the cause of the heisenbug.

goran-peoski-work (commented):
Couple of comments on possible places the bug might have occured

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
```diff
@@ -87,7 +95,7 @@ const VenueLandingPageContent: React.FC<VenueLandingPageContentProps> = ({
   const containerClasses = classNames("header", containerVars);
 
   return (
-    <div className="container venue-entrance-experience-container">
+    <div className="VenueLandingPageContent container venue-entrance-experience-container">
```
alexey-petrov, line 98: this `container` doesn't seem to be used anywhere. might as well remove it

src/utils/url.ts
```diff
@@ -44,7 +46,10 @@ export const generateUrl: (options: {
     return fallback;
   }
 
-  const relativePath = generatePath(route, params);
+  // NOTE: ?? {} stops TS from crying and the check makes the shorter generatePath is used
+  const relativePath = params
+    ? generatePath(route, params ?? {})
```
alexey-petrov, line 51: is there any way we can use `ALWAYS_EMPTY_OBJECT` here? I think I had some issues using that placeholder constant in such situation, but not sure

src/pages/VenueLandingPage/VenueLandingPageContent.tsx
goran-peoski-work: Deliberate moniker. It isn't the only measure to be used in a SCSS file. This one makes the top element of the component visible in the browser inspector and hopefully some day make it easier to have custom styles by users.

So, my rule of thumb is:
- Always put the component name as a class to the top level element in it

goran-peoski-work: The other ones.... didn't want to get into a whole cleaning process, wanted to just have it most as it is.

src/utils/url.ts
goran-peoski-work: The comment should've explained it, but isn't clear enough. If `params` is `null` or `undefined`, the other version `generatePath(route)` is used, so the one you've commented on will never go pass the `??` operator, but TS still complains about type 🤷

PR #2720, opened 2021-12-14
Fix days difference counting in the schedule

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1625

### DIFF
--- a/src/hooks/useVenueScheduleEvents.ts
+++ b/src/hooks/useVenueScheduleEvents.ts
@@ -13,7 +13,7 @@ import {
 
 import { VenueEvent } from "types/venues";
 
-import { getEventDayRange, isEventLiveOrFuture } from "utils/event";
+import { isEventLiveOrFuture } from "utils/event";
 import { WithVenueId } from "utils/id";
 import { isDateRangeStartWithinToday } from "utils/time";
 
@@ -25,7 +25,7 @@ import { useRelatedVenues } from "./useRelatedVenues";
 
 const emptyRelatedEvents: WithVenueId<VenueEvent>[] = [];
 const minRangeValue = 0;
-const todaysDate = new Date();
+const todaysDate = startOfToday();
 
 const useVenueScheduleEvents = ({
   userEventIds,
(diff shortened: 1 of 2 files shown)


src/hooks/useVenueScheduleEvents.ts
```diff
@@ -25,7 +25,7 @@ import { useRelatedVenues } from "./useRelatedVenues";
 
 const emptyRelatedEvents: WithVenueId<VenueEvent>[] = [];
 const minRangeValue = 0;
-const todaysDate = new Date();
+const todaysDate = startOfToday();
```
sunny-viktoryia, line 28: this is the fix

goran-peoski-work: I assume this one uses local browser TZ

src/hooks/useVenueScheduleEvents.ts
```diff
@@ -114,7 +110,8 @@ const useVenueScheduleEvents = ({
     fromUnixTime(firstRangeDateInSeconds)
   );
 
-  const dayDifference = getEventDayRange(daysInBetween, isOneEventAndLive);
+  // add 1 day to daysInBetween to form a full timeline
```
goran-peoski-work, line 113: This is just English for the line under, adds no more meaning. Would be nice to instead have the reason of why there is a need to add a day.

goran-peoski-work (approved):
LGTM

src/hooks/useVenueScheduleEvents.ts
sunny-viktoryia: yes, it uses a local browser timezone.

src/hooks/useVenueScheduleEvents.ts
sunny-viktoryia: changed the comment: https://github.com/sparkletown/sparkle/pull/2720/commits/658dc03ac29ab71d1ea74795e3bd5e23be09a05e

PR #2723, opened 2021-12-15
sync world entrance updates

alexxpetrov:
Additional changes to:
- https://github.com/sparkletown/internal-sparkle-issues/issues/974
- https://github.com/sparkletown/internal-sparkle-issues/issues/1619

Makes entrance form changes sync with the form of everyone that has this settings tab open. So that if user A updates it -> user B has the up-to-date form state

* Fixed form `isDirty` state ( now also depends on entrance step button list, which changes were previously ignored )


src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
```diff
@@ -22,6 +29,7 @@ import { ButtonNG } from "components/atoms/ButtonNG";
 
 import "./EntranceStepsInputFieldSet.scss";
 
+let buttonsCopy: EntranceStepButtonConfig[] | undefined = [];
```
goran-peoski-work, line 32: Mutable shared state, `let` used instead of `const` on top of it. What's the point of this?

src/hooks/useArray.ts
```diff
@@ -82,14 +90,27 @@ export const useArray = <T>(
     },
     [items]
   );
-
+  console.log(isDirty);
```
goran-peoski-work, line 93: This shouldn't be making regular PR appearances

```suggestion
```

src/hooks/useArray.ts
```diff
@@ -30,18 +31,25 @@ export type UseArrayResult<T> = {
...
+  const [prepared, setPrepared] = useState(
+    options?.prepare ? array?.map(options.prepare) : array
+  );
 
   const [items, setItems] = useState(prepared ?? []);
-  const [isDirty, setIsDirty] = useState(false);
 
+  useEffect(() => {
+    setItems(prepared ?? []);
+  }, [prepared]);
```
goran-peoski-work, line 50: My take on this:

No side effects inside `useArray`, this will make it do unexpected things. Have the appropriate array operation be called inside `useEffect()` inside a component, that makes it clear that it's part of the component lifecycle, not hidden mutation of the array data that may or may not trigger component re-render.

Also, the above is a comment on first view, still don't see the need for all of this.

src/components/organisms/WorldEntranceForm/WorldEntranceForm.tsx
```diff
@@ -37,13 +38,18 @@ export interface WorldEntranceFormProps {
   world: WithId<World>;
 }
 
+let codeQuestionsCopy: Question[] | undefined = [];
+let profileQuestionsCopy: Question[] | undefined = [];
+let entranceCopy: EntranceStepConfig[] | undefined = [];
```
goran-peoski-work, line 43: `const` instead of `let`.
no shared mutable state (which is why there is `ALWAYS_EMPTY_ARRAY`)

src/hooks/useArray.ts
```diff
@@ -82,14 +90,27 @@ export const useArray = <T>(
...
+      const result = options?.prepare
+        ? newArray?.map(options.prepare)
+        : newArray;
+      clearDirty();
+      setPrepared(result);
+
+      return result;
+    },
+    [clearDirty, options?.prepare]
+  );
```
goran-peoski-work, line 112: The idea of `useArray()` is to always wrap the same object, the same array, hence the `clear` function.

If there is a need of new wrapped array in a component somewhere, it can be dealt with there (a simple `[...arrayFromUseArray]` creates new reference for outside purpose.

goran-peoski-work (changes requested):
On a first view of the code, it appears to me `useArray()` has been given concerns that should remain in their respective components.

The use of shared mutable state is also a code smell, don't yet see why it is necessary

src/components/organisms/WorldEntranceForm/WorldEntranceForm.tsx
goran-peoski-work: I guess, the question would be, should this be `const a = ALWAYS_EMPTY_ARRAY` or should be `const [a,setA]=useState()`

alexey-petrov: got rid of those `let`s 941192c

src/hooks/useArray.ts
alexey-petrov: 941192c

src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
alexey-petrov: removed 941192c

src/components/organisms/WorldEntranceForm/WorldEntranceForm.tsx
```diff
@@ -90,25 +95,25 @@ export const WorldEntranceForm: React.FC<WorldEntranceFormProps> = ({
   });
 
   useEffect(() => {
-    if (!isEqual(world.questions?.profile, profileQuestionsCopy)) {
-      profileQuestionsCopy = world.questions?.profile;
+    if (!isEqual(world.questions?.profile, profileQuestionsList)) {
```
goran-peoski-work, line 98: Maybe fail fast for more linear code:

```ts
if( isEqual() ) return;
```

src/components/organisms/WorldEntranceForm/WorldEntranceForm.tsx
```diff
@@ -38,18 +38,23 @@ export interface WorldEntranceFormProps {
...
-
 export const WorldEntranceForm: React.FC<WorldEntranceFormProps> = ({
   world,
 }) => {
   const worldId = world.id;
   const { user } = useUser();
 
-  // @debt sync useArray with the form changes or try useFieldArray
   const [isDirtyButtons, setDirtyButtons] = useState(false);
+  const [codeQuestionsList, setCodeQuestionsList] = useState<
```
goran-peoski-work, line 48: I think "this is where the rabbit lies".

The meaning of "we need to add replace"?

Well if you put `useArray(codeQuestionsList)` instead of ` useArray<Question>(world.questions?.code)`, does't that remove the need of that entire `replace` function thing being part of the hook?

goran-peoski-work (changes requested):
I might now know how the code smell can probably be refactored so the concern of the array replacement stays within the component and is not pushed down to the `useArray()` hook.

Still need to research if/how that should work.

src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
```diff
@@ -30,6 +36,7 @@ export interface EntranceStepsInputFieldSetProps {
   onUpdate: UseArrayUpdate<EntranceStepConfig>;
   onRemove: UseArrayRemove<EntranceStepConfig>;
   register: (Ref: unknown, RegisterOptions?: unknown) => void;
+  setDirtyButtons: Dispatch<SetStateAction<boolean>>;
```
goran-peoski-work, line 39: Extra dependency that isn't needed. `onUpdate` and `onRemove` already exist and can be used to handle the dirty state in the parent. Also, they are made to return a context object which aside from `index` can be extended to return any other field of interest.

```suggestion
```

src/components/molecules/EntranceStepsInputFieldSet/EntranceStepsInputFieldSet.tsx
```diff
@@ -86,8 +94,13 @@ export const EntranceStepsInputFieldSet: React.FC<EntranceStepsInputFieldSetProp
     update: updateButton,
     clear: clearButtons,
     remove: removeButton,
+    isDirty: isDirtyButtons,
   } = useArray<EntranceStepButtonConfig>(item?.buttons);
 
+  useEffect(() => {
+    setDirtyButtons(isDirtyButtons);
+  }, [isDirtyButtons, setDirtyButtons]);
+
```
goran-peoski-work, line 103: See https://github.com/sparkletown/sparkle/pull/2723/files#r771133948

```suggestion
```

src/components/organisms/WorldEntranceForm/WorldEntranceForm.tsx
```diff
@@ -149,7 +183,8 @@ export const WorldEntranceForm: React.FC<WorldEntranceFormProps> = ({
     isSubmitting ||
     isDirtyCode ||
     isDirtyProfile ||
-    isDirtyEntrance
+    isDirtyEntrance ||
+    isDirtyButtons
```
goran-peoski-work, line 187: This is out of place. Since buttons are part of the entrance, adding or removing or changing a button should make `idDirtyEntrance` be true

alexxpetrov:
Not relevant anymore. Fixed by redesign/react-hook-form upgrade to 6.x.x

PR #2724, opened 2021-12-15
Move social login from the space setting to the world one

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1636

Makes Social Login in the World Advanced Settings work.
image

slack thread https://doesitsparkle.slack.com/archives/C01H9D1P1PE/p1639423276001400


src/components/organisms/AuthenticationModal/LoginForm/LoginForm.tsx
```diff
@@ -53,7 +53,7 @@ const LoginForm: React.FunctionComponent<LoginFormProps> = ({
     reValidateMode: "onChange",
   });
 
-  if (!space) return null;
+  if (!space || !world) return null;
```
goran-peoski-work, line 56: Maybe a loading screen, message (we got the components for it), but if hard to determine which is appropriate even `null` is OK, just maybe with `@debt` so we figure it out later

goran-peoski-work (approved):
LGTM

PR #2727, opened 2021-12-17
Create determineAvatar utility and use it for avatars

goran-peoski-work:
Holistic approach to resolving 
- https://github.com/sparkletown/internal-sparkle-issues/issues/1615

Now with a bit more logic and consolidation of defaults.


src/utils/image.ts
```diff
@@ -67,3 +74,50 @@ export const getFirebaseStorageResizedImage = (
...
+
+  // checking user.anonMode ruled against by https://github.com/sparkletown/internal-sparkle-issues/issues/1615#issuecomment-991054012
+
+  if (url.startsWith("/static") || url.startsWith("http")) {
+    return url;
+  }
+
+  if (isDefined(index) && Number.isSafeInteger(index) && index >= 0) {
+    return list[index % list.length];
+  }
```
mike-lvov, line 105: Given that this is only used in one place and this place randoms the index, can we remove this just to use the seed & hash implementation for a cleaner view?

mike-lvov (commented):
Overall looks good. One more thing that is not covered is if the link is broken. We did use some checks in the code before. eg. load the image, if it's not valid, use the default one. I wonder if we should implement this here too

src/utils/image.ts
goran-peoski-work: I went with the simplest solution first so that we get more consistent output and there are number of fallbacks that go from the most to the least consistent. so far, the index is my least favorite ones, but had to add because of AnimateMap.

I say this PR is good enough for solving the current problem and the hashing can be improved once all the places that require an avatar have similar data to provide.

goran-peoski-work: The above rationalization is more or less the same with the image checks. If and when we need it, can be added with a subsequent PR. Right now it might just be an overkill and OOS

goran-peoski-work: Ideally, what I envisioned for a future PR was the addition of another field in DB that sets default avatar once, on user creation, and picks that one consistently if/when the chosen avatar fails.

mike-lvov (approved):
LGTM

PR #2729, opened 2021-12-17
Allow worlds to be hidden

colinhowe:


src/components/organisms/AppRouter/Provided.tsx
```diff
@@ -13,12 +14,18 @@ export const Provided: React.FC<ProvidedProps> = ({
...
 }) => {
   const { worldSlug, spaceSlug } = useSpaceParams();
-  const { space, spaceId } = useWorldAndSpaceBySlug(worldSlug, spaceSlug);
+  // @debt This provider is used in cases where we don't have a space slug
+  // Refactoring that is more risky than makes sense right now (big events
+  // coming up). Therefore, we do best effort here. This provider shouldn't
+  // need a space at all. Related venues can be derived entirely from a world
+  // now.
+  const { world } = useWorldBySlug(worldSlug);
+  const { spaceId } = useWorldAndSpaceBySlug(worldSlug, spaceSlug);
```
mike-lvov, line 23: This hook returns world. I don't think you need to use `useWorldBySlug` hook here

scripts/migration/v0.0.9__add-is-hidden-field.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { MigrateOptions } from "fireway";
+
+export const migrate = async ({ firestore }: MigrateOptions) => {
+  console.log(`Updating worlds...`);
+
+  const { docs: worldDocs } = await firestore.collection("worlds").get();
+
+  for (const worldDoc of worldDocs) {
+    if (Object.keys(worldDoc.data()).indexOf("isHidden") !== -1) {
```
mike-lvov, line 11:
```suggestion
    if ("isHidden" in worldDoc.data()) return;
```

mike-lvov (approved):
A few moments and good to go

src/components/organisms/AppRouter/Provided.tsx
colinhowe: See the big comment :(

scripts/migration/v0.0.9__add-is-hidden-field.ts
colinhowe: Of course. I was knee deep in firestore rules which doesn't allow this :(

colinhowe: Fixed in ea34b8ac

src/components/organisms/AppRouter/Provided.tsx
colinhowe: Improved comment in ea34b8ac

scripts/migration/v0.0.9__add-is-hidden-field.ts
goran-peoski-work: Major difference between `in` which goes up the prototype chain and `Object.keys()` which only return own keys. The only suggestion I'd make here was to use `.includes()` instead of `indexOf`

src/api/world.ts
```diff
@@ -246,6 +246,7 @@ export const findWorldBySlug = async ({
   const worldsRef = await firebase
     .firestore()
     .collection(COLLECTION_WORLDS)
+    .where("isHidden", "==", false)
```
goran-peoski-work, line 249: Considering `isHidden` is a new field, more prudent action would be to count `undefined`  same as `false`. This also eliminates the need to run a migration script to "unhide"...

On second thought, why not just `isVisible` and set to `true` only for those that should be seen?

colinhowe: Firestore rules don't play nice with undefined fields. At all. I spent hours trying to get it to work :(

src/hooks/spaces/useWorldAndSpaceBySlug.ts
```diff
@@ -41,7 +41,11 @@ export const useWorldAndSpaceBySlug = (
 
   const worldsRef = firestore
     .collection("worlds")
-    .where("slug", "==", worldSlug ?? "")
+    .where("isHidden", "==", false)
```
goran-peoski-work, line 44: I think less effort solutions might be `isVisible` or check `!= true` and leave the current state be same as when `undefined` is the value of the field (since it's being added as a new one)

src/hooks/worlds/useWorldBySlug.ts
```diff
@@ -27,6 +27,7 @@ export const useWorldBySlug: (worldSlug?: string) => UseWorldBySlugResult = (
   const worldsRef = firestore
```
goran-peoski-work, line 27: Looks like duplicate. Probably should just remove it and rely on the function from `./api/world.ts`

src/api/world.ts
goran-peoski-work: This is why I've also suggested `!=true` as an option elsewhere or turning the tables from `isHidden` to `isVisible`, but at the end of the day, I'm just providing few different opinions for good measure and will leave it up to your best judgement and possibly @mike-lvov's second review after he'd already approven

PR #2733, opened 2021-12-20
Deletes unused backed functions

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1635


goran-peoski-work (approved):
LGTM

Next stop (PR) we rename what is left to be more in line with current practice

PR #2734, opened 2021-12-20
Update the create world with slug display and few more details

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1427
- https://github.com/sparkletown/internal-sparkle-issues/issues/1426

Changes:
- added an URL that shows where the world could be accessed in the admin
- added the missing _Back to Dashboard_ button for creating a world (like the one for Space)
- the button for _Save_ will now alternate between _Create_ when world has no ID and _Update_ when it has
- the tab bar will be hidden when creating a new world (i.e. no ID found) and shown upon editing a world

**Before**:

sparkle-create-world-slug-02

sparkle-create-world-slug-03

**After**:

sparkle-create-world-slug-06

sparkle-create-world-slug-05


src/components/molecules/WorldCard/WorldCard.tsx
```diff
@@ -48,7 +48,11 @@ export const WorldCard: React.FC<WorldCardProps> = ({ world }) => {
         variant="dark"
         isLink
         disabled={!world.slug}
-        linkTo={adminWorldSpacesUrl(world.slug)}
+        linkTo={generateUrl({
+          route: ADMIN_IA_WORLD_PARAM_URL,
+          required: ["worldSlug"],
+          params: { worldSlug: world.slug },
+        })}
```
colinhowe, line 55: I'm not a fan of using `generateUrl` everywhere. It's more verbose. It requires someone using it to write out which parts of the URL are required. It loses type validation. The thin wrapper function `adminWorldSpaceUrl` gives all of that.

colinhowe (approved):
The PR itself looks fine. We should probably talk about generateUrl somewhere else.

src/components/molecules/WorldCard/WorldCard.tsx
goran-peoski-work: @colinhowe  It has already been done elsewhere https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1638790538446100

goran-peoski-work: Here is a quote of it:
>  generateUrl should be preferred to any other URL generator for your links, since we can upgrade it in the future, but for now provides a bit more explicit DX. Future upgrades might be adding a generic <T> for the params it expects and making that optional required array not necessary (if we parse the route itself for : and ? - simple regex does it)

PR #2735, opened 2021-12-20
removed stats dependency from firebase functions

alexey-petrov:
Fix for failed builds due to missing `stats` dependency in firebase functions

image
image


sunny-viktoryia (approved):
Looks good. 🙃 
Thanks for the fix!!

mike-lvov (approved):
Ah, missed this

goran-peoski-work (approved):
LGTM

PR #2736, opened 2021-12-20
remove stats exports + replace ?? for firebase function

alexey-petrov:
Follow up for https://github.com/sparkletown/sparkle/pull/2735
* Removed `??` from the venue.js and replaced with `||`


sunny-viktoryia (approved):
Approving. 🙃

PR #2743, opened 2021-12-21
replaced dot more icon with string

alexey-petrov:
Closes:
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1629

Before: 
image

After:
image


src/components/molecules/UserList/UserList.tsx
```diff
@@ -38,7 +36,8 @@ export const UserList: React.FC<UserListProps> = ({
   showTitle = true,
   attendeesTitle,
 }) => {
-  const hasExcessiveUserCount = userCount > usersSample.length;
+  const excessiveUserCount =
+    userCount - usersSample.length > 0 ? userCount - usersSample.length : 0;
```
goran-peoski-work, line 40: In case of `NaN` (is that possible?), fallback to empty string ?

I think the code from bellow `and {excessiveUserCount} more` would actually read `and more` in case of NaN

goran-peoski-work (approved):
LGTM

src/components/molecules/UserList/UserList.tsx
alexey-petrov: since we have this true/false check in the render that shouldn't be the case 
`{excessiveUserCount && (
          <div className="UserList__excessive-number">
            and {excessiveUserCount} more
          </div>
        )}
`

PR #2744, opened 2021-12-22
Nicer UI/UX for AnimateMap initialization errors

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1523

by providing an error message at the place the bug is supposed to show up.

**Before**:
Screenshot_20211228-173728

**After**:
Screenshot_20211229-113453

An attempt to use animated map on mobile device that doesn't support WebGL 2:
sparkle-webgl-support-07

An attempt to use AnimateMap that has initialization error:
sparkle-error-boundary-07


goran-peoski-work:
Marked as do-not-merge since this PR is useful for the preview link, not necessarily the solution will end up with until properly tested on phone

PR #2750, opened 2022-01-05
Add warning for mobile devices

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1222

it was: 
image

it is now:
image

also added the ability to hide an announcement message, because the close button was hidden under the chat buttons.


src/components/molecules/NavBar/NavBar.scss
```diff
@@ -305,4 +305,16 @@ $nav-bar-title-width: 120px;
...
+    .NavSearchBar {
+      display: none;
+    }
+  }
+
+  @media (max-width: $media-breakpoint--md) {
+    .VenuePartygoers {
+      display: none;
+    }
+  }
```
mike-lvov, line 319: Based on what I see, you make it appear on small browser sizes which is not !== mobile. Is it an intended effect?

sunny-viktoryia: not sure if I got the comment. Can you give me an example, how would you do that? and what do you think is a better way to do that?

mike-lvov: So, what you do here is you say that everything that has a smaller screen than `$media-breakpoint--md` is a mobile. It's not a wrong solution. Just saying out loud that this includes not only moblies. More like, all the people with the small screens

goran-peoski-work:
>  We are best experienced using your Chrome browser.

This makes very little sense on iOS since Chrome and Firefox and basically any non-Apple browser is still required to use the OS engine.

src/components/molecules/NavBar/NavBar.scss
goran-peoski-work: Use case: split screen, like having two or 4 windows side by side using a windowing manager (and even  the new Windows 11 has something similar). On a laptop, this can easily make your browser window shrink to the break-point set for mobile device

sunny-viktoryia: ah, exactly these changes are related to the responsive view of the NavBar, because when the screen gets smaller there's not enough space for all the components of the NavBar, so I decided to hide them.

So I believe the answer is - yes, it's an intended effect. I just hide elements when there is not enough space for them.

goran-peoski-work: I can suggest some capability sniffing like in https://stackoverflow.com/questions/7838680/detecting-that-the-browser-has-no-mouse-and-is-touch-only/52854585#answer-52854585

mike-lvov: @sunny-viktoryia in this case, the wording in the warning is misleading. Might be useful to change it to reflect the absence of the space rather than mobile/desktop

sunny-viktoryia: @goran-peoski-work @mike-lvov, looks like we should better go to the issue and discuss wording and conditions of the banner visibility there.

Left a comment: 
https://github.com/sparkletown/internal-sparkle-issues/issues/1222#issuecomment-1008944184

whoaitsaimz: Thanks for the discussion guys, I [added a comment](https://github.com/sparkletown/internal-sparkle-issues/issues/1222#issuecomment-1009336879) to the ticket

sunny-viktoryia: the message is updated: https://github.com/sparkletown/sparkle/pull/2750/commits/538553ee0c0ba387080f0af8b6d74c06234c649e

ready for re-review. 🙃

goran-peoski-work (approved):
Message looks OK and is up to revised issue. Don't know about the `MobileWarning` name but it seems to serve the purpose.

Approving

PR #2756, opened 2022-01-10
add world schedule and show space events

denisdimitrov:
- Adds world schedule that shows all of the world events
- Renames Timing to Schedule and shows only the space specific events instead of all events

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1624


src/components/organisms/AdminVenueView/components/EventsView/EventsView.tsx
```diff
@@ -112,20 +123,24 @@ export const EventsView: React.FC<EventsViewProps> = ({ venueId, venue }) => {
...
-            checked={showSplittedEvents}
-            onChange={toggleSplittedEvents}
-            label="Split by space"
-            labelClassName="EventsView__checkbox"
-          />
+          {isWorldEvent && (
+            <Checkbox
+              checked={showSplittedEvents}
+              onChange={toggleSplittedEvents}
+              label="Split by space"
```
alexey-petrov, line 137: let's replace with `Split by ${SPACE_TAXON.lower}`

src/pages/WorldSchedule/WorldSchedule.tsx
```diff
@@ -0,0 +1,253 @@
...
+  startTime: string;
+  endDate: string;
+  endTime: string;
+}
+
+export const WorldSchedule = () => {
+  const { worldSlug } = useWorldParams();
+  const { isLoaded, world } = useWorldBySlug(worldSlug);
+
+  console.log(world);
```
alexey-petrov, line 115: let's remove this

alexxpetrov (approved):
A couple of minor things to fix

src/api/admin.ts
```diff
@@ -42,6 +42,20 @@ export interface EventInput {
...
+  description: string;
+  startDate: string;
+  startTime: string;
+  durationHours: number;
+  durationMinutes?: number;
+  host: string;
+  space?: {
+    id?: string;
+    slug?: string;
+  };
```
sunny-viktoryia, line 56: do we need `slug` here? can't we just save `spaceId`?

src/components/organisms/AdminVenueView/components/EventsView/EventsView.tsx
```diff
@@ -21,23 +24,33 @@ import { Checkbox } from "components/atoms/Checkbox";
...
     currentVenueId: venueId,
   });
+
   const { events, isLoaded: isEventsLoaded } = useSpaceEvents({
-    worldId: venue.worldId,
+    worldId: venue?.worldId ?? worldId,
     spaceIds: relatedVenueIds,
   });
 
+  const isWorldEvent = variant === EventsVariant.world;
```
sunny-viktoryia, line 52: Looks like now I can see all the world experiences in all the spaces. And current space and world schedules differ only in the Checkbox "Split by Space", although we can see all the experiences in the space schedule as well.

shouldn't we filter and show the Experiences that are related to the specific space if it's the Schedule tab in the space settings? 
and if it's a world schedule, then we show all the experiences.

sunny-viktoryia (commented):
some questions and comments from me. 🙃

src/components/organisms/AdminVenueView/components/EventsView/EventsView.tsx
whoaitsaimz: > shouldn't we filter and show the Experiences that are related to the specific space if it's the Schedule tab in the space settings?
> and if it's a world schedule, then we show all the experiences.

That is exactly what we are trying to achieve, yes

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -114,6 +135,24 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
+      spacesMap.map((space) => {
+        return (
+          <ReactBootstrapDropdown.Item
+            key={space.id}
+            ref={register}
+            name={"space"}
+            onClick={() => setValue("space", space)}
+            className="SpacesDropdown__item"
+          >
+            {space.slug}
```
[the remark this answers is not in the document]
whoaitsaimz: Definitely use **names** on anything admin or attendee with very few exceptions.

(The only exceptions I can think of are: the slug input field on the admin, and URL routing. Can't think of a reason we would need it showing to the user anywhere else.)

src/pages/SpacesDashboard/SpacesDashboard.tsx
```diff
@@ -112,6 +113,17 @@ export const SpacesDashboard: React.FC = () => {
                 Settings
               </ButtonNG>
             </div>
+            <ButtonNG
```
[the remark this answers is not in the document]
denisdimitrov: I added only the button, we can easily add the Tools section later on.

src/api/admin.ts
denisdimitrov: `slug` is not replaced with `name` we need the name to show in the dropdown but we need the `id` to be passed to the function and update the venue.

src/types/events.ts
```diff
@@ -0,0 +1,4 @@
+export enum EventsVariant {
```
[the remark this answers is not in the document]
denisdimitrov: Good point, I say we keep it as it is for now just to keep the consistency, however it would be good to rename all of these to avoid confusion in the future.

src/pages/WorldSchedule/WorldSchedule.tsx
```diff
@@ -0,0 +1,251 @@
...
+  startDate: Yup.string()
+    .required()
+    .test(
+      "startDate",
+      "Start date must not be after the end date",
+      testEndDate
+    ),
+  startTime: Yup.string()
+    .required()
+    .test("startTime", "Start time must be before the end time", testEndTime),
```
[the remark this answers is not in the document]
whoaitsaimz: If it's reasonable to make this text change...

Global ending time
~~"End date must not be before the starting date"~~ negative/inverse language... use instead: 
"End date must be after the start date"

denisdimitrov:
@sunny-viktoryia I just fixed all of your comments, had to merge with staging twice and I had quite a lot of conflicts and another quick review of the code, wouldn't hurt.

P.S. Please ignore the big package-lock.json change, I'm looking into it atm, it was probably my fault in one of the merges. :)

src/types/events.ts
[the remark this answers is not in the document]
goran-peoski-work: Rename now what you can. Put into constant what you can't.
Example: name your types and variables experience, but use `COLLECTION_EXPERIENCE="events"` to pass to DB queries.

Also, please do avoid enums, they're hard to deal with since it's TS trying to add C# thing by compiling into a complex JS. The same can be accomplished by:

`export type ExperiencePlace = "world" | "space"; `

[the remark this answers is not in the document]
goran-peoski-work: Oh, you can also make it a branded string. More info at  https://github.com/sparkletown/sparkle/blob/staging/src/types/id.ts

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -18,84 +17,102 @@ import { useRelatedVenues } from "hooks/useRelatedVenues";
...
 
 export type TimingEventModalProps = {
   show: boolean;
   onHide: () => void;
-  venueId: string | undefined;
   event?: WorldEvent;
-  template?: VenueTemplate;
-  venue: WithId<AnyVenue>;
+  venueId?: string;
   setEditedEvent: Function | undefined;
```
goran-peoski-work, line 29: [code smell] usually handlers are like `onEventEdit` and usually the `Function` is discouraged in favor of `(args, here)=>result` kind of type


```suggestion
  setEditedEvent: Function | undefined;
```

goran-peoski-work (changes requested):
Few suggestions

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -44,7 +44,7 @@ import "./AdminVenueView.scss";
 
 export enum AdminVenueTab {
   spaces = "spaces",
-  timing = "timing",
+  schedule = "schedule",
   run = "run",
 }
```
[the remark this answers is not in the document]
denisdimitrov: I think `enum` is in favor here because we are also using it for values as well. Ex: `AdminVenueTab.run` returns `"run"`

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -18,84 +17,102 @@ import { useRelatedVenues } from "hooks/useRelatedVenues";
...
   setShowDeleteEventModal: () => void;
+  worldId?: string;
```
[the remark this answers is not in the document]
denisdimitrov: The `Space` interface has `worldId: string` and `space.worldId` is passed to this prop. Changing it here to `WorldId` forces me to change it in the `Space` interface as well, which then also forces me to change the prop types in another 20 usages, which is too much imo for this PR.

src/components/organisms/TimingEventModal/TimingEventModal.tsx
denisdimitrov: I completely agree this is old code and should be fixed.

src/components/organisms/AdminVenueView/components/EventsView/EventsView.tsx
```diff
@@ -19,24 +23,36 @@ import { Checkbox } from "components/atoms/Checkbox";
 
 import "./EventsView.scss";
 
-type EventsViewProps = {
-  spaceId: SpaceId | string;
-  space: SpaceWithId;
+export type EventsViewProps = {
+  spaceId?: SpaceId | string;
+  space?: SpaceWithId;
```
goran-peoski-work, line 28: Hm... will these have different space ID values?

src/components/organisms/TimingEventModal/TimingEventModal.tsx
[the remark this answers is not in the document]
goran-peoski-work: This is the reason why I did some changes as `SpaceId | string` to be noticeable we'll want to move to the branded string in the future, but it requires too many changes.

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -18,84 +17,102 @@ import { useRelatedVenues } from "hooks/useRelatedVenues";
...
-  } = useForm<EventInput>({
+    watch,
+    setValue,
+  } = useForm<WorldScheduleEvent>({
     mode: "onSubmit",
     reValidateMode: "onChange",
     validationSchema: eventEditSchema,
+    validationContext: {
+      eventSpaceId,
+    },
```
goran-peoski-work, line 60: Interesting little schema :)

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -18,84 +17,102 @@ import { useRelatedVenues } from "hooks/useRelatedVenues";
...
     if (event?.id) {
       reset({
         name: event.name,
         description: event.description,
-        start_date: dayjs
+        startDate: dayjs
           .unix(event.startUtcSeconds)
           .format(DAYJS_INPUT_DATE_FORMAT),
-        start_time: dayjs
+        startTime: dayjs
```
goran-peoski-work, line 83: [OOS] this reminds me, we still haven't replaced  `dayjs`

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -105,23 +122,51 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
         <form className="form" onSubmit={handleSubmit(onUpdateEvent)}>
           <p>Your experience is in {eventSpace?.name}</p>
 
+          {!eventSpace?.name && (
+            <div className="TimingEventModal__input-group">
+              <Dropdown title={values.spaceName ?? "None"}>
+                {renderedSpaceIds}
+              </Dropdown>
+              {errors.spaceId && (
+                <span className="input-error">{errors.spaceId.message}</span>
```
goran-peoski-work, line 165: [NOTE] to self: eradicate `input-error` classes

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -194,75 +239,111 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
-                <span className="input-error">{errors.start_time.message}</span>
-              )}
+            <div className="TimingEventModal__input-group">
+              <label htmlFor="date">
+                Start date and time (use your own time zone; it will be
+                automatically localized)
+              </label>
+              <div className="TimingEventModal__container">
+                {/*
+                  These wrapper divs are here to unmuddle some precedance issues
```
goran-peoski-work, line 249: `@debt` tag?

src/pages/WorldSchedule/WorldSchedule.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+export const WorldSchedule = () => {
+  const { worldSlug } = useWorldParams();
+  const { isLoaded, world } = useWorldBySlug(worldSlug);
+
+  const defaultValues = useMemo<WorldScheduleFormInput>(() => {
+    const {
+      inputFormattedDateSegment: startDate,
+      inputFormattedTimeSegment: startTime,
+    } = convertDateFromUtcSeconds(world?.startTimeUnix ?? NaN);
```
goran-peoski-work, line 63: Interesting approach `?? NaN`.

Could be useful to have a comment explaining how `null` or `undefined` just don't work but `NaN` does.

goran-peoski-work (approved):
LGTM

Could use a code comment or two

src/pages/WorldSchedule/WorldSchedule.tsx
denisdimitrov: It was copy pasted from the regular schedule, good thing that you pointed it out. The reason it because `NaN` is actually a number. But maybe we should just return `0` instead.

denisdimitrov:
This is stale and very old design. It's unnecessary.

PR #2758, opened 2022-01-11
Refactor useRoom -> usePortal; and some room -> portal refactoring

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1533

mostly refactoring of the variable names: room -> portal
fixed some tech debts on the way which are no longer relevant with the new architecture


src/components/molecules/NavSearchBar/NavSearchBar.tsx
```diff
@@ -194,20 +177,10 @@ export const NavSearchBar: React.FC<NavSearchBarProps> = ({
...
       <PortalModal
-        show={isDefined(selectedRoom)}
-        portal={selectedRoom}
-        venue={selectedRoomVenue}
+        show={isDefined(selectedPortal)}
+        portal={selectedPortal}
         onHide={hidePortalModal}
       />
-
-      {/* @debt use only one EventModal instance with state controlled with redux */}
```
alexey-petrov, line 203: Do we no longer need this modal? If so is there any confirmation for it?

alexey-petrov: If it's not needed, then let's remove the modal files as well

alexey-petrov (approved):
LGTM, except for the modal thing above

src/components/molecules/NavSearchBar/NavSearchBar.tsx
sunny-viktoryia: not sure, Events were removed from the search list because we changed the way we fetch them, so I guess it was faster to remove that functionality rather than fix it everywhere. But I do not know if we're going to bring them back. 

@colinhowe, will we bring the Experiences (aka WoldEvents) back to the search results? or we won't need an EventModal anymore?

sunny-viktoryia: removed it: https://github.com/sparkletown/sparkle/pull/2758/commits/9e3f251fe8c55fbfdcc0c7dda3ce905572d77e2f

PR #2760, opened 2022-01-12
Show the default avatar on load error

sunny-viktoryia:
BEFORE
image

AFTER
image


src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -40,6 +40,13 @@ const AVATAR_SIZE_MAP: { [key in UserAvatarSize]: number | null } = {
   full: null,
 };
 
+const onProfileImageLoadError: React.ReactEventHandler<HTMLImageElement> = ({
+  currentTarget,
+}) => {
+  currentTarget.onerror = null; // prevents looping
+  currentTarget.src = DEFAULT_PROFILE_IMAGE;
```
goran-peoski-work, line 47: @sunny-viktoryia please check this PR https://github.com/sparkletown/sparkle/pull/2727 before using the constant.
I was hoping @mike-lvov would approve the PR since his ask can be done in a subsequent one

sunny-viktoryia: I see that in your PR you return an image without checking if the image exists. 
image

Mine works for the broken links. Let's say https://google.com/broken-image.png - if the image doesn't exist it'll be replaced by the default picture. 

Probably, we can combine both approaches. 🙃

goran-peoski-work: Please use `determineAvatar()` instead of the constant

goran-peoski-work: and upgrade it if necessary to solve your use case

alexey-petrov: @sunny-viktoryia do you plan to use that `determineAvatar` or proceed without it?
It's used in many places in our app, so probably would be a good idea to extend it's logic

sunny-viktoryia: I do. But haven't had enough time for that so far. 😬

sunny-viktoryia: I have updated `determineAvatar` function and made it return an `onLoadError` function.
So now you can get
```ts
const [src, onError] = determineAvatar(options);
...
<img src={src} onError={onError}>
```
`onError` function chooses the fallback image in case the `src` does not exist.

The default pictures now are different - you can check the screenshot attached to the description. 

<hr>

My first approach was to check if the avatar exists in the `determineAvatar` directly, but those changes came with some inconveniences:
* the function would become `async`, which would complicate its usage
* we would get the avatar several times, which could affect our performance in the auditorium, for example, where we have to check many of avatars

@alexey-petrov @goran-peoski-work the PR is ready for re-review. 🙃 will be happy to hear your feedback once you have free time.

goran-peoski-work: This looks like a good idea, I might just request we switch from an array to an object, since that will give us more flexibility for backwards and forwards compat down the road. 

The idea is: a function that takes only one object as an argument and returns only one object as a result can just add more fields and/or ignore fields to those objects without requiring us to go over the project and update every single use.

src/utils/image.ts
```diff
@@ -78,32 +78,41 @@ export const getFirebaseStorageResizedImage = (
 // @see https://crypto.stackexchange.com/questions/8533/why-are-bitwise-rotations-used-in-cryptography/8534#8534
 const DIFFUSION_PRIME = 31;
 
-type DetermineAvatar = (options?: {
+type DetermineAvatarProps = {
```
goran-peoski-work, line 81: Please reserve the Props suffix for types that actually are props of components and add Options and Result to utility functions. This is a little convention that makes it easily determine the use.

src/utils/image.ts
```diff
@@ -78,32 +78,41 @@ export const getFirebaseStorageResizedImage = (
...
   email?: string;
   index?: number;
   partyName?: string;
   pictureUrl?: string;
   userInfo?: firebase.UserInfo;
   user?: User;
-}) => string;
+};
+
+type DetermineAvatarFunc = (
```
goran-peoski-work, line 91: No need of `Func` suffix. 

`DetermineAvatar` vs `determineAvatar` is the convention used elsewhere

src/utils/image.ts
```diff
@@ -78,32 +78,41 @@ export const getFirebaseStorageResizedImage = (
...
+  options?: DetermineAvatarProps
+) => [string, React.ReactEventHandler<HTMLImageElement>];
```
goran-peoski-work, line 93: This makes the function backward and forward compatible in the future. Adding and ignoring fields of an object is easier than dealing with array indices.

```suggestion
) => { src: string, onError: React.ReactEventHandler<HTMLImageElement> };
```

src/utils/image.ts
```diff
@@ -78,32 +78,41 @@ export const getFirebaseStorageResizedImage = (
...
 
   if (isDefined(index) && Number.isSafeInteger(index) && index >= 0) {
-    return list[index % list.length];
+    return [list[index % list.length], onImageError];
   }
 
+  return [url, onImageError];
+};
+
+type GenerateFallbackFunc = (options?: DetermineAvatarProps) => string;
```
goran-peoski-work, line 110:
```suggestion
type GenerateFallback = (options?: DetermineAvatarProps) => string;
```

src/utils/image.ts
```diff
@@ -78,32 +78,41 @@ export const getFirebaseStorageResizedImage = (
...
+
+export const generateFallback: GenerateFallbackFunc = (options) => {
```
goran-peoski-work, line 112:
```suggestion
export const generateFallback: GenerateFallback = (options) => {
```

src/utils/image.ts
```diff
@@ -121,3 +130,10 @@ export const determineAvatar: DetermineAvatar = (options) => {
 
   return list[hash % list.length];
 };
+
+const makeProfileImageLoadErrorHandler = (
+  src: string
+): React.ReactEventHandler<HTMLImageElement> => ({ currentTarget }) => {
+  currentTarget.onerror = null; // prevents looping
```
goran-peoski-work, line 137: Should an error be at least logged to console (in case the browser already doing it)?
Is it an important one to also involve Bugsnag? (if yes, we can just put a `@debt use Bugsnag` comment for now)

src/components/molecules/TableComponent/TableComponent.tsx
```diff
@@ -82,6 +82,32 @@ const TableComponent: React.FunctionComponent<TableComponentPropsType> = ({
 
   const itemClasses = classNames("TableComponent__item", itemStyles);
 
+  const userPictures = useMemo(
```
goran-peoski-work, line 85: In line with other uses in the project, `rendered` prefix provides visual cue for the intended use.

```suggestion
  const renderedUserPictures = useMemo(
```

src/hooks/useProfileModalFormDefaultValues.ts
```diff
@@ -24,7 +24,7 @@ export const useProfileModalFormDefaultValues: (
   useMemo(
     () => ({
       profileLinks: user.profileLinks ?? [],
-      pictureUrl: determineAvatar({ user }),
+      pictureUrl: determineAvatar({ user })[0],
```
goran-peoski-work, line 27: Switching from array to object as the result will make this easier to understand:

```suggestion
      pictureUrl: determineAvatar({ user }).src,
```

goran-peoski-work (changes requested):
Few minor naming convention suggestions and one greater request of changing the return type of `determineAvatar()`

src/utils/image.ts
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/utils/image.ts
sunny-viktoryia: nice suggestion, like it. 
fixed: https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/utils/image.ts
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/utils/image.ts
sunny-viktoryia: not sure if it's important... it'll be errored if our default image is missing. 
Anyway, added a debt comment.

src/components/molecules/TableComponent/TableComponent.tsx
sunny-viktoryia: done https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/hooks/useProfileModalFormDefaultValues.ts
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/utils/image.ts
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2760/commits/9cc6e165b25244d4599206263d7e1e47ab5f583f

src/components/molecules/ProfilePictureInput/ProfilePictureInput.tsx
```diff
@@ -65,14 +65,20 @@ export const ProfilePictureInput: React.FunctionComponent<ProfilePictureInputPro
...
   return (
     <div className="ProfilePictureUploadForm">
       <div
         className="ProfilePicturePreviewContainer"
         onClick={() => uploadRef.current?.click()}
       >
         <img
-          src={determineAvatar({ pictureUrl, userInfo: user })}
+          src={pictureSrc}
+          onError={onPictureSrcError}
```
goran-peoski-work, line 81: Interesting enough, we can also try to be clever and use spread `{ ... determineAvatar() }` here... But current code is also good, it's easy to follow.

src/components/molecules/TableComponent/TableComponent.tsx
```diff
@@ -81,6 +81,34 @@ const TableComponent: React.FunctionComponent<TableComponentPropsType> = ({
...
+        });
+
+        return (
+          <img
+            onClick={() => openUserProfileModal(user.id)}
+            key={user.id}
+            className="TableComponent__profile-icon"
+            src={profilePic}
+            onError={onLoadError}
+            title={(!user.anonMode && user.partyName) || DEFAULT_PARTY_NAME}
```
goran-peoski-work, line 100: `anonMode` was removed previousle (original comment in `image.ts`). Is that one back on the table now?

```
// checking user.anonMode ruled against by https://github.com/sparkletown/internal-sparkle-issues/issues/1615#issuecomment-991054012
```

goran-peoski-work (approved):
LGTM. Approving.

Also, @sunny-viktoryia you might want to check that one comment about `annonMode`. I'm not sure what the current requirement for it is.

src/components/molecules/TableComponent/TableComponent.tsx
sunny-viktoryia: removed `anonMode` check here: https://github.com/sparkletown/sparkle/pull/2760/commits/db98df9b57d0e55acca8872b685e36307a96aa59

PR #2781, opened 2022-01-25
Adds world owners; Displays proper world & spaces

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1530
is relevant(but doesn't close) to https://github.com/sparkletown/internal-sparkle-issues/issues/1642

Changes: you can be either the world owner or a space owner.

If you are a world owner, you can see and edit the world & all spaces inside the world:

image

If you are a space owner, you can see the world and all the spaces inside, but you can't see settings nor can you edit the spaces that you don't own

image

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -53,9 +53,13 @@ service cloud.firestore {
         return get(/databases/$(database)/documents/worlds/$(venueData().worldId)).data;
       }
 
+      function checkIfSpaceOrWorldOwner(userId) {
+        return $(userId) in role('admin').users && ($(userId) in venueData().owners || $(userId) in world().owners)
+      }
+
       match /events/{event} {
         allow read: if true;
-        allow write: if request.auth.uid in venueData().owners;
+        allow write: if checkIfSpaceOrWorldOwner(request.auth.uid);
       }
 
       match /chatMessagesCounter/{shardId} {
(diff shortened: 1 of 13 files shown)


firestore.rules
```diff
@@ -53,9 +53,13 @@ service cloud.firestore {
         return get(/databases/$(database)/documents/worlds/$(venueData().worldId)).data;
       }
 
+      function checkIfSpaceOrWorldOwner(userId) {
+        return $(userId) in role('admin').users && ($(userId) in venueData().owners || $(userId) in world().owners)
```
colinhowe, line 57:
```suggestion
        return $(userId) in role('admin').users || $(userId) in venueData().owners || $(userId) in world().owners
```
My understanding of the "admin" role is that is for Sparkle staff to be able to edit anything.

This is a nice solution btw. Glad it worked out nice and tidy :)

src/hooks/worlds/useWorlds.ts
```diff
@@ -7,11 +7,10 @@ import { World } from "api/world";
 import { worldConverter } from "utils/converters";
 import { WithId } from "utils/id";
 
-export const useOwnWorlds = (userId?: string): WithId<World>[] => {
+export const useWorlds = (): WithId<World>[] => {
   const firestore = useFirestore();
   const worldsRef = firestore
     .collection("worlds")
-    .where("owners", "array-contains", userId ?? "")
```
colinhowe, line 14: Why have you refactored this? I feel like I'm missing a motivation for this change

src/pages/WorldsDashboard/WorldsDashboard.tsx
```diff
@@ -20,9 +20,21 @@ import ARROW from "assets/images/admin/dashboard-arrow.svg";
...
 
-  const worlds = useOwnWorlds(user.userId);
+  const ownedUniqueWorldIds = useMemo(
+    () => [...new Set(ownedVenues.map((venue) => venue.worldId))],
+    [ownedVenues]
+  );
+
+  const worlds = useWorlds();
+
+  // NOTE: We could do this by forming a firebase query where([world.id, 'in', ownedUniqueWorldIds]), but it has a limit of 10 items in ownedUniqueWorldIds.
```
colinhowe, line 32: Or, by doing `world.owners array-contains user.id`

colinhowe: I'm happy with how you've done the filtering in the frontend btw

colinhowe (changes requested):
Minor questions. Otherwise, looks great!

src/hooks/worlds/useWorlds.ts
mike-lvov: Think of this change as of 2 actions: 
1) Create new `useWorlds` hook to fetch all available worlds
2) Deletion of an old `useOwnWorlds` because it was not used anywhere

src/pages/WorldsDashboard/WorldsDashboard.tsx
mike-lvov: Any of these queries would still result in a hard limit of 10 entries. I wanted to avoid it

colinhowe: There isn't a hard limit of 10 when using `array-contains`. There is a hard limit of 10 when using `array-contains-any`.
Docs here: https://firebase.google.com/docs/firestore/query-data/queries#limitations_2

Just confirmed by creating a document with an array of 12 things and querying it with `array-contains`

colinhowe: No need to fix anything here. Just an FYI

colinhowe (approved):

PR #2782, opened 2022-01-26
fixed create table reference value

alexey-petrov:
Reference: https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1643123467022200

* Fixed table reference value
* Changed new table name from `Table #X` to `New Table` ( avoided complex logic )


src/utils/table.ts
```diff
@@ -6,19 +6,22 @@ export const generateTable: (props: {
   tableNumber: number;
```
alexey-petrov, line 6: 4 files to review, everything else has been `prettiered`

functions/venue.js
```diff
@@ -711,31 +711,42 @@ exports.deleteTable = functions.https.onCall(async (data, context) => {
   const tables = docData.config.tables || defaultTables;
 
   const index = tables.findIndex((val) => val.reference === tableName);
+
```
alexey-petrov, line 714: this

src/components/molecules/TablesUserList/TablesUserList.tsx
```diff
@@ -213,7 +210,10 @@ export const TablesUserList: React.FC<TablesUserListProps> = ({
       {allowCreateEditTable && (
         <StartTable
           defaultTables={defaultTables}
-          newTable={generateTable({ tableNumber: tables.length + 1 })}
+          newTable={generateTable({
```
alexey-petrov, line 213: this

functions/venue.js
```diff
@@ -711,31 +711,42 @@ exports.deleteTable = functions.https.onCall(async (data, context) => {
...
+  tables.splice(index, 1);
+
+  const recalculatedTables = tables.map((table, i) => ({
+    ...table,
+    reference: `Table ${i + 1}`,
```
goran-peoski-work, line 719: Not sure why this `reference` is needed and more important where.
Is it breaking some cloud function? Maybe that one should be updated. 
Is it used in other places? Maybe even the simulator.

Ideally tables would have `id` or Firestore key to reference them, but in this case, if only their position in an array is relevant, the change of the `reference` value is either not necessary or used in specific places

alexey-petrov:
Closed in favor of https://github.com/sparkletown/sparkle/pull/2788

PR #2785, opened 2022-01-26
prettier formatted files

alexey-petrov:
Ref: https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1643194178030800

Not seeing any recent `prettier` config/rules update, but still experiencing the same broken build issue due to invalid formatting

So this PR contains all those files reformatted


alexey-petrov:
Closed as it was a local issue

PR #2789, opened 2022-01-27
table create fix + catch error type declarations

alexey-petrov:
Reference: https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1643123467022200

Fixed table reference value
Changed new table name from Table #X to New Table ( avoided complex logic )

Follow up for #2782


functions/venue.js
```diff
@@ -711,31 +711,26 @@ exports.deleteTable = functions.https.onCall(async (data, context) => {
...
-  } else {
-    docData.config.tables.splice(index, 1);
   }
 
-  admin.firestore().collection("venues").doc(spaceId).update(docData);
+  tables.splice(index, 1);
+
+  const recalculatedTables = tables.map((table, i) => ({
+    ...table,
+    reference: `Table ${i + 1}`,
```
denisdimitrov, line 723: I think remapping the references might be a bad idea, and it can lead to edge cases. I think everything is fine except for this part, maybe we should use uuid for new table references and keep the old ones as they are.

What happens if you and me join a table and at that point in time the reference is `Table 10` after that someone removes another table, adds a new one and our table reference becomes `Table 9` If a new person tries to join our table with you and me, he will probably join another reference and be alone, it's a wild guess, but sounds like this is going to happen.

alexey-petrov: fixed 131ac0b

denisdimitrov (approved):
LGTM!

PR #2790, opened 2022-01-31
add unique table reference for new tables

denisdimitrov:
The reference was not unique and removing -> create a table was not working because of that.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1686


denisdimitrov:
Dupe of https://github.com/sparkletown/sparkle/pull/2790 which is already closed.

PR #2793, opened 2022-02-02
Remove height specification on tables outside jazz bar

colinhowe:
min-height is already set by the CSS on TableComponent and specifying the
height directly can lead to clipping issues when the component will
resize nicely anyway

### DIFF
--- a/src/components/molecules/TableComponent/TableComponent.tsx
+++ b/src/components/molecules/TableComponent/TableComponent.tsx
@@ -75,7 +75,6 @@ const TableComponent: React.FunctionComponent<TableComponentPropsType> = ({
     isJazzBar
       ? {}
       : {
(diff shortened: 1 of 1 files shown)


src/components/molecules/TableComponent/TableComponent.tsx
```diff
@@ -75,9 +75,8 @@ const TableComponent: React.FunctionComponent<TableComponentPropsType> = ({
     isJazzBar
       ? {}
       : {
-          height: `${table.rows && table.rows * 50 + 65}px`,
```
colinhowe, line 78: Removed this. There's a min-height on the component already and it doesn't need anything further. It grows as it should.

goran-peoski-work (approved):
LGTM

PR #2794, opened 2022-02-02
Show world events on entrance pages

colinhowe:
This had not been updated after the world event
refactor.

Resolves https://github.com/sparkletown/internal-sparkle-issues/issues/1691

### DIFF
--- a/src/hooks/events.ts
+++ b/src/hooks/events.ts
@@ -16,10 +16,7 @@ export interface VenueEventsData {
   events: WorldEvent[];
 }
 
-export const useSpaceEvents = ({
-  worldId,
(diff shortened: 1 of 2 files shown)


src/hooks/events.ts
```diff
@@ -16,10 +16,7 @@ export interface VenueEventsData {
   events: WorldEvent[];
 }
 
-export const useSpaceEvents = ({
```
colinhowe, line 19: Refactored into two methods: `useSpaceEvents` and `useWorldEvents`. Makes the APIs simpler and cleaner.

alexey-petrov (approved):
LGTM

PR #2802, opened 2022-02-03
Add posterhall and animatemap icons

sunny-viktoryia:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1655

in the designs, I've found a missing `animatemap` icon as well, so I've added it too.
Also, worth noting that although `posterhall` icon is added, there is no `posterpage` icon which is also a space template. 

image
image


goran-peoski-work (approved):
Would have been nice to see the icons as a screenshot in the description.

Posters don't (or didn't) exist since I've taken few from my screen, cropped them and put them in the assets. Same can be done for the missing I guess.

mike-lvov (approved):
LGTM. NOTE: It's known that AnimateMap doesn't work as of now, right?

PR #2808, opened 2022-02-07
use related venues for parent back options

denisdimitrov:
The options for the parent back button are now using the related venues only (world spaces)

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1685


sunny-viktoryia (approved):
Looks good to me. 🙃 

Pearl has also requested in the issue to modify the label of the input. Please, remove the word "parent":
> Select the ~~parent~~ space for the "back" button

PR #2817, opened 2022-02-08
Ran npm run reformat

goran-peoski-work:
Two files with added "ignore unused variables" directives that should be removed later on:
- `src/components/templates/Jazzbar/JazzBar/JazzBar.tsx`
- `src/components/attendee/VideoHuddle/VideoHuddle.tsx`


mike-lvov (approved):
LGTM

PR #2818, opened 2022-02-08
fix prettier to the specific version of 2.2.1

goran-peoski-work:
Just because `Prettier` makes code uglier, even more so if you upgrade the `2.2.1` to `2.2.5`


sunny-viktoryia (approved):
approving. 🙃

PR #2821, opened 2022-02-09
Boring video

colinhowe:
This PR creates a new set of components for doing video / audio calls.

For anyone reviewing this: please focus on the APIs and the logic within them. There is no real attempt to style this beyond what was convenient to get this tested. Specifics of the DOM and CSS will be changed in a future PR.


src/api/venue.ts
```diff
@@ -230,3 +230,14 @@ export const unsetTableSeat = async (
   userId: string,
   { venueId }: Pick<VenueTablePath, "venueId">
```
goran-peoski-work, line 231: Single `options` arg?

Something like:
```ts
type UnsetTableSeatOptions = {
  userId: UserId;
  spaceId: typeof VenueTablePath["venueId"]; // or just SpaceId ?
}
```

src/api/venue.ts
```diff
@@ -230,3 +230,14 @@ export const unsetTableSeat = async (
...
 ) => getUserSeatedTableRef(userId, venueId).delete();
+
+export const setProjectedVideoTrackId = async (
+  venueId: string,
+  trackId: string
```
goran-peoski-work, line 236:
```suggestion
options: { spaceId: SpaceId; trackId: TrackId }
```

src/components/attendee/AttendeeLayout.tsx
```diff
@@ -21,20 +22,30 @@ export const AttendeeLayout: React.FC = () => {
     };
   }, []);
 
+  if (!userId) {
+    return <p>Loading</p>;
```
goran-peoski-work: There is a `<Loading />` component

src/components/attendee/AttendeeLayout.tsx
```diff
@@ -21,20 +22,30 @@ export const AttendeeLayout: React.FC = () => {
...
-          <ChatContainer />
-          {userId && <VideoHuddle userId={userId} />}
-        </div>
-      </main>
-
-      <AttendeeFooter />
-    </VideoHuddleProvider>
+    <VideoCommsProvider userId={userId}>
+      <HuddleProvider>
+        <header></header>
```
goran-peoski-work, line 33: Used for anything?

src/components/attendee/AttendeeLayout.tsx
```diff
@@ -21,20 +22,30 @@ export const AttendeeLayout: React.FC = () => {
...
+        <main>
+          <section className={styles.Space}>
```
goran-peoski-work, line 35: We should make better names than this, it's too general.

src/components/attendee/AttendeeLayout.tsx
```diff
@@ -21,20 +22,30 @@ export const AttendeeLayout: React.FC = () => {
...
+            <VenuePage />
+          </section>
+          <div className={styles.LayerUi}>
+            <ChatContainer />
+            {/*
+          TODO VideoHuddle being available depends on the context of the space we're in.
+          This probably needs more thought as to what API it provides.
+        */}
+            {userId && <VideoHuddle userId={userId} />}
```
goran-peoski-work: Any place in the context for these edge cases https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1644330820946409?thread_ts=1643886209.361969&cid=C01ER4M61TL ?

src/components/attendee/VideoComms/VideoCommsParticipant.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface VideoCommsParticipantProps {
+  participant: Participant;
+  isLocal?: boolean;
+  extraButtons?: ButtonConfig[];
+}
+
+export const VideoCommsParticipant: React.FC<VideoCommsParticipantProps> = ({
+  participant,
+  isLocal,
+  extraButtons = [],
```
goran-peoski-work, line 20: These might be better passed as `children`. Could also allow for other kinds of components.

src/components/attendee/VideoComms/VideoCommsParticipant.tsx
```diff
@@ -0,0 +1,42 @@
...
+}) => {
+  return (
+    <>
```
goran-peoski-work, line 23: How about...
```suggestion
    <div className="VideoCommsParticipant">
```

src/components/attendee/VideoComms/VideoCommsProvider.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { noop } from "lodash";
+
+import { TwilioImpl } from "./internal/TwilioImplementation";
```
goran-peoski-work, line 4: `Impl`? So much Java...

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+export enum VideoCommsStatus {
+  Disconnected = "DISCONNECTED",
+  Connecting = "CONNECTING",
+  Connected = "CONNECTED",
+  Errored = "ERRORED",
+}
+
+export interface VideoCommsProviderProps {
+  userId: string;
+  children: React.ReactNode;
```
goran-peoski-work, line 12: This should not be necessary, `React.FC<VideoCommsProviderProps>` will add the `children` field

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+interface Track {
+  kind: "video" | "audio";
+  id: string;
```
goran-peoski-work, line 17: Maybe we add one more line in `./src/types/id.ts` ?

```suggestion
  id: TrackId;
```

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+  enabled: boolean;
+}
+
+export enum VideoSource {
```
goran-peoski-work, line 21: Same concerns about enums as before

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+  Webcam = "WEBCAM",
+  Screenshare = "SCREENSHARE",
+}
+
+export interface VideoTrack extends Track {
```
goran-peoski-work, line 26: Not sure if `interface`+`extends` would be better than just a union with `type`+`&`.
After all, why even have a base type?

Can just make them separate and add
`export type Track = VideoTrack | AudioTrack` 
with the `kind` already fixed at one specific value for them.

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+type JoinChannelFunc = (
+  userId: string,
+  channelId: string,
+  enableVideo: boolean,
+  enableAudio: boolean
+) => void;
+type DisconnectFunc = () => void;
+type ShareScreenFunc = () => void;
+
+export interface LocalParticipant extends Participant {}
```
goran-peoski-work, line 67: No need of both I guess.

```suggestion
export interface LocalParticipant extends Participant {}
```

Maybe something similar like the previous suggestion:
`export type Participant = LocalParticipant | RemoteParticipant;`

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export interface VideoCommsContextType {
+  channelId?: string;
```
goran-peoski-work, line 70: branded?

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export interface StateUpdateCallbackParams {
+  localParticipant?: LocalParticipant;
+  remoteParticipants: Participant[];
+  status: VideoCommsStatus;
+  isTransmittingAudio: boolean;
+  isTransmittingVideo: boolean;
+}
+
+export type StateUpdateCallback = (update: StateUpdateCallbackParams) => void;
```
goran-peoski-work, line 94: From usage elsewhere, kind of easy to determine by looking at the name:
- ends in `Props` it's for a component
- ends in `Options` it's for a utility function
- ends in `Result` it's from a utility function
- ends in `Context` it is for a listener/callback (I guess we're avoiding `event` because of components)

And we also avoid `args`, `params` and combos of those for naming... well, arguments/parameters.

```suggestion
export type StateUpdateCallback = (context: StateUpdateCallbackContext) => void;
```

src/components/attendee/VideoHuddle/VideoHuddle.tsx
```diff
@@ -1,57 +1,35 @@
...
-  setChannelId: () => {},
-});
-
 interface VideoHuddleProps {
   userId: string;
 }
 
 export const VideoHuddle: React.FC<VideoHuddleProps> = ({ userId }) => {
-  const { channelId } = useContext(VideoHuddleContext);
+  const { inHuddle } = useVideoHuddle();
```
goran-peoski-work: Shouldn't at least the `userId` be passed in here? Will there ever be a case for not the currently used user?

src/components/attendee/VideoHuddle/VideoHuddle.tsx
```diff
@@ -72,41 +50,25 @@ const ActiveVideoHuddle: React.FC<ActiveVideoHuddleProps> = ({
...
+        {/*  TODO shouldMirrorVideo */}
         {localParticipant && (
-          <div className={styles.VideoContainer} key={localParticipant.sid}>
-            <VideoParticipant
+          <div className={styles.VideoContainer} key={localParticipant.id}>
+            <VideoCommsParticipant
+              isLocal={true}
               participant={localParticipant}
-              shouldMirrorVideo
+              extraButtons={extraButtons}
```
goran-peoski-work, line 48: Suggested above, instead of just buttons, if `VideoCommsParticipant`...
(how about `Vidcom` as a system? Sounds weird? O.o) 
... takes in children, we can be flexible at what kind of buttons/icons/components we can add

src/components/attendee/VideoHuddle/useVideoHuddle.tsx
```diff
@@ -1,24 +1,103 @@
-import { useContext } from "react";
+import React, { useContext, useMemo, useState } from "react";
+import { IconDefinition } from "@fortawesome/free-solid-svg-icons";
 
-import { VideoHuddleContext } from "./VideoHuddle";
+import { useVideoComms } from "../VideoComms/hooks";
+import { VideoTrack } from "../VideoComms/types";
+
+interface HuddleContextType {
```
goran-peoski-work, line 7: Yes, it is a type :) 

```suggestion
interface HuddleContext {
```

src/components/attendee/VideoHuddle/useVideoHuddle.tsx
```diff
@@ -1,24 +1,103 @@
...
+  inHuddle: boolean;
+  setInHuddle: (cb: (prevInHuddle: boolean) => void) => void;
```
goran-peoski-work, line 9: Should we just not make callbacks accept objects? Something like `ctx: SetInHuddleContext` that maybe has only one field now, but can be expanded later on w/o rewriting other listeners. This way we can pass more info as new use cases arise.

src/components/attendee/VideoHuddle/useVideoHuddle.tsx
```diff
@@ -1,24 +1,103 @@
...
+
+  extraButtons: ButtonConfig[];
+  setExtraButtons: (buttons: ButtonConfig[]) => void;
```
goran-peoski-work, line 12: Shouldn't listeners start with `onInHuddle` and `onSetExtraButtons`? How about `onChildren` instead? (see above)

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
```diff
@@ -0,0 +1,3 @@
+.mirrored {
```
goran-peoski-work, line 1: no BEM?

goran-peoski-work (changes requested):
Reviewed half the files. Added few comments, questions and suggestions.

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+  return {
+    joinChannel,
+    disconnect,
+    shareScreen,
+    startAudio,
+    stopAudio,
+    startVideo,
+    stopVideo,
+  };
```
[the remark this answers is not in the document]
colinhowe: That's my next possible refactor. When I first built this is maintained more internal state.. Then I simplified it right down to re-generating state as that led to fewer bugs ;)

src/components/attendee/VideoComms/types.tsx
colinhowe: Twilio has different types for Video and Audio with no shared base. This has led to some annoying code in our implementation. e.g. look at `Tracklike`. That wouldn't be needed if they had a shared base interface.

Can't say I care particularly strongly though.

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
colinhowe: I don't think it's necessary when we're using styled components. Just looks like more noise to me.

mike-lvov: I'd still say that it improves the overall project navigation. When you see .mirrored-${id}, you might still have multiple occurrences of this class. Making the naming consistent helps to avoid such situations

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
+import Twilio from "twilio-video";
+
+export enum VideoCommsStatus {
+  Disconnected = "DISCONNECTED",
+  Connecting = "CONNECTING",
+  Connected = "CONNECTED",
+  Errored = "ERRORED",
+}
```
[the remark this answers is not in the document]
mike-lvov: The argumentation in the article is a bit opinionated as for me. 

> This brings us to the practical problem with TypeScript's violations of its own type-level extension rule. Tools like Babel, webpack, and their plugins are all designed for JavaScript first, so TypeScript support is just one of their many features. Sometimes, TypeScript support doesn't receive as much attention as JavaScript support, which can lead to bugs.

Enums is not a new feature, nor it is something exotic. It's a well-known and supported feature and I don't see any reasonable arguments against its usage

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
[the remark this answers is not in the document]
goran-peoski-work: @mike-lvov are we thinking of the same definition of "pure function"?

https://en.wikipedia.org/wiki/Pure_function

I'm considering the "react way" of at least having side effects  (`like a Reac.FC` is pure), but having the functions depend on explicit input, something like an object that has the state/context outside so that we can add/remove functions in the future directly at the file top level, not being tied to a closure.

I'm considering any hook that calls those functions will provide the closure (have the state/context object)

src/components/attendee/VideoComms/types.tsx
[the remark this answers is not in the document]
goran-peoski-work: > Enums is not a new feature, nor it is something exotic.
I'm not sure the article claims either of those two, not least as a reason for anything.

Enums get transpiled to a quirky JS object, you can't even just do `Object.values()` and get what you've written in the TS code, it's extra.

But, the article is there so I don't have to retype the argument of "there is an object created in JS, it is not working like all the other types do, doesn't work like `type`, doesn't work like `interface`)

src/components/attendee/VideoComms/types.tsx
goran-peoski-work: I always consider it a code smell naming something `*Base`. There is usually a better way of dealing with. In this case, I think union type is more robust approach as it doesn't try to invent shared interface but gives you the ability to use one or the other.

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
goran-peoski-work: @mike-lvov @colinhowe I have shared some good example here that I've noticed a 3rd party library has used and it was helpful for us:
https://doesitsparkle.slack.com/archives/C01ER4M61TL/p1644404537802149?thread_ts=1644335099.082349&cid=C01ER4M61TL

@colinhowe semantical meaning can be added in other ways just as well. we've not touched `data-*` props much.
I guess people in React avoid them, not necessarily only because they don't have camel casing.
The BEM approach is more versatile than just project navigation. It makes it easy to report issues with precise data and makes it easy to test solutions by just the browser dev tools.

firestore.rules
```diff
@@ -45,6 +45,11 @@ service cloud.firestore {
     match /venues/{venue} {
       allow read: if true;
 
+      // Allow anyone to set the centrally project track.
+      // TODO Decide whether this belongs on the venue, or, somewhere else.
+      allow update: if (request.resource.data.diff(resource.data).affectedKeys()
+        .hasOnly(['projectedVideoTrackId']));
+
```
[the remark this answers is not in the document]
goran-peoski-work: The problem with just using roles is that this kind of logic will pop up in different places.

How will moving this data to a different place work?
Can you have a sub-collection or fields that aren't being read because you don't have the rule permission for them?
Does it require a function or does it work out of the box?

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
```diff
@@ -0,0 +1,377 @@
...
+
+const wrapLocalAudioPublication = (
+  publication: Twilio.LocalAudioTrackPublication
+): AudioTrack => wrapTrack(publication.track, "audio", publication.trackSid);
+
+/**
+ * Used with filter to allow Typescript to convert arrays that might have nulls
+ * in them into arrays of just a specific type.
+ */
+const notNull = <T,>(value: T | null): value is T => value !== null;
```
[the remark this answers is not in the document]
goran-peoski-work: Most likely where `isDefined` resides, if not, `lodash` has two already `isNil()` and `isNotNil()` that I think are already typed properly.

Personally, I think we can clean up and beef up that `./src/utils/types.ts` file for our common uses/idioms.

On other (OOS) notes:
- it's kind of confusing having files like `./src/utils/types.ts` and `./src/types/utility.ts` but they do serve their purpose;
- we could think about having some isomorphic code with distinguishable naming usable by the TS functions/scripts as well

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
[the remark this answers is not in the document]
goran-peoski-work: > I doubt it would make this code more readable

Readability is one issue.

Being able to easily test something is another. The more predictable a function is (like one that doesn't depend in implicit input), the easier is to give it value, check the result, confirm it works or doesn't.

firestore.rules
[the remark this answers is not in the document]
colinhowe: This is just for a prototype. I have no intention of this going into production. Agree with all the concerns here.

I think a cloud function is the right move here.

src/components/attendee/VideoComms/VideoTrackDisplay.tsx
```diff
@@ -0,0 +1,32 @@
+import { VideoTrack } from "../VideoComms/types";
+
+import { TwilioTrackDisplay } from "./internal/TwilioTrackDisplay";
+
+interface VideoTrackDisplayProps {
+  track: VideoTrack;
+  isMirrored?: boolean;
+}
```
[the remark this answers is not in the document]
colinhowe: There was a thread from Jess about this just last week:
> it’s pretty disorienting for our videos to not be mirrored

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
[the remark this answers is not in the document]
colinhowe: I started moving these functions out into top-level functions. The number of arguments that would need passing in would make them worse in my opinion. I might come back to this later and see what can be done to make this easier to work with.

src/components/attendee/VideoComms/internal/VideoCommsControls.tsx
```diff
@@ -0,0 +1,29 @@
...
+    startAudio,
+    stopAudio,
+    startVideo,
+    stopVideo,
+    isTransmittingAudio,
+    isTransmittingVideo,
+  } = useVideoComms();
+
+  return (
+    <>
```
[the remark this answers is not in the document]
colinhowe: There's going to be a follow-on PR that makes sure all the HTML generated is what we actually want

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
colinhowe: I think CSS source maps make it easy to find where class is defined when you're debugging things.

That said, I don't feel particularly strongly about this.

`TwilioTrackDisplay__mirrored`?

src/components/attendee/VideoComms/internal/TwilioImplementation.tsx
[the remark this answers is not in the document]
goran-peoski-work: Number of arguments is 1. Just make a type for it

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
goran-peoski-work: That should be a modifier, no?

firestore.rules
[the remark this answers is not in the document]
goran-peoski-work: I'm unclear about the intent of the code itself. I'm guessing I will have to know some Twillio specifics like what a project video track is and how it relates to our use cases.

Maybe when ready (out of DRAFT) we can check this PR out with the preview link.

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
mike-lvov: `TwilioTrackDisplay--mirrored`

firestore.rules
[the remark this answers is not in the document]
colinhowe: https://7.sparkle.space/in/w/devplayground/s/experiment ;)

src/components/attendee/VideoComms/types.tsx
```diff
@@ -0,0 +1,94 @@
+import Twilio from "twilio-video";
+
```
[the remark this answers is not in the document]
colinhowe: Renamed. I'm not convinced by moving it to the top-level folder. I think these types belong with the video comms stuff

PR #2822, opened 2022-02-10
Move global styles to admin global styles

colinhowe:
This prevents tailwind from interfering with the attendee side


src/core/AppRouter/AdminSubRouter.tsx
```diff
@@ -25,6 +25,8 @@ import { WorldsDashboard } from "pages/WorldsDashboard";
 
 import { AdminVenueView } from "components/organisms/AdminVenueView";
 
+import "scss/adminGlobal.scss";
```
goran-peoski-work, line 28:
```suggestion
import "scss/admin.scss";
```

Two notes, though:

1. Should not be here. Router has too many concerns already.

    This import should be made at page level.
    Assuming all admin pages would have a single framing layout,
    that's only one component the file should be included it.

    Example: if both `WorldsDashboard` and `SpacesDashboard` components have `AdminLayout` one, then only   `AdminLayout` needs to import the SCSS file.

2. import may have time dimension, not just spatial one, what would the issue/solution be in that case?
Example:
   1. Navigate to attendee URL (no style added to DOM)
   2. Navigate to admin URL (the admin styles added to the DOM)
   3. Navigate back to the attendee URL (are the admin styles still part of the DOM?)

goran-peoski-work (changes requested):
The principle is there. Some implementation questions to puzzle about.

src/core/AppRouter/AdminSubRouter.tsx
colinhowe: 2. Just checked and this is a problem. I'll see if there's another way to do this.

goran-peoski-work (approved):
Looks good, with a caveat:

Login screens that get displayed from `/admin` but will not use tailwind should be checked if they will open new tab or if there isn't a real issue

PR #2823, opened 2022-02-10
move the scss files to the correct dir

denisdimitrov:
Moves the scss files to the `scss` dir to make the import paths easier and global


src/components/attendee/AttendeeFooter.module.scss
```diff
@@ -1,4 +1,4 @@
-@import "./scss/AttendeeMixins.scss";
+@import "scss/AttendeeMixins.scss";
```
goran-peoski-work, line 1: How about `@use scss/attendee/mixins.scss` ?

I'm guessing there might be `functions.scss`, `spacing.scss` etc.
And this also goes away with not adding camel case :)

NOTE:

`@use mixins.scss` makes module level items accessible with `mixins.$whitespace` since it has automatic namespace. I suggest using this approach.

The less robust alternative is `@use mixins.scss as *` or similar (can check https://sass-lang.com/documentation/at-rules/use)

goran-peoski-work (changes requested):
Can't attest to the actual styles, did a code review only.

General comment about the import should apply to all the CSS files.

src/components/attendee/AttendeeFooter.module.scss
denisdimitrov: I chose the robust way because I don't want to go through all of the variables and functions. Check it here: https://github.com/sparkletown/sparkle/pull/2823/commits/22cfe933d386ed598c282f06fdd7d62a1558177f

PR #2825, opened 2022-02-10
redesign jazzbar and conversation space tables

denisdimitrov:
Redesigns the jazzbar and conversation space tables. The design looks almost identical to the design mock ups and the styles are the same, however the jazzbar layout is missing and the end result might be slight different.

tables-redesign

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1717

### DIFF
--- a/src/components/molecules/TableComponent/TableComponent.scss
+++ b/src/components/molecules/TableComponent/TableComponent.scss
@@ -1,0 +1,99 @@
+@use "scss/attendee/mixins.scss"as *;
+@use "scss/attendee/variables.scss"as *;
+
+.TableComponent {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: $whitespace * 2 $whitespace * 3.5 $button-height * 2;
+  flex-wrap: wrap;
+
+  &__delete-button {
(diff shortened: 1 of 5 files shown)


src/components/molecules/TableComponent/TableComponent.scss
```diff
@@ -0,0 +1,99 @@
+@use "scss/attendee/mixins.scss"as *;
+@use "scss/attendee/variables.scss"as *;
```
colinhowe, line 2: Weird spacing going on.
```suggestion
@use "scss/attendee/mixins.scss" as *;
@use "scss/attendee/variables.scss" as *;
```

src/components/molecules/TableComponent/TableComponent.scss
```diff
@@ -0,0 +1,99 @@
+@use "scss/attendee/mixins.scss"as *;
```
colinhowe, line 1: This file (`TableComponent`) should be a SCSS module. Take a look at how some of the other code is written.

colinhowe (changes requested):

PR #2841, opened 2022-02-16
 Docs minor reformat and update of code-style.md

goran-peoski-work:
Minor reformat of markdown `./docs` and addition of a document for code style and related issues.


docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+1. use your own best judgement;
+2. try to follow style of code already in place;
+3. notify others of issues and possible improvements.
+
+---
+
+## Comments (general)
+
+There usually isn't a formal way of writing comments. Most common are simple
+line comments, not block ones. However, few tags, due to familiarity with JSDoc,
```
colinhowe, line 20:
```suggestion
line comments, not block ones. However, a few tags, due to familiarity with JSDoc,
```

docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Previous Different systems
+
+There has been a mix of different style systems in use in the past. Some
+include:
+
+- _Styled Components_
+- SCSS provided by _Create React App_ (_CRA_), not including modules
+- _Bootstrap_ styles
+- global overrides for the most basic HTML tags
```
colinhowe, line 42: Given that we're deleting all this in the redesign branch... do we need this section?

docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+
+### SCSS modules + Tailwind parity
+
+In order to make code style more uniform while using two different systems, here
+are some naming conventions that can be used:
+
+- import `classnames` library as `cn` for brevity since it will be used
+  extensively;
+- import Tailwind class strings from another file with `import * as TW from`;
+- import SCSS module class maps as `import CN from`;
```
colinhowe, line 68: I think `cn` and `CN` is going to lead to confusion in the future.

docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+Prefer shorthand (absolute) imports over relative ones. You can access every
+directory under `. /src` with just typing its name. Example: instead of
+`../../../../settings`, you can write`settings`.
+
+##
+
+Exports
+
+Should try to not export types and interfaces that aren 't used elsewhere
+outside that single file.There is a tool(`npm run prune`) that can generate a
```
colinhowe, line 154:
```suggestion
outside that single file.There is a tool (`npm run prune`) that can generate a
```

docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+execute before each commit.As for components, if they're only imported by a
+sibling `index.ts` file, then there should be a check if that component is used
+anywhere and if it's good to keep or delete it.
+
+###
+
+Types vs Interfaces
+
+Usually types declared with `interface` were used to define a components
+props.However`type` s syntax allows for more robust definitions and should be
```
colinhowe, line 165:
```suggestion
props. However, `type`s syntax allows for more robust definitions and should be
```

docs/code-style.md
```diff
@@ -0,0 +1,333 @@
...
+}
+```
+
+would be
+
+```ts
+type AllowedColor = "red" | "green" | "blue";
+```
+
+For extensive range in values, branded types could be more useful.
```
colinhowe, line 195: Disagree with this. @mike-lvov does too if I remember right.
Enums are a feature of the language we are using. We shouldn't be afraid to use those features.

colinhowe (approved):
Few minor things, otherwise, good work!

docs/code-style.md
goran-peoski-work: A historical overview as an intro. Maybe just a tad of "here is what we got away from" before someone recommends Bootrstrap again a year from now :)

Anyway, different systems are in use now as well, the need of co-exist is there, just the named changed.

docs/code-style.md
goran-peoski-work: Maybe.  Anything that's close to `TW` as a short and all caps due to being a constant would work. 
`M` for module? `CSS`? 

I'd argue low caps and all caps is indicative enough of what is function and what is constant even before compiler does the check. Still, the document is written in a way to not be mandating anything, but more like giving an example.

So, if there is a possibility of confusion, we can add a line about that, maybe provide different naming ideas in it.

I will merge this as is now and we can update it with the next idea we get after the experience.

docs/code-style.md
goran-peoski-work: **TL;DR**:

If there is a need to remove the passage from this document, be it for it causing worse code, adding more confusion, stating in-factual claims or simply just taking up space, we can do so with a next PR.

----

The longer part (recorded here so it will be safe to forget):

My idea isn't to remove passages though, but add more, maybe even to the point of a single document like this one becoming a reference for several ones covering more specific topics. 

In that light, I was more interested if I should put some naming suggestions for the `enums` we're already using and might be using in the future since right now their casings are all over the place. I just figured I might be overdoing it to try to account for every single naming in this first draft.

On the comment above, I don't know what precisely `this` is being disagreed to, I'm guessing `enums` usage,
but if there is a signal of an issue with this document, it can resolved by means mentioned above.

However, the comment has several notions I find puzzling and can answer only by a bit of deconstruction:
- The `Enums` passage is a simplified explanation of how the language works, meant to be factual as to explain the recommendation;
- Just like the entire document isn't a mandate, but a collection of suggestions, ideas and examples as opposed to restrictions, thus if anything should be followed, I'd like it to be the very first one: `use your own best judgement`;
- I define feature as only what one notices and not every functionality is a feature, but only those different persons notice or are significant to them. Writing this point because of the examples that follow;
- Not every functionality of a language must be used, like:
  - `namespace` is one of TypeScript, yet not a single use in the codebase - we just don't notice it;
  -  `with` is one of JavaScript, we all stay clear of it - our lives aren't worse for it;
- I find "don't be afraid to use it" a misleading argument since I don't see a reason to input emotions like "fear" into a place there aren't.
 
We can simply use `enum` or not, and as long as we know the pros and cons of it, I think it's good enough.

PR #2842, opened 2022-02-16
Fix table default references

colinhowe:
This fixes a bug where the `uuid` function was being used for setting the default references in *every single client*. Oops.


src/components/molecules/TableHeader/components/EditTableTitleModal/EditTableTitleModal.tsx
```diff
@@ -59,6 +59,7 @@ export const EditTableTitleModal: React.FC<EditTableTitleModalProps> = ({
   // use useAsyncFn for easier error handling, instead of state hook
   const [{ error: httpError, loading: isUpdating }, updateTables] = useAsyncFn(
     async (values: EditTableForm) => {
+      console.log("smash");
```
mike-lvov:
```suggestion
```

src/utils/table.ts
```diff
@@ -4,17 +4,21 @@ import { DEFAULT_TABLE_COLUMNS, DEFAULT_TABLE_ROWS } from "settings";
...
+  generateTableReference?: (title: string) => string;
 }) => Table = ({
   tableNumber,
   columns = DEFAULT_TABLE_COLUMNS,
   rows = DEFAULT_TABLE_ROWS,
+  generateTableReference = generateUniqueTableReference,
 }) => {
   const title = `Table ${tableNumber}`;
-  const reference = `${title}-${uuid()}`;
+  const reference = generateTableReference(title);
```
mike-lvov, line 21: It still is going to create unique id for every user, or am I missing something?

colinhowe: The change in `src/settings/index.ts` passes through a dumb function that makes the reference the same as the table title. **Only** for the default tables.

mike-lvov: Ah, I see. Yeah, makes sense

mike-lvov (approved):

PR #2844, opened 2022-02-17
Redesign create space screen

denisdimitrov:
Draft PR for the create space screen redesign

Partially fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1695


src/components/molecules/AdminInput/AdminInput.module.scss
```diff
@@ -0,0 +1,4 @@
+.input {
+  padding: 0.5rem 0.75rem;
```
goran-peoski-work, line 2: Maybe we'll need to copy the old `font-size--root` and `font-size-parent` from the old `constants.scss` file or write similar ones so we can add some meaning to these 0.5 and 0.75...

But it should be for attendee side. Considering this is admin side, shouldn't Tailwind solve it in the `className` instead? 

I'd suggest we create an `Input` component in `./src/components/admin` that follows that style system.

src/components/molecules/AdminInput/AdminInput.tsx
```diff
@@ -5,7 +5,7 @@ import { get } from "lodash";
 
 import { generateId } from "utils/string";
 
-import "./AdminInput.scss";
+import CN from "./AdminInput.module.scss";
```
goran-peoski-work, line 8: Probably should remove the SCSS module and create `AdminInput.tailwind.ts` file, possibly as a new component in a new place.

src/components/molecules/AdminInput/AdminInput.tsx
```diff
@@ -41,7 +41,11 @@ export const AdminInput: React.FC<AdminInputProps> = ({
     "AdminInput--visible": !hidden,
   });
 
-  const hiddenClasses = classNames(parentClasses, "AdminInput__input");
+  const inputClasses = classNames(
+    "w-full shadow-sm focus:ring-indigo-500 focus:border-indigo-500 block w-full sm:text-sm border-gray-300 rounded-md",
```
goran-peoski-work, line 45: Make the sausage as `TW.input` and just stick directly into `<input className={TW.input}`

In case of variants or other flag props, should probably shorten `classNames` as `cn` and just use `className={cn(TW.input, TW[variant])}`. Or put into variable, whichever allows for a more readable code.

src/components/molecules/AdminSection/AdminSection.tsx
```diff
@@ -28,7 +28,7 @@ export const AdminSection: React.FC<AdminSectionProps> = ({
   );
 
   return (
-    <section className="AdminSection">
+    <section className="mb-6 mt-6 flow-root">
```
goran-peoski-work, line 31: should probably do a copy-paste these atom/molecule components into `./src/components/admin/` and give them the full TW treatement while removing any `.module.scss` file

goran-peoski-work (changes requested):
Reviewed only 4 files. Some suggestions there can be extrapolated/implemented for the rest of the components before another more full review.

src/components/admin/Header/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+          fill="currentColor"
+          viewBox="0 0 20 20"
+          aria-hidden="true"
+        >
+          <path d="M5.555 17.776l8-16 .894.448-8 16-.894-.448z" />
+        </svg>
+      </div>
+      <div className={TW.title}>Create space</div>
+      <div className={TW.subtitle}>
+        {selectedItem?.text ?? "Select a template"}
```
denisdimitrov, line 48: [Note] Should we export `Select a template` to constants or similar?

src/components/molecules/AdminSection/AdminSection.tsx
denisdimitrov: All `AdminSection` components have been reverted to their original state and they have been extracted into new ones located at `/admin/...`

src/scss/admin/constants.scss
```diff
@@ -0,0 +1,2 @@
+$inputSpacingX: 0.75rem;
+$inputSpacingY: 0.5rem;
```
goran-peoski-work, line 2: [code smell] Might be perfectly OK, might be an indication there is some alternative solution.

Usually .tailwind.scss or .module.scss would compartmentalize these constants if they're specific for a single compoenent.

Alas, if they're shared, (and of course `$kebob-naming--scheme`) they could be put in a specific file as to not bloat `constants.scss`. 

I'm thinking a file like https://github.com/sparkletown/sparkle/blob/rnd/attendee-styles/src/scss/attendee/spacing.scss which is part of a re-organization PR https://github.com/sparkletown/sparkle/pull/2846/files

src/scss/colors.scss
```diff
@@ -0,0 +1 @@
+$white: #ffffff;
```
goran-peoski-work, line 1: There already is one (surprisingly empty) `colors.scss` file, but I think it should be moved under `./src/scss/admin/colors.scss`

goran-peoski-work (approved):
looks good. approving.

BUT

would be nice if some `@debt` and `@see` comments are added and maybe the SCSS constants moved

src/components/organisms/SpaceCreateForm/SpaceCreateForm.tailwind.ts
```diff
@@ -0,0 +1,5 @@
+export const saveButton =
```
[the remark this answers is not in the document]
denisdimitrov: I suppose you are right, I'll move it

src/scss/colors.scss
denisdimitrov: I didn't move the file on purpose because I want this one to be the shared one, between attendee and admin. There is no point in adding `$white: #FFFFFF` in both admin and attendee `colors.scss` files

PR #2849, opened 2022-02-18
Enable new users to proceed to their space

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1726

Fixed reading `userId` from correct place so the visited space and world would get saved thus not dropping the new users out of the space and into the entrance steps.


mike-lvov (approved):
LGTM

PR #2850, opened 2022-02-18
Fix bugs on the Spaces page

sunny-viktoryia:
Use the correct create-space URL
Fix width of the Edit buttons

### DIFF
--- a/src/components/admin/SpaceCard.tsx
+++ b/src/components/admin/SpaceCard.tsx
@@ -132,7 +132,11 @@ export const SpaceCard: React.FC<SpaceCardProps> = ({
           </p>
         </div>
         <div className="flex justify-center align-center mt-4">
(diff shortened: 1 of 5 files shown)


src/components/admin/SpaceCard.tsx
```diff
@@ -132,7 +132,11 @@ export const SpaceCard: React.FC<SpaceCardProps> = ({
           </p>
         </div>
         <div className="flex justify-center align-center mt-4">
-          <LinkButton href={editSpaceUrl} disabled={!isEditable} width="6/12">
+          <LinkButton
+            href={editSpaceUrl}
+            disabled={!isEditable}
+            width="halfContainer"
```
goran-peoski-work, line 138: [NOTE] Few musings:
Is this an exception? Would normally buttons be full width and only rarely half width? 

On that note, the strings can be `full` and `half`. Easy to deal with the `.tailwind.ts` names

PR #2854, opened 2022-02-21
New video API ported to master

colinhowe:
This ports the new video API to master and re-introduces reactions which were turned off in the new video API


src/components/attendee/VideoComms/VideoCommsProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const contextState: VideoCommsContextType = {
+    ...commsStatus,
+    joinChannel: twilio.joinChannel,
+    disconnect: twilio.disconnect,
+    shareScreen: twilio.shareScreen,
+    startAudio: twilio.startAudio,
+    stopAudio: twilio.stopAudio,
+    startVideo: twilio.startVideo,
+    stopVideo: twilio.stopVideo,
+  };
```
goran-peoski-work, line 70: [NOTE]
`useMemo` to not recreate the object w/o any `commsStatus` change although with limited use of the context (high up in the component tree) it should not make any difference

src/components/attendee/VideoComms/hooks.tsx
```diff
@@ -0,0 +1,9 @@
+import { useContext } from "react";
+
+import { VideoCommsContext } from "./VideoCommsProvider";
+
+export const useVideoComms = () => {
+  // This might seem a little pointless. However, we expect that in the future
+  // this hook might do a little bit more than just get the context.
```
goran-peoski-work, line 7: [NOTE]
Anything that can be added to this hook will in all probability work added to the context.
The only thing this hook would be useful for is local state for a component.

src/components/attendee/VideoComms/internal/scss/TwilioTrackDisplay.module.scss
```diff
@@ -0,0 +1,3 @@
+.mirrored {
+  transform: scale(-1, 1);
```
goran-peoski-work, line 2: [NOTE]
There should be a way to do the mirroring effect w/o flipping the DOM video element controls as well. Maybe through `filter` instead of `transform`, but I think there should be a flip on the video itself, before the stream is picked up by the DOM element.

src/components/attendee/VideoComms/internal/useMute.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback, useState } from "react";
+
+export const useMute = () => {
```
goran-peoski-work, line 3: [NOTE] `useMute` is basically `useShowHide`. Both just have too specific names for something that generically be named `useToggle`.

src/hooks/twilio/useVideoRoomState.tsx
```diff
@@ -87,43 +83,30 @@ export const useVideoRoomState = (
...
 
         if (message.toLowerCase().includes("unknown")) {
           setVideoError(
             `${message}; common remedies include closing any other programs using your camera, and giving your browser permission to access the camera.`
           );
         } else setVideoError(message);
-      });
-  }, [userId, roomName, dismissVideoError, isActiveParticipant]);
+      } else {
+        throw error;
```
goran-peoski-work, line 103: Hook shouldn't be throwing an error but returning it.

goran-peoski-work (changes requested):
Most of the code looks the same as already merged to the N16 branch.

`src/hooks/twilio/useVideoRoomState.tsx` throws an error instead of returning it. This is counter to how hooks are expected to work and counter to how it used to work.

Other than that' I'd approve this PR after a little local run and click tests, make sure nothing major is broken.

PR #2870, opened 2022-02-24
Add new live hooks to avoid reactfire bugs

goran-peoski-work:
Add few new hooks to replace old buggy ones.

These will be needed for checking the login/logout state as well as access to certain worlds and spaces.

Additionally there are few login and permission restriction components to use later. There is  a followup PR with them
- https://github.com/sparkletown/sparkle/pull/2872


src/components/admin/AdminRestrictedMessage.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { generateUrl } from "utils/url";
+
+import { useSpaceParams } from "hooks/spaces/useSpaceParams";
+import { useUserId } from "hooks/user/useUserId";
+
+import { ButtonNG } from "components/atoms/ButtonNG";
+import { SparkleLogo } from "components/atoms/SparkleLogo";
+
+import SHAPE_DENIED from "assets/images/access-forbidden.svg";
```
colinhowe, line 20: Might make sense to move this over to the ban icon from font awesome? https://fontawesome.com/icons/ban?s=solid

src/hooks/user/usePermission.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+type UsePermissionOptions = {
+  userId: UserId;
+  worldId?: WorldId;
+  spaceId?: SpaceId;
+};
+
+export const usePermission = (options: UsePermissionOptions) => {
+  const path = useMemo(() => [COLLECTION_ROLES, "admin"], []);
```
colinhowe, line 23: Could this just be a boring constant outside of the component?

src/hooks/user/usePermission.ts
```diff
@@ -0,0 +1,95 @@
...
+    const ids = !roleLoading && users ? role?.users : ALWAYS_EMPTY_ARRAY;
+    const isSuperAdmin = userId && ids.includes(userId);
+
+    const isLoading = worldLoading || spacesLoading || roleLoading;
+    const isLoaded = !isLoading;
+
+    const isAdmin = isSuperAdmin || isAnyWorldOwner || isAnySpaceOwner;
+    const isNotAdmin = !isAdmin;
+
+    const error = roleError ?? worldsError ?? spacesError;
```
colinhowe, line 64: I worry about the performance impact of this function. In particular, if we call this twice with the same arguments from two different places do we end up doing the same queries against the database multiple times? Do we end up with multiple subscriptions for update to those bits of data?

Can you do some checking to see exactly what happens in terms of queries hitting the database? It might make sense to move some of these checks into a top level permissions context provider that does most of the work once and once only.

src/hooks/fire/useLiveCollection.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  const memoizedPath = useMemo(
+    () => (path ?? ALWAYS_EMPTY_ARRAY).filter(isValidSegment),
+    [path]
+  );
+
+  const memoizedConstraints = useMemo(
+    () => (constraints ?? ALWAYS_EMPTY_ARRAY).filter(isValidConstraint),
+    [constraints]
+  );
```
colinhowe, line 55: I think quietly allowing for bad queries by filtering out broken parts is going to leave to hard to find bugs. If I write a query that is wrong I expect to have a very noisy error as soon as I try it.

src/hooks/fire/useLiveCollection.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const isValidSegment = (segment: string | undefined): segment is string =>
+  "string" === typeof segment && "" !== segment;
+
+const isValidConstraint = (
+  constraint: QueryConstraint | null | undefined
+): constraint is QueryConstraint =>
+  constraint !== null && constraint !== undefined;
+
+export const useLiveCollection = <T extends object>(
```
colinhowe, line 37: I think this could do with a comment saying why you'd use this one over all the other alternatives that we have in the codebase

src/hooks/fire/useLiveCollection.ts
colinhowe: Similarly for the other new hooks.

src/hooks/fire/useLiveCollection.ts
colinhowe: Similarly for the other new hooks.

src/hooks/worlds/useWorldsByNotHidden.ts
```diff
@@ -0,0 +1,33 @@
...
+    []
+  );
+
+  const { error, isLoading, isLoaded, data } = useLiveQuery(memoizedQuery);
+
+  return useMemo(
+    () => ({ error, isLoading, isLoaded, worlds: data ?? ALWAYS_EMPTY_ARRAY }),
+    [error, isLoading, isLoaded, data]
+  );
+};
```
colinhowe, line 33: I don't think this should exist at all. Hidden worlds should be invisible from **all** queries.

colinhowe (changes requested):
Some performance concerns that need checking

src/hooks/user/usePermission.ts
goran-peoski-work: > Do we end up with multiple subscriptions for update to those bits of data?
Yes. There is no project wide caching involved. For that I can suggest a follow up PR that adds a context that uses this hook, maybe even straight as part of those `LoginRestricted` and `WithPermission` components.

> Can you do some checking to see exactly what happens in terms of queries hitting the database? 

Short answer: can if you provide with a baseline. It's the same SDK be it used through one wrapper or another. the `useLive*` hooks are just more simplistic in design.

The long answer: 
it's not like a "number of queries hitting the database", 
but a regular protobuf channel that works the same for the API be it react-redux-connect or these `onSnapshot` listeners.

The "database driver" i.e. Firestore SDK has its own `*FromCache` and `*FromServer` versions that it decides which to use when a bare call like `getDoc` is made and the communication (you can notice in the browser's network tab) is made up of packaged deltas for the various connections and queries it has.

The listener itself already employs local caching and options to read only deltas or metadata for local changes that still are to be persisted to server etc.
https://firebase.google.com/docs/firestore/query-data/listen

Here is an excerpt from that documentation page:
>The initial state can come from the server directly, or from a local cache. If there is state available in a local cache, the query snapshot will be initially populated with the cached data, then updated with the server's data when the client has caught up with the server's state.

src/hooks/fire/useLiveCollection.ts
goran-peoski-work: >  If I write a query that is wrong I expect to have a very noisy error as soon as I try it.

And if I write a query that has missing `spaceSlug` because the previous hook is still loading I expect some noisy error to not break the entire page.

These expectations, however, aren't objective, but subjective.

What objectively can be done is to define what is "a query that is wrong" in the first place. 

Then, following from there, add proper tools, like switching the moniker from `undefined`, to some `Symbol()` or some other method of dealing with expected vs unexpected definition of a query, like populate the `error` field that every result has with a `SparkleError` type etc.

src/hooks/fire/useLiveCollection.ts
goran-peoski-work: The deprecated ones have a comment.

src/hooks/worlds/useWorldsByNotHidden.ts
goran-peoski-work: So far `  .where("isHidden", "==", false)` can be found in half of dozen lines. It is one that already exists and  `useWorldsByNotHidden` just copied the query:

```ts
export const useOwnWorlds: UseOwnWorlds = ({ userId }) => {
  const firestore = useFirestore();

  const ownWorldsRef = query(
    collection(firestore, COLLECTION_WORLDS),
    where("isHidden", "==", false),
    where("owners", "array-contains", userId)
  ).withConverter(withIdConverter<WorldWithId, WorldId>());
```

Is there a way of reading worlds from DB without putting `isHidden` constraint but not having hidden worlds in the result?

src/components/admin/AdminRestrictedMessage.tsx
goran-peoski-work: 🤷 They don't look like anything really. We'll have to ask for new design first, also for the few missing like "login with no world slug" and "login with only world but no space slug" and probably a new landing page for attendee side spaces etc.

src/hooks/user/usePermission.ts
goran-peoski-work: I think I have figured out a way and place for these.

src/hooks/fire/useLiveCollection.ts
colinhowe: If we're still waiting for a previous hook then we should carry on waiting rather than doing queries with crap arguments

src/hooks/fire/useLiveCollection.ts
colinhowe: Perfect

src/hooks/worlds/useWorldsByNotHidden.ts
colinhowe: No :(

goran-peoski-work (commented):
The `useLive*` hooks now with the possibility of outputting `error` if the query is wrong and the possibility to pause for a bit before trying to connect to DB data through the value of `DEFERRED`.

src/hooks/fire/useLiveCollection.ts
goran-peoski-work: Well, it's not like waiting, it's more like "bad connection". 

From 945342eb8af0a1591a833f0c785c70b1c61b60aa there is now a value named `DEFERRED` to signal the hook should wait for data before trying to connect. The simplest use would be 
```tsx
useLiveDocument(DEFERRED);
```
but usually some `if-else` check that either produces the correct path or puts DEFERRED in one of several possible places.

src/hooks/user/usePermission.ts
goran-peoski-work: 945342eb8af0a1591a833f0c785c70b1c61b60aa `PATH.rolesAdmin` from `apiSettings.ts` More paths can simply be added to `PATH` and will be constant array values.

colinhowe (approved):

PR #2883, opened 2022-02-28
Minor layout fixes

colinhowe:
Closes:
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1743
 - https://github.com/sparkletown/internal-sparkle-issues/issues/1742


goran-peoski-work (approved):
Would have been nice to have before/after shots since I'm approving only by looking at the code

PR #2884, opened 2022-03-01
Introduce safe zone to party maps

colinhowe:
image

Properly resize the party map so that the "safe" zone (as defined in the database) is always visible. Plus mobile responsive behaviour.


src/hooks/useCheckImage.ts
```diff
@@ -7,23 +7,29 @@ export const useCheckImage = (
   isValid: boolean;
   isLoading: boolean;
   error: string | undefined;
```
goran-peoski-work, line 9: [OOS] Looks like these can be replaced with adding `): LoadStatus & {` above as the result from this hook. Makes the output a bit more uniformed/standardized.

src/components/templates/PartyMap/components/RoomAttendance/RoomAttendance.module.scss
```diff
@@ -0,0 +1,7 @@
+@use "scss/attendee/space";
+
+.RoomAttendance {
```
goran-peoski-work, line 3: More visible it is no longer the old BEM.

```suggestion
.roomAttendance {
```

src/components/templates/PartyMap/components/RoomAttendance/RoomAttendance.tsx
```diff
@@ -1,23 +1,19 @@
-import React, { useMemo } from "react";
-
-import { ALWAYS_EMPTY_ARRAY, DEFAULT_ROOM_ATTENDANCE_LIMIT } from "settings";
+import React from "react";
```
goran-peoski-work, line 1: Empty undeleted file `src/components/templates/PartyMap/components/RoomAttendance/RoomAttendance.scss`

src/components/templates/PartyMap/components/Map/MapRoom.tsx
```diff
@@ -56,48 +65,68 @@ export const MapRoom: React.FC<MapRoomProps> = ({ venue, room }) => {
...
+                <p>TODO</p>
+                <span
+                  className={styles.PortalInfoButton}
+                  onClick={selectRoomWithSound}
+                >
+                  Enter
+                </span>
+              </div>,
+              popoverContainerElement
+            )}
```
goran-peoski-work, line 127: Maybe this `ReacDOM.createPortal()` call can be extracted in own popover component, have some nice comment about the need for it etc. In this place it would look like
```tsx
<Popover>
  <div> ... </div>
</Popover>
```

src/components/templates/PartyMap/components/Map/MapRoom.module.scss
```diff
@@ -6,14 +6,24 @@
...
 
 .MapRoom {
-  display: flex;
-  flex-direction: column;
-  overflow: hidden;
-  max-width: layout.$w-sidebar;
+  @media (max-width: calc(breakpoint.$md - 1px)) {
+    // Override inline styles
+    height: auto !important;
+    width: auto !important;
```
goran-peoski-work, line 17: Why the need to override? `@debt` comment would be nice if there is possible removal of `!important` in the future, but would be good to at least explain why, not what.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+    left: (browserWidth - desiredSafezoneWidth) / 2,
+    top: (browserHeight - desiredSafezoneHeight) / 2,
+    width: desiredSafezoneWidth,
+    height: desiredSafezoneHeight,
+  };
+
   const mapStyles = {
     backgroundImage: `url(${mapBackground})`,
+    backgroundSize: `${desiredWidth}px ${desiredHeight}px`,
   };
```
goran-peoski-work, line 161: I recommend `useCss()` if possible, there should be other examples through the code using background through it.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+    desiredSafezoneHeight = _scaledHeight;
+  } else {
+    desiredSafezoneWidth = _scaledWidth;
+    desiredSafezoneHeight = browserHeight - verticalPad;
+  }
+
+  const desiredWidth = desiredSafezoneWidth / safeWidthPerc;
+  const desiredHeight = desiredSafezoneHeight / safeHeightPerc;
+
+  const safeZoneBounds = {
```
goran-peoski-work, line 151: Maybe something like 
```suggestion
  const safeZone: Zone = {
```

Not a deal breaker, just thought would look nicer.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+    // TODO Tidy thisup
+    console.error("Failed to get image width/height");
+    return <>Error</>;
+  }
+
+  const { width: safeWidthPerc, height: safeHeightPerc } = venue.config
+    ?.safeZone || {
+    width: 1.0,
+    height: 1.0,
+  };
```
goran-peoski-work, line 121: this object could be put in some settings file like:

```ts
const DEFAULT_SAFE_ZONE: Zone = { width: 1, height: 1 };
Object.freeze(DEFAULT_SAFE_ZONE);
```

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+
+  // TODO Account for height of top and bottom nav
+
+  const verticalPad = 140; // 70px for top and bottom
```
goran-peoski-work, line 125: Same with this constant, along with `DEFAULT_SAFE_ZONE`.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+  // a basis
+  const safeZoneWidth = safeWidthPerc * imageWidth;
+  const safeZoneHeight = safeHeightPerc * imageHeight;
+
+  const _scaledWidth =
+    safeZoneWidth * ((browserHeight - verticalPad) / safeZoneHeight);
+  const _scaledHeight = safeZoneHeight * (browserWidth / safeZoneWidth);
+
+  let desiredSafezoneWidth = 1;
+  let desiredSafezoneHeight = 1;
```
goran-peoski-work, line 137: [NOTE] No need of change, just for amusement, a take on: 

`let` and single `if` 
VS 
`const` and two `= ? : `

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
+    height: 1.0,
+  };
+
+  // TODO Account for height of top and bottom nav
+
+  const verticalPad = 140; // 70px for top and bottom
+
+  // Figure out if the image should be rescaled using the width or height as
+  // a basis
+  const safeZoneWidth = safeWidthPerc * imageWidth;
```
goran-peoski-work, line 129: [NOTE] Generally the naming here can be made more self-documenting. Some ideas:
- `scale` or `factor` for `perc`
- prepending or appending `old`, `new` or `candidate` to the names instead of underscore
- `zone`, instead of `safeZone`, to keep them shorter since there isn't other type of zone

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -1,36 +1,31 @@
...
   selectPortal: (room: Room) => void;
+  safeZoneBounds: Dimensions & Position;
 }
 
-const Portals: React.FC<PortalsProps> = ({ space, selectPortal, portals }) => {
+const Portals: React.FC<PortalsProps> = ({
+  space,
+  selectPortal,
+  portals,
+  safeZoneBounds,
```
goran-peoski-work, line 27: [NOTE] There might be some other way to avoid passing the same prop three components deep, but then, probably it's clear enough by their names and location that these components are meant to only work together

goran-peoski-work (changes requested):
Looks good. 

Some suggestions and questions, wouldn't bother to touch on all, but it might be good to extract that `ReactDOM.createPortal` in a separate component and maybe try `useCss` and `ResizeObserver` if they can help.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
-  const [mapBackground] = useValidImage(
-    venue?.mapBackgroundImageUrl,
-    DEFAULT_MAP_BACKGROUND
-  );
+  const [
+    mapBackground,
+    { width: imageWidth, height: imageHeight },
+  ] = useValidImage(venue?.mapBackgroundImageUrl, DEFAULT_MAP_BACKGROUND);
+
+  const { width: _browserWidth, height: browserHeight } = useWindowSize();
```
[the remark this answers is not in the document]
colinhowe: `clientWidth` takes into account scrollbars. `useWindowSize` does not.

src/components/templates/PartyMap/components/Map/Map.tsx
```diff
@@ -56,28 +52,122 @@ interface MapProps {
...
     return <>Loading map...</>;
   }
 
+  if (!imageWidth || !imageHeight) {
+    // TODO Tidy thisup
+    console.error("Failed to get image width/height");
+    return <>Error</>;
+  }
+
+  const { width: safeWidthPerc, height: safeHeightPerc } = venue.config
```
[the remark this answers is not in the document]
colinhowe: Both percentage and scale could be used. I don't think there's a right answer here.

It's currently hackintheback. We'll have to add a UI for it when we create the party map admin UI

src/components/templates/PartyMap/components/Map/MapRoom.tsx
colinhowe: Potentially. I've created a placeholder issue to track this: https://github.com/sparkletown/internal-sparkle-issues/issues/1748

src/scss/attendee/theme.scss
```diff
@@ -39,6 +40,7 @@ $sparkle--border: var(--sparkle--border); // previously named $sparkle-border
 
 $c-panel-background: var(--panel-background);
 $c-glass-panel-background: var(--glass-panel-background);
+$c-color-glass-panel-background: var(--color-glass-panel-background);
```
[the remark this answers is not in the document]
colinhowe: This is where our naming falls down. This is for a <<coloured glass>> panel. Not just a plain glass panel.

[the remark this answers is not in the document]
goran-peoski-work: LOL

goran-peoski-work (approved):
LGTM

PR #2902, opened 2022-03-07
tweaks for world schedule and it's components

denisdimitrov:
Minor fixes and tweaks for the world schedule and the dropdown color. Also wraps the page with `<AdminRestricted/>` to prevent normal users from accessing the page.


goran-peoski-work (approved):
LGTM

PR #2903, opened 2022-03-08
Add informative error in `testWorldBySlugExists()` validation

goran-peoski-work:
I don't know the correct test that should be performed ATM, but this addition will at least not put the blame at Firebase due to hard to understand error.


sunny-viktoryia (approved):
Looks good 🙃

PR #2907, opened 2022-03-09
Adds new external room

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1712
designs at: https://www.figma.com/file/DltmqgBfdkgyX7z5Uv4r15/HiFi%3A-Visual-Redesign---Shared-w-Eng?node-id=1111%3A3431

image

image

### DIFF
--- a/src/components/templates/ExternalRoom/ExternalRoom.tsx
+++ b/src/components/templates/ExternalRoom/ExternalRoom.tsx
@@ -1,22 +1,11 @@
-import React, { useCallback, useEffect } from "react";
-import classNames from "classnames";
-
-import { ALWAYS_EMPTY_ARRAY, ENABLE_POPUPS_URL } from "settings";
+import React from "react";
 
 import { AnyVenue } from "types/venues";
 
 import { WithId } from "utils/id";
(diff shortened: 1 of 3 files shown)


src/components/templates/ExternalRoom/ExternalRoom.tsx
```diff
@@ -25,80 +14,48 @@ export interface ExternalRoomProps {
...
+      style={{
+        background: `url(${venue.config?.landingPageConfig.coverImageUrl}) center center`,
+      }}
+    >
+      <img
+        src={venue.host?.icon}
+        alt="Venue icon"
+        className={styles.VenueIcon}
+      />
+      <div className={styles.InfoContainer}>
```
goran-peoski-work, line 35: These would be nicer if they're `camelCase`, not `PascalCase` as is with the other class names. 

The class names mapping is another matter, since what is being import from `"./ExternalRoom.module.scss"` is a `Record<string,string>` not `CSSStyleDeclaration` so the name `styles` is misleading.

goran-peoski-work (approved):
LGTM

Added a comment about naming though.

PR #2908, opened 2022-03-09
fix table avatars appearing outside of the container

denisdimitrov:
Adds flex-wrap to fix the table avatars in the TableComponent for both jazzbar and conversation space.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1741

### P.S. The join button is in a wrong position because I manually added the rest of the html avatars after it's definition.
new-table

### DIFF
--- a/src/components/attendee/TableComponent/TableComponent.module.scss
+++ b/src/components/attendee/TableComponent/TableComponent.module.scss
@@ -42,6 +42,7 @@
 
 .tableUsers {
   display: flex;
+  flex-wrap: wrap;
   flex-direction: row;
   justify-content: center;
   align-items: center;


goran-peoski-work (approved):
LGTM

Anything on the square vs circle avatars?

denisdimitrov:
@goran-peoski-work Yeah, they look weird but I don't see an issue anywhere :/

genmon:
The avatars are supposed to be the same shape as the "join" button (there's an "avatar" section on the Figma for components too) -- drop a note in the redesign channel on Slack if you'd like to chat about it

goran-peoski-work:
@genmon @denisdimitrov I don't see it as a big, just thought to ask. If/when it is necessary to do the adjustment, we'll do it with an issue/PR as per usual

PR #2913, opened 2022-03-10
Edit the party map

colinhowe:
Getting this merged early so that the basic outline can be used for implementing the edit functionality for the other space templates.

This provides the main functionality for editing positions of portals etc.

image


src/pages/Account/Venue/VenueMapEditor/Container.tsx
```diff
@@ -0,0 +1,242 @@
...
+  useMemo(() => {
+    if (!imageDims || !imageDims.width || !imageDims.height) return;
+
+    const copy = iconsMapPerc.map(
+      (mapIcon) =>
+        ({
+          ...mapIcon,
+          width: imageDims.width * mapIcon.width * safeZonePerc.width,
+          height: imageDims.height * mapIcon.height * safeZonePerc.height,
+          top:
```
goran-peoski-work, line 148: I think there should be an explanation comment line or two here and above about the need to multiply or divide something by a factor. Also, since this is a function inside `.map()` it might be easy to extract it outside of this component

src/pages/Account/Venue/VenueMapEditor/Container.tsx
```diff
@@ -0,0 +1,242 @@
...
+      <div ref={drop}>
+        <div ref={ref} className={styles.mapContainer}>
+          <ReactResizeDetector handleWidth handleHeight>
+            {({ targetRef }) => <span ref={targetRef} />}
+          </ReactResizeDetector>
+          <img
+            alt="draggable background"
+            style={{
+              width: "100%",
+            }}
```
goran-peoski-work, line 216: Inline style. 

[Q] Would a class be more useful here?
That being asked, it is most likely OOS to determine if this container is admin or attendee side and deal accordingly

src/pages/Account/Venue/VenueMapEditor/Container.tsx
```diff
@@ -0,0 +1,242 @@
...
+              onDragStart={setDragItemId}
+              isSaving={isSaving}
+            />
+          ))}
+        </div>
+      </div>
+      {imageDims && interactive && (
+        <CustomDragLayer
+          snapToGrid={!!snapToGrid}
+          rounded={!!rounded}
```
goran-peoski-work, line 236: Not sure where these come from, but seeing it `!!rounded` used several times, it might be more clear to have something like

```tsx
const isRounded: boolean = !!rounded;
```

before the `return` keyword.

src/pages/Account/Venue/VenueMapEditor/DraggableSubvenue.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  background-size: 100% 100%;
+  background-repeat: no-repeat;
+}
+
+.resizeTab {
+  position: absolute;
+  width: 10px;
+  height: 10px;
+  background-color: white;
+  border: 1px solid var(--content--over);
```
goran-peoski-work, line 15: These styles look like just copied over from before. The `var(--content--over)` will most likely not be there and make border not show up. Similar comment for the rest of the hard-coded values.

src/pages/Account/Venue/VenueMapEditor/DraggableSubvenue.tsx
```diff
@@ -0,0 +1,120 @@
...
+  top: number,
+  isDragging: boolean
+): React.CSSProperties => ({
+  position: "absolute",
+  top,
+  left,
+  opacity: isDragging ? 0 : 1,
+  height: isDragging ? 0 : "",
+  background: "rgba(147, 124, 99, 0.2)",
+});
```
goran-peoski-work, line 24: This function is problematic, not only that there alternatives to it in `useCss()` and similar hooks, but also because its value is used directly for `style={}` which will probably trigger re-render.

That being said, I'm guessing it's not a big issue since dragging is limited to rare occasions of edit in Admin and it may probably provide for a desired side-effect to have components re-render on drag.

src/pages/Account/Venue/VenueMapEditor/DraggableSubvenue.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+export type PropsType = SubVenueIcon & {
+  imageStyle?: CSSProperties;
```
goran-peoski-work, line 27: [NOTE] This type is basically what something named `styles` would be since `import styles from "./some.module.scss"` doesn't import `CSSProperties` but `Record<string,string>` of the mapped class names.

src/pages/Account/Venue/VenueMapEditor/interfaces.ts
```diff
@@ -1,5 +1,5 @@
 export interface DragItem {
-  id: string;
+  itemId: number;
```
goran-peoski-work, line 2: If this is saved to DB, would it be somehow converted on read or write to `string`?

goran-peoski-work (approved):
Looks good. 

Approving but also think some comment suggestions should be accepted and few other comments added before merge.

The rest of the notes and questions, leaving up to your discretion.

PR #2905, opened 2022-03-08
Navigation menu for mobile + space info tab

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1707

image

PR #2916, opened 2022-03-11
add world users page to admin

denisdimitrov:
Add world users page to the admin with the ability to delete admins.

Partially fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1703


functions/src/world.ts
```diff
@@ -146,3 +146,76 @@ export const deleteWorld = functions.https.onCall(async (data, context) => {
...
+    .doc(worldId)
+    .get();
+
+  const worldData = worldRef.data();
+
+  if (!worldData) {
+    return;
+  }
+
+  const owners = worldData.owners as string[];
```
goran-peoski-work, line 173: [NOTE] might be OOS:

I think it would be more helpful to have some type guard here that uses `Array.isArray()` so we're not just blindly assuming it's an array. Might be useful to just use `Array.isArray()` without spending time on creating type guards like 

```ts
if( Array.isArray(owners) && owners.includes(userId) )
```

The `return` that follows, however, to avoid silent errors, might be useful to check if the simplest log like `console.error()` (or `warn` or `log`) is good enough to have some info of the problem in the Firestore functions console.

goran-peoski-work: The above note is valid for other similar instances, if deemed OOS, we should think of this kind of improvement for the future

src/api/world.ts
```diff
@@ -258,6 +258,19 @@ export const updateWorldScheduleSettings = async (
   )(await createFirestoreWorldScheduleInput(world));
 };
 
+export const addWorldAdmin = async (worldId: WorldId, userId: UserId) =>
+  await httpsCallable(
+    FIREBASE.functions,
+    "world-addWorldAdmin"
```
goran-peoski-work, line 264: [NOTE] [OOS]
I think we can put these strings on a frozen object for reference and maintainability.
A single utility file with something like https://github.com/sparkletown/sparkle/pull/2914/files#diff-2e14c70c6dbf9de86687e6d3b6a66d3f8492cccd3bf6d450d5324ef2a7d2687eR28-R42

src/components/admin/Button/Button.tailwind.ts
```diff
@@ -6,6 +6,8 @@ export const secondary =
   "border-gray-300 bg-white text-gray-700 hover:bg-gray-50 focus:ring-indigo-500";
 export const danger =
   "border-transparent text-white bg-warning-red focus:ring-indigo-500";
+export const secondaryBorderless =
+  "border-none shadow-none bg-white text-gray-700 hover:bg-gray-50 focus:ring-indigo-500";
```
goran-peoski-work, line 10: We're not constrained to have a single `variant` prop, we can also have a `border` one so we can do combinations if that makes sense. We might have similar issues with this button as the one for attendee.

@denisdimitrov Please check this comment and how it was resolved: https://github.com/sparkletown/sparkle/pull/2905/files#r823519370

And here it is how it's being used https://github.com/sparkletown/sparkle/pull/2905#discussion_r823546585

src/components/admin/Button/Button.tsx
```diff
@@ -5,7 +5,11 @@ import { Loading } from "components/molecules/Loading";
 
 import * as TW from "./Button.tailwind";
 
-export type ButtonVariant = "primary" | "secondary" | "danger";
+export type ButtonVariant =
+  | "primary"
+  | "secondary"
+  | "secondaryBorderless"
```
goran-peoski-work, line 11: This might be better if moved as `| borderless` for `ButtonBorders` as discussed above in the `.tailwind.ts` file

src/components/admin/DeleteAdminModal/DeleteAdminModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { UserWithId } from "types/id";
+
+import { useWorldAndSpaceByParams } from "hooks/spaces/useWorldAndSpaceByParams";
+
+import { Modal } from "components/molecules/Modal";
+
+import { Button } from "../Button";
+
+export interface DeleteAdminModalProps {
```
goran-peoski-work, line 11: If not used outside this file

```suggestion
interface DeleteAdminModalProps {
```

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import { Button } from "../Button";
+import { Input } from "../Input";
+
+export interface InviteAdminModalProps {
+  show: boolean;
+  onHide: () => void;
+}
+
+//@debt This component is not finished yet, algoliaSearch needs to be used for search functionality.
```
goran-peoski-work, line 16: This comment might make more sense if put in another file with a constant `DISABLED_DUE_TO_` as described in https://github.com/sparkletown/sparkle/blob/redesign-2021/docs/code-style.md#maybe-could-be-useful-later-dead-code

This means there shouldn't be a comment like `// await addWorldAdmin(worldId, "userId placeholder")` as well, maybe just an empty function that has the proper parameters but has the above constant and/or comment inside.

Please check https://github.com/sparkletown/sparkle/blob/redesign-2021/src/settings/disableSettings.ts

src/components/admin/WorldUserCard/WorldUserCard.tsx
```diff
@@ -0,0 +1,107 @@
...
+          {/* TODO: Add edit modal when design is provided. */}
+          {/* <Button
+            variant="secondaryBorderless"
+            className="flex"
+            onClick={showEditAdminModal}
+          >
+            <FontAwesomeIcon icon={faPen} className="px-1" size="lg" />
+
+            <div>Edit</div>
+          </Button> */}
```
goran-peoski-work, line 80: Same dead code comment as above

src/hooks/user/useProfileByIds.ts
```diff
@@ -0,0 +1,63 @@
...
+type UseProfileByIds = (options: {
+  userIds: UserId[];
+}) => UserWithId[] | undefined;
+
+export const useProfileByIds: UseProfileByIds = ({ userIds }) => {
+  const firestore = useFirestore();
+
+  const usersRequests = useMemo(
+    () =>
+      chunk(userIds, FIRESTORE_QUERY_IN_ARRAY_MAX_ITEMS).map((usersChunk) =>
```
goran-peoski-work, line 28: 👍 

[NOTE] Few more uses like this and we can start thinking of a utility function that abstracts the boilerplate away

src/pages/Admin/WorldUsers/WorldUsers.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { useShowHide } from "hooks/useShowHide";
+import { useUser } from "hooks/useUser";
+
+import { LoadingPage } from "components/molecules/LoadingPage";
+
+import { AdminRestricted } from "components/atoms/AdminRestricted";
+
+import * as TW from "./WorldUsers.tailwind";
+
+export interface WorldScheduleFormInput {
```
goran-peoski-work, line 25: Not sure if this one is used outside, but if it not, same removal for it and all the rest so we can have shorter output of `npm run prune` (different from `npm prune`)

```suggestion
interface WorldScheduleFormInput {
```

src/pages/Admin/WorldUsers/WorldUsers.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const spacesNames = userSpace.spaces
+      .map((space) => space.name.toLowerCase())
+      .join();
+
+    return includes(spacesNames, spaceSearchQuery.toLowerCase());
+  });
+
+  if (isWorldLoading) {
+    return <LoadingPage />;
+  }
```
goran-peoski-work, line 91: [OOS] 

There is a solution to not having these checks in a component, especially if we're replacing those with `compose()` by adding a simpler pedestrian HOC. Example:

`WorldUsersHoc.tsx` would have the hook and this kind of code:
```tsx

  if (isWorldLoading) {
    return <LoadingPage />;
  }
  
  return <WorldUsers world={world} />
```

`index.ts` would have 
```ts
export { WorldUsersHoc as WorldUsers } from './WorldUsers';
```

goran-peoski-work (approved):
LGTM

I have added few notes and comments. I can suggest doing some of the updates if they're short and quick, wouldn't spend too much time ATM.

functions/src/world.ts
denisdimitrov: The proposal is valid and better than nothing. However we need to think of a more global way to type these database responses.

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
denisdimitrov: I understand put the component is not disabled. It just has tech debt and it needs to be finished. Adding `DISABLED_DUE_TO` would be miss-leading. I can either finish the implementation or improve the comment if you'd like.

src/components/admin/WorldUserCard/WorldUserCard.tsx
denisdimitrov: Valid point.

PR #2917, opened 2022-03-11
replace new name for manually created tables

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1687

Spaces with tables with have default table names `table1` to `table10`
Manually created tables will be named `New Table`

### DIFF
--- a/src/utils/table.ts
+++ b/src/utils/table.ts
@@ -1,22 +1,22 @@
-import { uuid } from "uuidv4";
-
 import { DEFAULT_TABLE_COLUMNS, DEFAULT_TABLE_ROWS } from "settings";
 
 import { Table } from "types/Table";
 
-const generateUniqueTableReference = (title: string) => `${title}-${uuid()}`;
+import { generateTableReferenceId } from "./string";
 
 export const generateTable: (props: {
-  tableNumber?: number;
+  tableNumber: number;
   columns?: number;
   rows?: number;
+  generateTableReference?: (title: string) => string;
 }) => Table = ({
   tableNumber,
   columns = DEFAULT_TABLE_COLUMNS,
   rows = DEFAULT_TABLE_ROWS,
+  generateTableReference = generateTableReferenceId,
 }) => {
-  const title = tableNumber ? `Table ${tableNumber}` : "New table";
-  const reference = generateUniqueTableReference(title);
+  const title = `Table ${tableNumber}`;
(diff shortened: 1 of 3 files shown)


src/utils/table.ts
```diff
@@ -4,17 +4,19 @@ import { DEFAULT_TABLE_COLUMNS, DEFAULT_TABLE_ROWS } from "settings";
...
   columns?: number;
   rows?: number;
 }) => Table = ({
   tableNumber,
   columns = DEFAULT_TABLE_COLUMNS,
   rows = DEFAULT_TABLE_ROWS,
 }) => {
-  const title = `Table ${tableNumber}`;
-  const reference = `${title}-${uuid()}`;
+  const title = tableNumber ? `Table ${tableNumber}` : 'New table';
```
goran-peoski-work, line 18: Should this be just `New table` or could it be something like `New table`, `New table (2)`, `New table (3)` as seen in other software?


[Q] Can it be quick and easy to do something like 

```suggestion
  const title = tableNumber ? `Table ${tableNumber}` : `New Table (${someShortGeneratedUniqueToken})`;
```

goran-peoski-work (approved):
LGTM

I'd hope the moving function is a good suggestion for this PR, the other one... well... just an idea if only is a quick one to implement.

src/utils/table.ts
```diff
@@ -4,17 +4,19 @@ import { DEFAULT_TABLE_COLUMNS, DEFAULT_TABLE_ROWS } from "settings";
 
 import { Table } from "types/Table";
 
+const generateUniqueTableReference = (title: string) => `${title}-${uuid()}`;
```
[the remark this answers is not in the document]
alexey-petrov: done [d8c823c](https://github.com/sparkletown/sparkle/pull/2917/commits/d8c823ca7f62a8f092735f9313b0aceb021a2d93)

src/utils/table.ts
alexey-petrov: won't do, there's no way to guarantee that table name is unique unless we're using a `uuid`-like indicator, which would ruin UX ( seeing something like `New table XXXX-XXXX-XXXX` isn't really helpful ) and adding a lot of if/else checks to ensure that the next added table's name is exactly `X+1` might lead to other problems
discussed in slack

PR #2918, opened 2022-03-11
Show Embeddable space template in the list

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1669

decided to show Embeddable template as it's easier.
Finishing work with ViewingWindow template would require more efforts + migration of the existed ArtPiece and Embeddable spaces.


mike-lvov (approved):
Lgtm

PR #2919, opened 2022-03-11
fix schedule space id filter

alexey-petrov:
Fixes schedule event filtering. Previously it failed to match the events in the correct spaces due to an incorrect filter value comparison, resulting in errors such as empty schedule list or 
image

### DIFF
--- a/src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
+++ b/src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
@@ -90,7 +90,7 @@ export const ScheduleOverlay: React.FC = () => {
     isEventsLoading,
     sovereignVenue,
   } = useVenueScheduleEvents({ userEventIds });
-
+  console.log(dayDifference, liveAndFutureEvents);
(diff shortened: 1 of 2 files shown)


src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
```diff
@@ -202,11 +202,12 @@ export const ScheduleOverlay: React.FC = () => {
 
     const currentVenueBookMarkEvents = eventsFilledWithPriority.filter(
       ({ isSaved, spaceId: eventSpaceId }) =>
-        isSaved && eventSpaceId?.toLowerCase() === space?.id
+        isSaved && eventSpaceId?.toLowerCase() === space?.id.toLowerCase()
```
goran-peoski-work, line 205: [NOTE]

Useful utility function for the future `compareStringsNoCase( )` or similar that can be called every time we need to compare two strings and don't want to remember to check if `null`, `undefined` etc.

goran-peoski-work (approved):
LGTM

PR #2921, opened 2022-03-12
Schedule components refactor

alexey-petrov:
A follow up for 11.03.2022 Play in the Product that had this as an error
And additional refactoring & code cleanup for schedule components

image

### DIFF
--- a/src/components/attendee/IntersectingButton/IntersectingButton.tsx
+++ b/src/components/attendee/IntersectingButton/IntersectingButton.tsx
@@ -1,4 +1,4 @@
-import { ButtonHTMLAttributes, RefObject, useEffect } from "react";
+import { ButtonHTMLAttributes, RefObject, useEffect, useRef } from "react";
 import { useIntersection } from "react-use";
 import { Button } from "components/attendee/Button";
 
@@ -14,7 +14,7 @@ interface IntersectingButtonProps
   transparent?: boolean;
   border?: BorderVariant;
   className?: string;
-  forwardRef: RefObject<HTMLButtonElement>;
+  forwardRef?: RefObject<HTMLButtonElement>;
   updateIntersected: (el: RefObject<HTMLButtonElement>) => void;
 }
 
@@ -24,15 +24,17 @@ export const IntersectingButton: React.FC<IntersectingButtonProps> = ({
   updateIntersected,
   ...rest
 }) => {
-  const buttonIntersect = useIntersection(forwardRef, {
+  const fallbackRef = useRef<HTMLButtonElement>(null);
+  const targetRef = forwardRef ?? fallbackRef;
+  const buttonIntersect = useIntersection(targetRef, {
     rootMargin: "0px",
   });
 
   useEffect(() => {
     if (buttonIntersect?.isIntersecting) {
-      updateIntersected(forwardRef);
+      updateIntersected(targetRef);
     }
-  }, [buttonIntersect, updateIntersected, forwardRef]);
+  }, [buttonIntersect, updateIntersected, targetRef]);
 
-  return <Button {...rest} forwardRef={forwardRef} onClick={onClick} />;
+  return <Button {...rest} forwardRef={targetRef} onClick={onClick} />;
 };
--- a/src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
+++ b/src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
@@ -1,57 +1,8 @@
-import React, {
-  RefObject,
-  useCallback,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
-import { useIntersection } from "react-use";
-import { faBookmark as regularBookmark } from "@fortawesome/free-regular-svg-icons";
-import {
-  faAngleLeft as angleLeft,
-  faAngleRight as angleRight,
-  faBookmark as solidBookmark,
-} from "@fortawesome/free-solid-svg-icons";
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import classNames from "classnames";
-import { IntersectingButton } from "components/attendee/IntersectingButton";
+import React, { useMemo, useState } from "react";
 import { Toggler } from "components/attendee/Toggler";
-import {
-  addDays,
-  differenceInCalendarDays,
-  fromUnixTime,
-  isToday,
-  startOfDay,
-  startOfToday,
-} from "date-fns";
+import { fromUnixTime, isToday, startOfDay, startOfToday } from "date-fns";
 
-import {
-  ALWAYS_EMPTY_ARRAY,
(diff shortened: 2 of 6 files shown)


src/components/attendee/ScheduleOverlay/helpers.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { MutableRefObject, RefObject } from "react";
```
goran-peoski-work, line 1: This is a code smell. If it is helpers, it should probably be a `.ts` file in `utils`, but if it has `.tsx` extension, there is probably a component hidden somewhere. Might be good to extract it into a component that uses one or two utility functions.

src/components/attendee/ScheduleOverlay/helpers.tsx
```diff
@@ -0,0 +1,84 @@
...
+  dayDifference: number;
+  firstScheduleDate: Date;
+  selectedDayIndex: number;
+  liveAndFutureEvents: ScheduledEvent[];
+  isScheduleTimeshifted: boolean;
+  setSelectedDayIndex: React.Dispatch<React.SetStateAction<number>>;
+  allRefs?: MutableRefObject<RefObject<HTMLButtonElement>[]>;
+  updateIntersected?: (el: RefObject<HTMLButtonElement>) => void;
+};
+export const generatedWeekDays = ({
```
goran-peoski-work, line 30: This is basically a `<WeekdaysSomethingSomething>` component

src/components/attendee/ScheduleOverlay/ScrollableWeekdays.tsx
```diff
@@ -0,0 +1,188 @@
...
+    scrollRef?.scrollIntoView({
+      behavior: "smooth",
+      block: "nearest",
+      inline: "start",
+    });
+    recalculateSkipPrevValues(scrollRef);
+  };
+
+  // triggered on arrow click, sets prev arrow value to {currentIntersectingDay - 14} or the first day
+  // and next arrow value to {currentIntersectingDay + 7} or the last day
```
goran-peoski-work, line 123: Comments are good, but this one like the previous and next, doesn't need to have hard-coded numbers, there are already constants for it. Best to describe the why, not the what.

goran-peoski-work: [Q] Since this comment is repeated 3 times, isn't the logic inside also repeated? Can it be easily extracted into a helper function or would it take too much time thus making it OOS?

src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
```diff
@@ -338,32 +60,31 @@ export const ScheduleOverlay: React.FC = () => {
...
+          setSelectedDayIndex={setSelectedDayIndex}
+          selectedDayIndex={selectedDayIndex}
+          isScheduleTimeshifted={isScheduleTimeshifted}
+        />
+      ) : (
+        <ScrollableWeekdays
+          setSelectedDayIndex={setSelectedDayIndex}
+          selectedDayIndex={selectedDayIndex}
+          dayDifference={dayDifference}
+          isScheduleTimeshifted={isScheduleTimeshifted}
```
goran-peoski-work, line 74: [code smell] These names are contrary to common HTML and React practices. Handlers usually have name of `onSomething` and it's easy to make the parent take the arguments from the handler and call it's own `setWhatever` function.

[OOS] probably
The components as they are now look like tightly coupled almost as if having extra component adds more noise in passing values and listeners than helping compartmentalize logic.

There should be a way to improve the code, but I'm afraid it may take too much time and be OOS for this PR.

src/components/attendee/IntersectingButton/IntersectingButton.tsx
```diff
@@ -14,7 +14,7 @@ interface IntersectingButtonProps
   transparent?: boolean;
   border?: BorderVariant;
   className?: string;
-  forwardRef: RefObject<HTMLButtonElement>;
+  forwardRef?: RefObject<HTMLButtonElement>;
```
goran-peoski-work, line 17: Plenty of unnecessary field definitions here, `IntersectingButtonProps` can just extend from `ButtonProps` or be 

```tsx
type IntersectingButtonProps = ButtonProps & { forwardRef?: RefObject<HTMLButtonElement> };
```

Not even the `onClick` is modified in this HoC so it might just be 

```tsx
return <Button {...rest} forwardRef={targetRef} />
```

[NOTE] It's OK to export/import ButtonProps from "components/attendee/Button" as long as they're needed elsewhere

goran-peoski-work (changes requested):
Few suggestions, mostly about dividing logic into different components.

Not all needs to be refactored now, due to time constraints, just what's quick and easy to do and the rest might use `@debt` comments

src/components/attendee/ScheduleOverlay/helpers.tsx
alexey-petrov: done [ede5a78](https://github.com/sparkletown/sparkle/pull/2921/commits/ede5a78fee222b03794b50d81d1b5bb452d05814)

src/components/attendee/ScheduleOverlay/ScrollableWeekdays.tsx
alexey-petrov: replaced comments + added a `TODO` [ede5a78](https://github.com/sparkletown/sparkle/pull/2921/commits/ede5a78fee222b03794b50d81d1b5bb452d05814)

src/components/attendee/ScheduleOverlay/ScheduleOverlay.tsx
alexey-petrov: Replaced with `onSelectDayIndex`, maybe will decouple them in the future [ede5a78](https://github.com/sparkletown/sparkle/pull/2921/commits/ede5a78fee222b03794b50d81d1b5bb452d05814)

src/components/attendee/IntersectingButton/IntersectingButton.tsx
```diff
@@ -1,21 +1,12 @@
...
-  onClick?: () => void;
-  variant?: ButtonVariant;
-  transparent?: boolean;
-  border?: BorderVariant;
-  className?: string;
-  forwardRef?: RefObject<HTMLButtonElement>;
+  extends ButtonHTMLAttributes<HTMLButtonElement>,
+    ButtonProps {
   updateIntersected: (el: RefObject<HTMLButtonElement>) => void;
+  onClick?: () => void;
```
goran-peoski-work, line 9: I don't think this one is needed, by deleting `onClick` from 3 places it can just be passed through as part of `rest` props

alexey-petrov: this caused another TS error saying that `onClick` types differ from `htmlbutton` one
not sure why, leaving it as it is for now

PR #2924, opened 2022-03-14
Fix <Toggle> double click in <PortalStripForm>

goran-peoski-work:
Resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1759

The previous version of the `<Toggle >` component used `register` to handle data. The new version switched the handling to `onChange()` handler on the `<input>` element which doesn't always capture the clicks (without `register` even more so).

This update moves the handler to `onClick()` for the `<label>` component and adds a bit of cleanup and comments to explain the logic behind it.


sunny-viktoryia (approved):
Looks good to me.

PR #2926, opened 2022-03-15
profile links + validation fixes + minor formatting

alexey-petrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1761

Multiple fixes of profile form + minor reformatting


src/components/attendee/InputSelect/InputSelect.tsx
```diff
@@ -18,70 +18,83 @@ interface InputSelectProps
...
 
   return (
     <div className={containerClassNames}>
-      <div className={styles.wrapper}>
+      <form className={styles.wrapper} autoComplete="off">
         <div className={wrapperClassNames}>
           <input
             className={inputClassNames}
             placeholder={inputPlaceholder}
+            autoComplete="off"
```
mike-lvov, line 69: do we need to state `autocomplete=off` in both places? I think input should be enough

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
```diff
@@ -80,19 +80,12 @@ export const ProfileOverlay: React.FC<ProfileOverlayProps> = ({ profile }) => {
...
     shouldUnregister: true,
   });
 
-  const onDeleteLink = useCallback(
-    (i: number) => {
-      removeLink(i);
-    },
-    [removeLink]
-  );
-
```
mike-lvov, line 95: Why have you removed the `deletion` funcitonality?

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
```diff
@@ -167,13 +175,11 @@ export const ProfileOverlay: React.FC<ProfileOverlayProps> = ({ profile }) => {
           partyNameError={errors?.partyName}
         />
 
-        <form onSubmit={handleSubmit(onSubmit)}>
```
mike-lvov, line 170: Why did u change `form` to `div` here?

mike-lvov (commented):
A few questions

src/components/attendee/InputSelect/InputSelect.tsx
alexey-petrov: for some reason autocomplete disables only if we put it in the form and the inputs inside it, otherwise it still works

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
alexey-petrov: in the current version of profile links there's no delete mechanism via a button ( we can only delete if we erase link's url and save the form ), so there's no need for this function

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
alexey-petrov: otherwise I was getting `there couldn't be a form as a descendant of form` error due to the change in the `InputSelect`

src/components/attendee/InputSelect/InputSelect.tsx
```diff
@@ -18,70 +18,83 @@ interface InputSelectProps
...
+  errorInput,
   inputName,
   selectName,
   inputPlaceholder,
   selectPlaceholder,
   register,
-  inputRules,
-  selectRules,
+  inputRules = {},
+  selectRules = {},
```
goran-peoski-work, line 44: [NOTE]
I'm not sure about this one. Usually there should be no error, but in case there is one, do we need the other one?

Do we return `error: FIeldError | FieldError[]` or `error: FIeldError | Record<string,FieldError>` instead of having twice the props?

Just something to think about for later date.

goran-peoski-work: [OOS] There are some seed sparkle flavored errors that might work with this multiple errors returned thing, if for example there is `error: SparkleErrors` which is an error that has an array of `cause` errors inside.

Improvement for other times I guess, so can we please add a `@debt` comment here so we can return to it?

goran-peoski-work (approved):
LGTM

Some comments added, for future consideration.

src/components/attendee/InputSelect/InputSelect.tsx
alexey-petrov: added a debt comment to possibly unite error objects into a single entity [614df69](https://github.com/sparkletown/sparkle/pull/2926/commits/614df691194b102f72660673698d0af99c340241)

PR #2928, opened 2022-03-15
Makes chatbox display the latest messages

mike-lvov:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/176

https://user-images.githubusercontent.com/63194656/158488722-c0434e3e-0070-481b-87a5-a27abee5f8aa.mov

### DIFF
--- a/src/components/molecules/Chatbox/Chatbox.module.scss
+++ b/src/components/molecules/Chatbox/Chatbox.module.scss
@@ -7,6 +7,8 @@
 }
 
 .chatboxMessages {
+  display: flex;
+  flex-direction: column-reverse;
   flex-grow: 1;
   // Start tiny and grow so that the send box is always visible
   flex-basis: 1px;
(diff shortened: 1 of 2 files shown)


src/components/molecules/Chatbox/Chatbox.module.scss
```diff
@@ -7,6 +7,8 @@
 }
 
 .chatboxMessages {
+  display: flex;
+  flex-direction: column-reverse;
```
goran-peoski-work, line 11: I don't think depending on CSS is the way to go for displaying data properly. As a rule of thumb, data should be in order even if no styles are applied. 

So, can it be done in a way like `itemsToDisplay.sort().reverse()` or similar before outputing?

I mean, there are even nicer utilities from Flexbox, like `order` CSS property, but if we can display the data properly without resorting to style, I think we should use that.

goran-peoski-work (changes requested):
I don't think just CSS reverting of the sort is the way to go. There might be a simple JS way to do it as well `itemsToDisplay.reverse()`

goran-peoski-work (approved):
I took half an hour trying to figure this out. There might be something to do with the `inverse` prop of `<InfiniteScroll>` but this PR does solve the issue, even if in the CSS layer.

whoaitsaimz:
I dont think #176 is the correct issue number, but it didn't seem to auto-close it anyway, so prob not a big deal 

...unless someone else at Sparkle is as curious as me and clicks into related issues ;-)

PR #2929, opened 2022-03-16
Report error in rare cases secionId is missing

goran-peoski-work:
Adding an assert error as a type of sparkle error and utilities to capture its info, like the place of it.

This is probably not an isolated issue since types can assume data is there and in certain form, but reading from DB might not produce it as expected. This assertion in `SeatingBlock` component will note it (at least in console for now) and not make the page break.


alexey-petrov (approved):
LGTM

PR #2932, opened 2022-03-17
improve invite admin modal and add search functionality

denisdimitrov:
Adds user search functionality with algolia and invite admin logic has been reworked to allow bulk user selection.

Partially fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1703


functions/src/world.ts
```diff
@@ -172,13 +172,16 @@ export const addWorldAdmin = functions.https.onCall(async (data, context) => {
...
-  if (!Array.isArray(owners) || owners.includes(userId)) {
+  if (!Array.isArray(owners)) {
     return;
   }
 
-  const newOwners = [...owners, userId];
+  const newOwners = [...owners, ...userIds];
+
+  //Removes duplicates from the concat array.
+  const uniqueOwners = [...new Set(newOwners)];
```
goran-peoski-work: [NOTE]
No need to change this code, just pointing out if we use Lodash for functions, seeing function named `uniq` would replace the need of an explanation comment. This also means if we see too many of these [...new Set()] snippets, we might make our own `unique` function.

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +43,84 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
-  const { register } = useForm({
+  const { register, watch } = useForm({
     reValidateMode: "onChange",
     defaultValues,
   });
 
-  const addAdmin = async () => {
+  const searchQuery = watch("userName");
+
+  const { value: algoliaValue } = useAlgoliaSearch(searchQuery);
```
goran-peoski-work, line 54: [Q] Is the hook throttling too many changes at too short intervals?

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +43,84 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
+    setSelectedUsers((selectedUsers) => {
+      if (selectedUsers.includes(user)) {
+        const newUsers = selectedUsers.filter(
+          (selectedUser) => selectedUser !== user
+        );
+        return newUsers;
+      }
+
+      return [...selectedUsers, user];
+    });
```
goran-peoski-work: I think removing bracers, a variable, and changing `if-else` into `? :` might make these lines simpler and reduce syntax noise ratio

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +43,84 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
+  const checkAdminAccess = (user: User) => selectedUsers.includes(user);
+
+  const hasFoundUsers = !!foundUsers?.length;
+  const hasSelectedUsers = selectedUsers?.length;
+
   return (
     <Modal show={show} onHide={onHide} autoHide>
-      <Input placeholder="Search users" name="userName" register={register} />
+      {hasFoundUsers && (
+        <div className="flex justify-center">{`${foundUsers.length} users found`}</div>
```
goran-peoski-work: [Q] Should we bother with singular/plural here?

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +43,84 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
+
+      <Input placeholder="Type user name" name="userName" register={register} />
 
+      <div className="max-h-40 overflow-auto">
+        {foundUsers?.map((user) => (
+          <div
+            key={user.objectID}
+            className={`flex flex-row py-4 px-2 border-y hover:bg-gray-200 cursor-pointer ${
+              checkAdminAccess(user) && "bg-blue-100 hover:bg-blue-200"
+            }`}
```
goran-peoski-work: I recommend replacing this with call to the classnames function before the return. It makes it easy to spot as
```tsx
["bg-blue-100 hover:bg-blue-200"]: checkAdminAccess(user)
```

goran-peoski-work (approved):
LGTM

Some non-blocking suggestions and questions added

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
denisdimitrov: Honestly, I didn't notice any throttle/debounce behavior whatsoever. I think we need to check the hook and improve it.

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
denisdimitrov: Yup, I'll improve the wording.

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
goran-peoski-work: K, how about this?
```suggestion
  // @debt check if searchQuery is triggering too many algoliaValue changes per unit of time
  const searchQuery = watch("userName");
  const { value: algoliaValue } = useAlgoliaSearch(searchQuery);
```

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +46,91 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
     reValidateMode: "onChange",
     defaultValues,
   });
 
-  const addAdmin = async () => {
+  const searchQuery = watch("userName");
+
+  const { value: algoliaValue } = useAlgoliaSearch(searchQuery);
+
+  const foundUsers = algoliaValue?.[AlgoliaSearchIndex.USERS].hits;
```
goran-peoski-work, line 56: Not sure about the value of `AlgoliaSearchIndex.USERS` but since there are null/undefined guards already...
```suggestion
  const foundUsers = algoliaValue?.[AlgoliaSearchIndex.USERS]?.hits;
```

src/components/admin/InviteAdminModal/InviteAdminModal.tsx
```diff
@@ -27,29 +46,91 @@ export const InviteAdminModal: React.FC<InviteAdminModalProps> = ({
...
+            {
+              "bg-blue-100 hover:bg-blue-200": checkAdminAccess(user),
+            }
+          );
+
+          return (
+            <div
+              key={user.objectID}
+              className={foundUserClasses}
+              onClick={() => selectUser(user)}
```
goran-peoski-work, line 112: this could've been `useCallback()` I guess

src/utils/string.ts
```diff
@@ -16,3 +16,16 @@ export const generateId: (prefix: string, tailLength?: number) => string = (
...
+ * @param word The string you want to pluralize.
+ * @param array The array that the word is dependent on.
+ * @returns The provided string in either singular or plural state.
+ */
+export const pluralize = <T>(word: string, array: Array<T> | undefined) => {
+  const arrayExists = !!array?.length;
+  const hasMultipleResults = arrayExists ? array?.length > 1 : undefined;
+  const hasNoResults = arrayExists ? array?.length === 0 : undefined;
+  return hasNoResults || hasMultipleResults ? word + "s" : word;
+};
```
goran-peoski-work: I generally don't like `pluralize` functions, not only do they add time, but also complexity. Now you need to deal with edge cases, irregular plural forms etc.

Same problem is easily solved by providing 2 different strings at the place two different strings need to be shown.

PR #2934, opened 2022-03-17
Replaces prototypal flat with lodash flatten

mike-lvov:
The node version on staging doesn't have .flat() in its prototype.
That's why it worked locally, but didn't work in staging

closes https://github.com/sparkletown/internal-sparkle-issues/issues/1641

### DIFF
--- a/functions/analytics.js
+++ b/functions/analytics.js
@@ -10,39 +10,40 @@ const { formatSecondsAsHHMMSS } = require("./src/utils/time");
 const { checkIsAdmin } = require("./src/utils/permissions");
 const { checkAuth } = require("./src/utils/assert");
 
-const { chunk } = lodash;
+const { chunk, flatten } = lodash;
 
 const functionsConfig = functions.config();
 
 const getUsersWithVisits = async (venueIdsArray) => {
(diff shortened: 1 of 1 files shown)


functions/analytics.js
```diff
@@ -10,39 +10,40 @@ const { formatSecondsAsHHMMSS } = require("./src/utils/time");
...
-                }))
-              );
-
-            return { user, visits };
-          })
-        );
-    })
-    .map((el) => el.then((res) => res.flat()))
-    .flat();
+  const dto = flatten(
```
goran-peoski-work, line 18: I think this is what's needed https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap . I don't know why lodash utility is preferred over built-in JS functions though.

That being said, the solution that uses built-ins might need tweaks with Webpack and what polyfills it needs, so, for now, a minor suggestion:

```suggestion
  // @debt see if JS built in .flatMap is good fit for this logic (maybe would require polyfill to be set)
  const dto = flatten(
```

goran-peoski-work (approved):
LGTM

PR #2941, opened 2022-03-21
redesign space timing tab and show events

denisdimitrov:
The space timing tab is now redesign and shows the space's events properly. `Create new experience` button is also added, which allows users to create events directly from that screen. Some minor refactoring as well.

space-timing


src/components/admin/EventsPanel/EventsPanel.tsx
```diff
@@ -39,7 +41,7 @@ export const EventsPanel: React.FC<EventsPanelProps> = ({
     [spaces]
   );
 
-  const hasVenueEvents = events?.length !== 0;
+  const hasSpaceEvents = events?.length !== 0;
```
goran-peoski-work, line 44: Is this OK? 
If `events?.length` returns `undefined` the result will be `true`.

There are multiple ways to solve this, two examples are: `(events?.length ?? 0) !== 0` or simpler ` !! events?.length `

goran-peoski-work (approved):
LGTM

BUT, it think I might have spotted a bug from before the redesign. Not sure if OOS. Might be worth checking before merge.

src/components/admin/EventsPanel/EventsPanel.tsx
denisdimitrov: Fixed in: https://github.com/sparkletown/sparkle/pull/2941/commits/45c547cc49b25a3a7391b6468e28379ec01dd5b2

PR #2942, opened 2022-03-21
Router LoginRestricted and analytics core components cleanup

goran-peoski-work:
In general there is more work to be done on the hooks and Reactfire, 
but to avoid major issues, 
the decision is to add chunks of the merged-then-reverted PR 
- https://github.com/sparkletown/sparkle/pull/2894

piecemeal in order to catch culprits.


src/components/templates/Auditorium/components/SeatingBlock/SeatingBlock.tsx
```diff
@@ -44,7 +44,7 @@ export const SeatingBlock: React.FC<SeatingBlockProps> = ({ space }) => {
 
   return (
     <Switch>
-      <Section venue={space} sectionId={sectionId} user={user} />
```
colinhowe, line 47: We only allow for one section at the moment, this change isn't necessary

colinhowe (changes requested):
Minor change and then good to go

src/components/templates/Auditorium/components/SeatingBlock/SeatingBlock.tsx
goran-peoski-work: That's a leftover from before, no? It kind of snuck in from earlier

goran-peoski-work: What change?

src/components/templates/Auditorium/components/SeatingBlock/SeatingBlock.tsx
```diff
@@ -44,7 +44,7 @@ export const SeatingBlock: React.FC<SeatingBlockProps> = ({ space }) => {
...
+      {sectionId && <Section venue={space} sectionId={sectionId} user={user} />}
```
colinhowe, line 47:
```suggestion
      {<Section venue={space} sectionId={sectionId} user={user} />}
```

src/components/templates/Auditorium/components/SeatingBlock/SeatingBlock.tsx
goran-peoski-work: Regardless what other places of the code and the DB have, according to the line above:

`const sectionId = urlSectionId || sortedSections[0]?.id;` 

And according to the fact I did have page breaking error without the check above and the one on this line, I think it's better safe than sorry.

colinhowe (approved):

PR #2943, opened 2022-03-21
Allow setting of background images for all spaces

colinhowe:
Remove the ability to set a banner image as that isn't used in new design

image


src/components/admin/SpaceEditForm.tsx
```diff
@@ -132,9 +132,9 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
       space.parentId,
       space.sectionsCount,
       space.roomVisibility,
-      space.zoomUrl,
+      space?.zoomUrl,
       space?.host?.icon,
-      space?.config?.landingPageConfig?.coverImageUrl,
+      space.backgroundImageUrl,
```
goran-peoski-work, line 137: Not sure if `?.` is OK or `.` is OK, but would be good if all are the same

src/pages/VenuePage/TemplateWrapper.module.scss
```diff
@@ -4,10 +4,33 @@
...
+  background-repeat: no-repeat;
 }
 
 .shrunk {
   @media (min-width: breakpoint.$md) {
     width: calc(100% - layout.$w-sidebar);
   }
 }
+
+.gradients {
```
goran-peoski-work, line 18: Not sure if this is a style for `TemplateWrapper` since there is an `<AttendeeLayout />` component, but I guess is OK for the time being if this is simpler until `<TemplateWrapper />` gets fully absorbed by parent

goran-peoski-work (approved):
LGTM

PR #2944, opened 2022-03-21
Portal create/edit modal

alexxpetrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1758

* Create/edit portal modal
* confirmation modal style fixes

image
image
image


src/components/atoms/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -51,14 +49,24 @@ export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
...
-        <div className="ConfirmationModal__message">{message}</div>
-        <div className="ConfirmationModal__buttons">
-          <ButtonNG onClick={cancel}>{cancelBtnLabel}</ButtonNG>
-          <ButtonNG variant={confirmVariant} onClick={confirm}>
+        <div>{message}</div>
+        <div className="mt-4">
+          <Button
+            onClick={cancel}
+            variant="secondary"
+            className="mt-3 mb-3 ml-2 mr-2"
```
goran-peoski-work: Please remove this.

The reason this button didn't have `className` before is because it's very hard to make UI consistent if classes can be freely attached. 

On first look, it might seem like a good idea to follow the example of the props named `variant` and `borders` that in `Button` are used like: 
```tsx
className={ cn("Button", TW.general, TW[variant], TW[borders]) }
```
so a need of some new degree of visual variation would add a new prop with strictly controlled input values like:

`<Button variant="secondary" margins="normal"`
`<Button variant="secondary" margins="wide"`

On second look, most likely

**where a button is placed shouldn't be part of its own definition but part of the parent's layout**. 

This can be achieved by adding a new `<div>` container and/or tweaking the container's flexbox without intruding on the uniformity of buttons themselves.

src/components/admin/Button/Button.tsx
```diff
@@ -13,6 +13,7 @@ interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   borders?: ButtonBorders;
   loading?: boolean;
   forwardRef?: RefObject<HTMLButtonElement>;
+  className?: string;
```
goran-peoski-work: This should not be added as to limit the buttons' appearance to some small predictable uniformed appearances.
```suggestion
```

src/components/admin/Button/Button.tsx
```diff
@@ -22,12 +23,13 @@ export const Button: React.FC<ButtonProps> = ({
   loading = false,
   forwardRef,
   type = "button",
+  className,
```
goran-peoski-work:
```suggestion
```

src/components/admin/Button/Button.tsx
```diff
@@ -22,12 +23,13 @@ export const Button: React.FC<ButtonProps> = ({
...
   ...extraParams
 }) => (
   <button
     {...extraParams}
     type={type}
-    className={cn("Button", TW.general, TW[variant], TW[borders])}
+    className={cn("Button", TW.general, TW[variant], TW[borders], className)}
```
goran-peoski-work: Please see the comment at the place `className=""` is being used for alternative.

```suggestion
    className={cn("Button", TW.general, TW[variant], TW[borders])}
```

src/components/atoms/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -51,14 +49,24 @@ export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
...
+          >
+            {cancelBtnLabel}
+          </Button>
+          <Button
+            variant={confirmVariant}
+            onClick={confirm}
+            className="mt-3 mb-3 ml-2 mr-2"
```
goran-peoski-work: Same as above

```suggestion
```

src/components/atoms/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -51,14 +49,24 @@ export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
...
     <Modal show={isShown} onHide={hide}>
-      <div className="ConfirmationModal">
+      <div className="flex flex-col items-center">
         {hasHeader && <h4>{header}</h4>}
-        <div className="ConfirmationModal__message">{message}</div>
-        <div className="ConfirmationModal__buttons">
-          <ButtonNG onClick={cancel}>{cancelBtnLabel}</ButtonNG>
-          <ButtonNG variant={confirmVariant} onClick={confirm}>
+        <div>{message}</div>
+        <div className="mt-4">
```
goran-peoski-work, line 53: Instead of adding classes to the buttons themselves, this div flexbox can be tweaked by changing the gap, alignment/justification etc.

src/components/atoms/Dropdown/Dropdown.scss
```diff
@@ -0,0 +1,11 @@
+.dropdownSelected {
+  mask-image: url(/assets/icons/checkmark-icon.svg);
+}
+
+.optionContainer {
+  &:hover {
+    span {
+      background-color: white;
```
goran-peoski-work: [Q] Several questions pop up here:
- Should this one still be hardcoded white? 
- If not hardcoded, should `src/scss/attendee/theme.scss` color constant be used?
- But if used in admin, shouldn't Tailwind be preferred?
- But if needed for attendee, shouldn't it be converted to `.module.scss` and this component moved?


Most likely finding a good answer would take time, so, instead, there might be a way to put some `@deprecated` comment on this one and use another component (even started as copy paste from here to `src/component/admin` or `src/component/attendee`).

But, due to OOS, maybe just some short version of this comment as `@debt` in the code that we need to deprecate the dropdown and replace with appropriate one would be enough.

src/components/atoms/PortalVisibility/PortalVisibility.scss
```diff
@@ -0,0 +1,3 @@
+.portalImage {
```
goran-peoski-work: This does not look OK. File is `.scss` not `.module.scss` so most likely it should be deleted and Tailwind one used insted. Maybe even move this "atom" to the admin components.

src/components/atoms/PortalVisibility/PortalVisibility.tsx
```diff
@@ -40,46 +39,29 @@ export const PortalVisibility: React.FC<PortalVisibilityProps> = ({
...
-                    ))}
-                  </div>
-                )}
-              </div>
-              <span className="PortalVisibility__caption">{label}</span>
+        return (
+          <div
+            key={label}
+            onClick={(e) => {
+              e.preventDefault();
```
goran-peoski-work: We've been always trying to avoid `preventDefault()` in event handlers so that bubbling up isn't stopped.
The way to go is usually an `if() return;`  at the start of the handler.

In a very very unlikely workaround for some obscure bug or a problem, `preventDefault()` can be used with a good comment explaining the exception, but in any case I think `useCallback()` would be useful here.

src/components/atoms/PortalVisibility/PortalVisibility.tsx
```diff
@@ -40,46 +39,29 @@ export const PortalVisibility: React.FC<PortalVisibilityProps> = ({
...
-                        </span>
-                      </div>
-                    ))}
-                  </div>
-                )}
-              </div>
-              <span className="PortalVisibility__caption">{label}</span>
+        return (
+          <div
+            key={label}
```
goran-peoski-work: [NOTE]
This is interesting. Probably no change is needed, but nice to note.

This `label` might not always be unique if, for example, new items are added without labels (nulls, undefined or empty strings, whatever). `index` or pairing it as `index + '-' + label` is also viable since the order of labels will be very unlikely to change in `LABEL_VISIBILITY_OPTIONS` during the runtime.

src/components/atoms/PortalVisibility/PortalVisibility.tsx
```diff
@@ -40,46 +39,29 @@ export const PortalVisibility: React.FC<PortalVisibilityProps> = ({
...
+            onClick={(e) => {
+              e.preventDefault();
+              setSelected(value);
+              setValue(name, value, { shouldValidate: true });
+            }}
+          >
+            <div className={imageClasses}>
+              <img src="#" alt="" />
```
goran-peoski-work: [Q]
Is this supposed to be a placeholder to be implemented in the future?
Maybe a `@debt` or `TODO` comment is needed.
Otherwise, what's it's purpose?

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -19,12 +19,14 @@ const spaceNoneOption = Object.freeze({
   id: "",
   name: "",
   template: undefined,
+  host: { icon: "" },
```
goran-peoski-work, line 21: [Q] Why is `host` even used? Can it not be flat `icon` at the top level? Like

```tsx
const spaceNoneOption: SpacesDropdownPortal = {};
Object.freeze(spaceNoneOption);
```
Side note: name here should probably be like a constant, e.g. `SPACE_OPTION_NONE`

If  `host` must be there, we may need to add some Interface that is a kind of partial of the space.
In that case, please add a debt comment to formally define new interface and/or do an interface segregation of the target one

src/components/molecules/Modal/Modal.tsx
```diff
@@ -50,11 +50,11 @@ export const Modal: React.FC<ModalProps> = ({
     >
       <div
         className={
-          "max-h-screen relative px-4 w-full max-w-2xl h-full md:h-auto"
+          "max-h-screen relative px-4 w-full max-w-2xl h-full md:h-auto sm:max-w-lg sm:w-full"
         }
         ref={containerRef}
       >
-        <div className="relative bg-white rounded-lg shadow dark:bg-gray-700">
+        <div className="inline-block align-middle bg-white rounded-lg px-4 pt-5 pb-4 text-left overflow-hidden shadow-xl transform transition-all sm:my-8 sm:max-w-lg sm:w-full sm:p-6">
```
goran-peoski-work: These sausages would be more manageable (not make the code too wide) if we put them in `Modal.tailwind.ts` and would probably be best we move this component in the `src/components/admin` directory so it's not mistakenly used in the attendee side.

src/components/admin/Toggle/Toggle.tsx
```diff
@@ -1,32 +1,54 @@
...
 }
 
 export const Toggle: React.FC<ToggleProps> = ({
   label,
   checked = false,
   onChange,
-  name,
+  register,
+  name = "",
+  title,
```
goran-peoski-work:
```suggestion
  title = "",
```

src/components/admin/Toggle/Toggle.tsx
```diff
@@ -1,32 +1,54 @@
...
 }) => {
   // NOTE: the click handlers needs to be on the label to capture all its children clicks and there must always be a <label>
-  const handleClick = useCallback((event) => onChange?.(event), [onChange]);
+  const handleClick = useCallback(
+    (event) => {
+      // default action has to be prevented in order to avoid `double click`
+      onChange && event.preventDefault();
```
goran-peoski-work: Please remove this. The Toggle was fixed.

Whatever this "double click" problem is, if it is on the parent side, not inside this component, the parent's `onChange={changeHandler}` can do the `if() return`

On another note, I've already written about why we have avoided `preventDefault()` and this is also a case.

src/components/admin/Toggle/Toggle.tsx
```diff
@@ -1,32 +1,54 @@
...
+    <div className="flex flex-col w-full mb-3">
+      {title && (
+        <label className="block text-sm font-medium text-gray-700">
+          {title}
+        </label>
+      )}
+      <label
+        className="flex items-center cursor-pointer mt-3"
+        onClick={handleClick}
+      >
```
goran-peoski-work: This might be a start of another long term debug/fix circle with the Toggle component while it already had support for `register` and the code is still there in the `staging` branch. It was already debugged and worked. Thus, there are options:

1. copy the original one and not give it the possibility of `onChange` prop and name it differently
2. copy the original into this one and make some if/else logic that ignores `onChange` if `register` is present
3. start as 1, but have this one copied as well, let's say `ToggleWithHandler`, `ToggleWithRegister` and make `Toggle` just do  

```tsx
   return props.register? <ToggleWithRegister {...props} /> : <ToggleWithHandler {...props} />
``` 


There could be options 4 and 5 and 6... but as this version goes, it looks to me like a footgun.

src/components/admin/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <div className="relative inline-block w-full text-gray-700">
+      <select
+        className="w-full h-10 pl-3 pr-6 text-base placeholder-gray-600 border rounded-lg appearance-none focus:shadow-outline"
+        placeholder="Regular input"
+        {...register(name)}
+      >
+        {options?.map((option) => (
+          <option
+            key={option.name}
+            className="bg-white w-full h-10 pl-3 pr-6 text-base placeholder-gray-600 border rounded-lg appearance-none focus:shadow-outline"
```
goran-peoski-work: These could be moved  to `Select.tailwind.ts` so the code is more concise in here

src/components/admin/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { WithId } from "utils/id";
+
+export interface Option {
+  name: string;
+  value: string;
+}
+
+type SelectProps = {
+  options: WithId<AnyVenue, string>[];
```
goran-peoski-work: Maybe this can be tried. Might show up errors, due to `string` and `SpaceId` not being the same time.

```suggestion
  options: SpaceWithId[];
```

src/components/admin/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+  options,
+  register,
+  name,
+  error,
+}) => {
+  return (
+    <div className="relative inline-block w-full text-gray-700">
+      <select
+        className="w-full h-10 pl-3 pr-6 text-base placeholder-gray-600 border rounded-lg appearance-none focus:shadow-outline"
+        placeholder="Regular input"
```
goran-peoski-work: I don't think this is the intended value, but a copy paste from an example. More likely the placeholder would be `name` if nothing better

src/components/admin/PortalAddEditForm.tsx
```diff
@@ -226,86 +223,95 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
...
+            title={title}
+            type="submit"
+          >
+            {portal ? "Save" : "Create"}
+          </Button>
+        </div>
+      </form>
+      {isPortalDeleteShown && (
+        <ConfirmationModal
+          header={"Delete portal"}
```
goran-peoski-work:
```suggestion
          header="Delete portal"
```

src/components/admin/PortalAddEditForm.tsx
```diff
@@ -226,86 +223,95 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
...
-          <ButtonNG
-            variant="danger"
+        <SubmitError error={submitError || deleteError} />
+        <div className="mt-5 sm:mt-4 sm:flex justify-end">
+          {isEditMode && (
+            <Button
+              variant="danger"
+              disabled={isLoading || isDeleting}
+              onClick={showPortalDelete}
+              className="mr-auto"
```
goran-peoski-work:
```suggestion
```

src/components/admin/ImageInput/ImageInput.tsx
```diff
@@ -40,10 +40,11 @@ export interface ImageInputProps {
   setValue: UseFormSetValue<AnyForm>;
   register: UseFormRegister<AnyForm>;
   nameWithUnderscore?: boolean;
-  text?: string;
   subtext?: string;
   // isInputHidden?: boolean;
   variant?: ImageInputVariant;
+  label?: string;
+  buttonLabel?: string;
```
goran-peoski-work: This is hard to follow. If there is a single one, `label` is good enough, but with tho, it's easier to not have to remember what the short one stands for, with something like `inputLabel` and `buttonLabel`

src/components/admin/Input/Input.tsx
```diff
@@ -62,6 +64,11 @@ export const Input: React.FC<InputProps> = ({
   ) : (
     <div className="Input">
       <div className={parentClasses}>
+        {label && (
+          <label className="block text-sm font-medium text-gray-700">
+            {label}
+          </label>
+        )}
```
goran-peoski-work: Please check the use of `<label>` tag through this PR. There are gotchas that need fixing:

- a `<label />` must be associated with an input
- a label will **at most** have one input associated, not two or three
- it can be associated if it wraps the input, like `<label ><input /> </label>`
- it can be associated by `id`, like  <label htmlFor={id}>


For the `id`, please use something like 
```tsx
  const id = useMemo(() => generateId("Input"), []);
```

goran-peoski-work (changes requested):
There are some gotchas to navigate with the changes of basic input components.

src/components/admin/Button/Button.tsx
alexxpetrov: removed [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Button/Button.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Button/Button.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/ConfirmationModal/ConfirmationModal.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/ConfirmationModal/ConfirmationModal.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/Dropdown/Dropdown.scss
alexxpetrov: moved to `admin` and replaced with attendee theme colors, but I've also added a `debt` comment and we should probably add another set of style variables for admin side in the future [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/PortalVisibility/PortalVisibility.scss
alexxpetrov: moved to admin [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/PortalVisibility/PortalVisibility.tsx
alexxpetrov: removed [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/PortalVisibility/PortalVisibility.tsx
alexxpetrov: done [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/PortalVisibility/PortalVisibility.tsx
alexxpetrov: added a debt comment [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/atoms/SpacesDropdown/SpacesDropdown.tsx
alexxpetrov: using host this way matches the `space` entity type and doesn't trigger TS errors

src/components/molecules/Modal/Modal.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Toggle/Toggle.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Toggle/Toggle.tsx
alexxpetrov: removed [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Toggle/Toggle.tsx
alexxpetrov: replaced some `label`s with simple `div`s that now include input's label [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Select/Select.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Select/Select.tsx
alexxpetrov: removed this component [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/PortalAddEditForm.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/PortalAddEditForm.tsx
alexxpetrov: [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/ImageInput/ImageInput.tsx
alexxpetrov: done, [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/Input/Input.tsx
alexxpetrov: replaced `label`s with `div`s [3469e91](https://github.com/sparkletown/sparkle/pull/2944/commits/3469e91693f840fa4d04232bc8c99b452b8f6b6d)

src/components/admin/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -51,14 +47,16 @@ export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
 
   return (
     <Modal show={isShown} onHide={hide}>
-      <div className="ConfirmationModal">
+      <div className="flex flex-col items-center">
```
goran-peoski-work: To help us visually navigate DOM with browser inspect, but also not interfere with styles, first prop can be made a BEM style value named `data-bem=`

```suggestion
      <div data-bem="ConfirmationModal" className="flex flex-col items-center">
```

This comment is valid for all instances like this or any other instances we might have bare elements and might be helpful to annotate as such.

src/components/admin/Dropdown/Dropdown.module.scss
```diff
@@ -0,0 +1,16 @@
+@use "scss/attendee/theme";
+
+// @debt: we might need another set of variables for admin, just like we have for attendee
+// to cover cases when tailwind classes are not enough
```
goran-peoski-work: > to cover cases when tailwind classes are not enough

Well... ☹️ 

[OOS]
It is not as much as having a set of constants, but having separation between the Tailwind "world" and the `.module.scss` one. 

This might put the problem in perspective: if user clicks links from admin that take them to attendee side and from attendee to admin side, both systems shouldn't end up loaded in the same DOM.

And in this case, users don't even need to navigate between admin/attendee sides, they buy one, get one free 🤪 

👉 This debt should be repaid soon, not forgotten❗

goran-peoski-work: [Q]
@alexey-petrov can we have a copy-paste duplicates of this Dropdown for admin and attendee sides but removing the extra styles?

If it's something simple, we can do a follow up PR to divorce the uses.

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -10,21 +11,21 @@ import { VenueTemplate } from "types/VenueTemplate";
...
-
 import "./SpacesDropdown.scss";
 
 const noneOptionName = "None";
 const spaceNoneOption = Object.freeze({
   id: "",
   name: "",
   template: undefined,
+  host: { icon: "" },
 });
```
goran-peoski-work, line 22:
```suggestion
// @debt define an interface for spaceNoneOption that also factors in the Space interface to keep both in lockstep
// NOTE: using host this way matches the space entity type and doesn't trigger TS errors
const spaceNoneOption = Object.freeze({
  id: "",
  name: "",
  template: undefined,
  host: { icon: "" },
});
```

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -133,7 +131,10 @@ export const SpacesDropdown: React.FC<SpacesDropdownProps> = ({
 
   return (
     <>
-      <div className="SpacesDropdown">
+      <div className="mb-2">
```
goran-peoski-work:
```suggestion
      <div data-bem="SpacesDropdown" className="mb-2">
```

src/components/admin/Toggle/Toggle.tsx
```diff
@@ -1,32 +1,54 @@
...
-  name,
+  register,
+  name = "",
+  title = "",
 }) => {
   // NOTE: the click handlers needs to be on the label to capture all its children clicks and there must always be a <label>
-  const handleClick = useCallback((event) => onChange?.(event), [onChange]);
+  const handleClick = useCallback(
+    (event) => {
+      !register && onChange?.(event);
```
goran-peoski-work, line 32:
```suggestion
      // NOTE: for simplicity of forms, register excludes the use of onChange 
      // however, should such need arise that both are needed and this check is removed, 
      // a regression that does double accounting should be handled by parents of this component
      !register && onChange?.(event);
```

goran-peoski-work (approved):
Looks good.

Please add the debt and NOTE comments before merge

src/components/admin/Toggle/Toggle.tsx
alexxpetrov: [1ae2f8b](https://github.com/sparkletown/sparkle/pull/2944/commits/1ae2f8b185da5056b37e5478d84f29d2daf6cce8)

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
alexxpetrov: [1ae2f8b](https://github.com/sparkletown/sparkle/pull/2944/commits/1ae2f8b185da5056b37e5478d84f29d2daf6cce8)

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
alexxpetrov: [1ae2f8b](https://github.com/sparkletown/sparkle/pull/2944/commits/1ae2f8b185da5056b37e5478d84f29d2daf6cce8)

src/components/admin/Dropdown/Dropdown.module.scss
alexxpetrov: this dropdown is not used at the `attendee` side at all. I've removed this `attendee` usage here and edited the debt comment saying that we need another set of admin side variables added in the future
for now, this style will be hardcoded as `white` instead of using attendee variable
[1ae2f8b](https://github.com/sparkletown/sparkle/pull/2944/commits/1ae2f8b185da5056b37e5478d84f29d2daf6cce8)

src/components/admin/ConfirmationModal/ConfirmationModal.tsx
alexxpetrov: [1ae2f8b](https://github.com/sparkletown/sparkle/pull/2944/commits/1ae2f8b185da5056b37e5478d84f29d2daf6cce8)

PR #2947, opened 2022-03-22
Fix admin footer position

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1751


goran-peoski-work (approved):
LGTM

PR #2948, opened 2022-03-22
Chat emoji icon placement fix

alexxpetrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1763

Before: 
image

After:
image


src/components/attendee/InputField/InputField.module.scss
```diff
@@ -0,0 +1,21 @@
+@use "scss/attendee/space";
+
+.inputFieldWrapper {
+  position: relative;
+}
```
sunny-viktoryia: do you use anywhere relative position of the wrapper? looks like you're using flex now, so I guess it's not needed?

sunny-viktoryia (approved):
Looks good to me. 🙃

src/components/attendee/InputField/InputField.module.scss
alexxpetrov: yep, you're right
removed, [8cad6f9](https://github.com/sparkletown/sparkle/pull/2948/commits/8cad6f9d0156b47d6c22bf2b03bbf3ceac0e2c40)

PR #2950, opened 2022-03-23
Replace AdminRestricted -> WithPermission

goran-peoski-work:
In general there is more work to be done on the hooks and Reactfire, 
but to avoid major issues, 
the decision is to add chunks of the merged-then-reverted PR 
- https://github.com/sparkletown/sparkle/pull/2894

piecemeal in order to catch culprits.

Also in this PR adding `<LoginRequired />` wrapper to the `<WorldUsers />`.
Some of the components that have `<WithPermission />` can in the future decite to use logic like 
```tsx
check={ creatingNew ? 'super' : 'world' }
```
or provide different loading and/or fallback messages that the old `<AdminRestricted />` didn't handle.


alexxpetrov (approved):
LGTM

PR #2953, opened 2022-03-25
Add mini profile modal

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1769

It also
* removes Badges
* fixes appearance of the UserProfile on the admin side

MiniProfile appears on click on the:
* navigation bar avatars
* chat message avatars
should be also added on Tables (but couldn't test it there, because tables are broken for me for some reason now)

image
image

TODO:
* fix the modal of myself in the chat (most of it is disappearing out of the right side of the screen) 
* handle out of modal click to close the Modal
-- I belive it makes sense to create a general modal component which will handle all those cases


src/components/atoms/UserAvatar/UserAvatar.module.scss
```diff
@@ -16,4 +17,6 @@
 .userAvatar {
   width: 100%;
   height: 100%;
+  border-radius: 50%;
```
goran-peoski-work:
```suggestion
    border-radius: border.$radius--mx;
```

src/components/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -84,29 +82,34 @@ export const _UserAvatar: React.FC<UserAvatarProps> = ({
...
+    openUserProfileModal,
+    closeUserProfileModal,
+    selectedElementId,
+  } = useProfileModalControls();
 
-  const statusIndicatorStyles = useMemo(
-    () => ({ backgroundColor: userStatus.color }),
-    [userStatus.color]
-  );
+  const [elementId] = useState<ElementId>(uniqueId("UserAvatar-") as ElementId);
```
goran-peoski-work: If it is the same functionality, I recommend using our own id generator, like in

```tsx
  const id = useMemo(() => generateId("NormalRadio-popover"), []);
```

otherwise, I'd add a `@debt` comment for unifying these two

src/components/atoms/UserAvatar/UserStatus.tsx
```diff
@@ -0,0 +1,54 @@
...
+    // @debt until temporarily disable is online functionality
+    // isOnline &&
+    showStatus &&
+    !!venueUserStatuses.length;
+
+  const statusIndicatorClasses = classNames("UserAvatar__status-indicator", {
+    "UserAvatar__status-indicator--online": isOnline,
+    [`UserAvatar__status-indicator--${status}`]: isOnline && status,
+    [`UserAvatar__status-indicator--${size}`]: size,
+  });
```
goran-peoski-work, line 47: [Q] Are these used for style purpose or are they just informative at this point?

src/components/attendee/Attendance/Attendance.tsx
```diff
@@ -16,11 +18,11 @@ export const Attendance: React.FC<AttendanceOptions> = ({
...
 
   const renderedAvatars = usersSample?.map((user) => (
-    <div key={user.id} className={CN.Attendance__userAvatar}>
-      <div className={CN.Attendance__userAvatarImage}>
-        <img src={user.pictureUrl} alt={`Avatar of ${user.partyName}`} />
-      </div>
-    </div>
+    <UserAvatar
+      key={user.id}
+      containerClassName={CN.Attendance__userAvatar}
```
goran-peoski-work, line 23: [OOS] we should be removing these gradually

src/components/attendee/Button/Button.tsx
```diff
@@ -5,32 +5,39 @@ import CN from "./Button.module.scss";
 
 // Button And Border variant types are the same to allow variant mixing.
 // But we might have different variants for either button or border in the future
-type ButtonVariant = "primary" | "alternative";
+type ButtonVariant = "primary" | "alternative" | "intensive" | "danger";
 type BorderVariant = ButtonVariant;
+type BorderRadius = "regular" | "none";
+type ButtonMargin = "regular" | "none";
```
goran-peoski-work: I think most of CSS uses 'normal' in this case. Maybe we can follow that example to make it more familiar

src/components/attendee/Button/Button.tsx
```diff
@@ -5,32 +5,39 @@ import CN from "./Button.module.scss";
 
 // Button And Border variant types are the same to allow variant mixing.
 // But we might have different variants for either button or border in the future
-type ButtonVariant = "primary" | "alternative";
+type ButtonVariant = "primary" | "alternative" | "intensive" | "danger";
```
goran-peoski-work, line 8: [NOTE] If there is intensive, I'd like to see what a depressed button looks like 🤪

src/scss/attendee/layout.scss
```diff
@@ -21,3 +21,7 @@ $avatarSizes: (
   large: 504px,
   xlarge: 1000px,
 );
+
+$cardSizes: (
+  small: 200px,
+);
```
goran-peoski-work, line 29: [Q] Is this intended to be expanded in the future with more values and used in more places?

src/components/shared/SpentTime/SpentTime.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const { isLoaded: isWorldLoaded, spaceId } = useWorldAndSpaceByParams();
+  const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
+  const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const firestore = useFirestore();
+
+  const fetchAllVenues = useCallback(async () => {
+    const userSnapshot = await getUserRef(userId).get();
+    const visitsSnapshot = await userSnapshot.ref.collection("visits").get();
```
goran-peoski-work: if `getUserRef()` returns the reference, there is no need for backtracking by `userSnapshot.ref`. It can be

```suggestion
    const userRef = getUserRef(userId);
    const userSnapshot = await userRef.get();
    const visitsSnapshot = await userRef.collection("visits").get();
```

src/components/shared/SpentTime/SpentTime.tsx
```diff
@@ -0,0 +1,128 @@
...
+  userId: UserId;
+}
+
+export const SpentTime: React.FC<SpentTimeProps> = ({ userId }) => {
+  const { isLoaded: isWorldLoaded, spaceId } = useWorldAndSpaceByParams();
+  const [visits, setVisits] = useState<WithId<UserVisit>[]>([]);
+  const [venues, setVenues] = useState<WithId<AnyVenue>[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const firestore = useFirestore();
```
goran-peoski-work: All this firestore code looks like it doesn't belong here. At least it can use a `@debt` comment to be moved elsewhere.

I would really like it if Reactfire isn't used new new code, since it's on its way to be removed.

There is a `getFirestore()` function, a `FIREBASE.firestore` singleton as well as `useFire*` and `useLive*` hooks to pick from.

src/components/shared/SpentTime/SpentTime.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { UserVisit } from "types/Firestore";
+import { UserId } from "types/id";
+import { AnyVenue } from "types/venues";
+
+import { WithId } from "utils/id";
+import { isTruthy } from "utils/types";
+
+import { useWorldAndSpaceByParams } from "hooks/spaces/useWorldAndSpaceByParams";
+
+import styles from "./SpentTime.module.scss";
```
goran-peoski-work: [NOTE] 

These aren't styles, these are classes, a name mapping from human readable to Webpack generated.

In other words it is an object of `Record<string, string>` type which is different from `CSSStyleDeclaration` one (https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration)

src/components/shared/SpentTime/SpentTime.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return (
+      <div data-bem="SpentTime--loading" className={styles.general}>
+        Loading...
+      </div>
+    );
+  }
+
+  if (!isWorldLoaded) {
+    return null;
+  }
```
goran-peoski-work: These can be done similar to https://github.com/sparkletown/sparkle/blob/redesign-2021/src/components/templates/ConversationSpace/ConversationSpace.hoc.tsx and maybe that would be a good start for the elsewhere comment above

src/hooks/useIsCurrentUser.ts
```diff
@@ -1,7 +1,7 @@
-import { useUser } from "./useUser";
+import { useUserId } from "./user/useUserId";
 
 export const useIsCurrentUser = (otherUserId?: string) => {
-  const { userId } = useUser();
+  const { userId } = useUserId();
```
goran-peoski-work, line 4: 👍 love it, narrowing of dependencies just like those `useSpaceParams()`

src/types/id.ts
```diff
@@ -14,6 +14,7 @@ export type UserId = Branded<string, "UserId">;
 export type WorldId = Branded<string, "WorldId">;
 export type WorldEventId = Branded<string, "WorldEventId">;
 export type WorldSlug = Branded<string, "WorldSlug">;
+export type ElementId = Branded<string, "ElementId">;
```
goran-peoski-work: Please keep them in alphabetic order since the list might grow even larger

src/components/molecules/DefaultAvatars/DefaultAvatars.tsx
```diff
@@ -22,24 +22,19 @@ export const DefaultAvatars: React.FC<DefaultAvatarsProps> = ({
...
     loading: isLoadingCustomAvatars,
   } = useAsync(async () => {
-    if (!sovereignVenueId) return;
+    if (!worldId) return;
 
     const storageRef = firebase.storage().ref();
-    const list = await storageRef
-      .child(`/assets/avatars/${sovereignVenueId}`)
-      .listAll();
+    const list = await storageRef.child(`/assets/avatars/${worldId}`).listAll();
```
goran-peoski-work, line 35:
```suggestion
    // @debt access to DB should be move out of components and into `src/api` and/or `src/hooks` layers
    const list = await storageRef.child(`/assets/avatars/${worldId}`).listAll();
```

src/components/attendee/MiniProfile/components/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import { compose } from "lodash/fp";
+
+import { Loading } from "components/molecules/Loading";
+
+import { MiniProfileModal as _MiniProfileModal } from "./MiniProfileModal";
+
+export const MiniProfileModal = compose(
+  withProfileById,
+  withFallback(["isProfileByIdLoaded"], Loading)
+)(_MiniProfileModal);
```
goran-peoski-work: Please try to rework this like https://github.com/sparkletown/sparkle/blob/redesign-2021/src/components/templates/ConversationSpace/ConversationSpace.hoc.tsx 

The only difference is I'd not bother with doing interface extension and exclusion. Just a simple one would be good enough.


It will be easier for following/debugging the logic and for possible future Storyboard display of MoniProfileModal

src/components/attendee/MiniProfile/components/MiniProfileModalLinks.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const MiniProfileModalLinks: React.FC<MiniProfileModalLinksProps> = ({
+  profileLinks,
+}) => {
+  const renderedProfileModalLinks = useMemo(
+    () =>
+      profileLinks?.map((link: ProfileLink) => {
+        const linkIcon = getProfileModalLinkIcon(link.url);
+
+        return (
+          <a key={link.url} href={link.url} target="_blank" rel="noreferrer">
```
goran-peoski-work: This might also be a security risk, hard-coding `noreferrer` and not having `noopener`. It would be better if all these cases refer to the same source:

```ts
export const externalUrlAdditionalProps = {
  target: "_blank",
  rel: "noopener noreferrer",
};
```

src/components/attendee/MiniProfile/components/MiniProfileModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const actionButton = isCurrentUser ? (
+    <Card.Button variant="danger" onClick={logout}>
+      Log Out
+    </Card.Button>
+  ) : (
+    <Card.Button variant="intensive" onClick={openChat}>
+      Send message
+    </Card.Button>
+  );
```
goran-peoski-work: Most of this code is the same, it would be easy to follow it if it's not put into a variable. The `onClick` and inside text can use the `? :` treatment.

src/components/attendee/MiniProfile/components/MiniProfileModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { selectRecipientChat } = useChatSidebarControls();
+
+  const { closeUserProfileModal } = useProfileModalControls();
+
+  const openChat = useCallback(() => {
+    selectRecipientChat(profile);
+    closeUserProfileModal();
+  }, [selectRecipientChat, profile, closeUserProfileModal]);
+
+  const logout = useCallback(async () => {
```
goran-peoski-work: [Q] what will happen on rejection? would it be more useful to have `useAsyncFn` here and deal with the `error` field?

src/components/attendee/MiniProfile/MiniProfile.tsx
```diff
@@ -0,0 +1,20 @@
+import { ElementId } from "types/id";
+
+import { useProfileModalControls } from "hooks/useProfileModalControls";
+
+import { MiniProfileModal } from "./components";
+
+interface MiniProfileProps {
+  parentComponent: ElementId;
+}
+export const MiniProfile: React.FC<MiniProfileProps> = ({
```
goran-peoski-work, line 10: This is basically a HOC, just in case one is needed

src/components/attendee/MiniProfile/MiniProfile.tsx
```diff
@@ -0,0 +1,20 @@
...
+  parentComponent,
+}) => {
+  const { selectedUserId, selectedElementId } = useProfileModalControls();
+
+  if (!(selectedElementId && selectedElementId === parentComponent)) {
+    return <></>;
+  }
+
+  return <MiniProfileModal userId={selectedUserId} />;
```
goran-peoski-work: I will list few options, in case any one or combination of few makes the code here more readable

option 1: ` ! ( A && B === C )` can be converted to ` A || B !== C`

option 2: The `if-else` can be converted to `? :`

option 3: 
```ts
if ( ! condition ){
  // something
} else {
  // something else
}
```

can be converted into

```tsx
if ( condition ){
  // something else
} else {
  // something something
}
```

src/components/attendee/Card/CardButton.tsx
```diff
@@ -0,0 +1,9 @@
+import { Button, ButtonProps } from "components/attendee/Button";
+
+export const CardButton: React.FC<ButtonProps> = ({ children, ...rest }) => {
+  return (
+    <Button {...rest} borderRadius="none" margin="none">
```
goran-peoski-work: [NOTE]
I cant say if and how it will reflect this case, might be the above line is quite OK, so I will just note that generally, for DX reasons:

- margins and basically all of layout would be the concern of the parent, not the components inside. Flexbox for example can use `padding`, `gap` etc.

- some times props that work like booleans can be named `marginless` or `rounded` depending on the case itself

goran-peoski-work (changes requested):
Looks good.

I would like it if another pass has been done and maybe some suggestions implemented or if time forbidding noted with `@debt` comments and done in follow up PRs

src/components/atoms/UserAvatar/UserStatus.tsx
sunny-viktoryia: they're informative now, I believe, because we don't have those class names in the stylesheet.
By the way, it's a question if we're going to use user statuses at all.

src/components/shared/SpentTime/SpentTime.tsx
sunny-viktoryia: `isLoading` is actively used in the component, if we move all the logic to hook, it'll be almost the same as the component now but without the HTML. Once we rewrite the logic of calculating online hours, then we can refactor that I think.

src/components/atoms/UserAvatar/UserAvatar.module.scss
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/atoms/UserAvatar/UserAvatar.tsx
sunny-viktoryia: haven't payed attention to that generator earlier, thanks!
replaced: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/Button/Button.tsx
sunny-viktoryia: created `rounded` and `marginless` props instead like you suggested in another comment: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/scss/attendee/layout.scss
sunny-viktoryia: yep, added a `debt` comment above the variable: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/shared/SpentTime/SpentTime.tsx
sunny-viktoryia: you're right, the logic could be simplified a bit, made it one-line. Done here: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/shared/SpentTime/SpentTime.tsx
sunny-viktoryia: added a debt comment. It was copied from the `Badges` component. We can refactor it when needed. https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/types/id.ts
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/molecules/DefaultAvatars/DefaultAvatars.tsx
sunny-viktoryia: added: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/components/index.ts
sunny-viktoryia: created a hoc here: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/components/MiniProfileModalLinks.tsx
sunny-viktoryia: oh yes, thanks! Fixed: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/components/MiniProfileModal.tsx
sunny-viktoryia: I wouldn't say it would look better. As instead of 1 ternary operator, there would be 3 of them. But if you don't like the usage of `actionButton` const, I moved the ternary operator directly to the HTML part. Hopefully, it'll look more readable.
https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/components/MiniProfileModal.tsx
sunny-viktoryia: changed: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/MiniProfile.tsx
sunny-viktoryia: used the 1st option: https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/Card/CardButton.tsx
sunny-viktoryia: added `marginless` or `rounded` properties:
https://github.com/sparkletown/sparkle/pull/2953/commits/a52527d7fce74b5f15447723260cc7715c8614b3

src/components/attendee/MiniProfile/MiniProfile.tsx
sunny-viktoryia: MiniProfileModal is also a hoc, and I wouldn't like to join them together as it'd make all the modals to fetch the info about the users on page load. I suggest leaving it as is.

src/components/shared/SpentTime/SpentTime.tsx
sunny-viktoryia: renamed the constants here: https://github.com/sparkletown/sparkle/pull/2953/commits/9567a98c837f8f4ff16dfba4f38bb70ea2929ab8

src/components/atoms/UserAvatar/UserStatus.tsx
goran-peoski-work: Delete then? Put `data-bem`? Ignore as OOS? 🤷

src/components/attendee/MiniProfile/MiniProfile.tsx
goran-peoski-work: > I suggest leaving it as is.

Yeah, my bad, I forgot to put [NOTE] there. It does look like OOS

src/components/attendee/Button/Button.tsx
```diff
@@ -28,16 +27,16 @@ export const Button: React.FC<ButtonProps> = ({
   transparent,
   forwardRef,
   border,
-  borderRadius,
-  margin,
+  rounded = true,
```
goran-peoski-work: This might need tweaking I guess. `unrounded` works as both missing (`undefined`) and present (`true`) :/ Either way, we can revisit some DX changes later on, see what worked, what didn't.

src/components/shared/SpentTime/SpentTime.tsx
goran-peoski-work: I was hoping to at least have that `useFirestore()` replaced by `getFirestore()` so there isn't an import from `Reactfire`.

goran-peoski-work (approved):
LGTM

src/components/attendee/Button/Button.tsx
sunny-viktoryia: changed the prop: https://github.com/sparkletown/sparkle/pull/2953/commits/d110155d478fa8fcec0c3dce7443ed2d4b871a01

src/components/shared/SpentTime/SpentTime.tsx
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/2953/commits/d110155d478fa8fcec0c3dce7443ed2d4b871a01

PR #2954, opened 2022-03-25
Old login redesign pre-styling clean up

goran-peoski-work:
Mostly moving the old components into a single place that will be easy to remove later on after new Login flow is created.

There is slight addition / transfer of the `logout()` function so that attendee can do `SPARK.logout()` in their browser console.


src/core/spark.ts
```diff
@@ -0,0 +1,31 @@
...
+  BUILD_PULL_REQUESTS,
+  BUILD_REPOSITORY_URL,
+  BUILD_SHA1,
+  BUILD_TAG,
+} from "secrets";
+
+import { FIREBASE } from "./firebase";
+
+const logout = () =>
+  void FIREBASE.auth.signOut().then(console.log).catch(console.error);
```
alexxpetrov: weird console usages if you ask me

alexxpetrov (approved):
LGTM, note the console tho

src/core/spark.ts
goran-peoski-work: Yeah, I meant to replace it with more elaborate one, like the capture of error

goran-peoski-work: Done 4425bc5465061b654e0802f90fd88d3e4d895b70

PR #2955, opened 2022-03-25
fixed schedule double scroll + weekday render + title

alexxpetrov:
Closes:
- https://docs.google.com/spreadsheets/d/1MRbvI1m_OmDJpUgIWztmnjKV1HGOPvafte-Lkf6pnYY/edit#gid=891834841

Lines 22, 25


src/hooks/viewport/useDisableBodyScroll.tsx
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from "react";
+
+export const useDisableBodyScroll = (isOpen: boolean) => {
+  useEffect(() => {
+    const [layoutElement] = document.getElementsByClassName(
+      "AttendeeLayout"
```
goran-peoski-work: This is a hardcoded string used in other place as well. Most likely it needs to be a shared constant (or this hook produce it for the component that uses it).

src/components/attendee/AttendeeLayout.tsx
```diff
@@ -32,9 +32,9 @@ const _AttendeeLayout: React.FC<_AttendeeLayoutProps> = ({ userId }) => {
   const [backButtonSpace, setBackButtonSpace] = useState<SpaceWithId>();
 
   useEffect(() => {
-    document.documentElement.classList.add(styles.html);
+    document.documentElement.classList.add("AttendeeLayout", styles.html);
```
goran-peoski-work: I don't think there is a need of adding more classes to the body element (a god citizen of app would just not get wider from the screen). 

The idea here is if there is a component that goes wider than the parent component, the parent component should display scroll bars, not rely on the body element to do it.

src/hooks/useVenueScheduleEvents.ts
```diff
@@ -104,7 +104,7 @@ const useVenueScheduleEvents = ({
   );
 
   // +1 to include the latest day in the schedule (for example, there are events tomorrow and today -> tomorrow - today + 1 = 2 days)
-  const dayDifference = daysInBetween + 1;
+  const dayDifference = daysInBetween + 1 || 0;
```
goran-peoski-work: [NOTE] Not really necessary to change this code, maybe just add a line of comment saying `undefined + 1 is NaN which is falsy`.

On other hand, depending on `NaN` being falsy might not be obvious to everyone why it works.Can it be something like `isDefined()` check or `?? 0` ? Just a question, would keep the logic as is.

goran-peoski-work (changes requested):
I hope there is another way of dealing with style issues than changing the body element.

src/hooks/useVenueScheduleEvents.ts
alexxpetrov: added a comment + converted to ternary [80e0fc3](https://github.com/sparkletown/sparkle/pull/2955/commits/80e0fc3d5a99f391b995b7e6baf0a44ba5a0ec0f)

src/hooks/viewport/useDisableBodyScroll.tsx
alexxpetrov: i've added a const and reused it [80e0fc3](https://github.com/sparkletown/sparkle/pull/2955/commits/80e0fc3d5a99f391b995b7e6baf0a44ba5a0ec0f)

src/components/attendee/AttendeeLayout.tsx
alexxpetrov: I don't quite get your comment here. As far as I understand the only way to remove scroll from the body element, when another modal is up, is to remove the `overflow` off it, so here's why I'm adding this classname to use in the hook I've added in this PR

goran-peoski-work (approved):
LGTM

PR #2960, opened 2022-03-28
Meeting rooms

colinhowe:
Introduces meeting rooms and retunable media elements.
image

This also adds better diagnostics around twilio for future debugging


src/components/attendee/VideoComms/internal/TwilioTrackDisplay.tsx
```diff
@@ -25,8 +25,9 @@ export const TwilioTrackDisplay: React.FC<TwilioTrackDisplayProps> = ({
   const videoRef = useRef<HTMLVideoElement>(null);
   useEffect(() => {
     if (videoRef.current) {
+      const elForCleanup = videoRef.current;
       track.attach(videoRef.current);
-      return () => void track.detach();
+      return () => void track.detach(elForCleanup);
     }
```
colinhowe: The call to `detach` without any arguments will detach the Twilio video stream from **all** video elements. That's not what we want at all.

src/components/attendee/WebcamGrid/WebcamGrid.tsx
```diff
@@ -24,26 +24,15 @@ export const WebcamGrid: React.FC<TableGridProps> = ({ space }) => {
...
-        enableAudio: false,
-        enableVideo: false,
-      });
-      return () => {
+    return () => {
+      if (hasJoined) {
         disconnect();
-      };
-    }
-  }, [disconnect, joinChannel, space.id, userId]);
```
colinhowe, line 46: This change turns off the auto-joining behaviour which has been problematic. We can come back and do a more complex fix to this if the product team want us to.

src/hooks/useAnalytics.ts
```diff
@@ -40,7 +40,7 @@ export const setAnalyticsGroup = (groupKey: string, groupName: string) => {
 
 export const initAnalytics = (opts?: Object) => {
   if (!MIXPANEL_PROJECT_TOKEN) {
-    console.warn("Mixpanel is not set up correctly. The token is missing.");
```
colinhowe, line 43: This being at warning level led to a lot of stack traces in the console which made it hard to debug things that were genuinely broken.

src/components/attendee/MediaElement/MediaElement.module.scss
```diff
@@ -61,6 +61,13 @@
   height: 100%;
   @include rounded.normal;
   overflow: hidden;
+
+  video {
```
goran-peoski-work, line 65: [NOTE]
I'm guessing the implicit assumption here is that each media element will have one and only one `video` element.

src/components/attendee/MediaElement/MediaElement.tsx
```diff
@@ -47,16 +52,19 @@ export const MediaElement: React.FC<MediaElementProps> = ({
...
           <iframe
             title="main event"
             className={styles.iframe}
             src={embedIframeUrl}
             frameBorder="0"
             allow={IFRAME_ALLOW}
           />
-        ) : (
-          <div>Embedded Video URL not yet set up</div>
+        )}
```
goran-peoski-work, line 63: [Q] 
Is there a design for UI/UX to display any kind of feedback to user about this?
Is it now part of another component?

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [isTuning, setIsTuning] = useState(false);
+  const startTuning = useCallback(() => setIsTuning(true), [setIsTuning]);
+  const stopTuning = useCallback(() => setIsTuning(false), [setIsTuning]);
+  const { isLoading, settings } = useRetunableMediaElement({
+    spaceId: space.id,
+  });
+  const { localParticipant, remoteParticipants } = useVideoComms();
+
+  const mediaElement = useMemo(() => {
+    switch (settings.sourceType) {
```
goran-peoski-work, line 36: [code smell]

`switch` statements are usually a half-way between simpler and more robust solutions, that themselves fail to achieve.

The simpler one is reading from a key-value pair, a simple object or a map (in this case `Record<RetunableMediaSource, ReactNode``. This should make code more concise and easier to follow. 

The above is short way of making `RetunableMediaElement` be the facade (design pattern) object.

The facade is basically the more robust solution, a factory function to replace `switch` and return the correct element based on internal logic - the `if`  statements embedded under some `case` statements.

The latter solution also serves as going towards the single responsibility principle.

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
```diff
@@ -0,0 +1,90 @@
...
+          // left the call.
+          return <></>;
+        }
+
+        return <MediaElement track={videoTrack} autoPlay />;
+    }
+  }, [localParticipant, remoteParticipants, settings]);
+
+  if (isLoading) {
+    return <></>;
```
goran-peoski-work, line 86: If there are design UI/UX considerations for displaying feedback to user, this could use a debt/todo comment

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import { TuneBanner } from "./components/TuneBanner";
+import { Tuner } from "./components/Tuner";
+import { useRetunableMediaElement } from "./hooks";
+import { RetunableMediaSource } from "./types";
+
+import styles from "./RetunableMediaElement.module.scss";
+
+interface RetunableMediaElementProps {
+  space: WithId<AnyVenue>;
```
goran-peoski-work: helps reduce the number of direct imports of `WithId` in a gradual step towards moving it out of `utils/id`

```suggestion
  space: SpaceWithId;
```

src/components/attendee/RetunableMediaElement/components/TuneBanner.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Button } from "components/attendee/Button";
+
+import styles from "./TuneBanner.module.scss";
+
+interface TuneBannerProps {
+  isTuning: boolean;
+  startTuning: () => void;
+}
+
+export const TuneBanner = ({ isTuning, startTuning }: TuneBannerProps) => {
```
goran-peoski-work: `export const TuneBanner: React.FC<TuneBannerProps> = ({ isTuning, startTuning }) => {`

src/components/attendee/RetunableMediaElement/components/Tuner.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  background-color: theme.$c-tuner-background-color;
+  padding: space.empty(2);
+  margin: 0 auto space.empty(1) auto;
+  width: fit-content;
+  position: relative;
+
+  label {
+    display: grid;
+    grid-template-columns: 1em auto;
+    gap: 0.5em;
```
goran-peoski-work: there is `space.gap()` for this case

src/components/attendee/RetunableMediaElement/components/Tuner.module.scss
```diff
@@ -0,0 +1,75 @@
+@use "scss/attendee/space.scss";
+@use "scss/attendee/theme.scss";
+
```
goran-peoski-work, line 3: This file has a number of hard-coded values as `rem`, `em` and `px`. I think they could be relative to some constant(s), even if only local to this module.

src/api/retunableMediaElement.ts
```diff
@@ -0,0 +1,15 @@
+import { RetunableMediaElementSettings } from "components/attendee/RetunableMediaElement/types";
+import firebase from "firebase/compat/app";
+
+import { COLLECTION_RETUNABLE_MEDIA_ELEMENTS } from "settings";
+
+export const setRetunableMediaSettings = async (
+  spaceId: string,
+  settings: RetunableMediaElementSettings
+) => {
```
goran-peoski-work: This might help for more uniformed interface with other functions that try to make it easier to expand with new options and/or ignore old ones

```suggestion

type SetRetunableMediaSettings = ( options: {
  spaceId: SpaceId,
  settings: RetunableMediaElementSettings
}) => Promise<void>;

export const setRetunableMediaSettings: SetRetunableMediaSettings  = async ({
  spaceId,
  settings,
}) => {
```

src/components/attendee/RetunableMediaElement/components/Tuner.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import { setRetunableMediaSettings } from "api/retunableMediaElement";
+
+import { AnyVenue } from "types/venues";
+
+import { WithId } from "utils/id";
+
+import { useUserId } from "hooks/user/useUserId";
+
+import { RetunableMediaSource } from "../types";
```
goran-peoski-work: [code smell]
Relative import to parent directory. Usually it's only `./` and absolute ones.

src/components/attendee/RetunableMediaElement/types.tsx
```diff
@@ -0,0 +1,38 @@
+export enum RetunableMediaSource {
```
goran-peoski-work: [Q] Why is this file `.tsx`?

It might be a good idea to move this file in `./src/types` since it is meant to be shared across hooks and/or components.

Another suggestion, considering other component-local files in the code base, would be a rename to `RetunableMediaElement.types.ts`

src/components/attendee/RetunableMediaElement/types.tsx
```diff
@@ -0,0 +1,38 @@
...
+  notTuned = "NOT_TUNED",
```
goran-peoski-work: Maybe something simpler like

```suggestion
  none= "NONE",
```

or `missing`, `empty` or anything without negative?

Would a `default` have some logic behind it?

src/components/attendee/VideoComms/internal/TwilioClient.tsx
```diff
@@ -132,6 +132,27 @@ export const TwilioClient = (onStateUpdateCallback: StateUpdateCallback) => {
   let isTransmittingAudio = true;
   let isTransmittingVideo = true;
 
+  // @debt remove this. It is here to help with debugging until this has been
+  // battle tested
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  window.getTwilioInfo = () => {
```
goran-peoski-work, line 139: There is now a global we can use 
https://github.com/sparkletown/sparkle/blob/redesign-2021/src/core/spark.ts

And I think by adding a type to it as `Record<string, function|object|undefined>` we can add these kinds of utilities. Example:

`SPARK.twilio.getInfo = () => `

and it might make it easy to find them in code / reuse them in console / etc.

src/components/attendee/VideoComms/internal/TwilioClient.tsx
```diff
@@ -132,6 +132,27 @@ export const TwilioClient = (onStateUpdateCallback: StateUpdateCallback) => {
...
+    const localParticipant = room && wrapLocalParticipant(room);
+    const remoteParticipants =
+      room && Array.from(room.participants.values()).map(wrapRemoteParticipant);
+    return {
+      room,
+      status,
+      isTransmittingAudio,
+      isTransmittingVideo,
+      // Data that has been transformed for use in React
+      toReact: {
```
goran-peoski-work, line 149: Name sounds a lot like `toString`, `toJSON` so it might be mistaken for function. 

Will just naming it `react` be a problem? Is it usable only by react? I mean, if the types aren't `ReactNode[]` it can just be named `participants` and have fields `local` and `remote` 🤷

src/components/attendee/VideoComms/internal/TwilioTrackDisplay.tsx
```diff
@@ -25,8 +25,9 @@ export const TwilioTrackDisplay: React.FC<TwilioTrackDisplayProps> = ({
   const videoRef = useRef<HTMLVideoElement>(null);
   useEffect(() => {
     if (videoRef.current) {
+      const elForCleanup = videoRef.current;
```
goran-peoski-work, line 28: I don't think another name for `videoRef.current` is really necessary here.

src/components/attendee/VideoComms/internal/TwilioTrackDisplay.tsx
goran-peoski-work: @colinhowe Please add this comment inside the code prefixed with something like `// NOTE: `

src/components/attendee/WebcamGrid/WebcamGrid.tsx
goran-peoski-work: I propose ` if( !DISABLED_DUE_TO_AUTOJOIN_PROBLEMS && userId)` check before joining instead of outright deleting this code. The explanation can be put in short or long form in `disableSettings.ts` (the primary reason for that file existence)

src/hooks/useAnalytics.ts
goran-peoski-work: [NOTE]
I think the cleaner solution is to change from firing regularly, to only once per load. I wonder if we should include that in the `SPARK` object and if warning/info has already been shown, not spam it again and again.

src/pages/VenuePage/TemplateWrapper.tsx
```diff
@@ -107,6 +108,10 @@ export const TemplateWrapper: React.FC<TemplateWrapperProps> = ({ venue }) => {
       template = <ScreeningRoom venue={venue} />;
       break;
 
+    case VenueTemplate.meetingroom:
```
goran-peoski-work, line 112: [NOTE]
I think this `TemplateWrapper` should incrementally be deconstructed by removing concerns out of it and ultimately replaced by a simple `Record<VenueTemplate, ReactNode>` in `VenuePage.tsx`

src/scss/attendee/theme.scss
```diff
@@ -29,6 +29,9 @@
 
   --toggler-off-color: rgba(255, 255, 255, 0.3);
 
+  --button-disabled-background: rgb(183, 169, 217);
```
goran-peoski-work, line 37: [NOTE] Once `theme.scss` gets sufficiently large enough, we can analyze it and see if so many "colors" are needed (there are different alpha versions of the same color, as an example)

src/settings/apiSettings.ts
```diff
@@ -19,6 +19,7 @@ export const COLLECTION_WORLDS = "worlds";
 export const COLLECTION_WORLD_EVENTS = "worldEvents";
 export const COLLECTION_SEATED_USERS = "seatedUsers";
 export const COLLECTION_SEATED_USERS_CHECKINS = "recentSeatedUsers";
+export const COLLECTION_RETUNABLE_MEDIA_ELEMENTS = "retunableMediaElements";
```
goran-peoski-work, line 22: [NOTE]
We should've kept this list alphabetically sorted for easier visual search. I have failed to add a comment at the top like in other places.

src/components/attendee/RetunableMediaElement/hooks.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: useRetunableMediaElementDataOptions) => {
+  const queryPath = useMemo(
+    () => [COLLECTION_RETUNABLE_MEDIA_ELEMENTS, spaceId],
+    [spaceId]
+  );
+
+  const {
+    data: settings,
+    isLoading,
+  } = useRefiDocument<RetunableMediaElementSettings>(queryPath);
```
goran-peoski-work: Please try to replace it with `useFireDocument` to lessen the dependence of Reactfire. The alternative should also help with missing IDs by way of `DEFERRED`.

In case of problems with the suggested hook, please just add `@debt` comment here.

src/components/attendee/RetunableMediaElement/hooks.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  if (isLoading) {
+    return {
+      isLoading,
+      settings: NOT_TUNED_SETTINGS,
```
goran-peoski-work: I think just freezing this object will not prevent this hook triggering re-renders in cases it's output isn't deconstructed.

I think it's one of those "better be safe than sorry" to add `useMemo()` and do these `if` checks inside it.

src/components/attendee/RetunableMediaElement/hooks.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { useRefiDocument } from "hooks/fire/useRefiDocument";
+
+import {
+  NotTunedSettings,
+  RetunableMediaElementSettings,
+  RetunableMediaSource,
+} from "./types";
+
+interface useRetunableMediaElementDataOptions {
```
goran-peoski-work:
```suggestion
interface UseRetunableMediaElementDataOptions {
```

But I would go for a more holistic approach of `UseRetunableMediaElement ` type for the hook that has these options and returns ` {} & LoadStatus `

goran-peoski-work (changes requested):
Looks good. I hope it would benefit from another pass on the code, examine if some quick changes can be made before merging and leave some NOTE/TODO/@debt comments in cases it might be OOS

src/components/attendee/MediaElement/MediaElement.module.scss
colinhowe: yep

src/components/attendee/MediaElement/MediaElement.tsx
colinhowe: There's no design for it. I'll make a note to raise with design team

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
colinhowe: Generally, I agree with the map for factories. However, in this case we're using the type information from `settings.sourceType` to allow each case block to use the more specific type. I can't immediately see a way of getting Typescript to give us this in a map (I tried). I'll make a comment that this could be improved at a later date.

src/components/attendee/RetunableMediaElement/components/Tuner.tsx
colinhowe: I'm happy with this. The folder `Tuner` in is intended to be components used purely for the rendering of `RetunableMediaElement`. It knowing the relative location of types specific to this component is fine

src/components/attendee/RetunableMediaElement/types.tsx
colinhowe: I think it's clearer that the default case is in line with what is displayed on the UI

src/components/attendee/VideoComms/internal/TwilioClient.tsx
colinhowe: I'm expecting to delete this entirely in a week or two

src/components/attendee/VideoComms/internal/TwilioTrackDisplay.tsx
colinhowe: Making ESLint happy. It doesn't **know** that the callback will be invoked before the reference changes value

src/components/attendee/WebcamGrid/WebcamGrid.tsx
colinhowe: I'm not adding more flakes and more conditionals. If the product team want this it'll need fixing properly.

src/api/retunableMediaElement.ts
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/components/TuneBanner.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/components/Tuner.module.scss
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/components/Tuner.module.scss
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/hooks.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/hooks.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/hooks.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/RetunableMediaElement/types.tsx
colinhowe: Resolved in 40c9750b0

src/components/attendee/VideoComms/internal/TwilioTrackDisplay.tsx
colinhowe: Resolved in 40c9750b0

goran-peoski-work (approved):
LGTM

src/components/attendee/MediaElement/MediaElement.tsx
goran-peoski-work: I like the idea from another comment about a "slow load spinner" and I think it's easy to create (or update existing loading components)

PR #2962, opened 2022-03-28
Fix chat trying to read id of undefined

goran-peoski-work:
Previously chat wasn't checking the `isLoading` and `isLoaded` booleans before trying to go forward and read messages (and their IDs) resulting in type error reading from `undefined` values.

This PR revamps the HOC with easier to follow logic and updates the `useWorldAndSpaceBySlug` hook for more sensible error handling.


alexxpetrov (approved):
LGTM

PR #2963, opened 2022-03-28
Fix appearance of the space with not-whilisted template

sunny-viktoryia:
fixes spaces dashboard page for the worlds which have talkshowstudio template


mike-lvov (approved):
LGTM

PR #2965, opened 2022-03-29
redesign run tab and add announcement form

denisdimitrov:
- Redesign run tab
- Add announcement form that updates the banner
- Reduce the size of toggle component

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1777


denisdimitrov:
[Note] There is a cancel button in the design and page but there is nothing to be cancelled, really. It's not a modal, it's a static page form.

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
```diff
@@ -1,60 +1,163 @@
...
 import { MapPreview } from "../MapPreview";
 
 export interface RunTabViewProps {
   space?: SpaceWithId;
 }
 
 export const RunTabView: React.FC<RunTabViewProps> = ({ space }) => {
+  const defaultValues = useMemo(
+    () => ({
+      content: space?.banner?.content ?? "",
```
goran-peoski-work: maybe a

`const banner = space?.banner` 

before `useMemo` to improve on readability.

Not sure if de-structuring every field would be pro or con, but we can probably shave off the `space` in front of every access.

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
```diff
@@ -1,60 +1,163 @@
...
+            Save
+          </Button>
+        </div>
+      </AdminSidebar>
+
+      <AdminShowcase>
+        <MapPreview
+          isEditing={false}
+          mapBackground={mapBackground}
+          rooms={[]}
```
goran-peoski-work: Not sure if this array is supposed to be changed by `MapPreview`. It may be clearer if `rooms` can accept `undefined` or at least `[]` is replaced with `ALWAYS_EMPTY_ARRAY` to prevent re-renders.

src/components/organisms/AdminVenueView/components/RunTabView/RunTabView.tsx
```diff
@@ -1,60 +1,163 @@
...
+          setSelectedRoom={() => undefined}
```
goran-peoski-work: There is also `ALWAYS_NOOP_FUNCTION` for cases like this

goran-peoski-work (approved):
LGTM.

Gave some "nice to have" suggestions.
I don't think `rooms={[]}` will cause a bug, but it can be made safer.

PR #2967, opened 2022-03-29
add global time panel to space schedule

denisdimitrov:
- Adds global time panel to the space schedule in the edit space tab.
- Renames `timing` to `schedule`
- Extracts classes in `tailwind.ts` file

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1783


src/components/admin/SpaceSchedule/SpaceSchedule.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <div className={TW.container}>
+      <div className={TW.worldSchedulePanel}>
+        <div className={TW.worldScheduleSection}>
+          <div className={TW.worldScheduleTitle}>
+            Global start date and time
+          </div>
+          <div className={TW.worldScheduleDateTime}>
+            <div className={TW.worldScheduleDate}>{`${formatDate(
+              new Date(startDate)
+            )},`}</div>
```
goran-peoski-work, line 74: Must this be inside a template string? I'd think `formatDate` would return a string.

Also, [NOTE] to self: we should have few convenience date functions, maybe one that clones a date like the above but with easily recognizable name.

goran-peoski-work: Same comment for the next field

src/components/admin/SpaceSchedule/SpaceSchedule.tsx
```diff
@@ -0,0 +1,119 @@
...
+            Edit in World schedule
+          </div>
+        </div>
+      </div>
+
+      <div className={TW.eventsPanel}>
+        <div className={TW.createEventButtonWrapper}>
+          <Button onClick={showCreateEventModal}>Create new experience</Button>
+        </div>
+        <EventsPanel worldId={space.worldId as WorldId} spaces={[space]} />
```
goran-peoski-work, line 107: [NOTE] This is an interesting use of props: 

every space in `spaces` should already have `worldId` and `EventsPanel` doesn't control if all of them have the same `worldId`, but then `worldId` prop is also passed in.

goran-peoski-work (approved):
LGTM

PR #2968, opened 2022-03-30
Delete unused code

colinhowe:
Lots of deletions, all broken up into individual commits


src/utils/time.ts
```diff
@@ -347,5 +247,3 @@ export const getDateHoursAndMinutes = (
 ): string => format(dateOrTimestamp, DATEFNS_INPUT_TIME_FORMAT);
 
 export const currentMilliseconds = () => new Date().getTime();
-export const currentIso = () => new Date().toISOString();
-export const currentUtc = () => new Date().toUTCString();
```
goran-peoski-work, line 351: I added these 3 so that people wouldn't type `new Date()` all the time because of it's clunky interface. I think even if unused now, they're good alternative to constantly re-inventing the same when the need arises.

goran-peoski-work (approved):
LGTM

I like deleting legacy stuff. I just left one comment for keeping 2 functions that are "just in case" type

PR #2969, opened 2022-03-30
Introduce attendee popover component

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1748

* uses SpaceId instead of string in types
* adds basic info to the Portal popover
image

* fixes the layout of MiniProfiles

BEFORE
image

AFTER
image

* minor UI fixes on the way


src/components/admin/PortalAddEditForm.tsx
```diff
@@ -71,20 +72,21 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
       title: portal?.title ?? "",
       image_url: portal?.image_url ?? icon ?? spaceLogoImage,
       visibility: portal?.visibility ?? venueVisibility,
-      spaceId: portal?.spaceId ?? undefined,
       isClickable: portal?.type !== RoomType.unclickable,
       isEnabled: portal?.isEnabled ?? DEFAULT_PORTAL_IS_ENABLED,
+      // @debt we should use SpaceId type here; couldn't resolve error with form typing
```
goran-peoski-work:
```suggestion
      // @debt should use SpaceId type here, resolve error with form typing
```

src/components/admin/SpaceEditForm.tsx
```diff
@@ -105,7 +105,8 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
       columns: space.columns ?? 0,
       autoPlay: space.autoPlay ?? DEFAULT_VENUE_AUTOPLAY,
       isReactionsMuted: space.isReactionsMuted ?? DEFAULT_REACTIONS_MUTED,
-      parentId: space.parentId ?? "",
+      // @debt we should use SpaceId here when we figure out how to deal with branded types in react forms
```
goran-peoski-work:
```suggestion
      // @debt should use SpaceId type here, resolve error with form typing
```

src/components/attendee/MiniProfile/MiniProfile.tsx
```diff
@@ -20,5 +25,11 @@ export const MiniProfile: React.FC<MiniProfileProps> = ({
     return <></>;
   }
 
-  return <MiniProfileModal userId={selectedUserId} />;
+  return (
+    <div ref={setReferenceElement}>
```
goran-peoski-work, line 29: [NOTE]
Code is OK, no changes needed.

Just noting the object isn't exactly element, but has a field that is a reference to the element.
If there is a single one of these `ref` might be enough, and for multiple `divRef` or `containerRef` would be distinguishing enough.

src/components/attendee/Popover.tsx
```diff
@@ -0,0 +1,38 @@
...
+  referenceElement?: Element | null;
+}
+
+export const Popover: React.FC<PopoverProps> = ({
+  children,
+  referenceElement,
+}) => {
+  const popoverContainerElement = document.querySelector(
+    `#${POPOVER_CONTAINER_ID}`
+  );
```
goran-peoski-work, line 17: [Q] Why isn't the element itself passed instead (through ref maybe)? 

[NOTE] We can probably use the new global object `SPARK` to share few of these references. Might be OOS to this PR, but it seems like a tidying up we can do later on.

src/components/attendee/Popover.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const [popperElement, setPopperElement] = useState<HTMLElement | null>(null);
+  const { styles: popperStyles, attributes: popperAttributes } = usePopper(
+    referenceElement,
+    popperElement
+  );
+
+  return (
+    popoverContainerElement &&
+    ReactDOM.createPortal(
```
goran-peoski-work, line 29: Maybe we can put a comment or two of why `createPortal` is needed instead of a common JSX

src/components/attendee/PortalModal/PortalModal.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  @include sparkle.panel--glass;
+  @include rounded.normal;
+  border: 1px solid theme.$sparkle--border;
+  box-sizing: border-box;
+  z-index: 100;
+  max-width: 200px;
+  display: flex;
+  flex-direction: column;
+
+  h3 {
```
goran-peoski-work, line 18: [NOTE]

I'm not sure we should use headings, they have semantic meaning that is hard to track (at least at this stage).
A simple div with maybe a class should be enough to have the correct style, but not have headings that might confuse screen readers.

This should all be revisited in some accessibility / usability pass over the redesigned code.

src/components/attendee/PortalModal/PortalModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+  onEnter?: MouseEventHandler<HTMLButtonElement>;
+  portal: Room;
+}
+
+export const PortalModal: React.FC<PortalModalProps> = ({
+  referenceElement,
+  onEnter,
+  portal,
+}) => {
+  const { space } = useSpaceById({ spaceId: portal.spaceId });
```
goran-peoski-work, line 22: Or, maybe if TS doesn't complain, just

```suggestion
  const { space } = useSpaceById( portal );
```

Otherwise it's a nice to have polish we can do later on.

src/components/attendee/PortalModal/PortalModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <Popover referenceElement={referenceElement}>
+      <div className={CN.PortalPopupInfo}>
+        <h3>{space?.name}</h3>
+        <p>{space?.config?.landingPageConfig.description}</p>
+        <Button onClick={onEnter}>Enter</Button>
+      </div>
```
goran-peoski-work, line 30: [NOTE]

There might be other tags that convey the correct semantic meaning here, not just `<section />` but things like aside etc. Need to do research on those options.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
```diff
@@ -159,6 +158,7 @@ export const AdminVenueView: React.FC<AdminVenueViewProps> = ({
               to={visitSpaceUrl}
               name="Visit space"
               variant="primary"
+              target="_blank"
```
goran-peoski-work: I don't know if this is safe. Will the link have `rel="noreferrer noopener"`? Is it needed?

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -50,7 +50,7 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
     name: "",
   });
 
-  const eventSpaceId = event?.spaceId || venue?.id;
+  const eventSpaceId = event?.spaceId || (venue?.id as SpaceId);
```
goran-peoski-work: Please remember the `undefined` otherwise TS will override the `?.` in checks.

```suggestion
  const eventSpaceId = event?.spaceId || (venue?.id as SpaceId|undefined);
```

src/components/templates/AnimateMap/hooks/useRelatedPartymapRooms.ts
```diff
@@ -33,7 +34,7 @@ export const useRelatedPartymapRooms: (
       relatedPartymap?.rooms?.map((portal) => {
         const noTrailSlashPortalUrl = getUrlWithoutTrailingSlash(portal.url);
 
-        const [venueId] = getLastUrlParam(noTrailSlashPortalUrl);
+        const [venueId] = getLastUrlParam(noTrailSlashPortalUrl) as SpaceId[];
         const portalVenue = findVenueInRelatedVenues({ spaceId: venueId });
```
goran-peoski-work, line 38: [OOS]
I had a refactor of this function elsewhere in larger PR that got rolled back. It would have been nice to have it back

goran-peoski-work (approved):
LGTM

Few comments and puzzles, nothing blocking (unless that one `target` prop is for external link.

src/components/attendee/Popover.tsx
sunny-viktoryia: because the element `#${POPOVER_CONTAINER_ID}` is located not as a parent node, but somewhere as an independent node in the main layout, so when we don't have the reference to the  `#${POPOVER_CONTAINER_ID}` in the place where we use `<Popover />`.

It's needed for proper z-index layout.

here's the approximate schema:

```html
<main>
 <AttendeeContent>
 ....
      <PortalComponent>
          <Popover>
            ... looking for the #POPOVER_CONTAINER_ID node to put the content of the modal
           </Popover>
      </PortalComponent>
 .... 
 <AttendeeContent>
 
 <div id={POPOVER_CONTAINER_ID}></>
</main>
```

src/components/admin/PortalAddEditForm.tsx
sunny-viktoryia: updated the comment: https://github.com/sparkletown/sparkle/pull/2969/commits/54d690c68fa6b074181f6d18a88fb443a9797d0d

src/components/admin/SpaceEditForm.tsx
sunny-viktoryia: updated the comment: https://github.com/sparkletown/sparkle/pull/2969/commits/54d690c68fa6b074181f6d18a88fb443a9797d0d

src/components/attendee/Popover.tsx
sunny-viktoryia: added a comment: https://github.com/sparkletown/sparkle/pull/2969/commits/54d690c68fa6b074181f6d18a88fb443a9797d0d

src/components/attendee/PortalModal/PortalModal.tsx
sunny-viktoryia: ts is complaining.. although i can't really get why. spaceId is optional in both cases.
image

anyway, left it as is.

src/components/organisms/AdminVenueView/AdminVenueView.tsx
sunny-viktoryia: thanks! Fixed: https://github.com/sparkletown/sparkle/pull/2969/commits/54d690c68fa6b074181f6d18a88fb443a9797d0d

src/components/organisms/TimingEventModal/TimingEventModal.tsx
sunny-viktoryia: added: https://github.com/sparkletown/sparkle/pull/2969/commits/54d690c68fa6b074181f6d18a88fb443a9797d0d

PR #2970, opened 2022-03-30
add useMemo to fix infinite request loop in various hooks

denisdimitrov:
The problem was in the `{}` syntax that was passed to `useLiveCollection`. It was creating a new object reference on each re-render and forcing the `useLiveCollection` hook to retrigger. Also the `parseCollectionQueryOptions` was creating a new reference as well, and forcing the `useEffect` to retrigger and subscribe/unsubscribe.

### DIFF
--- a/src/hooks/fire/useLiveCollection.ts
+++ b/src/hooks/fire/useLiveCollection.ts
@@ -35,7 +35,7 @@ export const useLiveCollection = <T extends object, ID extends string = string>(
     hasDeferred,
     constraints,
     filteredConstraints,
-  } = parseCollectionQueryOptions(options);
+  } = useMemo(() => parseCollectionQueryOptions(options), [options]);
 
   // construction of the query is where the Firestore SDK may break if invalid values are given
   const memoizedQuery = useMemo(
--- a/src/hooks/spaces/useSpacesByOwner.tsx
+++ b/src/hooks/spaces/useSpacesByOwner.tsx
@@ -3,10 +3,10 @@ import { where } from "firebase/firestore";
 
 import {
   ALWAYS_EMPTY_ARRAY,
-  COLLECTION_SPACES,
   DEFERRED,
   FIELD_OWNERS,
   FIELD_WORLD_ID,
+  PATH,
 } from "settings";
 
 import { LoadStatus } from "types/fire";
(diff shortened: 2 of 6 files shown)


denisdimitrov:
@goran-peoski-work I was going to remove the `useLive` implementation and replace it with `useFire` but the website just crashes. It worked fine yesterday before I rebased with `redesign-2021` but now it doesn't.

src/hooks/fire/useLiveCollection.ts
```diff
@@ -35,7 +35,7 @@ export const useLiveCollection = <T extends object, ID extends string = string>(
     hasDeferred,
     constraints,
     filteredConstraints,
-  } = parseCollectionQueryOptions(options);
+  } = useMemo(() => parseCollectionQueryOptions(options), [options]);
```
goran-peoski-work, line 38: Interesting, I didn't think this could work considering `options` itself is an object and maybe it's fields would need direct reference.

Maybe it does not change much or anything if only `options` are passed into `useLiveCollection`, but that R&D would be for another PR.

src/hooks/spaces/useSpacesByOwner.tsx
```diff
@@ -35,7 +33,11 @@ export const useSpacesByOwner: UseSpacesByOwner = ({ worldId, userId }) => {
     [worldId, userId]
   );
 
-  const result = useLiveCollection<SpaceWithId>({ path, constraints });
+  const options = useMemo(() => ({ path: PATH.spaces, constraints }), [
```
goran-peoski-work, line 36: 👍 Nice find. I will have to update some hook APIs to make it harder to fall into this trap in the future.

src/hooks/user/usePermission.ts
```diff
@@ -12,13 +12,11 @@ type AdminRole = {
   allowAll: boolean;
   users: string[];
 };
-
```
goran-peoski-work, line 15: no rename to `useLivePermission`?

goran-peoski-work (approved):
LGTM

src/hooks/user/usePermission.ts
denisdimitrov: My bad, I'll add the change.

PR #2971, opened 2022-03-31
Small tweaks to HUD / Presence

csshsh:
* Fixes  https://github.com/sparkletown/internal-sparkle-issues/issues/1795
* Fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1796

### DIFF
--- a/src/components/attendee/Attendance/Attendance.tsx
+++ b/src/components/attendee/Attendance/Attendance.tsx
@@ -12,10 +12,14 @@ type AttendanceOptions = {
 };
 
 export const Attendance: React.FC<AttendanceOptions> = ({
-  totalUsersCount = 0,
+  totalUsersCount,
   usersSample,
 }) => {
   const containerRef = useRef(null);
+
+  if (!totalUsersCount) {
+    return null;
+  }
 
   const renderedAvatars = usersSample?.map((user) => (
     <UserAvatar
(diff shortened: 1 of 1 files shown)


src/components/attendee/Attendance/Attendance.tsx
```diff
@@ -12,11 +12,15 @@ type AttendanceOptions = {
 };
 
 export const Attendance: React.FC<AttendanceOptions> = ({
-  totalUsersCount = 0,
+  totalUsersCount,
   usersSample,
 }) => {
   const containerRef = useRef(null);
 
+  if (!totalUsersCount) {
```
denisdimitrov, line 20: Maybe it's OOS, but this prop should be required instead of optional. The complier should 'cry' and force us to pass it or maybe both props. WDYT?

Right now I can render this component like this, and it won't render anything in the DOM:

```tsx
<Attendance/>
```

csshsh: Yes, I think that makes sense. Mostly wasn't sure if there are some non-obvious reasons for it to be optional. Will change it quickly

csshsh: Ah! `recentUserCount` of `SpaceWithId` is optional. Let's tackle this at another point and just fix the issue like this for now?

denisdimitrov: Sounds good. A lot of components need this change, maybe it's best if we do it in a separate PR.

denisdimitrov (approved):
LGTM!

PR #2972, opened 2022-03-31
Old Login flow visual redesign

goran-peoski-work:
Resolves:

- https://github.com/sparkletown/internal-sparkle-issues/issues/1713

There are some minor refactors of the `Button` and `Input` APIs for greater uniformity and some experimental props on an experimental component named `Spacer` that can be later re-used for the aforementioned generic components

**BEFORE**:

sparkle-login-redesign-07

**AFTER**:

sparkle-login-redesign-06


src/components/attendee/Button/Button.tsx
```diff
@@ -1,11 +1,21 @@
-import { ButtonHTMLAttributes, MouseEventHandler, RefObject } from "react";
+import React, {
+  ButtonHTMLAttributes,
+  MouseEventHandler,
+  RefObject,
+} from "react";
```
denisdimitrov, line 5: [NOTE] `React` is no longer needed as import and you can directly use it in components. If you want to, you can remove the import to reduce boilerplate but then you'd have to add the prefix everywhere, ex: `RefObject` -> `React.RefObject` It's up to you.

src/components/attendee/Button/Button.tsx
```diff
@@ -22,27 +32,32 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
 export const Button: React.FC<ButtonProps> = ({
   onClick,
   children,
-  variant = "primary",
```
denisdimitrov, line 25: [NB] [OOS?] I know I've used this component in a couple of places without passing this prop because of the default value, maybe we should keep it for now, and make this change in a separate PR.

src/components/attendee/Input/Input.tsx
```diff
@@ -2,50 +2,47 @@ import React from "react";
...
-  containerClassName,
-  inputClassName,
   onLabelClick,
   error,
   label,
   register,
   rules,
   name,
   border = "borderless",
+  variant = "",
```
denisdimitrov, line 31: [NOTE] I think this prop should have some default value or remove the optional parameter and make it required.

src/components/attendee/RetunableMediaElement/RetunableMediaElement.tsx
```diff
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useMemo, useState } from "react";
```
denisdimitrov, line 1: [NOTE] Same as before, React is no longer required as an import, no need for the default import.

denisdimitrov: The rest are fine, it would look weird to use `React.useMemo` or `React.useCallback`

src/components/attendee/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,63 @@
+import React from "react";
```
denisdimitrov, line 1: [NOTE] Same here, no longer required.

src/components/attendee/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  if (element === "span") {
+    return (
+      <span data-bem="Spacer" className={className}>
+        {children}
+      </span>
+    );
+  }
+
+  if (element === "section") {
```
denisdimitrov, line 54: What is the idea here? All 3 cases look identical and there is no difference at all.

src/utils/url.ts
```diff
@@ -205,6 +205,7 @@ export const externalUrlAdditionalProps = {
   target: "_blank",
   rel: "noopener noreferrer",
 };
+Object.freeze(externalUrlAdditionalProps);
```
denisdimitrov, line 208: Can we just move `Object.freeze` inside the function and it's return?

denisdimitrov (approved):
LGTM!

Minor notes and maybe possible improvements.

src/components/attendee/Button/Button.tsx
goran-peoski-work: Maybe it is no longer needed for the code to run, it makes the IDE more quiet - it doesn't underscore `React.FC` for example.

src/components/attendee/Button/Button.tsx
goran-peoski-work: I've made the inverse adjustments and added `<Button variant="primary"` to the few places it was being left unspecified

src/components/attendee/Input/Input.tsx
goran-peoski-work: Required might be a good idea. Will be better done in a new PR later on to uniform some more generic props (see the `<Spacer />` experimental component)

src/components/attendee/Spacer/Spacer.tsx
goran-peoski-work: It was a simple solution to have `span` and `div`, but considering it is an experimental component, the idea was to try out few style props, see how they interact and maybe build a common interface to be added to the likes of `Button`, `Input` etc after this component along with the old login are removed.

src/utils/url.ts
goran-peoski-work: Some times TS might complain the interface is `Readonly<T>` instead of `T` even when just reading, so as a workaround, I've started adding the freeze after the `const` has inferred the correct type.

PR #2974, opened 2022-03-31
Allow configuration of channels for retunable media elements

colinhowe:
Adds channels to the tuner:
image

Adds channels to the admin side:
image

### DIFF
--- a/functions/src/venue.ts
+++ b/functions/src/venue.ts
@@ -1190,22 +1190,10 @@ export const upsertChannel = functions.https.onCall(async (data, context) => {
   await throwErrorIfNeitherWorldNorSpaceOwner({
     spaceId,
     worldId: space.worldId,
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    userId: context.auth.token.user_id,
+    userId: context.auth?.token.user_id,
   });
 
-  const doc = await admin.firestore().collection("venues").doc(spaceId).get();
-
-  if (!doc || !doc.exists) {
-    throw new HttpsError("not-found", `Space ${spaceId} not found`);
-  }
-  const docData = doc.data();
-  if (!docData) {
-    throw new HttpsError("internal", `Data not found`);
-  }
-
-  let channels = docData.channels || [];
+  let channels = space.channels || [];
 
   if (typeof channelIndex !== "number") {
     channels = [...channels, channel];
@@ -1224,21 +1212,10 @@ export const deleteChannel = functions.https.onCall(async (data, context) => {
   await throwErrorIfNeitherWorldNorSpaceOwner({
     spaceId,
     worldId: space.worldId,
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    userId: context.auth.token.user_id,
+    userId: context.auth?.token.user_id,
   });
-  const doc = await admin.firestore().collection("venues").doc(spaceId).get();
 
-  if (!doc || !doc.exists) {
-    throw new HttpsError("not-found", `Space ${spaceId} not found`);
-  }
-  const docData = doc.data();
-  if (!docData) {
-    throw new HttpsError("internal", `Data not found`);
-  }
-
-  const channels = docData.channels || [];
+  const channels = space.channels || [];
 
   channels.splice(channelIndex, 1);
 
--- a/functions/src/venue.ts
+++ b/functions/src/venue.ts
@@ -1179,3 +1179,68 @@ export const deleteScreeningRoomVideo = functions.https.onCall(
       .delete();
   }
 );
+
+export const upsertChannel = functions.https.onCall(async (data, context) => {
+  assertValidAuth(context);
+
+  const { spaceId, channelIndex, channel } = data;
+
+  const space = await getSpaceById(spaceId);
+
+  await throwErrorIfNeitherWorldNorSpaceOwner({
+    spaceId,
+    worldId: space.worldId,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    userId: context.auth.token.user_id,
+  });
+
+  const doc = await admin.firestore().collection("venues").doc(spaceId).get();
+
+  if (!doc || !doc.exists) {
+    throw new HttpsError("not-found", `Space ${spaceId} not found`);
+  }
+  const docData = doc.data();
+  if (!docData) {
+    throw new HttpsError("internal", `Data not found`);
+  }
+
+  let channels = docData.channels || [];
+
+  if (typeof channelIndex !== "number") {
+    channels = [...channels, channel];
+  } else {
+    channels[channelIndex] = channel;
+  }
+
+  admin.firestore().collection("venues").doc(spaceId).update({ channels });
+});
+
+export const deleteChannel = functions.https.onCall(async (data, context) => {
+  const { spaceId, channelIndex } = data;
+
+  const space = await getSpaceById(spaceId);
+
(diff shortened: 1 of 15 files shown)


functions/src/venue.ts
```diff
@@ -1141,3 +1141,68 @@ export const deleteScreeningRoomVideo = functions.https.onCall(
...
+  const { spaceId, channelIndex, channel } = data;
+
+  const space = await getSpaceById(spaceId);
+
+  await throwErrorIfNeitherWorldNorSpaceOwner({
+    spaceId,
+    worldId: space.worldId,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    userId: context.auth.token.user_id,
```
denisdimitrov: You can safely remove the `@ts-ignore` and do this.
```suggestion
    userId: context.auth?.token.user_id,
```

Plus, `assertValidAuth(context)` already checks for this context and it already exists, however we should probably return it in the function and get the new value that's going to have the correct type.

functions/src/venue.ts
```diff
@@ -1141,3 +1141,68 @@ export const deleteScreeningRoomVideo = functions.https.onCall(
...
+  });
+
+  const doc = await admin.firestore().collection("venues").doc(spaceId).get();
```
denisdimitrov: You are already getting the space above `const space = await getSpaceById(spaceId);` do you need this code? It might increase the request complete time, especially on slow internet.

functions/src/venue.ts
```diff
@@ -1141,3 +1141,68 @@ export const deleteScreeningRoomVideo = functions.https.onCall(
...
+  const space = await getSpaceById(spaceId);
+
+  await throwErrorIfNeitherWorldNorSpaceOwner({
+    spaceId,
+    worldId: space.worldId,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    userId: context.auth.token.user_id,
+  });
+  const doc = await admin.firestore().collection("venues").doc(spaceId).get();
```
denisdimitrov: Same comments like the `upsertChannel`

src/api/admin.ts
```diff
@@ -572,3 +577,44 @@ export const getSpaceEditors = async (spaceEditorIds: string[]) =>
   Promise.all(
     spaceEditorIds.map((admin) => getUserRef(admin).get())
   ).then((docs) => docs.map((doc) => withId(doc.data() as User, doc.id)));
+
+export const upsertChannel = async (
+  channel: Channel,
+  spaceId: string,
```
denisdimitrov: Can we use `spaceId: SpaceId` here?

src/api/admin.ts
```diff
@@ -572,3 +577,44 @@ export const getSpaceEditors = async (spaceEditorIds: string[]) =>
...
+      event.addMetadata("api/admin::upsertChannel", {
+        spaceId,
+        channelIndex,
+      });
+    });
+    throw e;
+  });
+};
+
+export const deleteChannel = async (spaceId: string, channelIndex?: number) => {
```
denisdimitrov: Can we use `spaceId: SpaceId` here?

src/components/admin/ChannelConfiguration/ChannelAddEditForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+          errors={errors}
+          name="iframeUrl"
+          register={register}
+          disabled={isLoading}
+          label="Embed URL"
+        />
+
+        <SubmitError error={submitError} />
+        <div className="mt-5 sm:mt-4 sm:flex justify-end">
+          <Button variant="secondary" disabled={isLoading} onClick={onDone}>
```
denisdimitrov, line 86: `onDone` is passed here because the prop functionality matches the usage here, but maybe it would be good to create another prop `onCancel` and pass the function again, it would be semantically better IMO.

src/components/admin/ChannelConfiguration/ChannelConfiguration.tsx
```diff
@@ -0,0 +1,143 @@
...
+      };
+    }
+  }, [space.channels?.length, isDeleting]);
+
+  const saveChannel = async (channelData: Channel) => {
+    await adminApi.upsertChannel(channelData, space.id, editingIndex);
+    setEditingIndex(undefined);
+  };
+
+  const deleteChannel = useAsyncFn(async () => {
```
denisdimitrov, line 67: `useAsyncFn` actually handles the loading state quite gracefully. You can see the implementation here: https://github.com/streamich/react-use/blob/master/docs/useAsyncFn.md

Ex: const [{loading: isDeleting}, deleteChannel] = useAsyncFn....

denisdimitrov (changes requested):
Some comments and possible improvements.

src/components/organisms/AdminVenueView/components/Spaces/Spaces.tsx
```diff
@@ -64,6 +65,11 @@ export const Spaces: React.FC<SpacesProps> = ({ space, world }) => (
             <ScreeningRoomVideosTable space={space} />
           </>
         )}
+        {space.template === VenueTemplate.meetingroom && (
+          <>
+            <ChannelConfiguration space={space} />
+          </>
+        )}
```
denisdimitrov: The fragment can be removed.

src/components/admin/ChannelConfiguration/channelSchema.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface ChannelSchemaShape {
+  name: string;
+  iframeUrl: string;
+}
+
+export const channelSchema = Yup.object().shape<ChannelSchemaShape>({
+  name: createNameSchema({ name: "Name", withMin: true }),
+  iframeUrl: Yup.string().required(`URL is required`),
+});
```
denisdimitrov, line 13: This file should be in `forms/` unless we have a new convention and I don't know about it, in that case ignore my comment.

src/components/admin/ChannelConfiguration/ChannelConfiguration.tsx
```diff
@@ -0,0 +1,143 @@
...
+    isDeleting,
+    showCreateModal,
+    showDeleteModal,
+    space.channels,
+  ]);
+
+  const finishEditing = useCallback(() => {
+    hideCreateModal();
+    setEditingIndex(undefined);
+  }, [hideCreateModal]);
```
denisdimitrov, line 112: [OOS] [NB] We might be overusing the caching hooks and this might me one of those cases, it's probably less expensive for react to re-initialize this on re-render instead of cache it and use memory. Just pointing it out. No change required.

src/components/admin/ChannelConfiguration/ChannelConfiguration.tsx
```diff
@@ -0,0 +1,143 @@
...
+  } = useShowHide(false);
+
+  const [editingIndex, setEditingIndex] = useState<number>();
+
+  const [isDeleting, setIsDeleting] = useState(false);
+
+  useEffect(() => {
+    // The cleanup below is only triggered when a deletion is complete and
+    // the number of channels is reduced because we have received new data
+    // The use of channels.length in the dependencies is important.
```
denisdimitrov, line 49: I understand the idea and smart implementation but why can't we just add the code after the request is complete?

denisdimitrov (commented):
A couple of non blocking and nit picking comments

src/components/admin/ChannelConfiguration/ChannelConfiguration.tsx
colinhowe: This is all related to the logic to handle that the data is updated in response to a push from the database rather than being the response from the function.

PR #2975, opened 2022-03-31
remove unnecessary element from WorldNav component

denisdimitrov:
Removes unnecessary element from the `WorldNav` component, that was breaking the tab's UI and their inherited width.

Before:
oldnav

After
newnavbar


goran-peoski-work (approved):
LGTM.

Seems like an unnecessary component being there is only a hook that can be moved to the parents of WorldNav, still, it's nice to have it separate like this.

PR #2977, opened 2022-04-01
Fix accessing length on undefined in SpacesDashboard

goran-peoski-work:
A two-liner fix for the case of `undefined?.filter` not returning an array but `undefined` which blows up if later accessing `length` without `?.`


denisdimitrov (approved):
LGTM!

PR #2978, opened 2022-04-01
Redesign admin general space settings

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1695

Some screenshots:

PartyMap
image

ArtPiece Embeddable Content
image

ExternalExperience Embeddable Content
image

JazzBar Embeddable Content
image

Auditorium
image

Screening Room Videos:
image

PR #2979, opened 2022-04-01
Follow-up of the 'Redesign admin general space settings'

sunny-viktoryia:
follow-up: https://github.com/sparkletown/sparkle/pull/2978


src/settings/index.ts
```diff
@@ -94,12 +94,12 @@ export const IFRAME_TEMPLATES = [
...
   VenueTemplate.viewingwindow,
-];
+]);
 
-export const EMBEDDABLE_CONTENT_TEMPLATES = [
+export const EMBEDDABLE_CONTENT_TEMPLATES = Object.freeze([
   ...IFRAME_TEMPLATES,
   ...ZOOM_URL_TEMPLATES,
-];
+]);
```
goran-peoski-work, line 102: [NOTE] This is nor request to change, just an FYI.

In most cases I use this form:

```ts
const object = {};
Object.freeze(object);
```

just because in the first case TS might infer the type as `Readonly<Record<T,R>>`, while in the snippet I provided is `Record<T,R>`. Usually I'd prefer the way you wrote it, but in some cases, it might be easier than dealing with type complaints.

goran-peoski-work (approved):
LGTM

PR #2991, opened 2022-04-04
Fix the behavior of Huddle videos on overflow

sunny-viktoryia:
fixes https://github.com/sparkletown/internal-sparkle-issues/issues/1764


src/components/attendee/VideoHuddle/VideoHuddle.module.scss
```diff
@@ -57,6 +57,13 @@ $-control-size: 32px;
   overflow-x: scroll;
   overflow-y: hidden;
   display: flex;
-  justify-content: center;
   gap: space.gap(1);
+
+  // use pseudoelements instead of justify-content:center to avoid hiding of the first child elements on content overflow
+  // @see https://stackoverflow.com/a/46059525
```
goran-peoski-work: Were other solutions considered? 
If so, it might be good to add the conclusion of those in the comment above `@see`

This is a solution from 2017 so the question is why should it be preferred over `display: flex` or `display: grid` with the proper `align-`, `justify-` and `place-` properties?

goran-peoski-work (changes requested):
Not sure about this one, but if there are some problems with newer technologies like grid and/or flexbox, at least a more exhaustive comment is needed to capture the knowledge.

src/components/attendee/VideoHuddle/VideoHuddle.module.scss
goran-peoski-work:
```suggestion

  // Use pseudoelements instead of justify-content:center to avoid hiding of the first child elements on content overflow
  // @see https://stackoverflow.com/a/46059525
  
  // Pseudoelements are quirky that if there is a standard flex/grid solution we should use it.
  // Currently there is a solution that isn't fully supported in all browsers that we might try in the future.
  // @see https://stackoverflow.com/a/33455342
  // @see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox#Flex_item_considerations
```

goran-peoski-work (approved):
Approving with the hope of the suggested comment addition is merged as well

src/components/attendee/VideoHuddle/VideoHuddle.module.scss
sunny-viktoryia: used your description of the issue to use in the comment: https://github.com/sparkletown/sparkle/pull/2991/commits/4be619acbf1cd0e0ef0759efa9d321d2f3feccf1
Thanks, Goran!

PR #2992, opened 2022-04-04
Space visual issues fix

alexxpetrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1799

* Remove 'assets/images/sparkle-bg-2.png' as the default background ( didn't find this asset )
* The background for the element should be light grey #e6e6e6 or its dark/light mode equivalent (not white) ( used theme color )
* fixed chat container moves space background
* fixed chat side bar absolute/relative location (stuck to bottom of the body when footer is seen/follows scroll when footer isn't seen )


src/components/attendee/AttendeeLayout/AttendeeLayout.module.scss
```diff
@@ -53,9 +54,10 @@ $-z-index--popover: layer.level($-z-index--ui);
   }
 }
 
-.LayerUiAbsolute {
-  position: absolute;
-  top: 0px;
+.LayerUiRelative {
```
goran-peoski-work: Wait, are we renaming this now? How about `layer`? Do we need it complicated?

goran-peoski-work: Wait, are we renaming this now? How about `layer`? Do we need it complicated?

goran-peoski-work (approved):
LGTM

src/components/attendee/AttendeeLayout/AttendeeLayout.module.scss
alexxpetrov: renamed into `relative`

PR #2993, opened 2022-04-04
More deletions

colinhowe:
Removed most of the unused components


alexxpetrov (approved):
LGTM
might've missed something tho, a lot has been removed
guess we'll see how it goes when we use the app

PR #2994, opened 2022-04-04
Redesign old 4xx pages for use in both admin and attendee sides

goran-peoski-work:
Resolves
- https://github.com/sparkletown/internal-sparkle-issues/issues/1786

until the new login flow is implemented in
- https://github.com/sparkletown/internal-sparkle-issues/issues/1800

Following
- https://github.com/sparkletown/sparkle/pull/2972

Note about this PR:
specific care has been taken to distinguish the admin and attendee sides to only (lazily) load the proper style system for it.

Some images of the change:
sparkle-4xx-05
sparkle-4xx-04
sparkle-4xx-03
sparkle-4xx-02
sparkle-4xx-01


src/components/admin/errors/BlankPage.tailwind.ts
```diff
@@ -0,0 +1,10 @@
+//   font-variation-settings: "wght" 350, "ital" 0, "SRFF" 20;
```
alexxpetrov, line 1: [Question]  Do we still need it?

src/components/attendee/errors/BlankPage.module.scss
```diff
@@ -0,0 +1,27 @@
+@use "scss/attendee/font";
+@use "scss/attendee/theme";
+
+.container {
+  height: 100vh;
```
alexxpetrov, line 5: [Question] is there any advantage using `vh`/`vw` in comparison to percents `%` in this case?

src/components/attendee/errors/Forbidden.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+import { BlankPage } from "./BlankPage";
+
+export const Forbidden: React.FC = () => (
+  <BlankPage>
+    <h1 data-bem="Forbidden__title" data-side="att">
+      403 Not authorised
```
alexxpetrov, line 8: We might want to have a shared config file with these error codes and description to use in both admin and attendee side. It's currently used in 2 places, so keeping it all in a single config would be a good thing

src/components/attendee/errors/NotFound.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+import { BlankPage } from "./BlankPage";
+
+export const NotFound: React.FC = () => {
+  return (
+    <BlankPage>
+      <h1 data-bem="NotFound__title" data-side="att">
+        404 Not found
```
alexxpetrov, line 9: same as above

src/components/attendee/errors/Unauthorized.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+import { BlankPage } from "./BlankPage";
+
+export const Unauthorized: React.FC = () => (
+  <BlankPage>
+    <h1 data-bem="Unauthorized__title" data-side="att">
+      401 Not authenticated
```
alexxpetrov, line 8: same as above

src/pages/auth/LoginForm/LoginForm.hoc.tsx
```diff
@@ -20,9 +20,9 @@ export const LoginFormHoc = (props: LoginFormHocProps) => {
     return <LoadingPage />;
   }
 
-  if (!space || !world) {
-    return <NotFoundFallback />;
-  }
-
-  return <LoginForm {...props} space={space} world={world} />;
+  return space && world ? (
```
alexxpetrov, line 23: [Question] doesn't it look better if we're rendering the `NotFound` in case any of `(!space || !world)` is invalid first?
so that we're eliminating this ternary and making render more straightforward?

src/pages/auth/RegisterForm/RegisterForm.hoc.tsx
```diff
@@ -41,6 +41,6 @@ export const RegisterFormHoc = (props: RegisterFormHocProps) => {
       isWorldLoaded={isLoaded}
     />
   ) : (
-    <NotFoundFallback />
+    <NotFound />
```
alexxpetrov, line 44: same question here

alexxpetrov (commented):
A few questions to discuss and some actions needed

src/components/admin/errors/BlankPage.tailwind.ts
goran-peoski-work: Maybe we do, the issue I had was how to translate it to Tailwind, so instead I just left it there as a reminder. Maybe a more complete comment would have  been great.

src/components/attendee/errors/BlankPage.module.scss
goran-peoski-work: `vh` and `vw` do work as intended: the screen (viewport) width and height. 
The `%` is (semantically) relative to parent so might not always be the same, thus it's a clearer intention to use `vh` and `vw`.

NOTE: there are more viewport units on the way in CSS that will make them even more useful than `%`.

src/components/attendee/errors/Forbidden.tsx
goran-peoski-work: They aren't codes but a display text that will most likely be replaced after the new login flow is implemented. Maybe even different for attendee and admin sides.

src/pages/auth/LoginForm/LoginForm.hoc.tsx
goran-peoski-work: It's some times easy to determine, other times harder to figure out what is the "looks better" solution.

In this case I opted to have the `space` and `world` front and more prominent with less syntax clutter than having (! || !) around them.

alexxpetrov (commented):
LGTM

alexxpetrov (approved):

PR #2997, opened 2022-04-05
show room participants in MeetingRoom before join

denisdimitrov:
Displays the room participants in the Meeting room before joining the room itself.

Current implementation will fetch them every 10 seconds, but this needs to be discussed with Pearl and the design team.


denisdimitrov:
New solution was introduced by Alexey and this was scraped because of technical limitations.

PR #2998, opened 2022-04-05
Fix make world owner script ts node path

alexxpetrov:
Fixes following error when trying to run the `make-owner-all-world-spaces.ts` script

image


goran-peoski-work (approved):
LGTM

PR #2999, opened 2022-04-05
Redesign WelcomeVideo component

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1714


src/components/attendee/AttendeeLayout/AttendeeLayout.tsx
```diff
@@ -13,6 +14,8 @@ import { ATTENDEE_LAYOUT_CLASSNAME, POPOVER_CONTAINER_ID } from "settings";
 
 import { SpaceWithId } from "types/id";
 
+import { isTruthy } from "utils/types";
```
goran-peoski-work, line 17: I notice this added, but don't see new line using it.

src/components/attendee/SpaceEntrancePage/index.ts
```diff
@@ -9,9 +9,9 @@ import { Login } from "pages/auth/Login";
 
 import { LoadingPage } from "components/molecules/LoadingPage";
 
-import { VenueEntrancePage as _VenueEntrancePage } from "./VenueEntrancePage";
+import { SpaceEntrancePage as _SpaceEntrancePage } from "./SpaceEntrancePage";
 
-export const VenueEntrancePage = compose(
+export const SpaceEntrancePage = compose(
```
goran-peoski-work: I think this change might have been a good opportunity to replace the `compose()` with `SpaceEntrancePageHoc` written as "in the good old style". 

There are about 9 `.hoc.tsx` components already. As an example from them, they no longer need to check if user is logged in since they're descendants of `LoginRestricted` etc.

If the above is OOS, please add `// @debt replace the compose() with SpaceEntrancePage.hoc.tsx

src/components/attendee/WelcomeVideo/WelcomeVideo.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import CN from "./WelcomeVideo.module.scss";
+
+const DEFAULT_BUTTON_TEXT = "Proceed";
+
+export const WelcomeVideo: React.FC<EntranceStepTemplateProps> = ({
+  config,
+  proceed,
+}) => {
+  const { world } = useWorldAndSpaceByParams();
```
goran-peoski-work, line 23: [NOTE] to self: combine both `useWorldParams` and `useWorldById` into similar `useWorldByParams()` that doesn't deal with space

src/components/attendee/WelcomeVideo/WelcomeVideo.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const { autoplay: autoPlay, buttons, videoUrl: url } = config;
+  const hasProceedButton = !!buttons?.find(({ isProceed }) => isProceed);
+
+  const backgroundCss = useCss({
+    backgroundImage:
+      world?.config.landingPageConfig.coverImageUrl &&
```
goran-peoski-work: Might be easier to read if there is a `const imageUrl = world?.config.landingPageConfig.coverImageUrl` before `useCss`

src/components/attendee/WelcomeVideo/WelcomeVideo.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const renderedButtons = useMemo(
+    () =>
+      (buttons ?? ALWAYS_EMPTY_ARRAY).map(
+        ({ className, href, text }, index) => (
+          <Button
+            key={href || `${index}`}
+            variant="intensive"
+            className={className}
+            onClick={() => (href ? window.open(href) && proceed() : proceed())}
```
goran-peoski-work, line 41: This is probably unsafe. Generally the `href` would be internal link, but there isn't a check for it.

[OOS] Add utility function that performs a safe `window.open()` and does necessary checks

src/components/attendee/WelcomeVideo/WelcomeVideo.tsx
```diff
@@ -0,0 +1,73 @@
...
+                </Button>
+              )}
+
+              {renderedButtons}
+            </div>
+          </div>
+        </MainSection>
+      </main>
+
+      <AttendeeFooter />
```
goran-peoski-work, line 70: [NOTE] I would think there wouldn't be a need of this fragment, nor the footer since this component would be inside a layout component (like `AttendeeLayout`), but that might be an improvement for later date.

goran-peoski-work (approved):
LGTM

Some comments left, mostly as FYI and for future consideration

src/components/attendee/AttendeeLayout/AttendeeLayout.tsx
sunny-viktoryia: it's on the 68th line. 🙃

src/components/attendee/SpaceEntrancePage/index.ts
sunny-viktoryia: added hoc: https://github.com/sparkletown/sparkle/pull/2999/commits/afffc7509388a525b1bf9d26d509d3719cc3a216

src/components/attendee/WelcomeVideo/WelcomeVideo.tsx
sunny-viktoryia: created a const: https://github.com/sparkletown/sparkle/pull/2999/commits/afffc7509388a525b1bf9d26d509d3719cc3a216

PR #3002, opened 2022-04-06
Add snippets to determine which fn in createVenue_v2 throws an error

goran-peoski-work:
The goal is twofold:
  1. Create a simple way of "annotating" rejections from cloud functions
  2. Use it to annotate which of the 3 calls inside `createVenue_v2` is the culprit

As a bonus, the form error messages will be possible to be made nicer and controlled by the front end.


src/api/admin.ts
```diff
@@ -217,40 +218,61 @@ export const createVenue_v2 = async (
...
-      ...input,
-      showShoutouts: input.showShoutouts ?? DEFAULT_SHOW_SHOUTOUTS,
-      showReactions: input.showReactions ?? DEFAULT_SHOW_REACTIONS,
-      rooms: [],
-    },
+    inputOptions,
     userId
+  ).catch(
+    createErrorRethrow({
+      where: "createVenue_v2",
```
denisdimitrov, line 232: [NB] Would be good to create variables to avoid typos, unless this is TS and it will cry with a compile error.

denisdimitrov (approved):
LGTM!

src/api/admin.ts
goran-peoski-work: It is purely informational. Later on it may be added to a Bugsnag context (easy place for the Bugsnag would be in the parent error)

PR #3003, opened 2022-04-06
Move logout button: MiniProfile -> ProfileOverlay

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1807

also:
* fixes bug of breaking overlay on the 2nd opening
* makes cancel button on the ProfileOverlay close the overlay
* adjusts Card sizes with/without CardButton

image
image
image


src/components/attendee/Button/Button.tsx
```diff
@@ -25,6 +25,7 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   border?: BorderVariant;
   unrounded?: boolean;
   marginless?: boolean;
+  large?: boolean;
```
goran-peoski-work, line 28: [NOTE][OOS]
There is a way of doing this in line with how `  CN["border-" + border],`  works.

I think it is highly likely we'd need more sizes in which case we can update this prop to use `SizeProp` like the experiment in https://github.com/sparkletown/sparkle/blob/redesign-2021/src/components/attendee/Spacer/Spacer.tsx

The style file in it has this idea realized as: 
```scss
$-sizes: (
  "none": 0,
  //
  "min": space.gap(1),
  "max": space.gap(10),
  //
  "small": space.gap(2.5),
  "medium": space.gap(5),
  "large": space.gap(10),
);
```

src/components/attendee/Card/Card.module.scss
```diff
@@ -2,6 +2,8 @@
 @use "scss/attendee/space";
 @use "scss/attendee/theme";
 
+$button-height: 42px;
```
goran-peoski-work: Private constant, just for this file

```suggestion
$-button-height: 42px;
```

src/components/attendee/Card/Card.module.scss
```diff
@@ -19,7 +21,11 @@
 @each $class, $size in layout.$card-sizes {
```
goran-peoski-work, line 21: [NOTE][OOS]
Just like above, we can probably uniform these sizes for all generic components (Button, Input, Toggle etc.) and then just provide the meaningful styles in the SCSS file.

src/components/attendee/Card/Card.tsx
```diff
@@ -10,14 +10,18 @@ type CardSize = "small";
...
 interface CardProps {
   size?: CardSize;
+  withoutButton?: boolean;
 }
 
 export const Card: React.FC<CardProps> & {
   Button: React.FC<ButtonProps>;
   Body: React.FC;
-} = ({ children, size = "small" }) => {
+} = ({ children, size = "small", withoutButton = false }) => {
```
goran-peoski-work, line 19: [NOTE]
👍 This is an interesting/good idea.

We can stop inventing proper adjectives and just add `with` and/or `without` in front of props. E.g. instead of `borderless`/`borderful`/`bordered` we can just use `withBorder` and `withoutBorder`

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
```diff
@@ -162,6 +169,16 @@ export const ProfileOverlay: React.FC<ProfileOverlayProps> = ({ profile }) => {
     ]
   );
 
+  const history = useHistory();
+  const { worldSlug, spaceSlug } = useWorldAndSpaceByParams();
```
goran-peoski-work: This is a bit heavy. `useWorldAndSpaceByParams` does two requests to DB. 
Yes, the result may usually be cached in the browser so it will be as performant. 

Yet, another hook just for the slugs is already present (even inside `useWorldAndSpaceByParams`) - `useSpaceParams`.

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
```diff
@@ -162,6 +169,16 @@ export const ProfileOverlay: React.FC<ProfileOverlayProps> = ({ profile }) => {
...
+
+  const [{ loading: isLoggingOut }, logout] = useAsyncFn(async () => {
+    await firebase.auth().signOut();
+    onOverlayClose();
+
+    history.push(generateAttendeeSpaceLandingUrl(worldSlug, spaceSlug));
```
goran-peoski-work: I think this `generateAttendeeSpaceLandingUrl` is (or should be) marked as `@deprecated` in favor of `generateUrl()`.
Can you please do  the replacement?

The idea is that once we remove all these and only `generateUrl()` remains, we can improve that one for nicer DX and more robust handling.

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
```diff
@@ -200,11 +217,19 @@ export const ProfileOverlay: React.FC<ProfileOverlayProps> = ({ profile }) => {
...
+              disabled={isSubmitting}
+              variant="alternative"
+              onClick={onOverlayClose}
+            >
               Cancel
             </Button>
             {isSuccess ? <span>✓ Saved</span> : ""}
           </div>
+
+          <Button variant="alternative" onClick={logout} large>
```
goran-peoski-work: Will this make the flag prop more visible?

```suggestion
          <Button large variant="alternative" onClick={logout}>
```

goran-peoski-work (approved):
Looks good.

Please go through the comments, some are FYI, others are some nice to have refactor suggestions.

src/components/attendee/Card/Card.module.scss
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/3003/commits/07ec924f308486ffff85bb746fdba69810c19e56

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
sunny-viktoryia: you're right. Confused the name of the hook. 🤦🏼‍♀️ 
fixed: https://github.com/sparkletown/sparkle/pull/3003/commits/07ec924f308486ffff85bb746fdba69810c19e56

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/3003/commits/07ec924f308486ffff85bb746fdba69810c19e56

src/components/attendee/ProfileOverlay/ProfileOverlay.tsx
sunny-viktoryia: moved: https://github.com/sparkletown/sparkle/pull/3003/commits/07ec924f308486ffff85bb746fdba69810c19e56

PR #3004, opened 2022-04-06
New presence tracking

colinhowe:
This creates a new close-to-realtime presence tracker. The data used
is as minimal as possible to reduce the amount of data sent
to clients. It is also debounced to reduce render thrash.


alexxpetrov (approved):
LGTM

PR #3005, opened 2022-04-06
Add onEnter handler to the search input

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1808

also:
* makes visible "Search" label

image


goran-peoski-work (changes requested):
I like the removal of the pseudoclass, but please do connect the label with the input. 

Example:
```
  const inputId = useMemo(() => generateId("Input"), []);
  <label htmlFor={id} />
  <input id={id} />

File: src/components/attendee/Input/Input.module.scss [hunk 1]
```diff
@@ -8,25 +8,17 @@
...
-    padding: 0 0 0 space.empty(0.75);
-    height: 26px;
-    color: theme.$c-glass-panel-text;
-    right: space.empty(0.8);
-    border-left: border.$width--slim solid theme.$sparkle--border;
-    color: theme.$c-inverted-text;
-    cursor: pointer;
-  }
+label {
+  position: absolute;
```
goran-peoski-work: Shouldn't the flexbox of the Input parent determine how to layout the children like `input` and `label` properly?

goran-peoski-work (commented):
Minor question about the necessity of the label positioning in such strict terms

src/components/attendee/Input/Input.module.scss
goran-peoski-work: [NOTE]

Also, I just realized this is not the original label of being just an explanatory text above the input, but rather a subtext label. In some earlier components there was a prop that determined where the label is shown, others had 2 labels taken as props

sunny-viktoryia: maybe we can rename it?

in props
label -> actionButton? (sideLabel?)

and in the component
use button or span for it?

goran-peoski-work: It can be renamed now or later, we should first have a clear picture of what makes an `Input` component. Does design require it, does separate label for any input make more sense, etc.

sunny-viktoryia: removed and refactored attendee input component a bit. Check out here: https://github.com/sparkletown/sparkle/pull/3005/commits/56622b7f4483b16d4282f57daa294296e318b5dc

goran-peoski-work (approved):
LGTM

PR #3007, opened 2022-04-07
set state properly for timing event modal

denisdimitrov:
The issue was not the form validation but the area that selects and sets the state. The problem was the callback being passed to the children div which was only a text, and clicking outside the text wasn't changing the state and the validation was failing.

This might happen in other forms as well and IMO we should pass an array of options to the dropdown instead of children, that's the way React Native handles dropdowns and lists and it works without any problems at all.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1811


src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -165,7 +172,9 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
           {!eventSpace?.name && (
             <>
               <div className="mb-6">
-                <Dropdown title={"None"}>{renderedSpaceIds}</Dropdown>
+                <Dropdown onSelect={selectSpace} title={"None"}>
```
goran-peoski-work:
```suggestion
                <Dropdown onSelect={selectSpace} title="None">
```

src/components/organisms/TimingEventModal/TimingEventModal.tsx
```diff
@@ -103,14 +104,20 @@ export const TimingEventModal: React.FC<TimingEventModalProps> = ({
...
         );
       }) ?? [],
     [spacesMap]
   );
 
+  const selectSpace = (option: Option) => {
+    if (option.value) {
+      setSelectedSpace(option.value as SpaceType);
+    }
+  };
```
goran-peoski-work: [Q] Does this mean once selected a space can't be un-selected? How does or should one be able to remove any space from the options?

A second, maybe related, scenario: user picks a space, later deletes that space, what id shown here?

goran-peoski-work (commented):
It seems to be resolving an issue, but unsure if causing new ones

src/components/organisms/TimingEventModal/TimingEventModal.tsx
denisdimitrov: The space is required and once selected it makes no sense to unselect it.

We haven't handled this delete case or any other similar ones on the admin side.

goran-peoski-work: good enough

goran-peoski-work (approved):
LGTM

PR #3009, opened 2022-04-07
Recalculate image ratio when changing image on portal

colinhowe:
This change was a bit more involved than expected as it required drilling the width and height of the map through.

There is also state maintained locally by the map component that needed dealing with.

On the bright side: removed some more unused code as that simplified implementation


src/components/admin/PortalAddEditForm.tsx
```diff
@@ -38,12 +38,52 @@ import { useUser } from "hooks/useUser";
...
+  if (existingWidthPx > existingHeightPx) {
+    return [
+      portal.width_percent,
+      (100 * existingWidthPx * newRatio) / mapHeightPx,
+    ];
+  } else {
+    return [
+      (100 * existingHeightPx) / newRatio / mapWidthPx,
+      portal.height_percent,
+    ];
```
goran-peoski-work, line 75: Could use a line of comment explaining the formula me thinks

src/components/admin/PortalAddEditForm.tsx
```diff
@@ -84,6 +128,8 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
       portal?.type,
       portal?.isEnabled,
       portal?.spaceId,
+      portal?.width_percent,
+      portal?.height_percent,
```
goran-peoski-work, line 132: [Q] Why aren't these camelCased already?
Is this PR a good opportunity or is it too much of a fix, thus OOS?

I think even if in DB it must remain backward compatible as `width_percent` and `height_percent` it is not necessarily useful to propagate the convention onto components.

src/components/admin/PortalAddEditForm.tsx
```diff
@@ -113,8 +159,24 @@ export const PortalAddEditForm: React.FC<PortalAddEditFormProps> = ({
...
+        const [newWidth, newHeight] = determineNewWidthAndHeight({
+          portal,
+          newWidthPx: checkImage.naturalWidth,
+          newHeightPx: checkImage.naturalHeight,
+          mapWidthPx,
+          mapHeightPx,
+        });
+        setValue("width_percent", newWidth);
+        setValue("height_percent", newHeight);
+      });
```
goran-peoski-work, line 177: I recommend not leaving uncaught rejections. Attaching `.catch()` here akin to https://github.com/sparkletown/sparkle/blob/redesign-2021/src/components/templates/Auditorium/components/Section/Section.tsx#L47-L53
would make it useful to know if a problem ocurred:

```suggestion
      }).catch(
        createErrorCapture({
          message: "Problem setting width and height percents",
          where: "PortalAddEditForm",
        })
      );
```

goran-peoski-work (approved):
Looks good.

I would like to have the `.catch()` suggestion in there before merge, the rest of the comments are for nice to have things, but not that important.

PR #3013, opened 2022-04-08
Some tidy up of code related to UserAvatar

goran-peoski-work:
Tried to resolve:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1806

but couldn't reproduce.

Instead, opted out to do some tidying up in the code of some easy and quick to refactor places like isolating non-shared constants and adding `@debt` comments.

There is still a lot of debt to be repaid by the `UserAvatar` component, especially in terms of removing the injected classes to achieve robustness and uniformity.


denisdimitrov (approved):
LGTM!

PR #3014, opened 2022-04-08
add consistent loading buttons around the platform

denisdimitrov:
Adds the correct loading styles from the mock ups repo and uses that around the platform for consistent UI.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1815

### DIFF
--- a/src/components/admin/Button/Button.tsx
+++ b/src/components/admin/Button/Button.tsx
@@ -28,10 +28,12 @@ export const Button: React.FC<ButtonProps> = ({
     data-bem="Button"
     {...extraParams}
     type={type}
-    className={cn(TW.general, TW[variant], TW[borders])}
+    className={cn(TW.general, TW[variant], TW[borders], {
+      [TW.loading]: loading,
+    })}
     ref={forwardRef}
   >
-    {loading && <Loading />}
+    {loading && <Loading containerClassName="pr-4" />}
     {children}
   </button>
 );
--- a/src/components/admin/Button/Button.tsx
+++ b/src/components/admin/Button/Button.tsx
@@ -33,7 +33,11 @@ export const Button: React.FC<ButtonProps> = ({
     })}
     ref={forwardRef}
   >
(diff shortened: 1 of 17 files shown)


src/components/admin/Button/Button.tsx
```diff
@@ -28,10 +28,12 @@ export const Button: React.FC<ButtonProps> = ({
...
     {...extraParams}
     type={type}
-    className={cn(TW.general, TW[variant], TW[borders])}
+    className={cn(TW.general, TW[variant], TW[borders], {
+      [TW.loading]: loading,
+    })}
     ref={forwardRef}
   >
-    {loading && <Loading />}
+    {loading && <Loading containerClassName="pr-4" />}
```
goran-peoski-work: [OOS] There are two not mutually exclusive ways of dealing with issues in this:
1. use variance props not injected classes
2. use a facade that works for both admin and attendee sides (similar to https://github.com/sparkletown/sparkle/blob/redesign-2021/src/components/shared/NotFound.tsx )

src/components/admin/SpaceCreateForm/SpaceCreateForm.tsx
```diff
@@ -35,7 +35,7 @@ import { PortalList } from "components/molecules/PortalList";
 import { SubmitError } from "components/molecules/SubmitError";
 import { YourUrlDisplay } from "components/molecules/YourUrlDisplay";
 
-import { ButtonNG } from "components/atoms/ButtonNG";
```
goran-peoski-work, line 38: Did we get to a point that will make it safe to just delete `ButtonNG`?

goran-peoski-work (approved):
LGTM

src/components/admin/SpaceCreateForm/SpaceCreateForm.tsx
denisdimitrov: Almost! Some of the Attendee components are still using the `NG` version of the button.

PR #3021, opened 2022-04-11
Add annotated interfaces and data props for spaces

goran-peoski-work:
There are few cases this PR tries to improve:
- Replace `WithId<>` types for spaces in favor of `*SpaceWithId` ones
- Add `data-bem`, `data-block` and `data-side` attributes for building blocks related to spaces on the attendee side
- Add few `.catch(createErrorCapture(())` handlers for ignored rejections in some of those blocks 

This should help beter navigation when searching for a bug to be resolved. Example:
sparkle-annotate-blocks-01


src/components/templates/ScreeningRoom/ScreeningRoom.tsx
```diff
@@ -120,7 +120,12 @@ export const ScreeningRoom: React.FC<ScreeningRoomProps> = ({ venue }) => {
   );
 
   return (
-    <div className="ScreeningRoom">
+    <div
+      data-bem="ScreeningRoom"
+      data-block="ScreeningRoom"
+      data-side="att"
+      className="ScreeningRoom"
```
sunny-viktoryia, line 127: why do you need so many of them? I thought we agreed data-bem should be enough for all the cases.. wouldn't `data-bem` serve the purpose of building the component tree?
moreover, I doubt className is used here.

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
```diff
@@ -62,50 +60,51 @@ export const AllSectionPreviews: React.FC<SectionPreviewsProps> = ({
...
-          dataLength={auditoriumSections.length}
-          className={`AllSectionPreviews ${containerClasses}`}
-          next={loadMore}
-          hasMore={
-            venue.sectionsCount
-              ? auditoriumSections.length < venue.sectionsCount
-              : true
-          }
-          loader={<Loading containerClassName="AllSectionPreviews__loader" />}
+    <VenueWithOverlay venue={venue} containerClassNames="">
```
sunny-viktoryia: [NOTE]
```suggestion
    <VenueWithOverlay venue={venue}>
```

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
```diff
@@ -62,50 +60,51 @@ export const AllSectionPreviews: React.FC<SectionPreviewsProps> = ({
...
+      <InfiniteScroll
+        dataLength={auditoriumSections.length}
+        className={`AllSectionPreviews ${containerClasses}`}
+        next={loadMore}
+        hasMore={
+          venue.sectionsCount
+            ? auditoriumSections.length < venue.sectionsCount
+            : true
+        }
```
sunny-viktoryia: [NOTE]
```suggestion
        hasMore={!venue.sectionsCount || auditoriumSections.length < venue.sectionsCount}
```

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
```diff
@@ -62,50 +60,51 @@ export const AllSectionPreviews: React.FC<SectionPreviewsProps> = ({
...
+              onChange={toggleFullAuditoriums}
+              containerClassName="AllSectionPreviews__toggler"
+              label="Hide full sections"
             />
-          ))}
-        </InfiniteScroll>
-      </VenueWithOverlay>
-    </>
+
+            <ButtonOG onClick={enterRandomSection}>
```
sunny-viktoryia, line 92: [NOTE] ButtonOG ? wow, i see it for the 1st time. 🙈

sunny-viktoryia (approved):
Looks good to me. 🙃

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
goran-peoski-work: You're right about the `className`, I just didn't want to make a blind assumption and didn't want to spend too much time on checking every instance of it.

W.r.t the large number - yes, `data-bem`  would usually be the only one we need.

However, `data-bem="Input"` can be added in two components, we also have two buttons - one attendee side and another admin side etc, hence `data-side`. 

The `data-block` on the other hand should be just one, the top level of those building blocks according to the visual redesign to make it easy to find them all like `document.querySelectorAll('[data-block]')`

src/components/templates/Auditorium/components/AllSectionPreviews/AllSectionPreviews.tsx
goran-peoski-work: This is harder to follow with `||` and `<` priorities 🤔 OK, I'll accept it

src/components/templates/ScreeningRoom/ScreeningRoom.tsx
sunny-viktoryia: ok, I got `data-side` usage, although I couldn't understand what "att" means. But still why do we need `data-block`? as it seems like the copy of `data-bem`.

I don't want us to become obsessed on all the `data-*` attributes now, and then just find out that we ignore them and realize some day that's it's again tons of legacy not-needed code. 
do you often use `document.querySelector` instead of just instecting elements on the html tab?

to me, it seems that data-bem should be pretty enoough.

PR #3023, opened 2022-04-11
Replace old useUser with newer useLiveUser hook (and similar useUser)

goran-peoski-work:
It is a simple replacement `hooks/useUser`  -> `hooks/user/useLiveUser` and a simple copy of `useLiveUser` as `hooks/user/useUser` that depends on `useProfile` instead of `useLiveProfile`


mike-lvov (approved):
LGTM

PR #3024, opened 2022-04-12
Fix/chat blowup

goran-peoski-work:
Fixing private chat messages blow up because of old hooks. Reference in:
- https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1649703838979769
- https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1649704736134679
- https://doesitsparkle.slack.com/archives/C02DPU9KF1A/p1649711367786149

NOTE: the current replacement is implemented via `useFireQuery()` but if live updating version is needed, it can be replaced with `useLiveQuery()`


src/components/organisms/ChatSidebar/ChatSidebar.tsx
```diff
@@ -19,17 +25,33 @@ export const _ChatSidebar: React.FC = () => {
...
   return (
     <div
       role="dialog"
-      aria-labelledby={isVenueChat ? venueTabId : privateTabId}
+      aria-labelledby={isVenueChat ? CHAT_TAB_SPACE_ID : CHAT_TAB_PRIVATE_ID}
       className={styles.chatSidebar}
     >
       {isVenueChat && <VenueChat />}
-      {isPrivateChat && <PrivateChats recipient={recipient} />}
+      {isPrivateChat && userId && (
```
alexxpetrov, line 52: [Question] Can this chat type be something other than `venue` or `private`?
If not we might not need the `isVenueChat` and use a ternary here rendering `PrivateChats` if `isPrivateChat && userId` are `true` and `VenueChat` if it's `false`

alexxpetrov (approved):
LGTM

src/components/organisms/ChatSidebar/ChatSidebar.tsx
goran-peoski-work: Maybe, but for now I tried to just keep the same logic (and only put the `userId` at the higher component for the error check

PR #3025, opened 2022-04-12
Make chat messages live update

goran-peoski-work:
Follow up PR to 
- https://github.com/sparkletown/sparkle/pull/3024
that makes the chat live update


alexxpetrov (approved):
LGTM

PR #3034, opened 2022-04-18
Failed to get image width/height message for map background

goran-peoski-work:
resolves:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1823

With a minor image hooks and error captures refactor, the map background will only print the error text in the browser console (and in the future an extension of the `SparkleError` may funnel to Bugsnag and/or similar systems)


alexxpetrov (approved):
LGTM

PR #3036, opened 2022-04-18
fix dropdown touchable area

denisdimitrov:
The dropdown component's touchable area is not the whole option row but the text instead. This now fixes the issue and allows the dropdown options to be passed as an array.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1842


alexxpetrov:
@denisdimitrov I think I had this solved in another, not so good looking, manner while encountering this issue with `SpacesDropdown` component. So if you have time to fix conflicts in this PR and re-check dropdowns that we use throughout the app, that would be great :)

src/components/templates/ExperimentalSpace/ExperimentalSpace.tsx
```diff
@@ -200,7 +195,9 @@ const _ExperimentalSpace: React.FC<ExperimentalSpaceProps> = ({
         {!inHuddle && <p onClick={connectCallback}>Connect</p>}
         {inHuddle && <p onClick={shareScreenCallback}>Share screen</p>}
       </div>
-      <Dropdown title={projectTitle}>{trackOptions}</Dropdown>
+      <Dropdown options={[]} title="Choose a track to project">
```
sunny-viktoryia, line 198: maybe we can make options default then? and add a default value in the Dropdown component?

src/components/atoms/SortDropDown/SortDropDown.tsx
```diff
@@ -13,19 +13,29 @@ interface SortDropDownProps {
...
-      ))}
-    </Dropdown>
-  </div>
-);
+}) => {
+  const sortingOptions = Object.values(SortingOptions).map((sortingOption) => ({
+    label: sortingOption,
+  }));
+
+  const renderSortingOption = (option: { label: SortingOptions }) => (
```
sunny-viktoryia: useCallback?

src/components/admin/UserStatusDropdown.tsx
```diff
@@ -29,21 +29,22 @@ export const UserStatusDropdown: React.FC<UserStatusDropdownProps> = ({
...
-          className="UserStatusDropdown__item"
-          key={`${index}-${userStatus.status}`}
-          onClick={() => changeUserStatus(userStatus.status)}
-          data-dropdown-value={userStatus.status}
-        >
-          {userStatus.status}
-        </div>
-      )),
-    [userStatuses, changeUserStatus]
+  const renderUserStatusOption = (userStatus: UserStatus) => (
```
sunny-viktoryia: useCallback?

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
```diff
@@ -69,75 +77,45 @@ export const SpacesDropdown: React.FC<SpacesDropdownProps> = ({
...
     <>
       <div data-bem="SpacesDropdown" className="mb-2">
         <label className="block text-sm font-medium text-gray-700 mb-1">
           Link portal to a space
         </label>
-        <Dropdown title={renderedTitle}>{renderedOptions}</Dropdown>
+        <Dropdown
+          title={selected?.name || noneOptionName}
+          options={spaceOptions}
+          onSelect={(space) => select(space)}
```
sunny-viktoryia, line 120: will that work?
```suggestion
          onSelect={select}
```

src/components/admin/Dropdown/Dropdown.tsx
```diff
@@ -125,12 +76,38 @@ export const Dropdown: React.FC<DropdownProps> = ({
...
               <li
-                key={`${index}-${option.value}`}
+                key={`${index}-${option.label}`}
                 className={listItem}
-                onClick={() => selectOption(option)}
+                onClick={() => {
+                  setSelectedOption(option);
+                  setOpened(false);
+                  onSelect?.(option);
+                }}
```
sunny-viktoryia: not to rereder it, can we use useCallback here?

sunny-viktoryia (approved):
Approving with some comments.
@denisdimitrov if you have time, please, take a look and fix, otherwise, feel free to merge as is.

src/components/templates/ExperimentalSpace/ExperimentalSpace.tsx
denisdimitrov: In my opinion, there is no need for it. This particular case is not going to show anywhere, this experimental space was developed by Colin for testing purposes. I didn't really know what the trackOptions are, so I just added empty array to avoid the compile error for the component. But it should definitely always accept options prop, which should be required.

sunny-viktoryia: ok, can we use an ALWAYS_EMPTY_ARRAY then not to create a new object on each rerender?

src/components/admin/SpacesDropdown/SpacesDropdown.tsx
denisdimitrov: Yes, it will pass the parameters automatically

PR #3038, opened 2022-04-20
Snagging: Chat

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1817

image


src/components/attendee/Input/Input.tsx
```diff
@@ -69,6 +102,8 @@ export const Input: React.ForwardRefRenderFunction<
           onKeyDown={handleKeyPress}
         />
 
+        {icon && renderIcon(icon, onIconClick)}
```
denisdimitrov, line 108: [NB] Some of the designs I've seen have this icon before the input on the left side, is there an easy way to do this or should we mark it OOS?

src/components/attendee/ChatContainer.module.scss
```diff
@@ -100,11 +100,15 @@
   margin-bottom: space.empty(3);
 
   @media (min-width: breakpoint.$md) {
-    min-height: 90vh;
+    min-height: calc(100vh - 69px);
```
denisdimitrov: [NB] Can we use a constant for this 69px value? Is it the height of the footer or something else?

denisdimitrov (approved):
LGTM!

src/components/attendee/Input/Input.tsx
sunny-viktoryia: I'd leave it for the next PR. 🙃

src/components/attendee/ChatContainer.module.scss
sunny-viktoryia: added more meaning to the calculations here with the comment: https://github.com/sparkletown/sparkle/pull/3038/commits/ac1c1c2c500155a189e76ade05363d5799e129be

PR #3039, opened 2022-04-20
Booths

colinhowe:
OoOoOo
image

### DIFF
--- a/src/api/presence.ts
+++ b/src/api/presence.ts
@@ -1,5 +1,5 @@
 import * as fs from "firebase/firestore";
-import { debounce, omit, uniqBy } from "lodash";
+import { debounce, groupBy, mapValues, omit, uniqBy } from "lodash";
 
 import {
   COLLECTION_USER_PRESENCE,
@@ -82,7 +82,7 @@ interface subscribeToCheckInsOptions {
   spaceIds: SpaceId[];
   limit?: number;
   debounceInterval?: number;
-  callback: (docs: UserPresenceDocument[]) => void;
+  callback: (docs: { [spaceId: SpaceId]: UserPresenceDocument[] }) => void;
 }
 
 export const subscribeToCheckIns: (
@@ -107,6 +107,10 @@ export const subscribeToCheckIns: (
     spaceIds.splice(10);
   }
 
+  if (!spaceIds.length) {
+    return () => {};
+  }
+
   const query = fs.query(
     collection,
     fs.where(FIELD_SPACE_ID, "in", spaceIds),
@@ -116,8 +120,11 @@ export const subscribeToCheckIns: (
 
   const onNext = debounce((snap: fs.QuerySnapshot<UserPresenceDocument>) => {
     const users = snap.docs.map((doc) => doc.data());
-    const dedupedUsers = uniqBy(users, ({ userId }) => userId);
-    callback(dedupedUsers);
+    const uniqueUsersBySpace = mapValues(
+      groupBy(users, ({ spaceId }) => spaceId),
+      (users) => uniqBy(users, ({ userId }) => userId)
+    );
+    callback(uniqueUsersBySpace);
   }, debounceInterval);
 
   const onError = (err: fs.FirestoreError) => {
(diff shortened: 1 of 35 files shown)


src/api/presence.ts
```diff
@@ -100,17 +100,31 @@ export const subscribeToCheckIns: (
     queryOptions.push(fs.limit(limit));
   }
 
+  if (spaceIds.length > 10) {
+    console.error(
+      "Too many space IDs provided. Firebase limits to 10 items in an IN query. Truncating"
+    );
+    spaceIds.splice(10);
+  }
```
alexxpetrov, line 108: [Question] I think we had an option to batch requests if the total number is larger than the max set by firebase. But I suppose we won't need it in this case?

src/components/admin/SpaceEditForm.tsx
```diff
@@ -110,6 +112,10 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
       showContent: space.showContent ?? DEFAULT_SHOW_CONTENT,
       backgroundImage: undefined,
       backgroundImageUrl: space.backgroundImageUrl,
+      boothsEnabled: space.boothsEnabled,
+      maxBooths: space.maxBooths || 1,
```
alexxpetrov, line 116: let's replace this `1` with a constant from `settings` ? MIN_MAX_BOOTHS

src/components/admin/SpaceEditForm.tsx
```diff
@@ -202,6 +211,22 @@ export const SpaceEditForm: React.FC<SpaceEditFormProps> = ({
...
+      Object.fromEntries(
+        relatedVenues
+          .filter(
+            ({ id, worldId, template, managedBy }) =>
+              space.worldId === worldId &&
+              id !== space.id &&
+              template === VenueTemplate.meetingroom &&
+              managedBy === undefined
+          )
+          .map((venue) => [venue.id, venue])
```
alexxpetrov, line 225: let's combine this `filter` and `map` into a `reduce` and create a separate function for the equality checks?
We'll avoid going through the array twice in this case

src/components/attendee/BoothGrid/BoothCard.module.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.contents {
+  padding: space.gap(6) space.gap(12);
+}
+
+.title {
+  display: block;
+  margin-bottom: space.gap(12);
+  @include font.xlarge;
+  border-bottom: 1px solid rgba(130, 130, 130, 0.3);
```
alexxpetrov: let's use `border.$width--slim` instead of this 1px and replace rgba with `theme.$sparkle--border`

src/pages/SpacesDashboard/SpacesDashboard.tsx
```diff
@@ -32,9 +32,12 @@ export const SpacesDashboard: React.FC<SpacesDashboardProps> = ({
   const isWorldAdmin = userId ? world?.owners.includes(userId) : undefined;
   const { spaces } = useWorldSpaces({ worldId: world.id });
 
-  const renderedMapCards: JSX.Element[] | undefined = useMemo(
-    () =>
-      spaces?.filter(isPartyMapVenue).map((space) => {
+  const managedSpaces = spaces.filter(({ managedBy }) => !!managedBy);
+  const unmanagedSpaces = spaces.filter(({ managedBy }) => !managedBy);
```
alexxpetrov, line 36: Hm, maybe we can replace this with a reduce that has 2 props in an object? Something like:
`{managedSpaces: [], unmanagedSpaces: []}` and then push into corresponding one based on the `managedBy` value?
Again, we'd avoid going through the space array twice this way

alexxpetrov (approved):
LGTM, with a couple of things to improve

src/api/presence.ts
colinhowe: We could do. We'd have to loop over them and setup multiple subscriptions. Might end up getting messy/slow. Matt is happy with the limit of 10 for now :)

PR #3041, opened 2022-04-20
Remove animate map

colinhowe:
Not wanted in the new design. If it is, we can resurrect from this commit


goran-peoski-work (approved):
LGTM

PR #3042, opened 2022-04-21
improve user management modals UI

denisdimitrov:
Some modifications and improvements for the invite admin and edit admin modals in the user management screen.


src/components/admin/WorldUserCard/WorldUserCard.tsx
```diff
@@ -16,6 +19,8 @@ export interface WorldUserCardProps {
   userId?: UserId;
 }
 
+const DEFAULT_AVATAR = DEFAULT_AVATAR_LIST?.[0];
```
alexxpetrov: [Question] I think think won't ever be undefined, unless we deleted it
So we can rewrite it like `const [DEFAULT_AVATAR] = DEFAULT_AVATAR_LIST;`

alexxpetrov (approved):
LGTM. 1 suggestion

PR #3057, opened 2022-04-27
Remove sovereign venues

colinhowe:
:fire:


alexxpetrov (approved):
With this change we'll effectively be fixing this thing too https://github.com/sparkletown/internal-sparkle-issues/issues/1830#issuecomment-1109520353
Thanks!
LGTM

PR #3058, opened 2022-04-27
Use inline styles for background images

colinhowe:
For reasons I cannot fathom, useCss is triggering a re-render of the
backgroudn image which is ugly


sunny-viktoryia (approved):

alexxpetrov (approved):
LGTM

PR #3059, opened 2022-04-28
Wait for the create venue function to finish before using the result

colinhowe:
Not sure why it wasn't doing this before


alexxpetrov (approved):
LGTM

PR #3065, opened 2022-04-29
Allow for the template to be unspecified for booths

colinhowe:
Fallback to generic arguments


sunny-viktoryia (approved):
Looks good. 🙃

oleksiinearmap (approved):

PR #3071, opened 2022-05-02
Make portals responsive

sunny-viktoryia:
closes https://github.com/sparkletown/internal-sparkle-issues/issues/1813

image
image
image


src/components/attendee/EventInfo/EventInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+  event: WorldEvent;
+  centered?: boolean;
+}
+
+const moreLessSpan = (text: string) => (
+  <span className={CN.moreLessSpan}>{text}</span>
+);
+
+export const EventInfo: React.FC<EventInfoProps> = ({ event, centered }) => (
+  <div data-bem="EventInfo" className={cn(CN.general, centered && CN.centered)}>
```
alexxpetrov: [Question] Can we transform this to
```suggestion
  const generalClassnames = cn(CN.general, {
  [CN.centered]: centered
  })
```
I thought we've agreed on this variant for classnames usage. It gets a little tough to see the resulting classnames when they're calculated inside the markup

src/components/attendee/EventInfo/EventInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <span>{STRING_SPACE + formatTimeLocalised(eventEndTime({ event }))}</span>
+    </div>
+
+    <div className={CN.paragraph}>
+      <ShowMoreText
+        lines={5}
+        anchorClass={CN.showMoreLessAnchor}
+        more={moreLessSpan("More")}
+        less={moreLessSpan("Less")}
+        expanded={false}
```
alexxpetrov: [Question] Do we need to include a falsy property at all? Why don't we provide a default `expanded = false` inside the `ShowMoreText` to avoid unnecessary props?

src/components/attendee/SpaceInfoText/SpaceInfoText.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { RenderMarkdown } from "components/organisms/RenderMarkdown";
+
+import CN from "./SpaceInfoText.module.scss";
+
+const descriptionLines: Record<SpaceInfoTextVariant, number> = {
+  space: 1,
+  mobilePortal: 5,
+  desktopPortal: 5,
+};
```
alexxpetrov, line 17: [Question] Are these values documented somewhere or can we export them from some settings file with comments to what they mean?

src/components/templates/PartyMap/components/Map/Map.module.scss
```diff
@@ -13,12 +13,30 @@
...
+  height: 66vh;
+  box-sizing: border-box;
+  padding-top: 54px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  font-size: 34vw;
+  letter-spacing: -0.07em;
+  line-height: 0.8;
```
alexxpetrov: [Suggestion] These hard coded values look like they would benefit if they were transformed into constants

src/components/templates/PartyMap/components/Map/MapPortal.module.scss
```diff
@@ -11,15 +12,21 @@
 
 $-title-height: 42px;
 
+$-w-max-tablet: calc(#{breakpoint.$md} - 1px);
```
alexxpetrov, line 15: [Question] Is it necessary to subtract that `1px` there? I can see that we have `min-width` a little down below, so wouldn't it mean that anything less than that would automatically mean that these `max-width` styles would be applied?

alexxpetrov (commented):
A couple suggestions and questions. Other than that - great job!

src/components/attendee/SpaceInfoText/SpaceInfoText.tsx
sunny-viktoryia: no, it's not documented. It's the number of lines showed by default if the text is long. If it's longer than xx lines than it's expandable.

I chose the default number of lines for description. Later we'll limit it in admin by character limit.
image

and in space (according to the design) there's just 1 line showed by default.
image

src/components/templates/PartyMap/components/Map/MapPortal.module.scss
sunny-viktoryia:
```
breakpoint.$md: 768px;
```
so we have 2 intervals:
[...; 768) -- excluding 768 ---------- mobile ---------- max-width: 768 - 1;
and [768; ...) -- including 768 ---------- desktop ---------- min-width: 768;

src/components/attendee/EventInfo/EventInfo.tsx
sunny-viktoryia: done: https://github.com/sparkletown/sparkle/pull/3071/commits/53ad693cb533a0eaa1fddc004d2ab374630c2be1

src/components/attendee/EventInfo/EventInfo.tsx
sunny-viktoryia: yes, good catch, thanks. done: https://github.com/sparkletown/sparkle/pull/3071/commits/53ad693cb533a0eaa1fddc004d2ab374630c2be1

src/components/templates/PartyMap/components/Map/Map.module.scss
sunny-viktoryia: moved to variables: https://github.com/sparkletown/sparkle/pull/3071/commits/53ad693cb533a0eaa1fddc004d2ab374630c2be1

alexxpetrov (approved):
LGTM

PR #3072, opened 2022-05-03
remove boothTemplateSpaceId null value from the functions

denisdimitrov:
- Remove unnecessary `null` check from the functions to prevent `null` being stored as a value for a field that's exclusively `string`.
- `.nullable()` has been added to the form for backwards compatibility to ignore the null value for documents that have it already stored in the database.

Fixes: https://github.com/sparkletown/internal-sparkle-issues/issues/1857

### DIFF
--- a/src/forms/spaceEditSchema.ts
+++ b/src/forms/spaceEditSchema.ts
@@ -48,5 +48,6 @@ export const spaceEditSchema = Yup.object().shape({
         : schema.notRequired()
   ),
   maxBooths: Yup.number().required(),
-  boothTemplateSpaceId: Yup.string().notRequired(),
+  // .nullable() is added for backwards compatibility after 'null' value was wrongly stored in some documents
+  boothTemplateSpaceId: Yup.string().notRequired().nullable(),
 });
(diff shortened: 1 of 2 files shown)


src/forms/spaceEditSchema.ts
```diff
@@ -48,5 +48,6 @@ export const spaceEditSchema = Yup.object().shape({
         : schema.notRequired()
   ),
   maxBooths: Yup.number().required(),
-  boothTemplateSpaceId: Yup.string().notRequired(),
+  // .nullable() is added for backwards compatibility after 'null' value was wrongly stored in some documents
```
alexxpetrov: [Question] Are there a lot of these? Maybe we can modify this value manually/via migration?

alexxpetrov (approved):
LGTM. But consider a migration/manual change in the firestore DB

src/forms/spaceEditSchema.ts
denisdimitrov: > Are there a lot of these?


I don't know. Firebase can't filter a field by `null` value from the GUI. We can run a script in all of the environments to change the `null` to `""` which would work fine IMO.

denisdimitrov: Firebase seems to be very unforgiving when it comes to `null` values and we should avoid them as much as possible.

denisdimitrov: There were only 5 documents in staging and sparkle 7. I changed them manually and this fix should be enough.

PR #3082, opened 2022-05-10
Fix creation of booth with no template specified

sunny-viktoryia:
Allows creation of booth when no template specified


alexxpetrov (approved):
LGTM

PR #3083, opened 2022-05-12
Snagging Webcam huddle

alexxpetrov:
Closes:
- https://github.com/sparkletown/internal-sparkle-issues/issues/1841


src/components/attendee/VideoHuddle/VideoHuddle.module.scss
```diff
@@ -35,28 +36,31 @@ $-control-size: 32px;
 
 .VideoHuddleControls {
   @include sparkle.panel--normal;
+  width: 5%;
```
sunny-viktoryia: why 5% ? in figma we have width in px, where did you get 5 and 10 %?
can't you specify just min-width / max-width for it to avoid the situation like on the screenshot?
image

src/components/attendee/VideoHuddle/VideoHuddle.module.scss
```diff
@@ -35,28 +36,31 @@ $-control-size: 32px;
...
   justify-self: start;
   border-left: 1px solid theme.$sparkle--border;
 
-  span {
-    border-bottom: 1px solid theme.$sparkle--border;
-    width: $-control-size;
-    height: $-control-size;
-    opacity: 0.5;
+  @media (max-width: breakpoint.$lg) {
+    width: 10%;
```
sunny-viktoryia: 10% ? looks a bit hacky.
I don't think we need here a different width depending on the different breakpoints. 
you can also use here a scss variable.

sunny-viktoryia (approved):
would be cool if you could fix the issue with magic percent numbers. Other than that, looks good.

PR #3084, opened 2022-05-12
Fix create Booth button placement in Safari

sunny-viktoryia:
fixes the position of the create booth button in Safari


alexxpetrov (approved):
LGTM

